PR #23326, opened 2019-07-05
hack/lib/constants.sh: remove no_openssl flag

sosiouxme:
Per https://jira.coreos.com/browse/ART-666 4.2 should no longer be disabling openssl linking.

@stevekuznetsov @smarterclayton I am guessing you will probably not want to do this until the latest golang-builder is being used in CI so that dynamic linking of openssl fails over as intended to static linked crypto. That builder is ready when you are.


sosiouxme:
@jwforres has indicated that we should go ahead with this even before the OCP-based golang builders are unleashed.

@smarterclayton @soltysh @eparis can i get a LGTM here?

sosiouxme:
bump - still needs approval/lgtm

PR #23281, opened 2019-06-26
 depend on master openshift-apiserver

deads2k:
This will let us clean up our temporary branch.

PR #23327, opened 2019-07-06
Set glide alias for goautoneg package

cblecker:
Set a glide repo alias for the `bitbucket.org/ww/goautoneg` package due to https://github.com/Masterminds/glide/issues/1057.

Upstream previously did this in https://github.com/kubernetes/kubernetes/pull/72138.

Modelled this PR after the latest dependency bump PR (https://github.com/openshift/origin/pull/23281).


cblecker:
Tests passing. This one is ready for review.

glide.yaml
```diff
@@ -264,3 +264,9 @@ import:
 - package: gonum.org/v1/gonum
   repo: https://github.com/gonum/gonum.git
   version: cebdade430ccb61c1feba4878085f6cf8cb3320e
+
+# set alias location for this package due to
+# https://github.com/Masterminds/glide/issues/1057; upstream did this in
+# https://github.com/kubernetes/kubernetes/pull/72138
+- package: bitbucket.org/ww/goautoneg
+  repo: https://github.com/munnerz/goautoneg.git
```
tnozicka: is that what upstream did? 

this does the trick without changing the repo location https://github.com/openshift/library-go/pull/469/files#diff-395f41b2a27b70ce44399c91c82158c2R68

cblecker: We don't run glide in upstream (went straight from Godeps to go modules) -- we just straight up replaced the dependency with the forked copy. We were also seeing flakiness in bitbucket for CI (but obviously at a much higher PR rate).

We already import the forked dependency too, because of the upstream Kubernetes changes. I think either change is acceptable with a slight preference to move to the github fork to align with upstream.

Let me know if you'd like me to update my PR, @tnozicka :)

cblecker: @tnozicka I confirmed that locally, this fixes openshift/origin for me as well. Happy to go with either option.

cblecker:
ping @tnozicka .. looking for feedback. I'm okay with either fix, but am blocked on getting dependency bumps into origin.

glide.yaml
tnozicka: given upstream switched to that fork, bot options look ok to me

tnozicka: /unassign
/assign @deads2k

PR #23224, opened 2019-06-20
automate OLM test case OCP-24074

jianzhangbjz:
As title, automate [test case OCP-24074](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-24074).
@ecordell @njhale Could you help take a look at it? And, could you give approve if it looks good to you? Thanks!
cc @bandrade @scolange @cuipinghuo @emmajiafan @chengzhang1016 @zihantang-rh 
Logs as below:
```console
mac:origin jianzhang$ openshift-tests run all --dry-run | grep "\[Feature:Platform\] OLM should"  | openshift-tests run --junit-dir=./ -f -
I0620 18:05:02.132668   73378 trace.go:81] Trace[939984059]: "Reflector k8s.io/client-go/informers/factory.go:133 ListAndWatch" (started: 2019-06-20 18:04:51.05113 +0800 CST m=+0.645920637) (total time: 11.08123279s):
Trace[939984059]: [11.081046228s] [11.081046228s] Objects listed
started: (0/1/12) "[Feature:Platform] OLM should check API version: operatorgroups [Suite:openshift/conformance/parallel]"

started: (0/2/12) "[Feature:Platform] OLM should check API version: catalogsources [Suite:openshift/conformance/parallel]"

started: (0/3/12) "[Feature:Platform] OLM should list catalogsources [Suite:openshift/conformance/parallel]"

started: (0/4/12) "[Feature:Platform] OLM should check API version: subscriptions [Suite:openshift/conformance/parallel]"

started: (0/5/12) "[Feature:Platform] OLM should list subscriptions [Suite:openshift/conformance/parallel]"

started: (0/6/12) "[Feature:Platform] OLM should check API version: packagemanifests [Suite:openshift/conformance/parallel]"

started: (0/7/12) "[Feature:Platform] OLM should list clusterserviceversions [Suite:openshift/conformance/parallel]"

started: (0/8/12) "[Feature:Platform] OLM should list packagemanifests [Suite:openshift/conformance/parallel]"

started: (0/9/12) "[Feature:Platform] OLM should list operatorgroups [Suite:openshift/conformance/parallel]"

started: (0/10/12) "[Feature:Platform] OLM should check API version: installplans [Suite:openshift/conformance/parallel]"

passed: (28.2s) 2019-06-20T10:05:31 "[Feature:Platform] OLM should check API version: installplans [Suite:openshift/conformance/parallel]"

[...]

PR #23329, opened 2019-07-08
automate OLM test case OCP-24074: check OLM resources version

jianzhangbjz:
Since some changes on 4.2, but not 4.1. So I think we can add automation test cases here. @bparees @njhale Could you help have a review? Thanks!
@stevekuznetsov @vikaslaad For 4.1.z automation, the branch `release-4.1` is the right place, right? I guess 4.1.z Errata conformance test will use this branch, right?

### DIFF
--- a/test/extended/operators/olm.go
+++ b/test/extended/operators/olm.go
@@ -1,12 +1,13 @@
 package operators
 
 import (
-	"reflect"
-
+	"fmt"
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 	yaml "gopkg.in/yaml.v2"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"reflect"
+	"strings"
 
 	exutil "github.com/openshift/origin/test/extended/util"
 )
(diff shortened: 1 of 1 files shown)


test/extended/operators/olm.go
```diff
@@ -37,6 +38,52 @@ var _ = g.Describe("[Feature:Platform] OLM should", func() {
...
+		"packagemanifests":       "v1",
+		"catalogsources":         "v1alpha1",
+		"subscriptions":          "v1alpha1",
+		"installplans":           "v1alpha1",
+		"clusterserviceversions": "v1alpha1",
+	}
+
+	for k, v := range versionMap {
+		g.It("check API version: "+k, func() {
+			output, err := oc.AsAdmin().Run("get").Args(k, "-n", "openshift-operator-lifecycle-manager", "-o=jsonpath={.items[0].apiVersion}").Output()
```
njhale: Even though the earlier test does this, I don't think we can continue to expect an instance of each kind to exist in the `openshift-operator-lifecycle-manager` namespace.

test/extended/operators/olm.go
```diff
@@ -37,6 +38,52 @@ var _ = g.Describe("[Feature:Platform] OLM should", func() {
...
+		})
+		g.It("check OLM's CRDs' version: "+k, func() {
+			var output string
+			var err error
+
+			if k == "packagemanifests" {
+				// packagemanifests is a resource created by Aggregation layer, not CRD.
+				output, err = oc.AsAdmin().Run("get").Args("apiservice", "v1.packages.operators.coreos.com", "-o=jsonpath={.spec.version}").Output()
+			} else {
+				output, err = oc.AsAdmin().Run("get").Args("crd", fmt.Sprintf("%s.operators.coreos.com", k), "-o=jsonpath={.spec.version}").Output()
```
njhale: (see https://github.com/openshift/origin/pull/23224#discussion_r300717900)

stevekuznetsov:
Yep, this is the appropriate branch to make this against.

test/extended/operators/olm.go
```diff
@@ -1,12 +1,13 @@
...
 import (
-	"reflect"
-
+	"fmt"
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 	yaml "gopkg.in/yaml.v2"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"reflect"
+	"strings"
```
eparis: These 2 stdlib imports belong up with "fmt".

eparis:
Where is the master/4.2 PR? I'm trying to connect the dots.

jianzhangbjz:
@eparis Here is that PR for master: #23224. But, it has been closed since PR #23331 covered this feature.

eparis:
if this was covered by 23331 should this be closed? should we backport 23331 instead of 23224 ?

jianzhangbjz:
@eparis Sorry for the late to reply. Yes, I think so.

PR #23336, opened 2019-07-08
make test more conventional to get better messages

deads2k:
This test appears to be flaking.  This makes the test more conventional to get better inline messages and tightens a wait window.


deads2k:
Hide whitespace for a cleaner diff

PR #23343, opened 2019-07-09
auto for case OCP-10284

zhouying7780:
Auto for case: OCP-10284, check add or remove volume from DC works well.

@xingxingxia Please help review. thanks.

### DIFF
--- a/test/extended/cli/set.go
+++ b/test/extended/cli/set.go
@@ -1,0 +1,68 @@
+package cli
+
+import (
+	"time"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	appsv1 "github.com/openshift/api/apps/v1"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const deploymentRunTimeout = 5 * time.Minute
+const deploymentChangeTimeout = 30 * time.Second
+
+var _ = g.Describe("[cli]oc set [Conformance]", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc                      = exutil.NewCLI("oc-set", exutil.KubeConfigPath())
+		simpleDeploymentFixture = exutil.FixturePath("testdata", "deployments", "deployment-simple.yaml")
+		appsLabel               = exutil.ParseLabelsOrDie("name=deployment-simple")
+	)
+
+	g.Describe("volume", func() {
+		dcName := "deployment-simple"
+
+		g.It("should works well with DC", func() {
+			namespace := oc.Namespace()
+			dc := exutil.ReadFixtureOrFail(simpleDeploymentFixture).(*appsv1.DeploymentConfig)
+			o.Expect(dc.Name).To(o.Equal(dcName))
+
+			dc, err := oc.AppsClient().AppsV1().DeploymentConfigs(namespace).Create(dc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(dc.Name).To(o.Equal(dcName))
+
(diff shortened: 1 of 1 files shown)


zhouying7780:
cc @openshift/cli-review

test/extended/cli/set.go
```diff
@@ -0,0 +1,79 @@
...
+const deploymentRunTimeout = 5 * time.Minute
+const deploymentChangeTimeout = 30 * time.Second
+
+var _ = g.Describe("[cli][Slow] oc set", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc                      = exutil.NewCLI("oc-set", exutil.KubeConfigPath())
+		simpleDeploymentFixture = exutil.FixturePath("testdata", "deployments", "deployment-simple.yaml")
+		imageSourceLabel        = exutil.ParseLabelsOrDie("name=deployment-simple")
```
xingxingxia: I searched `imageSourceLabel` appears in test/extended/builds/image_source.go
For var name, please name it so that the var name is close to the test context.

test/extended/cli/set.go
```diff
@@ -0,0 +1,79 @@
...
+	)
+
+	g.Describe("volume", func() {
+		dcName := "deployment-simple"
+
+		g.It("should works well with DC", func() {
+			namespace := oc.Namespace()
+			dc := exutil.ReadFixtureOrFail(simpleDeploymentFixture).(*appsv1.DeploymentConfig)
+			o.Expect(dc.Name).To(o.Equal(dcName))
```
xingxingxia: This 2 lines are covered in test/extended/deployments/deployments.go. I think they are redundant and thus could be removed.

test/extended/cli/set.go
zhouying7780: thanks, updated.

test/extended/cli/set.go
```diff
@@ -0,0 +1,79 @@
...
+
+			_, err = oc.Run("set", "volume").Args("dc/deployment-simple", "--add", "--type=emptyDir", "--mount-path=/opt1", "--name=v1").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			err = exutil.WaitForDeploymentConfig(oc.KubeClient(), oc.AppsClient().AppsV1(), oc.Namespace(), dcName, 2, true, oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			pods, err := exutil.WaitForPods(oc.KubeClient().CoreV1().Pods(oc.Namespace()), appsLabel, exutil.CheckPodIsRunning, 1, 4*time.Minute)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(len(pods)).To(o.Equal(1))
```
xingxingxia: I think this line of `len(pods)` is not the test case's checkpoint and thus is not needed. Same for below

test/extended/cli/set.go
```diff
@@ -0,0 +1,79 @@
...
+
+			for _, pod := range pods {
```
xingxingxia: No matter what the replica number is, checking against only one pod is enough for the test case's checkpoint. Thus need not use the `for`. Same for below

test/extended/cli/set.go
zhouying7780: Deleted.

test/extended/cli/set.go
zhouying7780: Updated

soltysh:
/assign 
I'd rather we eliminate one if introducing them in the other place. So if you are rewriting them to go, that's fine just look through that file I've mentioned and convert all.

zhouying7780:
@soltysh  Very glace to see your feedback, I've confirmed my  pr has been covered , I'll close it, thank you very much!

PR #23347, opened 2019-07-09
Load testPerson schema for ldap tests

mrogers950:
The groupsync test data was failing to load completely since the testPerson schema was missing.


enj:
/approve

?

PR #23350, opened 2019-07-10
auto for case OCP-10643

zhouying7780:
auto for case OCP-10643

@xingxingxia  Please help review , thanks.

### DIFF
--- a/test/extended/deployments/deployments.go
+++ b/test/extended/deployments/deployments.go
@@ -98,6 +98,7 @@ var _ = g.Describe("[Feature:DeploymentConfig] deploymentconfigs", func() {
 		envRefDeploymentFixture         = exutil.FixturePath("testdata", "deployments", "deployment-with-ref-env.yaml")
 		ignoresDeployersFixture         = exutil.FixturePath("testdata", "deployments", "deployment-ignores-deployer.yaml")
 		imageChangeTriggerFixture       = exutil.FixturePath("testdata", "deployments", "deployment-trigger.yaml")
+		noDefinedTriggerFixture         = exutil.FixturePath("testdata", "deployments", "deployment-withouttrigger.yaml")
 	)
 
 	g.Describe("when run iteratively [Conformance]", func() {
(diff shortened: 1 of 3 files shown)


test/extended/deployments/deployments.go
```diff
@@ -1554,4 +1554,42 @@ var _ = g.Describe("[Feature:DeploymentConfig] deploymentconfigs", func() {
...
 		})
 	})
+
+	g.Describe("when no triggers are defined [Conformance]", func() {
+		dcName := "paused"
+		g.AfterEach(func() {
+			failureTrap(oc, dcName, g.CurrentGinkgoTestDescription().Failed)
+		})
+
+		g.It("should manually make deploy", func() {
```
xingxingxia: In upstream contribution it is better to use better syntax. I suggest the verb "deploy" be corrected to a noun "deployment"

test/extended/deployments/deployments.go
```diff
@@ -1554,4 +1554,42 @@ var _ = g.Describe("[Feature:DeploymentConfig] deploymentconfigs", func() {
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(dc.Name).To(o.Equal(dcName))
+
+			g.By("remove all the triggers")
+			_, err = oc.Run("set").Args("triggers", "dc/"+dcName, "--remove-all").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("resume the DC")
+			_, err = oc.Run("rollout").Args("resume", "dc/"+dcName).Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
```
xingxingxia: Can we choose test data other than pausedDeploymentFixture so we can reduce the "resume" operation in this test?

xingxingxia:
The PR generally LGTM from peer review. Did you check no existing code already covers your PR? If yes, then could @ Workloads Dev to review/merge. Thanks

test/extended/deployments/deployments.go
zhouying7780: Updated.

test/extended/deployments/deployments.go
zhouying7780: Sorry, If we updated with other test data , we need to add other steps to check the non-auto deployment. So this one is appropriate。

xingxingxia: OK

test/extended/deployments/deployments.go
```diff
@@ -1554,4 +1554,42 @@ var _ = g.Describe("[Feature:DeploymentConfig] deploymentconfigs", func() {
...
+			failureTrap(oc, dcName, g.CurrentGinkgoTestDescription().Failed)
+		})
+
+		g.It("should manually make deployment", func() {
+			dc, err := createDeploymentConfig(oc, pausedDeploymentFixture)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(dc.Name).To(o.Equal(dcName))
+
+			g.By("remove all the triggers")
+			_, err = oc.Run("set").Args("triggers", "dc/"+dcName, "--remove-all").Output()
```
tnozicka: you can just create a fixture with `triggers: []`

tnozicka: prefer using api call unless you'd be explicitly testing cli

test/extended/deployments/deployments.go
```diff
@@ -1554,4 +1554,42 @@ var _ = g.Describe("[Feature:DeploymentConfig] deploymentconfigs", func() {
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("resume the DC")
+			_, err = oc.Run("rollout").Args("resume", "dc/"+dcName).Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("check the DC")
```
tnozicka: try being more descriptive to what property you are checking

xingxingxia: Good point :+1: , in **upstream** code contribution, we should take stricter care of sentences, syntax etc.

test/extended/deployments/deployments.go
zhouying7780: Thanks , updated.

zhouying7780:
For the confilicts will close this pr and open a new one.

PR #23355, opened 2019-07-10
[WIP] Run cmd tests in a pod against an e2e-aws cluster

stlaz:
The cmd-tests are keeping the `oc` dependency real
for the origin repository, let's move them into a
container and run them as a separate e2e test suite.

---
Currently not working, merely a concept


hack/test-cmd.sh
```diff
@@ -1,143 +0,0 @@
-#!/usr/bin/env bash
```
deads2k: keep this file so you don't break our current CI.  You can change it's content so that you have a clean spot to start the new tests after the clsuter comes up.

test/extended/cmd/cmd.go
```diff
@@ -0,0 +1,118 @@
...
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const (
+	testsConfigMapName = "tests"
+)
+
+var _ = g.Describe("[test-cmd] test-cmd:", func() {
```
deads2k: tag serial too.

test/extended/testdata/cmd/lib/build/environment.sh
```diff
@@ -0,0 +1,281 @@
+#!/usr/bin/env bash
```
deads2k: why have these been copied?

stlaz: to keep the `test-cmd.sh` working, perhaps not all the files here are necessary

stlaz:
Alright, this now ALMOST works, two issues though:
- the tests themselves expect initialized `hack/lib` so I don't see us getting rid of it (unless @deads2k thinks otherwise), we should not need `hack/local-up-master` and perhaps `hack/test-lib` is also unnecessary
- the `cli` image does not contain the `bc` library, which makes the logs unusable

There also seem to be some more issues with symlinks due to the nature of how the directory with tests is mounted, I'll see what I can do about that.

stlaz:
Removed previously mentioned `bc` dependency by using bash binary expressions instead.

stlaz:
I symlinked the `hack/lib` dir from the tests directory to avoid duplicities as bindata seems to hate symlinks, but this is starting to get real ugly

hack/lib
```diff
@@ -0,0 +1 @@
+../test/extended/testdata/cmd/hack/lib
```
deads2k: ok, I see

deads2k: > ok, I see

Pull out only the bits you need.  YOu cannot just copy things like rpm building.

test/extended/testdata/cmd/hack/lib/cmd.sh
```diff
@@ -210,7 +210,7 @@ function os::cmd::internal::expect_exit_code_run_grep() {
 	local test_succeeded=$( ${test_eval_func} "${test_result}"; echo $? )
 
 	local end_time=$(os::cmd::internal::seconds_since_epoch)
-	local time_elapsed=$(echo "scale=3; ${end_time} - ${start_time}" | bc | xargs printf '%5.3f') # in decimal seconds, we need leading zeroes for parsing later
+	local time_elapsed=$(echo $(( ${end_time} - ${start_time} )) | xargs printf '%5.3f') # in decimal seconds, we need leading zeroes for parsing later
```
deads2k: why has this changed?

stlaz: The `cli` image does not contain the `bc` binary `yum install -y bc` wouldn't work because it required subscription, this was the easiest way for me to work around that.

hack/lib
stlaz: It feels more feasible to have just one symlink to the whole lib rather than linking specific deps.

I myself thought that the `hack/lib/build` would not be required but there are dependecies in `hack/lib/` and even other `hack/lib` scripts, also the `basicresources.sh` test depends on `hack/lib/build/version.sh`.

I could perhaps try to cut the dependencies inside the library itself to achieve some kind of isolation that probably should have been there from the start, not sure that it is worth cutting the symlink, though.

stlaz:
/close
This got superseeded by https://github.com/openshift/origin/pull/23453

PR #23359, opened 2019-07-11
UPSTREAM: 70286: Use pvc uid instead of pvc name in glusterfs endpoint name

humblec:
Bug # https://bugzilla.redhat.com/show_bug.cgi?id=1714609

Signed-off-by: Humble Chirammal <hchiramm@redhat.com>


jsafrane:
Why this targets 3.10 if the bug wants 3.11.z?

humblec:
> Why this targets 3.10 if the bug wants 3.11.z?

@jsafrane the flags are not adjusted against 3.10 yet and it will be soon. However I would like to point out that the ask is on 3.10 as in https://bugzilla.redhat.com/show_bug.cgi?id=1714609#c6

PR #23360, opened 2019-07-11
move router related testdata to router folder

lihongan:
Moving router related testdata to router folder, it should be clearer to add more cases in future.
@openshift/openshift-team-network-edge Please help review. Thanks.


ironcladlou:
Nice little cleanup, thanks!

/lgtm

PR #23362, opened 2019-07-11
Fix APIServer CRD admission

sttts:
The admission stack was never calling the `Validate` method of the embedded plugin.

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validation_wrapper.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validation_wrapper.go
@@ -25,7 +25,7 @@ func Register(plugins *admission.Plugins) {
 }
 
 type validateCustomResourceWithClient struct {
-	admission.Interface
+	admission.ValidationInterface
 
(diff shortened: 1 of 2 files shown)


vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validation_wrapper.go
```diff
@@ -25,7 +25,7 @@ func Register(plugins *admission.Plugins) {
 }
 
 type validateCustomResourceWithClient struct {
-	admission.Interface
+	admission.ValidationInterface
```
deads2k: I think we need both.

deads2k:
your commit name needs to match the upstream pattern.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validation_wrapper.go
sttts: clarified on Slaxck.

sttts:
rpm build issues.

/retest

sttts:
Cluster doesn't come up. Do we already create broken APIServer objects?

PR #23370, opened 2019-07-12
Bug 1711538: Reenable 'should be rejected when no endpoints exist' test

rcarrillocruz:


rcarrillocruz:
do not review yet, testing the waters in CI as to possible failures.

PR #23372, opened 2019-07-12
move Image change build triggers

deads2k:
moves more integration tests.


stlaz:
/lgtm
looks more or less like what was in the integration suite before and given the CI is green, it apparently works

PR #23373, opened 2019-07-13
openshift-sdn: skip OPENSHIFT-MASQ for traffic already marked.

squeed:
If a packet has already been marked by other kube-proxy rules for masquerade, don't run it through the OPENSHIFT-MASQUERADE chain for further twiddling.

Most notably, this chain is used for Egress IPs.

This change fixes a bug where egress IPs can't access services via their ExternalIP. (bz1726045)

### DIFF
--- a/staging/src/github.com/openshift/sdn/pkg/network/node/iptables.go
+++ b/staging/src/github.com/openshift/sdn/pkg/network/node/iptables.go
@@ -22,19 +22,21 @@ type NodeIPTables struct {
 	syncPeriod         time.Duration
 	masqueradeServices bool
 	vxlanPort          uint32
+	masqueradeBitHex   string // the masquerade bit as hex value
 
 	mu sync.Mutex // Protects concurrent access to syncIPTableRules()
 
 	egressIPs map[string]string
 }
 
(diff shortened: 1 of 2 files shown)


squeed:
cc @danwinship

danwinship:
So this implies that connections to an ExternalIP from a namespace using an egress IP will not use the egress IP. Is that not fixable?

squeed:
> So this implies that connections to an ExternalIP from a namespace using an egress IP will not use the egress IP. Is that not fixable?

Basically, yes. Connections from pods to external IPs are always masqueraded (as opposed to static SNAT). That prevents the RPF issues, as the source IP is chosen after the routing decision.

Even after disabling rpf filtering, it was still being dropped by the conntrack INVALID filter, since it too doesn't like asymmetric routing.

So it does leave us with a bit of an odd situation, where the source IP is the tun0 address of the node holding the egress IP. The only other way I could think to do this would be to program every external IP in to the flow rules to skip egress IP forwarding. If you think that's a better choice, we could do that.

danwinship:
staging sdn is imminently about to be deleted and even if this merges first I think there still might be a race condition with the staging bot... keep an eye on it and be ready to re-file

squeed:
Moved to https://github.com/openshift/sdn/pull/13

PR #23379, opened 2019-07-15
scrub more integration tests

deads2k:
Moves more integration tests, fixes use of constants to avoid bad dependencies, removes integration tests that are already unit tests.


deads2k:
two test failures not from this pull.

/retest

deads2k:
Terraform

/retest

PR #23381, opened 2019-07-15
Add repo details for disconnected install/upgrade

mtnbikenc:
When performing a release mirror, provide details of repos where images
have been mirrored from to be used for disconnected install or upgrade.

https://jira.coreos.com/browse/CORS-1105

Command:
```
oc adm release mirror --from=registry.svc.ci.openshift.org/ocp/release:4.2 --to=registry.svc.ci.openshift.org/rteague-1/mirror-test
```

Example output:
```
To use the new mirrored release-image to install, provide the following list of release-image sources to the installer using the install-config.yaml:

imageContentSources:
- sources:
  - registry.svc.ci.openshift.org/rteague-1/mirror-test
  - registry.svc.ci.openshift.org/ocp/release
  - quay.io/openshift-release-dev/ocp-v4.0-art-dev
```

### DIFF
--- a/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
+++ b/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
@@ -513,16 +513,3 @@ func printInstallInstructions(out io.Writer, from, to string, repos map[string]s
 
 	return nil
 }
-
-func sourceImageRef(is *imagev1.ImageStream, name string) (string, bool) {
-	for _, tag := range is.Spec.Tags {
-		if tag.Name != name {
-			continue
-		}
-		if tag.From == nil || tag.From.Kind != "DockerImage" {
-			return "", false
-		}
-		return tag.From.Name, true
-	}
-	return "", false
-}
--- a/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
+++ b/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
@@ -469,7 +469,7 @@ func (o *MirrorOptions) Run() error {
 //    - registry.svc.ci.openshift.org/ocp/release       # source listed second
 //    - quay.io/openshift-release-dev/ocp-v4.0-art-dev  # repos from image-references
 //
-func printInstallInstructions(out io.Writer, from, to string, repos map[string]struct{}) {
+func printInstallInstructions(out io.Writer, from, to string, repos map[string]struct{}) error {
 	type installConfigSubsectionImageContentSources struct {
 		Sources []string `json:"sources"`
 	}
@@ -480,12 +480,18 @@ func printInstallInstructions(out io.Writer, from, to string, repos map[string]s
 	var repoList []string
 
 	// Append destination (mirror) repo first
-	mirrorRef, _ := imagereference.Parse(to)
+	mirrorRef, err := imagereference.Parse(to)
+	if err != nil {
+		return fmt.Errorf("Unable to parse '--to' image reference: %v", err)
+	}
 	mirrorRepo := mirrorRef.AsRepository().String()
 	repoList = append(repoList, mirrorRepo)
 
 	// Append source repo second
-	sourceRef, _ := imagereference.Parse(from)
+	sourceRef, err := imagereference.Parse(from)
+	if err != nil {
+		return fmt.Errorf("Unable to parse '--from' image reference: %v", err)
+	}
 	sourceRepo := sourceRef.AsRepository().String()
 	repoList = append(repoList, sourceRepo)
 
@@ -497,10 +503,15 @@ func printInstallInstructions(out io.Writer, from, to string, repos map[string]s
 		Sources: repoList}
 	imageContentSource := installConfigSubsection{
 		ImageContentSources: []installConfigSubsectionImageContentSources{sources}}
-	result, _ := yaml.Marshal(imageContentSource)
+	result, err := yaml.Marshal(imageContentSource)
+	if err != nil {
+		return fmt.Errorf("Unable to marshal imageContentSource yaml: %v", err)
+	}
 
 	fmt.Fprintf(out, "\nTo use the new mirrored release-image to install, provide the following list of release-image sources to the installer using the install-config.yaml:\n")
 	fmt.Fprintf(out, string(result))
+
+	return nil
 }
 
 func sourceImageRef(is *imagev1.ImageStream, name string) (string, bool) {
--- a/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
+++ b/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
@@ -11,7 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/ghodss/yaml"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
@@ -23,6 +22,7 @@ import (
 	"k8s.io/client-go/util/retry"
 	kcmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
 	"k8s.io/kubernetes/pkg/kubectl/util/templates"
+	"sigs.k8s.io/yaml"
 
 	imagev1 "github.com/openshift/api/image/v1"
 	imageclient "github.com/openshift/client-go/image/clientset/versioned"
@@ -271,9 +271,7 @@ func (o *MirrorOptions) Run() error {
 		}
 	}
 
-	var currentRepo string
-	exists := struct{}{}
-	repoMap := map[string]struct{}{}
+	repositories := make(map[string]struct{})
 
 	// build the mapping list for mirroring and rewrite if necessary
 	for i := range is.Spec.Tags {
@@ -290,8 +288,8 @@ func (o *MirrorOptions) Run() error {
 		}
 
 		// Create a unique map of repos as keys
-		currentRepo = from.AsRepository().String()
-		repoMap[currentRepo] = exists
+		currentRepo := from.AsRepository().String()
+		repositories[currentRepo] = struct{}{}
 
 		dstMirrorRef := targetFn(tag.Name)
 		mappings = append(mappings, mirror.Mapping{
@@ -453,7 +451,7 @@ func (o *MirrorOptions) Run() error {
 		fmt.Fprintf(o.Out, "\nSuccess\nUpdate image:  %s\nMirrored to: %s\n", to, o.To)
 	}
 
-	printInstallInstructions(o.Out, o.From, o.To, repoMap)
+	printInstallInstructions(o.Out, o.From, o.To, repositories)
 	return nil
 }
 
--- a/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
+++ b/staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"sort"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/ghodss/yaml"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
@@ -269,6 +271,10 @@ func (o *MirrorOptions) Run() error {
 		}
 	}
 
+	var currentRepo string
+	exists := struct{}{}
+	repoMap := map[string]struct{}{}
+
 	// build the mapping list for mirroring and rewrite if necessary
 	for i := range is.Spec.Tags {
 		tag := &is.Spec.Tags[i]
@@ -282,6 +288,10 @@ func (o *MirrorOptions) Run() error {
 		if len(from.Tag) > 0 || len(from.ID) == 0 {
 			return fmt.Errorf("image-references should only contain pointers to images by digest: %s", tag.From.Name)
 		}
+
+		// Create a unique map of repos as keys
+		currentRepo = from.AsRepository().String()
+		repoMap[currentRepo] = exists
 
 		dstMirrorRef := targetFn(tag.Name)
 		mappings = append(mappings, mirror.Mapping{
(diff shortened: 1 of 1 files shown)


abhinavdahiya:
Can you provide the example output for the changeset..

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,6 +452,14 @@ func (o *MirrorOptions) Run() error {
 	} else {
 		fmt.Fprintf(o.Out, "\nSuccess\nUpdate image:  %s\nMirrored to: %s\n", to, o.To)
 	}
+	
+	// Print list of repos mirrored from for use with installing or upgrading when disconnected
```
abhinavdahiya: the goal for the output is to be directly useful for installation and upgrade.

```
To use the new mirrored release-image to install, provide the following list of release-image sources to the installer using the install-config.yaml:
imageContentSource:
- sources:
  - <mirrored repo>
  - <origin release-image repo>
  - <repos referenced in the release-image>
  ...
```

Let's make sure the order is same as the example above..

mtnbikenc: Updated output as requested.  PR description is updated with example.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,6 +451,26 @@ func (o *MirrorOptions) Run() error {
...
+	fmt.Fprintf(o.Out, "\nTo use the new mirrored release-image to install, provide the following list of release-image sources to the installer using the install-config.yaml:\n")
+	fmt.Fprintf(o.Out, `
+imageContentSource:
+- sources:
+  - %s
+  - %s
+`, mirrorRepo, sourceRepo)
+
+	// Print unordered list of repos mirrored from for use with installing or upgrading when disconnected
+	for repo := range repoMap {
```
abhinavdahiya: wouldn't repoMap include the `o.From` already?

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,6 +451,26 @@ func (o *MirrorOptions) Run() error {
...
+	mirrorRef, _ := imagereference.Parse(o.To)
+	mirrorRepo := mirrorRef.AsRepository().String()
+
+	sourceRef, _ := imagereference.Parse(o.From)
+	sourceRepo := sourceRef.AsRepository().String()
+
+	fmt.Fprintf(o.Out, "\nTo use the new mirrored release-image to install, provide the following list of release-image sources to the installer using the install-config.yaml:\n")
+	fmt.Fprintf(o.Out, `
+imageContentSource:
+- sources:
```
abhinavdahiya: Is it better to create a internal struct that maps matches the output and use that to marshal yaml.. string manipulation is always great.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,6 +451,26 @@ func (o *MirrorOptions) Run() error {
 	} else {
 		fmt.Fprintf(o.Out, "\nSuccess\nUpdate image:  %s\nMirrored to: %s\n", to, o.To)
 	}
+	
+	mirrorRef, _ := imagereference.Parse(o.To)
```
abhinavdahiya: WDYT:
```
func printInstallInstuctions(out io.Writer, from, to, otherrepos []string)
```

be move useful..?

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
mtnbikenc: repoMap only includes (unordered) repos listed in image-references, so o.From is not included.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
mtnbikenc: I was not pleased with doing it this way.  I had considered creating a yaml type structure but since we want a specific order I just when with strings to maintain desired order.  Could you provide an example of the internal struct you are suggesting?

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
mtnbikenc: This section of the code could continue to evolve in complexity where it could be clearer to break it out into a function.  Will consider this.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
abhinavdahiya: somethings like this.
```go
type installConfigSubsection struct{
    ImageContentSources []installConfigSubsectionImageContentSource `json:"imageContentSources"`
}
type installConfigSubsectionImageContentSource struct {
    Sources []string `json:"sources"`
}
```

you can still control the order in the Sources list...?

mtnbikenc:
@abhinavdahiya Updated with types to marshal the yaml output and broke out the code into a function.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -283,6 +289,11 @@ func (o *MirrorOptions) Run() error {
 			return fmt.Errorf("image-references should only contain pointers to images by digest: %s", tag.From.Name)
 		}
 
+		// Create a unique map of repos as keys
+		currentRepo = from.AsRepository().String()
+		klog.V(2).Infof("From Repo: %s", currentRepo)
```
mtnbikenc: Removing this.  Was for debug.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,9 +453,57 @@ func (o *MirrorOptions) Run() error {
...
 
+// printInstallInstructions provides detail to the user for using the new mirror for installs.
+// The instructions include yaml formatted output which should be added to the install-config.yaml.
+//
+// Exmaple output:
+//
+//  <descriptive installer usage text>
+//  imageContentSource:
+//  - sources:
+//    - mycompany.com/myrepository/repo					# destination (mirror) listed first
```
mtnbikenc: Converting tabs to spaces for alignment.

mtnbikenc:
Updated to `imageContentSources`

abhinavdahiya:
/cc @smarterclayton 

as you've done most of the `oc adm release mirror`

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,9 +452,57 @@ func (o *MirrorOptions) Run() error {
...
+// The instructions include yaml formatted output which should be added to the install-config.yaml.
+//
+// Exmaple output:
+//
+//  <descriptive installer usage text>
+//  imageContentSources:
+//  - sources:
+//    - mycompany.com/myrepository/repo                 # destination (mirror) listed first
+//    - registry.svc.ci.openshift.org/ocp/release       # source listed second
+//    - quay.io/openshift-release-dev/ocp-v4.0-art-dev  # repos from image-references
```
DanyC97: @mtnbikenc 👋 

my 0.002$: in my humble opinion, would have been nice if the initial design took into account the separation of the source and target repo mirror in the same way `oc adm release mirror` has  `--from=` and `--to`

The yaml file could have had something like

```
imageContentSources:
//  - sources:
//     - registry.svc.ci.openshift.org/ocp/release       # source listed second
//     - quay.io/openshift-release-dev/ocp-v4.0-art-dev  # repos from image-references
//   - targets:
//     - mycompany.com/myrepository/repo                 # destination (mirror) listed first 

```
Doing so  i think everything would have been much more natural and align with the cli - source/target

mtnbikenc: @DanyC97 The context of these instructions is for the installer, not the mirror command.  If thinking about the mirror command, sources and targets make sense.  For the installer, we are providing a list of sources where images could be pulled from.  In this case, they can be pulled from the mirror, the original location or potentially from some other repo which an image in the images-references file pointed to.  In the end, the order of these items makes no difference in the installer code, but we wanted a consistent UX for how these repos were presented.  Have a look at https://github.com/openshift/installer/pull/1910 also.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,9 +452,57 @@ func (o *MirrorOptions) Run() error {
...
+		Sources []string `json:"sources"`
+	}
+	type installConfigSubsection struct{
+		ImageContentSources []installConfigSubsectionImageContentSources `json:"imageContentSources"`
+	}
+
+	var repoList []string
+
+	// Append destination (mirror) repo first
+	mirrorRef, _ := imagereference.Parse(to)
```
abhinavdahiya: should we be handling errors?

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,9 +452,57 @@ func (o *MirrorOptions) Run() error {
...
+	mirrorRepo := mirrorRef.AsRepository().String()
+	repoList = append(repoList, mirrorRepo)
+
+	// Append source repo second
+	sourceRef, _ := imagereference.Parse(from)
```
abhinavdahiya: same as https://github.com/openshift/origin/pull/23381/files#r304103409

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,9 +452,57 @@ func (o *MirrorOptions) Run() error {
...
+
+	// Append unordered repos from image-references in release-image
+	for repo := range repos {
+		repoList = append(repoList, repo)
+	}
+	sources := installConfigSubsectionImageContentSources{
+		Sources: repoList}
+	imageContentSource := installConfigSubsection{
+		ImageContentSources: []installConfigSubsectionImageContentSources{sources}}
+	result, _ := yaml.Marshal(imageContentSource)
```
abhinavdahiya: https://github.com/openshift/origin/pull/23381/files#r304103409

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
mtnbikenc: I had considered it but since we already parsed these refs previously, it would never trigger an error at the end here.  I'll add them for the same of completeness.

mtnbikenc: Updated.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
mtnbikenc: Updated.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
mtnbikenc: Updated.

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
DanyC97: thank you for taking the time to respond @mtnbikenc , i may have misunderstood the comment in the code `destination (mirror) listed first`.

is all clear now, thx

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -442,18 +452,64 @@ func (o *MirrorOptions) Run() error {
...
-			return "", false
-		}
-		return tag.From.Name, true
+// printInstallInstructions provides detail to the user for using the new mirror for installs.
+// The instructions include yaml formatted output which should be added to the install-config.yaml.
+//
+// Exmaple output:
+//
+//  <descriptive installer usage text>
+//  imageContentSources:
```
smarterclayton: Won't this be different between install and upgrade?  I would expect us to print both sets, or to have a consistent set across both.  I.e. how would I create an ICSP if the sources change?

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -269,6 +271,10 @@ func (o *MirrorOptions) Run() error {
 		}
 	}
 
+	var currentRepo string
+	exists := struct{}{}
```
smarterclayton: style: inline this, we don't use this form

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -269,6 +271,10 @@ func (o *MirrorOptions) Run() error {
 		}
 	}
 
+	var currentRepo string
```
smarterclayton: style: inline this variable into the map assignment directly (unless you use it in two places)

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -269,6 +271,10 @@ func (o *MirrorOptions) Run() error {
...
+	exists := struct{}{}
+	repoMap := map[string]struct{}{}
```
smarterclayton: style: always use `repositories := make(map[string]struct{}{})` to init variables, and don't use `Map` in variables (map is implicit in the name usually)

staging/src/github.com/openshift/oc/pkg/cli/admin/release/mirror.go
```diff
@@ -4,12 +4,14 @@ import (
...
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"sort"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/ghodss/yaml"
```
smarterclayton: This should probably be using the k8s yaml utility, not ghodss directly

mtnbikenc:
Resubmitted under https://github.com/openshift/oc/pull/6
/close

PR #23387, opened 2019-07-16
Route to SDN irrespective if macvlan mode is used or not

rcarrillocruz:
Fixes SDN-345

### DIFF
--- a/staging/src/github.com/openshift/sdn/cmd/sdn-cni-plugin/openshift-sdn_linux.go
+++ b/staging/src/github.com/openshift/sdn/cmd/sdn-cni-plugin/openshift-sdn_linux.go
@@ -137,6 +137,10 @@ func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
 	if err != nil {
 		return err
 	}
+	_, serviceIPNet, err := net.ParseCIDR(config.ServiceNetworkCIDR)
+	if err != nil {
+		return fmt.Errorf("failed to parse ServiceNetworkCIDR: %v", err)
+	}
 
 	var hostVeth, contVeth net.Interface
 	err = ns.WithNetNSPath(args.Netns, func(hostNS ns.NetNS) error {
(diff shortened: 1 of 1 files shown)


staging/src/github.com/openshift/sdn/cmd/sdn-cni-plugin/openshift-sdn_linux.go
```diff
@@ -194,6 +194,7 @@ func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
 			return fmt.Errorf("failed to configure container loopback: %v", err)
 		}
 
+		var addrs []netlink.Addr
```
danwinship: hm... I think it would be clearer to move the definition of `dsts` out to here, and leave `addrs` and the loop over it inside the macvlan-specific code

staging/src/github.com/openshift/sdn/cmd/sdn-cni-plugin/openshift-sdn_linux.go
```diff
@@ -222,26 +222,25 @@ func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
...
+			dsts = append(dsts, &net.IPNet{IP: addr.IP, Mask: net.CIDRMask(32, 32)})
+		}
 
-			var dsts []*net.IPNet
-			for _, addr := range addrs {
-				dsts = append(dsts, &net.IPNet{IP: addr.IP, Mask: net.CIDRMask(32, 32)})
-			}
+		_, serviceIPNet, err := net.ParseCIDR(config.ServiceNetworkCIDR)
+		if err != nil {
+			return fmt.Errorf("failed to parse ServiceNetworkCIDR: %v", err)
```
danwinship: Since we always parse `config.ServiceNetworkCIDR` now, how about moving it up to the top, right after `config` is read, so that if there is a problem, we don't bother half-setting-up the pod before we notice it.

staging/src/github.com/openshift/sdn/cmd/sdn-cni-plugin/openshift-sdn_linux.go
```diff
@@ -194,17 +198,21 @@ func (p *cniPlugin) CmdAdd(args *skel.CmdArgs) error {
...
 		// Set up macvlan0 (if it exists)
 		link, err = netlink.LinkByName("macvlan0")
 		if err == nil {
+			var addrs []netlink.Addr
 			err = netlink.LinkSetUp(link)
 			if err != nil {
 				return fmt.Errorf("failed to enable macvlan device: %v", err)
 			}
+			for _, addr := range addrs {
+				dsts = append(dsts, &net.IPNet{IP: addr.IP, Mask: net.CIDRMask(32, 32)})
```
danwinship: er, you can't move this here; `addrs` hasn't been set yet. Just leave the `addrs` code where it was

danwinship:
staging sdn is about to be deleted. Please refile against https://github.com/openshift/sdn
/close

PR #23389, opened 2019-07-16
[3.11] Bug 1719156: reenable TestRegistryClientAPIv1

dmage:


adambkaplan:
/retest

@dmage Given that this test is for schema v1 - which is no longer supported - do we want to just drop this test?

PR #23391, opened 2019-07-16
Fix documentation for cluster_up_down: host-data-dir => base-dir

christf:
The --host-data-dir flag has been renamed to --base-dir. This commit
lets the documentation reflect that.


christf:
well, what is the way forward here?

christf:
/remove-lifecycle rotten

Is there a problem with this change?

PR #23396, opened 2019-07-16
stop publishing oauth-server

deads2k:
publishing change only.  I plan to merge once https://github.com/openshift/oauth-server/pull/1 merges


deads2k:
publishing only.  merging

PR #23401, opened 2019-07-17
stop publishing repos and vendor instead

deads2k:


deads2k:
just publishing files.  merging

PR #23402, opened 2019-07-17
Bug: 1699146: UPSTREAM: 79940: Avoid RbdDiskManager's DetachDisk function never be executed again

jsafrane:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1699146
https://github.com/kubernetes/kubernetes/pull/79940
@openshift/sig-storage


jsafrane:
/approve
I should be owner of all k8s `pkg/volume`.

jsafrane:
hey, https://github.com/openshift/origin/blob/51a61347137a3f9f04e17a18ee365f93a9abdb95/vendor/k8s.io/kubernetes/pkg/volume/OWNERS#L6

PR #23408, opened 2019-07-17
Bug 1701422: Multiple ports for registry service

adambkaplan:
Allow the registry service to have multiple ports for auth.
This will let us drop the port from the internal registry hostname,
while maintaining back-compatibility for older pull specs.


adambkaplan:
/hold

This may have to be re-done against the new openshift-controller-manager repo

adambkaplan:
/close

Moving this to the new ocm repo!

PR #23411, opened 2019-07-17
stop producing oauth-server binary and images

deads2k:
Since https://jira.coreos.com/browse/ART-787 is ready, stop producing the oauth-server image from origin.

/assign @sanchezl


sanchezl:
I don't think you can remove the `Dockerfile.rhel` until you get rid of the CI config that uses it:
https://github.com/openshift/release/blob/cba7d53c382f35a31cf23ce302f5d95842cb7e56/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L32-L39

deads2k:
s3

/retest

tnozicka:
we should remove it from hypershift 
https://github.com/openshift/origin/blob/b21503e65c6bcc724f1c43b2c187a664779b3746/cmd/hypershift/main.go#L74-L78

deads2k:
> we should remove it from hypershift

already gone

PR #23412, opened 2019-07-17
stop building template-service-broker

deads2k:
images should come from openshift/template-service-broker now


deads2k:
terraform

/retest

PR #23414, opened 2019-07-18
images/kube-proxy/Dockerfile.rhel: COPY relative

sosiouxme:
There is no need to refer to the whole repo when only using the contents from the same directory.

Of course you can expect this to fail tests until CI config is adjusted to match.


sosiouxme:
I imagine this is deemed more trouble than it's worth to change. Well I won't push it, it's just a bit silly/wasteful.

PR #23407, opened 2019-07-17
Nuke sdn

tnozicka:
/cc @deads2k @sttts 

/hold
for tests and sync with image promotion in

PR #23420, opened 2019-07-18
switch SDN to be vendored

deads2k:
this completes the arc for moving SDN to openshift/sdn


danwinship:
Do we even need to vendor it? With https://github.com/openshift/sdn/pull/2 merged, the sdn node image should be getting build from that repo, so `origin/images/sdn` (and associated `Makefile`, etc, rules) can go away. And then when https://github.com/openshift/cluster-network-operator/pull/248 merges you can remove "`hypershift openshift-network-controller`" support from `cmd/hypershift/main.go`. And then nothing in origin will refer to openshift/sdn any more.

danwinship:
ah, which is to say, close in favor of #23407 ?

sttts:
> ah, which is to say, close in favor of #23407 ?

This one vendors. I think we still need it for some tests.

danwinship:
> I think we still need it for some tests.

I think @deads2k fixed all those. `cmd/hypershift/main.go` is the only go file in origin (outside of `staging/src/github.com/openshift/sdn`) that imports anything from `github.com/openshift/sdn`

tnozicka:
unconfirmed guess is that sdn might be hooked into test-cmd as that fails on it 
when removing it entirely in https://github.com/openshift/origin/pull/23407

PR #23426, opened 2019-07-18
Bug 1734009: Unidling: minimize iptables lock contention

squeed:
This PR includes

- backport of upstream PR 71735, which moves the userspace proxy over to an asynchronous model
- Changes needed to use that backport
- A small carry patch that lets us tweak the proxy internals slightly
- A patch that combines both proxy's sync loops, preventing lock contention

Fixes: BZ 1699341


squeed:
@dcbw PTAL, thanks!

squeed:
Uh, unit test failures have nothing to do with this. Is origin broken?

danwinship:
/hold
This needs to be fixed in master and 4.1 before it can be fixed in 3.11

squeed:
> This needs to be fixed in master and 4.1 before it can be fixed in 3.11

Yup. I just picked up 3.11 first, since there already was the backport *and* the sdn repo was being migrated as I started work.

dcbw:
@squeed are you sure you want the `hybrid-proxy: bump debug level` commit? I believe that was just for local debugging to make sure unidling was working.

dcbw:
@squeed could also squash these two together:

UPSTREAM: 71735: proxy/userspace: respect minSyncInterval
proxy: consolidate ServicesHandler/EndpointsHandler into ProxyProvider

vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier_openshift.go
```diff
@@ -0,0 +1,15 @@
+package iptables
+
+/*
```
dcbw: C-style comments make Go's brain hurt.

dcbw:
otherwise looks good to me

nstielau:
Are we good to merge this so we can get it in August 12th 3.11.z?

pkg/network/proxy/proxy.go
```diff
@@ -332,12 +332,37 @@ func (proxy *OsdnProxy) OnEndpointsDelete(ep *kapi.Endpoints) {
...
 	}
 }
 
 func (proxy *OsdnProxy) OnEndpointsSynced() {
-	proxy.baseEndpointsHandler.OnEndpointsSynced()
+	proxy.baseProxy.OnEndpointsSynced()
+}
+
+func (proxy *OsdnProxy) OnServiceAdd(service *kapi.Service) {
+	glog.V(2).Infof("sdn proxy: add svc %s/%s: %v", service.Namespace, service.Name, service)
```
dcbw: @squeed I guess this is a leftover from my original patches.

PR #23427, opened 2019-07-18
[WIP]Add plugin for mapping tolerations to RBAC and SAs

ravisantoshgudimetla:
@deads2k - First pass at mapping tolerations to RBAC

I did not register this plugin in the api server and there are some edge cases that I need to add.


vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/admission/initializer"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	coreapi "k8s.io/kubernetes/pkg/apis/core"
+)
+
+const (
+	PluginName       = "PodTolerations"
+	masterToleration = "node-role.kubernetes.io/master"
```
deads2k: this is our plugin, name it for us.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+	"io"
+
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/admission/initializer"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	coreapi "k8s.io/kubernetes/pkg/apis/core"
+)
+
+const (
+	PluginName       = "PodTolerations"
```
deads2k: this should be something int he openshift namespace.  `scheduling.openshift.io/RestrictPodTolerations`

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
deads2k: wrong line

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func NewPodTolerationsPlugin() admission.Interface {
+	return &podTolerations{
+		Handler: admission.NewHandler(admission.Create, admission.Update),
+	}
+}
+
+func (p *podTolerations) Validate(attr admission.Attributes, _ admission.ObjectInterfaces) error {
+	if attr.GetResource().GroupResource() != coreapi.Resource("pods") && attr.GetKind().GroupKind() != coreapi.Kind("Pod") {
```
deads2k: and subresource should be len zero

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil
+	}
+	tolerations := podSpec.Tolerations
+	for _, toleration := range tolerations {
+		allow, err := p.checkPodTolerationAccess(attr, toleration)
+		if err != nil {
+			return err
+		}
+		if !allow {
+			return fmt.Errorf("requested user %v doesn't have permission to provide the given tolerations", attr.GetUserInfo().GetName())
```
deads2k: you need to return API errors.  `NewForbidden` probably.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+func (p *podTolerations) validatePodTolerations(attr admission.Attributes, podSpec coreapi.PodSpec) error {
+	// if a pod has no tolerations, then no need to validate it's tolerations
+	if len(podSpec.Tolerations) == 0 {
+		return nil
+	}
+	tolerations := podSpec.Tolerations
+	for _, toleration := range tolerations {
+		allow, err := p.checkPodTolerationAccess(attr, toleration)
+		if err != nil {
+			return err
```
deads2k: apierrors, so probalby NewInternalError

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+/*
+tolerations:
+- effect: NoSchedule
+  key: node-role.kubernetes.io/master
+  operator: Exists
+are converted to
+
+apiGroup: toleration.scheduling.openshift.io
+Resource: node-role.kubernetes.io/master
+resourceName: <empty_string>/NoSchedule
```
deads2k: Maybe `<effect>::<value>`?  slashes aren't allowed in names and we don't want something unexpected to break.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+// checkPodTolerationAccess checks if pods are allowed to have a particular toleration. In this case, we're focusing
+// on master tolerations. For example,
+/*
+tolerations:
+- effect: NoSchedule
+  key: node-role.kubernetes.io/master
+  operator: Exists
+are converted to
+
+apiGroup: toleration.scheduling.openshift.io
```
deads2k: please write corresponding RBAC rules for default cluster roles.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+Resource: node-role.kubernetes.io/master
+resourceName: <empty_string>/NoSchedule
+verbs: ["Exists"]
+*/
+func (p *podTolerations) checkPodTolerationAccess(attr admission.Attributes, toleration coreapi.Toleration) (bool, error) {
```
deads2k: please extract the "toleration to attributesrecord" from the authorizer call so you can unit test it separately.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerations/admission.go
```diff
@@ -0,0 +1,114 @@
...
+	verb := toleration.Operator
+	tolerationKey := ""
+	// TODO: Add some more cases for blanket toleration etc.
+	if len(toleration.Key) == 0 {
+		tolerationKey = masterToleration
+	}
+	authzAttr := authorizer.AttributesRecord{
+		User:            attr.GetUserInfo(),
```
deads2k: this looks wrong.  The user is supposed to be a serviceaccount only, right?

deads2k: remember that serviceaccounts have hardcoded groups,  you'll want to specify those too.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,156 @@
...
+		if !allow {
+			return admission.NewForbidden(attr, fmt.Errorf("requested user %v doesn't have permission to provide the given tolerations", podSpec.ServiceAccountName))
+		}
+	}
+	return nil
+}
+
+func (p *podTolerations) IsAuthorized(podSAName string, attr admission.Attributes, toleration coreapi.Toleration) bool{
+	var saInfo user.Info
+	saInfo = serviceaccount.UserInfo(attr.GetNamespace(), podSAName, "")
```
deads2k: this one looks right and it appears to fill in all the groups for you.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func NewPodTolerationsPlugin() admission.Interface {
+	return &podTolerations{
+		Handler: admission.NewHandler(admission.Create, admission.Update),
+	}
+}
+
+func (p *podTolerations) Validate(attr admission.Attributes, _ admission.ObjectInterfaces) error {
+	if attr.GetResource().GroupResource() != coreapi.Resource("pods") && attr.GetKind().GroupKind() != coreapi.Kind("Pod") && len(attr.GetSubresource()) == 0 {
```
deads2k: I'm not so great at parsing these.  Mind breaking them apart?

```go
if attr.groupResource != thing{
    return
}
if attr.groupkind != thing{
    return
}
if len() != 0{
    return
}
```

I think your conditional on len is backwards.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (p *podTolerations) validatePodTolerations(attr admission.Attributes, podSpec coreapi.PodSpec) error {
+	// if a pod has no tolerations, then no need to validate it's tolerations
+	if len(podSpec.Tolerations) == 0 {
+		return nil
+	}
+	tolerations := podSpec.Tolerations
+	if len(podSpec.ServiceAccountName) == 0 {
+		return admission.NewForbidden(attr, fmt.Errorf("service account is not provided. The restrict pod tolerations plugin is completely lost"))
```
deads2k: more user focused.  `serviceaccounts are required on pods that want to use tolerations`

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	for _, toleration := range tolerations {
+		allow := p.IsAuthorized(podSpec.ServiceAccountName, attr, toleration)
```
deads2k: aggregate all the failures so the user gets the whole list back at once.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,156 @@
...
+/*
+tolerations:
+- effect: NoSchedule
+  key: node-role.kubernetes.io/master
+  operator: Exists
+
+are converted to
+
+apiGroup: toleration.scheduling.openshift.io
+Resource: toleration
```
deads2k: seems like this should be key

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
ravisantoshgudimetla: Sure thing.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,156 @@
...
+		APIGroup:        "toleration.scheduling.openshift.io",
+		ResourceRequest: true,
+		Name:            tolerationKey,
+	}
+	return authzAttr
+}
+
+
+// checkPodTolerationAccess checks if pods are allowed to have a particular toleration. In this case, we're focusing
+// on master tolerations. For example,
```
deads2k: the example is good, but please also list the generic transformation.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,156 @@
...
+tolerations:
+- effect: NoSchedule
+  key: node-role.kubernetes.io/master
+  operator: Exists
+
+are converted to
+
+apiGroup: toleration.scheduling.openshift.io
+Resource: toleration
+resourceName: <empty_string>/NoSchedule
```
deads2k: I don't think you understand why the `value/effect` in the resource name makes sense.  Also, you can't split on slash here.  For splitting, you probably want the effect first, right?

Empty names match all names.  Specified names match their specification.  This construct makes it easy to allow someone to tolerate a particular key with any value or any effect.  It makes it possible to enumerate the key/effect tuple as a unit.

ravisantoshgudimetla: The above example is wrong. Please look at https://github.com/openshift/origin/pull/23427/files#diff-b49487e41f10910366174f767ef67f90R99-R137

ravisantoshgudimetla: Ok, so, if someone wants to have toleration against any key with any value or any effect, they can use the https://github.com/openshift/origin/pull/23427/files#diff-b49487e41f10910366174f767ef67f90R121-R123

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+	if len(podSpec.Tolerations) == 0 {
+		return nil
+	}
+	tolerations := podSpec.Tolerations
+	if len(podSpec.ServiceAccountName) == 0 {
+		return admission.NewForbidden(attr, fmt.Errorf("serviceaccounts are required on pods that want to use tolerations"))
+	}
+	for _, toleration := range tolerations {
+		allow := p.IsAuthorized(podSpec.ServiceAccountName, attr, toleration)
+		var errs []error
```
deads2k: wrong scope. This needs to move out of the for loop.  Please add a multi-error unit test

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+		if !allow {
+			errs = append(errs, fmt.Errorf("serviceaccounts doesn't have permission to provide the given toleration %v on the pod", toleration.Key))
+		}
+		if len(errs) > 0 {
```
deads2k: this should be below the for loop

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	tolerations := podSpec.Tolerations
+	if len(podSpec.ServiceAccountName) == 0 {
+		return admission.NewForbidden(attr, fmt.Errorf("serviceaccounts are required on pods that want to use tolerations"))
+	}
+	for _, toleration := range tolerations {
+		allow := p.IsAuthorized(podSpec.ServiceAccountName, attr, toleration)
+		var errs []error
+		if !allow {
+			errs = append(errs, fmt.Errorf("serviceaccounts doesn't have permission to provide the given toleration %v on the pod", toleration.Key))
```
deads2k: serviceaccount

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+			return admission.NewForbidden(attr, err)
+		}
+	}
+	return nil
+}
+
+func (p *podTolerations) IsAuthorized(podSAName string, attr admission.Attributes, toleration coreapi.Toleration) bool {
+	saInfo := serviceaccount.UserInfo(attr.GetNamespace(), podSAName, "")
+	// check for hard-coded serviceaccount groups
+	SAInNamespace := serviceaccount.UserInfo(attr.GetNamespace(), "system:serviceaccounts:"+attr.GetNamespace(), "")
```
deads2k: this line and the one below it aren't necessary.  your `saInfo` is fully correct.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+	SAAcrossAllNamespace := serviceaccount.UserInfo("", "system:serviceaccounts", "")
+	return p.checkPodTolerationAccess(saInfo, attr, toleration) || p.checkPodTolerationAccess(SAInNamespace, attr, toleration) || p.checkPodTolerationAccess(SAAcrossAllNamespace, attr, toleration)
+}
+
+// buildTolerationsAuthorizationAttributes builds the tolerations attributes from the user
+func buildTolerationsAuthorizationAttributes(info user.Info, namespace string, toleration coreapi.Toleration) authorizer.Attributes {
+	verb := toleration.Operator
+	tolerationKey := toleration.Key
+	if len(toleration.Key) == 0 {
+		tolerationKey = masterToleration
```
deads2k: I think this default should be `*`

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+	saInfo := serviceaccount.UserInfo(attr.GetNamespace(), podSAName, "")
+	// check for hard-coded serviceaccount groups
+	SAInNamespace := serviceaccount.UserInfo(attr.GetNamespace(), "system:serviceaccounts:"+attr.GetNamespace(), "")
+	SAAcrossAllNamespace := serviceaccount.UserInfo("", "system:serviceaccounts", "")
+	return p.checkPodTolerationAccess(saInfo, attr, toleration) || p.checkPodTolerationAccess(SAInNamespace, attr, toleration) || p.checkPodTolerationAccess(SAAcrossAllNamespace, attr, toleration)
+}
+
+// buildTolerationsAuthorizationAttributes builds the tolerations attributes from the user
+func buildTolerationsAuthorizationAttributes(info user.Info, namespace string, toleration coreapi.Toleration) authorizer.Attributes {
+	verb := toleration.Operator
```
deads2k: tolower

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+A cluster role to access toleration resources should have rbac rules like this
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: tolerations-creator
+rules:
+- apiGroups: ["toleration.scheduling.openshift.io"]
+  resources: ["node-role.kubernetes.io/master", "node-role.kubernetes.io/worker",...]
+  verbs: ["Exists", "Equal"] This can also be verbs:["*"]
```
deads2k: do not suggest `*`

deads2k:
I'd like one e2e test that makes sure this is active please.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
ravisantoshgudimetla: Yeah, I did that yesterday evening but did not push it. I am on it.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
ravisantoshgudimetla: Fair enough.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
ravisantoshgudimetla: You mean, we just need saInfo? I thought you wanted hardcoded SA'es to be used as well.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
ravisantoshgudimetla: Silly me, what was I thinking. Thanks for catching it :)

ravisantoshgudimetla:
> I'd like one e2e test that makes sure this is active and working.

Started on it. Got distracted. Will get to it, once I am done with changes with other PRs.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,166 @@
...
+/*
+tolerations:
+- effect: NoSchedule
+  key: node-role.kubernetes.io/master
+  operator: Exists
+
+are converted to
+
+apiGroup: toleration.scheduling.openshift.io
+Resource: node-role.kubernetes.io/master
```
smarterclayton: Is there a way to wildcard by label namespace?  There will be certain label sets that nodes shouldn't be allowed to label.  I suspect we're going to end up needing to have at minimum by namespace, by prefix, etc.

Can we use `namespace` in the rule for `node-role.kubernetes.io` or are you leveraging namespace?  I.e. then the label "name" == `resourceName`, the label namespace == `namespace`, and the resource == `tolerations`?

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
deads2k: > You mean, we just need saInfo? I thought you wanted hardcoded SA'es to be used as well.

I clicked through. that method builds them out for you already.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
deads2k: Using the namespace as a real namespace means that an admin granted powers in one namespace can delegate those power to multiple service accounts.  If we override the meaning of namespace, then someone with the power to grant rbac roles in the those namespaces becomes unexpectedly  powerful.

How many different toleration keys do we really expect to have?

ravisantoshgudimetla: Limiting based on namespace is really problematic. @smarterclayton, is your concern more on the lines of we will end up having writing permissions for too many tolerations on the pod spec(across multiple repos)?

test/extended/authorization/tolerations_restriction.go
```diff
@@ -0,0 +1,136 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+			err = createSAWithTolerationsAccess(cs, f.Namespace.Name)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			err = createRoleBinding(cs, f.Namespace.Name)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			err = createPodWithSA(cs, f.Namespace.Name)
+			o.Expect(err).NotTo(o.HaveOccurred())
+		})
+	})
+	g.Describe("Pod without service account should fail", func() {
```
deads2k: This test doesn't do what you want. All pods have serviceaccounts enforced via admission.  You need to test and SA without permissions.

test/extended/authorization/tolerations_restriction.go
```diff
@@ -0,0 +1,136 @@
...
+		g.It(fmt.Sprintf("should fail"), func() {
+			err = createPodWithOutSA(cs, f.Namespace.Name)
+			o.Expect(err).Should(o.HaveOccurred())
+		})
+	})
+
+})
+
+func createClusterRole(cs clientset.Interface, nsName string) error {
```
deads2k: misnamed.  This is a local role. Also, it appears to be hardcoded so please name it for what it does

vendor/k8s.io/kubernetes/openshift-kube-apiserver/kubeadmission/register.go
```diff
@@ -62,6 +64,7 @@ var (
 		"security.openshift.io/SCCExecRestrictions",
 		"route.openshift.io/IngressAdmission",
 		"quota.openshift.io/ClusterResourceQuota",
+		"scheduling.openshift.io/RestrictPodTolerations",
```
deads2k: nothing comes after quota

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,168 @@
...
+	"k8s.io/klog"
+	coreapi "k8s.io/kubernetes/pkg/apis/core"
+	"k8s.io/kubernetes/pkg/serviceaccount"
+	"strings"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const (
+	PluginName       = "scheduling.openshift.io/RestrictPodTolerations"
+	masterToleration = "node-role.kubernetes.io/master"
```
deads2k: delete

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,168 @@
...
+func (p *podTolerations) Validate(attr admission.Attributes, _ admission.ObjectInterfaces) error {
+	if attr.GetResource().GroupResource() != coreapi.Resource("pods") {
+		return nil
+	}
+	if attr.GetKind().GroupKind() != coreapi.Kind("Pod") {
+		return nil
+	}
+	if len(attr.GetSubresource()) != 0 {
+		return nil
+	}
```
deads2k: You only need to run on create and update, right?

test/extended/authorization/tolerations_restriction.go
ravisantoshgudimetla: True but the plugin also enforces the existence of SA, so this case is still valid, I'll add one more test with a SA without enough privileges but I wanted to make sure that the happy path works fine.

test/extended/authorization/tolerations_restriction.go
ravisantoshgudimetla: Yup. Thanks for catching, will change it.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
deads2k: this comment is still outstanding.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/scheduler/podtolerationsrestriction/admission.go
```diff
@@ -0,0 +1,168 @@
...
+// buildTolerationsAuthorizationAttributes builds the tolerations attributes from the user
+func buildTolerationsAuthorizationAttributes(info user.Info, namespace string, toleration coreapi.Toleration) authorizer.Attributes {
+	verb := strings.ToLower(string(toleration.Operator))
+	tolerationKey := toleration.Key
+	if len(toleration.Key) == 0 {
+		tolerationKey = "*"
+	}
+	authzAttr := authorizer.AttributesRecord{
+		User:            info,
+		Verb:            string(verb),
```
deads2k: unnecessary cast

test/extended/authorization/tolerations_restriction.go
```diff
@@ -0,0 +1,136 @@
...
+	imageutils "k8s.io/kubernetes/test/utils/image"
+)
+
+var _ = g.Describe("[Feature: Toleration restrictions] Tolerations restrictions should be functional", func() {
+	defer g.GinkgoRecover()
+	f := e2e.NewDefaultFramework("tolerations-restriction")
+	cs, err := e2e.LoadClientset()
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.Describe("Pod with service account", func() {
```
deads2k: grant permission to a named toleration and make sure that passes, another fails, and tolerating everything fails.

derekwaynecarr:
If taints are recommended as a means to provide hard partitioning, I think it makes sense to enable an RBAC model for specifying specific tolerations on nodes.  Hard partitioning is desired for special resources like GPUs where you want to repel workloads from landing on that node due to cost (see: https://github.com/NVIDIA/gpu-operator#hard-partitioning).  Another potential example for hardware partitioning is around network security zones for particular instance types.  In that model, you may declare a node is in a particular DMZ and also want to require explicit RBAC to use that DMZ.

The problem with taints right now is it is intended for both hard and soft partitioning, but this PR doesn't give a great way to configure for both.  Is it possible we could either a) enable this behind a feature gate that says hard partitioning only, or b) identify a set of keys that require hard partitioning (role, dedicated hardware, network security zone)?

derekwaynecarr:
In general, I am worried about the friction this will cause for anyone that was use tolerations for soft partitioning schemes.

PR #23428, opened 2019-07-18
TEMP: Skip template conformance tests

adambkaplan:
Skip these until we determine the root cause of bug 1731222.


tnozicka:
/hold
(feel free to unhold if that PR doesn't fix it)

deads2k:
/hold cancel

That still didn't solve them all, but it got better

PR #23431, opened 2019-07-18
Bug 1683057: fix flaky router metrics test

ironcladlou:
In v4, there are two ingress controller replicas by default. The ingress
controllers have their own uncoordinated metrics state. The metrics test was
refactored to speak to the metrics endpoint via a Service in front of the
ingress controllers. This means the metrics requests are getting load balanced
to different Service endpoints. The test was originally designed in v3 to assume
communication with a single ingress controller endpoint and makes assertions
about the metrics state of the single ingress controller. By abstracting
communication through the Service, the test no longer knows which ingress
controller it's talking to, and so sometimes gets the answers it expects, and
sometimes not, resulting in flakiness.

Refactor the test to communicate with ingress controller endpoints directly as
appropriate rather than through a Service.

### DIFF
--- a/test/extended/router/metrics.go
+++ b/test/extended/router/metrics.go
@@ -13,17 +13,22 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 
 	corev1 "k8s.io/api/core/v1"
+
 	kapierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+
 	clientset "k8s.io/client-go/kubernetes"
 	watchtools "k8s.io/client-go/tools/watch"
+
 	"k8s.io/kubernetes/pkg/client/conditions"
+
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 
 	exutil "github.com/openshift/origin/test/extended/util"
@@ -34,9 +39,11 @@ var _ = g.Describe("[Conformance][Area:Networking][Feature:Router]", func() {
 	var (
 		oc = exutil.NewCLI("router-metrics", exutil.KubeConfigPath())
 
-		username, password, execPodName, ns, host string
-		statsPort                                 int
-		routerNamespace, serviceIP, bearerToken   string
+		username, password, execPodName, ns string
+
+		statsPort int
+
+		routerNamespace, routerIP, ingressEndpoint, bearerToken string
 	)
 
 	g.BeforeEach(func() {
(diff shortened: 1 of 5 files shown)


ironcladlou:
Monday I'll try harder to reproduce this in a cluster — if I don't have any luck I'd like to merge this and use the CI data to diagnose if there are still problems.

ironcladlou:
cc @knobunc

ironcladlou:
@Miciah we got one! https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/23431/pull-ci-openshift-origin-master-e2e-aws/11964

test/extended/router/metrics.go
```diff
@@ -134,7 +136,7 @@ var _ = g.Describe("[Conformance][Area:Networking][Feature:Router]", func() {
 					}
 					// send a burst of traffic to the router
 					g.By("sending traffic to a weighted route")
-					err = expectRouteStatusCodeRepeatedExec(ns, execPodName, fmt.Sprintf("http://%s", serviceIP), "weighted.metrics.example.com", http.StatusOK, times)
+					err = expectRouteStatusCodeRepeatedExec(ns, execPodName, fmt.Sprintf("http://%s", ingressEndpoint), "weighted.metrics.example.com", http.StatusOK, times)
```
Miciah: Why use `ingressEndpoint` rather than `routerIP`?

ironcladlou: Depending on the platform, the ingress controller might be speaking PROXY. To portably exercise a a real route in the context of a black box test we need to be somewhat contextual (e.g. go through the external LB).

IMO all these router tests needs shuffled/relocated/refactored. Things that need to talk directly to a router probably belong in openshift/router e2e tests. Stuff that remains in the origin suite should be using route hostnames.

ironcladlou:
Hang on a second here... something about https://github.com/openshift/origin/pull/23431#discussion_r312599038 is nagging me...

/hold

test/extended/router/metrics.go
ironcladlou: On second thought, doesn't this mean the traffic is going to be balanced between the routers?

ironcladlou: Maybe this test is a good candidate for biting the bullet and moving over to openshift/router?

Miciah: > On second thought, doesn't this mean the traffic is going to be balanced between the routers?

Yeah.  However, the test keeps sending traffic to the service till the router from which it gathers metrics reports receiving traffic, so it should be all right:  https://github.com/openshift/origin/blob/f9d13286aae646b1e5ea9fe26695c1dbfcd7fcc6/test/extended/router/metrics.go#L125-L134

ironcladlou: Seems brittle, but as long as the test is now stable, is it worth going ahead like this to restore coverage right away?

Miciah: I think so.

ironcladlou: Okay.

/hold cancel

ironcladlou:
The test failed again some new reason that I had better get to the bottom of before making a bad situation worse...

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/23431/pull-ci-openshift-origin-master-e2e-aws/12031

/hold

test/extended/router/metrics.go
ironcladlou: The test failure in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/23431/pull-ci-openshift-origin-master-e2e-aws/12031 looks like a failure waiting for traffic to be received by the expected router — given we don't have control over the balancing behavior, isn't this a new source of flakes?

PR #23438, opened 2019-07-19
UPSTREAM: <drop>: disable version check to unblock CI

deads2k:
***THIS MUST BE REVERTED***

We need to unblock CI so we can fix the problem that causes this issue.  This requires an urgent bugzilla.


sttts:
/hold

until this is linked to the issue this is going to fix. We have to document those things.

deads2k:
/hold

try https://github.com/openshift/oc/pull/18 first

PR #23439, opened 2019-07-19
remove old image files no longer produced from origin

deads2k:
These have all been moved to other repos now.


images/kube-proxy/Dockerfile.rhel
```diff
@@ -1,18 +0,0 @@
-FROM registry.svc.ci.openshift.org/ocp/4.2:hyperkube
```
tnozicka: I don't think we moved kube-proxy, nor did ART
https://github.com/openshift/release/blob/37e66a2d9acadc40bc91ba79b1d65e03c5a366bb/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L52-L59

PR #23443, opened 2019-07-21
UPSTREAM: 80393: Prevent a flake in a pod e2e test

smarterclayton:
We can fail to observe ready because pod completes so fast.

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-4.1/978


smarterclayton:
Can you tag the upstream?

PR #23467, opened 2019-07-24
UPSTREAM: <carry>: remove embedded oauth server

deads2k:
This removes the oauthserver from the kube-apiserver to eliminate a big dependency.


vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
```diff
@@ -107,7 +107,7 @@ func NewOpenShiftKubeAPIServerConfigPatch(delegateAPIServer genericapiserver.Del
 		// END ADMISSION
 
 		// HANDLER CHAIN (with oauth server and web console)
```
enj: Fix comment?

enj:
/lgtm

Can fix comment in followup.

PR #23468, opened 2019-07-24
remove binaries that we no longer use

deads2k:
The only binaries produced by origin are `hyperkube` and `openshift-tests`.  This pares us back.


mfojtik:
This looks good to me. It removes the Makefile targets and binaries we don't build in this repository anymore.

/lgtm

deads2k:
/hold

https://github.com/openshift/origin/pull/23470 got me into the origin spec file and this should have failed that rpm build.  fun stuff.

PR #23469, opened 2019-07-24
remove dead test code

deads2k:
test-cmd and test-integration have been replaced by e2e-cmd and e2e-tests plus bugzillas. 

This pull removes the dead test code to eliminate our surface area.


deads2k:
remove spurious hold on removed owners

PR #23470, opened 2019-07-24
remove oc tools from origin

deads2k:
These are confusing. They aren't actually testing our code.  We fixed that with https://github.com/openshift/oc/pull/27 .  This removes the confusing/incorrect values.


mfojtik:
/lgtm

This is supporting the cleanup and stabilization effort.

PR #23471, opened 2019-07-24
remove internal api usage from test

deads2k:
nothing should be using internal types anymore


enj:
/lgtm

Copy pasta oops during the port.

PR #23472, opened 2019-07-24
Bug 1711544: remove the long names test

rphillips:
Upstream changed the behavior of log path names resulting in the long Pod names test to fail. This PR proposes to remove the long names test to start the discussion of whether we really need this test, and to remove it from the tree.

https://github.com/kubernetes/kubernetes/issues/80042


bparees:
this isn't removing the test, it's removing it from the disabled bucket, no?

/cc @adambkaplan

test/extended/util/test.go
```diff
@@ -379,29 +379,28 @@ var (
...
-			"PersistentVolumes NFS when invoking the Recycle reclaim policy",   // failing for some reason
-			`Simple pod should handle in-cluster config`,                       // kubectl cp is not preserving executable bit
-			`Services should be able to up and down services`,                  // we don't have wget installed on nodes
-			`Network should set TCP CLOSE_WAIT timeout`,                        // possibly some difference between ubuntu and fedora
-			`Services should be able to create a functioning NodePort service`, // https://bugzilla.redhat.com/show_bug.cgi?id=1711603
-			`\[NodeFeature:Sysctls\]`,                                          // needs SCC support
-			`should check kube-proxy urls`,                                     // previously this test was skipped b/c we reported -1 as the number of nodes, now we report proper number and test fails
-			`extremely long build/bc names are not problematic`,                // there's a problem in kubelet when creating log directory in vendor/k8s.io/kubernetes/pkg/kubelet/kuberuntime/helpers.go:181, this changed in https://github.com/kubernetes/kubernetes/pull/74441
```
bparees: @adambkaplan this is the test in question that's being affected (long_names.go)

bparees:
@rphillips is there a new upper limit?  We can change the existing build test to conform to the new upper limit and keep the test.

rphillips:
@bparees This PR removes the actual test and removes the `extremely long build/bc names are not` string. gofmt formatted the section differently so it is hard to see.

Upstream doesn't really have a new limit... A feature they implemented unintentionally broke the limit based on the file path length.

/retest

bparees:
> @bparees This PR removes the actual test 

whoops, totally skimmed over the deleted file.

/approve

/assign @adambkaplan

adambkaplan:
The value of this test is to ensure that if the build config has a really long name, the resulting build pod and mounts are created. The build controller takes advantage of library-go code to hash the names of the pod and ConfigMap if we go over the 253 char limit.

Overall I don't think this e2e is necessary, since we have unit testing for all build strategies which create pod specs with very long build config names. However, those unit tests in openshift-controller-manager seem to be missing a few verification bits (ex: ensure the pod name is under the char limit).

/hold

adambkaplan:
/lgtm

Otherwise looks good.

rphillips:
@adambkaplan Was there another action item on this PR? or is it ok to merge?

test/extended/builds/long_names.go
```diff
@@ -1,94 +0,0 @@
...
-				err = oc.Run("get").Args("bc", bcB).Execute()
-				o.Expect(err).To(o.HaveOccurred())
-
-				err = oc.Run("get").Args(brB1.BuildPath).Execute()
-				o.Expect(err).To(o.HaveOccurred())
-
-				err = oc.Run("get").Args(brB2.BuildPath).Execute()
-				o.Expect(err).To(o.HaveOccurred())
-
-				g.By("Verifying build config A was untouched by delete")
```
adambkaplan: This implies that at some point, having a really long name in the buildConfig caused OpenShift to delete build objects by naming convention (rather than using owner refs?). 

@bparees is this something we should be concerned about?

/lgtm cancel

bparees: i think it was to ensure that when we truncated the names, we didn't cause duplicates?

bparees: ie that bcB and its builds, had their own ownerref tree that didn't have any relation to bcA's ownerref tree.

bparees: which, yes, is still something to be concerned about (that if/when we truncate names, we are doing it in a way that ensures sufficient uniqueness)

adambkaplan: okay - this is something that we can unit test (there are specific methods which generate the names for the pod and related ConfigMaps).

/hold cancel

/lgtm

PR #23448, opened 2019-07-22
Final SDN cleanup

danwinship:
Origin is currently vendoring openshift/sdn only so that it can (1) build an `openshift-controller-manager` subcommand that hasn't been used since 3.11, (2) build an image that no one is building from this repo any more, (3) test that `openshift-sdn --help` prints the expected output.

Obsoletes #23407 

/assign @deads2k @tnozicka

PR #23475, opened 2019-07-24
Be free!

deads2k:
Stop vendoring openshift binaries. They should never be re-added to they will effectively pin pulling in new levels of kube.  Right now, openshift/oc is still present because of the RPM build, but we should remove that as soon as we get the new RPM building.


deads2k:
/hold

this bump is fake.  We need to wire fake deps for hyperkube into our package or we drop transitive deps that we really need.  Also, publishing is broken.

deads2k:
/hold cancel

Built on top of https://github.com/openshift/origin/pull/23448

PR #23479, opened 2019-07-25
Skip another upstream test under multitenant

danwinship:
The multitenant test is still broken at the moment but once it's fixed, it shouldn't pass because we never implemented preservation of source pod IPs through service IPs in multitenant. So we need to skip that test too. Since it's possible there will be other tests we need to skip in the future too, this reorganizes a bit to put the multitenant skips in the same place as all the other skips.

/cc @rcarrillocruz @squeed

### DIFF
--- a/cmd/openshift-tests/e2e.go
+++ b/cmd/openshift-tests/e2e.go
@@ -49,9 +49,10 @@ var staticSuites = []*ginkgo.TestSuite{
 		Only the portion of the openshift/conformance test suite that applies to the openshift-sdn multitenant plugin.
 		`),
 		Matches: func(name string) bool {
-			return !strings.Contains(name, "[Feature:NetworkPolicy]") && strings.Contains(name, "[Suite:openshift/conformance/")
+			return strings.Contains(name, "[Suite:openshift/conformance/parallel") && !strings.Contains(name, "[Skipped:multitenant]")
 		},
-		Parallelism: 30,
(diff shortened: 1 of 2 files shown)


squeed:
Seems fine.
/lgtm
I wonder if the failing tests are flakes or are related.

danwinship:
no, those shouldn't be related
/retest

mfojtik:
Please create bugzilla that tracks the disable tests if you haven't already.

danwinship:
@mfojtik This test suite is just for the ovs-multitenant plugin, which is headed towards deprecation. The client IP test never passed with that plugin, and we don't plan to fix it. (The test passes fine and is not skipped with the ovs-networkpolicy plugin, which is the default in 4.x.)

danwinship:
@rcarrillocruz e2e-cmd appears to be broken but it's not a required test anyway

danwinship:
Just realized we were trying to run the serial tests in parallel, which was causing flakes... updated `openshift/conformance/multitenant` to now only include the `openshift/conformance/parallel` tests. Which means we won't be testing any serial tests under ovs-multitenant, but that's fine. We're still testing all the big-picture stuff, which will catch any differences between ovs-networkpolicy and ovs-multitenant.

danwinship:
@mfojtik can you approve?

danwinship:
merged into #23797

PR #23483, opened 2019-07-25
make test/cmd/basicresources work

deads2k:
match-version won't match since we build in separate repos


deads2k:
/hold

I think we do this after we enable the test overall.

mfojtik:
/lgtm

It passed.

PR #23485, opened 2019-07-25
UPSTREAM: <carry>: run ResourceQuota before ClusterResourceQuota

p0lyn0mial:
changes ClusterResourceQuota to be a validating admission plugin so that the correct order of admission is preserved, that is RQ is before CRQ.
this prevents a scenario where a resource cannot be created due to RQ limits but CRQ would allow for it because it doesn't know about RQ constraints.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,6 @@ notifications:
   irc: "chat.freenode.net#openshift-dev"
 
 sudo: false
+
+git:
+  depth: false
--- a/pkg/quota/admission/clusterresourcequota/admission.go
+++ b/pkg/quota/admission/clusterresourcequota/admission.go
@@ -56,6 +56,7 @@ type clusterQuotaAdmission struct {
 var _ oadmission.WantsInternalKubernetesInformers = &clusterQuotaAdmission{}
 var _ oadmission.WantsOpenshiftInternalQuotaClient = &clusterQuotaAdmission{}
 var _ oadmission.WantsClusterQuota = &clusterQuotaAdmission{}
+var _ admission.ValidationInterface = &clusterQuotaAdmission{}
 
 const (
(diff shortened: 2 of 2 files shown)


p0lyn0mial:
> Travis CI - Pull Request Failing after 6m — Build Failed

/home/travis/gopath/src/github.com/openshift/origin/hack/lib/constants.sh: line 131: OS_GIT_MAJOR: unbound variable

wtf :(

.travis.yml
```diff
@@ -14,3 +14,6 @@ notifications:
   irc: "chat.freenode.net#openshift-dev"
 
 sudo: false
+
+git:
+  depth: false
```
p0lyn0mial: a few hours later . . . I figured out that


the lack of this setting was causing `/home/travis/gopath/src/github.com/openshift/origin/hack/lib/constants.sh: line 131: OS_GIT_MAJOR: unbound variable` error

Previously the depth was set to 50 and that made the build to fail. The build scripts didn't find the version using `git describe` because the history was too short and didn't have the tags for `3.9`

p0lyn0mial:
@sttts I think `ci/openshift-jenkins/gcp` fails due to 
`An unhandled exception occurred while running the lookup plugin 'openshift_master_facts_default_predicates'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Unknown short_version 3.11`

do you know what might be wrong?

p0lyn0mial:
I think it never worked, see: https://openshift-gce-devel.appspot.com/builds/origin-ci-test/pr-logs/directory/test_pull_request_origin_extended_conformance_gce_39

mfojtik:
/override ci/openshift-jenkins/gcp

As this test is not working for 3.9 branch.

PR #23487, opened 2019-07-26
auto for smoke case OCP-10658

zhouying7780:
Auto for smoke case: OCP-10658. 
Checkpoint:  `oc rsh` command with specific flags should work well.

### DIFF
--- a/test/extended/testdata/cli/pod-with-two-containers.yaml
+++ b/test/extended/testdata/cli/pod-with-two-containers.yaml
@@ -1,0 +1,36 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: doublecontainers
+  labels:
+    name: hello-centos
+spec:
+  containers:
+  - name: hello-centos
+    image: docker.io/centos:centos7
+    command:
+      - /bin/sleep
+      - infinity
+    resources:
+      limits:
+        memory: 256Mi
+    terminationMessagePath: "/dev/termination-log"
+    imagePullPolicy: IfNotPresent
+    capabilities: {}
+    securityContext: {}
+  - name: hello-centos-2
+    image: docker.io/centos:centos7
+    command:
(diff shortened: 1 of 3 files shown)


adambkaplan:
Workloads team is in a better position to review.

/assign @damemi

test/extended/testdata/cli/pod-with-two-containers.yaml
```diff
@@ -0,0 +1,33 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: doublecontainers
+  labels:
+    name: hello-openshift
+spec:
+  containers:
+  - name: hello-openshift
+    image: aosqe/hello-openshift
```
soltysh: I'd prefer we stick what's already available and used everywhere so `openshift/hello-openshift` please.

test/extended/testdata/cli/pod-with-two-containers.yaml
```diff
@@ -0,0 +1,33 @@
...
+      protocol: TCP
+    resources:
+      limits:
+        memory: 256Mi
+    terminationMessagePath: "/dev/termination-log"
+    imagePullPolicy: IfNotPresent
+    capabilities: {}
+    securityContext: {}
+  - name: hello-openshift-fedora
+    image: aosqe/fedora_base
```
soltysh: Ditto, why not just running hello-openshift twice?

soltysh: openshift or centos orgs are preferred if you look through other test data.

zhouying7780: Thanks very much, I've updated.

zhouying7780:
@soltysh  Thanks for feedback, I'v updated the testdata, Please help review again.

PR #23491, opened 2019-07-26
Bug 1732193: UPSTREAM: 80518: Fix detachment of deleted volumes

jsafrane:
https://bugzilla.redhat.com/show_bug.cgi?id=1732193
upstream: https://github.com/kubernetes/kubernetes/pull/80518

cc @openshift/storage

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack_volumes.go
+++ b/vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack_volumes.go
@@ -26,7 +26,7 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -41,7 +41,7 @@ import (
 	"github.com/mitchellh/mapstructure"
 	"gopkg.in/gcfg.v1"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
bertinatto: nit: only change in the file

jsafrane: Interesting, it has been merged upstream this way. I removed it from this backport.

PR #23494, opened 2019-07-26
remove images/kube-proxy

sallyom:
Now that kube-proxy image is built from openshift/sdn can remove kube-proxy build from origin. 

depends on https://gitlab.cee.redhat.com/openshift-art/ocp-build-data/merge_requests/195


sallyom:
@deads2k 
remove https://github.com/openshift/origin/blob/master/publishing-kube-rules.yaml#L101-#L106 ?
https://github.com/openshift/origin/blob/master/glide.yaml#L54-#L56 / bump ?

PR #23496, opened 2019-07-27
Update gophercloud

gnufied:
Update gophercloud library.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1731282


jsafrane:
Where the gophercloud version comes from?  Upstream 73437 (commit [55a8dbcbfbe1833330d88fc67421532d3fbbb066](https://github.com/kubernetes/kubernetes/commit/55a8dbcbfbe1833330d88fc67421532d3fbbb066) contains slightly different code. And this PR contains thousands of files that are not in Kubernetes. What is our vendor pruning policy?

gnufied:
Yes - I brought forward gophercloud version mentioned in - https://github.com/kubernetes/kubernetes/pull/79637 but since upstream has moved to use gomodules, calling it `UPSSTREAM: xxx` is tricky. 

And I needed https://github.com/kubernetes/kubernetes/commit/55a8dbcbfbe1833330d88fc67421532d3fbbb066  because it contains necessary changes in kubernetes code to accomodate for version bump. 

`v0.1.0` of gophercloud seems to have more sane fixes than the SHA referred in dims commit. I can perhaps make this clear.

Also about vendor stipping - I am not sure why is glide not stripping files by scanning the imports. Ideally it should.

jsafrane:
> v0.1.0 of gophercloud seems to have more sane fixes than the SHA referred in dims commit. I can perhaps make this clear and split the commits.

Yes, more obvious commit message would be useful.

> Also about vendor stripping/pruning - I am not sure why is glide not stripping files by scanning the imports. Ideally it should.

Ack, I see other deps are not pruned either, so let it be as it is.

Still, this replaces lot of OpenStack code with something that was not tested by us and can bring regressions.

```
 895 files changed, 70436 insertions(+), 4307 deletions(-)
```

It's mostly adding support for new APIs, yet I am not sure we have power to retest whole OCP on OSP in 3.11.z timeframe.

gnufied:
> Still, this replaces lot of OpenStack code with something that was not tested by us and can bring regressions.

So the reason we need the entire 0.1.0 release rather than just the cherry-pick of the commit that fixes the issue is because - https://github.com/gophercloud/gophercloud/pull/1408 while fixed the issue, it had a regression. And the fix that fixes the regression had conflict with other related changes.  It is my understanding that - reauth. mechanism in gophercloud is quite problematic and suffers from several problems. Only around 0.1.0 release - dust seems to have settled down little bit.

I do not know gophercloud library and nor do I understand it deeply. So far our options are:

1. Just roll with this patch. Ask QE to test it and  provide affected customer with a hotfix release that includes this patch and ask them to test. If everything works - we are probably fine. 

2. Really delve into reauth. mechanism code and make a more targetted fix which hopefully does not have regression that upstream fix had.  There are several downsides of this approach: 
a. We will be creating a mini-fork of gophercloud. 
b. We will miss other fixes that were done to library including for reauthentication.
c. It will not be any less time consuming than option#1.

jsafrane:
There are also these options:

3) Find acceptable workaround
4) Update OCP to 4.1
5) Update OpenStack to 13

gnufied:
Just for a quick update - the hotfix build was deployed in customers environment and it was confirmed that, this patch fixes the customer's issue and no regression was found. We should be good to go with this PR.

gnufied:
I am currently waiting for https://github.com/openshift/origin/pull/23620 PR to merge first, so as some of the issues with missing dependencies can be resolved. I will have to update this PR once https://github.com/openshift/origin/pull/23620 merges.

gageorsburn:
@gnufied Any timeframe on getting conflicts resolved so this can get merged?

jsafrane:
The customer has confirmed it fixes the bug and has no obvious regressions.

/lgtm

gnufied:
The CI for 3.11 branches is broken. @dobbymoodge is working on fixing the CI

PR #23504, opened 2019-07-29
UPSTREAM: <carry>: run ResourceQuota before ClusterResourceQuota

p0lyn0mial:
changes ClusterResourceQuota to be a validating admission plugin so that the correct order of admission is preserved, that is RQ is before CRQ.
this prevents a scenario where a resource cannot be created due to RQ limits but CRQ would allow for it because it doesn't know about RQ constraints.


p0lyn0mial:
ports https://github.com/openshift/origin/pull/23485

p0lyn0mial:
/assign @deads2k

for approval

PR #23493, opened 2019-07-26
Bug 1733921:  ClusterResourceQuota usage is above project usage quota impacting all projects in ClusterResourceQuota

p0lyn0mial:
changes the order of admission plugins so that RQ is applied before CRQ.
this prevents a scenario where a resource cannot be created due to RQ limits but CRQ would allow for it because it doesn't know about RQ constraints.

PR #23509, opened 2019-07-30
UPSTREAM: <carry>: UT that checks if ResourceQuota is before ClusterR…

p0lyn0mial:
add a UT that verifies if `ResourceQuota` admission plugin is before `ClusterResourceQuota` plugin


p0lyn0mial:
covers https://github.com/openshift/origin/pull/23493

PR #23515, opened 2019-07-30
Use apiserver library go

deads2k:
switches our kube fork to use the library shared with openshift-apiserver


deads2k:
@sttts broke because publish bot broke.

PR #23519, opened 2019-07-31
Don't run [Disruptive] tests in Suite:openshift/csi

jsafrane:
The tests need SSH to the host && restart kubelet there, which is not possible now.

Therefore mark the tests into the suite only if they pass through `excludedTestsFilter` during `ginkgo.WalkTests()`.


smarterclayton:
If you want to exclude disruptive, do it from your test suite definition in cmd/openshift-tests

jsafrane:
moved to `cmd/openshift-tests`
@smarterclayton, PTAL.

PR #23526, opened 2019-07-31
UPSTREAM:<carry>: Don't supress the node update error while logging

ravisantoshgudimetla:
We can see that this particular test is flaking at the following line `Conflict when trying to add/update avoidPonds` in the test suite 

Ref: https://search.svc.ci.openshift.org/?search=pod+should+avoid&maxAge=168h&context=2&type=all, 

I want to enable debugging so that we can see why the test is failing. Note that this test is passing continuously upstream

/cc @mfojtik @tnozicka

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/framework/util.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/framework/util.go
@@ -2935,7 +2935,7 @@ func AddOrUpdateAvoidPodOnNode(c clientset.Interface, nodeName string, avoidPods
 			if !apierrs.IsConflict(err) {
 				ExpectNoError(err)
 			} else {
-				Logf("Conflict when trying to add/update avoidPonds %v to %v", avoidPods, nodeName)
+				Logf("Conflict when trying to add/update avoidPonds %v to %v with error %v", avoidPods, nodeName, err)
 			}
 		}
 		return true, nil


vendor/k8s.io/kubernetes/test/e2e/framework/util.go
```diff
@@ -2935,7 +2935,7 @@ func AddOrUpdateAvoidPodOnNode(c clientset.Interface, nodeName string, avoidPods
 			if !apierrs.IsConflict(err) {
 				ExpectNoError(err)
 			} else {
-				Logf("Conflict when trying to add/update avoidPonds %v to %v", avoidPods, nodeName)
+				Logf("Conflict when trying to add/update avoidPonds %v to %v with error %v", avoidPods, nodeName, err)
```
sttts: avoidPods

sttts:
This should be an upstream PR. No reason not to get this in upstream.

ravisantoshgudimetla:
I'll create a PR upstream as well since the issue is noticed downstream(test is working fine upstream), I am thinking of moving forward with this.

sttts:
> I'll create a PR upstream as well since the issue is noticed downstream(test is working fine upstream), I am thinking of moving forward with this.

We don't want to collect downstream patches if possible. Either make it `<drop>` or push it upstream, and then make it a "UPSTREAM: 12234: ..." patch.

PR #23532, opened 2019-08-01
Bug 1733600: kube-apiserver: wait for kube informers in openshift patch

sttts:
When SCCs are not synched, admission rejects pods with "no SecurityContextConstraints found in cluster".

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
@@ -128,6 +128,10 @@ func NewOpenShiftKubeAPIServerConfigPatch(delegateAPIServer genericapiserver.Del
 		// END CONSTRUCT DELEGATE
 
 		patchContext.informerStartFuncs = append(patchContext.informerStartFuncs, kubeAPIServerInformers.Start)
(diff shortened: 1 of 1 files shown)


vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
```diff
@@ -128,6 +128,10 @@ func NewOpenShiftKubeAPIServerConfigPatch(delegateAPIServer genericapiserver.Del
 		// END CONSTRUCT DELEGATE
 
 		patchContext.informerStartFuncs = append(patchContext.informerStartFuncs, kubeAPIServerInformers.Start)
+		patchContext.postStartHooks["openshift.io-kubernetes-informers-synched"] = func(context genericapiserver.PostStartHookContext) error {
+			kubeInformers.WaitForCacheSync(context.StopCh)
```
enj: Should this fail if you cannot sync?

sttts: potentially yes. But we seem to be very undecided elsewhere.

sttts: If I see it correctly, this can only return bools in the map if stopCh closed. In that case, we don't care about the result.

PR #23535, opened 2019-08-02
UPSTREAM: 80899: use set informers

Fedosin:
This patch allows to use the built-in function SetInformers to read data from a secret instead of the custom function that creates k8s client.


vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -42,11 +42,10 @@ import (
 	"gopkg.in/gcfg.v1"
 
 	"k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	netutil "k8s.io/apimachinery/pkg/util/net"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/informers"
+	listersv1 "k8s.io/client-go/listers/core/v1"
```
sttts: nit: corelistersv1

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -259,40 +260,21 @@ func configFromEnv() (cfg Config, ok bool) {
...
 	secretNamespace := os.secretNamespace
-	kubeconfigPath := os.kubeconfigPath
-
-	k8sClient, err := createKubernetesClient(kubeconfigPath)
-	if err != nil {
-		return fmt.Errorf("failed to get kubernetes client: %v", err)
-	}
 
-	secret, err := k8sClient.CoreV1().Secrets(secretNamespace).Get(secretName, metav1.GetOptions{})
+	secret, err := os.secretLister.Secrets(secretNamespace).Get(secretName)
```
sttts: I like this much more. Only concert: we ignore the kubeconfig now passed in `os`. I guess that's ok?

sttts: Is there some sensible fallback solution?

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -259,40 +260,21 @@ func configFromEnv() (cfg Config, ok bool) {
...
-	v, err := client.Discovery().ServerVersion()
-	if err != nil {
-		return nil, err
-	}
-
-	klog.Infof("Kubernetes API client created, server version %s", fmt.Sprintf("v%v.%v", v.Major, v.Minor))
-	return client, nil
+// SetInformers initializes secret lister
+func (os *OpenStack) SetInformers(informerFactory informers.SharedInformerFactory) {
+	os.secretLister = informerFactory.Core().V1().Secrets().Lister()
```
sttts: Is this implementing some interface? Add a type assertion.

sttts: Is this called in all contexts?

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
Fedosin: Yeah, I did it for consistency. If we remove this parameter from the config, but leave it in the file, there will be an error.
I think that the best option is to comment in the code that this parameter is deprecated and should not used.

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
Fedosin: agree

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
```diff
@@ -146,7 +145,10 @@ type OpenStack struct {
 	// Options for reading config data from a secret
 	secretName      string
 	secretNamespace string
+	// kubeconfigPath option is deprecated and shouldn't be used
 	kubeconfigPath  string
```
sttts: isn't there some struct you marshal into? I guess it is not this one.

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
Fedosin: What type assertion do you mean here?

vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go
Fedosin: done

sttts: Is this var still used. Can we reven delete it?

Fedosin: I think we shouldn't because it will break the installer, which sets the value. If it tries to set a non-existent value, the installation will fail.

Fedosin: I think we can merge this patch, then remove setting of the value from the installer
https://github.com/openshift/installer/blob/master/pkg/asset/manifests/openstack/cloudproviderconfig.go#L33 and, if necessary, remove kubeconfigPath from here with a separate patch.

sttts: I don't mean the config struct var, but this one (which is copied over from the former)

Fedosin: ah, you're right. for sure we can remove it

sttts:
/approve
/lgtm
/hold

@Fedosin hold cancel whenever you like

PR #23546, opened 2019-08-05
e2e: Fix namespace cleanup

smarterclayton:
The resourceToDelete refactoring resulted in namespaces not getting
deleted.


smarterclayton:
Verified test namespaces no longer exist at end of e2e (no pod logs captured from still running tests)

smarterclayton:
This likely contributed significantly to flakiness because we were still using resources in a large number (30, 40?) of e2e tests after they finished.

PR #23550, opened 2019-08-05
CL: Support waiting for arbitrary objects after creation

sjug:
The new concurrent created templates should also wait for object to be in good status.

### DIFF
--- a/test/extended/cluster/cl.go
+++ b/test/extended/cluster/cl.go
@@ -16,6 +16,7 @@ import (
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 
 	"github.com/openshift/origin/test/extended/cluster/metrics"
+	"github.com/openshift/origin/test/extended/util"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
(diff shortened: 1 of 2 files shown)


test/extended/cluster/cl.go
```diff
@@ -222,75 +223,8 @@ var _ = g.Describe("[Feature:Performance][Serial][Slow] Load cluster", func() {
...
-			}
-			if len(dcList.Items) > 0 {
-				// Get first deployment config name
-				deploymentName := dcList.Items[0].Name
-				e2e.Logf("Waiting for deployment: %q", deploymentName)
-				err = exutil.WaitForDeploymentConfig(oc.AdminKubeClient(), oc.AsAdmin().AppsClient().AppsV1(), ns, deploymentName, 1, true, oc)
-				o.Expect(err).NotTo(o.HaveOccurred())
-				e2e.Logf("Deployment %q completed", deploymentName)
-			}
+		if err := postWait(oc, namespaces); err != nil {
```
jmencak: Can we rename postWait() to give it a more descriptive meaning?  A few comments at the top of the function would also help.

test/extended/cluster/clt.go
```diff
@@ -45,27 +47,35 @@ var _ = g.Describe("[Feature:Performance][Serial][Slow] Load cluster", func() {
...
-			go clWorker(work, &wg, oc)
+			go clWorker(work, ns, &wg, oc)
 		}
 
 		for _, p := range project {
 			for j := 0; j < p.Number; j++ {
 				wg.Add(1)
 				unit := ProjectMeta{j, p, viperConfig}
 				work <- unit
+				namespace := <-ns
```
jmencak: Clever use of channels!

test/extended/cluster/cl.go
```diff
@@ -309,6 +243,83 @@ var _ = g.Describe("[Feature:Performance][Serial][Slow] Load cluster", func() {
 	})
 })
 
+func postWait(oc *util.CLI, namespaces []string) error {
+	// Wait for builds and deployments to complete
+	for _, ns := range namespaces {
+		rcList, err := oc.AdminKubeClient().CoreV1().ReplicationControllers(ns).List(metav1.ListOptions{})
+		if err != nil {
+			return err
```
jmencak: I wonder why the more specific error messages were dropped after the code was lifted into this function.  Are the generic err(s) that you return enough to tell the user what happened?

test/extended/cluster/cl.go
sjug: How's `postCreateWait()`

test/extended/cluster/cl.go
sjug: Wanted to pass the errors back to be handled in the main function, passing these new errors instead. What do you think?

PR #23551, opened 2019-08-05
UPSTREAM: 76910: webhook interception for: connect, proxy, binding, eviction

rmohr:
Allow intercepting pod subresources with webhooks. Needed for kubevirt to trigger migrations insead of shutdown when an eviction is intended.

This backports the following upstream PRs:

 * https://github.com/kubernetes/kubernetes/pull/76910

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/registry/core/pod/storage/eviction.go
+++ b/vendor/k8s.io/kubernetes/pkg/registry/core/pod/storage/eviction.go
@@ -86,6 +86,13 @@ func (r *EvictionREST) Create(ctx context.Context, obj runtime.Object, createVal
 		return nil, err
 	}
 	pod := obj.(*api.Pod)
+
+	if createValidation != nil {
+		if err := createValidation(eviction); err != nil {
+			return nil, err
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


rmohr:
/cc @eparis @derekwaynecarr 

that would be the change required for CNV.

derekwaynecarr:
@mfojtik - let’s discuss this tomorrow.

derekwaynecarr:
/hold

I would like to understand why we cannot wait to merge this when 4.3 opens.  Our 4.2 release is past feature freeze and I would like to scrutinize backporting upstream features no differently than features we author.

rmohr:
> I would like to understand why we cannot wait to merge this when 4.3 opens. Our 4.2 release is past feature freeze and I would like to scrutinize backporting upstream features no differently than features we author.

As discussed on the mailing thread regarding to kubevirt and drains, there are other options. It is at the end a decision also up to you what you find less risky for automated drain processes when kubevirt wants to do live migrations on drains. This shows what kubevirt would need.

mfojtik:
@rmohr i agree with @derekwaynecarr. We are past feature freeze for 4.2 at this point and the bar to get anything other than a bug fix is really high. I'm happy to approve this when the master branch open for 4.3 (you should also get this for free when we bump the kubernetes in 4.3, which should happen early and should be fast this time).

PR #23553, opened 2019-08-05
[gcp] Add gcp-sdk to tests image

csrwng:
Adds google cloud sdk to tests images so we can use gcloud in e2e


csrwng:
@smarterclayton @sdodson ptal 
For e2e tests on gcp we need the gcloud command. If you think this should be done elsewhere, please let me know.

smarterclayton:
/hold

Your test either needs to use an image or dynamically install gcp on demand.  We do not want to ship GCP to customers.

csrwng:
> Your test either needs to use an image or dynamically install gcp on demand. We do not want to ship GCP to customers.

I didn't realize this is an image we ship to customers. I can definitely layer the google sdk on top of this image and give it its own tag. Are you saying it shouldn't live in this repository?

PR #20031, opened 2018-06-18
Fix a crash on a certain type of unsupported NetworkPolicy

danwinship:
Similar to #19869 but different, and not caused by "new features" in this case. In theory, NetworkPolicy allows you to say:

    kind: NetworkPolicy
    spec:
      podSelector:
      ingress:
      - ports
        - protocol: TCP

meaning "allow all TCP ports but no UDP ports". This is a dumb feature, and we don't bother to implement it, but we were accidentally crashing if we saw a NetworkPolicy like that. (Which only happened because someone created a NetworkPolicy with broken indentation that caused parts of it to be parsed as the wrong subfield.)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1591584

PR #23558, opened 2019-08-06
Bug 1743747: Fix a crash on a certain type of unsupported NetworkPolicy

juanluisvaladas:
Backport to 3.10 https://github.com/openshift/origin/pull/20031
Fixes on 3.10: https://bugzilla.redhat.com/show_bug.cgi?id=1591584

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -373,15 +373,18 @@ func (np *networkPolicyPlugin) parseNetworkPolicy(npns *npNamespace, policy *net
 				return nil, fmt.Errorf("policy specifies unrecognized protocol %q", *port.Protocol)
 			}
 			var portNum int
-			if port.Port.Type == intstr.Int {
+			if port.Port == nil {
+				// FIXME: implement this?
+				return nil, fmt.Errorf("port fields with no port value are not implemented")
+			} else if port.Port.Type != intstr.Int {
+				// FIXME: implement this?
+				return nil, fmt.Errorf("named port values (%q) are not implemented", port.Port.StrVal)
+			} else {
 				portNum = int(port.Port.IntVal)
 				if portNum < 0 || portNum > 0xFFFF {
 					// FIXME: validation should catch this
 					return nil, fmt.Errorf("port value out of bounds %q", port.Port.IntVal)
 				}
-			} else {
-				// FIXME: implement this
-				return nil, fmt.Errorf("named port values (%q) are not yet implemented", port.Port.StrVal)
 			}
 			portFlows = append(portFlows, fmt.Sprintf("%s, tp_dst=%d, ", protocol, portNum))
 		}


juanluisvaladas:
In OpenShift CEE (support) we're doing a patch team to fix that will do some backports and eventually hotfixes. This is just some simple fix, easy to test easy to back port in order to get familiar with the whole process.

juanluisvaladas:
This is backported to 3.9 already:
https://github.com/openshift/ose/pull/1335/files

Which means it's currently fixed in 3.9 and 3.11, but 3.10 still has the bug.
Its referenced here https://github.com/openshift/origin/pull/20031

danwinship:
> Which means it's currently fixed in 3.9 and 3.11, but 3.10 still has the bug.

doh. It looks like 3.10 was branched shortly before the original PR merged and we didn't realize that we'd missed it and needed a separate backport

You'll need to make a clone of the original bugzilla bug as well, with version 3.10 and target release 3.10.z

/ok-to-test

juanluisvaladas:
> You'll need to make a clone of the original bugzilla bug as well, with version 3.10 and target release 3.10.z
Not quite sure if created correctly, I created this:
https://bugzilla.redhat.com/show_bug.cgi?id=1627991

As a clone of: https://bugzilla.redhat.com/show_bug.cgi?id=1591584
However I have access denied to my own bug

danwinship:
> Not quite sure if created correctly, I created this:
> https://bugzilla.redhat.com/show_bug.cgi?id=1627991

um... no you didn't :-) That's a bug from a year ago.

I'm not sure what you did before, but 1591584 doesn't list any new clones, so you didn't manage to create a clone of it.

To clone the bug, click the dark-square-over-a-dark-square icon in the group of icons at the top right of the page (right over the "Save Changes" button). Then choose "Red Hat" and "OpenShift Container Platform", and you'll get a mostly-filled-in bug report that you just have to change the Version and Target Release of. Oh, and the version number in the Summary line. And change the Assignee to either you or me.

Then once you have a bug number, come back here and add a comment with the command

    /retitle Bug XXXXXX: Fix a crash on a certain type of unsupported NetworkPolicy

which will change the PR title, and then the bot will link the PR with the bugzilla bug.

juanluisvaladas:
I see, I cannot clone it becuase I'm not allowed to see this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1627991

And probably it's going to be a recursive problem  so I just need to get into the group that has the permissions. I already requested permissions but I don't know how long it's gonna take me to get them.

danwinship:
oh, I see. The original bug is marked as blocking a tracking bug for a particular customer, which is apparently a problem since you aren't yet in the group that can see private bugs.
I've cloned the bug for you, it's https://bugzilla.redhat.com/show_bug.cgi?id=1743747. So now...

/retitle Bug 1743747 Fix a crash on a certain type of unsupported NetworkPolicy

danwinship:
oops forgot the colon
/retitle Bug 1743747: Fix a crash on a certain type of unsupported NetworkPolicy

danwinship:
$#@%!#@$% the bot can't see security bugs and I can't remove the "security" group from the bug. This bug is just all kinds of exciting.
Anyway
/lgtm

PR #23560, opened 2019-08-06
Install CSI driver on openshift-test start.

jsafrane:
Goal: create a prow job that tests CSI sidecars shipped as part of OCP / OKD. In the future, use the same approach to test CSI drivers too.

There are 6 CSI sidecars in total (and the number is growing), so to simplify job definition in each repostory, I embedded AWS EBS driver yaml files into openshift-tests.

Prow configs for each sidecar repo would then contain something like this:

```
- as: e2e-aws-csi
  commands: TEST_INSTALL_CSI_DRIVERS=aws-ebs TEST_SUITE=openshift/csi run-tests
  openshift_installer:
    cluster_profile: aws
    upgrade: false
```

Env. variables are the easiest to pass between parent openshift-test and child openshift-test processes. It's a bit hacky though. I tried command line arguments, you can find the unsuccessful attempt in the commits in this PR. 

The architecture is open for multiple CSI drivers, we will ship drivers for all clouds that we support eventually. In addition, we can easily replace yaml files with CSI drivers with CRs for an operator that installs CSI drivers, when we have them.

Future direction:
- Add such prow job to all CSI sidecar repos + origin (to catch openshift-test changes & rebases).
- When we add operators for CSI drivers (4.4?), use OLM manifests / CRDs for driver installation instead current yaml files.

/assign @smarterclayton

### DIFF
--- a/cmd/openshift-tests/csi.go
+++ b/cmd/openshift-tests/csi.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/openshift/origin/test/extended/csi"
+
+	"k8s.io/kubernetes/test/e2e/storage/external"
+)
+
+const (
+	manifestEnvVar       = "TEST_CSI_DRIVER_FILES"
+	installDriversEnvVar = "TEST_INSTALL_CSI_DRIVERS"
+)
+
+// Initialize openshift/csi suite, i.e. define CSI tests from TEST_CSI_DRIVER_FILES.
+func initCSITests(dryRun bool) error {
+	driverList := os.Getenv(installDriversEnvVar)
+	if driverList != "" {
+		drivers := strings.Split(driverList, ",")
+		for _, driver := range drivers {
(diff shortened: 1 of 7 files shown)


cmd/openshift-tests/csi.go
```diff
@@ -0,0 +1,55 @@
...
+	"k8s.io/kubernetes/test/e2e/storage/external"
+)
+
+const (
+	manifestEnvVar       = "TEST_CSI_DRIVER_FILES"
+	installDriversEnvVar = "TEST_INSTALL_CSI_DRIVERS"
+)
+
+// Initialize openshift/csi suite, i.e. define CSI tests from TEST_CSI_DRIVER_FILES.
+func initCSITests() error {
```
smarterclayton: Who calls this?

jsafrane: It's called from `initProvider`: https://github.com/openshift/origin/blob/d88cc07845b4e368145da435feb9e1bc318a54ac/cmd/openshift-tests/openshift-tests.go#L277-L286

PR #23565, opened 2019-08-07
Bug 1723753: OpenShift incorrectly applies RFC 6902 JSON PATCH

p0lyn0mial:
this pull pins the library to the specific commit - the one that is required by k8s.

### DIFF
--- a/vendor/github.com/evanphx/json-patch/patch_test.go
+++ b/vendor/github.com/evanphx/json-patch/patch_test.go
@@ -127,6 +127,11 @@ var Cases = []Case{
 		`{ "foo": [ "all", "cows", "grass", "eat" ] }`,
 	},
 	{
+		`{ "foo": [ "a", "b", "c", "d", "e" ] }`,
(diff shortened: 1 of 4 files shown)


vendor/github.com/evanphx/json-patch/patch_test.go
```diff
@@ -126,6 +126,11 @@ var Cases = []Case{
 		`[ { "op": "move", "from": "/foo/1", "path": "/foo/2" } ]`,
 		`{ "foo": [ "all", "cows", "grass", "eat" ] }`,
 	},
+	{
+		`{ "foo": [ "a", "b", "c", "d", "e" ] }`,
+		`[ { "op": "move", "from": "/foo/1", "path": "/foo/2" } ]`,
+		`{ "foo": [ "a", "c", "b", "d", "e" ] }`,
```
p0lyn0mial: before this pull, the patch would give `a, c, d, b, e` - which is incorrect.

PR #23567, opened 2019-08-07
Bug 1739445: Remove oc rpms

tnozicka:
Those RPMs have moved into oc repo. We are now trying to switch ART team to use them but we shouldn't build from 2 places at once. 

hyperkube needs to stay for providing kubelet to the base image. (https://github.com/openshift/origin/blob/7aedc3cb19c39e22b382eff96380bef245f92352/images/os/Dockerfile#L7)

/cc @smarterclayton @eparis @deads2k @soltysh

### DIFF
--- a/images/os/Dockerfile
+++ b/images/os/Dockerfile
@@ -4,7 +4,9 @@ COPY --from=registry.svc.ci.openshift.org/openshift/origin-v4.0:machine-os-conte
 RUN set -x && yum install -y ostree yum-utils selinux-policy-targeted && \
     commit=$( find /srv -name *.commit | sed -Ee 's|.*objects/(.+)/(.+)\.commit|\1\2|' | head -1 ) && \
     mkdir /tmp/working && cd /tmp/working && \
-    yumdownloader --enablerepo=built --destdir=/tmp/rpms openshift-hyperkube openshift-clients && \
+    PACKAGES=(openshift-hyperkube openshift-clients) && \
+    yumdownloader --enablerepo=built --destdir=/tmp/rpms "${PACKAGES[@]}"  && \
+    ls /tmp/rpms/ && (cd /tmp/rpms/ && ls ${PACKAGES[@]/%/*}) && \
     for i in $(find /tmp/rpms/ -name openshift-* -iname *.rpm); do echo "Extracting $i ..."; rpm2cpio $i | cpio -div; done && \
     mv etc usr/ && \
     mkdir -p /tmp/tmprootfs/etc && \
--- a/images/os/Dockerfile
+++ b/images/os/Dockerfile
@@ -4,11 +4,11 @@ COPY --from=registry.svc.ci.openshift.org/openshift/origin-v4.0:machine-os-conte
 RUN set -x && yum install -y ostree yum-utils selinux-policy-targeted && \
     commit=$( find /srv -name *.commit | sed -Ee 's|.*objects/(.+)/(.+)\.commit|\1\2|' | head -1 ) && \
     mkdir /tmp/working && cd /tmp/working && \
-    PACKAGES=(openshift-hyperkube openshift-clients) && \
+    PACKAGES=(openshift-hyperkube) && \
     yumdownloader --enablerepo=built --destdir=/tmp/rpms "${PACKAGES[@]}"  && \
     ls /tmp/rpms/ && (cd /tmp/rpms/ && ls ${PACKAGES[@]/%/*}) && \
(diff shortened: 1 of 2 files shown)


tnozicka:
/hold
(for a sync with https://jira.coreos.com/browse/ART-844) when this is tagged)

tnozicka:
rpm repos timeouts for other packages
/retest

tnozicka:
```
https://mirror.umd.edu/fedora/linux/updates/29/Modular/x86_64/repodata/5ddd47c03de6cd5e4b1d102c96f23f7e794c404611c49820745047dfbfbd4320-primary.xml.gz: [Errno 14] HTTPS Error 404 - Not Found
Trying other mirror.
No Match for argument openshift-clients
++ find /tmp/rpms/ -name 'openshift-*' -iname '*.rpm'
```
sounds like we don't support cross repo rpms

tnozicka:
It is failing on missing `openshift-clients` package - not sure what to do to make it pull a promoted RPM from `oc` repo like images do... @smarterclayton ideas?

images/os/Dockerfile
```diff
@@ -4,7 +4,9 @@ COPY --from=registry.svc.ci.openshift.org/openshift/origin-v4.0:machine-os-conte
 RUN set -x && yum install -y ostree yum-utils selinux-policy-targeted && \
     commit=$( find /srv -name *.commit | sed -Ee 's|.*objects/(.+)/(.+)\.commit|\1\2|' | head -1 ) && \
     mkdir /tmp/working && cd /tmp/working && \
-    yumdownloader --enablerepo=built --destdir=/tmp/rpms openshift-hyperkube openshift-clients && \
+    PACKAGES=(openshift-hyperkube openshift-clients) && \
+    yumdownloader --enablerepo=built --destdir=/tmp/rpms ${PACKAGES[@]}  && \
```
stevekuznetsov: If you are using `[@]` you _must_ use double quotes or it's useless.

stevekuznetsov:
If you're trying to install RPMs built in a different repo from the ephemeral RPM server, then no, that does not work.

tnozicka:
> If you're trying to install RPMs built in a different repo from the ephemeral RPM server, then no, that does not work.

Sounds like a blocker for 4.2; given oc lives in openshift/oc and it needs to be installed into machine-os image build from origin, we need a way to promote those as we do for images. Can DPTP team help support promoting RPMs?

stevekuznetsov:
Can you build machine-os-content from `openshift/oc`?

stevekuznetsov:
Or, we could move `openshift/oc` back into this repo.

deads2k:
> Or, we could move `openshift/oc` back into this repo.

No. That is an anti-goal.  The two are separate and upstream is pushing them even further apart.



> Can you build machine-os-content from `openshift/oc`?

What does machine-os-content mean?



How about we simply eliminate the `Requires` of clients?  There's not actual need for a dependency between them.

tnozicka:
talked to @smarterclayton and we can drop the oc from the machine-os-image as nothing in CI uses it and ART will install oc there

tnozicka:
test-cmd should be fixed by https://github.com/openshift/origin/pull/23649
/test e2e-cmd

sosiouxme:
At this point, if an image needs an oc client, we would prefer copying it over from ose-cli image rather than picking it up as an RPM.

That doesn't really matter here, though, as we still need to produce a client RPM, it just needs to come from a different location.

soltysh:
> That doesn't really matter here, though, as we still need to produce a client RPM, it just needs to come from a different location.

@sosiouxme not quite, aside from the binary RPM must deliver man pages, I don't think we put those in the image, although maybe we should. @tnozicka can you check if we do and update our image to have it? That'd be
a) handy for users to have man there
b) handy for art

sosiouxme:
We actually had a bug open about oc not having man pages - I think we passed it on, but it is certainly desirable for oc to have man pages wherever it might be used by a human.

tnozicka:
I am not sure about copying it out from the image. What if the man pages or some of the shared libraries differ? We link differently for RHEL, more dynamically and with special libraries. If we need to install those into both RHEL 7 and 8 those libraries and man pages could differ and we have the CLI image build just for one of them, so copying it out might not be the safest option.

tnozicka:
ok, this is ready to be unhold when ART switches the oc RPM to be built from oc repo

tnozicka:
https://jira.coreos.com/browse/ART-844 is done
/hold cancel

PR #23568, opened 2019-08-07
Bug 1721380: more exhaustive verification of imagestream import completeness

gabemontero:
Employs verifications noted in https://bugzilla.redhat.com/show_bug.cgi?id=1721380#c11 
but with my rationale for verifying the latest OCM rollout does include the internal image registry hostname, instead of doing a canary build.

- verifies the internal registry hostname value in the OCM config matches the image config
- verifies the OCM rollout occurs  after the image config was created/populated with the hostname
- verifies the OCM rollout is complete, and the pods are updated 
- verifies the samples operator conditions indicate successful completion of imagestream import (this replaces my porting of the verification logic to the tests this PR originally had.

/assign @adambkaplan 
/assign @bparees 
@openshift/openshift-team-developer-experience FYI

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -1,6 +1,8 @@
 package util
 
 import (
+	"bufio"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -22,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -36,7 +39,9 @@ import (
 
 	appsv1 "github.com/openshift/api/apps/v1"
 	buildv1 "github.com/openshift/api/build/v1"
+	configv1 "github.com/openshift/api/config/v1"
 	imagev1 "github.com/openshift/api/image/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	appsv1clienttyped "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
 	buildv1clienttyped "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
 	imagev1typedclient "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
@@ -52,23 +57,120 @@ import (
 func WaitForInternalRegistryHostname(oc *CLI) (string, error) {
 	e2e.Logf("Waiting up to 2 minutes for the internal registry hostname to be published")
 	var registryHostname string
+	foundOCMLogs := false
+	isOCMProgressing := true
+	podLogs := map[string]string{}
 	err := wait.Poll(2*time.Second, 2*time.Minute, func() (bool, error) {
 		imageConfig, err := oc.AsAdmin().AdminConfigClient().ConfigV1().Images().Get("cluster", metav1.GetOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				e2e.Logf("Image config object not found")
+				return false, nil
+			}
+			e2e.Logf("Error accessing image config object: %#v", err)
+			return false, err
+		}
+		if imageConfig == nil {
+			e2e.Logf("Image config object nil")
+			return false, nil
+		}
+		registryHostname = imageConfig.Status.InternalRegistryHostname
+		if len(registryHostname) == 0 {
+			e2e.Logf("Internal Registry Hostname is not set in image config object")
+			return false, nil
+		}
+
+		// verify that the OCM config's internal registry hostname matches
+		// the image config's internal registry hostname
+		ocm, err := oc.AdminOperatorClient().OperatorV1().OpenShiftControllerManagers().Get("cluster", metav1.GetOptions{})
 		if err != nil {
 			if errors.IsNotFound(err) {
 				return false, nil
 			}
 			return false, err
 		}
-		if imageConfig == nil {
+		observedConfig := map[string]interface{}{}
+		err = json.Unmarshal(ocm.Spec.ObservedConfig.Raw, &observedConfig)
+		if err != nil {
 			return false, nil
 		}
-		registryHostname = imageConfig.Status.InternalRegistryHostname
-		if len(registryHostname) == 0 {
+		internalRegistryHostnamePath := []string{"dockerPullSecret", "internalRegistryHostname"}
+		currentRegistryHostname, _, err := unstructured.NestedString(observedConfig, internalRegistryHostnamePath...)
+		if err != nil {
+			e2e.Logf("error procesing observed config %#v", err)
 			return false, nil
 		}
-		return true, nil
+		if currentRegistryHostname != registryHostname {
+			e2e.Logf("OCM observed config hostname %s does not match image config hostname %s", currentRegistryHostname, registryHostname)
+			return false, nil
+		}
+		// check pod logs for messages around image config's internal registry hostname has been observed and
+		// and that the build controller was started after that observation
+		pods, err := oc.AdminKubeClient().CoreV1().Pods("openshift-controller-manager").List(metav1.ListOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
+		}
+		for _, pod := range pods.Items {
+			req := oc.AdminKubeClient().CoreV1().Pods("openshift-controller-manager").GetLogs(pod.Name, &corev1.PodLogOptions{})
+			readCloser, err := req.Stream()
+			if err == nil {
+				b, err := ioutil.ReadAll(readCloser)
+				if err == nil {
+					podLog := string(b)
+					podLogs[pod.Name] = podLog
+					scanner := bufio.NewScanner(strings.NewReader(podLog))
+					firstLog := false
+					for scanner.Scan() {
+						line := scanner.Text()
+						if strings.Contains(line, "docker_registry_service.go") && strings.Contains(line, registryHostname) {
+							firstLog = true
+							continue
+						}
+						if firstLog && strings.Contains(line, "build_controller.go") && strings.Contains(line, "Starting build controller") {
+							e2e.Logf("the OCM pod logs indicate the build controller was started after the internal registry hostname has been set in the OCM config")
+							foundOCMLogs = true
+							break
+						}
+					}
+				}
+			} else {
+				e2e.Logf("error getting pod logs: %#v", err)
+			}
+		}
+		if !foundOCMLogs {
+			e2e.Logf("did not find the sequence in the OCM pod logs around the build controller getting started after the internal registry hostname has been set in the OCM config")
+			return false, nil
+		}
+
+		if !isOCMProgressing {
+			return true, nil
+		}
+		// now cycle through the OCM operator conditions and make sure the Progressing condition is done
+		for _, condition := range ocm.Status.Conditions {
+			if condition.Type != operatorv1.OperatorStatusTypeProgressing {
+				continue
+			}
+			if condition.Status != operatorv1.ConditionFalse {
+				e2e.Logf("OCM rollout still progressing or in error: %v", condition.Status)
+				return false, nil
+			}
+			e2e.Logf("OCM rollout progressing status reports complete")
+			isOCMProgressing = true
+			return true, nil
+		}
+		e2e.Logf("OCM operator progressing condition not present yet")
+		return false, nil
 	})
+
+	if !foundOCMLogs {
+		e2e.Logf("dumping OCM pod logs since we never found the internal registry hostname and start build controller sequence")
+		for podName, podLog := range podLogs {
+			e2e.Logf("pod %s logs:\n%s", podName, podLog)
+		}
+	}
 	if err == wait.ErrWaitTimeout {
 		return "", fmt.Errorf("Timed out waiting for internal registry hostname to be published")
 	}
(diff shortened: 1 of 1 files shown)


bparees:
i'm not seeing how this would address my theory that the openshift-controller-manager has cached a copy of the imagestream that does not have the `status.dockerImageRepository` field populated?

bparees:
(remember that status.dockerImageRepository is added by a decorator on GET, so a client that is watching the imagestream for updates will not get an event/update when the internal registry starts up, thus a client that was watching the imagestream *before* the internal registry existed, would have seen the imagestream created/updated with no status.DockerImageRepository field value, and cached it)

bparees:
(and if there's no status.DockerImageRepository value, then a "local" imagestreamtag resolution will still resolve to a pullspec that points to the upstream registry, not the internal registry).

gabemontero:
thanks for re-stating / writing down your theory from last week @bparees ... hopefully I retain it better this time

I'll start an investigative thread into what you've outlined.

But as I noted in the description, there is conceivable value add in improving the readiness check for the openshift imagestreams.

Do we still want to consider this change in that context @bparees @adambkaplan ?  Or would you prefer I go straight to the library-go path, and create a common function between the extended tests and the samples operator?

/retest

adambkaplan:
@gabemontero I'm inclined to prefer a shared function if that would be accepted in library-go. We can leave this PR open until that consensus is reached.

Separately we should add logic to the test to confirm Ben's hypothesis.

gabemontero:
ok @adambkaplan sounds good re: leave this for now, but pursue library-go

and yeah my "investigation" into @bparees 's hypothesis most likely will include adding diagnostics, logic, etc.

gabemontero:
FYI ... @bparees 's last point re: his hypothesis is confirmed:  https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L329-L337

gabemontero:
Also, a possible prior reference to @bparees 's theory in the build controller:  https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/build/build_controller.go#L876-L880

Going to see if I can find `"the integrated registry is not configured"` in the build controller log from the last failed run.

gabemontero:
And to complete the cross referencing, the decorate logic in the apiserver that sets Status.DockerImageRepository:  https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/registry/imagestream/strategy.go#L617-L634

gabemontero:
OK @openshift/openshift-team-developer-experience @bparees I've updated this PR so that the tests validate internal registry hostname propagation to the OCM pods and that the imagestream imports in the openshift namespace are in fact successful.

PTAL / see if you agree with the approach/rationale

test/extended/util/framework.go
```diff
@@ -56,18 +60,99 @@ func WaitForInternalRegistryHostname(oc *CLI) (string, error) {
...
 		}
-		return true, nil
+
+		// now confirm the OCM was restarted after the image registry was set;
+		// to do that, we are going to compare the create time of the ocm pods
+		// with the last transition times of the ocm operator object progressing condition;
+		// we'll also confirm timings between the OCM rollout and image config creation,
+		// and we'll verify that the OCM config's internal registry hostname matches
+		// the image config's internal registry hostname
+		// get the OCM cfg
```
bparees: does the controller manager log anything when it starts up, about what it knows the internal registry hostname to be?  perhaps checking the logs of the controller manager pod could be an easier/more reliable way to check this?  (i'd even be ok adding such a log output to the controller manager if necessary... it's probably something we want to know anyway)

gabemontero: This does show up in the logs:

```
I0813 15:21:47.865941       1 docker_registry_service.go:284] Updating registry URLs from map[172.30.122.228:5000:{} image-registry.openshift-image-registry.svc.cluster.local:5000:{} image-registry.openshift-image-registry.svc:5000:{}] to map[172.30.122.228:5000:{} image-registry.openshift-image-registry.svc.cluster.local:5000:{} image-registry.openshift-image-registry.svc:5000:{}]
I0813 15:21:47.866011       1 deleted_token_secrets.go:72] caches synced
I0813 15:21:47.899404       1 build_controller.go:466] Starting build controller
```

that said, though less code for sure, parsing the pod logs seems more fragile, no?

bparees: We can make it not fragile by introducing a test that ensures that log message is always output (so that we do not accidentally remove the log output in the future).

bparees: (in fairness i have not looked through your current implementation of the check based on timestamps/etc, so let me do that and give you a more informed opinion)

gabemontero: dumping those pod logs if this logic finds a problem, or may dump the pod  logs and filter for build_controller.go and docker_registry_service.go regardless, would be useful info if the hiccup occurred again

So maybe the approach here *plus* that?

test/extended/util/framework.go
```diff
@@ -56,18 +60,99 @@ func WaitForInternalRegistryHostname(oc *CLI) (string, error) {
...
+				continue
+			}
+			if condition.Status != operatorv1.ConditionFalse {
+				e2e.Logf("OCM rollout still progressing or in error: %v", condition.Status)
+				return false, nil
+			}
+			// make sure OCM rolled out after image config created; since the image registry operator sets the internal
+			// registry hostname when it creates the image config object, then making sure the latest OCM rollout is
+			// after its create time is another indication that config value was processed by the OCM
+			if condition.LastTransitionTime.Before(&imageConfig.CreationTimestamp) {
```
bparees: i don't know that looking at the imageconfig creationtimestamp is meaningful.  Pretty sure we create that object when the cluster is created (even before we populate it w/ the internal registry hostname).

I think that's going to be the most difficult part of this approach....  being confident that a given controller pod is "newer" than the time at which the internal registry hostname config value became available. Because there's no specific timestamp for that value being set in a config resource.

gabemontero: My initial thought as well re: image config create time stamp.

But in my prototyping, the image config was always at a generation of 1 though, and in the image registry operator, I see that on the create of the image.config object, they are setting the internal registry hostname.

```
apiVersion: config.openshift.io/v1
kind: Image
metadata:
  annotations:
    release.openshift.io/create-only: "true"
  creationTimestamp: "2019-08-12T20:19:52Z"
  generation: 1
  name: cluster
  resourceVersion: "8460"
  selfLink: /apis/config.openshift.io/v1/images/cluster
  uid: 8abeb8c5-bd3e-11e9-a339-02e2cbcb7c6e
spec: {}
status:
  internalRegistryHostname: image-registry.openshift-image-registry.svc:5000
```

And here is the creation code, where they immediately set the status after the create:  https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/imageconfig.go#L78-L108

I took those pieces to mean we could effectively leverage the create timestamp.

thoughts?

gabemontero:
discussion on the approach with @bparees aside, my checks are proving too aggressive when compared to my local testing:

Several instances of `Aug 13 15:54:20.132: INFO: OCM progressing condition last transition time 2019-08-13 15:29:58 +0000 UTC is younger than oldest OCM pod 2019-08-13 15:29:56 +0000 UTC`

Need to sort out, at least while the approach discussion is not finalized.

gabemontero:
I'm error'ing out too soon ... need to let the poll continue.

test/extended/util/framework.go
bparees: that logic will take effect only if the resource does not already exist.  But it could already exist (in which case the registry operator just does an Update path), and in fact i'd expect it to already exist due to:

https://github.com/openshift/cluster-config-operator/blob/master/empty-resources/0000_05_config-operator_02_image.cr.yaml

it's possible it's non-deterministic who creates the image config resources (the cluster config operator, or the image registry operator).

it's also possible that generation value isn't being incremented properly, as I just added a route to my registry (which causes the image config resource to be updated w/ an external hostname) and the generation is still "1".  So i'd say that's not a reliable indicator that the creation state == current state of the resource.


In any case the image config object's timestamp isn't really what matters, it's the controller manager's config timestamp.  You need to know that all the controller pods are newer than the time at which the controller manager operator config resource was updated to include the internal registry hostname (which it gets from the image config resource).  Which again I think will be hard to determine since you'd have to know not the creation timestamp but specifically when a particular field was updated.

gabemontero: OK then ... punting on correlating creation times of the OCM pods against any of the config 

I propose leaving in:
- at least making sure the OCM config's setting matches the image config's setting (in addition to the current check that the image config is set)
- leave in the check to make sure an OCM rollout is not in progress (including all 3 pods properly recreted) .... current e2e results already show that an OCM rollout is still in progress when the e2e-aws-build run starts .... I've seen this in other recent test exercises ... OCM progressing goes to false before the daemonset/pod rollout finishes ... I was reluctant to change that earlier, but maybe something to revisit.... I'm referring to my recent changes in https://github.com/openshift/origin/blob/master/test/extended/builds/cluster_config.go as part of moving integration tests to extended tests
- either contrive log parsing based on the current log noted above, or add new logging that makes said parsing easier ... and use that to confirm the config has been consumed

- or would adding an annotation for last update time on the OCM config be too rebellious?

thoughts?

bparees: > at least making sure the OCM config's setting matches the image config's setting (in addition to the current check that the image config is set)

makes sense

> leave in the check to make sure an OCM rollout is not in progress (including all 3 pods properly recreted) 

makes sense, and may well be "sufficient" to resolve the core issue here.

> OCM progressing goes to false before the daemonset/pod rollout finishes

really?  is that a bug?  it sounds wrong.

> either contrive log parsing based on the current log noted above, or add new logging that makes said parsing easier ... and use that to confirm the config has been consumed

per my earlier suggestion i certainly support this one :)


> or would adding an annotation for last update time on the OCM config be too rebellious?

i'm pretty sure that's been done elsewhere, so not too rebellious per se.

gabemontero: > > at least making sure the OCM config's setting matches the image config's setting (in addition to the current check that the image config is set)
> 
> makes sense

cool

> 
> > leave in the check to make sure an OCM rollout is not in progress (including all 3 pods properly recreted)
> 
> makes sense, and may well be "sufficient" to resolve the core issue here.

that thought occurred to me as well, though seemed perhaps hopeful at first blush

> 
> > OCM progressing goes to false before the daemonset/pod rollout finishes
> 
> really? is that a bug? it sounds wrong.

@adambkaplan and I were of the same thinking back with the integration test to extended test work, but I have not had time to research the history there and reach out to whomever touched that piece last.

I'm more of a mind at the moment to handle that separately.

I've opened https://bugzilla.redhat.com/show_bug.cgi?id=1740887 to track.
> 
> > either contrive log parsing based on the current log noted above, or add new logging that makes said parsing easier ... and use that to confirm the config has been consumed
> 
> per my earlier suggestion i certainly support this one :)
> 
> > or would adding an annotation for last update time on the OCM config be too rebellious?
> 
> i'm pretty sure that's been done elsewhere, so not too rebellious per se.

OK cool I'll add that as well, though will take on log parsing initially.  Based if/how it lands, if possible I'll nuke the log parsing once it is ready.

gabemontero:
ok @bparees I've pushed the log parsing 

I ditched all the timesstamp compares, even the pod create with OCM progressing (I had hoped I could keep that one), as my testing saw the occasionally saw the those two items land on the wrong side of typical results, like with in a second or so ... classic timing anomaly wrt watch events.

As we opined in the comment, quite possibly the additional checks, including the progressing still true one (as I saw that in some PR runs) probably mitigates the flake associated with the bugzilla.  

But if future instances occur, I think we pursue the annotation on the OCM to indicate the internal image registry hostname update has occurred (though an initial peek into the unstructured observed config flow makes me think it will be at least a little cumbersome, so I want to defer if possible).

test/extended/util/framework.go
```diff
@@ -56,18 +61,104 @@ func WaitForInternalRegistryHostname(oc *CLI) (string, error) {
...
+			if err == nil {
+				b, err := ioutil.ReadAll(readCloser)
+				if err == nil {
+					podLog := string(b)
+					podLogs[pod.Name] = podLog
+					scanner := bufio.NewScanner(strings.NewReader(podLog))
+					firstLog := false
+					for scanner.Scan() {
+						line := scanner.Text()
+						if strings.Contains(line, "docker_registry_service.go") && strings.Contains(line, registryHostname) {
```
bparees: does this output get printed in the OCM w/o increasing the OCM's verbosity level?  this is why i was thinking we might need to make it an info message (which, again, I don't think is a bad thing in general for helping us investigate customer clusters)

gabemontero: that comes from the standard verbosity level

gabemontero:
rpm problems on e2e runs ... reported to 4-dev-triage and build-cop

test/extended/util/framework.go
bparees: sorry, i did a bad grep.

gabemontero: no worries

test/extended/util/framework.go
```diff
@@ -105,6 +196,36 @@ func WaitForOpenShiftNamespaceImageStreams(oc *CLI) error {
 					return false
 				}
 			}
+			// check the samples operator to see about imagestream import status
```
bparees: doing this check inside the "for each lang" loop doesn't seem to make sense?  you just need to do it once, right?

test/extended/util/framework.go
```diff
@@ -56,18 +61,104 @@ func WaitForInternalRegistryHostname(oc *CLI) (string, error) {
...
+				}
+			} else {
+				e2e.Logf("error getting pod logs: %#v", err)
+			}
+		}
+		if !foundOCMLogs {
+			e2e.Logf("did not find the sequence in the OCM pod logs around the build controller getting started after the internal registry hostname has been set in the OCM config")
+			for podName, podLog := range podLogs {
+				e2e.Logf("pod %s logs:\n%s", podName, podLog)
+			}
```
bparees: you might only want to dump these logs after you've completely given up on waiting...as it is you're potentially going to dump a *lot* of logs with this being called with each iteration of the poll loop.

test/extended/util/framework.go
gabemontero: yep part of next push

test/extended/util/framework.go
gabemontero: yep part of next push

gabemontero:
updates pushed @bparees thanks

PR #23569, opened 2019-08-07
[gcp]: Skip failing e2e tests

csrwng:
Skips currently failing e2e tests for gce provider.

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -440,6 +440,8 @@ var (
 			`DynamicProvisioner should test that deleting a claim before the volume is provisioned deletes the volume`, // test is very disruptive to other tests
 
 			`Should be able to support the 1\.7 Sample API Server using the current Aggregator`, // down apiservices break other clients today https://bugzilla.redhat.com/show_bug.cgi?id=1623195
+
+			`\[HPA\] Horizontal pod autoscaling \(scale resource: Custom Metrics from Stackdriver\)`, // down custom metrics apiservices break other clients
 		},
 		"[Skipped:azure]": {
 			"Networking should provide Internet connection for containers", // Azure does not allow ICMP traffic to internet.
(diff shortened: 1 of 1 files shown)


csrwng:
@abhinavdahiya needs an lgtm

csrwng:
Removed skip for auth test. Fix delivered here: https://github.com/openshift/origin/pull/23577
Added comments to consolidate with: https://github.com/openshift/origin/pull/23582

jsafrane:
/lgtm
from aos-storage point of view

test/extended/util/test.go
```diff
@@ -476,6 +478,10 @@ var (
 			`\[sig-storage\] In-tree Volumes \[Driver: azure\] \[Testpattern: Pre-provisioned PV \(ext4\)\] volumes should be mountable`,
 			`\[sig-storage\] In-tree Volumes \[Driver: azure\] \[Testpattern: Pre-provisioned PV \(filesystem volmode\)] volumeMode should create sc, pod, pv, and pvc, read/write to the pv, and delete all created resources`,
 		},
+		"[Skipped:gce]": {
+			// Requires creation of a different compute instance in a different zone and is not compatible with volumeBindingMode of WaitForFirstConsumer which we use in 4.x
+			`\[sig-scheduling\] Multi-AZ Cluster Volumes \[sig-storage\] should only be allowed to provision PDs in zones where nodes exist \[Suite:openshift/conformance/parallel\] \[Suite:k8s\]`,
```
abhinavdahiya: I think we need to drop the `[Suite:openshift/conformance/parallel\] \[Suite:k8s\]`

The test is not skipped in the run

test/extended/util/test.go
```diff
@@ -476,6 +478,13 @@ var (
 			`\[sig-storage\] In-tree Volumes \[Driver: azure\] \[Testpattern: Pre-provisioned PV \(ext4\)\] volumes should be mountable`,
 			`\[sig-storage\] In-tree Volumes \[Driver: azure\] \[Testpattern: Pre-provisioned PV \(filesystem volmode\)] volumeMode should create sc, pod, pv, and pvc, read/write to the pv, and delete all created resources`,
 		},
+		"[Skipped:gce]": {
+			// Requires creation of a different compute instance in a different zone and is not compatible with volumeBindingMode of WaitForFirstConsumer which we use in 4.x
+			`\[sig-scheduling\] Multi-AZ Cluster Volumes \[sig-storage\] should only be allowed to provision PDs in zones where nodes exist`,
+
+			// Tries to ssh directly to a node. None of our nodes have external IPs
+			`[k8s.io] [sig-node] crictl should be able to run crictl on the node`,
```
abhinavdahiya: we should add `\[sig-storage\] Flexvolumes should be mountable` as they require ssh too...

abhinavdahiya: also this probably needs to be `\[k8s.io\] \[sig-node\] crictl should be able to run crictl on the node`

abhinavdahiya:
I think we can skip `\[sig-auth\] Metadata Concealment should run a check-metadata-concealment job to completion`

https://github.com/kubernetes/kubernetes/blob/5a7b978c7401efad5672bcc876bcf5d3dfa71bd0/test/e2e/auth/metadata_concealment.go#L30

we already use openshift-sdn to conceal metadata https://github.com/openshift/sdn/blob/c9cbc6e2c8864a5421c8c63810610333179d4e52/cmd/sdn-cni-plugin/openshift-sdn_linux.go#L129-L133

abhinavdahiya:
Also skip
`\[sig-storage\] Detaching volumes should not work when mount is in progress`

As it is also depends on SSH.

abhinavdahiya:
based on https://github.com/kubernetes/kubernetes/issues/80767 esp https://github.com/kubernetes/kubernetes/pull/80769

it seems like we should open a bug for `\[sig-api-machinery\] AdmissionWebhook Should be able to deny pod and configmap creation` and skip it for now.

abhinavdahiya:
https://github.com/openshift/release/pull/4735 should fix the `[sig-scheduling] Multi-AZ Cluster Volumes [sig-storage] should schedule pods in the same zones as statically provisioned PVs [Suite:openshift/conformance/parallel] [Suite:k8s]` error.

wking:
/retest

@csrwng just ran `oc delete project ci-op-9ijgiymj` to clear any busted cache that might have lead to the broken `e2e-cmd` and `e2e-gcp` jobs.

PR #23570, opened 2019-08-08
Ocp 21041

wmuha:
Link to polarion tc:
https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-21041

step1: 
Check pods by
# oc -n openshift-monitoring get pod -o wide | grep -e alertmanager-main -e prometheus-k8s

step2:

alertmanager-main pod
*********************************************************************
affinity:
    podAntiAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - podAffinityTerm:
          labelSelector:
            matchExpressions:
            - key: alertmanager
             operator: In
              values:
              - main
*********************************************************************
prometheus-k8s pod
*********************************************************************
  affinity:
    podAntiAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - podAffinityTerm:
          labelSelector:
            matchExpressions:
            - key: prometheus
              operator: In
              values:
              - k8s
*********************************************************************
--

PR #23571, opened 2019-08-08
Ocp 21041

wmuha:
fyi: https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-21041


s-urbaniak:
as @brancz mentioned in https://github.com/openshift/origin/pull/23570#issuecomment-519421281, the right place for unit tests like these is in https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/manifests_test.go. I agree here is not the best place for these.

wmuha:
Thank you @s-urbaniak and @brancz for your comments! will move test there

stlaz:
/close
Please use a better name for the PR next, although this certainly catches one's eye

PR #23576, opened 2019-08-08
Bug 1739601: UPSTREAM: 80491: Changed IsCriticalPod to return true in case of static pods

sjenning:
https://github.com/kubernetes/kubernetes/pull/80491

Proactive pick for upstream fix.

Some fuzz around https://github.com/kubernetes/kubernetes/commit/ca6003bc75a3b9c81f38738c7363c9453247682e#diff-9fe046de3c6aaa377bb7fa24a34509c9 but otherwise it applies cleanly.

Extra +/-1 line changed is `goimports` adding `v1` alias
https://github.com/openshift/origin/pull/23576/files#diff-5cdd56e52dd392c4f849386f73dc723aR23

@derekwaynecarr @mrunalp


mfojtik:
Need a BZ reference after feature freeze :-)

/hold

sjenning:
I'm fine leaving it until 4.3 unless @derekwaynecarr wants to push it in for 4.2

derekwaynecarr:
This should go all the way back to 4.1.

PR #23577, opened 2019-08-08
Bug 1738690: UPSTREAM: 76637: Add missing node.address != "" condition in tests

sjenning:
https://github.com/kubernetes/kubernetes/pull/76637

xref https://bugzilla.redhat.com/show_bug.cgi?id=1738690

@csrwng


csrwng:
lgtm but not familiar with how upstream carries are handled. Will let @bparees or @deads2k take a look

PR #23579, opened 2019-08-09
test/operators: whitelist gcp-routes-controller static pod as it is part of infra

abhinavdahiya:
More info in https://github.com/openshift/machine-config-operator/pull/1031 as to why the gcp-routes-controller pod is required to keep running.

/cc @csrwng


csrwng:
@ironcladlou or @bparees ptal

PR #23581, opened 2019-08-09
Fixed jenkins verifyservice e2e tests 

waveywaves:
Fixed as containers were going to crashloopbackoff.
The image loaded before was an s2i image without the source. 
In this iteration we are using the redis-epemeral template with reduced memory which also provides us with the basic service to test and in the template jenkins-verifyservice-pipeline we are creating the headless service and the buildconfig with the jenkins file to test the feature.

cc @akram @gabemontero

### DIFF
--- a/test/extended/builds/pipeline_origin_bld.go
+++ b/test/extended/builds/pipeline_origin_bld.go
@@ -274,33 +274,44 @@ var _ = g.Describe("[Feature:Builds][Feature:Jenkins][Slow] openshift pipeline b
 			})
 
 			g.By("should verify services successfully", func() {
-				// instantiate the bc
+				redisTemplate := "redis-ephemeral"
+				redisAppName := "redis"
+				verifyServiceBuildConfig := "jenkins-verifyservice-pipeline"
+
+				newAppRedisEphemeralArgs := []string{redisTemplate, "--name", redisAppName, "-p", "MEMORY_LIMIT=128Mi"}
+
+				// Redis deployment with the redis service
+				g.By("instantiate the test application")
+				err := oc.Run("new-app").Args(newAppRedisEphemeralArgs...).Execute()
+				o.Expect(err).NotTo(o.HaveOccurred())
+				err = exutil.WaitForDeploymentConfig(oc.KubeClient(), oc.AppsClient().AppsV1(), oc.Namespace(), redisAppName, 1, false, oc)
+				if err != nil {
+					exutil.DumpApplicationPodLogs(redisAppName, oc)
+				}
+
+				// Redis headless service and jenkinsFile which runs verify service on both the services
 				g.By("create the jenkins pipeline strategy build config that leverages openshift client plugin")
-				err := oc.Run("new-app").Args("-f", verifyServiceClientPluginPipelinePath).Execute()
+				err = oc.Run("new-app").Args("-f", verifyServiceClientPluginPipelinePath).Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
 
-				err = exutil.WaitForDeploymentConfig(oc.KubeClient(), oc.AppsClient().AppsV1(), oc.Namespace(), "nodejs-example", 1, false, oc)
-				if err != nil {
-					exutil.DumpApplicationPodLogs("nodejs-example", oc)
-				}
-
-				brservices, err := exutil.StartBuildAndWait(oc, "jenkins-verifyservice-pipeline")
+				brservices, err := exutil.StartBuildAndWait(oc, verifyServiceBuildConfig)
 				if err != nil || !brservices.BuildSuccess {
-					debugAnyJenkinsFailure(brservices, oc.Namespace()+"-jenkins-verifyservice-pipeline", oc, true)
+					debugAnyJenkinsFailure(brservices, oc.Namespace()+"-"+verifyServiceBuildConfig, oc, true)
 					exutil.DumpBuilds(oc)
-					exutil.DumpDeploymentLogs("nodejs-example", 1, oc)
-					exutil.DumpBuildLogs("jenkins-verifyservice-pipeline", oc)
+					exutil.DumpDeploymentLogs(redisAppName, 1, oc)
+					exutil.DumpBuildLogs(verifyServiceBuildConfig, oc)
 				}
 				brservices.AssertSuccess()
+
 				g.By("get build console logs and see if succeeded")
 				_, err = j.GetJobConsoleLogsAndMatchViaBuildResult(brservices, "Finished: SUCCESS")
 				o.Expect(err).NotTo(o.HaveOccurred())
 				g.By("clean up openshift resources for next potential run")
-				err = oc.Run("delete").Args("bc", "jenkins-verifyservice-pipeline").Execute()
+				err = oc.Run("delete").Args("bc", verifyServiceBuildConfig).Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				err = oc.Run("delete").Args("dc", "nodejs-example").Execute()
+				err = oc.Run("delete").Args("dc", redisAppName).Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				err = oc.AsAdmin().Run("delete").Args("all", "-l", "app=nodejs-example").Execute()
+				err = oc.AsAdmin().Run("delete").Args("all", "-l", fmt.Sprintf("app=%v", redisAppName)).Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 			})
--- a/test/extended/testdata/verifyservice-pipeline-template.yaml
+++ b/test/extended/testdata/verifyservice-pipeline-template.yaml
@@ -1,81 +1,28 @@
 apiVersion: v1
 kind: Template
 labels:
-  template: nodejs-example
+  template: jenkins-verifyservice-pipeline
 metadata:
-  name: nodejs-example
+  name: redis-verifyservice-test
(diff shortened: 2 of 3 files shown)


test/extended/builds/pipeline_origin_bld.go
```diff
@@ -274,33 +274,41 @@ var _ = g.Describe("[Feature:Builds][Feature:Jenkins][Slow] openshift pipeline b
 			})
 
 			g.By("should verify services successfully", func() {
-				// instantiate the bc
-				g.By("create the jenkins pipeline strategy build config that leverages openshift client plugin")
-				err := oc.Run("new-app").Args("-f", verifyServiceClientPluginPipelinePath).Execute()
-				o.Expect(err).NotTo(o.HaveOccurred())
 
-				err = exutil.WaitForDeploymentConfig(oc.KubeClient(), oc.AppsClient().AppsV1(), oc.Namespace(), "nodejs-example", 1, false, oc)
+				newTestAppArgs := []string{"--template", "redis-ephemeral", "-p", "MEMORY_LIMIT=128Mi"}
```
akram: minor change, but, can you just rename the newTestAppArgs into newAppRedisEphemeralArgs?

akram: Can you also add the --name=redis parameter, just to ensure that the name is explicitly set here? You can use a local variable to no repeat the hardcoded name in the subsequent calls.

test/extended/builds/pipeline_origin_bld.go
```diff
@@ -274,33 +274,41 @@ var _ = g.Describe("[Feature:Builds][Feature:Jenkins][Slow] openshift pipeline b
...
+
+				// Redis deployment with the redis service
+				g.By("instantiate the test application")
+				err := oc.Run("new-app").Args(newTestAppArgs...).Execute()
+				o.Expect(err).NotTo(o.HaveOccurred())
+				err = exutil.WaitForDeploymentConfig(oc.KubeClient(), oc.AppsClient().AppsV1(), oc.Namespace(), "redis", 1, false, oc)
 				if err != nil {
-					exutil.DumpApplicationPodLogs("nodejs-example", oc)
+					exutil.DumpApplicationPodLogs("redis", oc)
```
akram: like here, yep, use a local variable: redisAppName for example.

test/extended/builds/pipeline_origin_bld.go
```diff
@@ -274,33 +274,41 @@ var _ = g.Describe("[Feature:Builds][Feature:Jenkins][Slow] openshift pipeline b
...
+				g.By("create the jenkins pipeline strategy build config that leverages openshift client plugin")
+				err = oc.Run("new-app").Args("-f", verifyServiceClientPluginPipelinePath).Execute()
+				o.Expect(err).NotTo(o.HaveOccurred())
+
 				brservices, err := exutil.StartBuildAndWait(oc, "jenkins-verifyservice-pipeline")
 				if err != nil || !brservices.BuildSuccess {
 					debugAnyJenkinsFailure(brservices, oc.Namespace()+"-jenkins-verifyservice-pipeline", oc, true)
 					exutil.DumpBuilds(oc)
-					exutil.DumpDeploymentLogs("nodejs-example", 1, oc)
+					exutil.DumpDeploymentLogs("redis", 1, oc)
```
akram: use the local variable name here also.

test/extended/builds/pipeline_origin_bld.go
```diff
@@ -274,33 +274,41 @@ var _ = g.Describe("[Feature:Builds][Feature:Jenkins][Slow] openshift pipeline b
...
 				_, err = j.GetJobConsoleLogsAndMatchViaBuildResult(brservices, "Finished: SUCCESS")
 				o.Expect(err).NotTo(o.HaveOccurred())
 				g.By("clean up openshift resources for next potential run")
 				err = oc.Run("delete").Args("bc", "jenkins-verifyservice-pipeline").Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				err = oc.Run("delete").Args("dc", "nodejs-example").Execute()
+				err = oc.Run("delete").Args("dc", "redis").Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				err = oc.AsAdmin().Run("delete").Args("all", "-l", "app=nodejs-example").Execute()
+				err = oc.AsAdmin().Run("delete").Args("all", "-l", "app=redis-ephemeral").Execute()
```
akram: this one needs to be fixed: Change to app=redis or better to "app=" + redisAppName .

also, note that local imageStreams in the namespace are not holding the level.
so, maybe add an additional `oc delete is redis` or `oc delete is --all`.

test/extended/builds/pipeline_origin_bld.go
```diff
@@ -274,33 +274,42 @@ var _ = g.Describe("[Feature:Builds][Feature:Jenkins][Slow] openshift pipeline b
...
 				_, err = j.GetJobConsoleLogsAndMatchViaBuildResult(brservices, "Finished: SUCCESS")
 				o.Expect(err).NotTo(o.HaveOccurred())
 				g.By("clean up openshift resources for next potential run")
 				err = oc.Run("delete").Args("bc", "jenkins-verifyservice-pipeline").Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				err = oc.Run("delete").Args("dc", "nodejs-example").Execute()
+				err = oc.Run("delete").Args("dc", redisAppName).Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				err = oc.AsAdmin().Run("delete").Args("all", "-l", "app=nodejs-example").Execute()
+				err = oc.AsAdmin().Run("delete").Args("all", "-l", fmt.Sprintf("app=%v", redisTemplate)).Execute()
```
akram: this one should be redisAppName not redisTemplate.

test/extended/testdata/verifyservice-pipeline-template.yaml
```diff
@@ -1,81 +1,28 @@
...
-          name: nodejs:latest
-          namespace: openshift
-      type: ImageChange
-    - type: ConfigChange
+      name: redis
 - apiVersion: v1
   kind: BuildConfig
   metadata:
     name: jenkins-verifyservice-pipeline
+    app: redis-ephemeral
```
akram: All the other components will be create with app=redis and not app=redis-ephemeral

gabemontero:
yeah adding the test data will require @adambkaplan (perhaps another candidate for OWNERS update)

gabemontero:
@waveywaves there is an error with your changes in the latest e2e-aws-builds run ... I'll refrain from highlighting it for now ... see if you can sort it out.

test/extended/builds/pipeline_origin_bld.go
```diff
@@ -274,33 +274,42 @@ var _ = g.Describe("[Feature:Builds][Feature:Jenkins][Slow] openshift pipeline b
 			})
 
 			g.By("should verify services successfully", func() {
-				// instantiate the bc
-				g.By("create the jenkins pipeline strategy build config that leverages openshift client plugin")
-				err := oc.Run("new-app").Args("-f", verifyServiceClientPluginPipelinePath).Execute()
-				o.Expect(err).NotTo(o.HaveOccurred())
+				redisTemplate := "redis-ephemeral"
+				redisAppName := "redis"
+				newAppRedisEphemeralArgs := []string{"--template", fmt.Sprintf("--name=%v", redisAppName), redisTemplate, "-p", "MEMORY_LIMIT=128Mi"}
```
akram: The syntax is not correct, it should be:

```
newAppRedisEphemeralArgs := []string{ fmt.Sprintf("--template=%v", redisTemplate),  fmt.Sprintf("--name=%v", redisAppName), "-p", "MEMORY_LIMIT=128Mi"}
```

or simplier:

```
newAppRedisEphemeralArgs := []string{ redisTemplate, "--name", redisAppName), "-p", "MEMORY_LIMIT=128Mi"}
```

akram:
/assign @bparees 

As of @adambkaplan is on PTO, can you approve this one ? 
I think it was breaking origin e2e tests, so the sooner the 
Thanks

bparees:
/approve

but i would like to understand this:

> Fixed as containers were going to crashloopbackoff.
> The image loaded before was an s2i image without the source.

why was the original nodejs image crashlooping? was the nodejs image changed such that it can no longer be run without source code being added?

waveywaves:
> /approve
> 
> but i would like to understand this:
> 
> > Fixed as containers were going to crashloopbackoff.
> > The image loaded before was an s2i image without the source.
> 
> why was the original nodejs image crashlooping? was the nodejs image changed such that it can no longer be run without source code being added?

It is an S2I image with no source code and nothing defined as the entrypoint. To make the container run with the image, commands need ot be given in the Deployment for it to be able to run. Hence we had to fall bakc to using a predefined template.

bparees:
> It is an S2I image with no source code and nothing defined as the entrypoint. To make the container run with the image, commands need ot be given in the Deployment. Hence we had to fall bakc to using a predefined template.

yes but presumably this test was working at some point, so why did it start failing?

waveywaves:
> yes but presumably this test was working at some point, so why did it start failing?

After testing out the deployment config seperately, it did fail with a crashloopback off. When the changes were pushed to github, the tests which ran did not fail and seems like changes in these tests weren't picked up by the CI unlike travisCi where the changes in the PR are applied to the test env.

bparees:
hm.  i'm still a little puzzled but it sounds like this test basically wasn't being run normally so we didn't notice it was broken.

PR #23595, opened 2019-08-12
test/extended: set ErrorPassthrough true for URL checker

abhinavdahiya:
based on https://github.com/openshift/origin/blob/1f707e35708112ac19446570268f0b916578119a/test/extended/util/url/url.go#L80
the test fails on first invalid response that cannot be decoded rather than retying for expected time in Within.

Esp. when the DNS resolution hasn't propagated, which we are seeing on GCP and Azure.. empty response causes the test to immediatly fail with error like
```
response 0 could not be decoded: invalid character '0' after object key:value pair
```
With passthrough set to true, the test should keep retrying.


ironcladlou:
/approve

Let's try it, thanks!

PR #23613, opened 2019-08-15
os/Dockerfile: Use dnf, pin repos, and clean cache

ashcrow:
- Clean the dnf/yum cache before running commands
- Pin Fedora repos
- Call dnf directly instead of yum

### DIFF
--- a/images/os/Dockerfile
+++ b/images/os/Dockerfile
@@ -1,7 +1,12 @@
 FROM fedora:29 AS build
 
 COPY --from=registry.svc.ci.openshift.org/openshift/origin-v4.0:machine-os-content /srv/ /srv/
-RUN set -x && yum install -y ostree yum-utils selinux-policy-targeted && \
+# Don't use mirrors to avoid errors in build
+COPY repos/* /etc/yum.repos.d/
+RUN set -x && \
+    dnf clean all && \
+    rm -rf /var/cache/dnf/* && \
+    dnf install -y ostree yum-utils selinux-policy-targeted && \
     commit=$( find /srv -name *.commit | sed -Ee 's|.*objects/(.+)/(.+)\.commit|\1\2|' | head -1 ) && \
     mkdir /tmp/working && cd /tmp/working && \
     yumdownloader --enablerepo=built --destdir=/tmp/rpms openshift-hyperkube openshift-clients && \
--- a/images/os/repos/fedora-modular.repo
+++ b/images/os/repos/fedora-modular.repo
@@ -1,0 +1,11 @@
+[fedora-modular]
+name=Fedora Modular $releasever - $basearch
+failovermethod=priority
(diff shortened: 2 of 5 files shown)


ashcrow:
```
Error getting repository data for built, repository not found
```

ashcrow:
I'm looking to see where `built.repo` is defined. I'm wondering if it may have been accidentally removed somewhere but is still expected by this build.

ashcrow:
Looks like `f29` required in the name for other dependencies in the CI set up. Switched back to f29 to see if it passes.

ashcrow:
Looking to pin repos instead of using mirrors.

ashcrow:
The image builds fine, just need to get past the other CI targets.

ashcrow:
`no such image` :confused: 

/test e2e-cmd

jlebon:
Is the motivation here using mirrors because they can flake? If a build observes many of them failing in a single run, there might something else wrong which using a specific mirror won't fix.

ashcrow:
@jlebon more or less yes. The Fedora mirrors have been flaky at random times causing significant failures.

ashcrow:
`s: "error waiting for the pod 'test-cmd' to complete: the pod errored trying to run the command",`

/test e2e-cmd

ashcrow:
Failed: ```
[Suite:openshift/test-cmd][Serial][Disruptive] test-cmd: test/cmd/policy-storage-admin.sh [Suite:openshift] expand_less
```

/test e2e-cmd

**Comment by cgwalters** [comment]
Ultimately though, if we're going to use Fedora content in CI, we need to make the mirroring more robust, not work around it at every call site.

That said another way to work around this would be changing the system to use coreos-assembler instead like https://github.com/openshift/machine-config-operator/pull/682

**Comment by ashcrow** [comment]
> Ultimately though, if we're going to use Fedora content in CI, we need to make the mirroring more robust, not work around it at every call site.

True. This PR is attempting to work around the pain for the short term, not a forever fix.

**Comment by vrutkovs** [comment]
```
FAILURE after 0.000s: test/cmd/basicresources.sh:175: executing 'oc set probe -f /var/tests/test/cmd/testdata/hello-openshift/hello-pod.json --local -o yaml --liveness --get-url=https://127.0.0.1:port/path' expecting success and text 'port: port': the command returned the wrong error code; the output content test failed
Standard error from the command:
error: --get-url could not be parsed as a valid URL: parse https://127.0.0.1:port/path: invalid port ":port" after host
[ERROR] hack/lib/cmd.sh:241: `return "${return_code}"` exited with status 1.
```
/retest

jlebon:
I guess if it helps in the short-term...
/approve

(But again, it's not clear to me if this would fix the original issue. IIRC, multiple mirrors were failing in a single run, right?)

PR #23621, opened 2019-08-16
Bug 1741532: tests/router: remove redundant and flaky reencrypt test

ironcladlou:
The specific router reencrypt test here doesn't seem to be adding any value over
the existing test in routable.go which verifies system routes (one of which is
the console, whose route topology looks similar to what this test wants).

The test also relies on a public nginx image and in general has a relatively
complex test setup, but the value gained over routable.go is unclear.

Since the routable.go test is also flaking on Azure/GCP, removing this redundant
test could increase the signal to noise ratio around router tests generally.

@smarterclayton @openshift/sig-network-edge for your consideration


Miciah:
I'm hoping https://github.com/openshift/origin/pull/23622 will fix the test failure.

smarterclayton:
Console uses reencrpyt?  Can you add a comment to that test that it is used as a verifier so no one accidentally removes it?

Do we know why this flakes?

PR #23626, opened 2019-08-19
Remove oc from origin

deads2k:
`oc` and it's RPMs come from openshift/oc now.  This removes the false dependency


tnozicka:
oc rpm is out (https://github.com/openshift/origin/pull/23567), wanna rebase?

tnozicka:
(I don't see conflict here per say)
/retest

PR #23628, opened 2019-08-19
Bug 1642530 - Custom Resource is stuck if deleted repeatedly with PropagationPolicy Foreground

p0lyn0mial:
backports https://github.com/kubernetes/kubernetes/pull/81081 and https://github.com/kubernetes/kubernetes/pull/78650 (for `hasFinalizer` method)


p0lyn0mial:
please check the commit message.

PR #23629, opened 2019-08-19
fix broken link to kinds-of-tests

pruan-rht:


pruan-rht:
@adambkaplan can you PTAL, it's just a simple fix to a dead link.  Thanks!

PR #23630, opened 2019-08-19
e2e: test ingress controller prometheus integration

ironcladlou:
Add an e2e test which ensures that ingress controllers are successfully scraped
by Prometheus. Cover the default ingresscontroller and a non-default instance.

cc @smarterclayton @openshift/sig-network-edge

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -12,20 +12,26 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 
 	v1 "k8s.io/api/core/v1"
+
 	kapierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+
 	clientset "k8s.io/client-go/kubernetes"
 	watchtools "k8s.io/client-go/tools/watch"
+
 	kapi "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/client/conditions"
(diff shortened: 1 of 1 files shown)


ironcladlou:
My last run of this single test against an AWS cluster took 2m45s. The `servicemonitor` created by the ingress operator was 104s old when Prometheus started reporting the job for the first time.

smarterclayton:
Test looks ok to me, although it’s only verifying we get the target, not that a scrape returned any metric.  I’d like to ensure we scrape at least one expected metric from those jobs (something stats related, possibly from the health endpoint).

ironcladlou:
>Test looks ok to me, although it’s only verifying we get the target, not that a scrape returned any metric. I’d like to ensure we scrape at least one expected metric from those jobs (something stats related, possibly from the health endpoint).

Good idea, I can do that

smarterclayton:
that's fine, that's not particularly long

ironcladlou:
@smarterclayton Seeking a good proxy for readiness from existing stats... the readiness endpoint is backed by haproxy's `monitor-uri` endpoint combined with a check for whether the router has ever synced, and AFAICT neither of those conditions are exposed by any existing metric.

Maybe a rough approximation would be to check:

```
haproxy_server_up{namespace="",pod="",route="",server="fe_sni",service=""} 1
template_router_reload_seconds_count 33
```

`haproxy_server_up` would be a proxy for haproxy readiness, `template_router_reload_seconds_count` for the sync check. Probably a lossy alternative. Good? Other ideas?

ironcladlou:
To clarify, the rules would be:

```
haproxy_server_up == 1
template_router_reload_seconds_count > 0
```

ironcladlou:
@smarterclayton bb42dfa adds some assertions on metrics keys which were the closest proxy for readiness I could easily find.

smarterclayton:
Those metrics look fine to me.

ironcladlou:
Test failures have so far been unrelated.

@smarterclayton any other feedback before tagging?

/test all

PR #23636, opened 2019-08-20
Bug 1743881: Use dns go bindings instead of dig command for dns resolution in egress networkpolicies

juanluisvaladas:
Use dns go bindings instead of dig command for dns resolution in egress networkpolicies

- Query all nameservers in /etc/resolv.conf to populate ovs rules for domain names in egress policy
  When the given domain is using DNS load balancing algorithm, depending on which nameserver it queries,
  we may get different IP addr.

This is a backport to 3.11 of https://github.com/openshift/sdn/commit/4c01b7ba5023a49e541fccdeb687df538e0d6f44 , however there were a few conflicts and it's not exactly the same.

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -11,6 +11,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	kexec "k8s.io/utils/exec"
 )
 
 const (
@@ -27,6 +28,9 @@ type dnsValue struct {
 }
 
 type DNS struct {
+	// Runs shell commands
+	execer kexec.Interface
+
 	// Protects dnsMap operations
 	lock sync.Mutex
 	// Holds dns name and its corresponding information
@@ -38,13 +42,14 @@ type DNS struct {
 	port string
 }
 
-func NewDNS(resolverConfigFile string) (*DNS, error) {
+func NewDNS(execer kexec.Interface, resolverConfigFile string) (*DNS, error) {
 	config, err := dns.ClientConfigFromFile(resolverConfigFile)
 	if err != nil || config == nil {
 		return nil, fmt.Errorf("cannot initialize the resolver: %v", err)
 	}
 
 	return &DNS{
+		execer:      execer,
 		dnsMap:      map[string]dnsValue{},
 		nameservers: filterIPv4Servers(config.Servers),
 		port:        config.Port,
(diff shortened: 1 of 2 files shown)


pkg/network/common/dns.go
```diff
@@ -11,6 +11,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	kexec "k8s.io/utils/exec"
```
juanluisvaladas: Part of the conflict, not quite sure if I should have replaced os/exec for this

pkg/network/common/dns.go
```diff
@@ -27,6 +28,9 @@ type dnsValue struct {
 }
 
 type DNS struct {
+	// Runs shell commands
+	execer kexec.Interface
```
juanluisvaladas: Again part of the conflict, not quite sure if there is any advantage of replacing exec for kexec

pkg/network/common/dns.go
```diff
@@ -38,13 +42,14 @@ type DNS struct {
 	port string
 }
 
-func NewDNS(resolverConfigFile string) (*DNS, error) {
+func NewDNS(execer kexec.Interface, resolverConfigFile string) (*DNS, error) {
```
juanluisvaladas: Same as before, not quite sure if there is a point on this

pkg/network/common/dns.go
```diff
@@ -38,13 +42,14 @@ type DNS struct {
...
 	config, err := dns.ClientConfigFromFile(resolverConfigFile)
 	if err != nil || config == nil {
 		return nil, fmt.Errorf("cannot initialize the resolver: %v", err)
 	}
 
 	return &DNS{
+		execer:      execer,
```
juanluisvaladas: And the same

pkg/network/common/egress_dns.go
```diff
@@ -41,7 +42,7 @@ func NewEgressDNS() *EgressDNS {
 }
 
 func (e *EgressDNS) Add(policy networkapi.EgressNetworkPolicy) {
-	dnsInfo, err := NewDNS("/etc/resolv.conf")
+	dnsInfo, err := NewDNS(kexec.New(), "/etc/resolv.conf")
```
juanluisvaladas: again the kexec.

juanluisvaladas:
/retitle Bug 1743881: Use dns go bindings instead of dig command for dns resolution in egress networkpolicies

juanluisvaladas:
/test verify
Not quite sure if a flake or if failing because I modified the DNS struct. It's not in types.go so AFAIK it shouldn't break anything related to swagger.

juanluisvaladas:
Wrong branch. Sorry.

PR #23637, opened 2019-08-20
Add a test to ensure that provider API version is not in k8s.io

mrunalp:
@derekwaynecarr ptal to see if this works.

cc: @ingvagabund 
Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

### DIFF
--- a/test/extended/machines/machines.go
+++ b/test/extended/machines/machines.go
@@ -98,6 +98,61 @@ var _ = g.Describe("[Feature:Machines][Smoke] Managed cluster should", func() {
 			e2e.Failf("Machine resources missing for nodes: %s", strings.Join(nodeNames.List(), ", "))
 		}
 	})
+
+	g.It("not have machine resources with provider APIVersion in k8s.io namespace", func() {
+		cfg, err := e2e.LoadConfig()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		c, err := e2e.LoadClientset()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		dc, err := dynamic.NewForConfig(cfg)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("checking for the openshift machine api operator")
+		// TODO: skip if platform != aws
+		skipUnlessMachineAPIOperator(c.CoreV1().Namespaces())
+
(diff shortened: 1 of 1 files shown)


test/extended/machines/machines.go
```diff
@@ -98,6 +98,61 @@ var _ = g.Describe("[Feature:Machines][Smoke] Managed cluster should", func() {
...
+			items := objects(machines.Get("items"))
+			if len(items) == 0 {
+				e2e.Logf("No machine objects found")
+				return true, nil
+			}
+
+			for _, machine := range items {
+				providerAPIVersion := providerAPIVersionFromMachine(machine)
+				e2e.Logf(providerAPIVersion)
+				if strings.Contains(providerAPIVersion, "k8s.io") {
```
ingvagabund: Can we go with `!strings.Contains(providerAPIVersion, "openshift.io")` unless we plan to also allow groups other than `openshift.io`.

ingvagabund:
Checking test logs [1], I don't see `[Feature:Machines][Smoke] Managed cluster should not have machine resources with provider APIVersion in k8s.io namespace` running.

[1] https://storage.googleapis.com/origin-ci-test/pr-logs/pull/23637/pull-ci-openshift-origin-master-e2e-aws-serial/9474/artifacts/e2e-aws-serial/container-logs/test.log

mrunalp:
@ingvagabund It is running as part of ` [Suite:openshift/conformance/parallel] ` in `e2e-aws`. See logs here  https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/23637/pull-ci-openshift-origin-master-e2e-aws/12221.

test/extended/machines/machines.go
mrunalp: I think one of the metal providers isn't running in openshift.io namespace. Hence, I went with the not in k8s.io logic.

ingvagabund: In that case I would still go with white list rather than black list. We can always extend the white list later. Btw. what's the metal provider group? @cynepco3hahue do you know?

cynepco3hahue: `metal3.io`

PR #23640, opened 2019-08-21
Bug 1734673: UPSTREAM: 81306: Scheduler should terminate on loosing leader lock

soltysh:
This is a backport of kubernetes PR that Ravi pointed out in https://bugzilla.redhat.com/show_bug.cgi?id=1734673#c17

/assign @damemi


damemi:
/lgtm
e2e-cmd still flaking it looks like

PR #23620, opened 2019-08-16
Bug 1741679: Bump cadvisor [3.11.z]

sjenning:
master PR https://github.com/openshift/origin/pull/23585
4.1.z PR https://github.com/openshift/origin/pull/23619

pull in https://github.com/openshift/google-cadvisor/pull/7

The delta is larger because a `make update-deps` on a clean `origin/master` makes a lot of changes and the `github.com/apcera/gssapi` dep needed to be redirected to the openshift fork because that repo doesn't exist anymore.  I've split up the bump(*) on the clean master into a separate commit.

Fixes #23492

@derekwaynecarr @deads2k @mfojtik

PR #23646, opened 2019-08-22
Fix gssapi dependency

gnufied:
Use openshift fork

cc @enj @eparis


enj:
@sjenning already has this in #23620

gnufied:
Closing in favour of @sjenning PR. 

/close

PR #23651, opened 2019-08-22
UPSTREAM: 81776: apimachinery: hide 'suppressing panic for copyResponse' error' in ReverseProxy

sttts:
The Golang ReverseProxy prints out `suppressing panic for copyResponse error in test; copy error: context canceled` on termination of the process. This is highly misleading in logs and people think this is fatal, leading to lots of noise when debugging systems.

Backport of https://github.com/kubernetes/kubernetes/pull/81776


sttts:
/hold

Let's check the logs that it is effective.

sttts:
Verified.

/hold cancel

PR #23658, opened 2019-08-23
Bug 1718878: remove commented code

dmage:
There is no reason to have commented out code. We'll restore this code if we decided to implement this feature.


adambkaplan:
@dmage per this PR deleting layer links has not been implemented, correct?

dmage:
@adambkaplan correct.

PR #23659, opened 2019-08-23
Add additional bit of information from pod status when it fails running

soltysh:
I've seen in the logs:
```
fail [github.com/openshift/origin/test/extended/cmd/cmd.go:117]: Expected
    <[]error | len:1, cap:1>: [
        {
            s: "error waiting for the pod 'test-cmd' to complete: the pod errored trying to run the command",
        },
    ]
to have length 0
```
several times, I'd like to know why it failed, so I'm adding pod status to that error. 
/assign @tnozicka

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -1563,7 +1563,7 @@ func RunOneShotCommandPod(
 		}
 
 		if podHasErrored(cmdPod) {
-			return true, fmt.Errorf("the pod errored trying to run the command")
+			return true, fmt.Errorf("the pod errored trying to run the command: %#v", cmdPod.Status)
 		}
 		return podHasCompleted(cmdPod), nil
 	})


soltysh:
/hold
let's try to gather some data in this PR, if not we'll merge it in.

test/extended/util/framework.go
```diff
@@ -1563,7 +1563,7 @@ func RunOneShotCommandPod(
 		}
 
 		if podHasErrored(cmdPod) {
-			return true, fmt.Errorf("the pod errored trying to run the command")
+			return true, fmt.Errorf("the pod errored trying to run the command: %#v", pod.Status)
```
tnozicka: cmdPod ?

JAORMX: Even if it referes to the same pod, it does sound like it would be cmdPod instead; given that it would contain the latest status (perhaps?)

JAORMX:
LGTM (not sure if I'm allowed to actually vote on this, hence why I didn't use the "/")

soltysh:
> LGTM (not sure if I'm allowed to actually vote on this, hence why I didn't use the "/")

It's not a vote question, you need to be part of the openshift team to be allowed to tag PRs. Approvals are coming from files in each directory, or its parents.

test/extended/util/framework.go
```diff
@@ -1563,7 +1563,7 @@ func RunOneShotCommandPod(
 		}
 
 		if podHasErrored(cmdPod) {
-			return true, fmt.Errorf("the pod errored trying to run the command")
+			return true, fmt.Errorf("the pod errored trying to run the command: %#v", cmdPod.Status)
```
tnozicka: /lgtm cancel

use spew to get full dump `Terminated:(*v1.ContainerStateTerminated)(0xc002a40230)`

JAORMX:
> It's not a vote question, you need to be part of the openshift team to be allowed to tag PRs. Approvals are coming from files in each directory, or its parents.

ack

soltysh:
I'm going to close this for now. Will re-open when needed.

PR #23660, opened 2019-08-23
Bug 1744995: Request headers fix x509 [4.1]

stlaz:
We are using kube authnentication to wrap the oauth-server
endpoints and, typically, let the oauth-server itself fetch
the clientCA from the cluster for cluster admin
authenticatication.

When using Request Header identity provider, we also need to use
client cert authentication. However, oauth-server's own authentication
is logically preceded by the kube one. Since this one is configured
to fail on errors, and the client CA for Request Header IdPs is
usually different from the one that issued cluster admin cert,
the kube cert authnetication fails sooner than the request is passed
to the oauth-server.

The behavior is fixed here by using a union authenticator of the
kube authenticator and an authenticator that will ignore kube
authentication errors on paths that any user is authorized to access
(as configured for the delegated authorization).

### DIFF
--- a/pkg/cmd/openshift-integrated-oauth-server/server.go
+++ b/pkg/cmd/openshift-integrated-oauth-server/server.go
@@ -2,8 +2,14 @@ package openshift_integrated_oauth_server
 
 import (
 	"errors"
+	"net/http"
 
+	"k8s.io/apiserver/pkg/authentication/authenticator"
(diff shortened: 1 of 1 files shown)


stlaz:
cc @enj @deads2k 
This is a backport of https://github.com/openshift/oauth-server/pull/8

enj:
The 4.2 change was manually tested but we need #23662 to keep it from regressing in the future.

PR #23661, opened 2019-08-23
Bug 1734385: [3.11] Don't use strategic merge patch on Node.Status.Addresses

danwinship:
Backport of #23345 / #23367 to release-3.11 for a customer. (Their symptom is slightly different, involving IPv4 vs IPv6 addresses, but it appears to be the same underlying bug.)

/cc @RobertKrawitz @sjenning


danwinship:
e2e-gcp-crio seems to have been broken since forever. can someone big-green-button this?

PR #23674, opened 2019-08-26
1.14.6 rebase

deads2k:
About 6 hours

Updates to 1.14.6 and eliminates patches that are only needed by other repos.

Picklist: https://docs.google.com/spreadsheets/d/12txg87Jq-BRfPsgMYWgdy_eH94lVOwGtseIe4OWPmZ0/edit#gid=1413655187


deads2k:
/hold

to ensure this isn't merged until we re-check the carry list

derekwaynecarr:
agree we should take the latest upstream prior to 4.2.0.

mfojtik:
/lgtm

Please open a follow up BZ for the file locking issue

/hold

Holding so @soltysh and @sttts can take a stab

vendor/k8s.io/kubernetes/pkg/kubectl/cmd/config/config_test.go
```diff
@@ -87,6 +87,7 @@ func TestCurrentContext(t *testing.T) {
 }
 
 func TestSetCurrentContext(t *testing.T) {
+	t.Skip()
```
sttts: document in the commit message why

glide.yaml
```diff
@@ -71,7 +71,7 @@ import:
   version: origin-4.2-kubernetes-1.14.0
```
sttts: while probably not super important. Would switch all of them.

sttts:
Looks good overall. Some nits.

deads2k:
/hold

holding again because there's a new upstream commit

vendor/k8s.io/kubernetes/pkg/kubectl/cmd/config/config_test.go
deads2k: > document in the commit message why

removed that commit.  Found the lock file culprit

deads2k:
/hold cancel

updated

PR #23689, opened 2019-08-28
Added the .git extension to the examples in --help

joellord:
Added the .git extension to the examples that had a github repo for the

oc new-app --help

command.

I've had some errors and only realized later that I needed to add the .git extension to my repositories in order for new-app to work. 
This does not apply to new-build which will work with or without the .git extension.


soltysh:
oc lives in https://github.com/openshift/oc please provide any fixes against that repository.
/close

PR #23693, opened 2019-08-29
Bug 1746870: Do not cache egressLink details

rcarrillocruz:


eparis:
Seems like months without progress here. Is this going to get fixed in master? It failed QE. there, right?

PR #23694, opened 2019-08-29
test/extended/prometheus: Remove machine api operator metrics test

LiliC:
This test was failing because machine-api-operator component is not
installed everywhere, so we often see this test failing. We will remove
it for now, and rethink the test in a future PR.

cc @s-urbaniak @paulfantom @smarterclayton

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -217,17 +217,6 @@ var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
 			}
 			runQueries(tests, oc, ns, execPodName, url, bearerToken)
 		})
-		g.It("should have machine api operator metrics", func() {
-			oc.SetupProject()
-			ns := oc.Namespace()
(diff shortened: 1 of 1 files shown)


enxebre:
Machine API Operator is deployed in UPI,l it's just no-operating. https://github.com/openshift/origin/pull/23694/files#diff-68b78becedd34dfeef78260ab7c23952L227 should be updated to validate the collector is up https://github.com/openshift/machine-api-operator/blob/master/pkg/metrics/metrics.go#L32 instead of expecting a machineSet cc @vikaschoudhary16 @ingvagabund

s-urbaniak:
@enxebre as this e2e test is not a cross component metrics test, we advocate that it should rather live either in a separate directory owned by the machine API team i.e. in `test/extended/machine-api` or in the machine API operator repository and invoked as part of that e2e test suite.

enxebre:
https://github.com/openshift/origin/tree/master/test/extended/machines sounds good

PR #23698, opened 2019-08-29
Bug 1746499: Ensure images push without full tag spec

adambkaplan:
If an image push doesn't have a tag, ensure that `:latest` is appended and doesn't fail due to an "identifier is not an image" error.


adambkaplan:
Needs openshift/builder#92 to land in order to pass

adambkaplan:
@openshift/openshift-team-developer-experience test is passing now that openshift/builder#92 merged. ptal

PR #23707, opened 2019-09-02
WIP: Rebase 1.16-beta.1

sttts:


test/extended/util/framework.go
```diff
@@ -1446,9 +1454,62 @@ func ParseLabelsOrDie(str string) labels.Selector {
 	return ret
 }
 
+// LaunchWebserverPod launches a pod serving http on port 8080 to act
+// as the target for networking connectivity checks.  The ip address
+// of the created pod will be returned if the pod is launched
+// successfully.
+func LaunchWebserverPod(f *e2e.Framework, podName, nodeName string) (ip string) {
```
deads2k: used by what?

mrunalp:
Can we add this diff to pull in cri-o 1.16?

```
diff --git a/images/os/Dockerfile b/images/os/Dockerfile
index a22f956e75..81c1f56915 100644
--- a/images/os/Dockerfile
+++ b/images/os/Dockerfile
@@ -6,6 +6,7 @@ RUN set -x && yum install -y ostree yum-utils selinux-policy-targeted && \
     mkdir /tmp/working && cd /tmp/working && \
     PACKAGES=(openshift-hyperkube) && \
     yumdownloader --enablerepo=built --destdir=/tmp/rpms "${PACKAGES[@]}"  && \
+    wget -P /tmp/rpms/ https://lsm5.fedorapeople.org/cri-o-1.16.0-0.1.dev.rhaos4.3.git35bae61.el8.x86_64.rpm && \
     ls /tmp/rpms/ && (cd /tmp/rpms/ && ls ${PACKAGES[@]/%/*}) && \
     for i in $(find /tmp/rpms/ -name openshift-* -iname *.rpm); do echo "Extracting $i ..."; rpm2cpio $i | cpio -div; done && \
     if [[ -d etc ]]; then mv etc usr/; fi && \
```

mrunalp:
https://github.com/cri-o/cri-o/pull/2770 should fix it for 1.14 if we want to wait for that instead.

test/extended/util/framework.go
sttts: networking e2e

/cc @squeed

rphillips:
as discussed in slack, we might need to add a 'selinux' golang build tag because of this [commit](https://github.com/kubernetes/kubernetes/commit/51177c1ebbb105e354e114a56fa12cc2aee0993e#diff-6b0572fcf7dadf28d35ae2f18847035fR1).

test/extended/networking/util.go
```diff
@@ -203,7 +186,7 @@ func checkConnectivityToHost(f *e2e.Framework, nodeName string, podName string,
 	savedErr := err
 
 	// Debug
```
danwinship: This part doesn't work right any more and was only added to debug a long-ago test flake. You can just `return err` above and drop everything from here to the end of the function

test/extended/util/framework.go
```diff
@@ -1446,9 +1454,62 @@ func ParseLabelsOrDie(str string) labels.Selector {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: podName,
+		},
+		Spec: kapiv1.PodSpec{
+			Containers: []kapiv1.Container{
+				{
+					Name:  containerName,
+					Image: image.GetE2EImage(image.Agnhost),
+					Env:   []kapiv1.EnvVar{{Name: fmt.Sprintf("SERVE_PORT_%d", port), Value: "foo"}},
+					Ports: []kapiv1.ContainerPort{{ContainerPort: int32(port)}},
```
danwinship: drop the `Env:`, add

    Args:  []string{"netexec", "--http-port", fmt.Sprintf("%d", port)},

that should fix all the broken `[Area: Networking]` tests

p0lyn0mial:
running one more time as I see a lot of `connect: connection refused` errors

/test e2e-aws

test/extended/networking/util.go
sttts: done

test/extended/util/framework.go
sttts: done

PR #23708, opened 2019-09-02
Bug 1747343: Disable unsupported Ceph tests

jsafrane:
OpenShift 4.x does not support Ceph, Ceph CSI driver should be used instead.

https://docs.openshift.com/container-platform/4.1/storage/understanding-persistent-storage.html#types-of-persistent-volumes_understanding-persistent-storage


jsafrane:
Slightly related to https://github.com/kubernetes/kubernetes/pull/82241
(`ceph` tests flake a bit, https://bugzilla.redhat.com/show_bug.cgi?id=1747343)

bertinatto:
/assign @mfojtik

for approval

PR #23645, opened 2019-08-21
 Bug 1731263: Re-enable preemption test with fixes

damemi:
Rebased changes from https://github.com/openshift/origin/pull/23050

PR #23719, opened 2019-09-03
Bug 1748150: Modify scheduler preemption tests (fixup #23645)

damemi:
This fixes a bug introduced in https://github.com/openshift/origin/pull/23645 by initializing the pod array with a length of 0 and a capacity of 4, due to the fact that some nodes may be unschedulable and thus we can't assume that `len(pods)` will always have 4 pods

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/scheduling/preemption.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/scheduling/preemption.go
@@ -27,7 +27,7 @@ import (
 	. "github.com/onsi/gomega"
 	_ "github.com/stretchr/testify/assert"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


soltysh:
/hold 
it looks like there's a another problem, in line 336 of this file, see:
```
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/openshift/origin/vendor/k8s.io/kubernetes/test/e2e/scheduling.glob..func4.5()
	/go/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/vendor/k8s.io/kubernetes/test/e2e/scheduling/preemption.go:336 +0x15f2
```

soltysh:
@damemi stupid me, when you create a slice like this `make([]*v1.Pod, 0, numPods)` you're actually creating an empty slice with given capacity, this means you need to use `append` to add pods into it.

damemi:
@soltysh good catch, could you /lgtm this again?

PR #23722, opened 2019-09-04
Bug 1748501: UPSTREAM: <carry>: fix openshift admission plugin registration

mfojtik:
Without this, no openshift admission plugin is registered and trying to enable them via unsupported config overrides error out with "unknown" plugin.

This is because of wrong order :-)

Credits to @sttts with debugging help.

/cc @sttts 
/cc @deads2k


sttts:
/lgtm

/hold

for some more eyes.

deads2k:
You'll want to make sure that the CR validation works. That's probably e2e worthy

mfojtik:
@deads2k @sttts said Standa have test

PR #23723, opened 2019-09-04
Machine metric e2e

vikaschoudhary16:
This is a follow-up PR for https://github.com/openshift/origin/pull/23694#issuecomment-527040801

Initially mao metrics e2e was depending on machineset count. This was making test fail for platforms which do not support machinesets. Test was removed in #23694  

In this PR:

- Adding MAO metrics e2e test which depend on whether mao metrics collector is up or not. This would be up for all the platforms. 
- Making some APIs global in prometheus e2e package so that code can be reused in mao metrics e2e

### DIFF
--- a/test/extended/machines/metrics.go
+++ b/test/extended/machines/metrics.go
@@ -1,0 +1,41 @@
+package operators
+
+import (
+	g "github.com/onsi/ginkgo"
+
+	prom "github.com/openshift/origin/test/extended/prometheus"
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = g.Describe("[Feature:Machines][Serial] Prometheus", func() {
+	defer g.GinkgoRecover()
+	var (
+		oc = exutil.NewCLIWithoutNamespace("prometheus")
+
+		url, bearerToken string
+	)
+
+	g.BeforeEach(func() {
+		var ok bool
+		url, bearerToken, ok = prom.LocatePrometheus(oc)
+		if !ok {
+			e2e.Skipf("Prometheus could not be located on this cluster, skipping prometheus test")
+		}
+	})
+	g.Describe("when installed on the cluster", func() {
+		g.It("should have machine api operator metrics", func() {
+			oc.SetupProject()
+			ns := oc.Namespace()
+			execPodName := e2e.CreateExecPodOrFail(oc.AdminKubeClient(), ns, "execpod", func(pod *v1.Pod) { pod.Spec.Containers[0].Image = "centos:7" })
+			defer func() { oc.AdminKubeClient().CoreV1().Pods(ns).Delete(execPodName, metav1.NewDeleteOptions(1)) }()
+
+			tests := map[string][]prom.MetricTest{
+				`mapi_mao_collector_up`: {prom.MetricTest{GreaterThanEqual: true, Value: 1}},
+			}
+			prom.RunQueries(tests, oc, ns, execPodName, url, bearerToken)
+		})
+	})
+})
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -47,7 +47,7 @@ var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
 
 	g.BeforeEach(func() {
 		var ok bool
-		url, bearerToken, ok = locatePrometheus(oc)
+		url, bearerToken, ok = LocatePrometheus(oc)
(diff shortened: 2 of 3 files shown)


test/extended/prometheus/prometheus.go
```diff
@@ -335,12 +335,12 @@ func (t *prometheusTargets) Expect(l labels, health, scrapeURLPattern string) er
 
 type labels map[string]string
 
-func (l labels) With(name, value string) labels {
+func (l labels) With(name, Value string) labels {
```
enxebre: why is this needed?

enxebre:
/lgtm
cc @s-urbaniak @LiliC

s-urbaniak:
sounds good, but you should also loop in @smarterclayton , and add an OWNERS file in `test/extended/machines` as I am not sure how the formal process here is.

test/extended/machines/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+	})
+	g.Describe("when installed on the cluster", func() {
+		g.It("should have machine api operator metrics", func() {
+			oc.SetupProject()
+			ns := oc.Namespace()
+			execPodName := e2e.CreateExecPodOrFail(oc.AdminKubeClient(), ns, "execpod", func(pod *v1.Pod) { pod.Spec.Containers[0].Image = "centos:7" })
+			defer func() { oc.AdminKubeClient().CoreV1().Pods(ns).Delete(execPodName, metav1.NewDeleteOptions(1)) }()
+
+			tests := map[string][]prom.MetricTest{
+				`mapi_mao_collector_up`: {prom.MetricTest{GreaterThanEqual: true, Value: 1}},
```
smarterclayton: Is this how you know a platform supports machine api?  This seems... pretty opaque.  Why don't we report that on Status of Infrastructure?

vikaschoudhary16: > Is this how you know a platform supports machine api?

No, for any platform machine api operator will be deployed. No-Op for some of platforms like BM. But in all the cases, if machine api operator has started successfully, this metric will be reported. So the test case will not fail on specific providers.

smarterclayton: Sorry, asking something different.  How does a given bit of code in our product know definitively whether machine api "works" or not on a given cluster?  I would expect some API object somewhere (ideally global config) to return a boolean or something more rich in status to say "dynamic compute possible".  Same for "dynamic load balancers possible" or "dynamic storage possible".  Otherwise, tests will start to accrue one off's like this.

I don't think we have to change it here if we don't already have something like that, but we should probably have a card for it ("As an API user of the platform I can see whether dynamic compute is enabled")

test/extended/machines/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+	})
+	g.Describe("when installed on the cluster", func() {
+		g.It("should have machine api operator metrics", func() {
+			oc.SetupProject()
+			ns := oc.Namespace()
+			execPodName := e2e.CreateExecPodOrFail(oc.AdminKubeClient(), ns, "execpod", func(pod *v1.Pod) { pod.Spec.Containers[0].Image = "centos:7" })
+			defer func() { oc.AdminKubeClient().CoreV1().Pods(ns).Delete(execPodName, metav1.NewDeleteOptions(1)) }()
+
+			tests := map[string][]prom.MetricTest{
```
LiliC: If I understand correctly this metric `mapi_mao_collector_up` according to the comment, is a metric that makes sure that the internal metrics are collected?

smarterclayton:
/approve

Rebase needed then I'll tag

test/extended/machines/metrics.go
brancz: > No, for any platform machine api operator will be deployed. No-Op for some of platforms like BM. But in all the cases, if machine api operator has started successfully, this metric will be reported. So the test case will not fail on specific providers.

That makes it sound like this should be a success/error counter. 0/1 metrics are very rarely the right thing to do, almost always should they be counters instead.

vikaschoudhary16: @brancz i dont have much background in metrics. What is problem in 0/1 metrics? Why counters would be better in this case? Can you share any reference to such counters anywhere else in openshift?

brancz: Reference: https://github.com/openshift/cluster-monitoring-operator/blob/48f061fd796d3f92053553185251cec5e24a561f/pkg/operator/operator.go#L131-L139

A counter allows for a gradual escalation for an alert, for example, a single failure may not mean anything, also this put the monitoring logic into your operator as opposed to an unopinionated data-point, whose meaning of "available" is determined at query time rather than within your go program, plus your go program would actually need to do the right thing.

Does that make sense?

test/extended/prometheus/prometheus.go
LiliC: I believe this still needs addressing, I don't see a need for this change.

PR #23726, opened 2019-09-04
Install python-openstackclient in test image

dprince:
python-openstackclient is required to run some of the the
"parallel" tests. See requirements here https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/drivers/in_tree.go#L956


images/tests/Dockerfile.rhel
```diff
@@ -14,3 +14,13 @@ RUN yum install --setopt=tsflags=nodocs -y git gzip util-linux && yum clean all
 LABEL io.k8s.display-name="OpenShift End-to-End Tests" \
       io.k8s.description="OpenShift is a platform for developing, building, and deploying containerized applications." \
       io.openshift.tags="openshift,tests,e2e"
+
+# Install python-openstackclient for Cinder test support here: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/drivers/in_tree.go
+
+COPY --from=builder /go/src/github.com/openshift/origin/images/tests/rdo-stein.repo /etc/yum.repos.d/rdo-stein.repo
```
smarterclayton: This is definitely, 100% not allowed.  You can only install RPMS from our base.

smarterclayton: /hold

PR #23728, opened 2019-09-04
Bug 1743102: UPSTREAM: 80821: Remove duplicate anti-affinity scheduler e2e

damemi:
This test has been flaky, but fortunately its removal approval was given a while back in https://github.com/kubernetes/kubernetes/pull/80821.


wking:
/bugzilla refresh

Take advantage of the bug at least temporarily targeting 4.2.0 ;).

PR #22324, opened 2019-03-15
Bug 1620608: Fix dc adoption

tnozicka:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1620608

/cc @mfojtik

PR #22683, opened 2019-04-26
Bug 1741133: Fix dc adoption

tnozicka:
forward port of https://github.com/openshift/origin/pull/22324

replaces https://github.com/openshift/origin/pull/22409

/cc @soltysh @mfojtik

PR #23731, opened 2019-09-05
Bug 1741135: Fix dc adoption

tnozicka:
forwardport of https://github.com/openshift/origin/pull/22324 - can't really do backport from 4.2 because the repo split, but this fix started in 3.11 anyways (and missed 4.1)

/cc @soltysh


tnozicka:
picked test fixes from https://github.com/openshift/origin/pull/22683

PR #23733, opened 2019-09-05
Bug 1744422: UPSTREAM: 82335: defer the close to after the error check

rphillips:
Trivial fix for a crash in e2e tests.

ref: https://github.com/kubernetes/kubernetes/pull/82335


rphillips:
requesting a review and approval

/cc @sjenning @derekwaynecarr @deads2k

PR #23740, opened 2019-09-06
Bug 1743104: DS - detect schedulable nodes on demand

tnozicka:
DS e2e detects schedulable nodes at startup which results in flakes if cluster has some nodes not schedulable yet when e2e start. This limits the occurence but doesn't fix the issue of us starting e2e on a cluster that is not ready for it.

backport of https://github.com/kubernetes/kubernetes/pull/75349 via https://github.com/kubernetes/kubernetes/pull/82415

/hold
(for https://github.com/kubernetes/kubernetes/pull/82415 to merge upstream first)

/cc @soltysh @smarterclayton @sjenning


smarterclayton:
Also approve - conformance tests have a high bar.

PR #23744, opened 2019-09-06
Bug 1716697: Verify valid push specs in logs

adambkaplan:
Ensure that the push spec printed by builds does not include the tag (only sha).

### DIFF
--- a/test/extended/builds/build_timing.go
+++ b/test/extended/builds/build_timing.go
@@ -69,6 +69,8 @@ var _ = g.Describe("[Feature:Builds][timing] capture build stages and durations"
 			g.By("starting the test source build")
 			br, _ := exutil.StartBuildAndWait(oc, "test", "--from-dir", sourceBuildBinDir)
 			br.AssertSuccess()
+			// Bug 1716697 - ensure push spec doesn't include tag, only SHA
+			o.Expect(br.Logs()).To(o.MatchRegexp(`pushed image-registry\.openshift-image-registry\.svc:5000/.*/test@sha256:`))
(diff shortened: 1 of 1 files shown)


adambkaplan:
/hold

Blocked by https://github.com/openshift/builder/pull/89

adambkaplan:
/assign @gabemontero 

fyi @nalind

test/extended/builds/build_timing.go
```diff
@@ -90,6 +92,8 @@ var _ = g.Describe("[Feature:Builds][timing] capture build stages and durations"
 			g.By("starting the test docker build")
 			br, _ := exutil.StartBuildAndWait(oc, "test", "--from-file", dockerBuildDockerfile)
 			br.AssertSuccess()
+			// Bug 1716697 - ensure push spec doesn't include tag, only SHA
+			o.Expect(br.Logs()).To(o.MatchRegexp(`pushed image-registry\.openshift-image-registry\.svc:5000/.*/test@sha256:`))
```
gabemontero: is it possible that if an error was introduced where both a tag followed by sha were in the spec, the tag would fall under the `*` in your regexp above

nalind: According to [the distribution sources](https://github.com/docker/distribution/blob/master/reference/regexp.go#L37), a valid tag value can't include `/`, and the `.*` would have to match the `:` separator for a value that included a tag to be matched, so I think it's alright.

gabemontero: thanks

PR #23747, opened 2019-09-07
origin.spec OKD RPM build readiness

LorbusChris:
- Prepared the specfile for builds of OKD packages
- Removed mentions of Tito

### DIFF
--- a/hack/build-rpms.sh
+++ b/hack/build-rpms.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
 # This script generates release zips and RPMs into _output/releases.
-# tito and other build dependencies are required on the host. We will
-# be running `hack/build-cross.sh` under the covers, so we transitively
-# consume all of the relevant envars.
+# All build dependencies are required on the host.
+# We will be running `hack/build-cross.sh` under the covers,
+# so we transitively consume all of the relevant envars.
 source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
 function cleanup() {
@@ -22,7 +22,7 @@ os::build::setup_env
 
 if [[ "${OS_ONLY_BUILD_PLATFORMS:-}" == 'linux/amd64' ]]; then
 	# when the user is asking for only Linux binaries, we will
-	# furthermore not build cross-platform clients in tito
+	# furthermore not build cross-platform clients
 	make_redistributable=0
 else
 	make_redistributable=1
@@ -58,12 +58,13 @@ if [[ -n "${dirty}" && "${OS_GIT_TREE_STATE}" == "dirty" ]]; then
 		--define "_sourcedir ${rpm_tmp_dir}/SOURCES" \
 		--define "_builddir ${rpm_tmp_dir}/BUILD" \
 		--define "skip_prep 1" \
-		--define "skip_dist 1" \
+		--define "skip_dist ${SKIP_DIST:-1}" \
 		--define "make_redistributable ${make_redistributable}" \
-		--define "version ${OS_RPM_VERSION}" --define "release ${OS_RPM_RELEASE}" \
+		--define "version ${OS_RPM_VERSION}" \
+		--define "release ${OS_RPM_RELEASE}" \
 		--define "commit ${OS_GIT_COMMIT}" \
 		--define "os_git_vars ${OS_RPM_GIT_VARS}" \
-		--define 'dist .el7' --define "_topdir ${rpm_tmp_dir}"
+		--define "_topdir ${rpm_tmp_dir}"
 	
 	mkdir -p "${OS_OUTPUT_RPMPATH}"
 	mv -f "${rpm_tmp_dir}"/RPMS/*/*.rpm "${OS_OUTPUT_RPMPATH}"
@@ -72,16 +73,18 @@ else
 	mkdir -p "${rpm_tmp_dir}/SOURCES"
 	tar czf "${rpm_tmp_dir}/SOURCES/${OS_RPM_NAME}-${OS_RPM_VERSION}.tar.gz" \
 		--owner=0 --group=0 \
-		--exclude=_output --exclude=.git --transform "s|^|${OS_RPM_NAME}-${OS_RPM_VERSION}/|rSH" \
+		--exclude=_output --exclude=.git \
+		--transform "s|^|${OS_RPM_NAME}-${OS_RPM_VERSION}/|rSH" \
 		.
 
 	rpmbuild -b${srpm} "${OS_RPM_SPECFILE}" \
-		--define "skip_dist 1" \
+		--define "skip_dist ${SKIP_DIST:-1}" \
 		--define "make_redistributable ${make_redistributable}" \
-		--define "version ${OS_RPM_VERSION}" --define "release ${OS_RPM_RELEASE}" \
+		--define "version ${OS_RPM_VERSION}" \
+		--define "release ${OS_RPM_RELEASE}" \
 		--define "commit ${OS_GIT_COMMIT}" \
 		--define "os_git_vars ${OS_RPM_GIT_VARS}" \
-		--define 'dist .el7' --define "_topdir ${rpm_tmp_dir}"
+		--define "_topdir ${rpm_tmp_dir}"
 
 	output_directory="$( find "${rpm_tmp_dir}" -type d -path "*/BUILD/${OS_RPM_NAME}-${OS_RPM_VERSION}/_output/local" )"
 	if [[ -z "${output_directory}" ]]; then
@@ -97,7 +100,7 @@ else
 	# an NFS volume exported with root_squash set.  This can occur when running this
 	# script on a Vagrant box.  The error shown is "mv: failed to preserve ownership
 	# for $FILE: Operation not permitted".  As a workaround, if
-	# ${tito_output_directory} and ${OS_OUTPUT} are on different devices, use cp and
+	# ${output_directory} and ${OS_OUTPUT} are on different devices, use cp and
 	# rm instead.
 	if [[ $(stat -c %d "${output_directory}") == $(stat -c %d "${OS_OUTPUT}") ]]; then
 		mv "${output_directory}"/* "${OS_OUTPUT}"
--- a/origin.spec
+++ b/origin.spec
@@ -7,8 +7,8 @@
 %global import_path github.com/openshift/origin
 
 %global golang_version 1.12
-# %commit and %os_git_vars are intended to be set by tito custom builders provided
-# in the .tito/lib directory. The values in this spec file will not be kept up to date.
+# commit and os_git_vars are intended to be set by the build system.
+# NOTE: The values in this spec file will not be kept up to date.
 %{!?commit:
 %global commit 86b5e46426ba828f49195af21c56f7c6674b48f7
 }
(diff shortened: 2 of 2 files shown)


origin.spec
```diff
@@ -46,16 +46,16 @@
...
 %{!?make_redistributable: %global make_redistributable %{need_redistributable_set}}
 
-%if "%{dist}" == ".el7aos"
+%if 0%{?epel}
 %global package_name openshift
 %global product_name OpenShift
 %else
-%global package_name openshift
-%global product_name OpenShift
+%global package_name okd
```
DanyC97: aren't we breaking OCP here?

LorbusChris: hm, on epel it'll still be openshift. See the lines above

origin.spec
```diff
@@ -7,15 +7,15 @@
...
+# commit and os_git_vars are intended to be set by the build system.
 %{!?commit:
-%global commit 86b5e46426ba828f49195af21c56f7c6674b48f7
+%global commit e795244084
 }
 %global shortcommit %(c=%{commit}; echo ${c:0:7})
 # os_git_vars needed to run hack scripts during rpm builds
 %{!?os_git_vars:
-%global os_git_vars OS_GIT_VERSION='' OS_GIT_COMMIT='' OS_GIT_MAJOR='' OS_GIT_MINOR='' OS_GIT_TREE_STATE=''
+%global os_git_vars OS_GIT_COMMIT='e795244084' OS_GIT_TREE_STATE='dirty' OS_GIT_VERSION='v4.1.0+e795244-1411-dirty' OS_GIT_MAJOR='4' OS_GIT_MINOR='1+' OS_GIT_PATCH='0' KUBE_GIT_MAJOR='1' KUBE_GIT_MINOR='14+' KUBE_GIT_COMMIT='e8fcd6a' KUBE_GIT_VERSION='v1.14.0+e795244084' ETCD_GIT_VERSION='v3.3.10' ETCD_GIT_COMMIT='27fc7e2'
```
DanyC97: any reason why we added the version in too ? 
i may miss something here but if i remember in the past the hack/* script was dynamically working out the version ... not sure if this breaks OCP

hack/build-rpms.sh
```diff
@@ -22,7 +21,7 @@ os::build::setup_env
 
 if [[ "${OS_ONLY_BUILD_PLATFORMS:-}" == 'linux/amd64' ]]; then
 	# when the user is asking for only Linux binaries, we will
-	# furthermore not build cross-platform clients in tito
```
DanyC97: am i right in thinking that currently tito is no longer in used (e.g it was in 3.x but world moved on since then? )

origin.spec
LorbusChris: Afaict this is just what OCP passes as os_git_vars right now and line just hasn't been kept up to date before. Putting them all in one var does obscure things here a bit.

hack/build-rpms.sh
LorbusChris: I will need some clarification on where it is still used today, if at all.

origin.spec
DanyC97: _line just hasn't been kept up to date before_ => that is what i was afraid you will say after i posted my comment, thanks for the info !

LorbusChris:
Verification is slightly tricky, because one revision of the specfile can only contain a git ref to the previous commit, before the spec itself was commited.
Changed the script to verify the spec against HEAD~1 (to be done after the commit) and removed the it from the `make verify` target for now. This isn't ideal.

To me, the best flow would look something like this in order to make sure the spec is updated with a git ref that contains all other updates:
1 make update (as is now, !spec),
2 git commit,
3 make update spec,
4 git commit
5 make verify all

I think Packit doesnt require the right commit and version to be set (at least on PRs). Will investigate.

In the meantime, does anybody have an idea how to solve this elegantly?

origin.spec
```diff
@@ -102,9 +103,14 @@ Summary:        %{product_name} Kubernetes server commands
...
-Provides:       hyperkube
-Provides:       atomic-openshift-hyperkube
-Provides:       atomic-openshift-node
+Provides:       hyperkube = %{version}
+Provides:       atomic-openshift-hyperkube = %{version}
+Provides:       atomic-openshift-node = %{version}
+%if 0%{?epel}
+Conflicts:      okd-hyperkube
+%else
+Conflicts:      openshift-hyperkube
```
smarterclayton: Why add this?

origin.spec
```diff
@@ -142,16 +148,10 @@ OS_ONLY_BUILD_PLATFORMS="${BUILD_PLATFORM}" %{os_git_vars} OS_BUILD_RELEASE_ARCH
...
 PLATFORM="$(go env GOHOSTOS)/$(go env GOHOSTARCH)"
 install -d %{buildroot}%{_bindir}
-
-# Install linux components
-for bin in hyperkube
-do
-  echo "+++ INSTALLING ${bin}"
-  install -p -m 755 _output/local/bin/${PLATFORM}/${bin} %{buildroot}%{_bindir}/${bin}
-done
+echo "+++ INSTALLING hyperkube"
```
smarterclayton: Leave the loop, we may have additional binaries in this package.

origin.spec
```diff
@@ -46,16 +46,16 @@
...
-%global package_name openshift
-%global product_name OpenShift
+%global package_name okd
+%global product_name OKD
 %endif
 
-%{!?version: %global version 0.0.1}
-%{!?release: %global release 1}
+%{!?version: %global version 4.1.0}
+%{!?release: %global release 1.1417.98b35a9}
```
smarterclayton: Are you generating this?  It would change on ever commit if so which won’t work.

origin.spec
```diff
@@ -46,16 +46,16 @@
...
 %else
-%global package_name openshift
-%global product_name OpenShift
+%global package_name okd
+%global product_name OKD
 %endif
 
-%{!?version: %global version 0.0.1}
-%{!?release: %global release 1}
+%{!?version: %global version 4.1.0}
```
smarterclayton: I don’t think we should be tying this version to a top level version, the current plan is that component version numbers are completely component dependent and likely not to be managed under the covers (the rpm may vary wildly over time).  Probably better to start at 4.0.0

hack/build-rpms.sh
smarterclayton: I’m not aware of any use in this repo, and it hasn’t been updated in a year or more.

smarterclayton:
Looks like this broke the ostree override for PR testing

origin.spec
LorbusChris: Intention was to signal that the user can't install both at the same time due to conflicting files.

However, on second thought, as the openshift-hypershift rpm is not distributed stand-alone, I'd be fine with removing it again altogether because they might not ever ever come across each other.
Should I remove?

origin.spec
LorbusChris: Yeah..I realized after the fact that committing the vars can never work with CI.
The right way is probably to have a proper make target for OKD builds and pass the vars at build time in a proper CI pipeline.

origin.spec
LorbusChris: This did actually break things in OCP. Should be fixed by the `.openshift` dist tag now

origin.spec
LorbusChris: Removed the spec update generation.

origin.spec
LorbusChris: This was also taken from the auto-generation. Set to 4.0.0 now.

origin.spec
DanyC97: @LorbusChris @smarterclayton beat me on this ;) after our chat over the weekend i started to dig a bit and found https://github.com/openshift/release/pull/4466 & https://github.com/openshift/installer/pull/2131 hence the hint i guess.

LorbusChris: Above links have something to do with removing `hypershift`.  This is about `hyperkube`, though. I don't have a strong opinion about this one and can remove it, but my interpretation of https://docs.fedoraproject.org/en-US/packaging-guidelines/Conflicts/#_implicit_conflicts would still be to add it.

LorbusChris:
I just noticed that we shouldn't inject `Provides: bundled(x)`, but rather `BuildRequires: bundled(x)` as per https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/#_buildrequires, and there is a macro for that:
```
%generate_buildrequires
%go_generate_buildrequires
```
Will try to get that to work next.

origin.spec
DanyC97: sorry, ignore what i've said, monday morning w/o coffee.

sgreene570: > This is about `hyperkube`, though. I don't have a strong opinion about this one and can remove it, but my interpretation of https://docs.fedoraproject.org/en-US/packaging-guidelines/Conflicts/#_implicit_conflicts would still be to add it.

+ 1

Is there any disadvantage to being verbose and including the conflict in this situation?

hack/build-rpms.sh
```diff
@@ -58,12 +58,14 @@ if [[ -n "${dirty}" && "${OS_GIT_TREE_STATE}" == "dirty" ]]; then
 		--define "_sourcedir ${rpm_tmp_dir}/SOURCES" \
 		--define "_builddir ${rpm_tmp_dir}/BUILD" \
 		--define "skip_prep 1" \
-		--define "skip_dist 1" \
+		--define "skip_dist ${SKIP_DIST:-1}" \
+		--define "ose_build ${OSE_BUILD:-1}" \
```
LorbusChris: ose or ocp build?

hack/build-rpms.sh
```diff
@@ -72,16 +74,19 @@ else
...
 		--define "make_redistributable ${make_redistributable}" \
-		--define "version ${OS_RPM_VERSION}" --define "release ${OS_RPM_RELEASE}" \
+		--define "version ${OS_RPM_VERSION}" \
+		--define "release ${OS_RPM_RELEASE}" \
 		--define "commit ${OS_GIT_COMMIT}" \
 		--define "os_git_vars ${OS_RPM_GIT_VARS}" \
-		--define 'dist .el7' --define "_topdir ${rpm_tmp_dir}"
```
LorbusChris: can I safely remove the `dist` override here? Looks like it it is skipped anyway by `skip_dist`

origin.spec
smarterclayton: In this case we shouldn't have `atomic-openshift` anymore, OKD can use `openshift` for package names.

origin.spec
smarterclayton: Let's just use package_name openshift for everything.  Do we *need* to call it OKD?

origin.spec
smarterclayton: `openshift-hyperkube`

hack/build-rpms.sh
smarterclayton: `ocp_build` if you have too.

origin.spec
```diff
@@ -46,15 +46,15 @@
...
 
-%if "%{dist}" == ".el7aos"
+%if 0%{?ose_build}
 %global package_name openshift
 %global product_name OpenShift
 %else
-%global package_name openshift
-%global product_name OpenShift
+%global package_name okd
+%global product_name OKD
```
smarterclayton: What are the rules on Product Name in general?

origin.spec
smarterclayton: Oh, I see.  Yeah, let's call the variable `ocp_build`, keep the package name the same (no reason not to), and change product name based on ocp_build

origin.spec
LorbusChris: not sure. If we could reuse the `openshift` naming scheme across the board, we wouldn't need the ose/ocp distinction in rpmbuild

hack/build-rpms.sh
LorbusChris: won't be necessary if we can use `openshift` name schema

origin.spec
smarterclayton: Yeah, let's reuse the openshift naming scheme across the board for now.  We call component pieces OpenShift and they are allowed to use the word.  It's only the top level that is officially called OKD for Reasons(tm) but we allow the use of the term to refer to the code base.

smarterclayton:
/lgtm

If this has an impact on ART builds we'll revert

PR #23751, opened 2019-09-09
Bug 1749645 - watch cache causes update failure on stale data

deads2k:
looks like https://github.com/kubernetes/kubernetes/issues/82130 which is fixed by https://github.com/kubernetes/kubernetes/pull/82303

/assign @JacobTanenbaum


JacobTanenbaum:
fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1749645

PR #23758, opened 2019-09-10
Bug 1731263: Reenable basic pod preemption e2es

damemi:
This test was marked as flaky but has had a number of fixes applied to it (pr #23645 and #23719) tht have adjusted the node utilization calculations and structure of the test to hopefully make it more stable. In order to address the comments in the BZ for this test's flakiness (https://bugzilla.redhat.com/show_bug.cgi?id=1731263#c9) we should remove the [flaky] tag to re-enable the tests.


damemi:
@deads2k you mentioned that the `[Flaky]` tag is how to disable these tests, could you confirm on this PR that it will reenable it?

damemi:
/retest
this still flaking?

damemi:
/hold
I think these tests need to just be rewritten, there are some spots that don't make sense to me

soltysh:
> I think these tests need to just be rewritten, there are some spots that don't make sense to me

@damemi how about checking what the newer version will look like? I mean that from k8s 1.16 introduced in the rebase and look at it only after rebase lands. I'm moving the BZ to 4.3

damemi:
@soltysh This test is actually ahead of 1.16 right now (origin already has the changes in https://github.com/kubernetes/kubernetes/pull/82350 which hasn't merged)

soltysh:
@damemi what's the call here?

damemi:
@soltysh I need to revisit this test entirely, it's not flaking at all upstream but when I submitted this PR to remove `[flaky]` it still fails CI

soltysh:
@damemi what's the current state of this?

damemi:
It seems that this test was removed from `[flaky]` status with the k8s bump in https://github.com/openshift/origin/pull/23811, so it should be re-enabled now I think...

Though, I can't find the test in [4.2 testgrid](https://testgrid.k8s.io/redhat-openshift-ocp-release-4.2-blocking#release-openshift-origin-installer-e2e-aws-serial-4.2) or 4.3

but I do see preemption tests in, for example, 4.1 blocking: https://testgrid.k8s.io/redhat-openshift-ocp-release-4.1-blocking#release-openshift-origin-installer-e2e-aws-serial-4.1. Maybe I'm just not finding it, so I'll switch the BZ back to modified and have QA take a look as well

damemi:
So I noticed that, with our changes to this test, the pod wasn't always being scheduled onto the node we expected (for example, if there is a low priority pod, we were assuming that the preemptor pod would be placed on the same node as the low priority pod, which wasn't always the case, and so the test failed with no pods preempted and the preemptor pod running).

Now I've made an update that ensures that the preemptor pod will land on the appropriate node. However, it seems that preemption just isn't running, because *now* the preemptor pod is attempting to schedule on the correct node, but failing due to lack of cpu:
```
event for preemptor-pod: {kubelet ip-10-0-141-255.ec2.internal} OutOfcpu: Node didn't have enough resource: cpu, requested: 1686, used: 2176, capacity: 3500
...
Nov  8 18:37:20.016: INFO: pod0-sched-preemption-low-priority     ip-10-0-141-255.ec2.internal  Running         [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:09 +0000 UTC  } {Ready True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:18 +0000 UTC  } {ContainersReady True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:18 +0000 UTC  } {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:09 +0000 UTC  }]
Nov  8 18:37:20.016: INFO: pod1-sched-preemption-medium-priority  ip-10-0-142-166.ec2.internal  Running         [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:09 +0000 UTC  } {Ready True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:18 +0000 UTC  } {ContainersReady True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:18 +0000 UTC  } {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:09 +0000 UTC  }]
Nov  8 18:37:20.016: INFO: pod2-sched-preemption-medium-priority  ip-10-0-154-84.ec2.internal   Running         [{Initialized True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:09 +0000 UTC  } {Ready True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:18 +0000 UTC  } {ContainersReady True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:18 +0000 UTC  } {PodScheduled True 0001-01-01 00:00:00 +0000 UTC 2019-11-08 18:37:09 +0000 UTC  }]
Nov  8 18:37:20.016: INFO: preemptor-pod                          ip-10-0-141-255.ec2.internal  Failed          []
```

@ravisantoshgudimetla, do we need to enable podpreemption in our e2es somehow? Failing to schedule due to lack of cpu is exactly what we want to test

soltysh:
Irrelevant.
/close

PR #23720, opened 2019-09-03
Bug 1745720: Disable GCE CSI tests due to CRI-O bug

huffmanca:
Disabled the GCE CSI tests temporarily, as they are currently failing due to a CRI-O bug that prevents accessing IPs [1], currently being targeted for 4.3.

This resolves [BZ 1745720](https://bugzilla.redhat.com/show_bug.cgi?id=1745720).

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1718389

PR #23760, opened 2019-09-10
Bug 1750926: Disable failing storage GCE Serial tests due to openshift blocking pod access to cloud provider.

huffmanca:
This is a followup to https://github.com/openshift/origin/pull/23720, that disables the remaining failing tests. These tests appear to be failing due to a CRI-O bug that prevents accessing IPs [1], currently being targeted for 4.3.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1718389

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -491,6 +491,9 @@ var (
 
 			// https://bugzilla.redhat.com/show_bug.cgi?id=1745720
 			`\[sig-storage\] CSI Volumes \[Driver: pd.csi.storage.gke.io\]\[Serial\]`,
+
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1749882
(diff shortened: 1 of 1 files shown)


test/extended/util/test.go
```diff
@@ -491,6 +491,9 @@ var (
 
 			// https://bugzilla.redhat.com/show_bug.cgi?id=1745720
 			`\[sig-storage\] CSI Volumes \[Driver: pd.csi.storage.gke.io\]\[Serial\]`,
+
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1749882
+			`\[sig-storage\] CSI Volumes CSI Topology test using GCE PD driver \[Serial\]`,
```
bertinatto: We should probably remove the `[Serial] ` from here (and from the test above as well).

bertinatto:
/assign @mfojtik @eparis 

for approval

PR #23762, opened 2019-09-10
increase the scale up timeout to 10 minutes

iamemilio:
The OpenStack platform is preparing for GA and is currently timing out on the scale up phase of the scale machineSet test in their serial test plan. We would like to bump the time out up to 10 minutes, which should provide a reasonable margin of error, so that our platform can consistently pass its tests.


iamemilio:
Sample flake: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/canary-openshift-ocp-installer-e2e-openstack-serial-4.2/7

mandre:
I ran the full `openshift/conformance/serial` suite against my cluster deployed on OpenStack platform and can confirm that the `grow and decrease cluster failed when scaling different machineSets simultaneously` test passes when increasing the timeout.

test/extended/machines/scale.go
```diff
@@ -25,7 +25,7 @@ import (
 const (
 	machineAPIGroup       = "machine.openshift.io"
 	machineSetOwningLabel = "machine.openshift.io/cluster-api-machineset"
-	scalingTime           = 7 * time.Minute
+	scalingTime           = 12 * time.Minute
```
mandre: Thanks for raising the timeout limit even further, I was about to say 10 min doesn't give us a lot of room.
With my tests locally, the first time it took 9:30, and the second time 10:45.

Could you just update the commit message?

mandre: Commit title, sorry...

mandre:
This patch should help making the `openshift/conformance/serial` test suite green on OpenStack platform.

https://testgrid.k8s.io/redhat-openshift-release-informing#redhat-canary-openshift-ocp-installer-e2e-openstack-serial-4.2

sdodson:
/cc @vikaschoudhary16 @enxebre 
(authors of this test)

PR #23765, opened 2019-09-11
Update Jenkins Secret Sync Test for new feature to change name

waveywaves:
Checks todo :

- [x] Check change in the credential name after annotating secret 
- [ ] Check credential name for pre annotated secret
- [ ] User submits a customCredentialName which is the same of the generatedCredentialName
- [ ] User submits a customCredentialName which is the same as an existing credential Name
- [ ] User submits 2 subsquents credentialsNames.

cc @akram @gabemontero


gabemontero:
so far looks good @waveywaves 

will await the additional changes per your description checklist

PR #23767, opened 2019-09-11
Bug 1750809: [4.1] Fix bootstrap IdP challenges ruining RequestHeaders IdP

stlaz:
If RequestHeaders are configured for challenges, don't set up
the bootstrap IdP for challenges either. This would otherwise
add the WWW-Authenticate header on top of the redirecting request
which then in turn causes oauth-server internal errors.


stlaz:
cc @enj

PR #23768, opened 2019-09-11
Disable local-storage tests that use gce local-ssd

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1751367

/sig storage

cc @huffmanca @knobunc


gnufied:
cc @deads2k

PR #23771, opened 2019-09-12
add selinux go build tag on linux

rphillips:
With the 1.16 rebase we found the selinux build tag was found to be missing. We need this patch to enable proper labeling within the go-selinux package for 1.14 origin releases and prior.

ref: https://github.com/openshift/origin/blob/master/vendor/github.com/opencontainers/selinux/go-selinux/label/label_selinux.go#L1

/cc @sjenning @mrunalp @smarterclayton @derekwaynecarr @eparis @deads2k


smarterclayton:
Is there an selinux e2e test that would capture this?

mrunalp:
@haircommander is looking into e2e test for this.

smarterclayton:
Updates?  This is concerning for lots of reasons, why aren't be positive that this is doing the right thing when enabled?

rphillips:
@haircommander is actively working on a test

haircommander:
After a couple of days testing (with the help of @mrunalp and @rphillips) I have determined SELinux is working as expected (properly setting labels on processes and volumes) in 4.2. I don't know if we need this PR

rphillips:
Sounds like we can close it. Thanks for verifying that @haircommander .

PR #23783, opened 2019-09-13
Bug 1752045: UPSTREAM: <carry>:Add a RBAC checker for external IP ranger

abhat:
This PR adds a RBAC checker for adding an External IP range for a service.

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
@@ -7,15 +7,15 @@ import (
 	"net"
 	"strings"
 
-	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"github.com/openshift/library-go/pkg/config/helpers"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/admission/initializer"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
 	"k8s.io/klog"
-	kapi "k8s.io/kubernetes/pkg/apis/core"
-
-	"github.com/openshift/library-go/pkg/config/helpers"
 	"k8s.io/kubernetes/openshift-kube-apiserver/admission/network/apis/externalipranger"
 	v1 "k8s.io/kubernetes/openshift-kube-apiserver/admission/network/apis/externalipranger/v1"
+	kapi "k8s.io/kubernetes/pkg/apis/core"
 )
 
 const ExternalIPPluginName = "network.openshift.io/ExternalIPRanger"
@@ -63,11 +63,13 @@ type externalIPRanger struct {
 	*admission.Handler
 	reject         []*net.IPNet
 	admit          []*net.IPNet
+	authorizer     authorizer.Authorizer
 	allowIngressIP bool
 }
 
 var _ admission.Interface = &externalIPRanger{}
 var _ admission.ValidationInterface = &externalIPRanger{}
+var _ = initializer.WantsAuthorizer(&externalIPRanger{})
 
 // ParseRejectAdmitCIDRRules calculates a blacklist and whitelist from a list of string CIDR rules (treating
 // a leading ! as a negation). Returns an error if any rule is invalid.
@@ -99,6 +101,17 @@ func NewExternalIPRanger(reject, admit []*net.IPNet, allowIngressIP bool) *exter
 		admit:          admit,
 		allowIngressIP: allowIngressIP,
 	}
+}
+
+func (r *externalIPRanger) SetAuthorizer(a authorizer.Authorizer) {
+	r.authorizer = a
+}
+
+func (r *externalIPRanger) ValidateInitialization() error {
+	if r.authorizer == nil {
+		return fmt.Errorf("missing authorizer")
+	}
+	return nil
 }
 
 // NetworkSlice is a helper for checking whether an IP is contained in a range
@@ -165,8 +178,31 @@ func (r *externalIPRanger) Validate(ctx context.Context, a admission.Attributes,
 			}
 		}
 	}
+
 	if len(errs) > 0 {
-		return apierrs.NewInvalid(a.GetKind().GroupKind(), a.GetName(), errs)
+		//if there are errors reported, resort to RBAC check to see
+		//if this is an admin user who can over-ride the check
+		allow, err := r.checkAccess(a)
+		if err != nil {
+			return err
+		}
+		if !allow {
+			return admission.NewForbidden(a, errs.ToAggregate())
+		}
 	}
+
 	return nil
+}
+
+func (r *externalIPRanger) checkAccess(attr admission.Attributes) (bool, error) {
+	authzAttr := authorizer.AttributesRecord{
+		User:            attr.GetUserInfo(),
+		Verb:            "create",
+		Resource:        "service",
+		Subresource:     "externalips",
+		APIGroup:        "network.openshift.io",
+		ResourceRequest: true,
+	}
+	authorized, _, err := r.authorizer.Authorize(authzAttr)
+	return authorized == authorizer.DecisionAllow, err
 }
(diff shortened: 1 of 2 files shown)


vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
```diff
@@ -167,5 +183,29 @@ func (r *externalIPRanger) Validate(a admission.Attributes, _ admission.ObjectIn
...
+
 	return nil
 }
+
+func (r *externalIPRanger) checkAccess(attr admission.Attributes) (bool, error) {
+	authzAttr := authorizer.AttributesRecord{
+		User:            attr.GetUserInfo(),
+		Verb:            "create",
+		Namespace:       attr.GetNamespace(),
+		Resource:        "endpoints",
```
danwinship: "service"

soltysh:
/cc @deads2k @sttts 
since this is your area and on top of this is adding another carry patch

squeed:
/retitle Bug 1752045: UPSTREAM: <carry>:Add a RBAC checker for external IP ranger

squeed:
FWIW, this used to live in pkg/network, but since it's part of the kube-apiserver, vendor/... is the authoritative source.

abhat:
Update: fixed a typo in the testname in one of the admission tests in the recent force-push.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
```diff
@@ -155,6 +171,7 @@ func (r *externalIPRanger) Validate(a admission.Attributes, _ admission.ObjectIn
 			ip := net.ParseIP(s)
 			if ip == nil {
 				errs = append(errs, field.Forbidden(field.NewPath("spec", "externalIPs").Index(i), "externalIPs must be a valid address"))
+				invalidIp = true
```
danwinship: Kubernetes [already validates](https://github.com/openshift/origin/blob/17f1107c/vendor/k8s.io/kubernetes/pkg/apis/core/validation/validation.go#L3712) that all `service.Spec.ExternalIPs` values are valid IPs, so we don't need to check that here. So we don't need to distinguish admin-overridable from non-admin-overridable errors.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
```diff
@@ -164,8 +181,37 @@ func (r *externalIPRanger) Validate(a admission.Attributes, _ admission.ObjectIn
...
-		return apierrs.NewInvalid(a.GetKind().GroupKind(), a.GetName(), errs)
+		//if there are errors reported, resort to RBAC check to see
+		//if this is an admin user who can over-ride the check
+		if !invalidIp {
+			allow, err := r.checkAccess(a)
+			if err != nil {
+				return err
+			}
+			if !allow {
+				return apierrs.NewInvalid(a.GetKind().GroupKind(), a.GetName(), errs)
```
danwinship: as in the restricted endpoints controller, this should be `apierrs.NewForbidden`

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission_test.go
```diff
@@ -1,15 +1,45 @@
...
+}
+
+func (a *fakeTestAuthorizer) Authorize(attributes authorizer.Attributes) (authorizer.Decision, string, error) {
+	ui := attributes.GetUser()
+	if ui == nil {
+		return authorizer.DecisionNoOpinion, "", fmt.Errorf("No valid UserInfo for Context")
+	}
+	// system:serviceaccount:herpy:derpy user aka admin user is allowed to set
+	// external IPs
+	if ui.GetName() == "system:serviceaccount:herpy:derpy" {
```
danwinship: use a more self-documenting name. No one is going to understand why a test is expected to pass for "herpy:derpy" but fail for "slurpy:derpy".

eg `"system:serviceaccount:test:admin"` vs `"system:serviceaccount:test:ordinary-user"`

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
abhat: I only moved the existing error inside the `!allow` block. But for correctness, I can change it to `apierrs.NewForbidden`

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission_test.go
abhat: I was following the example in the `podnodeconstraints` admission plugin, but I suppose I can change the account name for clarity. The fake authorizer comments do make it clear that `herpy:derpy` is an admin user, but i guess i can make it unambiguous further using appropriate account names. Thanks for keeping me honest.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
abhat: Isn't it better to validate them early in the cycle before it gets to the spec level validation?

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
abhat: Actually, the only apierror that takes an error list as argument is `NewInvalid`. Do we want to use `NewForbidden` and only pick the first element in the `errs` slice?

deads2k:
/hold

I'd like to see an enhancement for this that outlines the problem, links prior art, and suggests different options for the RBAC permissions.

deads2k:
1 Is this really a namespace scoped check?
2. Does subdivision of this power by namespace make sense?
3. Should regular users with this power in one namespace be able delegate that power to other users?
4. Is this an escalating permission for the sake of authorization scope grants?

danwinship:
We don't really expect anyone to actually be granted permission to do this. The RBAC check is just an escape hatch for cluster admins to be able to bypass the checks if absolutely needed. The design was based on what the restricted endpoints admission controller does, but that's a few years old and maybe that's no longer considered the right thing?

That said:

> 1 Is this really a namespace scoped check?
> 2. Does subdivision of this power by namespace make sense?
> 3. Should regular users with this power in one namespace be able delegate that power to other users?

No to all three: being able to bypass external IP restrictions in any namespace gives you the power to attack people in every namespace.

> 4. Is this an escalating permission for the sake of authorization scope grants?

I don't know what you mean by that

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
danwinship: validation should happen before admission controllers; there's no point in doing admission control on an object that will be rejected anyway

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
danwinship: It seems like it should have been Forbidden before... anyway, you could aggregate the errors into a single message perhaps

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/network/externalipranger/externalip_admission.go
```diff
@@ -165,8 +178,31 @@ func (r *externalIPRanger) Validate(ctx context.Context, a admission.Attributes,
...
+		if !allow {
+			return admission.NewForbidden(a, errs.ToAggregate())
+		}
 	}
+
 	return nil
 }
+
+func (r *externalIPRanger) checkAccess(attr admission.Attributes) (bool, error) {
+	authzAttr := authorizer.AttributesRecord{
```
deads2k: this matches our design.

PR #23784, opened 2019-09-13
e2e: test oc adm release mirror

jcpowermac:
Adds a test that creates an imagestream and executes
oc adm release mirror.  Then confirms
that the imagestreamtag release has
an identical clusterversion image url.


gabemontero:
@dmage @adambkaplan FYI

jcpowermac:
@smarterclayton when you have a moment PTAL.

PR #23796, opened 2019-09-16
UPSTREAM: 82367: Enable block tests for Cinder

Fedosin:
Cinder tests are failing in a downstream test suite. With disabled block support, we test that a block volume cannot be dynamically provisioned and it fails, because Cinder provisions block volumes just fine.

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/storage/drivers/in_tree.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/storage/drivers/in_tree.go
@@ -962,6 +962,7 @@ func InitCinderDriver() testsuites.TestDriver {
 				testsuites.CapPersistence: true,
 				testsuites.CapFsGroup:     true,
 				testsuites.CapExec:        true,
+				testsuites.CapBlock:       true,
 			},
 		},
 	}


mandre:
We're hoping this helps with some of the cinder failures we're seeing in the parallel test suite on OpenStack platform [1], specifically the `[sig-storage] In-tree Volumes [Driver: cinder] [Testpattern: Dynamic PV (block volmode)] volumeMode should fail in binding dynamic provisioned PV to PVC [Suite:openshift/conformance/parallel] [Suite:k8s]` and the `[sig-storage] In-tree Volumes [Driver: cinder] [Testpattern: Pre-provisioned PV (block volmode)] volumeMode should fail to create pod by failing to mount volume [Suite:openshift/conformance/parallel] [Suite:k8s]` tests.

[1] https://testgrid.k8s.io/redhat-openshift-release-informing#redhat-canary-openshift-ocp-installer-e2e-openstack-4.2

Fedosin:
Failing unit tests are not related to the patch, they are fixed in upstream should be backported to OpenShift soon
https://github.com/kubernetes/kubernetes/pull/82697

Fedosin:
Broken unit test fix backport: https://github.com/openshift/origin/pull/23800

smarterclayton:
Tagged the other one, retest when that merges and ping me again

Fedosin:
@smarterclayton ping :) tests are okay

PR #23797, opened 2019-09-16
Bug 1753324: Allow skipping tests based on detected network plugin

danwinship:
This fixes two problems:
- We need to be able to skip some tests on ovn-kubernetes (eg, the Service SessionAffinity tests, since that feature isn't implemented yet in 4.2).
- We need to be able to *not* skip some tests on ovn-kubernetes that we are skipping on openshift-sdn (eg, NetworkPolicy egress tests)

This also lets us simplify the way the multitenant suite is done.

(WIP until the 1.16 rebase lands since this will need to be updated for that)

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -131,12 +131,21 @@ func ExecuteTest(t ginkgo.GinkgoTestingT, suite string) {
 }
 
 func AnnotateTestSuite() {
-	testRenamer := newGinkgoTestRenamerFromGlobals(e2e.TestContext.Provider)
+	testRenamer := newGinkgoTestRenamerFromGlobals(e2e.TestContext.Provider, getNetworkSkips())
 
 	ginkgo.WalkTests(testRenamer.maybeRenameTest)
 }
 
-func newGinkgoTestRenamerFromGlobals(provider string) *ginkgoTestRenamer {
+func getNetworkSkips() []string {
+	out, err := e2e.KubectlCmd("get", "network.operator.openshift.io", "cluster", "--template", "{{.spec.defaultNetwork.type}}{{if .spec.defaultNetwork.openshiftSDNConfig}} {{.spec.defaultNetwork.type}}/{{.spec.defaultNetwork.openshiftSDNConfig.mode}}{{end}}").CombinedOutput()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


danwinship:
/assign @smarterclayton 
I don't want this to merge until after the rebase but can you comment on the approach?

If you don't like running `kubectl` at startup, I guess we could just set environment variables from the multitenant and ovn-kubernetes test jobs. Or update the test-running infrastructure to allow jobs to pass a `--skip` arg or something. But I liked this approach because it lets us fix the "some tests should be skipped under openshift-sdn but not ovn-kubernetes" problem without needing to treat openshift-sdn specially and without having to change all the existing e2e-aws definitions to include something saying "skip the tests that only be skipped under openshift-sdn".

danwinship:
OK, QE is annoyed that there's a multitenant release job and it's not passing. So let's do this, so we'll have the multitenant and ovn-kubernetes tests available in 4.2, and it will just get updated a little bit with the rebase.

squeed:
flake
/retest

knobunc:
@soltysh @mfojtik -- Can either of you please approve the change to /cmd?

test/extended/util/test.go
```diff
@@ -131,12 +131,21 @@ func ExecuteTest(t ginkgo.GinkgoTestingT, suite string) {
...
 
 func AnnotateTestSuite() {
-	testRenamer := newGinkgoTestRenamerFromGlobals(e2e.TestContext.Provider)
+	testRenamer := newGinkgoTestRenamerFromGlobals(e2e.TestContext.Provider, getNetworkSkips())
 
 	ginkgo.WalkTests(testRenamer.maybeRenameTest)
 }
 
-func newGinkgoTestRenamerFromGlobals(provider string) *ginkgoTestRenamer {
+func getNetworkSkips() []string {
```
smarterclayton: This really doesn't belong here, since it's making network calls deep in the guts of the initialization path.  If you want to provide plugin specific skips you need to configure that in the top level of openshift-tests (in the same place the provider is initialized) and then follow the same pattern as the cloud skips.

If you're too busy to rework this it'll be in my next pass, but I'd like this fixed soon.

PR #23798, opened 2019-09-16
Bug #1752646: Disable the router header test for all platforms

iamemilio:
It seems that most platforms are either having trouble with, or have already disabled this test. This test occasionally passes but usually doesn't in our canaries, and we believe the best course of action is to disable it.

Test Grid:
https://testgrid.k8s.io/redhat-openshift-release-informing#redhat-canary-openshift-ocp-installer-e2e-openstack-4.2

Example Failure:
https://prow.k8s.io/view/gcs/origin-ci-test/logs/canary-openshift-ocp-installer-e2e-openstack-4.2/194

### DIFF
--- a/test/extended/router/headers.go
+++ b/test/extended/router/headers.go
@@ -33,17 +33,6 @@ var _ = g.Describe("[Conformance][Area:Networking][Feature:Router]", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 		metricsIP, err = exutil.WaitForRouterInternalIP(oc)
 		o.Expect(err).NotTo(o.HaveOccurred())
-
-		if routerIP != metricsIP {
-			// On 4.x clusters, there's couple of bugs to fix before
-			// we can enable this test:
-			//  1. Selector on router-internal-default service
-			//  2. There is a weird 10.131.0.1 IP in the mix which
-			//     shows up as the client ip and not the node IP.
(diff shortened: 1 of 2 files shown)


ironcladlou:
@Miciah PTAL (although I thought this was [already skipped](https://github.com/openshift/origin/blob/master/test/extended/router/headers.go#L44), I'm probably misunderstanding something)

Miciah:
> @Miciah PTAL (although I thought this was [already skipped](https://github.com/openshift/origin/blob/master/test/extended/router/headers.go#L44), I'm probably misunderstanding something)

It is skipped on platforms that use the `LoadBalancerService` endpoint publishing strategy type, but presumably OpenStack is using `HostNetwork`.

ironcladlou:
@iamemilio can you go ahead and delete the [inline skip](https://github.com/openshift/origin/blob/master/test/extended/router/headers.go#L44) as part of this? Thank you!

iamemilio:
Yes, should I move it to one of the disabled pools so it doesnt re-enable for other platforms?

ironcladlou:
Yeah, let's skip it everywhere. I opened https://bugzilla.redhat.com/show_bug.cgi?id=1752646 which you can cite in the comment to track re-enabling it. Thanks again!

ironcladlou:
Thanks!

/lgtm

PR #23806, opened 2019-09-16
UPSTREAM: 80004: Prefer to delete doubled-up pods of a ReplicaSet

Miciah:
When scaling down a ReplicaSet, delete doubled up replicas first, where a "doubled up replica" is defined as one that is on the same node as an active replica belonging to a related ReplicaSet.  ReplicaSets are considered "related" if they have a common controller (typically a Deployment).

My intention with this change is to make a rolling update of a Deployment scale down the old ReplicaSet as it scales up the new ReplicaSet by deleting pods from the old ReplicaSet that are colocated with ready pods of the new ReplicaSet.  This change in the behavior of rolling updates can be combined with pod affinity rules to preserve the locality of a Deployment's pods over rollout.

A specific scenario that benefits from this change is when a Deployment's pods are exposed by a Service that has type "LoadBalancer" and external traffic policy "Local".  In this scenario, the load balancer uses health checks to determine whether it should forward traffic for the Service to a particular node.  If the node has no local endpoints for the Service, the health check will fail for that node.  Eventually, the load balancer will stop forwarding traffic to that node.  In the meantime, the service proxy drops traffic for that Service.  Thus, in order to reduce risk of dropping traffic during a rolling update, it is desirable preserve node locality of endpoints.

* `vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go` (`ActivePodsWithRanks`): New type to
sort pods using a given ranking.
* `vendor/k8s.io/kubernetes/pkg/controller/replicaset/replica_set.go` (`getReplicaSetsWithSameController`): New method.  Given a ReplicaSet, return all ReplicaSets that have the same owner.
(`manageReplicas`): Call `getIndirectlyRelatedPods`, and pass its result to `getPodsToDelete`.
(`getIndirectlyRelatedPods`): New method.  Given a ReplicaSet, return all pods that are owned by any ReplicaSet with the same owner.
(`getPodsToDelete`): Add an argument for related pods.  Use related pods and `ActivePodsWithRanks` to take into account whether the pod is doubled up.
* `vendor/k8s.io/kubernetes/pkg/controller/replicaset/replica_set_test.go` (`newReplicaSet`): Set `OwnerReferences` on the ReplicaSet.
(`newPod`): Set a unique UID on the pod.
(`byName`): New type to sort pods by name.
(`TestRelatedPodsLookup`): New test for `getIndirectlyRelatedPods`.
(`TestGetPodsToDelete`): Add a "ready and colocated with another ready pod vs not colocated, diff < len(pods)" test case to verify that a doubled-up pod gets preferred for deletion.  Augment the "various pod phases and conditions, diff = len(pods)" test case to ensure that scale-down still selects doubled-up pods if there are not enough other pods to scale down. Augment the "various pod phases and conditions, diff < len(pods)" test case to ensure that not-ready pods are preferred over ready but doubled-up pods.
* `vendor/k8s.io/kubernetes/pkg/controller/replicaset/BUILD`: Regenerate.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go
+++ b/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go
@@ -771,6 +771,94 @@ func (s ActivePods) Less(i, j int) bool {
 	return false
 }
 
+// ActivePodsWithRanks is a sortable list of pods and a list of corresponding
+// ranks which will be considered during sorting.  The two lists must have equal
+// length.  After sorting, the pods will be ordered as follows, applying each
+// rule in turn until one matches:
+//
+// 1. If only one of the pods is assigned to a node, the pod that is not
+//    assigned comes before the pod that is.
+// 2. If the pods' phases differ, a pending pod comes before a pod whose phase
+//    is unknown, and a pod whose phase is unknown comes before a running pod.
+// 3. If exactly one of the pods is ready, the pod that is not ready comes
+//    before the ready pod.
+// 4. If the pods' ranks differ, the pod with greater rank comes before the pod
+//    with lower rank.
+// 5. If both pods are ready but have not been ready for the same amount of
+//    time, the pod that has been ready for a shorter amount of time comes
+//    before the pod that has been ready for longer.
+// 6. If one pod has a container that has restarted more than any container in
+//    the other pod, the pod with the container with more restarts comes
+//    before the other pod.
+// 7. If the pods' creation times differ, the pod that was created more recently
+//    comes before the older pod.
+//
+// If none of these rules matches, the second pod comes before the first pod.
+//
+// The intention of this ordering is to put pods that should be preferred for
+// deletion first in the list.
+type ActivePodsWithRanks struct {
+	// Pods is a list of pods.
+	Pods []*v1.Pod
+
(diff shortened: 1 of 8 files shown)


Miciah:
Looks like problems pulling from the internal registry.
/test e2e-aws-upgrade

ironcladlou:
@smarterclayton PTAL, we need to get this in soon.

tnozicka:
this broke https://prow.svc.ci.openshift.org/job-history/origin-ci-test/logs/release-openshift-openshift-ansible-e2e-aws-scaleup-rhel7-4.3

p0lyn0mial:
I lost that commit during rebase, let me know if we want to bring it back.

Miciah:
Ah, are you referring to https://bugzilla.redhat.com/show_bug.cgi?id=1765756?

tnozicka:
> Ah, are you referring to https://bugzilla.redhat.com/show_bug.cgi?id=1765756?

correct, https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-openshift-ansible-e2e-aws-scaleup-rhel7-4.3/184#1:build-log.txt%3A5981

Miciah:
#24057 brings this commit back and adds a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1765756.

PR #23809, opened 2019-09-17
Bug 1752582: e2e: wait longer for authz checks

enj:
Group based permissions can take a long time to be effective due to the extra layer of indirection via the group reverse index.

Signed-off-by: Monis Khan <mkhan@redhat.com>

@deads2k


benjaminapetersen:
/lgtm

on behalf of @deads2k

PR #23813, opened 2019-09-17
UPSTREAM: <carry>: make the namespace controller recheck once a minute

deads2k:
This should make the ns-controller more reactive.


enj:
/retitle Bug 1752578:  UPSTREAM: <carry>: make the namespace controller recheck once a minute

PR #23814, opened 2019-09-17
Bug 1752646: Move templateinsance impersonation to broken skip tag 

iamemilio:
For some reason this test doesn't always get skipped on the openstack platform, so we are moving it with the other broken tests. This should be more consistent with other skipped tests, and will prevent the test from running on all platforms as originally intended.

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -415,7 +415,7 @@ var (
 
 			// Test fails on platforms that use LoadBalancerService and HostNetwork endpoint publishing strategy
 			`\[Conformance\]\[Area:Networking\]\[Feature:Router\] The HAProxy router should set Forwarded headers appropriately`, // https://bugzilla.redhat.com/show_bug.cgi?id=1752646
-
+			`\[Conformance\]\[templates\] templateinstance impersonation tests should pass impersonation creation tests`,         // https://bugzilla.redhat.com/show_bug.cgi?id=1752910
 			// requires a 1.14 kubelet, enable when rhcos is built for 4.2
 			"when the NodeLease feature is enabled",
 			"RuntimeClass should reject",
--- a/test/extended/templates/templateinstance_impersonation.go
+++ b/test/extended/templates/templateinstance_impersonation.go
@@ -250,14 +250,6 @@ var _ = g.Describe("[Conformance][templates] templateinstance impersonation test
 	})
 
 	g.It("should pass impersonation update tests", func() {
-		g.Skip("Bug 1731222: skip template tests until we determine what is broken")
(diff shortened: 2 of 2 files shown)


test/extended/util/test.go
```diff
@@ -415,7 +415,7 @@ var (
 
 			// Test fails on platforms that use LoadBalancerService and HostNetwork endpoint publishing strategy
 			`\[Conformance\]\[Area:Networking\]\[Feature:Router\] The HAProxy router should set Forwarded headers appropriately`, // https://bugzilla.redhat.com/show_bug.cgi?id=1752646
-
+			`\[Conformance\]\[templates\] templateinstance impersonation tests should pass impersonation creation tests`,         // https://bugzilla.redhat.com/show_bug.cgi?id=1752910
```
bparees: what happened to bug 1731222?   it looks like it was fixed, so shouldn't we be re-enabling this test?

bparees: also not clear why this is a preferred way to skip the test over how it was being done previously?

bparees: (can ignore my second comment, I see your PR summary, though that is weird).

iamemilio: I know very little of the status of the template tests, but what I do know is that this test was already disabled on all platforms, which is what this was for:
```go
g.Skip("Bug 1731222: skip template tests until we determine what is broken")
```

do you know anyone who may be familiar with this test suite who could comment?

bparees: it's owned by @openshift/sig-developer-experience.  Given that the bug being referenced as the reason for disabling the test, has been closed, they should investigate re-enabling the test or why it is still failing (And open a new bug).

(but if you need to do this PR in the meantime to make progress, that's fine, i just don't want the test to be forever disabled as we lose track of the history of why it was disabled/plans to re-enable it)

iamemilio: OpenStack needs its tests green to GA with 4.2, so if we dont hear from them in a few hours lets disable it.

adambkaplan:
nack on this - per @bparees we should consider re-enabling these tests. These tests are flaking in the `BeforeEach` block, which I believe does run and is not flaking on other platforms.

As an example, https://testgrid.k8s.io/redhat-openshift-release-informing#redhat-canary-openshift-ocp-installer-e2e-openstack-4.2 ultimately fails due to the following error:

```
Sep 17 12:50:11.948: INFO: Post https://api.kqy0cx0l-f21ca.shiftstack.ci:6443/apis/oauth.openshift.io/v1/oauthclients: read tcp 172.16.221.49:37684->128.31.27.245:6443: read: connection reset by peer
```

iamemilio:
ACK, Closing this PR

PR #23815, opened 2019-09-17
[4.1] Bug 1753996: UPSTREAM: 77661: Kubelet status manager sync the status of local pods

rphillips:
This enables the kubelet pod list to have correctly updated statuses for
static pods.

ref: https://github.com/kubernetes/kubernetes/pull/77661

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/kubelet/kubelet_getters.go
+++ b/vendor/k8s.io/kubernetes/pkg/kubelet/kubelet_getters.go
@@ -162,7 +162,15 @@ func (kl *Kubelet) getPodResourcesDir() string {
 // GetPods returns all pods bound to the kubelet and their spec, and the mirror
 // pods.
 func (kl *Kubelet) GetPods() []*v1.Pod {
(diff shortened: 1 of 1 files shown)


bparees:
@rphillips the 4.2 BZ referenced by the 4.1 BZ was closed as "not a bug", we need to see this fix verified/validated in 4.2 before backporting to 4.1.z.

rphillips:
@bparees fixed up the BZ chain. thanks!

bparees:
hmm..  https://bugzilla.redhat.com/show_bug.cgi?id=1753996 (4.1.z) is listed as dependent on https://bugzilla.redhat.com/show_bug.cgi?id=1720174 (3.11.z) which is unusual.

(it's also listed as dependent on https://bugzilla.redhat.com/show_bug.cgi?id=1753995 (4.2) which makes sense).

As things stand now, we'll need to see both of those dependent bugs in verified state before this can merge to 4.1.z.  (and of course it needs reviews).

eparis:
But the team needs to review/lgtm/approve.

PR #23817, opened 2019-09-17
Bug 1752983: Ensure the OLM e2e test loops correctly

smarterclayton:
The closure was capturing the last i value instead of testing each individual item. Change the test to loop correctly.

```
fail [github.com/openshift/origin/test/extended/operators/olm.go:36]: subscriptions list should have at least one item
Expected
    <bool>: true
to be false
```

All of the tests in the suite return that when a 4.1 cluster is run against 4.2.  Blocks us starting the skew test (there's almost certainly an actual bug here, this is just a blocker for knowing what the other bugs are).

No 4.2 change necessary because the test was already refactored as part of an epic.

### DIFF
--- a/test/extended/operators/olm.go
+++ b/test/extended/operators/olm.go
@@ -1,12 +1,10 @@
 package operators
 
 import (
-	"reflect"
+	"fmt"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
-	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


smarterclayton:
Nm, this is also broken similarly in 4.2.  Will spawn second bug.

jwforres:
/hold

still in progress, don't merge it yet

smarterclayton:
/hold cancel

The tests pass in 4.2

/test all

@jwforres can you verify the upstream bug?

PR #23818, opened 2019-09-17
Bug 1752988: OLM e2e test closes over a loop variable, thus doesn't test correctly

smarterclayton:
Correct the test to not use the loop variable, which should ensure each group is tested.


smarterclayton:
/assign @ecordell 

@jwforres

smarterclayton:
#23817 is 4.1

PR #23819, opened 2019-09-17
UPSTREAM: 82803: dump namespace object in e2e when it doesn't get deleted

deads2k:
Adds debugging information when namespaces don't get cleaned up.

/assign @p0lyn0mial


p0lyn0mial:
/lgtm

will be useful once we have condition

PR #23821, opened 2019-09-18
Bug 1753071: Allow certain tests to be skipped when performing skew testing

smarterclayton:
Skew tests create payloads that don't completely upgrade a cluster in order to get reproducible partial upgrade states. Some tests need to be disabled in order to allow the suite to succeed. Uses an env var for now, although future changes will switch to an argument flag.

Backport of #23820 to enable skewed node testing.


smarterclayton:
/assign @jwforres 

You'll have to verify the upstream bug (sorry)

PR #23822, opened 2019-09-18
Bug 1752800: UPSTREAM: 79235: kubectl attach test: wait for input before proceeding

giuseppe:
This way we know stdin is attached and we are in a known state.

ref: https://github.com/kubernetes/kubernetes/pull/79235

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/kubectl/kubectl.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/kubectl/kubectl.go
@@ -514,25 +514,14 @@ var _ = SIGDescribe("Kubectl client", func() {
 			nsFlag := fmt.Sprintf("--namespace=%v", ns)
 
 			ginkgo.By("executing a command with run and attach with stdin")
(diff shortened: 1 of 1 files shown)


mrunalp:
@smarterclayton we need approve for this. It should help with a test flake in vsphere upi.

mrunalp:
@derekwaynecarr @smarterclayton ptal

smarterclayton:
I’m concerned that changing this test broke the intent of the test.  Attach isn’t supposed to be racy.

giuseppe:
in the log file I see:
```
Sep 17 01:31:33.404: INFO: Pod "run-test-twgqt" of Job "run-test" should still be running
```

since the container exits immediately, I think it could happen that the pod terminates between the failed `ContainSubstring` tests and the next iteration (so that `CheckPodsRunningReady` fails).

nee1esh:
@smarterclayton @giuseppe What's next for this PR? We have a 4.2 blocker associated with this.

mrunalp:
I will look into this today to see if we want to retain older test.

nee1esh:
Giuseppe, can you please rebase and bring this PR and associated bug to closure?

giuseppe:
the fix is already present as part of the rebase, so closing it

PR #23828, opened 2019-09-18
Add validation for apiserver.spec.encryption.type

enj:
Signed-off-by: Monis Khan <mkhan@redhat.com>

Needs https://github.com/openshift/api/pull/429


sttts:
/hold

We have OpenAPI for this. Just define an enum. It's a oneliner in the API.

PR #23829, opened 2019-09-18
Bug 1733581: Change timeout in eviction e2e test to account for new kubelet delay in pod deletion

damemi:
This test is occasionally flaking with the following error
```
fail [k8s.io/kubernetes/test/e2e/scheduling/taints.go:442]: Sep 16 20:50:26.054: Failed to evict all Pods. 1 pod(s) is not evicted.
```
It seems that possibly the changes in https://github.com/openshift/machine-config-operator/pull/1120 affecting the kubelet timeout have also affected the time this test should wait to look for a proper pod deletion. The kubelet's default wait time is now 2 minutes, this updates the test's timeout to 130 seconds (to account for fudge).

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/scheduling/taints.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/scheduling/taints.go
@@ -17,18 +17,21 @@ limitations under the License.
 package scheduling
 
 import (
+	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


damemi:
cc @smarterclayton @rphillips @soltysh

damemi:
@ingvagabund suggested that this could also be due to Delete events being dropped, so I updated this test to:
1. increase the timeout
2. replace the listwatch with a regular wait loop that simply checks if the pods were deleted

PR #23833, opened 2019-09-20
Bug 1753953: exclude openstack infra namespace from the tolerations tests

Fedosin:
We don't need to test static pods in openshift-openstack-namespace here.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1753953

### DIFF
--- a/test/extended/operators/tolerations.go
+++ b/test/extended/operators/tolerations.go
@@ -29,7 +29,7 @@ var _ = Describe("[Feature:Platform][Smoke] Managed cluster should", func() {
 		// a pod in a namespace that begins with kube-* or openshift-* must come from our release payload
 		// TODO components in openshift-operators may not come from our payload, may want to weaken restriction
 		namespacePrefixes := sets.NewString("kube-", "openshift-")
-		excludedNamespaces := sets.NewString("openshift-kube-apiserver", "openshift-kube-controller-manager", "openshift-kube-scheduler", "openshift-etcd")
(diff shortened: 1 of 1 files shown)


test/extended/operators/tolerations.go
```diff
@@ -29,7 +29,7 @@ var _ = Describe("[Feature:Platform][Smoke] Managed cluster should", func() {
 		// a pod in a namespace that begins with kube-* or openshift-* must come from our release payload
 		// TODO components in openshift-operators may not come from our payload, may want to weaken restriction
 		namespacePrefixes := sets.NewString("kube-", "openshift-")
-		excludedNamespaces := sets.NewString("openshift-kube-apiserver", "openshift-kube-controller-manager", "openshift-kube-scheduler", "openshift-etcd")
+		excludedNamespaces := sets.NewString("openshift-kube-apiserver", "openshift-kube-controller-manager", "openshift-kube-scheduler", "openshift-etcd", "openshift-openstack-infra")
```
mandre: Instead of having a namespace per platform, how about using a generic namespace such as `openshift-infra` for all platforms that need static pods for their infra (right now openstack and kni).

I understand this will require a code change in MCO, so it doesn't have to happen now. Just throwing ideas out there.

Fedosin: As I know this is not recommended, because it can lead to a collision of names when different pods with the same name will be created in the same namespace depending on the platform.
But the unique namespace would be great if we called our like "openstack-infra-haproxy" and kni team "metal-infra-haproxy"

mandre: > As I know this is not recommended, because it can lead to a collision of names when different pods with the same name will be created in the same namespace depending on the platform.

This shouldn't be an issue if we assume openshift will only be installed on one platform at a time. Isn't it the case?

PR #23848, opened 2019-09-23
cmd/openshift-tests: Autodetect cloud

smarterclayton:
Ensure the cloud provider is autodetected where possible, although
not all cloud providers will behave this way.

### DIFF
--- a/test/extended/util/cloud/cloud.go
+++ b/test/extended/util/cloud/cloud.go
@@ -1,0 +1,91 @@
+package cloud
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	"github.com/openshift/origin/test/extended/util/azure"
+)
+
+// LoadConfig uses the cluster to setup the cloud provider config.
+func LoadConfig() (string, *e2e.CloudConfig, error) {
+	coreClient, err := e2e.LoadClientset()
+	if err != nil {
+		return "", nil, err
+	}
+	clientConfig, err := e2e.LoadConfig()
+	if err != nil {
+		return "", nil, err
+	}
+	client := configclient.NewForConfigOrDie(clientConfig)
+
+	infra, err := client.ConfigV1().Infrastructures().Get("cluster", metav1.GetOptions{})
+	if err != nil {
+		return "", nil, err
+	}
+	p := infra.Status.PlatformStatus
+	if p == nil {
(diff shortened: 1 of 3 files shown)


test/extended/util/cloud/cloud.go
```diff
@@ -0,0 +1,75 @@
...
+
+	var provider string
+	var projectID string
+	var region string
+	switch {
+	case p.AWS != nil:
+		provider = "aws"
+		region = p.AWS.Region
+	case p.GCP != nil:
+		provider = "gcp"
```
abhinavdahiya: this needs to be `gce`

abhinavdahiya: openshift Platform is `GCP`, the cloudprovider in k8s is `gce`

test/extended/util/cloud/cloud.go
```diff
@@ -0,0 +1,75 @@
...
+		return "", nil, err
+	}
+	p := infra.Status.PlatformStatus
+	if p == nil {
+		return "", nil, fmt.Errorf("status.platformStatus must be set")
+	}
+	if p.Type == configv1.NonePlatformType {
+		return "", nil, nil
+	}
+
```
abhinavdahiya: IMO we should restructure as 
```

## all the platform agnostic settings like zones, zone, multimaster etc.

## switch and return directly to avoid these variables `projectID` `region`

```

test/extended/util/cloud/cloud.go
```diff
@@ -0,0 +1,75 @@
...
+	switch {
+	case p.AWS != nil:
+		provider = "aws"
+		region = p.AWS.Region
+	case p.GCP != nil:
+		provider = "gcp"
+		projectID = p.GCP.ProjectID
+		region = p.GCP.Region
+	case p.Azure != nil:
+		provider = "azure"
```
abhinavdahiya: why stop here? why not setup the platform correctly when asked.

This assumes that somebody else will handle the azure stuff, if it's due to the tmp file handling it would be more appropriate to get that as option to this functions...

because when we need overrides like how to set service-endpoint-overrides for aws client when testing the config file could be handy..

test/extended/util/cloud/cloud.go
smarterclayton: ok

test/extended/util/cloud/cloud.go
smarterclayton: The caller covers that today, I can move it in here, but this is a utility method for the default.

smarterclayton:
This goes a step further and *forces* cloud provider lookup when it is a known string (one of the named ones).  We can't start using this in the templates, but should dramatically simplify end user flow.

smarterclayton:
Any other comments?

abhinavdahiya:
LGTM

/test e2e-azure

/test e2e-gcp

smarterclayton:
Wierd, I thought those two tests were disabled?

abhinavdahiya:
> Wierd, I thought those two tests were disabled?

the PRs are in flight
https://github.com/openshift/origin/pull/23889
https://github.com/openshift/origin/pull/23863

PR #23811, opened 2019-09-17
Bump to kubernetes 1.16.0 rc.2

p0lyn0mial:

PR #23860, opened 2019-09-25
Bug 1698543: bump libcontainer/runc and deps

rphillips:
Bump opencontainer/runc fork pin to pull in [this](https://github.com/opencontainers/runc/pull/2117/commits/518c855833c4920f9901f47f2a520425a33cceb4) commit to fix delegation not available errors.

**Note** This patch also pins openshift/{api,client-go,library-go,apiserver-library-go} to release branches.

The diff is large based on running `make update-deps` which is standard.

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1698543

/cc @mrunalp @sjenning


rphillips:
Compile error.

```
vendor/github.com/openshift/api/authorization/v1/generated.pb.go:59:28: undefined: proto.InternalMessageInfo
```

rphillips:
/hold

working on a different fix with @mrunalp

mrunalp:
@smarterclayton ptal

rphillips:
bot got updated to target 4.2.0... 

/bugzilla refresh

glide.yaml
```diff
@@ -87,13 +87,13 @@ import:
 
 # openshift second
 - package: github.com/openshift/api
-  version: master
+  version: release-4.2
```
wking: This feels like something that should happen automatically when we fork an origin branch off master.  Is there a plan to make that happen, or is it up to whoever makes the first post-fork vendor bump to do this?

wking:
Also, the Bugzilla bot is currently accepting [rhbz#1698543][1] because there is no blocking 4.3.0 bug; otherwise it would need a VERIFIED [4.3.0 bug][2].  Z patch managers could override that by setting labels manually, but we definitely don't want to fix this in 4.2 and then forget to fix it in 4.3.0.  Maybe 88baf7f651 (#23811) obsoletes this bump in the master branch, but I think our current policy is to have blocking 4.3.0 bugs so QE can verify that we aren't leaking the fix.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1698543
[2]: https://github.com/openshift/release/blob/ea151068df17ed44219ff2097ec7fec2228eaeaf/core-services/prow/02_config/_plugins.yaml#L1621

glide.yaml
rphillips: Yeah, I think this was missed upon branching.

rphillips:
@wking Got the 4.3 BZ and PR available.

blomquisg:
Had to reverse the BZ dependencies (4.3.0 should block 4.2.0).  This PR will be blocked until the 4.3.0 BZ is verified

nee1esh:
@eparis Can we get this approved for BZ process exception and cherry-picked for 4.2? 4.3 PR is merged. According to Seth "we don't have a recreate procedure for it and don't really know how it happens.  we just removed the code that makes it possible."

PR #23862, opened 2019-09-25
Bug 1753071: Skip alert test when test environment is known to be skewed

smarterclayton:
Alerts are fired in skewed environments which prevent tests from
passing. This is required for the 4.2 and 4.1 skew tests to pass.


smarterclayton:
@jwforres for 4.1.21 to allow the disruptive suite to run

smarterclayton:
@jwforres can you tag so we can get 4.1 to 4.2 skew green?

PR #23871, opened 2019-09-26
Remove travis CI from origin

smarterclayton:
All functionality is moved into other jobs.


rphillips:
Does this need cherry picked to the other release branches?

PR #23872, opened 2019-09-26
Bug 1727441: kube-proxy periodic iptables reloads are extremely disruptive in large clusters

danwinship:
kubelet and kube-proxy periodically resync all of their rules to iptables, even when nothing has changed, to ensure that things keep working even if the user runs `systemctl restart iptables` and accidentally deletes all of kube's rules. In large clusters, this get very disruptive and interferes with "real" iptables updates, sometimes causing service updates to take many minutes to appear.

Part of the problem was RHEL 7-specific (we were telling kube-proxy "wait 5 seconds to get the iptables lock" but telling all other iptables users "wait forever to get the iptables lock", meaning that when there was lots of lock contention, kube-proxy would normally end up being the loser). This was fixed upstream by https://github.com/kubernetes/kubernetes/pull/80368 and https://github.com/kubernetes/kubernetes/pull/82602, which are in 1.16 and so already in origin 4.3.

The rest of the problem was fixed upstream just-post-1.16 by https://github.com/kubernetes/kubernetes/pull/81517, by changing kubelet and kube-proxy to only reload their rules if the old ones were actually deleted.

In OCP 4.3, the fix is to backport 81517 to origin (this PR, for kubelet and openshift-tests) and sdn (https://github.com/openshift/sdn/pull/39, for kube-proxy and openshift-sdn's internal usage). (The origin and sdn PRs complement each other, but are independent, and can merge in either order.)

Then for 4.2 we'll need all that plus 80368+82602. Then in 4.1 and 3.11 the same thing but entirely in origin.


danwinship:
/hold
upstream PR needs some followup

nee1esh:
@danwinship what is the next step for this PR or bug?

knobunc:
https://github.com/kubernetes/kubernetes/pull/83498 and https://github.com/kubernetes-sigs/kind/pull/947 appear to be the follow-on work that should also be included.  @danwinship is that correct?  And will you have time to include these in the backport?  Or should we try to find someone else to pick this up?

knobunc:
@danwinship If the above PR includes that, can you pull the hold please?

danwinship:
83498 only applies to kube-proxy and so is needed in the sdn PR but not this one. The kind bug is not related. This PR is ready to go in. Sorry, I had updated it and then waited to see if the tests passed before commenting but then never commented. 
/hold cancel

nee1esh:
@danwinship , @tbielawa doesn't think this PR is within his domain and Seth is on vacation AFAICT. Is there anyone else who can review and approve this?

knobunc:
Thanks Dan

PR #23875, opened 2019-09-27
Bug 1756458: Separate upgrade flags for safety instead of abusing force

smarterclayton:
The --force flag is dangerous and potentially allows untrusted
content to be upgraded to accidentally. Instead, introduce two
new flags `--allow-explicit-upgrade` (for upgrading to something not
in availableVersions) and `--allow-unsafe-upgrade` (for upgrading
when another upgrade is in progress or the cluster is reporting
an error) and remove those checks from `--force`.

While this is an API change, it is necessary to ensure that users
do not accidentally get access to untrusted content when
performing upgrades across major versions in advance of graph
updates, or when they are upgrading in disconnected environments.

Backport of openshift/oc#109


enj:
/uncc

@stlaz @sttts @mfojtik

PR #23877, opened 2019-09-27
Bug 1758682: Make the quorum restore disruptive test to be resilient and add disruptive suite

smarterclayton:
Address some common failure modes in the quorum restore test by
externalizing SSH bastion support (provided by the test env),
hardening retry logic and other loops to take into account partial
failures during recovery, and using the framework SSH code.


smarterclayton:
@jwforres

PR #23878, opened 2019-09-27
[4.2.z] UPSTREAM: 79370: Migrate TaintManager to use watch for listing pods instead of expensive listing pods call

rphillips:
Backports https://github.com/kubernetes/kubernetes/pull/79370 to improve performance.


rphillips:
/hold

holding for 4.2.z

eparis:
You need to get a 4.2 and 4.3 bz. (and you need to QA to verify that the real problem is actually fixed in 4.3)

PR #23885, opened 2019-09-30
Bug 1694935: Sanitize user input when erroring on an unknown redirect handler [3.11]

stlaz:
Certain browsers might interpret the 500 error page as an HTML
page even though it's text/plain in some corner cases.

cc @enj


enj:
/unassign

@stlaz @sttts @mfojtik

PR #23863, opened 2019-09-25
Disable GPUDevicePluginAcrossRecreate test

rphillips:
We missed a Nvidia GPU test when we rebased to 1.16. Disable from E2E for now.

PR #23889, opened 2019-09-30
Skip 'recreate nodes' test until it is fixed upstream.

patrickdillon:
The test ` [Feature:Recreate] recreate nodes and ensure they function upon restart [Suite:openshift/conformance/parallel] [Suite:k8s]` should be labeled disruptive. It is causing all GCP CI tests to fail. We should disable until it is fixed upstream.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1756428


abhinavdahiya:
can we also file a BZ for `[sig-scheduling] GPUDevicePluginAcrossRecreate [Feature:Recreate] run Nvidia GPU Device Plugin tests with a recreation [Suite:openshift/conformance/parallel] [Suite:k8s]` to Node component and disable that test too?

rphillips:
Already PRed here: https://github.com/openshift/origin/pull/23863

rphillips:
We need to add `\[Feature:GPUDevicePlugin\]` as well to the skip list.

patrickdillon:
/assign @smarterclayton

we discussed this on slack on Friday. let's skip this test in gcp e2e until we can resolve the bz (get this test labeled `disruptive` upstream)

PR #23879, opened 2019-09-27
Bug 1755936: e2e: fix ingress metrics parallelism flake

ironcladlou:
This ingress metrics test manages an ingresscontroller that is not unique,
violating parallel test invariants. Because parallel tests try to own the same
ingresscontroller and tend to overlap, one test deletes the resource of the
other, causing assertion failures.

Use a unique name for newly created ingresscontrollers to ensure the tests can
safely execute in parallel.

PR #23892, opened 2019-10-01
Bug 1755936: e2e: stabilize ingress metrics tests

ironcladlou:
Sometimes Prometheus seems to report not having scraped the target
for the ingresscontroller created during this test, even after 4 minutes.

Since this test for the user-defined ingresscontroller seems inessential
compared to the latest ingress-operator e2e tests, rather than increasing
the timeout or spending additional effort to root-cause the scrape delay,
just delete the test to stabilize the test suite.

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -31,8 +31,6 @@ import (
 
 	"github.com/openshift/origin/test/extended/networking"
 	exutil "github.com/openshift/origin/test/extended/util"
-
-	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
 const waitForPrometheusStartSeconds = 240
@@ -226,49 +224,6 @@ var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
 			execPod := exutil.CreateCentosExecPodOrFail(oc.AdminKubeClient(), ns, "execpod", nil)
 			defer func() { oc.AdminKubeClient().CoreV1().Pods(ns).Delete(execPod.Name, metav1.NewDeleteOptions(1)) }()
 
-			g.By("creating a new ingresscontroller")
-			replicas := int32(1)
-			ingress := &operatorv1.IngressController{
-				ObjectMeta: metav1.ObjectMeta{
-					Namespace: "openshift-ingress-operator",
-					Name:      "prometheus",
-				},
-				Spec: operatorv1.IngressControllerSpec{
-					Domain:   "prometheus.e2e.openshift.example.com",
-					Replicas: &replicas,
-					EndpointPublishingStrategy: &operatorv1.EndpointPublishingStrategy{
-						Type: operatorv1.PrivateStrategyType,
-					},
-					RouteSelector: &metav1.LabelSelector{
-						MatchLabels: map[string]string{
-							"e2e-test": ns,
-						},
-					},
-				},
-			}
-			_, err := oc.AdminOperatorClient().OperatorV1().IngressControllers(ingress.Namespace).Create(ingress)
-			o.Expect(err).NotTo(o.HaveOccurred())
-
-			defer func() {
-				dump := func(resources ...string) {
-					for _, resource := range resources {
-						out, err := oc.AsAdmin().Run("get").Args("--namespace", "openshift-ingress", resource, "-o", "json").Output()
-						if err == nil {
-							e2e.Logf("%s", out)
-						} else {
-							e2e.Logf("error getting %s: %v", resource, err)
-						}
-					}
-				}
-				dump("servicemonitors", "services", "endpoints", "pods")
-
-				if err := oc.AdminOperatorClient().OperatorV1().IngressControllers(ingress.Namespace).Delete(ingress.Name, metav1.NewDeleteOptions(1)); err != nil {
-					e2e.Logf("WARNING: failed to delete ingresscontroller '%s/%s' created during test cleanup: %v", ingress.Namespace, ingress.Name, err)
-				} else {
-					e2e.Logf("deleted test ingresscontroller '%s/%s'", ingress.Namespace, ingress.Name)
-				}
-			}()
-
 			var lastErrs []error
 			o.Expect(wait.PollImmediate(10*time.Second, 4*time.Minute, func() (bool, error) {
 				contents, err := getBearerTokenURLViaPod(ns, execPod.Name, fmt.Sprintf("%s/api/v1/targets", url), bearerToken)
@@ -282,7 +237,6 @@ var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
 				lastErrs = all(
 					// Is there a good way to discover the name and thereby avoid leaking the naming algorithm?
 					targets.Expect(labels{"job": "router-internal-default"}, "up", "^https://.*/metrics$"),
-					targets.Expect(labels{"job": "router-internal-prometheus"}, "up", "^https://.*/metrics$"),
 				)
 				if len(lastErrs) > 0 {
 					e2e.Logf("missing some targets: %v", lastErrs)
@@ -293,10 +247,8 @@ var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
 
 			g.By("verifying standard metrics keys")
 			queries := map[string][]metricTest{
-				`template_router_reload_seconds_count{job="router-internal-default"}`:    {metricTest{greaterThanEqual: true, value: 1}},
-				`haproxy_server_up{job="router-internal-default"}`:                       {metricTest{greaterThanEqual: true, value: 1}},
-				`template_router_reload_seconds_count{job="router-internal-prometheus"}`: {metricTest{greaterThanEqual: true, value: 1}},
-				`haproxy_server_up{job="router-internal-prometheus"}`:                    {metricTest{greaterThanEqual: true, value: 1}},
+				`template_router_reload_seconds_count{job="router-internal-default"}`: {metricTest{greaterThanEqual: true, value: 1}},
+				`haproxy_server_up{job="router-internal-default"}`:                    {metricTest{greaterThanEqual: true, value: 1}},
 			}
 			runQueries(queries, oc, ns, execPod.Name, url, bearerToken)
 		})


ironcladlou:
cc @smarterclayton @brancz 

See https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_release/5195/rehearse-5195-pull-ci-openshift-installer-master-e2e-vsphere/15/build-log.txt — in that case the target was returned but wasn't scraped after 4 minutes:

```son
"labels": {
    "endpoint": "metrics",
    "instance": "10.129.1.128:1936",
    "job": "router-internal-prometheus",
    "namespace": "openshift-ingress",
    "pod": "router-prometheus-857ff5bb67-mllk6",
    "service": "router-internal-prometheus"
  },
  "scrapeUrl": "https://10.129.1.128:1936/metrics",
  "lastError": "",
  "lastScrape": "0001-01-01T00:00:00Z",
  "health": "unknown"
```

It doesn't explain https://github.com/openshift/origin/pull/23879#issuecomment-536131153, in which the entire `labels` and scrape fields are missing from the _same job_.

From the e2e-vsphere run (all of these now include openshift-ingress state dumps), you can see something else confusing:

```json
{
        "discoveredLabels": {
          "__address__": "10.129.1.128:80",
          "__meta_kubernetes_endpoint_address_target_kind": "Pod",
          "__meta_kubernetes_endpoint_address_target_name": "router-prometheus-857ff5bb67-mllk6",
          "__meta_kubernetes_endpoint_node_name": "compute-0",
          "__meta_kubernetes_endpoint_port_name": "http",
          "__meta_kubernetes_endpoint_port_protocol": "TCP",
          "__meta_kubernetes_endpoint_ready": "true",
          "__meta_kubernetes_endpoints_name": "router-internal-prometheus",
          "__meta_kubernetes_namespace": "openshift-ingress",
          "__meta_kubernetes_pod_annotation_k8s_v1_cni_cncf_io_networks_status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.129.1.128\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
          "__meta_kubernetes_pod_annotation_openshift_io_scc": "restricted",
          "__meta_kubernetes_pod_annotationpresent_k8s_v1_cni_cncf_io_networks_status": "true",
          "__meta_kubernetes_pod_annotationpresent_openshift_io_scc": "true",
          "__meta_kubernetes_pod_container_name": "router",
          "__meta_kubernetes_pod_container_port_name": "http",
          "__meta_kubernetes_pod_container_port_number": "80",
          "__meta_kubernetes_pod_container_port_protocol": "TCP",
          "__meta_kubernetes_pod_controller_kind": "ReplicaSet",
          "__meta_kubernetes_pod_controller_name": "router-prometheus-857ff5bb67",
          "__meta_kubernetes_pod_host_ip": "139.178.73.4",
          "__meta_kubernetes_pod_ip": "10.129.1.128",
          "__meta_kubernetes_pod_label_ingresscontroller_operator_openshift_io_deployment_ingresscontroller": "prometheus",
          "__meta_kubernetes_pod_label_pod_template_hash": "857ff5bb67",
          "__meta_kubernetes_pod_labelpresent_ingresscontroller_operator_openshift_io_deployment_ingresscontroller": "true",
          "__meta_kubernetes_pod_labelpresent_pod_template_hash": "true",
          "__meta_kubernetes_pod_name": "router-prometheus-857ff5bb67-mllk6",
          "__meta_kubernetes_pod_node_name": "compute-0",
          "__meta_kubernetes_pod_phase": "Running",
          "__meta_kubernetes_pod_ready": "true",
          "__meta_kubernetes_pod_uid": "ba94137b-401a-41a3-a258-7865f9813fbd",
          "__meta_kubernetes_service_annotation_service_alpha_openshift_io_serving_cert_secret_name": "router-metrics-certs-prometheus",
          "__meta_kubernetes_service_annotation_service_alpha_openshift_io_serving_cert_signed_by": "openshift-service-serving-signer@1569943242",
          "__meta_kubernetes_service_annotation_service_beta_openshift_io_serving_cert_signed_by": "openshift-service-serving-signer@1569943242",
          "__meta_kubernetes_service_annotationpresent_service_alpha_openshift_io_serving_cert_secret_name": "true",
          "__meta_kubernetes_service_annotationpresent_service_alpha_openshift_io_serving_cert_signed_by": "true",
          "__meta_kubernetes_service_annotationpresent_service_beta_openshift_io_serving_cert_signed_by": "true",
          "__meta_kubernetes_service_label_ingresscontroller_operator_openshift_io_owning_ingresscontroller": "prometheus",
          "__meta_kubernetes_service_labelpresent_ingresscontroller_operator_openshift_io_owning_ingresscontroller": "true",
          "__meta_kubernetes_service_name": "router-internal-prometheus",
          "__metrics_path__": "/metrics",
          "__scheme__": "https",
          "job": "openshift-ingress/router-default/0"
        }
      },
```

Same thing was observed in the GCP run. The question is, how can `"job": "openshift-ingress/router-default/0"` be correct here given the servicemonitor/service/pod/endpoints reported in the cluster at the time? See the data starting at:

```
Oct  1 16:17:07.611: INFO: Running 'oc --namespace=e2e-test-prometheus-72bls --config=/tmp/admin.kubeconfig get --namespace openshift-ingress servicemonitors -o json'
```

I've now identified what look like two anomalies with Prometheus here.

1. Target appears but is missing scrape data/labels entirely and has _incorrect labels associated with another service_, and,
2. Target appears with correct scrape data but failed to be scraped in a timely fashion.

If everything in the openshift-ingress state dump looks right, this is now looking to me like some sort of issue with Prometheus?

brancz:
Can you share the ServiceMonitor definition you are using here? The discovered labels are misleading as that's just what prometheus saw during discovery time, and it will have seen a whole lot :) . Discovered labels, are _not_ the target labels, they are what Prometheus discovered and what it had _before_ relabeling. Relabeling typically overrides the job label with some combination of discovered labels, which is what is likely happening here. The initial job label is just what is generated from the ServiceMonitor definition you are using.

ironcladlou:
@brancz the state dumps are linked in https://github.com/openshift/origin/pull/23892#issuecomment-537210795. Here's the servicemonitor:

```json
{
  "apiVersion": "monitoring.coreos.com/v1",
  "kind": "ServiceMonitor",
  "metadata": {
    "creationTimestamp": "2019-10-01T16:12:58Z",
    "generation": 1,
    "name": "router-prometheus",
    "namespace": "openshift-ingress",
    "ownerReferences": [
      {
        "apiVersion": "apps/v1",
        "controller": true,
        "kind": "Deployment",
        "name": "router-prometheus",
        "uid": "0e95685c-4de9-46a7-8230-10a5fcd9eda5"
      }
    ],
    "resourceVersion": "123215",
    "selfLink": "/apis/monitoring.coreos.com/v1/namespaces/openshift-ingress/servicemonitors/router-prometheus",
    "uid": "d1ddea0c-8522-4012-8429-e017975f2584"
  },
  "spec": {
    "endpoints": [
      {
        "bearerTokenFile": "/var/run/secrets/kubernetes.io/serviceaccount/token",
        "interval": "30s",
        "path": "/metrics",
        "port": "metrics",
        "scheme": "https",
        "tlsConfig": {
          "caFile": "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt",
          "serverName": "router-internal-prometheus.openshift-ingress.svc"
        }
      }
    ],
    "namespaceSelector": {
      "matchNames": [
        "openshift-ingress"
      ]
    },
    "selector": {
      "matchLabels": {
        "ingresscontroller.operator.openshift.io/owning-ingresscontroller": "prometheus"
      }
    }
  }
}
```

Do you have any thoughts on the two specific points I raised in https://github.com/openshift/origin/pull/23892#issuecomment-537210795?

>1. Target appears but is missing scrape data/labels entirely and has incorrect labels associated with another service, and,
>2. Target appears with correct scrape data but failed to be scraped in a timely fashion.

If nothing obvious stands out, I'll open a monitoring BZ to follow up.

brancz:
The discovered labels can't come from the ServiceMonitor you liked, it has to be the ServiceMonitor in the `openshift-ingress` namespace called `router-default`. The one you shared is called `router-prometheus`. Please excuse my ignorance but either I don't know how to manually mangle the URL in https://github.com/openshift/origin/pull/23892#issuecomment-537210795 to see the artifacts or they're not linked.

In general this sounds like there might be two ServiceMonitors discovering the same targets.

ironcladlou:
>The discovered labels can't come from the ServiceMonitor you liked, it has to be the ServiceMonitor in the openshift-ingress namespace called router-default. The one you shared is called router-prometheus. Please excuse my ignorance but either I don't know how to manually mangle the URL in #23892 (comment) to see the artifacts or they're not linked.

>In general this sounds like there might be two ServiceMonitors discovering the same targets.

Yes, there are two monitors, that's my point. Anyway, I've spent too much time on this and all the data's there for somebody else to research.

brancz:
> Yes, there are two monitors, that's my point. Anyway, I've spent too much time on this and all the data's there for somebody else to research.

Two ServiceMonitors discovering the same target is a problem, we shouldn't be doing that, as it causes duplicate time-series. My understanding is that you control both of these, so why not fix that?

ironcladlou:
>Two ServiceMonitors discovering the same target is a problem, we shouldn't be doing that, as it causes duplicate time-series. My understanding is that you control both of these, so why not fix that?

**Please look at the data I've posted here and elsewhere.** The servicemonitor selectors _do not appear to intersect_. Nor do the resources selected at each layer. The data our operator is persisting appears to be correct.

I'm considering this conversation closed unless someone wants to actually engage with the data and claims.

If you are willing to rely on speculation to discount my concerns as the ingress operator doing something wrong, fine, we'll agree to disagree and consider it case closed. I'm unwilling to continue producing concrete diagnostic data nobody wants to take the time to understand.

brancz:
I'm sorry I don't think I explained myself enough in my previous comments. Let me explain in a bit more in-depth why the discovered labels are not wrong, but I agree they are misleading.

Each ServiceMonitor generates a Prometheus scrape config. Each of those scrape configs will discover _all_ Services, but then filter down those to the ones that match the selector. In a way a ServiceMonitor generates a config that immitates the Kubernetes selector behavior, but it will still discover _all_ Services initially, that why we see all possible combinations, even those that don't actually produce a target, because they don't match the selector in the discovered labels.

Judging by the up metric, the targets are being scraped just fine.

Screenshot from 2019-10-02 18-52-54

Further I double checked, that there are actually samples being ingested, and it seems that is also the case:

Screenshot from 2019-10-02 18-52-01

To me it seems all metrics are being ingested correctly, we're definitely ingesting metrics from the OpenShift router. My feeling is that we should change the metrics/query that we test to include the exact namespace + route, instead of a job label, as the end of the day we don't care about the job.

ironcladlou:
>To me it seems all metrics are being ingested correctly, we're definitely ingesting metrics from the OpenShift router. My feeling is that we should change the metrics/query that we test to include the exact namespace + route, instead of a job label, as the end of the day we don't care about the job.

You're right, they are ingested correctly. Sometimes. Except when they're apparently not, which is what the bug is about, and is what all of my anomalous data points are about. So far I see nothing which explains the data in front of us.

PR #23893, opened 2019-10-01
Bug 1776595: Enable conformance-k8s for 1.16

smarterclayton:


smarterclayton:
This has to be backported now, may have to set up parallelism for it to pass within 3h

smarterclayton:
I have to run a `rename` on the e2e tests after serial to prevent them from getting stomped (since we have to submit them)

smarterclayton:
This is ready and results submitted.

/retest

PR #23897, opened 2019-10-02
Bug 1711600: UPSTREAM: 82424:Use pod + nsenter instead of SSH in mount propagation tests

jsafrane:
\+ enable the test downstream

https://github.com/kubernetes/kubernetes/pull/82424
https://bugzilla.redhat.com/show_bug.cgi?id=1711600

@openshift/storage, PTAL


jsafrane:
/assign @mfojtik @deads2k 
for approval

PR #23901, opened 2019-10-02
Bug 1757807: e2e: stabilize ingress metrics tests

ironcladlou:
Sometimes Prometheus seems to report not having scraped the target
for the ingresscontroller created during this test, even after 4 minutes.

Since this test for the user-defined ingresscontroller seems inessential
compared to the latest ingress-operator e2e tests, rather than increasing
the timeout or spending additional effort to root-cause the scrape delay,
just delete the test to stabilize the test suite.

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -32,8 +32,6 @@ import (
 
 	"github.com/openshift/origin/test/extended/networking"
 	exutil "github.com/openshift/origin/test/extended/util"
-
-	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
 const waitForPrometheusStartSeconds = 240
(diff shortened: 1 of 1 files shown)


s-urbaniak:
@ironcladlou thanks a lot for the test improvement. I think that 4.2.z is not open yet for development, but independently of that I have a question/request. Could we move this test to a different subfolder, owned by the ingress team that performs that test?

Inside here we really should only contain tests that are targeted at the monitoring stack itself, not tests owned by different operators.

wdyt?

brancz:
I do agree, that only cross component things should be in origin, these types of tests should be in the respective component test suites I believe.

ironcladlou:
Ingress and monitoring are both required/installed-by-default/OOTB components of the platform which are represented in conformance testing. Can you help me understand why it's inappropriate to test the integration of two OOTB components? Changes to the monitoring stack which could impact ingress will be surfaced sooner if the tests run here, and vice versa.

cc @smarterclayton

s-urbaniak:
@ironcladlou i am not arguing about having the e2e test here in origin, just in a different subfolder, i.e. `test/extended/ingress/ingress.go` with a separate `OWNERS` file.

ironcladlou:
@s-urbaniak 
>i am not arguing about having the e2e test here in origin, just in a different subfolder, i.e. test/extended/ingress/ingress.go with a separate OWNERS file.

I see, thanks. Sure.

brancz:
Looks like this fails consistently.

/hold

ironcladlou:
The failures I see are API throttling... let's try again

/hold cancel

ironcladlou:
@knobunc @smarterclayton ci/prow/e2e-aws-upgrade has been failing way too long on some throttling issue, please override it or help me understand what's happening?

crawford:
AWS throttled our API requests.

/retest

ironcladlou:
@knobunc @smarterclayton I have no idea what's up with these throttling errors, but this stability fix has been stuck in the merge queue for like 2 weeks now. Can somebody please override the failing job?

PR #22855, opened 2019-05-17
Bug 1709178: Prevent egress DNS request flooding

ironcladlou:
Before this change, zero-TTL records would cause the `EgressDNS` controller to
query for updates to those records in a tight loop with no rate limiting.

Prevent the flooding behavior by defaulting TTL to 30 minutes when the provided
TTL is zero

An alternative considered was to store the true TTL internally but enforce a
minimum sync interval through `EgressDNS.GetMinQueryTime`, which may also be
useful in its own right as a hedge against future bugs. Using a rate limited
work queue would be another possibility.

PR #23904, opened 2019-10-02
Bug 1709178: Prevent egress DNS request flooding

ironcladlou:
Before this change, zero-TTL records would cause the `EgressDNS` controller to
query for updates to those records in a tight loop with no rate limiting.

Prevent the flooding behavior by defaulting TTL to 30 minutes when the provided
TTL is zero.

An alternative considered was to store the true TTL internally but enforce a
minimum sync interval through `EgressDNS.GetMinQueryTime`, which may also be
useful in its own right as a hedge against future bugs. Using a rate limited
work queue would be another possibility.

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -14,6 +14,7 @@ import (
 )
 
 const (
+	// defaultTTL is used if an invalid or zero TTL is provided.
 	defaultTTL = 30 * time.Minute
 )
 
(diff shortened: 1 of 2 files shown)


ironcladlou:
@danwinship @dcbw simple cherry-pick of https://github.com/openshift/origin/pull/22855

ironcladlou:
@knobunc

PR #23905, opened 2019-10-03
test: Don't block on cloud providers without zones in autodetect

smarterclayton:
Autodetection was failing on vsphere because it has no zone labels.
Instead only set zone if we have one.


wking:
Any way to kick off a vSphere rehearsal?

smarterclayton:
You can launch a cluster and run the tests against it as I did.

wking:
Tested by someone works for me ;)

/lgtm

PR #23830, opened 2019-09-19
UPSTREAM: 82189: add conditions for remaining object totals during ns termination (fixes integration test)

p0lyn0mial:
fixes the broken integration test.

PR #23910, opened 2019-10-03
Bug 1766365: TestNamespaceCondition integration test fails

p0lyn0mial:
this pull deprecates https://github.com/openshift/origin/pull/23830


eparis:
are we still working on this? I'm don't understand how https://github.com/openshift/origin/pull/23830 and this relate.

p0lyn0mial:
> are we still working on this? I'm don't understand how #23830 and this relate.

I think it is worth taking as this pull fixes an integration test. The previous pull #23830 had targeted `master` which became `4.3` and already has the fix.

p0lyn0mial:
@eparis please tag the pull.

crawford:
I don't see where this change has first been verified against 4.3. Can you update the BZ with its parent (targeted against 4.3)?

derekwaynecarr:
the test had already merged upstream.

PR #23913, opened 2019-10-03
Bug 1758235: NetworkPolicy performance fixes [3.11]

danwinship:
Backport of https://github.com/openshift/sdn/pull/36 and https://github.com/openshift/sdn/pull/42 to 3.11.

This ended up being slightly different from the 4.3/4.2/4.1 version because 3.11 doesn't have namespaceSelector+podSelector support so the `selectNamespacesInternal` changes didn't apply completely cleanly and had to be rewritten a bit.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/golang/glog"
 
-	kapierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	ktypes "k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -304,6 +339,49 @@ func (np *networkPolicyPlugin) SyncVNIDRules() {
...
+//
+// Yes, if a selector matches against multiple labels then the order they appear in
+// cacheKey here is non-deterministic, but that just means that, eg, we might compute the
+// results twice rather than just once, and twice is still better than 10,000 times.
+func (np *networkPolicyPlugin) selectNamespacesInternal(selector labels.Selector) map[string]uint32 {
+	cacheKey := selector.String()
+	match := np.nsMatchCache[cacheKey]
+	if match == nil {
+		match = &npCacheEntry{selector: selector, matches: make(map[string]uint32)}
+		for vnid, npns := range np.namespaces {
```
squeed: This is a bit different from https://github.com/openshift/sdn/pull/36/files#diff-770c2f485d06d6a759ebe04d107047eaR317 - any reason for that?

danwinship: > This ended up being slightly different from the 4.3/4.2/4.1 version because 3.11 doesn't have namespaceSelector+podSelector support so the selectNamespacesInternal changes didn't apply completely cleanly and had to be rewritten a bit.

squeed: OK, as long as it wasn't accidentally a backport of an old patch.

yocum137:
\o/

PR #23921, opened 2019-10-04
Bug 1751640: UPSTREAM: 83108: Use pod + nsenter instead of SSH in block volume tests

jsafrane:
cc @openshift/storage 
https://bugzilla.redhat.com/show_bug.cgi?id=1751640
https://github.com/kubernetes/kubernetes/pull/83108


jsafrane:
/assign @mfojtik @deads2k 
for approval

PR #23922, opened 2019-10-04
billion laughs

deads2k:
How I did it.

1. ran hack/cherry-pick.sh.  This picked the reaction commit, but not the bump due to a quick in the script.
2. manually updated glide.yaml to pull in the "correct" level of yaml
3. ran make `update-deps`
4. saw that the publishing bot is broken again, paging @sttts and @mfojtik 
5. created two commits, one bump to pick to previous release with just the adjustment we need.  The second contains everything except k8s.io/*.  ordinarily we wouldn't do this and we would fix publishing, but see CVE problem.


jottofar:
@sttts and @mfojtik, can one of you check this please

glide.lock
```diff
@@ -2133,7 +2125,7 @@ imports:
   subpackages:
   - config/v1beta1
 - name: k8s.io/kubernetes
-  version: a3d846c6abdf03e54525c3ebca35c40441462261
+  version: 3538c7842b1e43174e181b732cd3147a56711427
```
sttts: how did you do this bump without overwriting the changes of the UPSTREAM commit? Did you just leave out the kube files in the bump, i.e. these SHAs are now wrong?

PR #23925, opened 2019-10-07
tests/extented/prometheus: Disable telemetry e2e for remote-write move

brancz:
This will be temporarily required in order to switch sending telemetry to Prometheus native remote-write capabilities (https://github.com/openshift/cluster-monitoring-operator/pull/434). We will want to merge this PR and https://github.com/openshift/cluster-monitoring-operator/pull/434 quickly after each other and then enable the here commented out check that verifies, that remote-write successfully sent data instead of the bespoke telemeter client.

Telemetry via remote-write can only be accepted by the v2 receiving infrastructure, so we need to hold off on merging this and https://github.com/openshift/cluster-monitoring-operator/pull/434 until the infrastructure is completely migrated.

@s-urbaniak @paulfantom @LiliC @squat @smarterclayton 

/hold

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -51,6 +51,15 @@ var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
 
 	g.Describe("when installed on the cluster", func() {
 		g.It("should report telemetry if a cloud.openshift.com token is present", func() {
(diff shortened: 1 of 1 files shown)


brancz:
https://github.com/openshift/cluster-monitoring-operator/pull/434 is now ready to go, so we can move forward here. 

/hold cancel

PR #23918, opened 2019-10-04
[release-3.11] Bug 1758375: fix haproxy reload crash when processing ECDSA keys

Miciah:
This is manual cherry-pick of openshift/router#39.

PR #23927, opened 2019-10-07
Bug 1749653: (attempt to) fix ECDSA formatted private key usage in route

Alveel:
An ECDSA private key only specifies "EC PRIVATE KEY", not "ECDSA PRIVATE KEY": https://tools.ietf.org/html/rfc5915

Wouldn't it be better to use `MarshalPKCS8PrivateKey` in the `*KeyBlockVerifier` functions? The [go documentation](https://golang.org/pkg/crypto/x509/#MarshalECPrivateKey) specifies

>For a more flexible key format which is not EC specific, use MarshalPKCS8PrivateKey.

This is probably incomplete, I tried to set up a test environment to build everything but was having quite some issues there. I'd need to spend quite some more time to set that up. Any/all comments/additions/criticism etc. welcome.

Also, `gofmt` changed some of the spacing :)

### DIFF
--- a/pkg/route/controller/routeapihelpers/validation.go
+++ b/pkg/route/controller/routeapihelpers/validation.go
@@ -62,7 +62,7 @@ func privateKeyBlockVerifier(block *pem.Block) (*pem.Block, error) {
 		}
 	case *ecdsa.PrivateKey:
 		block = &pem.Block{
-			Type: "ECDSA PRIVATE KEY",
+			Type: "EC PRIVATE KEY",
 		}
 		if block.Bytes, err = x509.MarshalECPrivateKey(t); err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


ironcladlou:
This appears to be a duplicate of https://github.com/openshift/origin/pull/23918.

ironcladlou:
Closing as a dupe of https://github.com/openshift/origin/pull/23918. Let me know if I've made a mistake.

/close

Alveel:
I believe you are correct @ironcladlou. Maybe it's a good idea to internally link the bugzilla reports?

ironcladlou:
I believe all the bugzillas are appropriately cross-referenced from GitHub. The backport BZs won't link up automatically until https://bugzilla.redhat.com/show_bug.cgi?id=1723400 is verified, then everything should sort out. Thanks!

PR #23929, opened 2019-10-07
UPSTREAM: 83261: limit yaml/json decode size

jottofar:


jottofar:
@deads2k can you review

deads2k:
Can you add a description of how you produced this?  It appears to be missing the reaction commit, it doesn't have a glide.yaml update, and it seems to be updating many unrelated dependencies which should remain untouched.

jottofar:
- I started with release-4.2
- Did a make update-deps
- cherry-picked db8f23c2d39fb3ad89b0e68dc64df94dbf70743b, 9a706b3a6785029b1669e7aeae38c0851d3f5016 and c6f6f00d7cfab15ab1a3773755c5ccfe10aa3fb2 but NOT the glide.yaml commit. I thought the 'update-deps' would take care of that. 
- squashed the three commits into 1 and did the push

PR #23936, opened 2019-10-09
UPSTREAM: <drop>: Lower DescribeVolumes polling frequency

jsafrane:
Call `DescribeVolumes` less frequently so OpenShift is not throttled by AWS. `DescribeVolumes` is basically the only Kubernetes call that suffers API throttling by AWS.

This is a temporary patch that adds extra logs here and there to get accurate numbers. We should revert it in a week or two, analyze if it helped with API throttling and either post it upstream or find a better solution.


jsafrane:
Removal is tracked here: https://bugzilla.redhat.com/show_bug.cgi?id=1759998

PR #23944, opened 2019-10-10
UPSTREAM: 00000: add debugging for upgrade tests

deads2k:
Adding debugging lines for failed upgrade tests.


wking:
is going to need a 4.2 bug and a VERIFIED master bug (possibly via a master PR) to land, or we'll need the z patch managers to waive that requirement for this PR and twiddle labels manually.

PR #23948, opened 2019-10-10
Bug 1757710: CVE-2019-11253 billion laughs

jottofar:
Process followed:
1. modified and committed glide.yaml
2. 'make update-deps' and committed
3. cherry-picked 880a053dde315c44c0f2c5e905eeb9bfc6d44074


sttts:
sigs.k8s.io/yaml was not used at that time. Probably we relied on gopkg.in/yaml.v2.

sttts:
Replace sigs.k8s.io/yaml with `sigsyaml "github.com/ghodss/yaml"` and it will fix the new unit test.

vendor/k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/util/yaml/decoder_test.go
```diff
@@ -197,8 +197,8 @@ stuff: 1
 		t.Fatal("expected error with yaml: violate, got no error")
 	}
 	fmt.Printf("err: %s\n", err.Error())
-	if !strings.Contains(err.Error(), "yaml: line 2:") {
-		t.Fatalf("expected %q to have 'yaml: line 2:' found a tab character", err.Error())
```
sttts: have you checked what is happening here?

sttts:
Seeing >1GB yaml test memory. But the same is true in upstream.

```
BenchmarkYAMLLimits/3MB_map_without_alias/sigsyaml-12            	       1	2108367117 ns/op	1186317152 B/op	15728782 allocs/op
BenchmarkYAMLLimits/3MB_map_without_alias/utilyaml-12            	       1	2017328812 ns/op	1186317024 B/op	15728781 allocs/op
```

We also have the same amount of memory limit for the unit tests for 4.2 and 3.11. Wondering why it OOMs with 3.11.

vendor/k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/util/yaml/decoder_test.go
jottofar: Not sure I understand the question. Are you seeing that error in the logs? If so can you point me there because I can't find it.

pkg/apps/strategy/support/acceptor.go
```diff
@@ -72,7 +72,8 @@ func (c *acceptAvailablePods) Accept(rc *corev1.ReplicationController) error {
 		case nil:
 			return nil
 		case watch.ErrWatchClosed:
-			fmt.Fprint(c.out, "Warning: acceptAvailablePods encountered %T, retrying", watch.ErrWatchClosed)
+			//fmt.Fprint(c.out, "Warning: acceptAvailablePods encountered %T, retrying", watch.ErrWatchClosed)
```
deads2k: this looks spurious.

pkg/apps/strategy/support/acceptor.go
```diff
@@ -72,7 +72,8 @@ func (c *acceptAvailablePods) Accept(rc *corev1.ReplicationController) error {
 		case nil:
 			return nil
 		case watch.ErrWatchClosed:
-			fmt.Fprint(c.out, "Warning: acceptAvailablePods encountered %T, retrying", watch.ErrWatchClosed)
```
deads2k: why did this need to change?

vendor/k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/util/yaml/decoder_test.go
deads2k: > Not sure I understand the question. Are you seeing that error in the logs? If so can you point me there because I can't find it.

he wants to know the same thing I do.  Why has the line number changed?  Was it consistently incorrect before?  Can you provide an example?

pkg/apps/strategy/support/acceptor.go
jottofar: I just changed that to get it to build in my env and forgot to finalize. Why it fails to build in my env I'm not sure. I'm building with go 1.11 vs. 1.7 but wouldn't think it would cause this issue.

vendor/k8s.io/kubernetes/pkg/kubectl/cmd/testdata/edit/testcase-apply-edit-last-applied-syntax-error/2.original
```diff
@@ -1,7 +1,7 @@
 # Please edit the 'last-applied-configuration' annotations below.
 # Lines beginning with a '#' will be ignored, and an empty file will abort the edit.
 #
-# The edited file had a syntax error: error parsing edited-file: error converting YAML to JSON: yaml: line 12: could not find expected ':'
+# The edited file had a syntax error: error parsing edited-file: error converting YAML to JSON: yaml: line 13: could not find expected ':'
```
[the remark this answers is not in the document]
jottofar: @deads2k I'm not sure how this worked before or if it did. So 'spec' is missing the ':'. In the 2.original file it would be normally indicated by the editor at line 12 (it does not include the two comment lines the editor adds telling you what the issue is so that's why it's 12 versus 14). However the 2.edited file adds a line, 'new-label1: foo1', which therefore makes the line number 13. So two changes occur, add a line and remove the ':'. These tests are not in 4.x.

jottofar:
/cc @deads2k @sttts To get unit tests to run to success:
1. Increased mem resources to avoid OOMKilled. I ended up doubling them but that's not necessarily what's needed as I just wanted to get pass the OOMKilled to get to the next issue.
2. TestYAMLLimits would get 2 minute timeout panic. I increased timeout to 10 minutes and then the test just hung and was killed (I'm assuming as I got no log) after 4 hours. I suspect possibly something is not thread safe so noop-ed TestJSONLimits which runs at same time as TestYAMLLimits and unit test completes. Based on that I added TestJSONLimits back in and ran 'got test' with '-p 1' to single thread each package level test and that completed successfully.

deads2k:
Wow that's a long unit test run.  i think I can live with it though.

pkg/apps/strategy/support/acceptor.go
jottofar: @deads2k I backed this change out. Not sure what caused the initial issue but I was using different version of go.

vendor/k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/util/yaml/decoder_test.go
jottofar: TestDecodeBrokenYAML: This same change was actually made by @deads2k in 4.1 (8c90407) although I didn't know it a t the time. So this is really a back port from 4.1. As to the cause, the error message lists the line number of yaml, which is 3, vs line number of json which is 2. I do not know why it was ever 2, perhaps the underlying ymal.v2 impl changed as that is the only diff -> moving to v2.2.4.

vendor/k8s.io/kubernetes/pkg/kubectl/cmd/testdata/edit/testcase-apply-edit-last-applied-syntax-error/2.original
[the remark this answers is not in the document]
jottofar: This is the test case that fails without my change:
vendor/k8s.io/kubernetes/pkg/kubectl/cmd TestEdit apply-edit-last-applied-syntax-error
See full failure message below.
See my full explanation in comment above. In short, I tested this on a 3.11 and 3.10 cluster and both are consistent with my change. The parsing is fine. What's driving the change is the 'editor' function leaves the added line in when it displays to the user so perhaps that functionality has changed at some point, e.g. if it displayed the original file with only the missing colon and not the added line the line number would be correct.

--- FAIL: TestEdit (1.43s)
    --- PASS: TestEdit/apply-edit-last-applied (0.07s)
    --- PASS: TestEdit/apply-edit-last-applied-list (0.06s)
    --- PASS: TestEdit/apply-edit-last-applied-list-fail (0.09s)
	edit_test.go:126: apply-edit-last-applied-syntax-error, step 2: diff in edit content:
		# Please edit the 'last-applied-configuration' annotations below.
		# Lines beginning with a '#' will be ignored, and an empty file will abort the edit.
		#
		# The edited file had a syntax error: error parsing edited-file: error converting YAML to JSON: yaml: line 1
		
		A: 3: could not find expected ':'
		#
		kind: Service
		metadata:
		  annotations: {}
		  labels:
		    app: svc1
		    new-label: foo
		  name: svc1
		  namespace: myproject
		spec
		  ports:
		    name: "80"
		    port: 81
		    protocol: TCP
		    targetPort: 81
		  sessionAffinity: None
		  type: ClusterIP
		status:
		  loadBalancer: {
		
		
		B: 2: could not find expected ':'
		#
		kind: Service
		metadata:
		  annotations: {}
		  labels:
		    app: svc1
		    new-label: foo
		  name: svc1
		  namespace: myproject
		spec
		  ports:
		    name: "80"
		    port: 81
		    protocol: TCP
		    targetPort: 81
		  sessionAffinity: None
		  type: ClusterIP
		status:
		  loadBalancer: {

PR #23950, opened 2019-10-11
[proof] TLS 1.3 does not brake origin

stlaz:
/hold

Fake bumps library-go to see if adding 1.3 ciphers breaks the cluster or not. The ciphers should automatically be picked for the kube-apiserver by `SetRecommendedGenericAPIServerConfigDefaults()`

https://github.com/openshift/library-go/pull/553


stlaz:
Looks like kube-apiserver might have been broken by the change, most probably because go 1.12 did not allow TLS 1.3 ciphers to be configurable, lets try with go 1.13

stlaz:
/close
proven

PR #23951, opened 2019-10-11
Fixes Bug 1763340: Fix targetPort in route created by ingress

### LINKED ISSUE #23937: Incorrect reconciliation of OpenShift Route based on K8s Ingress
When OpenShift reconciles and therefore creates an OpenShift `Route` for each Kubernetes `Ingress`, the property `spec.port.targetPort` is set with the name of the port on the `Pod` instead of the name of the port on the referenced `Service`. This gets problematic when they differ in their names, as the referenced name doesn't exist on the `Service`.

Responsible controller: https://github.com/openshift/origin/blob/v3.11.0/pkg/route/controller/ingress/ingress.go#L576

##### Version
OpenShift Master: v3.11.141
Kubernetes Master: v1.11.0+d4cacc0

##### Steps To Reproduce

1. Create a k8s `Service` which targets a named targetPort `default`(on the `Pod`) which differs from the name of the port on the service `http-default`.

```
oc apply -f - << EOF
apiVersion: v1
kind: Service
metadata:
  name: app
spec:
  ports:
  - name: http-default
    port: 8080
    protocol: TCP
    targetPort: default
  selector:
    deployment: app
EOF
```

2. Create a k8s `Ingress` which targets the previously created `Service` on its named service port `http-default`.

```
oc apply -f - << EOF
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: app
spec:
  rules:
  - host: app.test.com
    http:
      paths:
      - backend:
          serviceName: app
          servicePort: http-default
        path: /
EOF
```

##### Current Result

OpenShift creates a `Route` which targets the service `app` but the targetPort is the name of the `Pod` (`default`)  instead of the one of the `Service` (`http-default`).

```
oc get route app-6dqpg -o yaml
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  ...
spec:
  host: app.test.com
  path: /
  port:
    targetPort: default
  to:
    kind: Service
    name: app
...
```

Therefore no connection is possible via the `Route`. Even the OpenShift webconsole warns about this ("Route target port is set to "default", but service "app" has no port with that name.")

##### Expected Result

The created `Route` should target the name of the port on the `Service` instead of the one of the `Pod`.

##### Additional Information

* Relevant sourcecode location: https://github.com/openshift/origin/blob/v3.11.0/pkg/route/controller/ingress/ingress.go#L576
* Error seems to be present in 4.x: https://github.com/openshift/openshift-controller-manager/blob/release-4.4/pkg/route/ingress/ingress.go#L576 
* Creating an Route with `oc expose service app` creates a correct Route with the name of the service port
* The description of the property `route.spec.port.targetPort` is quite irritating. maybe this is the source of the error. (_The target port on pods selected by the service this route points to. If
     this is a string, it will be looked up as a named port in the target
     endpoints port list._)

derkoe:
Use the service port as the Route's targetPort (the service port may differ from the service targetPort).

### DIFF
--- a/pkg/route/controller/ingress/ingress.go
+++ b/pkg/route/controller/ingress/ingress.go
@@ -487,9 +487,9 @@ func newRouteForIngress(
 			return nil
 		}
 		tlsConfig = &routev1.TLSConfig{
-			Termination: routev1.TLSTerminationEdge,
-			Certificate: string(secret.Data[v1.TLSCertKey]),
-			Key:         string(secret.Data[v1.TLSPrivateKeyKey]),
+			Termination:                   routev1.TLSTerminationEdge,
+			Certificate:                   string(secret.Data[v1.TLSCertKey]),
+			Key:                           string(secret.Data[v1.TLSPrivateKeyKey]),
 			InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
 		}
 	}
@@ -583,14 +583,16 @@ func targetPortForService(namespace string, path *extensionsv1beta1.HTTPIngressP
 		expect := path.Backend.ServicePort.StrVal
 		for _, port := range service.Spec.Ports {
 			if port.Name == expect {
-				return &port.TargetPort
+				targetPort := intstr.FromString(port.Name)
+				return &targetPort
 			}
 		}
 	} else {
 		for _, port := range service.Spec.Ports {
 			expect := path.Backend.ServicePort.IntVal
 			if port.Port == expect {
-				return &port.TargetPort
+				targetPort := intstr.FromInt(int(port.Port))
+				return &targetPort
 			}
 		}
 	}
--- a/pkg/route/controller/ingress/ingress_test.go
+++ b/pkg/route/controller/ingress/ingress_test.go
@@ -369,7 +369,6 @@ func TestController_sync(t *testing.T) {
 			Spec: v1.ServiceSpec{
 				Ports: []v1.ServicePort{
 					{
-						Name:       "80-tcp",
 						Port:       80,
 						TargetPort: intstr.FromInt(8080),
 					},
@@ -635,7 +634,7 @@ func TestController_sync(t *testing.T) {
 							Name: "service-1",
 						},
 						Port: &routev1.RoutePort{
-							TargetPort: intstr.FromInt(8080),
+							TargetPort: intstr.FromString("http"),
 						},
 					},
 				},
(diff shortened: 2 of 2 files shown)


danehans:
Associated GH issue: https://github.com/openshift/origin/issues/23937

ironcladlou:
Thanks for the PR! Can you please open this against the `master` branch instead? We'll use an automated process (if possible) to handle any backports.

ironcladlou:
@smarterclayton https://github.com/openshift/api/blob/master/route/v1/types.go#L134

This PR _seems_ correct but I want to double check with you since this is a tricky subject.

Also, the nature of the unit tests changes here (i.e. no behavioral changes) lead me to believe that this case isn't actually being tested properly...

pkg/route/controller/ingress/ingress_test.go
```diff
@@ -635,7 +635,7 @@ func TestController_sync(t *testing.T) {
 							Name: "service-1",
 						},
 						Port: &routev1.RoutePort{
-							TargetPort: intstr.FromInt(8080),
+							TargetPort: "http",
```
danehans: Since the changes you introduce in `pkg/route/controller/ingress/ingress.go ` support using a port number or name, can you update the test cases accordingly?

danehans:
`could not wait for build: the build bin failed after 3m21s with reason DockerBuildFailed: Docker build strategy has failed.`

/retest

ironcladlou:
/ok-to-test cancel

This is targeting the wrong branch

derkoe:
> Thanks for the PR! Can you please open this against the `master` branch instead? We'll use an automated process (if possible) to handle any backports.

@ironcladlou this controller does not exist in the master branch - I think this has been replaced with https://github.com/openshift/cluster-ingress-operator in OpenShift 4.x

pkg/route/controller/ingress/ingress_test.go
derkoe: Actually I did not introduce that feature (this was already there) - just used the right port (which in this case has a name). I agree some more meaningful tests would be good - trying to add some.

mhofstetter:
@derkoe thanks for the PR which fixes my issue.

the code has been moved to https://github.com/openshift/openshift-controller-manager/blob/release-4.4/pkg/route/ingress/ingress.go#L576 in openshift 4.x

ironcladlou:
>@derkoe thanks for the PR which fixes my issue.

>the code has been moved to https://github.com/openshift/openshift-controller-manager/blob/release-4.4/pkg/route/ingress/ingress.go#L576 in openshift 4.x

This is correct, sorry! I forgot we split the repo out recently. We do eventually want to consider moving this into the ingress operator.

I think we can keep this PR targeting the 3.11 branch for now, but it'll have to merge to https://github.com/openshift/openshift-controller-manager master first.

danehans:
@derkoe do you plan on fixing this in master of https://github.com/openshift/openshift-controller-manager? If not, I'll submit a PR for the fix.

derkoe:
@danehans created the pr in https://github.com/openshift/openshift-controller-manager, added a test case for an int port and fixed some issues.

danehans:
/retitle Fixes Bug 1763340: Fix targetPort in route created by ingress

ironcladlou:
/hold

I perceive some confusion around this issue. Please keep the hold until @smarterclayton has weighed in.

derkoe:
I checked now why this tests works: https://github.com/openshift/origin/blob/release-3.11/test/extended/router/router.go

The reason for this is that OpenShift Routes ignore the `targetPort` when the service has only a single, unnamed port. OpenShift Console tells you this in the UI:
grafik

Maybe this test should be adapted too?

pkg/route/controller/ingress/ingress.go
```diff
@@ -583,14 +583,16 @@ func targetPortForService(namespace string, path *extensionsv1beta1.HTTPIngressP
...
+				return &targetPort
 			}
 		}
 	} else {
 		for _, port := range service.Spec.Ports {
 			expect := path.Backend.ServicePort.IntVal
 			if port.Port == expect {
-				return &port.TargetPort
+				targetPort := intstr.FromInt(int(port.Port))
+				return &targetPort
```
Miciah: This change is breaking the test in Origin.  The router uses this target port to look up the port in the _endpoints_ resource, not the _service_ resource, and although a port in a service resource and the corresponding port in the endpoints resource have the same _name_, the port _number_ in the endpoints resource comes from the _target port_ of the service.  For example, the following service resource:
```yaml
kind: Service
# ...
spec:
  ports:
  - name: http-default
    port: 8888
    targetPort: 8080
```
will have the following endpoints resource:
```yaml
kind: Endpoints
# ...
subsets:
- # ...
  ports:
  - name: http-default
    port: 8080
```
The following route resource will match the endpoints resource:
```yaml
kind: Route
# ...
spec:
  # ...
  port:
    targetPort: 8080
```
and so will the following:
```yaml
kind: Route
# ...
spec:
  # ...
  port:
    targetPort: http-default
```
but the following will not:
```yaml
kind: Route
# ...
spec:
  # ...
  port:
    targetPort: 8888
```

derkoe: OK - now I understand. The original bug #23937 only happens when the targetPort of the Service is a named (string) port of the Pod. When the targetPort is an int the endpointPort is already correct in the current implementation.

Miciah:
Forgot to close this PR.  https://github.com/openshift/openshift-controller-manager/pull/53 fixes the problem.  Thanks!
/close

PR #23962, opened 2019-10-14
test: Update junit naming to ensure disruption tests are recorded

smarterclayton:
Testgrid would omit disruption tests.


smarterclayton:

which is broken due to gap firewalls

smarterclayton:
/override ci/prow/e2e-aws-builds

was stuck, not relevant

PR #23981, opened 2019-10-16
Add validation for apiserver.spec.tlsSecurityProfile

stlaz:
Adds validation for the `apiserver.spec.tlsSecurityProfile` field

/assign @sanchezl 

Will require adding OpenSSL to IANA cipher suite names mapping in lib-go: https://github.com/openshift/library-go/pull/559


sanchezl:
Id the profile.Type value case-sensitive? e.g. : `tlsSecurityProfile: { type: "old" }` vs. `tlsSecurityProfile: { type: "Old" }`?

stlaz:
It is case-sensitive, the type is basically an enum of hard-coded values

stlaz:
I did not notice the build error, I was merely using the unit tests to run the code. Must be something in the bump, I'll see what can be done

stlaz:
Old grpc did not have the `WithCOntextDialer` function exported, bumped from `1.13` to `1.19.1`

stlaz:
Eh, I did not notice that the bump removed all of the changes to the apiserver admission? I did not realize that yesterday, but, eh, the code for kube apiserver should have been pinned?

stlaz:
Eh, my bad, this is probably a wrong repository now

stlaz:
There appears to be the fix for billion laughs patch on top in the branch that is pinned for the kube-apiserver, so that is probably (?) fine. Moved my patch after the bump so that the changes are not annihilated.

glide.yaml
```diff
@@ -95,7 +95,7 @@ import:
 
 # see if old grpc can let kubelet/crio communicate
 - package: google.golang.org/grpc
-  version: 168a6198bcb0ef175f7dacec0b8691fc141dc9b8
+  version: 3507fb8e1a5ad030303c106fef3a47c9fdad16ad
```
sttts: why?

stlaz: kube-apiserver would not compile as it relies on `grpc.WithCOntextDialer` being public, and it wasn't until this gRPC version

stlaz: code in question: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L134

stlaz:
cc @openshift/openshift-team-network-edge

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validate_apiserver.go
```diff
@@ -146,3 +156,101 @@ func validateAdditionalCORSAllowedOrigins(fieldPath *field.Path, cors []string)
...
+	errs := field.ErrorList{}
+
+	switch profile.Type {
+	case "":
+		if profile.Old != nil || profile.Intermediate != nil || profile.Modern != nil || profile.Custom != nil {
+			errs = append(errs, field.Invalid(typePath, profile.Type, "one of the profiles if set but 'type' field is empty"))
+		}
+	case configv1.TLSProfileOldType:
+		if profile.Old == nil {
+			errs = append(errs, field.Invalid(fieldPath.Child("old"), profile.Old, fmt.Sprintf(typeProfileMismatchFmt, profile.Type)))
```
sttts: field.Required

stlaz:
/hold
let me fix the field err

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validate_apiserver.go
```diff
@@ -146,3 +156,101 @@ func validateAdditionalCORSAllowedOrigins(fieldPath *field.Path, cors []string)
...
+func validateTLSSecurityProfileType(fieldPath *field.Path, profile *configv1.TLSSecurityProfile) field.ErrorList {
+	const typeProfileMismatchFmt = "type set to %s, but the corresponding field is unset"
+	typePath := fieldPath.Child("type")
+
+	errs := field.ErrorList{}
+
+	switch profile.Type {
+	case "":
+		if profile.Old != nil || profile.Intermediate != nil || profile.Modern != nil || profile.Custom != nil {
+			errs = append(errs, field.Required(typePath, "one of the profiles if set but 'type' field is empty"))
```
Miciah: "if" → "is".

PR #23983, opened 2019-10-17
UPSTREAM: 82439: Delete e2e namespaces asynchronously

smarterclayton:
This reduces the total runtime of tests by a few minutes and is a better
representation of our expectations around clusters.


smarterclayton:
/hold

Some interaction with our test runner and this cleanup is causing e2e-aws to fail.

smarterclayton:
One of the tests is double deleting the namespace.  The DELETE namespace call returns an error the second time you try to delete it, unlike every other API object we have.  I think that's a Kube API bug and should be fixed.  Will sync with Jordan and then back port that.

smarterclayton:
Upstream is https://github.com/kubernetes/kubernetes/pull/84122

smarterclayton:
/retest
/hold cancel

Picked the upstream error that was lost

PR #23992, opened 2019-10-20
Bug 1751820: test/extended/operators: ensure clusterversion current version is set

abhinavdahiya:
The check was dropped in 040f521 [1] during the rebase. reinstating the check correctly.

[1]: https://github.com/openshift/origin/commit/040f521d78039a06be7c5393bc6968285ae4ba4e

### DIFF
--- a/test/extended/operators/operators.go
+++ b/test/extended/operators/operators.go
@@ -58,6 +58,11 @@ var _ = g.Describe("[Feature:Platform][Smoke] Managed cluster should", func() {
 			cv := objx.Map(obj.UnstructuredContent())
 			lastErr = nil
 			lastCV = cv
+			payload := cv.Get("status.desired.payload").String()
+			if len(payload) == 0 {
+				e2e.Logf("ClusterVersion has no current payload version")
+				return false, nil
+			}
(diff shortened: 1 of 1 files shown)


wking:
History, because I was initially unclear on how this commit was fixing the issue:

* 5ff7aed012e2 (2018-11-09, #21459), added the check against `status.current.payload`, but it was a no-op gate because `.String()` returned a version or `<nil>`.
* openshift/api@cae4137f1a7f94f91916f2affc4a2e3f7c02ce31 (2019-01-04, openshift/api#153) changed `current` -> `desired` in the ClusterVersion status.  This should have caused the tests to fail to notice the CVO leveling, but because of the `<nil>` thing we didn't notice.
* The rebase in #23811 picked up a new objx, fixing `<nil>` -> empty string, but the tests started failing saying "I don't see the CVO ever setting `current`!".
* a07da12106f1007507ad51c395891371641e6baf (2019-09-11, #23811) dropped the check, because we don't really care about current/desired when judging CVO leveling.
* This PR's 2f21b747a119d1427111987255a508e4e26b086f restores the check, but fixes `status.current.payload` -> `status.desired.payload`, so now we're looking in the right place again and break through this gate when the CVO sets `desired`.

/lgtm

PR #23894, opened 2019-10-02
Bug 1755039: Enable storage tests that require 1.16 kubelet

jsafrane:
\+ remove one unused pattern
\+ reshuffle remaining disabled storage tests a bit and add BZ link to one
(see the individual commits)

@openshift/storage

PR #23999, opened 2019-10-22
Bug 1751640: Enable block bad-case tests

jsafrane:
The test was enabled in https://github.com/openshift/origin/pull/23921, but then accidentally disabled by https://github.com/openshift/origin/pull/23894 (github did not catch it as conflict).

@openshift/storage


jsafrane:
/test e2e-gcp
/assign @knobunc 
for approval

PR #24000, opened 2019-10-22
UPSTREAM: 83419: Insecure backend proxy

deads2k:
/hold

generated code needs to be regenerated and I don't know how right off.


deads2k:
@p0lyn0mial got it, thanks

/close

PR #24004, opened 2019-10-23
Bug 1759998: Bug 1690588: UPSTREAM: 84181: Lower AWS DescribeVolume frequency

jsafrane:
This replaces our temporary patch introduced in #23936 with proper upstream patch. This removes our extra logging of Attach/Detach.

https://github.com/kubernetes/kubernetes/pull/84181
https://bugzilla.redhat.com/show_bug.cgi?id=1759998


jsafrane:
cc @openshift/storage, PTAL

jsafrane:
I would let it soak in 4.3 for another week or two and then maybe we could fix also 4.2/4.1. WDYT?

jsafrane:
/bugzilla refresh
trying to fix to bugs with single PR :-)

PR #24010, opened 2019-10-23
reduce parallelism to avoid pushing as many concurrent writes

deads2k:
/assign @eparis


smarterclayton:
Which suites and tests are you trying to fix?

/hold

deads2k:
> Which suites and tests are you trying to fix?
> 
> /hold

Fix is relative. This was just a demonstrate for @eparis so he could see how he could reduce parallelism to plumb an option back for gcp

PR #24012, opened 2019-10-23
Bug 1765177: e2e: fix flaky route wait function

ironcladlou:
Many router tests use `waitForRouterOKResponseExec()` to wait for test route
connectivity. However, due to a bug in the check script, if the underlying
`curl` command times out once during checks, the `waitForRouterOKResponseExec()`
function will always time out even if the target route becomes responsive.

If the `curl` command itself never times out, the function works properly.

Fix the check script so that `curl` timeouts are correctly handled.

This probably fixes many e2e flakes associated with router tests which use
`waitForRouterOKResponseExec()`.

### DIFF
--- a/test/extended/router/scoped.go
+++ b/test/extended/router/scoped.go
@@ -235,6 +235,7 @@ func waitForRouterOKResponseExec(ns, execPodName, url, host string, timeoutSecon
 		set -e
 		STOP=$(($(date '+%%s') + %d))
 		while [ $(date '+%%s') -lt $STOP ]; do
+			rc=0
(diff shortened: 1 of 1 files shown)


knobunc:
/lgtm
/approve

This is great!  Thanks

ironcladlou:
As @Miciah said,

>It's not that it "will always time out"; more that it will continue to report the timeout error if subsequent curl commands succeed.

ironcladlou:
While cherry-picking I found some more of these, patched in https://github.com/openshift/origin/pull/24028. Will do a manual back port of the consolidated fix after https://github.com/openshift/origin/pull/24028 merges.

PR #24014, opened 2019-10-24
bumps to kubernetes 1.16.2

p0lyn0mial:


p0lyn0mial:
It seems like `ci/prow/e2e-aws` job is failing because `ClusterVersion has no current payload version`. This check was brought back in https://github.com/openshift/origin/pull/23992 and removed in https://github.com/openshift/origin/commit/040f521d78039a06be7c5393bc6968285ae4ba4e

It reads [`status.desired.payload`](https://github.com/openshift/origin/blob/master/test/extended/operators/operators.go#L61) but I can't find the definition of that field for [`config.openshift.io/v1/ClusterVersion`](https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L25) resource unless it reads a different resource?

p0lyn0mial:
> It seems like ci/prow/e2e-aws job is failing because ClusterVersion has no current payload version. This check was brought back in #23992 and removed in 040f521

> It reads status.desired.payload but I can't find the definition of that field for config.openshift.io/v1/ClusterVersion resource unless it reads a different resource?

fixed in https://github.com/openshift/origin/pull/24029

PR #24016, opened 2019-10-24
adds publishing rules for origin-4.3-kubernetes-1.16.2

p0lyn0mial:


p0lyn0mial:
I am leaving `origin-4.3-kubernetes-1.16.0-rc.2` rules for now (will remove them in a separate pull). They will also take `master` but that shouldn't be an issue.

publishing-rules.yaml
```diff
@@ -3,6 +3,10 @@ skip-gomod: true
 rules:
 - destination: kubernetes
   branches:
+  - name: origin-4.3-kubernetes-1.16.2
```
sttts: shouldn't we remove the old one at the same time?

p0lyn0mial: I wasn't sure if we want that https://github.com/openshift/origin/pull/24016#issuecomment-545872592. Will remove then.

p0lyn0mial: done

PR #24021, opened 2019-10-24
Bug 1753502: UPSTREAM: 82503: Reorder symlinks to prevent path escapes

sallyom:
/cc @soltysh


sallyom:
need this: https://github.com/openshift/origin/pull/24143  and then backport to 3.10 b4 this will be green

stevekuznetsov:
/hold

This has been re-tested so many times we've reached the limit of how many times we can report a failed message to GitHub. Please rebase.

PR #24022, opened 2019-10-24
Bug 1719686: Fix ceph expansion

gnufied:
Fixes  https://bugzilla.redhat.com/show_bug.cgi?id=1719686

This PR basically fixes 3 issues with ceph expansion:

1. upstream PR#72431 fixes problem with parsing of output by converting output to json. But this is not enough because `rbd` command seems to spew random warnings to stderr when ceph keyring file is missing 
2. upstream PR#75588 fixes silencing some of those warnings logged to stderr so as JSON output could be parsed correctly. But this is not enough because in Openshift we seem to have bundled rather old `ceph-common` package and not all warnings are silenced by `-k=/dev/null` option
3. The carry patch uses bit of heavy handed approach and skips using `mount.Exec` interface which combines stdout and stderr, in favour of `os/exec` so as we could just read `stdout`. This fixes ceph volume expansion. IMO - this should be relatively safe.

FWIW I have tested, reproduced and verified the fix. It works and I can't think of any negative consequence. 

/sig storage

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/volume/rbd/rbd_test.go
+++ b/vendor/k8s.io/kubernetes/pkg/volume/rbd/rbd_test.go
@@ -594,3 +594,28 @@ func TestConstructVolumeSpec(t *testing.T) {
 		}
 	}
 }
+
+func TestGetRbdImageSize(t *testing.T) {
+	for i, c := range []struct {
+		Output     string
+		TargetSize int
+	}{
+		{
+			Output:     `{"name":"kubernetes-dynamic-pvc-18e7a4d9-050d-11e9-b905-548998f3478f","size":10737418240,"objects":2560,"order":22,"object_size":4194304,"block_name_prefix":"rbd_data.9f4ff7238e1f29","format":2}`,
+			TargetSize: 10240,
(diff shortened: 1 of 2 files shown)


gnufied:
cc @jsafrane @childsb @bertinatto

gnufied:
Looks like openshift-ansible is failing for unrelated reason:

```
            {
                            "lastHeartbeatTime": "2019-10-25T11:26:01Z", 
                            "lastTransitionTime": "2019-10-25T11:22:45Z", 
                            "message": "runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialized", 
                            "reason": "KubeletNotReady", 
                            "status": "False", 
                            "type": "Ready"
                        }
```

/retest

petr-muller:
/hold

It really does not make sense to retest this automatically after 18 days of failures.

gnufied:
The flake is not in this PR. It is because of this - https://bugzilla.redhat.com/show_bug.cgi?id=1767579

let see if the flake is fixed now.

/hold cancel

PR #24028, opened 2019-10-25
e2e: fix more flaky route wait functions

ironcladlou:
Follow-up to 34297f29e11c14918bbc098cf06598755419f89b with newly discovered
instances where a return code variable is uninitialized leading to flaky
wait behavior.

### DIFF
--- a/test/extended/router/config_manager.go
+++ b/test/extended/router/config_manager.go
@@ -141,6 +141,7 @@ func waitForRouteToRespond(ns, execPodName, proto, host, abspath, ipaddr string,
 		set -e
 		STOP=$(($(date '+%%s') + %d))
 		while [ $(date '+%%s') -lt $STOP ]; do
+			rc=0
 			code=$( curl -k -s -m 5 -o /dev/null -w '%%{http_code}\n' --resolve %s:%d:%s %q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
 				echo $code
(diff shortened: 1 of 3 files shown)


test/extended/router/config_manager.go
```diff
@@ -141,6 +141,7 @@ func waitForRouteToRespond(ns, execPodName, proto, host, abspath, ipaddr string,
 		set -e
 		STOP=$(($(date '+%%s') + %d))
 		while [ $(date '+%%s') -lt $STOP ]; do
+			rc=0
 			code=$( curl -k -s -m 5 -o /dev/null -w '%%{http_code}\n' --resolve %s:%d:%s %q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
```
Miciah: It might make sense to replace `${rc:-0}` with `$rc` now that we know `rc` is always initialized, but using `${rc:-0}` doesn't actually hurt anything.

test/extended/router/headers.go
```diff
@@ -113,6 +113,7 @@ func dumpRouterHeadersLogs(oc *exutil.CLI, name string) {
 func getRoutePayloadExec(ns, execPodName, url, host string) (string, error) {
 	cmd := fmt.Sprintf(`
 		set -e
+		rc=0
```
Miciah: I was going to say that initializing `rc` is unnecessary because there is no loop, but it's probably a good idea anyway in case the parent process set `rc`.

PR #24023, opened 2019-10-25
test: Add additional debugging to flaky pull secret test

smarterclayton:

PR #24034, opened 2019-10-28
Bug 1765294: Use ginkgo in OCM extended tests

adambkaplan:
Improving logging of test failures


adambkaplan:
/close

https://github.com/openshift/origin/pull/24023 added more debug info, and increased timeout to 5 minutes

PR #24036, opened 2019-10-28
Upstream 84466: gce: ensureInternalInstanceGroups: reuse instance-groups for internal load balancers

abhinavdahiya:
upstream PR: https://github.com/kubernetes/kubernetes/pull/84466

This is required for internal cluster support in GCP, otherwise we can't use internal LB for API and also have other internal LB service type load balancer because control-plane is also part of the LB backend for all such k8s services.


abhinavdahiya:
/cc @jstuever 

installer PR that's blocked: https://github.com/openshift/installer/pull/2522

smarterclayton:
I'm going to grant approval to carry this even if upstream drags their
heels.  If I can get a rough agreement from bowei today i'll merge in the
morning.

smarterclayton:
Discussion upstream has clarified the use case, I think this is ok to merge now as long as any feedback gets addressed as a PR post merge.  Merging now enables us to test with this and expose any weakness.

/lgtm

PR #24053, opened 2019-10-30
e2e requestheader: make client follow TLS correctly

stlaz:
Don't force our client certificates on the server, instead respect
the selection of CAs the server sends us to pick the correct trusted
client cert.

This should be breaking tests until https://github.com/openshift/oauth-server/pull/19 get merged


stlaz:
/test e2e-aws-serial
oauth-server PR merged, this should now work

PR #24047, opened 2019-10-29
UPSTREAM: 80004: Prefer to delete doubled-up pods of a ReplicaSet

p0lyn0mial:
brings back https://github.com/openshift/origin/pull/23806

PR #24057, opened 2019-10-30
Bug 1765756: UPSTREAM: 80004: Prefer to delete doubled-up pods of a ReplicaSet

Miciah:
This commit brings back https://github.com/openshift/origin/pull/23806, which was dropped during a rebase scuffle (see https://github.com/openshift/origin/pull/24014#commitcomment-35709732), and adds a fix for [BZ#1765756](https://bugzilla.redhat.com/show_bug.cgi?id=1765756).

---

#### UPSTREAM: 80004: Prefer to delete doubled-up pods of a ReplicaSet

When scaling down a ReplicaSet, delete doubled up replicas first, where a "doubled up replica" is defined as one that is on the same node as an active replica belonging to a related ReplicaSet.  ReplicaSets are considered "related" if they have a common controller (typically a Deployment).

The intention of this change is to make a rolling update of a Deployment scale down the old ReplicaSet as it scales up the new ReplicaSet by deleting pods from the old ReplicaSet that are colocated with ready pods of the new ReplicaSet.  This change in the behavior of rolling updates can be combined with pod affinity rules to preserve the locality of a Deployment's pods over rollout.

A specific scenario that benefits from this change is when a Deployment's pods are exposed by a Service that has type "LoadBalancer" and external traffic policy "Local".  In this scenario, the load balancer uses health checks to determine whether it should forward traffic for the Service to a particular node.  If the node has no local endpoints for the Service, the health check will fail for that node.  Eventually, the load balancer will stop forwarding traffic to that node.  In the meantime, the service proxy drops traffic for that Service.  Thus, in order to reduce risk of dropping traffic during a rolling update, it is desirable preserve node locality of endpoints.

* `vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go` (`ActivePodsWithRanks`): New type to sort pods using a given ranking.
* `vendor/k8s.io/kubernetes/pkg/controller/controller_utils_test.go` (`TestSortingActivePodsWithRanks`): New test for `ActivePodsWithRanks`.
* `vendor/k8s.io/kubernetes/pkg/controller/replicaset/replica_set.go`
(`getReplicaSetsWithSameController`): New method.  Given a ReplicaSet, return all ReplicaSets that have the same owner.
(`manageReplicas`): Call `getIndirectlyRelatedPods`, and pass its result to `getPodsToDelete`.
(`getIndirectlyRelatedPods`): New method.  Given a ReplicaSet, return all pods that are owned by any ReplicaSet with the same owner.
(`getPodsToDelete`): Add an argument for related pods.  Use related pods and the new getPodsRankedByRelatedPodsOnSameNode function to take into account whether the pod is doubled up when sorting pods for deletion.
(`getPodsRankedByRelatedPodsOnSameNode`): New function.  Return an `ActivePodsWithRanks` value that wraps the given slice of pods and computes ranks where each pod's rank is equal to the number of active related pods that are colocated on the same node.
* `vendor/k8s.io/kubernetes/pkg/controller/replicaset/replica_set_test.go` (`newReplicaSet`): Set `OwnerReferences` on the ReplicaSet.
(`newPod`): Set a unique UID on the pod.
(`byName`): New type to sort pods by name.
(`TestRelatedPodsLookup`): New test for `getIndirectlyRelatedPods`.
(`TestGetPodsToDelete`): Augment the "various pod phases and conditions, diff = len(pods)" test case to ensure that scale-down still selects doubled-up pods if there are not enough other pods to scale down.  Add a "various pod phases and conditions, diff = len(pods), relatedPods empty" test case to verify that getPodsToDelete works even if related pods could not be determined.  Add a "ready and colocated with another ready pod vs not colocated, diff < len(pods)" test case to verify that a doubled-up pod gets preferred for deletion.  Augment the "various pod phases and conditions, diff < len(pods)" test case to ensure that not-ready pods are preferred over ready but doubled-up pods.
* `vendor/k8s.io/kubernetes/pkg/controller/replicaset/BUILD:` Regenerate.
* `vendor/k8s.io/kubernetes/test/e2e/apps/deployment.go` (`testRollingUpdateDeploymentWithLocalTrafficLoadBalancer`): New end-to-end test.  Create a deployment with a rolling update strategy and affinity rules and a load balancer with "Local" external traffic policy, and verify that set of nodes with local endponts for the service remains unchanged during rollouts.
(`setAffinity`): New helper, used by `testRollingUpdateDeploymentWithLocalTrafficLoadBalancer`.
* `vendor/k8s.io/kubernetes/test/e2e/apps/types.go` (`AgnhostImageName`, `AgnhostImage`): New constants for the agnhost image.
* `vendor/k8s.io/kubernetes/test/e2e/framework/service/jig.go` (`GetEndpointNodes`): Factor building the set of node names out...
(`GetEndpointNodeNames`): ...into this new method.

#### UPSTREAM: 84568: test/e2e/apps: Skip or scale LB test per node count

Skip the "Deployment should not disrupt a cloud load-balancer's connectivity during rollout" test if the number of nodes is less than 2; otherwise, set the deployment's replicas equal to the lesser of 5 and the number of nodes.

The test would fail if there were fewer nodes than replicas, but the test needs at least 2 nodes, and the likelihood of failure absent the feature under test increases with the number of replicas, so it is desirable to set replicas to a higher value, within reason.

* `vendor/k8s.io/kubernetes/test/e2e/apps/deployment.go`: Skip the load-balancer connectivity test unless there are at least 2 nodes.
(`testRollingUpdateDeploymentWithLocalTrafficLoadBalancer`): Set `replicas` to the min of 5 and the number of nodes.

---

This PR supersedes #24047.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go
+++ b/vendor/k8s.io/kubernetes/pkg/controller/controller_utils.go
@@ -771,6 +771,94 @@ func (s ActivePods) Less(i, j int) bool {
 	return false
 }
 
+// ActivePodsWithRanks is a sortable list of pods and a list of corresponding
+// ranks which will be considered during sorting.  The two lists must have equal
+// length.  After sorting, the pods will be ordered as follows, applying each
+// rule in turn until one matches:
+//
+// 1. If only one of the pods is assigned to a node, the pod that is not
+//    assigned comes before the pod that is.
+// 2. If the pods' phases differ, a pending pod comes before a pod whose phase
+//    is unknown, and a pod whose phase is unknown comes before a running pod.
+// 3. If exactly one of the pods is ready, the pod that is not ready comes
+//    before the ready pod.
+// 4. If the pods' ranks differ, the pod with greater rank comes before the pod
+//    with lower rank.
+// 5. If both pods are ready but have not been ready for the same amount of
+//    time, the pod that has been ready for a shorter amount of time comes
+//    before the pod that has been ready for longer.
+// 6. If one pod has a container that has restarted more than any container in
+//    the other pod, the pod with the container with more restarts comes
+//    before the other pod.
+// 7. If the pods' creation times differ, the pod that was created more recently
+//    comes before the older pod.
+//
+// If none of these rules matches, the second pod comes before the first pod.
+//
+// The intention of this ordering is to put pods that should be preferred for
+// deletion first in the list.
+type ActivePodsWithRanks struct {
+	// Pods is a list of pods.
+	Pods []*v1.Pod
+
+	// Rank is a ranking of pods.  This ranking is used during sorting when
+	// comparing two pods that are both scheduled, in the same phase, and
+	// having the same ready status.
+	Rank []int
+}
+
+func (s ActivePodsWithRanks) Len() int {
+	return len(s.Pods)
+}
+
+func (s ActivePodsWithRanks) Swap(i, j int) {
+	s.Pods[i], s.Pods[j] = s.Pods[j], s.Pods[i]
+	s.Rank[i], s.Rank[j] = s.Rank[j], s.Rank[i]
+}
+
+// Less compares two pods with corresponding ranks and returns true if the first
+// one should be preferred for deletion.
+func (s ActivePodsWithRanks) Less(i, j int) bool {
+	// 1. Unassigned < assigned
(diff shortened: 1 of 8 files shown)


Miciah:
I wonder whether I can do this:
/test e2e-aws-scaleup-rhel7-4.3

PR #23614, opened 2019-08-15
Bug 1703885: test/extended/router: Fix curl timeout, log dumping, template parameters, and router readiness probes

Miciah:
#### `test/extended/router`: Fix curl timeout logic

Fix `waitForRouteToRespond`, `waitForRouterOKResponseExec`, and `expectRouteStatusCodeRepeatedExec` to honor their timeout parameter values and to specify a timeout for each curl invocation.

Before this change, each function used its timeout parameter value as the number of times to retry the curl command and did not specify a timeout value to that command.  As a result, each command invocation could keep trying indefinitely, and the function could keep retrying the command for far longer than the timeout period that the caller indicated.

Follow-up to https://github.com/openshift/origin/pull/12346 and https://github.com/openshift/origin/pull/19073.

* `test/extended/router/config_manager.go` (`waitForRouteToRespond`):
* `test/extended/router/scoped.go` (`waitForRouterOKResponseExec`)
(`expectRouteStatusCodeRepeatedExec`): Keep retrying until the amount of time that is indicated by the timeout value has elapsed.  Specify a timeout value of 5 seconds to curl using the `-m` option.

#### `test/extended/router/unprivileged`: Fix log dump

Fix the prefix for the pod name used in the unprivileged router test so that the pod's logs will be dumped on test failures.

Follow-up to https://github.com/openshift/origin/pull/21557.

* `test/extended/router/unprivileged.go`: Call `DumpPodLogsStartingWith` with the "router-" prefix instead of the "unprivileged-router" prefix.

#### `test/extended/router`: Delete `SCOPE` template param

Delete the `SCOPE` template parameter in the various router template fixtures, and add `ROUTER_NAME` (default value "test-scoped") and `UPDATE_STATUS` (default value "true") parameters to the `router-scoped.yaml` template fixture.

Most tests were not using the `SCOPE` parameter, and in the one test that did, its use evoked the following warning from `oc new-app`:

    warning: --param no longer accepts comma-separated lists of values.

* `test/extended/router/unprivileged.go`:
* `test/extended/testdata/router/router-scoped.yaml`: Replace the `SCOPE` template parameter with `ROUTER_NAME` and `UPDATE_STATUS` parameters.
* `test/extended/testdata/router/router-override-domains.yaml`:
* `test/extended/testdata/router/router-override.yaml`: Delete the `SCOPE` parameter.
* `test/extended/testdata/bindata.go`: Regenerate.

#### `test/extended/router`: Enable metrics, fix readiness

[...]

PR #24064, opened 2019-10-31
[release-4.1] Bug 1765181: e2e: fix flaky route wait function

ironcladlou:
Many router tests use waitForRouterOKResponseExec() to wait for test route
connectivity. However, due to a bug in the check script, if the underlying
curl command times out once during checks, the waitForRouterOKResponseExec()
function will always time out even if the target route becomes responsive.

If the curl command itself never times out, the function works properly.

Fix the check script so that curl timeouts are correctly handled.

This probably fixes many e2e flakes associated with router tests which use
waitForRouterOKResponseExec().

### DIFF
--- a/test/extended/router/config_manager.go
+++ b/test/extended/router/config_manager.go
@@ -140,6 +140,7 @@ func waitForRouteToRespond(ns, execPodName, proto, host, abspath, ipaddr string,
 	cmd := fmt.Sprintf(`
 		set -e
 		for i in $(seq 1 %d); do
+			rc=0
 			code=$( curl -k -s -o /dev/null -w '%%{http_code}\n' --resolve %s:%d:%s %q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
 				echo $code
--- a/test/extended/router/scoped.go
+++ b/test/extended/router/scoped.go
@@ -242,6 +242,7 @@ func waitForRouterOKResponseExec(ns, execPodName, url, host string, timeoutSecon
 	cmd := fmt.Sprintf(`
 		set -e
 		for i in $(seq 1 %d); do
+			rc=0
 			code=$( curl -k -s -o /dev/null -w '%%{http_code}\n' --header 'Host: %s' %q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
 				echo $code
@@ -272,6 +273,7 @@ func expectRouteStatusCodeRepeatedExec(ns, execPodName, url, host string, status
 	cmd := fmt.Sprintf(`
 		set -e
 		for i in $(seq 1 %d); do
+			rc=0
 			code=$( curl -s -o /dev/null -w '%%{http_code}\n' --header 'Host: %s' %q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
 				echo $code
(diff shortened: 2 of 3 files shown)


test/extended/router/scoped.go
```diff
@@ -242,6 +242,7 @@ func waitForRouterOKResponseExec(ns, execPodName, url, host string, timeoutSecon
 	cmd := fmt.Sprintf(`
 		set -e
 		for i in $(seq 1 %d); do
+			rc=0
```
Miciah: Would it be worth backporting https://github.com/openshift/origin/pull/23614/commits/743b473afe412ae14061e3ee741c472962df1fe7 as well in order to keep the code in synch?

ironcladlou: I think any test stability back ports would be welcome

Miciah: https://github.com/openshift/origin/commit/743b473afe412ae14061e3ee741c472962df1fe7 is older, so it would make sense to backport it first.  Would you like a separate PR, or will you incorporate it into this PR?

Miciah: (Presumably we'd want all the commits in https://github.com/openshift/origin/pull/23614.)

ashcrow:
This hasn't merged because:
- unit testing failed (looks to be non-blocking?)
- No lgtm

If unittesting is expected to fail and someone LGTM I'm happy to approve this :-)

ashcrow:
Since this is specifically test fixes I believe BZ's `POST` to be sufficient.

ashcrow:
Once I have the power once more I'll approve :smile:

ironcladlou:
unit job looks like it was being OOMKilled — bumped the job memory limit in https://github.com/openshift/release/pull/5986 to see if it helps


/retest

ashcrow:
New changes and test failures. Removing `cherry-pick-approved` for now.

ironcladlou:
>New changes and test failures. Removing cherry-pick-approved for now.

The change was a rebase (not to the patch) and the failures are something systemic which I'm surprised more people aren't concerned about... I'm going to look a little more closely. Check out https://github.com/openshift/release/pull/5986. Something's wrong.

ashcrow:
Re-added label

ashcrow:
Removing `cherry-pick-approved` for now. The next patch manager should re-review.

test/extended/router/config_manager.go
```diff
@@ -140,6 +140,7 @@ func waitForRouteToRespond(ns, execPodName, proto, host, abspath, ipaddr string,
 	cmd := fmt.Sprintf(`
 		set -e
 		for i in $(seq 1 %d); do
+			rc=0
 			code=$( curl -k -s -o /dev/null -w '%%{http_code}\n' --resolve %s:%d:%s %q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
```
frobware: nit: If we are defaulting the value this can just be `$rc`.

frobware: Can this not be simplified to:

```foo.sh
#!/bin/bash

if ! code=$(curl -s -o /dev/null -w '%{http_code}\n' "$@"); then
    echo "curl failed"
    exit 1
fi

echo $code
```

```console
$ bash foo.sh http://google.comx
curl failed

$ bash foo.sh http://google.com/
301

$ bash foo.sh -L https://google.com/
200

$ bash foo.sh -L https://google.com/404
404
```

ironcladlou:
@Miciah @knobunc this got lost in the shuffle, might as well merge it for CI's sake

Miciah:
Should we backport #23614 first to maintain chronological ordering of related cleanups?

ironcladlou:
Go for it

PR #24066, opened 2019-10-31
Bug 1772467: UPSTREAM: 84612: Set TypeSetter in get print flags, like we do everywhere else

soltysh:
/assign @damemi @sallyom


damemi:
/lgtm cancel
to end the bot spam

PR #24067, opened 2019-10-31
test: must-gather pods may tolerate all taints

smarterclayton:
Because the test runs when other tests run, it can sometimes fail
if must gather happens to be running. It's ok for must-gather pods
to tolerate all taints.


smarterclayton:
unit has passed before, this is just a flake

/override ci/prow/unit

PR #24070, opened 2019-11-01
Add Ben as top-level OWNER

jsafrane:
He was in vendor/OWNERS that got accidentally removed in #24014. Since it's common to remove files in vendor/ during rebases, adding him to top-level OWNERS is more stable.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -10,6 +10,7 @@ approvers:
   - deads2k
   - derekwaynecarr
   - eparis
+  - knobunc  # Network, Multi-cluster, Storage
   - mfojtik
   - pweil-
   - sjenning


jsafrane:
I removed @childsb :-(. Storage does not have a new teamlead yet.

mfojtik:
/approve

Storage team need to deliver fixes from upstream and deserve to have top-level approver.
This repository is not what it used to be before split, and they should not place OWNERS into `vendor` directory that is managed by glide and it will be by `go.mod` in future.

jsafrane:
@knobunc FYI, you're top level approver now. Enjoy our endless requests for approval :-)

PR #24072, opened 2019-11-01
apiserver crd validation: remove tls1.3 and modern profile

stlaz:
I made some wrong assumptions about turning TLS1.3 on for 1.12 golang binaries...

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validate_apiserver.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validate_apiserver.go
@@ -1,7 +1,6 @@
 package apiserver
 
 import (
-	"crypto/tls"
 	"fmt"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


stlaz:
@openshift/sig-network-edge please take a look

stlaz:
cc @ironcladlou @Miciah

sttts:
/hold 

until https://github.com/openshift/api/pull/508/files is merged.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/apiserver/validate_apiserver.go
```diff
@@ -219,10 +216,7 @@ func validateCipherSuites(fieldPath *field.Path, suites []string) field.ErrorLis
 
 	if len(suites) > 0 { // something was configured
```
Miciah: Isn't it invalid for `suites` to be empty?

stlaz: Hm, the field does not have `+required` so I was just falling back to intermediate profile ciphers in this case, but the behavior of the field being empty is really not documented. 

I guess forcing it makes sense as any automatic behavior may be counterintuitive.

PR #24075, opened 2019-11-01
Bug 1683057: Fix router metrics test flakes

ironcladlou:
The router metrics tests assume they're talking to a single router pod. During
the v4 refactor, this assumption broke, and the test connection to the router
was load-balanced to router pods behind the router service. This made the test
flaky.

Refactor the test to discover and operate against a single ingress endpoint.

### DIFF
--- a/test/extended/router/metrics.go
+++ b/test/extended/router/metrics.go
@@ -16,15 +16,16 @@ import (
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 
+	clientset "k8s.io/client-go/kubernetes"
+	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/kubernetes/pkg/client/conditions"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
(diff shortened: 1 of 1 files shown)


ironcladlou:
>LGTM, though I think doing _ := foo() was clearer without the assignment. Either way, it is ignored.

Oops... my editor's linter auto-refactored these! Wonder if the std linter likes one or the other better.

And yeah, I guess I should actually do something with the error anyway. Too bad the editor couldn't do that for me too...

ironcladlou:
@frobware rolled back the unnecessary refactor, PTAL

frobware:
> @frobware rolled back the unnecessary refactor, PTAL

/lgtm

smarterclayton:
I think this substantially increased flakes in AWS

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_oc/143/pull-ci-openshift-oc-master-e2e-aws/478

Going to revert, then we can retest in e2e-aws premerge

ironcladlou:
https://github.com/openshift/origin/pull/24085 should fix the AWS flake

PR #24076, opened 2019-11-01
add ci image to launch gce instance from openshift

sallyom:
This image plus scripts allow a gcp instance to launch from OpenShift CI pod, with a 'real' user, to be able to ssh to the gcp instance to run commands. 

We use a similar setup to launch a libvirt cluster nested in a gcp instance in CI. 

This PR is necessary for https://github.com/openshift/release/pull/5741
/cc @mrunalp


sallyom:
closing, will add this image to CI rather than origin, if needed.

PR #24085, opened 2019-11-04
e2e: fix router metrics test flake on AWS

ironcladlou:
On AWS, the default router speaks PROXY protocol. The fix in
https://github.com/openshift/origin/pull/24075 switched some router tests to
(correctly) speak to the router directly. However, the fix did not update client
code to speak PROXY to the router on AWS. The tests still sometimes pass on AWS
by coincidence (as other non-test clients send traffic to the router through the
LB, causing router stats to sometimes match test expectations.)

Fix the tests so that test clients talking to routers use PROXY protocol on AWS.

### DIFF
--- a/test/extended/router/scoped.go
+++ b/test/extended/router/scoped.go
@@ -262,13 +262,19 @@ func waitForRouterOKResponseExec(ns, execPodName, url, host string, timeoutSecon
 	return nil
 }
 
-func expectRouteStatusCodeRepeatedExec(ns, execPodName, url, host string, statusCode int, times int) error {
+func expectRouteStatusCodeRepeatedExec(ns, execPodName, url, host string, statusCode int, times int, proxy bool) error {
+	var extraArgs []string
+	if proxy {
+		extraArgs = append(extraArgs, "--haproxy-protocol")
+	}
+	args := strings.Join(extraArgs, " ")
+
 	cmd := fmt.Sprintf(`
 		set -e
 		STOP=$(($(date '+%%s') + %d))
 		while [ $(date '+%%s') -lt $STOP ]; do
 			rc=0
-			code=$( curl -s -m 5 -o /dev/null -w '%%{http_code}\n' --header 'Host: %s' %q ) || rc=$?
+			code=$( curl %s -s -m 5 -o /dev/null -w '%%{http_code}\n' --header 'Host: %s' %q ) || rc=$?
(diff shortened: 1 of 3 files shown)


ironcladlou:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24085/pull-ci-openshift-origin-master-e2e-gcp/3548

```
Flaky tests:

[Feature:Prometheus][Conformance] Prometheus when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog [Suite:openshift/conformance/parallel/minimal]
[sig-scheduling] Multi-AZ Clusters should spread the pods of a service across zones [Suite:openshift/conformance/parallel] [Suite:k8s]

Failing tests:

[Feature:Builds][Conformance] oc new-app  should succeed with a --name of 58 characters [Suite:openshift/conformance/parallel/minimal]
```

test/extended/router/scoped.go
```diff
@@ -262,13 +262,19 @@ func waitForRouterOKResponseExec(ns, execPodName, url, host string, timeoutSecon
 	return nil
 }
 
-func expectRouteStatusCodeRepeatedExec(ns, execPodName, url, host string, statusCode int, times int) error {
+func expectRouteStatusCodeRepeatedExec(ns, execPodName, url, host string, statusCode int, times int, proxy bool) error {
+	var extraArgs []string
+	if proxy {
+		extraArgs = append(extraArgs, "--haproxy-protocol")
```
squeed: This argument isn't in the version of `curl` bundled in UBI...

smarterclayton:
/retest

How did AWS pass then?

test/extended/router/scoped.go
ironcladlou: It's so old though... https://github.com/curl/curl/commit/6baeb6df35d24740c55239f24b5fc4ce86f375a5

In my dev clusters, the image that gets used is `gcr.io/kubernetes-e2e-test-images/agnhost:2.6` which has curl 7.61.1 (x86_64-alpine-linux-musl) libcurl/7.61.1 LibreSSL/2.5.5 zlib/1.2.11 libssh2/1.8.2

ironcladlou: https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/test/e2e/framework/pod/resource.go#L475

smarterclayton:
Force merging to unblock queues

PR #24088, opened 2019-11-05
UPSTREAM: 84123: Handle namespace shutdown gracefully

smarterclayton:
Several controllers spam the apiserver during namespace deletion. Make
it easier for clients to detect namespace deletion and stop trying
when a namespace is deleting.


smarterclayton:
@deads2k @sttts 

It looks like this reduces total APIServer requests per second from ~450/s during a normal GCP e2e 4.3 run to ~380/s.  Cuts kube-apiserver CPU during e2e run by 5-15% roughly.

smarterclayton:
And I'll open openshift-controller-manager changes after this merges.

smarterclayton:
@soltysh as well (since these are your controllers in many cases)

PR #24093, opened 2019-11-05
Bug 1752646: e2e: Re-enable HAProxy router header test

frobware:
This change loosens the test expectations for AWS only.

On AWS the packet (GET request) leaves the cluster and will be NAT'd;
source addresses will now reflect the elastic IP address of the node.
Matching on the PODs IP address is therefore not possible on AWS; we
now assert that we have an 'X-Forwarded-For' header only.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1752646

### DIFF
--- a/test/extended/router/headers.go
+++ b/test/extended/router/headers.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 
+	configv1 "github.com/openshift/api/config/v1"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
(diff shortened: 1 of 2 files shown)


ironcladlou:
Very nice, thanks!

@Miciah PTAL and tag if ready?

/approve

ironcladlou:
/retitle Bug 1752646: e2e: Re-enable HAProxy router header test

ironcladlou:
Hmm..

/retest

mandre:
We'll have to handle the OpenStack platform as a special case too as it's failing there. Is it the same fix as for AWS platform?

https://storage.googleapis.com/origin-ci-test/logs/release-openshift-ocp-installer-e2e-openstack-4.3/400/build-log.txt

Issue #24068, opened 2019-11-01
[Feature:ImageExtract] Image extract should extract content from an image [Suite:openshift/conformance/parallel] fails on ppc64le platform


mkumatag:
[provide a description of the issue]

##### Version
```
[root@localhost origin]# oc version
Client Version: version.Info{Major:"4", Minor:"2+", GitVersion:"v4.2.0", GitCommit:"c569285e9", GitTreeState:"clean", BuildDate:"2019-07-31T11:06:03Z", GoVersion:"go1.12.6", Compiler:"gc", Platform:"linux/ppc64le"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6+2e5ed54", GitCommit:"2e5ed54", GitTreeState:"clean", BuildDate:"2019-10-16T20:38:49Z", GoVersion:"go1.12.8", Compiler:"gc", Platform:"linux/ppc64le"}
OpenShift Version: 0.0.1-2019-10-22-203947
[root@localhost origin]#
```
##### Steps To Reproduce
1. Build openshift-tests
2. Run `_output/local/bin/linux/ppc64le/openshift-tests run-test "[Feature:ImageExtract] Image extract should extract content from an image [Suite:openshift/conformance/parallel]"`

##### Current Result
```
---->
+ oc image extract --insecure image-registry.openshift-image-registry.svc:5000/e2e-test-image-extract-xwvzg/1:busybox --path=/:/tmp/doesnotexist
error: destination path does not exist: /tmp/doesnotexist
+ oc image extract --insecure image-registry.openshift-image-registry.svc:5000/e2e-test-image-extract-xwvzg/1:busybox --path=/:/
error: directory / must be empty, pass --confirm to overwrite contents of directory
+ mkdir -p /tmp/test
+ oc image extract --insecure image-registry.openshift-image-registry.svc:5000/e2e-test-image-extract-xwvzg/1:busybox --path=/:/tmp/test
+ '[' -d /tmp/test/etc ']'
+ '[' -d /tmp/test/bin ']'
+ '[' -f /tmp/test/bin/ls ']'
+ /tmp/test/bin/ls /tmp/test
/tmp/test/bin/ls: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/test/bin/ls)
/tmp/test/bin/ls: /lib64/libc.so.6: version `GLIBC_2.27' not found (required by /tmp/test/bin/ls)
<----end of log for "append-test"/"test"
[...]

PR #24095, opened 2019-11-05
Remove command execution from the extended image

mkumatag:
Fixes: https://github.com/openshift/origin/issues/24068

### DIFF
--- a/test/extended/images/extract.go
+++ b/test/extended/images/extract.go
@@ -76,7 +76,7 @@ var _ = g.Describe("[Feature:ImageExtract] Image extract", func() {
 			mkdir -p /tmp/test
 			oc image extract --insecure %[2]s/%[1]s/1:busybox --path=/:/tmp/test
 			[ -d /tmp/test/etc ] && [ -d /tmp/test/bin ]
-			[ -f /tmp/test/bin/ls ] && /tmp/test/bin/ls /tmp/test
+			[ -f /tmp/test/etc/passwd ] && grep root /tmp/test/etc/passwd
 
 			# extract multiple individual files
 			mkdir -p /tmp/test2


test/extended/images/extract.go
```diff
@@ -76,7 +76,7 @@ var _ = g.Describe("[Feature:ImageExtract] Image extract", func() {
 			mkdir -p /tmp/test
 			oc image extract --insecure %[2]s/%[1]s/1:busybox --path=/:/tmp/test
 			[ -d /tmp/test/etc ] && [ -d /tmp/test/bin ]
-			[ -f /tmp/test/bin/ls ] && /tmp/test/bin/ls /tmp/test
+			[ -f /tmp/test/bin/ls ]
```
bparees: hmm... is there something else we can do to confirm the content was extracted successfully?  perhaps extract a text file and confirm the content?  i'd prefer to have some assurance we are not corrupting or truncating the content on extraction.

mkumatag: I'm not really sure how to that! here we are just importing files from baseimage and checking the content! I'm wondering is there a way where we bundle something in the baseimage and extract and check!

mkumatag: How about grepping `root` in the `/etc/passwd` file?

bparees: i was hoping there might be something already in the busybox image you could check, but i guess there's not much in there.  can you at least confirm the extracted file is non-zero sized?

bparees: > How about grepping root in the /etc/passwd file?

sounds good to me.

mkumatag: Pushed the changes, hope it works fine 🤞

PR #24098, opened 2019-11-06
enables encryption during extended e2e tests

p0lyn0mial:


p0lyn0mial:
After enabling encryption the tests fail due to network errors (`GOAWAY`, `EOF`, `connection refused`). I think this is because handling `HTTP2` connections in a graceful way don't work. 
Enabling encryption rolls out 3 new revisions causing servers to restart. During that time servers forcefully close connections causing clients to fail.

In order to prove that I did some further tests, in particular, the following:

1. I turned off all API servers except one.
2. I wrote a simple client that created new namespaces at a controlled pace - 5 requests per second.
3. Then I had first run the client and then I sshed into the node with the running API server and told the API server to start the graceful shutdown process (I sent `SIGTERM` signal). In that state, all new requests should be rejected and those already in flight have 60 seconds to finish (after that time they will be forcefully closed).
4. Finally, I returned to the terminal with the running client and observed that it kept creating new namespaces (no errors). The new requests were not rejected by the server until the very last moment when the server started forcefully close opened connections (60 seconds after sending `SIGTERM`) causing network errors (`GOAWAY`, `EOF`, `connection refused`).

Further questions/observations:

1. How many connections do e2e tests create? Do all tests use the same client?
2. Load balancing probably doesn't work because a client maintains a single connection to the server (HTTP2).

PR #24100, opened 2019-11-06
React to openshift/ruby-hello-world update

adambkaplan:
Update tests to use ruby-25 image.

/assign @bparees 
fyi @openshift/openshift-team-developer-experience


adambkaplan:
Looks like there were a few tests that were not disabled in the ruby:2.2 deprecation.

adambkaplan:
/hold

`cmd` fails because the pod it runs in does not have git installed :/

adambkaplan:
self-tagging

adambkaplan:
/retitle Bug 1769949: React to openshift/ruby-hello-world update

PR #24103, opened 2019-11-07
Bug 1765294:  Changing from passive to active approach on secret deletion.

ricardomaraschini:
The passive way of waiting for events to come in through a Watch may be
causing some problems with this specific test. This PR converts it into
an active polling approach.

### DIFF
--- a/test/extended/controller_manager/pull_secret.go
+++ b/test/extended/controller_manager/pull_secret.go
@@ -8,8 +8,9 @@ import (
 	o "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


ricardomaraschini:
/retitle Bug 1765294:  Changing from passive to active approach on secret deletion.

test/extended/controller_manager/pull_secret.go
```diff
@@ -175,26 +168,23 @@ var _ = g.Describe("[Feature:OpenShiftControllerManager]", func() {
...
+			select {
+			case <-timeout.C:
+				t.Fatalf("timeout waiting for secret deletion")
+			case <-ticker.C:
+				if _, err := clusterAdminKubeClient.CoreV1().Secrets(sa.Namespace).Get(
+					dockercfgSecretName,
+					metav1.GetOptions{},
+				); errors.IsNotFound(err) {
+					return
+				}
```
adambkaplan: nit: use `wait.Poll` from `k8s.io/apimachinery/pkg/util/wait`. It effectively does the same thing, returning a specific timeout error if it is exceeded.

See https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L65-L189 as an example.

ricardomaraschini: Done, PTAL.

PR #24104, opened 2019-11-07
Bug 1768002: Stabilize router stress test

ironcladlou:
The stress.go test makes some brittle assumptions about the number of writes to
Route resources in a namespace. The assumptions used to hold true on v3
clusters, but in v4 there are other routers which break the assumptions.

For now, relax the assumptions. In the future, this test should probably be
refactored to not rely on specific write counts.


ironcladlou:
Day 6...

/retest

PR #24107, opened 2019-11-07
Add tests for image references after oc tag

dmage:
These tests ensure that we won't regress like in the [bug 1757889](https://bugzilla.redhat.com/show_bug.cgi?id=1757889).

### DIFF
--- a/test/extended/images/oc_tag.go
+++ b/test/extended/images/oc_tag.go
@@ -1,0 +1,115 @@
+package images
+
+import (
+	"fmt"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[Feature:Image] oc tag", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("image-oc-tag", exutil.KubeConfigPath())
+
+	g.It("should preserve image reference for external images", func() {
+		const (
+			externalRepository = "busybox"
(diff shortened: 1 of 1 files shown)


bparees:
it seemed like the issue in the bug occurred when tagging between namespaces, not just tagging one imagestreamtag to a new name within a single namespace.  I'm not sure why that should be different, but i think the test should probably cover that scenario.

dmage:
@bparees yes, the original issue is about two namespaces, but I wanted to make the test simpler so I used only one namespace. Would you prefer to have tests for both cases (copying within the namespace and across namespaces)?

bparees:
> Would you prefer to have tests for both cases (copying within the namespace and across namespaces)?

i think that would be best.  Thanks for creating this!

test/extended/images/oc_tag.go
```diff
@@ -0,0 +1,115 @@
...
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[Feature:Image] oc tag", func() {
```
smarterclayton: Do we not already test something like this via the cmd tests?  I thought we did.

dmage: We did, but some of them are disabled because we don't have the registry for cmd tests in 4.x:

https://github.com/openshift/origin/blob/496eb7cd1b389529b55d2c154421b3be21e93db2/test/extended/testdata/cmd/test/cmd/images.sh#L40

Eventually we are going to get rid of cmd tests (https://jira.coreos.com/browse/DEVEXP-238).

dmage:
up

PR #24117, opened 2019-11-08

smarterclayton:

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -33,7 +33,9 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
-const waitForPrometheusStartSeconds = 240
+const (
+	maxPrometheusQueryRetries = 5
+)
 
 var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
 	defer g.GinkgoRecover()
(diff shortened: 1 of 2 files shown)


test/extended/prometheus/prometheus.go
```diff
@@ -173,6 +173,29 @@ var _ = g.Describe("[Feature:Prometheus][Conformance] Prometheus", func() {
...
+			execPod := exutil.CreateCentosExecPodOrFail(oc.AdminKubeClient(), ns, "execpod", nil)
+			defer func() { oc.AdminKubeClient().CoreV1().Pods(ns).Delete(execPod.Name, metav1.NewDeleteOptions(1)) }()
+
+			tests := map[string]bool{
+				// track infrastructure type
+				`cluster_infrastructure_provider{type!=""}`: true,
+				`cluster_feature_set`:                       true,
+
+				// track installer type
+				`cluster_installer{type!="",invoker!=""}`: true,
```
LiliC: How come we do not check that this value is > 0 as well?

smarterclayton: good point

s-urbaniak:
just one little concern from my side: Is this asserting a functioning monitoring stack? If not I suggest to create a dedicated `platforms` e2e test folder.

smarterclayton:
Openshift e2e requires a cluster monitoring stack (non optional part), unless you mean something different?

smarterclayton:
/retest

With Michal’s fixes

smarterclayton:
To catch regressions

PR #24118, opened 2019-11-08

smarterclayton:


s-urbaniak:
hmm .. is this asserting if we have functioning monitoring stack? If not I suggest to have a dedicated `platforms` e2e test folder which is asserting those metrics.

PR #24123, opened 2019-11-11
Add "get configmaps from openshift namespace" rule to expected rules

JAORMX:
The PR https://github.com/openshift/openshift-apiserver/pull/42 adds
this rule to the defaults; so, we need to expect it as part of the
tests.


sttts:
/hold cancel

openshift/openshift-apiserver#42 was lgtm'ed by @soltysh

PR #24126, opened 2019-11-11
Bug 1711535: Update test logic to simulate NodeReady/False and NodeReady/Unknown events correctly

damemi:
Picks upstream PR https://github.com/kubernetes/kubernetes/pull/84883 in order to attempt to fix taintbasedevictions e2es again
However this may also require the changes from upstream PR 82072, which switches from using channels to the `context` package in the scheduler. That change doesn't apply directly to our current vendored state because of the many other recent changes to the scheduler. It can apply, but for simplicity I'm first going to see if sticking with a `stopCh` (in the one spot it's used) can still work for us


damemi:
As it turns out, upstream integration tests aren't run in our CI. So there's no need for us to continue trying to carry patches for this and I've closed the associated bug.

PR #24135, opened 2019-11-12
Bug 1767295: UPSTREAM: 84466: gce: skip ensureInstanceGroup for a zone that has no remaining nodes for k8s managed IG

abhinavdahiya:
Make sure we don't ensure cluster-managed instance group for a zone that has no more nodes left. Creating am IG for a zone that has no instances causes errors when attaching to backend service.

```
failed to ensure load balancer: googleapi: Error 400: Invalid value for field 'resource.backends[5].group': 'https://www.googleapis.com/compute/v1/projects/openshift-qe/zones/us-central1-c/instanceGroups/k8s-ig--8d8682bc12c7a717'. Instance group must have a network to be attached to a backend service. Add an instance to give the instance group a network., invalid
```


sdodson:
/retest
Failing to attach/detach volumes on AWS, code only affects gce provider. Will override on next flake.

PR #24136, opened 2019-11-12
Bug 1771662: OpenStack: show original cloud provider init errors

Fedosin:
Now if an error happens during cloud provider initialization, users just see an abstract message:

"cannot initialize cloud provider using data from the secret"

without any details.

This commit wraps original errors to make them shown in the output.


pierreprinetti:
I suspect that `err` was purposely ignored in order to avoid printing secrets contained in the `clouds.yaml`.

WDYT?

Fedosin:
@pierreprinetti I don't think so :) The rest of the errors are shown, so there's no point in hiding them.

pierreprinetti:
> The rest of the errors are shown

Well if other secrets are logged elsewhere, we might want to fix that! WDYT?

(unless I'm wrong and the line you changes do _not_ log any secret, then I'm absolutely fine LGTM'ing)

mandre:
/hold
We certainly don't want to leak out secrets in logs.

nee1esh:
What are the next steps for this PR and bug? given that details are secret, do we close the bug as WONTFIX?

PR #24145, opened 2019-11-13
e2e: run router header test on AWS/GCP/Azure only

frobware:
This is a follow-up to https://github.com/openshift/origin/pull/24093 which breaks (or flakes on) OpenStack.

Rather than trying to run down all the flakes on all the platforms in
4.3 I've changed the router header test to run against the following
platforms only:

- AWS
- Azure
- GCP

A new BZ[1] has been opened to ensure that the other platforms are not
forgotten.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1772125

### DIFF
--- a/test/extended/router/headers.go
+++ b/test/extended/router/headers.go
@@ -27,10 +27,13 @@ var _ = g.Describe("[Conformance][Area:Networking][Feature:Router]", func() {
 
 		routerIP  string
 		metricsIP string
+		infra     *configv1.Infrastructure
 	)
 
 	g.BeforeEach(func() {
 		var err error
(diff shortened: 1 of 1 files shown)


ironcladlou:
To avoid doing anything at all on those platforms (and also providing output about the skipping),  I believe you can add a `g.Skip()` call during a platform check added to `g.BeforeEach()`

frobware:
> To avoid doing anything at all on those platforms (and also providing output about the skipping), I believe you can add a `g.Skip()` call during a platform check added to `g.BeforeEach()`

Done in https://github.com/openshift/origin/pull/24145/commits/0e40910ff426967a37505acf618ff8b9401f0382

mandre:
Thanks for looking into it. LGTM.

ironcladlou:
Thanks!

/lgtm

PR #24150, opened 2019-11-13
dynamic certs and cluster auth

deads2k:
we should change how our authenticator overrides what's there


vendor/k8s.io/kubernetes/pkg/master/controller/clusterauthenticationtrust/cluster_authentication_trust_controller.go
```diff
@@ -0,0 +1,516 @@
...
+
+	if rhs != nil {
+		for _, curr := range rhs.Value() {
+			if present.Has(curr) {
+				continue
+			}
+			ret = append(ret, curr)
+			present.Insert(curr)
+		}
+	}
```
stlaz: Why not just
```
if lhs != nil {
    present.Insert(lhs.Value()...)
}
if rhs != nil {
    present.Insert(rhs.Value()...)
}

return headerrequest.StaticStringSlice(present.List())
```

Does the input order matter or is it just so that the behavior matches the other `combine*` functions?

deads2k: commenting on changes to upstream code.  We could change the upstream, but I don't think we want to.

stlaz: pfft, I thought <drop> means our changes on top of upstream that are supposed to be dropped on new kube version rebase. My bad.

deads2k: > pfft, I thought `<drop>` means our changes on top of upstream that are supposed to be dropped on new kube version rebase. My bad.

The features came in over the course of 8 or so PRs.  made more sense to bring it all in here at once

PR #24155, opened 2019-11-14
Bug 1761346: e2e: populate pending pods map everytime with most recent data

vareti:


test/extended/operators/cluster.go
```diff
@@ -23,7 +23,7 @@ var _ = g.Describe("[Feature:Platform] Managed cluster should", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		var lastPodsWithProblems []*corev1.Pod
-		var pending map[string]*corev1.Pod
+		var lastPending map[string]*corev1.Pod
```
stlaz: We don't want any of the pods to be crashlooping anywhere throughout the two minutes, but we also don't want the pods to be pending for the whole time.

By checking the `lastPending` you may lose the information about any infra pod crashlooping if it was replaces later by a healthy/semi-healthy pod.

I think you may need to climb `ownerReferences` to check whether by any chance the pending pods you found somewhere through the loop (and which were pending only because they did not start yet, not because they crashed) aren't pods that had already been replaced any current pods which are running just fine.

vareti: I might have jumped quickly to conclusions on this. Let me go back and take a look at this again.

vareti: If we use owner references to track if pods are replaced, the logic seems to be getting little complicated when crash looping pods are involved. The way I solved it is that `lastPending` still has the latest copy of pending pods, `podsWithProblems`  is updated every loop with pods that encountered problems and it state is retained for final processing after 2 minutes. With this test will not miss out on any pods that are crash looping and also account for pods that are previously pending(only because they did not start) and are later replaced.

stlaz:
/lgtm

pending pods at the end belong to the set of pods that were pending the entire time and no time checks were performed before anyway, so this is an improvement when the pods that were replaced by a properly running pod (under the same replicaset/deployment) are no longer reported

PR #24168, opened 2019-11-18
use upstream authentication initialization for kas.

deads2k:
builds on https://github.com/openshift/origin/pull/24150

aligns us closer with upstream authentication stack, making our diff very clear


vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch_authenticator.go
```diff
@@ -178,34 +179,42 @@ func newAuthenticator(
 	// build cert authenticator
 	// TODO: add "system:" prefix in authenticator, limit cert to username
 	// TODO: add "system:" prefix to groups in authenticator, limit cert to group name
-	dynamicCA := certs.NewDynamicCA(apiClientCABundle)
-	if err := dynamicCA.CheckCerts(); err != nil {
+	clientCAContentProvider, err := dynamiccertificates.NewDynamicCAContentFromFile("client-ca-bundle", apiClientCABundle)
```
stlaz: You remove all of this file's content in favour of upstream, do we still need this commit?

stlaz:
```
Error: unable to add "quota.openshift.io-clusterquotamapping" because it was already registered by: goroutine 1 [running]:

```
I seem to have been too eager to tag

stlaz:
/retest
hm, this is weird, there are minor line differences in some of the logs? As if it weren't testing the same thing?

vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
```diff
@@ -87,10 +65,10 @@ func NewOpenShiftKubeAPIServerConfigPatch(delegateAPIServer genericapiserver.Del
 
 		// ADMISSION
 		clusterQuotaMappingController := newClusterQuotaMappingController(kubeAPIServerInformers.KubernetesInformers.Core().V1().Namespaces(), kubeAPIServerInformers.OpenshiftQuotaInformers.Quota().V1().ClusterResourceQuotas())
-		patchContext.postStartHooks["quota.openshift.io-clusterquotamapping"] = func(context genericapiserver.PostStartHookContext) error {
+		genericConfig.AddPostStartHookOrDie("quota.openshift.io-clusterquotamapping", func(context genericapiserver.PostStartHookContext) error {
```
stlaz: This causes double-registration as `genericConfig` is later used to contruct `openshiftNonAPIServer`, which is passed to the `StartingDelegate` global variable from `patch_openshift.go` during `buildGenericConfig`.

The delegate variable is then used in `createAPIExtensionsServer()` during its config completion alongside config derived from the `genericConfig` above, so it dies as it tries to merge post-start hooks of these two.

Even if I move this to just `enablement.AddPostStartHookOrDie()`, the API servers initialization fails later in `CreateKubeAPIServer()`with registration of `"openshift.io-kubernetes-informers-synched"`, but would perhaps also fail with the `quota` as it's going to use more or less the same delegate and config for its config completion.

I am not sure how to get out of this double-registration hell.

PR #24169, opened 2019-11-19
Bug 1771809: add PodDisruptionBudgetAtLimit alert e2e

sallyom:
test for PodDisruptionBudgetAtLimit Alert fires when a PodDisruptionBudget with MinAvailable is created. 

see also: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/316  
https://bugzilla.redhat.com/show_bug.cgi?id=1771809#c1
/cc @soltysh

### DIFF
--- a/test/extended/prometheus/alerts.go
+++ b/test/extended/prometheus/alerts.go
@@ -1,0 +1,94 @@
+package prometheus
+
+import (
+	"time"
+
+	g "github.com/onsi/ginkgo"
+
+	v1 "k8s.io/api/core/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	intstr "k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	clientset "k8s.io/client-go/kubernetes"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/framework/pod"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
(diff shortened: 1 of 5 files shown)


sallyom:
Failing bc alert is 'pending' not 'firing' -via slack discussion it's ok to check for 'pending' alert.  Now, have to figure out best place for this test (best place is not here where I've put it, but conveniently the extended/prometheus has set up testing alerts w/ queries already so was convenient place to start.

test/extended/prometheus/alerts.go
```diff
@@ -0,0 +1,94 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	clientset "k8s.io/client-go/kubernetes"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/framework/pod"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[Feature:PodDisruptionBudgetAtLimitAlert][Conformance] Prometheus", func() {
```
sallyom: This is so if/when this test fails cluster-kube-controller-manager-operator (API/Workloads) peeps might get the bug/issue rather than monitoring folks.

sallyom:
> Now, have to figure out best place for this test

I've created a new file test/extended/prometheus/alerts.go and have included all necessary fns to make the query in one file test/extended/prometheus/util.go, so anyone else wanting to check on a pending alert might find them.
I don't believe the 'pending' alert will interfere with the current check that no alerts are 'firing' other than Watchdog alert, since the alert will remain pending for 15 min, and as soon as the test completes (as soon as the alert is 'pending') the pod/ns is deleted.

sallyom:
closing in favor of openshift/cluster-kube-controller-manager-operator#316, https://github.com/openshift/library-go/pull/602

PR #24001, opened 2019-10-22
Disable the s2i build with quota memory test

umohnani8:
It is currently failing with cri-o 1.16 so disabling
the test for now to unblock cri-o.
Have opened a high severity bug for this https://bugzilla.redhat.com/show_bug.cgi?id=1764323

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

PR #24174, opened 2019-11-19
Bug 1787457: Read cgroup settings from the right cgroup for build-quota tests

nalind:
The assemble script was reading from the top of the controller's group hierarchy, even though the script itself is run inside of a different control group, which is where the settings are applied.  This is related to investigation in #24157.

### DIFF
--- a/test/extended/builds/s2i_quota.go
+++ b/test/extended/builds/s2i_quota.go
@@ -53,11 +53,9 @@ var _ = g.Describe("[Feature:Builds][Conformance] s2i build with a quota", func(
 				o.Expect(br.Build.Status.Duration).To(o.Equal(duration), "Build duration should be computed correctly")
 
 				g.By("expecting the build logs to contain the correct cgroups values")
-				// buildLog, err := br.LogsNoTimestamp()
-				_, err = br.LogsNoTimestamp()
+				buildLog, err := br.LogsNoTimestamp()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				// TODO: re-enable this check when https://bugzilla.redhat.com/show_bug.cgi?id=1764323 is resolved
-				//o.Expect(buildLog).To(o.ContainSubstring("MEMORY=419430400"))
+				o.Expect(buildLog).To(o.ContainSubstring("MEMORY=419430400"))
 				// TODO: re-enable this check when https://github.com/containers/buildah/issues/1213 is resolved.
 				//o.Expect(buildLog).To(o.ContainSubstring("MEMORYSWAP=419430400"))
 


umohnani8:
Nice! the s2i build quota test passes now. Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1764323.
LGTM
@adambkaplan @bparees PTAL

umohnani8:
The test was disabled in https://github.com/openshift/origin/pull/24001 to unblock cri-o 1.16

adambkaplan:
/retitle Bug 1764323: Read cgroup settings from the right cgroup for build-quota tests

test/extended/builds/s2i_quota.go
```diff
@@ -53,11 +53,9 @@ var _ = g.Describe("[Feature:Builds][Conformance] s2i build with a quota", func(
...
 				g.By("expecting the build logs to contain the correct cgroups values")
-				// buildLog, err := br.LogsNoTimestamp()
-				_, err = br.LogsNoTimestamp()
+				buildLog, err := br.LogsNoTimestamp()
 				o.Expect(err).NotTo(o.HaveOccurred())
-				// TODO: re-enable this check when https://bugzilla.redhat.com/show_bug.cgi?id=1764323 is resolved
-				//o.Expect(buildLog).To(o.ContainSubstring("MEMORY=419430400"))
+				o.Expect(buildLog).To(o.ContainSubstring("MEMORY=419430400"))
 				// TODO: re-enable this check when https://github.com/containers/buildah/issues/1213 is resolved.
 				//o.Expect(buildLog).To(o.ContainSubstring("MEMORYSWAP=419430400"))
```
adambkaplan: @nalind can we re-enable this check, too?

bparees:
> The assemble script was reading from the top of the controller's group hierarchy, even though the script itself is run inside of a different control group

this has implications for how s2i images compute their container memory and set JVM limits.  @gabemontero this is probably related to the customer wildfly BZ you had.

When/why did this change?  it seems like a change in behavior from 4.2 to 4.3 that's going to effect customer builds if their assemble scripts are reading the cgroup to find the memory limit, as many of our images do.

nalind:
This looks like a change in how we're running privileged containers and pods.  In 1.14 (4.2), runc would be given these in its configuration for a privileged container:
```
        {
            "destination": "/sys",
            "type": "sysfs",
            "source": "sysfs",
            "options": [
                "nosuid",
                "noexec",
                "nodev",
                "rw"
            ]
        },
	{
            "destination": "/sys/fs/cgroup",
            "type": "cgroup",
            "source": "cgroup",
            "options": [
                "nosuid",
                "noexec",
                "nodev",
                "relatime",
                "rw"
            ]
        },
```
An unprivileged container would get the same mounts, but marked read-only.

In 1.16 (4.3), a privileged container instead gets
```
        {
            "destination": "/sys",
            "type": "bind",
            "source": "/sys",
            "options": [
                "nosuid",
                "noexec",
                "nodev",
                "rw",
                "rbind"
            ]
        }
```
I bisected the change down to https://github.com/cri-o/cri-o/pull/2627.

giuseppe:
the cgroup handling code LGTM

nalind:
/hold

https://github.com/cri-o/cri-o/pull/2997 is preferable, since it gets us back to the behavior we had in 4.2.

nalind:
Updated to only re-enable the test that was failing.

nalind:
Depends on https://github.com/cri-o/cri-o/pull/2997.

haircommander:
/retitle Read cgroup settings from the right cgroup for build-quota tests

(this now targets 4.4, where the above bug targeted 4.3)

haircommander:
/retitle Bug 1787457: Read cgroup settings from the right cgroup for build-quota tests

haircommander:
LGTM PTAL @bparees @mrunalp @adambkaplan

test/extended/builds/s2i_quota.go
bparees: @nalind bump..I saw some discussion in the bugs around this but i didn't follow it closely enough to understand if the memoryswap values should be correct now or not.

nalind: I don't expect it to be set to the value that the test would expect, since cri-o can't be told to do it for us (there doesn't appear to be a `k8s.io/api/core/v1.ResourceName` for swap limits - it can be told to set memory usage, and https://github.com/cri-o/cri-o/pull/2997 fixed a problem with it, allowing us to re-enable the `MEMORY=` test), and we don't currently have logic in buildah's chroot isolation that would attempt to set control group settings.

adambkaplan: @nalind if that's the case, can you add a commend to this effect in https://github.com/containers/buildah/issues/1213?

nalind: Added that [note](https://github.com/containers/buildah/issues/1213#issuecomment-571277448).  I'm _really_ hoping we can lean on libcontainer for that.

PR #24184, opened 2019-11-21
bump library-go to fix wrapping not found error with forbidden in scc admission

mfojtik:
Details: https://github.com/openshift/apiserver-library-go/pull/11

This also pin 2 deps that are moving almost every bump and they diverged from what kube 1.16.2 use.
The env var in hack/update-deps.sh allows one to skip the check whether the k8s staging repos are in GOPATH, because it is super annoying and I always had to edit that file and already committed it twice ;-)


mfojtik:
/cc @smarterclayton 

Since you asked for this :-)

PR #24185, opened 2019-11-21
Bug 1763936: Disable node-Service and pod-Service granular checks

pierreprinetti:
https://bugzilla.redhat.com/show_bug.cgi?id=1763936

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -457,6 +457,10 @@ var (
 		"[Flaky]": {
 			`Job should run a job to completion when tasks sometimes fail and are not locally restarted`, // seems flaky, also may require too many resources
 			`openshift mongodb replication creating from a template`,                                     // flaking on deployment
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1763936
+			`\[sig-network\] Networking Granular Checks: Services should function for node-Service`,
+			`\[sig-network\] Networking Granular Checks: Services should function for pod-Service`,
+			`\[sig-network\] Networking Granular Checks: Services should function for endpoint-Service`,
 
 			// TODO(node): test works when run alone, but not in the suite in CI
 			`\[Feature:HPA\] Horizontal pod autoscaling \(scale resource: CPU\) \[sig-autoscaling\] ReplicationController light Should scale from 1 pod to 2 pods`,


pierreprinetti:
/hold
needs further investigation.

pierreprinetti:
/hold cancel

These tests are still impacting the CI (see [OpenStack tests](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-openstack-4.3/466)).

iamemilio:
@pierreprinetti to synchronize bugzilla and a pull request automatically, the title of the pull request should be in this format.
```
Bug ###: <whatever title you would like>
```

wking:
/retitle Bug 1763936: Disable node-Service and pod-Service granular checks

wking:
Tests all failed for unrelated reasons.

/retest

pierreprinetti:
I have rebased, in the hopes we're lucky and the YAML failures are fixed by someone else.

pierreprinetti:
@mfojtik 
Do you think this has chances to get /approval?

pierreprinetti:
/honk
:bird:

test/extended/util/test.go
```diff
@@ -533,6 +533,9 @@ var (
 			`\[sig-network\] Services should have session affinity work for service with type clusterIP`,
 			// SDN-587: OVN-Kubernetes doesn't support hairpin services
 			`\[sig-network\] Services should allow pods to hairpin back to themselves through services`,
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1763936
+			`\[sig-network\] Networking Granular Checks: Services should function for node-Service`,
+			`\[sig-network\] Networking Granular Checks: Services should function for pod-Service`,
```
danwinship: This is in the `[Skipped:Network/OVNKubernetes]` section, which means it will only be skipped for clusters using ovn-kubernetes, which is not what you mean. You presumably want to skip regardless of network plugin, so this would go in the main skip list.

pierreprinetti: Thank you, that's valid information finally!

pierreprinetti:
@danwinship Thanks again for reviewing.

While we assess whether disabling the tests is the right thing to do, do you mind having a look over whether this second try is the right way to do it?

test/extended/util/test.go
```diff
@@ -533,7 +537,6 @@ var (
 			`\[sig-network\] Services should have session affinity work for service with type clusterIP`,
 			// SDN-587: OVN-Kubernetes doesn't support hairpin services
 			`\[sig-network\] Services should allow pods to hairpin back to themselves through services`,
-			`\[sig-network\] Networking Granular Checks: Services should function for endpoint-Service`,
```
danwinship: Sorry, this one (endpoint-Service) needs to stay in the ovn-kubernetes skip list as well because even if we fix the flake, this particular test doesn't pass under ovn-kubernetes. So we'll have this test listed in two different places (for two different reasons). Other than that, the patch lgtm.

smarterclayton:
I see no justification in the bug for why these tests would be disabled.

We should not be removing tests in SDN unless we have clear justifications why the test (and not SDN) is the root cause.

smarterclayton:
Without sufficient documentation I will revert this PR to ensure we aren't dropping coverage.

smarterclayton:
I see no flakes in the network stress test 

https://testgrid.k8s.io/redhat-openshift-ocp-release-4.4-informing#release-openshift-origin-installer-e2e-aws-sdn-network-stress-4.4

If the tests become flaky under load, you need to fix load, or fix the test.  We do not disable tests that fail on alternate platforms, we identify their root cause and fix them.

pierreprinetti:
Thank you Clayton for commenting.

The issue seems to be impacting different infrastructures. As far as OpenStack is concerned, our plan is to do whatever reasonably possible to get a stable CI, and then start extending the coverage from there.

The debate on what is reasonable is, of course, open. A proper investigation of this particular issue is currently beyond our capacity, so we decided to propose a skip.

PR #24186, opened 2019-11-21
Bug 1765294: Adding debug to test.

ricardomaraschini:
Added debug to a test. This test is constantly flaking and we need to
discover why, this debug will report all attempts to get the secret.


adambkaplan:
/close

Replaced by #24187 (I won? the race to get the first flake failure).

PR #24191, opened 2019-11-22
Bug 1775757: KubeAPI server doesn't accept large patch requests

p0lyn0mial:


vendor/k8s.io/kubernetes/test/integration/apiserver/max_request_body_bytes_test.go
```diff
@@ -109,12 +125,12 @@ func TestMaxResourceSize(t *testing.T) {
 			t.Errorf("expected success or bad request err, got %v", err)
 		}
 	})
-	t.Run("ApplyPatchType should handle a patch just under the max limit", func(t *testing.T) {
```
p0lyn0mial: almost a clean pick, I had to remove tests for `ApplyPatchType` because it doesn't apply here

PR #23909, opened 2019-10-03
Bug 1758159: aws: sort addresses of multiple interfaces correctly [4.1]

danwinship:
Backport of #23870 to 4.1

PR #24193, opened 2019-11-24
WIP: fix AWS NodeAddress sorting

danwinship:
#23909 did not actually fix this problem in 4.1... there must be something different somewhere in the 4.1 code?


danwinship:
4.1 is basically dead and the customer has upgraded to 4.2

PR #24194, opened 2019-11-25
WIP add extended test for operator metrics

sanchezl:
Add tests to verify expected metrics are being exposed by operators and scraped by Prometheus.

### DIFF
--- a/test/extended/operators/metrics.go
+++ b/test/extended/operators/metrics.go
@@ -1,0 +1,144 @@
+package operators
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"k8s.io/client-go/kubernetes"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"k8s.io/client-go/dynamic"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	config "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+)
+
+var _ = g.Describe("OpenShift Operators", func() {
+
+	var (
+		clusterOperatorInfos []clusterOperatorInfo
+		kubeClient           kubernetes.Interface
+		dynamicClient        dynamic.Interface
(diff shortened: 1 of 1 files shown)


test/extended/operators/metrics.go
```diff
@@ -0,0 +1,144 @@
...
+			"openshift-samples",
+			"operator-lifecycle-manager",
+			"operator-lifecycle-manager-catalog",
+			"operator-lifecycle-manager-packageserver",
+			"service-ca",
+			"service-catalog-apiserver",
+			"service-catalog-controller-manager",
+			"storage",
+			"support",
+		)
```
sttts: with a black list we are super far in terrain of the monitoring team. Either we switch to a whitelist of our operators, or we gift this to @s-urbaniak and @brancz.

test/extended/operators/metrics.go
```diff
@@ -0,0 +1,144 @@
...
+	config "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+)
+
+var _ = g.Describe("OpenShift Operators", func() {
+
+	var (
+		clusterOperatorInfos []clusterOperatorInfo
+		kubeClient           kubernetes.Interface
+		dynamicClient        dynamic.Interface
```
sttts: reused globals are an go anti-pattern.

test/extended/operators/metrics.go
sanchezl: I would feel more comfortable with that approach also. There was a desire though to catch any "new" operator that did not get added to the whitelist, but I didn't find (yet) a reliable method of identifying "our" operators.

s-urbaniak: @sanchezl The monitoring team does not maintain a list or a "registry" of operators that expose metrics in a given OpenShift installation. Setting up service monitors should be in the responsibility of a given operator itself and the correctness of that setup should be covered in e2e tests in the corresponding operator code base. I have doubts we should maintain such a list (be it blacklist or whitelist) in origin itself.

That also corresponds to the bug process. If exposing metrics for a given operator don't work because the service monitor setup is not correct, it is the concrete operator's responsibility to fix it.

/cc @smarterclayton @brancz

marun:
@sanchezl While I appreciate your initiative, I don't think this PR as proposed should be merged. I would echo [the comment](https://github.com/openshift/origin/pull/24194#discussion_r350658611) from a member of the monitoring team that it is the responsibility of each operator to validate that prometheus is  correctly configured to read their metrics. Centralizing the testing of metrics endpoints in this repo could also complicate any metrics changes since it could require coordinating the merge of PRs across an operator repo and this repo.

@sallyom's is still working on [the PR to simplify querying of metrics](https://github.com/openshift/library-go/pull/602), and once that is available each operator will be responsible for implementing a test (if they haven't already). A periodic job testing service ca rotation will likely have to duplicate or consume the operator testing.

PR #24200, opened 2019-11-25
hack: fix unbound variable

mfojtik:


mfojtik:
self-lgtm, this is 1 character fix that already merged to 4.3

PR #24207, opened 2019-11-25
UPSTREAM: <carry>: use a filter for well-known instead of an apiserver delegate

deads2k:
simplifying the kube-apiserver patches to avoid rewiring bugs.


vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch_handlerchain.go
```diff
@@ -12,10 +12,23 @@ import (
...
+func BuildHandlerChain(consolePublicURL string, oauthMetadataFile string) (func(apiHandler http.Handler, kc *genericapiserver.Config) http.Handler, error) {
+	// load the oauthmetadata when we can return an error
+	oAuthMetadata := []byte{}
+	if len(oauthMetadataFile) > 0 {
+		var err error
+		oAuthMetadata, err = loadOAuthMetadataFile(oauthMetadataFile)
+		if err != nil {
+			return nil, err
+		}
+	}
```
sttts: as in the other PR: odd to see loading code so deep here, and not high level, with a `[]byte` passed in here.

PR #24217, opened 2019-11-26
extended util oauthserver - increase pod ready wait time

stlaz:
It seems that the images are being pulled for up to 55 seconds in
the CI test runs, which would leave the oauth-server to start for
only about 5 seconds.

Increase the timeout for pod readiness in order to stop jobs from
failing too often.


stlaz:
let's just merge this and use the BZ for a backport

PR #24218, opened 2019-11-26
Bug 1741417: UPSTREAM: 81330: Hide bearer token in logs

sttts:


mfojtik:
approving on last minute as this is a CVE fix

PR #24224, opened 2019-11-26
Bug 1775826: [release-4.3] UPSTREAM: drop 83592: added --reserved-cpus kubelet command option

vladikr:
Backport of https://github.com/kubernetes/kubernetes/pull/83592


rphillips:
/lgtm cancel

this should be applied to master first, then backported

sjenning:
this is really backporting a feature.  we typically do not do this.  features that go in upstream come into origin via rebasing typically.  is there an exception in this case?

vladikr:
> this is really backporting a feature. we typically do not do this. features that go in upstream come into origin via rebasing typically. is there an exception in this case?

@sjenning Yes, this feature is needed in 4.3 as per the bz. Perhaps, @eparis could provide the justification?

vladikr:
> this should be applied to master first, and then backported. But since this is a new feature, we will pick this up with the 1.17 rebase.

@rphillips Yes, I wasn't sure about the process. I only knew that kubernetes/kubernetes#83592
was merged upstream and thought it should already appear in master.

What is the right course of action should be?

derekwaynecarr:
Is this actually settable without a corresponding bump of the machine config operator?

vladikr:
> Is this actually settable without a corresponding bump of the machine config operator?

Yes, if you manually set it in kubelet.conf or hyperkube manually. However, MCO will require an update, to include the schema to be able to set reservedSystemCPUs.

I'm not sure where do I need to make this change in MCO, but I thought to take care of this backport first.

derekwaynecarr:
required for isolcpu integration.

/approve

PR #24227, opened 2019-11-27
Check the oc explain about opsrc&csc

emmajiafan:
This auto cover the `oc explain` about opsrc&csc .


test/extended/marketplace/marketplace_explain.go
```diff
@@ -0,0 +1,26 @@
...
+	)
+
+	//OCP-23670 description info for opsrc&csc CRDs
+	g.It("ensure the opsrc & csc can be explain by oc command", func() {
+
+		explainOpsrc, _ := oc.AsAdmin().WithoutNamespace().Run("explain").Args("operatorsource").Output()
+		o.Expect(explainOpsrc).ShouldNot(o.ContainSubstring("<empty>"))
+		explainCsc, _ := oc.AsAdmin().WithoutNamespace().Run("explain").Args("catalogsourceconfig").Output()
+		o.Expect(explainCsc).ShouldNot(o.ContainSubstring("<empty>"))
+	})
```
awgreene: @emmajiafan this test looks great - but with the longterm plans of deprecating the two CRDs this would add additional work.

Thoughts @kevinrizza

emmajiafan: Then the related PRs can be in branch and removed from the master? I think so.

kevinrizza: This significantly adds complexity to the development workflow of these features, since a single pr can't be created to update these tests at the same time as the pr that modifies the source. Essentially the workflow becomes:

1. Create a PR in this repository that _turns off_ the test
2. Create a PR that modifies the feature
3. Create a PR that reenables this test with modifications

I think this workflow doesn't make much sense, and we should consider how to write these tests with the expectation that we should only be writing tests based on API guarantees -- my opinion being that a log message is _not_ something we should be guaranteeing.

awgreene: I agree with @kevinrizza - this is a shortcoming that we need to address.

emmajiafan:
@shawn-hurley Could you help add  "/approved" label please? Thank you!

test/extended/marketplace/marketplace_lables.go
```diff
@@ -104,7 +104,6 @@ var _ = g.Describe("[Feature:Marketplace] Marketplace resources with labels prov
 			{"catalogsource", "csctestlabel", "-o=jsonpath={.spec.displayName}", allNs},
 			{"catalogsource", "csctestlabel", "-o=jsonpath={.spec.publisher}", allNs},
 		}
-
```
kevinrizza: Please avoid editing this unrelated file in the pr.

PR #24234, opened 2019-11-27
Rewrite commitchecker so it validates bumps as well

tnozicka:
/cc @sttts @deads2k 

@openshift/sig-master

### DIFF
--- a/tools/rebasehelpers/commitchecker/validate.go
+++ b/tools/rebasehelpers/commitchecker/validate.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bytes"
 	"fmt"
-	"os/exec"
 	"regexp"
 	"strings"
 	"text/template"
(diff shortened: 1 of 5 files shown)


sttts:
Naive question: why do we need a full rewrite?

tnozicka:
> Naive question: why do we need a full rewrite?

Can't say I wasn't expecting that question. The godeps have been pretty bundled to it and all the function were assuming they never get bump commits.

tools/rebasehelpers/commitchecker/validate.go
```diff
@@ -3,19 +3,53 @@ package main
...
+		ValidateUpstreamCommit,
+		ValidateBumpCommit,
+	}
+)
+
+func ValidateCommitAuthor(commit util.Commit) []string {
+	var allErrors []string
+
+	if strings.HasPrefix(commit.Email, "root@") {
+		allErrors = append(allErrors, fmt.Sprintf("Commit %s has invalid email %q", commit.Sha, commit.Email))
```
sttts: why strings for errors?

tools/rebasehelpers/util/git.go
```diff
@@ -9,25 +9,87 @@ import (
...
-	"k8s.io":            2,
-	"speter.net":        2,
+var (
+	UpstreamSummaryPattern = regexp.MustCompile(`^UPSTREAM: (revert: )?(([\w\.-]+\/[\w-\.-]+)?: )?(\d+:|<carry>:|<drop>:)`)
+	BumpSummaryPattern     = regexp.MustCompile(`^bump[\(\w].*`)
+
+	// patchRegexps holds regexps for paths inside vendor dir that are allowed to be patched directly.
+	// It must corresponds to published repositories.
+	PatchRegexps = []*regexp.Regexp{
+		regexp.MustCompile("^k8s.io/.*"),
```
sttts: actually only k8s.io/kubernetes. Staging dirs also have that prefix. `k8s.io/{utils,kube-openapi,gengo,....}` must not be patched.

tools/rebasehelpers/commitchecker/validate.go
tnozicka: errors may not start with capital letter

tools/rebasehelpers/util/git.go
tnozicka: yeah, looks like we only publish staging dirs, this should match that list

tnozicka: done in https://github.com/openshift/origin/compare/ee8636db3b0609dfa40c81534c2fd4303769d6e7..80cfb21aad257b5dadcbc9893f2573e05b5f8613

sttts: Technically not even the staging dirs, but just kube. Your linked commit looks good.

PR #24236, opened 2019-11-28
Check the different name of opsrc&csc

emmajiafan:
cluster doesn't support install the same name opsrc & csc.


jianzhangbjz:
LGTM, thanks! @shawn-hurley Could you help give approval? Thanks!
/assign @shawn-hurley

test/extended/marketplace/marketplace_diffname.go
```diff
@@ -0,0 +1,86 @@
...
+			o.Expect(msg).Should(o.BeTrue())
+		}
+
+		// create the csc samename
+		cscYaml, err := oc.AsAdmin().Run("process").Args("--ignore-unknown-parameters=true", "-f", cscYamltem, "-p", "NAME=samename", fmt.Sprintf("NAMESPACE=%s", allNs), fmt.Sprintf("MARKETPLACE=%s", marketplaceNs), "PACKAGES=camel-k-marketplace-e2e-tests", "DISPLAYNAME=samename", "PUBLISHER=samename").OutputToFile("config.json")
+		err = createResources(oc, cscYaml)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		//check the csc status
+		output, _ := oc.AsAdmin().WithoutNamespace().Run("get").Args("catalogsourceconfig", "samename", "-o=jsonpath={.status.currentPhase.phase.message}", "-n", marketplaceNs).Output()
+		o.Expect(output).Should(o.ContainSubstring("Deployment samename exists"))
```
kevinrizza: Rather than check this string, why not just check that `status.currentPhase.phase` is `Failed`?

emmajiafan: Because the "failed" reason is the "same name" . Hmm I will also add the check of "status.currentPhase.phase".

emmajiafan:
@kevinrizza  could you help add "/lgtm" label if this pr is ok?

test/extended/marketplace/marketplace_diffname.go
kevinrizza: Yes but by saying that that we expect the string `"Deployment samename exists"` to be in the status we are essentially guaranteeing that as a contract. What if we wanted to wrap that error, or add additional details? Then this test will fail. I don't think it's necessarily valuable to make this guarantee.

PR #24238, opened 2019-11-28
Explain tests

soltysh:
This PR:
1. prints CRD yaml when it fails match
2. uncoments CRDs which pass explain tests
/assign @sttts


adambkaplan:
/hold

`verify` is spinning.

PR #24250, opened 2019-12-03
Check the multi type operators support

emmajiafan:
Auto check the multi operators support and the permissive to errors in manifests


test/extended/marketplace/marketplace_multi.go
```diff
@@ -0,0 +1,163 @@
...
+				return true, nil
+			}
+			return false, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		logOfFlat, _ := oc.AsAdmin().WithoutNamespace().Run("logs").Args(podNameFlat, "-n", marketplaceNs).Output()
+		o.Expect(logOfFlat).Should(o.ContainSubstring("decoded 2 flattened and 0 nested"))
+
+		////create the opsrc that only has the nested operator
```
njhale: nit: extra `//`

test/extended/marketplace/marketplace_multi.go
```diff
@@ -0,0 +1,163 @@
...
+	var (
+		oc            = exutil.NewCLI("marketplace", exutil.KubeConfigPath())
+		marketplaceNs = "openshift-marketplace"
+		resourceWait  = 60 * time.Second
+
+		opsrcYamltem = exutil.FixturePath("testdata", "marketplace", "opsrc", "02-opsrc.yaml")
+	)
+
+	g.AfterEach(func() {
+		//clear the sub,csv resource
```
njhale: nit: most conventions prescribe a space between the comment directive and the first letter of the comment

test/extended/marketplace/marketplace_multi.go
```diff
@@ -0,0 +1,163 @@
...
+		allresourcelist := [][]string{
+			{"operatorsource", "flatted", marketplaceNs},
```
njhale: nit: "flattened"

test/extended/marketplace/marketplace_multi.go
```diff
@@ -0,0 +1,163 @@
...
+			{"operatorsource", "nested", marketplaceNs},
+			{"operatorsource", "multiformat", marketplaceNs},
```
njhale: Would it be sufficient to only test with a multi-format Appregistry namespace? That way, if we updated the `marketplace_e2e` appr namespace to be multi-format, we could avoid creating new namespaces.

test/extended/marketplace/marketplace_multi.go
```diff
@@ -0,0 +1,163 @@
...
+			output, err := oc.AsAdmin().Run("get").Args(podNameFlat, "-n", marketplaceNs).Output()
+			if err != nil {
+				e2e.Failf("Failed to get the pods podNameFlat status, error:%v", err)
+				return false, err
+			}
+			if strings.Contains(output, "1/1") {
+				return true, nil
+			}
+			return false, nil
+		})
```
njhale: I think it would be more readable if this was generalized and factored out into a helper function.

test/extended/marketplace/marketplace_multi.go
```diff
@@ -0,0 +1,163 @@
...
+
+		podNameMult, _ := oc.AsAdmin().WithoutNamespace().Run("get").Args("pods", "-n", marketplaceNs, "-l marketplace.operatorSource=multiformat", "-oname").Output()
+		//wait for pod is running
+		err = wait.Poll(5*time.Second, resourceWait, func() (bool, error) {
+			output, err := oc.AsAdmin().Run("get").Args(podNameMult, "-n", marketplaceNs).Output()
+			if err != nil {
+				e2e.Failf("Failed to get the pods podNameMult status, error:%v", err)
+				return false, err
+			}
+			if strings.Contains(output, "1/1") {
```
njhale: Is this sufficient to determine pod readiness? We should probably check the [`Ready` status condition](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions) instead.

test/extended/marketplace/marketplace_multi.go
emmajiafan: will remove

test/extended/marketplace/marketplace_multi.go
emmajiafan: will remove

test/extended/marketplace/marketplace_multi.go
emmajiafan: will update

test/extended/marketplace/marketplace_multi.go
emmajiafan: Hmm the original story and test case is : support only flattented , only nested, nested+ flattented, 3 kinds of opsrc . If @kevinrizza  say it is ok , then we need to push one "nested" operator and one "with error" operator to `marketplace_e2e` appr namespace to be multi-format and "with error operator" . Then we could avoid creating new ns.

test/extended/marketplace/marketplace_multi.go
emmajiafan: Good idea. will update

test/extended/marketplace/marketplace_multi.go
emmajiafan: Hmm seems like the "Ready status consition" uneasy to get from "oc get" command. Will search about it. Thank you for your all comment @njhale .

test/extended/marketplace/marketplace_multi.go
emmajiafan: will update

test/extended/marketplace/marketplace_multi.go
kevinrizza: I think it would be fine to just test the multi-format registry namespace. But I think it should not be the `marketplace_e2e` test namespace because that should, ideally, only be updated for the e2e tests in the marketplace repository. Otherwise someone could inadvertently break a test in a repo while trying to write a new test for the other repo. Can we just use this `multiformat` namespace for these tests?

emmajiafan:
@njhale  Could you help add "/lgtm" if the pr is ok? Thank you!

test/extended/marketplace/marketplace_multi.go
```diff
@@ -0,0 +1,74 @@
...
+		for _, source := range allresourcelist {
+			err := clearResources(oc, source[0], source[1], source[2])
+			o.Expect(err).NotTo(o.HaveOccurred())
+		}
+	})
+
+	// OCP-23090 check the multi format operators support
+	g.It("[ocp-23090 25675]create opsrc with multi format operators and operator with error", func() {
+
+		// OCP-25675 create the opsrc that has the flatted & nested operator and one operator has error
```
jianzhangbjz: @kuiwang02 Please help review it, thanks!

PR #24258, opened 2019-12-05
Bug 1779563: fix oc adm node logs and add tests

soltysh:
This should fix it once and for all, and I'm also adding tests. I'll create a followup shortly to add tests for the `--since` with date parameter, but given there's some time between when this lands in kubelet I'll hold that other PR.
/assign @smarterclayton 

@p0lyn0mial fyi, since you're working on the k8s bump


vendor/k8s.io/kubernetes/pkg/kubelet/kubelet_server_journal.go
```diff
@@ -241,7 +241,7 @@ var (
 )
 
 const (
-	dateFormat         = `2006-01-02T15:04:05.999999Z`
+	dateFormat         = `2006-01-02 15:04:05.999999`
```
smarterclayton: Why this change?

smarterclayton: Aren't you breaking the API here?

soltysh: journalctl accepts w/o those, currently if you run node-logs with ` --since "2019-12-05T08:00:00Z"` that will pass through here, but journalctl will return you error about failed parsing, from what I checked in the man page (https://www.freedesktop.org/software/systemd/man/journalctl.html): `Date specifications should be of the format "2012-10-30 18:17:16".` That's why we need this change, honestly I'm surprised this came up this late :cry:

smarterclayton: ok

PR #24259, opened 2019-12-05
Add adm node-logs with --since=date parameter test

soltysh:
This builds on top of https://github.com/openshift/origin/pull/24258 but should not be merged until the former is merged and a new image with new kubelet is delivered otherwise this will block the queue. 
I'll drop the hold once I'm certain the prereq is deployed. 
/hold

/assign @smarterclayton

### DIFF
--- a/test/extended/cli/adm.go
+++ b/test/extended/cli/adm.go
@@ -1,7 +1,9 @@
 package cli
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


soltysh:
The other merged some time ago, so this should work by now.
/hold cancel 
and let's see what the tests show.

PR #24265, opened 2019-12-07
Bug 1748073: UPSTREAM: 86009: kubelet: guarantee at most only one cinfo per containerID

rphillips:
Backports my fix from upstream with the kubelet injecting duplicate metrics causing 500 on the Kubelet's metric endpoint.

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1748073

cc @mrunalp @sjenning


rphillips:
Ci failure on pushing images 

/retest

PR #24269, opened 2019-12-09
bump(*): verify the branch reset works

mfojtik:


deads2k:
The k8s.io bits looks good, but release 4.3 glide.yaml has the wrong library-go and api/client branch deps

PR #24273, opened 2019-12-09
Bug 1781302: Handle build controller skew in tests

adambkaplan:
Update 4.2 tests so they can detect the internal registry hostname was
published and loaded by the build controller.


adambkaplan:
@smarterclayton fyi - up to you if you want to bypass the patch manager as this is needed for the skew testing for 4.3.

PR #24279, opened 2019-12-10
Remove gradle e2e on 3.11

akram:
gradle binary is not availaible anymore. 
Removing the e2e tests related to gradle.


gabemontero:
/assign @bparees 

cause bindata had to be updated

bparees:
this is removing a test, not modifying shipping code and not adding a new test, so overriding the BZ requirement.

PR #24282, opened 2019-12-11
Bug 1779694: bump to pull latest scope-escalating resources from apiserver-lib-go

stlaz:


stlaz:
/bugzilla refresh
let's see if it's fixed already
_duck and cover_

PR #24291, opened 2019-12-12
test for excessive etcd leadership changes

sanchezl:
Adds an extended test which watches the `etcd_server_leader_changes_seen_total` metric and fails if the total across all nodes is ever greater than 9.

### DIFF
--- a/test/extended/etcd/leader_changes.go
+++ b/test/extended/etcd/leader_changes.go
@@ -1,0 +1,26 @@
+package etcd
+
+import (
+	"context"
+	"time"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	"github.com/prometheus/common/model"
+
+	"github.com/openshift/origin/test/extended/prometheus/client"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("etcd", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("etcd-leader-change", exutil.KubeConfigPath())
+	g.It("leader changes are not excessive", func() {
+		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		g.By("Examining the rate of increase in the number of etcd leadership changes for last five minutes")
+		result, _, err := prometheus.Query(context.Background(), "increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[15m:1m])", time.Now())
+		o.Expect(err).ToNot(o.HaveOccurred())
+		o.Expect(result.(model.Vector)[0].Value).To(o.BeNumerically("==", 0))
+	})
+})
--- a/test/extended/prometheus/client/via_route.go
+++ b/test/extended/prometheus/client/via_route.go
@@ -1,0 +1,87 @@
+package client
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/prometheus/client_golang/api"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/transport"
+
+	routev1 "github.com/openshift/api/route/v1"
+
+	"github.com/openshift/origin/test/extended/util"
+)
+
+// NewE2EPrometheusRouterClient returns a Prometheus HTTP API client configured to
+// use the Prometheus route host, a bearer token, and no certificate verification.
+func NewE2EPrometheusRouterClient(oc *util.CLI) (prometheusv1.API, error) {
+	kubeClient := oc.AdminKubeClient()
+	routeClient := oc.AdminRouteClient()
+
+	// wait for prometheus service to exist
+	err := wait.PollImmediate(time.Minute, time.Second, func() (bool, error) {
+		_, err := kubeClient.CoreV1().Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+		return err == nil, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// wait for the prometheus route to exist
+	var route *routev1.Route
+	err = wait.PollImmediate(time.Minute, time.Second, func() (bool, error) {
+		route, err = routeClient.RouteV1().Routes("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+		return err == nil, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// retrieve an openshift-monitoring service account secret
+	var secret *corev1.Secret
+	secrets, err := kubeClient.CoreV1().Secrets("openshift-monitoring").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, currSecret := range secrets.Items {
+		if currSecret.Type == corev1.SecretTypeServiceAccountToken && strings.HasPrefix(currSecret.Name, "prometheus-") {
+			secret = &currSecret
+			break
+		}
+	}
+	if secret == nil {
+		return nil, fmt.Errorf("unable to locate service prometheus service account secret")
+	}
+
+	// prometheus API client, configured for route host and bearer token auth, and no cert verification
+	client, err := api.NewClient(api.Config{
+		Address: "https://" + route.Status.Ingress[0].Host,
+		RoundTripper: transport.NewBearerAuthRoundTripper(
+			string(secret.Data[corev1.ServiceAccountTokenKey]),
+			&http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout:   30 * time.Second,
+					KeepAlive: 30 * time.Second,
+				}).DialContext,
+				TLSHandshakeTimeout: 10 * time.Second,
+				TLSClientConfig:     &tls.Config{InsecureSkipVerify: true},
+			},
+		),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// return prometheus API
+	return prometheusv1.NewAPI(client), nil
+}


sanchezl:
I need to make this a "long running" test.

deads2k:
https://github.com/openshift/origin/blob/254adc6c36edf6fefb25e0bf20d7428aa4a8abb8/test/extended/prometheus/prometheus.go#L246-L247

deads2k:
creative.  I wonder if we'll get lucky and flake

/retest

test/extended/etcd/leader_changes.go
```diff
@@ -0,0 +1,91 @@
...
+			body, err := ioutil.ReadAll(response.Body)
+			o.Expect(err).ToNot(o.HaveOccurred())
+			result := map[string]interface{}{}
+			err = json.Unmarshal(body, &result)
+			o.Expect(err).ToNot(o.HaveOccurred())
+			g.By("Expect total number of ectd leadership changes seen to be less than 10")
+			totalLeaderChanges := result["data"].(map[string]interface{})["result"].([]interface{})[0].(map[string]interface{})["value"].([]interface{})[1]
+			e2e.Logf("sum(etcd_server_leader_changes_seen_total) = %v", totalLeaderChanges)
+			return strconv.Atoi(totalLeaderChanges.(string))
+		}, "5ms", "30s").Should(o.BeNumerically("<", 10))
```
sttts: please make tests readable

sanchezl: @sttts I have added some more comments. Let me know if its still unclear.

sanchezl: @sttts I've re-written it again. PTAL

mfojtik: @sttts it is pretty readable to me :-)

test/extended/etcd/leader_changes.go
```diff
@@ -0,0 +1,33 @@
...
+	o "github.com/onsi/gomega"
+	"github.com/prometheus/common/model"
+
+	"github.com/openshift/origin/test/extended/prometheus/client"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("etcd", func() {
+	defer g.GinkgoRecover()
+	var (
```
mfojtik: nit: make this one liner

test/extended/prometheus/client/via_route.go
```diff
@@ -0,0 +1,88 @@
...
+
+	routev1 "github.com/openshift/api/route/v1"
+
+	"github.com/openshift/origin/test/extended/util"
+)
+
+// NewE2EPrometheusRouterClient returns a Prometheus HTTP API client configured to
+// use the Prometheus route host, a bearer token, and no certificate verification.
+func NewE2EPrometheusRouterClient(oc *util.CLI) (prometheusv1.API, error) {
+	var err error
```
mfojtik: nit: remove and make the one in 32 `err := wait.PollImmediate(`

mfojtik:
/lgtm

We can iterate, this looks good and get us what we want.

/cc @hexfusion

test/extended/etcd/leader_changes.go
```diff
@@ -0,0 +1,33 @@
...
+	)
+
+	g.It("leader changes are not excessive", func() {
+		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		g.By("Watching the total number of etcd leadership changes for the next five minutes")
+		// check sum(etcd_server_leader_changes_seen_total) every 30s for 5m.
+		// the value should consistently be less than 10.
+		o.Consistently(func() model.SampleValue {
+			result, err := prometheus.Query(context.Background(), "sum(etcd_server_leader_changes_seen_total)", time.Now())
```
deads2k: `max(etcd_server_leader_changes_seen_total)` should equal one.  Per a buried thread in the incident channel.

sanchezl: Updated.

wking:
can we touch something to trigger an Azure rehearsal?  Azure disks usually give us more trouble, and we want the threshold to be high enough for them to get through without too many flakes.

test/extended/etcd/leader_changes.go
hexfusion: It’s important to understand this is optimal (max leader change 1) and only true to my knowledge with AWS. So we will most likely fail on most clouds. I think we will need a knob to tune this per cloud. So we can adjust per cloud or only run on AWS to watch for regression.

hexfusion:
> can we touch something to trigger an Azure rehearsal? Azure disks usually give us more trouble, and we want the threshold to be high enough for them to get through without too many flakes.

Exactly this will never pass azure see my note above on what is practical for this test.

wking:
Also in this space, we have an existing alert on this metric [in the cluster operator][1], which didn't fire because it's supposed to be:

```
increase((max(etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total{}))[15m:1m]) >= 1
```

or some such.  @smarterclayton has openshift/cluster-monitoring-operator#592 in flight to fix that, which may obsolete this test-suite effort (because we already have a test checking for surprising firing alerts)?

[1]: https://github.com/openshift/cluster-monitoring-operator/blob/d531ecb19943f445a3b8ed46d593dec4fd05674a/assets/prometheus-k8s/rules.yaml#L1368-L1376

wking:
/hold

To see if folks want this in addition to openshift/cluster-monitoring-operator#592 as a belt-and-suspenders, or if we're fine just fixing the existing alert.

hexfusion:
WRT alert. what is reasonable and actionable for admin in the wild is not the same metric we are tracking for regression. Meaning if 2/3 of clouds trigger alert and fail because of known storage we aren’t gaining anything as we already are watching p99 fsync. But if p99.9 elections for cloud A is N and test returns N > that  we do want to fail/track. 

The reason is the spike that triggers this can be short lived and missed in prom resolution as granularity of metrics is not that precise.

deads2k:
> /hold
> 
> To see if folks want this in addition to [openshift/cluster-monitoring-operator#592](https://github.com/openshift/cluster-monitoring-operator/pull/592) as a belt-and-suspenders, or if we're fine just fixing the existing alert.

clarified by @hexfusion . I'm glad the alert is fixed.  We should be better

/hold cancel

deads2k:
/lgtm cancel

@sanchezl fix the bit about per-cloud.

test/extended/etcd/leader_changes.go
sanchezl: I need some guidance on what the numbers should be for other clouds. 

@wking, if you have a suggestion for azure, I could at least start with that.

test/extended/etcd/leader_changes.go
```diff
@@ -0,0 +1,31 @@
...
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("etcd-leader-change", exutil.KubeConfigPath())
+
+	g.It("leader changes are not excessive", func() {
+		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		g.By("Watching the number of etcd leadership changes for the next five minutes")
+		// check max(etcd_server_leader_changes_seen_total) every 30s for 5m.
+		// the value should consistently be 1.
+		o.Consistently(func() model.SampleValue {
```
smarterclayton: This is really a post condition test, or a snapshot test.  Instead, you should be doing

```
increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[15m:1m]) > 0
```

smarterclayton: You shouldn't be waiting for 5 minutes.

sanchezl:
I have updated to use the query provided by @smarterclayton in https://github.com/openshift/origin/pull/24291#discussion_r359007429  . 

@hexfusion , @wking, this new query, looking at the rate of increase over the last five mionutes, should remove the need for cloud specific parameters.

test/extended/etcd/leader_changes.go
sanchezl: Updated

deads2k: > This is really a post condition test, or a snapshot test. Instead, you should be doing
> 
> ```
> increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[15m:1m]) > 0
> ```

I have nothing against it, but what does this say?  How can `0*` anything be not zero?

test/extended/prometheus/client/via_route.go
```diff
@@ -0,0 +1,87 @@
...
+			break
+		}
+	}
+	if secret == nil {
+		return nil, fmt.Errorf("unable to locate service prometheus service account secret")
+	}
+
+	// prometheus API client, configured for route host and bearer token auth, and no cert verification
+	client, err := api.NewClient(api.Config{
+		Address: "https://" + route.Spec.Host,
```
sallyom: This is preferred: https://github.com/openshift/library-go/pull/602#discussion_r362421809   
`route.Status.Ingress[0].Host`  
Note: In my PR I'm not doing any verifying/checking of that Ingress, I'm assuming the [0] one is good, rather than this:
https://github.com/openshift/library-go/blob/master/pkg/oauth/oauthserviceaccountclient/oauthclientregistry.go#L425-#L426

test/extended/etcd/leader_changes.go
```diff
@@ -0,0 +1,26 @@
...
+var _ = g.Describe("etcd", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("etcd-leader-change", exutil.KubeConfigPath())
+	g.It("leader changes are not excessive", func() {
+		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		g.By("Examining the rate of increase in the number of etcd leadership changes for last five minutes")
+		result, _, err := prometheus.Query(context.Background(), "increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[15m:1m])", time.Now())
+		o.Expect(err).ToNot(o.HaveOccurred())
+		o.Expect(result.(model.Vector)[0].Value).To(o.BeNumerically("==", 0))
```
sttts: @sanchezl nicely readable :+1:

test/extended/etcd/leader_changes.go
sttts: it is! Thanks :)

test/extended/etcd/leader_changes.go
sttts: Does this mean that the rate is zero?

sttts: And why only the first element?

sanchezl: Only one value is returned.

sttts: Still open: Does this mean that the rate is zero?

Everywhere else where this code is used, we compare with `3`, not zero.

sanchezl: The query was suggested in https://github.com/openshift/origin/pull/24291#discussion_r359007429.
I think this is saying the rate of increase over the last five minutes is `0`. 
This should take care of the issue brought up in https://github.com/openshift/origin/pull/24291#discussion_r358430475 where clusters on some cloud providers might have some different levels of etcd thrashing before settling down.  We use to compare the `etcd_server_leader_changes_seen_total` to `3`, but that value was only know to be appropriate on AWS (and even there, some preferred a stricter value of `1`).

sttts:
Based on https://github.com/openshift/origin/pull/24291#discussion_r369359229

/lgtm

PR #24296, opened 2019-12-13
UPSTREAM: 86177: make test framework easier to re-use

deads2k:
baseline to use to add more test error handling.  I'd like oc inspect output on all involved namespaces and all clusteroperators

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/framework/framework.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/framework/framework.go
@@ -31,6 +31,8 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


vendor/k8s.io/kubernetes/test/e2e/framework/framework.go
```diff
@@ -31,6 +31,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/runtime"
```
mfojtik: nit: move down

deads2k: > nit: move down

it's a pick from upstream

PR #24311, opened 2019-12-17
Bug 1690031: UPSTREAM: 86250: AWS: Don't report deletion of attached volume as warning

jsafrane:
https://bugzilla.redhat.com/show_bug.cgi?id=1690031
cc @openshift/storage


jsafrane:
/assign @knobunc 
for approval

PR #24312, opened 2019-12-17
Revert "Reenable 'should be rejected when no endpoints exist' test"

vrutkovs:
Reverts openshift/origin#23370

This test is broken on RHEL7 and OKD-on-FCOS, see https://bugzilla.redhat.com/show_bug.cgi?id=1781575


squeed:
This test has caught regressions in the past, so I'd rather keep it if possible.

vrutkovs:
Replaced by #24358

PR #23352, opened 2019-07-10
check ServiceCatalog basic usages in OCP 4.x

jianzhangbjz:
This PR will cover ServiceCatalog test cases: OCP-24062, OCP-24049, OCP-15600. @jmrodri @shawn-hurley @bparees Could you help have a review? Thanks! cc: @bandrade @scolange @cuipinghuo @chengzhang1016 @zihantang-rh @emmajiafan

PR #24315, opened 2019-12-17
Bug 1784334: Revert "check ServiceCatalog basic usages in OCP 4.x"

shawn-hurley:
Reverts openshift/origin#23352

/cc @smarterclayton @jianzhangbjz @jmrodri 

this test is causing consistent CI flakes and we should remove it.


shawn-hurley:
Disabling because:
Service Catalog will throw alerts once it is installed. 
Other tests will fail on those alerts. 

this will cause flakes in 4.3 and 4.4 tests we should move these to serial tests IIUC

jianzhangbjz:
@shawn-hurley Glad to know it. Thanks! I guess Service Catalog throw alerts only for 4.4. I was wondering if we can cherry-pick these automation tests to 4.1, 4.2, 4.3. And, in the future, can we contribute Service Catalog automation test cases to the 4.1, 4.2, 4.3 directly? No more submit PRs to the master branch.
One more question, will Service Catalog be removed from the 4.4?

jmrodri:
@jianzhangbjz my understanding is you should be able to submit a PR directly to the 4.3, 4.2, & 4.1 branches directly. The tests in 4.4 should actually be looking for service catalog to be removed. The 4.3 service catalog operator does create a prometheusrule and alerts.

jianzhangbjz:
> my understanding is you should be able to submit a PR directly to the 4.3, 4.2, & 4.1 branches directly. 

@jmrodri Thanks! But, I was told the workflow should be we submit the PR to the master branch, and then, the bot can cherry-pick it to the old versions. @bparees Can I submit the PR to the old version branch directly? Thanks!

wking:
Needs a rebase, @shawn-hurley

wking:
/retitle Bug 1784837: Revert "check ServiceCatalog basic usages in OCP 4.x"

wking:
/retitle Bug 1784334: Revert "check ServiceCatalog basic usages in OCP 4.x"

jmrodri:
Please close this PR in favor of PR #24323

bparees:
> @jmrodri Thanks! But, I was told the workflow should be we submit the PR to the master branch, and then, the bot can cherry-pick it to the old versions. @bparees Can I submit the PR to the old version branch directly? Thanks!

you can do it either way, but you must have merged to each prior branch before it can be merged to the next.  So you need to merge into master before your PR against release-4.3 will be merged (you can open it, but it will not be merged).  And then release-4.3 must merge before release-4.3, etc.

Also before you can merge to release-4.3 you need a BZ against 4.4 that is in "modified".
before you can merge to release-4.2 you need a BZ against 4.3.0 that is in "Verified"
before you can merge to release-4.1 you need a BZ against 4.2.z that is in "verified".

The patch manager+group leads will enforce those rules for release-4.3/release-4.2/release-4.1.

jianzhangbjz:
@bparees Thanks! Must we create a bug for the automation PR? It's a little bit of unreasonable. Do we consider a new workflow for automation?

> Service Catalog will throw alerts once it is installed.
Other tests will fail on those alerts.

@shawn-hurley Service Catalog will throw alerts in 4.4 as we designed. However, these alerts lead to other tests to fail. Actually, in my opinion, the customers will encounter this scenario too. I think we should find out the root causes, not remove Service Catalog tests simply. Or am I missing something?

bparees:
> @bparees Thanks! Must we create a bug for the automation PR? It's a little bit of unreasonable. Do we consider a new workflow for automation?

we require bugs for any merge into a non-master branch.  you can ask the patch-manager to override the requirement, but it's generally going to be faster to just create the bugs and follow the process than to argue for an exception.

And this should not be a normal thing (needing to merge automation/tests into non-master branches).  particularly once we put all the missing automation in place.

jianzhangbjz:
@bparees Many thanks for your explanation! Yeah, I see.

PR #24316, opened 2019-12-17
Disable tests causing e2e issues failures when run in parallel suite

kevinrizza:
Tests are flaky because they are run in parallel with tests that verify that no crash looping pods are running in openshift-* namespaces.  We either need to fix the tests to not launch crash looping pods, or we need to move them to another namespace, or we need to move these tests to the [Serial] suite.


emmajiafan:
I think I can move these tests to the [Serial] suite, since almost all the test cases in ns openshift-marketplace and openshift-operators.

PR #24318, opened 2019-12-18
[release-3.11] pkg/network: Fix pod startup reattach/kill loop

SriRamanujam:
This backports openshift/sdn commit
3101e183a787edfe6356765d93ae20786ca8635d which was merged in openshift/sdn#58.

Gonna be honest, I'm not sure if I did any of this correctly. The bug that led me to this MR only shows up occasionally in my environment, so I don't really have a proper build/test setup for OKD available. I couldn't even get the repo to build the artifacts locally, so I'm really flying blind here and hoping the test suite will catch anything wrong.

Feel free to close this if it's obvious to you guys that it's not going to work correctly. I just wanted to get the ball rolling on this and do what I could because I think this fixes my weird bug (if anybody's interested in that, it's openshift/sdn#80).

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -329,7 +329,7 @@ func (node *OsdnNode) Start() error {
 	if !node.useConnTrack {
 		node.watchServices()
 	}
-	existingSandboxPods, err := node.getPodSandboxes()
+	existingSandboxPods, err := node.getSDNPodSandboxes()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


squeed:
gofmt is failing:
```
./pkg/network/node/runtime.go
[FATAL] Try running 'gofmt -s -d [path]'
[FATAL] Or autocorrect with 'hack/verify-gofmt.sh | xargs -n 1 gofmt -s -w'
[ERROR] hack/verify-gofmt.sh exited with code 1 after 00h 00m 07s
```

danwinship:
the fix needs to be backport to every intermediate release, and the 3.11 version is going to need some additional backports
/close

PR #24319, opened 2019-12-18
build: disable timeouting test

mfojtik:
Attempt to mitigate https://bugzilla.redhat.com/show_bug.cgi?id=1784848


mfojtik:
/cc @adambkaplan 
/cc @bparees 

Guys, can you link the BZ here?

mfojtik:
/hold

https://github.com/openshift/origin/pull/24322 seems to be the fix

bparees:
/hold cancel

just in case my PR hits issues i'd rather unblock the org first and then fix the test later.

we can close this if my PR gets through first though.

PR #24320, opened 2019-12-18
e2e: disable check basic usages test for service catalog

mfojtik:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1784406


wking:
dup of #24315?

PR #24324, opened 2019-12-18
bug 1784880: fix up ruby-25-centos7 tag check logic

bparees:


bparees:
unblocks CI so cherry pick approving.

PR #24325, opened 2019-12-18
bug 1784881: fix up ruby-25-centos7 tag check logic

bparees:


bparees:
cherry pick approved to unblock CI and BZ approved on the basis that this can't merge w/o fixing the broken test.

PR #24326, opened 2019-12-18
bug 1784882: fix up ruby-25-centos7 tag check logic

bparees:


bparees:
overriding BZ requirement as this can't merge unless it fixes the BZ in question.
cherry pick approved to unblock CI

PR #24313, opened 2019-12-17
Revert "Bug 1763936: Disable node-Service and pod-Service granular checks"

smarterclayton:
Reverts openshift/origin#24185, insufficient justification provided for disablement.

PR #24330, opened 2019-12-19
Disable flaky tests on OpenStack

pierreprinetti:
In order to increase chances to get a voting CI for OpenStack soon, I
propose to skip Network Granular Check tests that are intermittently
failing on our infrastructure.

This commit must be reverted and the issue properly investigated once
the minimum result of a voting CI is achieved.

ref.: https://bugzilla.redhat.com/show_bug.cgi?id=1763936
ref.: https://github.com/openshift/origin/pull/24313

The bug reference is not in the PR title nor leading the commit message, because this proposed change does not fix it.

/assign @smarterclayton 
/cc @knobunc 

@danwinship Do you think this work for OpenStack? Sorry I am not familiar with the skip logic, and I could not find specific documentation on the matter.

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -516,6 +516,12 @@ var (
 			// should be serial if/when it's re-enabled
 			`\[HPA\] Horizontal pod autoscaling \(scale resource: Custom Metrics from Stackdriver\)`,
 		},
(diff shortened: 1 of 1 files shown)


smarterclayton:
This is fine for me

PR #24331, opened 2019-12-19
Bug 1785268: bump(*): pin ghodss/yaml

mfojtik:


sdodson:
/hold 
This needs to depend on a 4.3 bug rather than a 4.4 bug.

PR #24332, opened 2019-12-19
Bug 1785326: Disable build OOM test

adambkaplan:
RHCOS-related change broke the reason reported by the build container.
Disabling until it is fixed.


adambkaplan:
/hold

Fix has been tagged, but not pulled in to RHCOS/OCP yet.

smarterclayton:
Supposedly fixed in master 44.81.201912191631.0,

more 

/hold

just in case :)

adambkaplan:
/close

Obsolete - patch to RHCOS landed which fixed the underlying bug.

PR #24334, opened 2019-12-19
fix router ingress fixture to use real apigroup

deads2k:
found because it failed the kube-update job


smarterclayton:
/lgtm

too

deads2k:
/hold

might get insta-merge with kube bump

deads2k:
merged with kube bump

PR #24348, opened 2019-12-24
skip psp test consistently

p0lyn0mial:
This PR:
 - brings back disabled e2e test
 - changes Kube API server in a way that it stops serving `PodSecurityPolicies` resources.

Note:
We decided to disable PSP explicitly in the code instead of using `runtime-config` flag. At the moment the flag allows only for disabling the entire group, not individual resources. We don't want to disable the group as it contains critical resources like `PodDisruptionBudgets`.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/registry/policy/rest/storage_policy.go
+++ b/vendor/k8s.io/kubernetes/pkg/registry/policy/rest/storage_policy.go
@@ -55,11 +55,13 @@ func (p RESTStorageProvider) v1beta1Storage(apiResourceConfigSource serverstorag
 	storage["poddisruptionbudgets"] = poddisruptionbudgetStorage
 	storage["poddisruptionbudgets/status"] = poddisruptionbudgetStatusStorage
 
-	rest, err := pspstore.NewREST(restOptionsGetter)
-	if err != nil {
-		return storage, err
+	if apiResourceConfigSource.ResourceEnabled(policyapiv1beta1.SchemeGroupVersion.WithResource("podsecuritypolicies")) {
+		rest, err := pspstore.NewREST(restOptionsGetter)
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
I was hoping that disabling `PSP` would reveal tests that were messing with `PSP`. Previously, the disabled `PSP` test was occasionally enabled which initiated that there are some other tests that must be adding `PSP` resources to the system. Surprisingly no tests are complaining when `PSP` is disabled.

deads2k:
I would prefer to disable it via a flag so that our overall distro stays closer to upstream kube.

p0lyn0mial:
> I would prefer to disable it via a flag so that our overall distro stays closer to upstream kube.

We decided to disable PSP explicitly in the code instead of using runtime-config flag. At the moment the flag allows only for disabling the entire group, not individual resources. We don't want to disable the group as it contains critical resources like PodDisruptionBudgets.

vendor/k8s.io/kubernetes/pkg/registry/policy/rest/storage_policy.go
```diff
@@ -55,11 +55,13 @@ func (p RESTStorageProvider) v1beta1Storage(apiResourceConfigSource serverstorag
 	storage["poddisruptionbudgets"] = poddisruptionbudgetStorage
 	storage["poddisruptionbudgets/status"] = poddisruptionbudgetStatusStorage
 
-	rest, err := pspstore.NewREST(restOptionsGetter)
```
deads2k: In the quest for the simplest patch, is this actually just wrapping it in an `if false` block?

p0lyn0mial:
closing this pull as we decided to actually change/improve the e2e test instead.

/close

PR #24323, opened 2019-12-18
Bug 1784334: Revert "check ServiceCatalog basic usages for OCP 4.x"

jmrodri:
Supersedes PR #24315

PR #24349, opened 2019-12-25
Bug 1786409: Merge pull request #23352 from jianzhangbjz/service-catalog

jianzhangbjz:
check ServiceCatalog basic usages in OCP 4.1.

### DIFF
--- a/test/extended/operators/service_catalog.go
+++ b/test/extended/operators/service_catalog.go
@@ -1,0 +1,268 @@
+package operators
+
+import (
+	"fmt"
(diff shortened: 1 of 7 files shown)


eparis:
What are we doing here? just updating some tests in 4.1?  why?

jianzhangbjz:
@eparis In OCP 4.3 or 4.4, Service Catalog will throw alerts once it is installed. Other tests will fail on those alerts. So we remove these tests from the master branch(see #24323 #24315). But, we should test it in released 4.1.z, 4.2.z.

PR #24351, opened 2019-12-25
Exclude openshift-ovirt-infra from tolerations tests

rgolangh:
Exclude openshift-ovirt-infra namespaces when testing
"[Feature:Platform] Managed cluster should ensure control plane operators do not make themselves unevictable"


rgolangh:
Please review this

PR #24358, opened 2020-01-02
tests: skip tests based on worker OS

vrutkovs:
Some tests - e.g. one SDN test - doesn't pass on OKD-on-FCOS. This would 
ensure some tests can be skipped on working running Fedora.

/cc @smarterclayton

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -133,7 +133,7 @@ func ExecuteTest(t ginkgo.GinkgoTestingT, suite string) {
 }
 
 func AnnotateTestSuite() {
-	testRenamer := newGinkgoTestRenamerFromGlobals(e2e.TestContext.Provider, getNetworkSkips())
+	testRenamer := newGinkgoTestRenamerFromGlobals(e2e.TestContext.Provider, getNetworkSkips(), getOSSkips())
 
 	ginkgo.WalkTests(testRenamer.maybeRenameTest)
 }
@@ -147,7 +147,17 @@ func getNetworkSkips() []string {
 	return strings.Split(string(out), " ")
 }
 
-func newGinkgoTestRenamerFromGlobals(provider string, networkSkips []string) *ginkgoTestRenamer {
+func getOSSkips() []string {
+	out, err := e2e.KubectlCmd("get", "nodes", "--selector", "node-role.kubernetes.io/worker",
+		"--template", "{{ range .items }}{{index .metadata.labels \"node.openshift.io/os_id\"}} {{end}}").CombinedOutput()
+	if err != nil {
+		e2e.Logf("Could not get nodes list: not adding any plugin-specific skips")
+		return nil
+	}
+	return strings.Split(string(out), " ")
+}
+
+func newGinkgoTestRenamerFromGlobals(provider string, networkSkips []string, osSkips []string) *ginkgoTestRenamer {
 	var allLabels []string
 	matches := make(map[string]*regexp.Regexp)
 	stringMatches := make(map[string][]string)
@@ -180,6 +190,9 @@ func newGinkgoTestRenamerFromGlobals(provider string, networkSkips []string) *gi
 	}
 	for _, network := range networkSkips {
 		excludedTests = append(excludedTests, fmt.Sprintf(`\[Skipped:Network/%s\]`, network))
+	}
+	for _, os := range osSkips {
+		excludedTests = append(excludedTests, fmt.Sprintf(`\[Skipped:OS/%s\]`, os))
 	}
 	klog.V(4).Infof("openshift-tests excluded test regex is %q", strings.Join(excludedTests, `|`))
 	excludedTestsFilter := regexp.MustCompile(strings.Join(excludedTests, `|`))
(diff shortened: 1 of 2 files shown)


test/extended/util/test.go
```diff
@@ -147,7 +147,17 @@ func getNetworkSkips() []string {
 	return strings.Split(string(out), " ")
 }
 
-func newGinkgoTestRenamerFromGlobals(provider string, networkSkips []string) *ginkgoTestRenamer {
+func getOSSkips() []string {
+	out, err := e2e.KubectlCmd("get", "nodes", "--selector", "node-role.kubernetes.io/worker",
```
smarterclayton: No, you can’t run a command to get this info.

Detection would have to be done when the cloud provider info is filled out by reading from a key object.

It’s going to be better overall to just fix agnhost since this is a pretty key test and a future version of rhel could break.  As Casey says the error message is wrong, and that’s an indicator something about the environment is impacting the test.

vrutkovs: >Detection would have to be done when the cloud provider info is filled out by reading from a key object.

So read the AMI info instead? There is no other nice way to distinct OKD vs. OCP payload afaik.

smarterclayton: The info about the os should be on nodes.

I’m still inclined to say we need to fix the test.

vrutkovs: `os_id` label is set by MCO (as a [kubelet param](https://github.com/openshift/machine-config-operator/blob/master/templates/worker/01-worker-kubelet/_base/units/kubelet.yaml#L24)) so its seems to be a safe way of find out worker's OS.

Test definitely needs fixing, but its doesn't look its gonna get attention from SDN team :/

danwinship: > It’s going to be better overall to just fix agnhost

Casey's claim that it was agnhost's fault was wrong. Agnhost is correctly reporting that the test is not getting the expected result.

Vadim said on Slack that this is only failing on RHEL 7? In that case, I think it's a problem with RHEL 7 / legacy iptables mode then; and that explains why it was that it suddenly started working for us even though we hadn't changed anything. It didn't pass in 3.x but we thought it was because of openshift-sdn problems, so we had disabled it. But then when we tried re-enabling the test in 4.2, it just worked without us needing to change anything and we didn't know why. But it makes sense if the thing that changed was that OCP went from RHEL 7 to RHEL 8 underneath.

vrutkovs: This is reproducible on RHEL7 workers of OCP and Fedora CoreOS 31 on OKD.

I tried switching to nft tables on OKD ([replaced symlinks](https://github.com/openshift/release/commit/7e4dd6fcfa19bd537f56cb1c3139b84065158d9c)) but that didn't help much. It appears RHCOS (or just RHEL8) has some other secret fix / change

danwinship: I guess it doesn't really make sense that the difference would be iptables-legacy vs iptables-nft since we know that the test passes in upstream e2e with iptables-legacy.

And Fedora 31 has openvswitch 2.11 (which is what we're currently using in OCP) so it can't be OVS either...

LorbusChris: I'll try to summarize this issue:
The test was disabled because it wasn't passing.
Then _something_ happened and it started to pass again on RHCOS - however nobody seems to be able to pinpoint to what change exactly caused this.
The test was re-enabled, but continues to fail on RHEL7 and FCOS nodes.
Vadim and I have tried to solve this by making FCOS use iptables-nft instead of -legacy, but to no avail.
The SDN team is busy and won't have time to look into it this sprint.
This is now the main blocker for OKD 4 Beta 1.

I'd really like to move past this asap and disable the test temporarily until the proper solution is found, so we can get the OKD4 Beta out this week. 

@smarterclayton, the `getNetworkSkips` func above works similarly, and I agree with @vrutkovs that it seems to be a safe way to get the OS id from the node. If this is not the right way to exclude the test on FCOS, could you provide some more details on how we should do it instead?

smarterclayton: First, getNetworkSkips is also broken in the same way.  You cannot invoke Kubectl command at skip generation time.  All lookups like this need to be part of the default provider initialization, and use API calls to fetch info (if needed).

Second, this test is fundamental to how services are supposed to work (rejecting connections when there are no backends).  If it's broken, it's as if "pods don't start", and thus must be fixed.  A cluster that doesn't pass this isn't conformant to a Kube cluster.  It was temporarily skipped for OVN while they provided a fix, and should have already been removed.

vrutkovs: >If it's broken, it's as if "pods don't start", and thus must be fixed

Its rather "a different error message is thrown when a service with endpoints is used". It complicates the debugging, but its not as bad as "SDN doesn't route packages at all"

LorbusChris: Can we not skip this test exclusively on FCOS for now and make fixing this (as well as the getXSkips funcs) a blocker for OKD 4 GA? The change would only affect the OKD4 Beta, and with it being a Beta, I think we can get away with it not being entirely conformant in that regard.

danwinship: > Second, this test is fundamental to how services are supposed to work (rejecting connections when there are no backends). If it's broken, it's as if "pods don't start", and thus must be fixed.

The test name/description is somewhat misleading; what it's testing is "connections to a service with no backends are rejected immediately with 'connection refused' rather than timing out after several seconds with 'no route to host'". Everything still gets the expected results even when this test fails, it's just that when things are broken, clients will take longer to notice.

(This test was only added a year ago, at which point it was discovered that the "immediately reject" feature had been broken several months earlier in a kube-proxy refactoring, and no one had noticed.)

> A cluster that doesn't pass this isn't conformant to a Kube cluster.

This is not a `[Conformance]` test. (And before this test was added all of the other e2e tests had been passing for months while this feature was broken).

> It was temporarily skipped for OVN while they provided a fix, and should have already been removed.

It is still disabled for ovn-kubernetes (though it's expected to be fixed for 4.4).

PR #24363, opened 2020-01-06
Allow for customization of storageClass in the CSI test suite

bertinatto:
This patch is the result of running the CSI certification process outlined [here](https://github.com/openshift/enhancements/pull/162/files) against the EBS CSI driver. In other words, we need this change so partners can test and certify their CSI drivers with OpenShift.

We also want to use the EBS CSI driver in our CI, so getting it to work is the first step.

CC @openshift/storage

### DIFF
--- a/cmd/openshift-tests/csi.go
+++ b/cmd/openshift-tests/csi.go
@@ -3,10 +3,12 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/openshift/origin/test/extended/csi"
 
+	"k8s.io/kubernetes/test/e2e/framework/testfiles"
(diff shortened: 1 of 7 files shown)


vendor/k8s.io/kubernetes/test/e2e/storage/external/external.go.orig
```diff
@@ -0,0 +1,329 @@
+/*
```
jsafrane: external.go.orig file should be removed.

jsafrane:
/hold

for https://github.com/kubernetes/kubernetes/pull/86912 to be merged upstream first

bertinatto:
/assign @knobunc @mfojtik 
for approval

jsafrane:
/hold cancel
kubernetes/kubernetes#86912 is not needed in the end

PR #24367, opened 2020-01-07
improves verify-generated-bindata to verify also extended test bind data

p0lyn0mial:


hack/verify-generated-bindata.sh
```diff
@@ -13,7 +13,9 @@ trap "cleanup" EXIT
 TMP_GENERATED_BOOTSTRAP_DIR="_output/verify-bindata"
 
 os::test::junit::declare_suite_start "verify/bindata"
-os::cmd::expect_success "OUTPUT_ROOT=${TMP_GENERATED_BOOTSTRAP_DIR} ${OS_ROOT}/hack/update-generated-bindata.sh"
+os::cmd::expect_success "${OS_ROOT}/hack/update-generated-bindata.sh"
```
tnozicka: why is it dropping the TMP_GENERATED_BOOTSTRAP_DIR reference? and where is the TMP_GENERATED_BOOTSTRAP_DIR used after that?

I am not familiar with these scripts though, maybe @sttts would be a better choice

sttts: I wished I were familiar with them ;-) @p0lyn0mial what's the rational here?

p0lyn0mial: Dropping `TMP_GENERATED_BOOTSTRAP_DIR` enables the script to change the original file and diffing its content. 

Previously a new file was created in the directory pointing by `TMP_GENERATED_BOOTSTRAP_DIR` each time this script was executed.

p0lyn0mial: It seems like `TMP_GENERATED_BOOTSTRAP_DIR` is not used anymore and can be removed

sttts: this means that the checkout is dirty when there is a change?

p0lyn0mial: yes, exactly the same as for `vendor/k8s.io/kubernetes/test/e2e/generated/bindata.go` and `vendor/k8s.io/kubernetes/staging/src/k8s.io/kubectl/pkg/generated/bindata.go`

tnozicka: anything named verify should never change the original file
/hold

tnozicka: I suppose the rest of this script using git diff already breaks that invariant :( so
/hold cancel

PR #24378, opened 2020-01-08
 Bug 1782467: pin image-eco sclorg samples usage to specific versions to shield fro…

gabemontero:
…m upstream churn wrt EOL image removals

### DIFF
--- a/test/extended/image_ecosystem/s2i_ruby.go
+++ b/test/extended/image_ecosystem/s2i_ruby.go
@@ -19,7 +19,7 @@ import (
 var _ = g.Describe("[image_ecosystem][ruby][Slow] hot deploy for openshift ruby image", func() {
 	defer g.GinkgoRecover()
 	var (
-		railsTemplate = "https://raw.githubusercontent.com/openshift/rails-ex/master/openshift/templates/rails-postgresql.json"
+		railsTemplate = "rails-postgresql-example"
 		oc            = exutil.NewCLI("s2i-ruby", exutil.KubeConfigPath())
 		modifyCommand = []string{"sed", "-ie", `s%render :file => 'public/index.html'%%`, "app/controllers/welcome_controller.rb"}
 		removeCommand = []string{"rm", "-f", "public/index.html"}
@@ -46,8 +46,8 @@ var _ = g.Describe("[image_ecosystem][ruby][Slow] hot deploy for openshift ruby 
 			g.It(fmt.Sprintf("should work with hot deploy"), func() {
 
 				exutil.WaitForOpenShiftNamespaceImageStreams(oc)
-				g.By(fmt.Sprintf("calling oc new-app -f %q", railsTemplate))
-				err := oc.Run("new-app").Args("-f", railsTemplate).Execute()
+				g.By(fmt.Sprintf("calling oc new-app %q", railsTemplate))
+				err := oc.Run("new-app").Args(railsTemplate).Execute()
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 				g.By("waiting for build to finish")
--- a/test/extended/image_ecosystem/sample_repos.go
+++ b/test/extended/image_ecosystem/sample_repos.go
@@ -55,7 +55,8 @@ func NewSampleRepoTest(c sampleRepoConfig) func() {
 					err := exutil.WaitForOpenShiftNamespaceImageStreams(oc)
 					o.Expect(err).NotTo(o.HaveOccurred())
 					g.By(fmt.Sprintf("calling oc new-app with the " + c.repoName + " example template"))
-					err = oc.Run("new-app").Args("-f", c.templateURL).Execute()
+					newAppArgs := []string{c.templateURL}
+					err = oc.Run("new-app").Args(newAppArgs...).Execute()
 					o.Expect(err).NotTo(o.HaveOccurred())
 
 					// all the templates automatically start a build.
@@ -121,7 +122,7 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
 	g.Describe("[image_ecosystem][ruby] test ruby images with rails-ex db repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "rails-postgresql",
-			templateURL:            "https://raw.githubusercontent.com/openshift/rails-ex/master/openshift/templates/rails-postgresql.json",
+			templateURL:            "rails-postgresql-example",
 			buildConfigName:        "rails-postgresql-example",
 			serviceName:            "rails-postgresql-example",
 			deploymentConfigName:   "rails-postgresql-example",
@@ -135,7 +136,7 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
 	g.Describe("[image_ecosystem][python] test python images with django-ex db repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "django-psql",
-			templateURL:            "https://raw.githubusercontent.com/openshift/django-ex/master/openshift/templates/django-postgresql.json",
+			templateURL:            "django-psql-example",
 			buildConfigName:        "django-psql-example",
 			serviceName:            "django-psql-example",
 			deploymentConfigName:   "django-psql-example",
@@ -149,7 +150,7 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
 	g.Describe("[image_ecosystem][nodejs] test nodejs images with nodejs-ex db repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "nodejs-mongodb",
-			templateURL:            "https://raw.githubusercontent.com/openshift/nodejs-ex/master/openshift/templates/nodejs-mongodb.json",
+			templateURL:            "nodejs-mongodb-example",
 			buildConfigName:        "nodejs-mongodb-example",
 			serviceName:            "nodejs-mongodb-example",
 			deploymentConfigName:   "nodejs-mongodb-example",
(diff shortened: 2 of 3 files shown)


gabemontero:
/assign @bparees 
/assign @adambkaplan 

whoever sees first ... per the discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1782467#c14 here is the image_eco e2e changes for pinning down versions when using upstream sclorg templates

test/extended/image_ecosystem/s2i_ruby.go
```diff
@@ -19,7 +19,7 @@ import (
 var _ = g.Describe("[image_ecosystem][ruby][Slow] hot deploy for openshift ruby image", func() {
 	defer g.GinkgoRecover()
 	var (
-		railsTemplate = "https://raw.githubusercontent.com/openshift/rails-ex/master/openshift/templates/rails-postgresql.json"
+		railsTemplate = "https://raw.githubusercontent.com/sclorg/rails-ex/master/openshift/templates/rails-postgresql.json"
```
bparees: would it be better to pull this from openshift/library which we at least sort of control?

bparees: (or even openshift samples operator's repo, if we're shipping this template?)

bparees: looks like we are shipping it....in which case you don't even need to get it from the repo, it's available in the cluster (as installed by the samples operator)

test/extended/image_ecosystem/sample_repos.go
```diff
@@ -121,56 +132,68 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
 	g.Describe("[image_ecosystem][ruby] test ruby images with rails-ex db repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "rails-postgresql",
-			templateURL:            "https://raw.githubusercontent.com/openshift/rails-ex/master/openshift/templates/rails-postgresql.json",
+			templateURL:            "https://raw.githubusercontent.com/sclorg/rails-ex/master/openshift/templates/rails-postgresql.json",
```
bparees: same comment about template source

test/extended/image_ecosystem/sample_repos.go
```diff
@@ -121,56 +132,68 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
...
 			dbServiceName:          "postgresql",
+			//TODO upstream template does not expose lang/db version params yet
 		},
 	))
 
 	g.Describe("[image_ecosystem][python] test python images with django-ex db repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "django-psql",
-			templateURL:            "https://raw.githubusercontent.com/openshift/django-ex/master/openshift/templates/django-postgresql.json",
+			templateURL:            "https://raw.githubusercontent.com/sclorg/django-ex/master/openshift/templates/django-postgresql.json",
```
bparees: same comment about template source

test/extended/image_ecosystem/sample_repos.go
```diff
@@ -121,56 +132,68 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
...
+			dbVersion:              "10",
+			dbVersionParameter:     "POSTGRESQL_VERSION",
 		},
 	))
 
 	g.Describe("[image_ecosystem][nodejs] test nodejs images with nodejs-ex db repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "nodejs-mongodb",
-			templateURL:            "https://raw.githubusercontent.com/openshift/nodejs-ex/master/openshift/templates/nodejs-mongodb.json",
+			templateURL:            "https://raw.githubusercontent.com/sclorg/nodejs-ex/master/openshift/templates/nodejs-mongodb.json",
```
bparees: same comment about template source

test/extended/image_ecosystem/sample_repos.go
```diff
@@ -121,56 +132,68 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
...
+			dbVersion:              "3.6",
+			dbVersionParameter:     "MONGODB_VERSION",
 		},
 	))
 
 	var _ = g.Describe("[image_ecosystem][php] test php images with cakephp-ex db repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "cakephp-mysql",
-			templateURL:            "https://raw.githubusercontent.com/openshift/cakephp-ex/master/openshift/templates/cakephp-mysql.json",
+			templateURL:            "https://raw.githubusercontent.com/sclorg/cakephp-ex/master/openshift/templates/cakephp-mysql.json",
```
bparees: same comment about template source

test/extended/image_ecosystem/sample_repos.go
```diff
@@ -193,15 +216,16 @@ var _ = g.Describe("[image_ecosystem][Slow] openshift sample application reposit
 	g.Describe("[image_ecosystem][python] test python images with django-ex repo", NewSampleRepoTest(
 		sampleRepoConfig{
 			repoName:               "django",
-			templateURL:            "https://raw.githubusercontent.com/openshift/django-ex/master/openshift/templates/django.json",
+			templateURL:            "https://raw.githubusercontent.com/sclorg/django-ex/master/openshift/templates/django.json",
```
bparees: same comment about template source

test/extended/image_ecosystem/s2i_ruby.go
gabemontero: I debated making that change, but  wavered on the value of some element of upstream verification.

This was sort of a meet in the middle sort of compromise.  No longer the wild wild west but some eye to helping to keep upstream honest.

Thoughts @bparees ?  I'll admit I don't feel super strongly about it though.

bparees: moving to library probably doesn't have a huge benefit. moving to using the in-cluster template would make the test more robust since it wouldn't be at risk of:

1) github flakes
2) upstream merges that break the test and block the queue

i don't feel that strongly about it though (you guys own the test and will be dealing w/ the fall out if it starts breaking).

adambkaplan: +1 to using the template on the cluster. In addition to the benefits @bparees mentioned, this lets us catch breaking changes when we bump the samples operator.

Hypothetically we could then create a job that regularly bumps the samples from openshift/library and opens a PR for us to review.

gabemontero:
@adambkaplan @bparees moved to leveraging the templates installed via samples operator for image-eco

bparees:
the template usage lgtm, but don't you still need to pass the version pinning param values?

gabemontero:
> the template usage lgtm, but don't you still need to pass the version pinning param values?

My thought was "no" because if the template we install does not work out of the gate, we have a problem.  So if upstream and openshift/library has disconnects, the reality is we have to patch the template in the samples operator after importing from openshift/library to address the disconnect, where we essentially pin the version there, while we work with upstream.

We already have had a case or two where I've had to do such things.

Otherwise, the versions in the older release templates should point to older versions.

This way will also make the cherry pick back to 4.3 and 4.2 tenable.  Where as if we cite versions, the versions can vary between 4.2, 4.3, 4.4, etc.

bparees:
> My thought was "no" because if the template we install does not work out of the gate, we have a problem. So if upstream and openshift/library has disconnects, the reality is we have to patch the template in the samples operator after importing from openshift/library to address the disconnect, where we essentially pin the version there, while we work with upstream.

ah, yeah that makes sense.

gabemontero:
/hold

CI is struggling all afternoon. Will unhold this evening an see if it is healthier.

gabemontero:
sorry @bparees had a gofmt error ... can you re-lgtm ?

gabemontero:
I've engaged on slack with the e2e anomalies but no response yet

see if we get lucky

/hold cancel

/retest

adambkaplan:
/hold

image-eco seems to be failing hard.

gabemontero:
yep there is a hiccup with the switch to the installed templates @adambkaplan thanks

I was missing it while chasing the CI problems in my other PRs today.

gabemontero:
`oc new-app` is not finding templates in the `openshift` namespace:

```
Cmd: "oc --namespace=e2e-test-nodejs-repo-test-nvk6q --config=/tmp/configfile396527657 new-app nodejs-example",
        StdErr: "error:  local file access failed with: stat nodejs-example: no such file or directory\nerror: unable to locate any images in image streams, templates loaded in accessible projects, template files, local docker images with name \"nodejs-example\"\n\nArgument 'nodejs-example' was classified as an image, image~source, or loaded template reference.\n\nThe 'oc new-app' command will match arguments to the following types:\n\n  1. Images tagged into image streams in the current project or the 'openshift' project\n     - if you don't specify a tag, we'll add ':latest'\n  2. Images in the Docker Hub, on remote registries, or on the local Docker engine\n  3. Templates in the current project or the 'openshift' project\n  4. Git repository URLs or local paths that point to Git repositories\n\n--allow-missing-images can be used to point to an image that does not exist yet.\n\nSee 'oc new-app -h' for examples.",
```

I thought I double-checked the names but we'll redo that.  Or we have a `oc new-app` bug.

bparees:
I'm OK dropping tests for templates we no longer ship. Surprised we don't ship those 3 though. Is it because we ship the versions that include databases?

gabemontero:
green tests gentlemen

gabemontero:
> I'm OK dropping tests for templates we no longer ship. Surprised we don't ship those 3 though. Is it because we ship the versions that include databases?

Yes that is my recollection @bparees

gabemontero:
iam / gce flakes 

/retest

PR #24391, opened 2020-01-13
Bug 1782467: move more image-eco db template tests from upstream repos to sample o…

gabemontero:
…perator installed

/assign @bparees 
/assign @adambkaplan

### DIFF
--- a/test/extended/image_ecosystem/mariadb_ephemeral.go
+++ b/test/extended/image_ecosystem/mariadb_ephemeral.go
@@ -12,7 +12,7 @@ import (
 var _ = g.Describe("[image_ecosystem][mariadb][Slow] openshift mariadb image", func() {
 	defer g.GinkgoRecover()
 	var (
-		templatePath = exutil.FixturePath("..", "..", "examples", "db-templates", "mariadb-ephemeral-template.json")
+		templatePath = "mariadb-ephemeral"
 		oc           = exutil.NewCLI("mariadb-create", exutil.KubeConfigPath())
 	)
 
@@ -33,8 +33,8 @@ var _ = g.Describe("[image_ecosystem][mariadb][Slow] openshift mariadb image", f
 			g.It(fmt.Sprintf("should instantiate the template"), func() {
 				exutil.WaitForOpenShiftNamespaceImageStreams(oc)
 
-				g.By(fmt.Sprintf("calling oc process -f %q", templatePath))
-				configFile, err := oc.Run("process").Args("-f", templatePath).OutputToFile("config.json")
+				g.By(fmt.Sprintf("calling oc process %q", templatePath))
+				configFile, err := oc.Run("process").Args("openshift//" + templatePath).OutputToFile("config.json")
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 				g.By(fmt.Sprintf("calling oc create -f %q", configFile))
--- a/test/extended/image_ecosystem/mysql_ephemeral.go
+++ b/test/extended/image_ecosystem/mysql_ephemeral.go
@@ -12,7 +12,7 @@ import (
 var _ = g.Describe("[image_ecosystem][mysql][Slow] openshift mysql image", func() {
 	defer g.GinkgoRecover()
 	var (
-		templatePath = exutil.FixturePath("..", "..", "examples", "db-templates", "mysql-ephemeral-template.json")
+		templatePath = "mysql-ephemeral"
 		oc           = exutil.NewCLI("mysql-create", exutil.KubeConfigPath())
 	)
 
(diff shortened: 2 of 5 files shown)


gabemontero:
```
Cmd: "oc --namespace=e2e-test-mariadb-create-txmwv --config=/tmp/configfile103497059 process mariadb-ephemeral -n openshift",
        StdErr: "error: unable to process template\n  processedtemplates.template.openshift.io is forbidden: User \"e2e-test-mariadb-create-txmwv-user\" cannot create resource \"processedtemplates\" in API group \"template.openshift.io\" in the namespace \"openshift\"",
```

and 

```
Cmd: "oc --namespace=e2e-test-mysql-create-qnpsw --config=/tmp/configfile840281241 process mysql-ephemeral -n openshift",
        StdErr: "error: unable to process template\n  processedtemplates.template.openshift.io is forbidden: User \"e2e-test-mysql-create-qnpsw-user\" cannot create resource \"processedtemplates\" in API group \"template.openshift.io\" in the namespace \"openshift\"",
```
```

**Comment by gabemontero** [comment]
fyi @akram @waveywaves   ... as part of removing EOL SCL images (like nodejs:8) I'm updating with this PR one of the jenkins test template to leverage the template we install as part of openshift install vs. the template as it exists in the upstream SCL github repo

aside from better testing what openshift install's, we've seen enough problems with leveraging upstream SCL content that we have started systematically moving off of it.  This becomes even more paramount in support existing releases, as they won't be subject to upstream changes.

**Comment by gabemontero** [comment]
Hmm ... perhaps a new underlying change is now impacting us.  Looks like the scl tests as is can no longer create pods:

```
            Status: "Failure",
            Message: "pods \"test-pod-c7f22e7d-e000-4637-bb97-0bffa7775534\" is forbidden: unable to validate against any security context constraint: []",
            Reason: "Forbidden",
```

will change them to run as admin and see what happens

**Comment by gabemontero** [comment]
ok the permission failure on the scl enable tests were flakes

we have green image-ecosystem tests @bparees @adambkaplan  ... if one of you could lgtm, with retests to move past the infra/unrelated component flakes in the other tests, we can get this merged

File: test/extended/image_ecosystem/mariadb_ephemeral.go [hunk 1]
```diff
@@ -33,8 +33,8 @@ var _ = g.Describe("[image_ecosystem][mariadb][Slow] openshift mariadb image", f
 			g.It(fmt.Sprintf("should instantiate the template"), func() {
 				exutil.WaitForOpenShiftNamespaceImageStreams(oc)
 
-				g.By(fmt.Sprintf("calling oc process -f %q", templatePath))
-				configFile, err := oc.Run("process").Args("-f", templatePath).OutputToFile("config.json")
+				g.By(fmt.Sprintf("calling oc process %q", templatePath))
+				configFile, err := oc.AsAdmin().Run("process").Args(templatePath, "-n", "openshift").OutputToFile("config.json")
```
bparees: why did this need to switch to AsAdmin?

test/extended/image_ecosystem/mysql_ephemeral.go
```diff
@@ -31,8 +31,8 @@ var _ = g.Describe("[image_ecosystem][mysql][Slow] openshift mysql image", func(
 		g.Describe("Creating from a template", func() {
 			g.It(fmt.Sprintf("should instantiate the template"), func() {
 
-				g.By(fmt.Sprintf("calling oc process -f %q", templatePath))
-				configFile, err := oc.Run("process").Args("-f", templatePath).OutputToFile("config.json")
+				g.By(fmt.Sprintf("calling oc process %q", templatePath))
+				configFile, err := oc.AsAdmin().Run("process").Args(templatePath, "-n", "openshift").OutputToFile("config.json")
```
bparees: same question on this one

test/extended/image_ecosystem/mariadb_ephemeral.go
gabemontero: see https://github.com/openshift/origin/pull/24391#issuecomment-574193003 ... I started getting those permission errors consistently when the `oc process` calls tried to access templates in the openshift namespace instead of access a local file like was done previously.

Conversely, `oc new-app` did not get those errors (you'll see uses of that in this file that did not have to change).

The help for `oc process` does mention you can use `oc process openshift//foo` to access templates in the openshift namespace.  

But admittedly, I went for the path of least resistance at this point, given the delays in trying to get this in between CI infra flakes and 2nd and 3rd level discoveries of accessing upstream templates.

So we either:
a) go with this short cut
b) try `oc process openshift//` and see what happens
c) investigate the `oc` nuances (my least favorite option ;-) )

test/extended/image_ecosystem/mysql_ephemeral.go
gabemontero: same answer

test/extended/image_ecosystem/mariadb_ephemeral.go
gabemontero: I'll try b)

bparees: ah yes, i understand the issue i think.  It does look like "openshift//templatename" with no `-n` argument should work.

bparees: the reason is that "oc process" posts a "processtemplate" request object to whatever namespace you pass with "-n".  And regular users can't create that request object in the "openshift" namespace.  that's why the namespace//templatename syntax exists, so you say "get the template from namespace X, but process it in namespace Y"

gabemontero: Interestingly  .... `oc adm policy who-can process templates -n openshift ` and `oc adm policy who-can get templates -n openshift ` return different results

gabemontero: ah gotcha

gabemontero:
updated @bparees

gabemontero:
ok image-eco is green ... just circumnavigating non-related flakes now

PR #24393, opened 2020-01-14
UPSTREAM: <carry>: don't add service-ca to SA tokens

stlaz:
This PR is here for 4.5 where the `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` will be remove as announced in 4.1 release notes.

https://github.com/openshift/openshift-docs/issues/18426#issuecomment-574239835

/cc @deads2k @marun 

/hold


marun:
@stlaz As per [this bz](https://bugzilla.redhat.com/show_bug.cgi?id=1813894) this PR is deferred to 4.6.

stlaz:
> @stlaz As per [this bz](https://bugzilla.redhat.com/show_bug.cgi?id=1813894) this PR is deferred to 4.6.
> 
> Edit: Would it make sense to remove `UPSTREAM: <carry>: kube-controller-manager: add service serving cert signer to token controller` as well?

Yes, we should do that as well to clear this all off

marun:
afaik we can't disable this since we can't prove it won't negatively impact customers.

slaskawi:
@stlaz @marun Keycloak Operator uses the `service-ca.crt`, so I'm very interested if this PR got descoped for good or it was just delayed to another release?

stlaz:
@slaskawi it did, but I hope we'll get to do it one day because the amount of unused CA certs that take up space in etcd for no reason is ridiculous.

You can just do https://docs.openshift.com/container-platform/4.5/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate and mount the CM to your pod instead.

slaskawi:
Thanks for the hint @stlaz !

PR #24400, opened 2020-01-15
Bug 1778856: Add MIN/MAX variables to SSL/TLS versions

briantward:
Add same variables as in OCP 4.x routers to support Min and Max TLS versions:

Related 4.x activity:
https://github.com/openshift/router/pull/30
https://bugzilla.redhat.com/show_bug.cgi?id=1746467

RFE on 3.11:
https://issues.redhat.com/browse/RFE-167

I've set the default for 3.11 to TLSv1.0 since that was the previous setting without these variables.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -57,8 +57,9 @@ global
   tune.maxrewrite 8192
   tune.bufsize 32768
 
-  # Prevent vulnerability to POODLE attacks
-  ssl-default-bind-options no-sslv3
+  # Configure the TLS versions we support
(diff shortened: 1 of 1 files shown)


briantward:
What is not tested, and I don't know about, is whether or not any 3.11 components require TLS 1.0 or 1.1.

ehashman:
Both look like flakes

/retest

briantward:
@knobunc @pecameron is this something you can take a look at? I'm not sure why gcp tests were failing last. I can rebase this if necessary since it has been a few weeks.

ehashman:
@briantward the GCP test failure just looks like a failure to report/timeout (all tests pass but build killed after 4h)

ehashman:
https://bugzilla.redhat.com/show_bug.cgi?id=1571401 has been reopened to cover this.

PR #24411, opened 2020-01-17
Bug 1792569: UPSTREAM: : admission/restrictusers ensure groups cache synced

sanchezl:
The restrictusers admission plugin uses a groups.user.openshift.io informer without waiting for its cache to sync. This results in misleading permission denials when cache is not synced.

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/authorization/restrictusers/subjectchecker.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/authorization/restrictusers/subjectchecker.go
@@ -2,10 +2,12 @@ package restrictusers
 
 import (
 	"fmt"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/pkg/apis/rbac"
 
@@ -102,6 +104,13 @@ func (ctx *RoleBindingRestrictionContext) labelSetForUser(subject rbac.Subject) 
 func (ctx *RoleBindingRestrictionContext) groupsForUser(subject rbac.Subject) ([]*userv1.Group, error) {
 	if subject.Kind != rbac.UserKind {
 		return []*userv1.Group{}, fmt.Errorf("not a user: %q", subject.Name)
+	}
+
+	err := wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		return ctx.groupCache.HasSynced(), nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("groups.user.openshift.io cache is not synchronized")
 	}
 
 	return ctx.groupCache.GroupsFor(subject.Name)
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/authorization/restrictusers/usercache/groups.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/authorization/restrictusers/usercache/groups.go
@@ -12,7 +12,8 @@ import (
 // GroupCache is a skin on an indexer to provide the reverse index from user to groups.
 // Once we work out a cleaner way to extend a lister, this should live there.
 type GroupCache struct {
-	indexer cache.Indexer
+	indexer      cache.Indexer
+	groupsSynced cache.InformerSynced
 }
 
 const ByUserIndexName = "ByUser"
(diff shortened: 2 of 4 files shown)


vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/authorization/restrictusers/usercache/groups.go
```diff
@@ -31,10 +34,17 @@ func ByUserIndexKeys(obj interface{}) ([]string, error) {
 func NewGroupCache(groupInformer userinformer.GroupInformer) *GroupCache {
 	return &GroupCache{
 		indexer: groupInformer.Informer().GetIndexer(),
+		groupsSynced: groupInformer.Informer().HasSynced,
 	}
 }
 
 func (c *GroupCache) GroupsFor(username string) ([]*userapi.Group, error) {
+	err := wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
```
deads2k: wrong spot. You can track the `groupsSynced` in this struct, but leave the handling up to the callers of this method.  I doubt they will be uniform.  Admission handlers may choose to block on partial lists and then fail. Authentication is likely to block and then make the best choice it can.

sanchezl: @deads2k , please review update. I also have https://github.com/openshift/origin/pull/24414 for the authentication use of this utility.

deads2k:
@sanchezl fix the import

/lgtm
/hold

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/authorization/restrictusers/subjectchecker.go
```diff
@@ -104,6 +106,13 @@ func (ctx *RoleBindingRestrictionContext) groupsForUser(subject rbac.Subject) ([
 		return []*userv1.Group{}, fmt.Errorf("not a user: %q", subject.Name)
 	}
 
+	err := wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
```
sttts: I would rather expect this much higher up in the outer plumbing code that sets up the group cache, and therefore delay the readiness of the whole apiserver.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/patch.go
```diff
@@ -84,6 +86,12 @@ func NewOpenShiftKubeAPIServerConfigPatch(kubeAPIServerConfig *kubecontrolplanev
 		genericConfig.AddPostStartHookOrDie("openshift.io-startkubeinformers", func(context genericapiserver.PostStartHookContext) error {
 			go kubeInformers.Start(context.StopCh)
 			go openshiftInformers.Start(context.StopCh)
+			err := wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
```
sanchezl: > I would rather expect this much higher up in the outer plumbing code that sets up the group cache, and therefore delay the readiness of the whole apiserver.

@deads2k what do you think of waiting for the groups cache to sync here, as suggested by @sttts ? I think I would then be able to undo the other commit's changes to the `GroupCache`.

sttts: Don't we wire the WaitForSync funcs somewhere already?

mfojtik:
/hold

this will break bootstrap kube apiserver

vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/authorization/restrictusers/subjectchecker.go
deads2k: > I would rather expect this much higher up in the outer plumbing code that sets up the group cache, and therefore delay the readiness of the whole apiserver.

that's an undesireable outcome.  This affects a minor part of a minor feature that is ultimately optional.  Delaying readiness of the server is inappropriate in this case.  The authentication side fails closed and isn't always used either, so it's also inappropriate to delay readiness.

Also, if you delayed readiness, you would create a bootstrap loop with the openshift apiserver that would never be resolved on installation.

sttts: Yes, we realized the bootstrap loop in slack discussion. Some delay here is probably the best we can do if we care about new kube-apiservers to give correct results of the admission plugin. Any other suggestion?

sttts:
/topic Bug 1792569: UPSTREAM: <carry>: admission/restrictusers ensure groups cache synced

sttts:
/topic "Bug 1792569: UPSTREAM: : admission/restrictusers ensure groups cache synced"

sttts:
/title Bug 1792569: UPSTREAM: : admission/restrictusers ensure groups cache synced

sttts:
/retitle Bug 1792569: UPSTREAM: : admission/restrictusers ensure groups cache synced

PR #24413, opened 2020-01-17
UPSTREAM: 86895: Support dual-stack and singlestack ipv6 services with old clients

smarterclayton:
Fix a number of issues around IPFamily when the DualStack gate is
enabled that prevented services from being updated. This change
is a prerequisite for enabling DualStack and single-stack IPv6
due to how services are handled and defaulted.


smarterclayton:
Waiting for upstream merge.

smarterclayton:
Going to be rebased on a longer change set

PR #24414, opened 2020-01-18
UPSTREAM: <carry>: authentication ensure groups cache synced

sanchezl:


deads2k:
Don't wait in the authenticator, just check and drop the V(4) message right away.

PR #24417, opened 2020-01-20
Fix counting of junit tests and failures

mandre:
Passed tests were counted as failures. This resulted in wrong counts
for failures and tests in the generated junit files.


stevekuznetsov:
This is a good candidate for backports as well FYI

PR #24422, opened 2020-01-20
test: Allow imagetags resource to be added to group RBAC

smarterclayton:
This resource is new in 4.4 and is created by openshift-apiserver.


smarterclayton:
Flaking due to e2e-cmd, unrelated to this change.  Merging to unblock imagetags.

PR #24018, opened 2019-10-24
Bug 1744077: DROP: Cause Sandboxes to be killed if they have no default route

squeed:
When restarting kubelet, we can kill in-flight CNI requests. If they're killed between setting the interface up and adding the default route, then kubelet upon restart thinks the container is up.

Add a check so we kill the sandbox in that case.

We do not need to bring this forward to 4.x, as we use CRIO.

ref: rhbz 1744077

PR #24426, opened 2020-01-21
DROP: Cause Sandboxes to be killed if they have no default route

juanluisvaladas:
Taken from: https://github.com/openshift/origin/pull/24018
When restarting kubelet, we can kill in-flight CNI requests. If they're killed between setting the interface up and adding the default route, then kubelet upon restart thinks the container is up.

Add a check so we kill the sandbox in that case.

We do not need to bring this forward to 4.x, as we use CRIO.

Doing a separate PR so that I can get in the cluster and see what is actually failing


juanluisvaladas:
/hold
I'm just doing this to see if this passes the tests, I still need to verify manually as I'm taking over https://github.com/openshift/origin/pull/24018 and I don't know how much has Casey done here, if it's just a quick check or what.

PR #24435, opened 2020-01-21
bug 1797740: UPSTREAM: <carry>: simplify the authorizer patch to allow the flags t…

deads2k:
…o function


deads2k:
terraform

/retest

vendor/k8s.io/kubernetes/openshift-kube-apiserver/enablement/intialization.go
```diff
@@ -4,6 +4,8 @@ import (
 	"io/ioutil"
 	"path"
 
+	"k8s.io/apiserver/pkg/server"
+
```
sttts: download-1

vendor/k8s.io/kubernetes/pkg/kubeapiserver/authorizer/modes/patch.go
```diff
@@ -0,0 +1,8 @@
+package modes
+
+var ModeScope = "Scope"
+var ModeSystemMasters = "SystemMasters"
```
sttts: why not also BrowserSafe?

vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/patch.go
```diff
@@ -0,0 +1,7 @@
+package server
+
+var skipSystemMastersAuthorizer = false
+
+func SkipSystemMastersAuthorizer() {
```
sttts: godoc

vendor/k8s.io/kubernetes/openshift-kube-apiserver/enablement/intialization.go
```diff
@@ -78,4 +80,7 @@ func ForceGlobalInitializationForOpenShift() {
 	// TODO, we should scrub these out
 	bootstrappolicy.ClusterRoles = bootstrappolicy.OpenshiftClusterRoles
 	bootstrappolicy.ClusterRoleBindings = bootstrappolicy.OpenshiftClusterRoleBindings
+
+	// we need to have the authorization chain place something before system:masters
+	server.SkipSystemMastersAuthorizer()
```
sttts: // disable implicitly added system/master authz, and turn it into another authz mode "SystemMasters", to be added via authorization-mode

vendor/k8s.io/kubernetes/pkg/kubeapiserver/authorizer/modes/patch.go
deads2k: > why not also BrowserSafe?

Because that change the resources, but the point of SystemMasters is to approve everything, no matter the resource.

sttts: as another authz-mode? Which resource?

sttts:
/lgtm

Out of curiosity: https://github.com/openshift/origin/pull/24435/files#r370282776

vendor/k8s.io/kubernetes/pkg/kubeapiserver/authorizer/modes/patch.go
deads2k: > as another authz-mode? Which resource?

it wraps the attributes with an adapter than can conditionally return a different verb and resource for /proxy to require a second rule to allow it so that always comes from an authenticated request.

sttts:
Serial test is always red. Conincidence?

PR #24437, opened 2020-01-22
UPSTREAM: 85898: Fix multinode storage e2e tests for multizone clusters

tsmetana:
The multi-node PV test creates a PV, starts a pod that writes data to a file on
the PV, then kills the pod and starts another with node anti-affinity to read
the data from the PV attached to a different node. This is problematic on
multi-AZ clusters: there may only be one node in a zone and for volumes that
can't be transferred among zones (AWS EBS, GCE PD, ...) the test would fail.


bertinatto:
/assign @soltysh 
for approval (also approved the upstream patch)

PR #24441, opened 2020-01-22
UPSTREAM: 86895: service: Handle single stack service errors when dual-stack flag enabled

p0lyn0mial:


p0lyn0mial:
/hold

as it's not clear if we really want it.

PR #24443, opened 2020-01-22
Add delete-namespace* flags

mkumatag:
These flags will be helpful while debugging the tests

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -134,6 +134,7 @@ func newRunCommand() *cobra.Command {
 		},
 	}
 	bindOptions(opt, cmd.Flags())
+	bindE2EOptions(cmd.Flags())
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


pweil-:
/assign @soltysh 

Maciej, are you the best one to review this or can you assign it to the right person?  Thanks!

mkumatag:
> Isn't setting `DELETE_NAMESPACE` envvar, see here: https://github.com/openshift/origin/blob/master/test/extended/util/test.go#L62

seems like that does magic for deleting the samespace but I was more interested in DeleteNamespaceOnFailure which retains the data only for the failures!

soltysh:
@mkumatag you can still run an individual test manually and with `DELETE_NAMESPACE` keep the namespace. This is only for local dev so that shouldn't be a problem. We will never allow setting this anywhere in our CI or alike.

mkumatag:
That is how I'm running it now.. 😁

PR #24445, opened 2020-01-22
test: Generate test name updates rather than calculate dynamically

smarterclayton:
It is expensive to perform 100 checks against 1k names on every startup.
Instead, move these to generated files that can be reviewed out of bounds
on change. Add them to generated-bindata for verification and regen.


test/extended/util/annotate/OWNERS
```diff
@@ -0,0 +1,5 @@
+reviewers:
+  - smarterclayton
+approvers:
+  - smarterclayton
+  - decarr
```
derekwaynecarr: derekwaynecarr

smarterclayton:
Merging to reduce CPU use and unblock David.

PR #24447, opened 2020-01-22
test: Remove wildcard use in the test regex files

smarterclayton:
Regexes can be expensive, more correct fix coming shortly.


smarterclayton:
Fixed in the other Pr

PR #24450, opened 2020-01-23
bug 1797788: add openshift-tests as a future delegation binary

deads2k:
Let's learn about images!

This eases me into https://github.com/openshift/enhancements/pull/185 by trying to get the image builds in order first.

1. tests-base - includes `openshift-tests` which is now an execve call and will later be a binary union
2. tests - builds on tests-base, includes `openshift-test-kubernetes` which is the same as the previous `openshift-tests`
3. tests-kubernetes - the same as 2 right now, but eventually tests will be built out of openshift/openshift-tests instead.


deads2k:
disappointing that github didn't do the straight move as I would have liked

soltysh:
Closing in favor of https://github.com/openshift/origin/pull/24560
/close

PR #24455, opened 2020-01-24
Test internal port connectivity internally for 9000-9999

smarterclayton:
We realized UPI was not allowing these ports, which is wrong.  This test requires the entire cluster to allow 9000-9999 connectivity from host to host and tests that.


test/extended/networking/internal_ports.go
```diff
@@ -0,0 +1,80 @@
...
+	})
+}
+
+var _ = ginkgo.Describe("[sig-network][Feature:Platform] Internal connectivity", func() {
+	f := framework.NewDefaultFramework("nettest")
+
+	ginkgo.It("should function for pod-Service: http", func() {
+		config := newNetworkingTestConfig(f)
+		ginkgo.By(fmt.Sprintf("dialing(http) %v --> %v:%v (config.clusterIP)", config.TestContainerPod.Name, config.ClusterIP, nodeTCPPort))
+		config.DialFromTestContainer("http", config.ClusterIP, nodeTCPPort, config.MaxTries, 0, config.EndpointHostnames())
```
wking: I looked through `vendor/k8s.io/kubernetes/test/e2e/framework/networking_utils.go` briefly to try and understand `EndpointHostnames`, but it's like an onion.  Can anyone give a high-level summary about whether we can expect these to run on compute and/or control-plane machines?

smarterclayton: They are randomly chosen.

smarterclayton: In this PR this is validated by tolerating master taints.

PR #24458, opened 2020-01-27
image: add ImageTag to etcd test

mfojtik:
This should address following failure:

```
fail [github.com/openshift/origin/test/extended/etcd/etcd_test_runner.go:72]: test failed:
no test data for image.openshift.io/v1, Kind=ImageTag.  Please add a test for your new type to etcdStorageData.
etcd data does not match the types we saw:
in etcd data but not seen:
[]
seen but not in etcd data:
[
	image.openshift.io/v1, Resource=imagetags
```

Found by @smarterclayton in https://testgrid.k8s.io/redhat-openshift-ocp-release-4.4-informing#release-openshift-ocp-installer-e2e-metal-serial-4.4

Question: Why was this test not failing before and we only found this on metal?

/cc @bparees 
/cc @adambkaplan

### DIFF
--- a/test/extended/etcd/etcd_storage_path.go
+++ b/test/extended/etcd/etcd_storage_path.go
@@ -185,6 +185,7 @@ var openshiftEtcdStorageData = map[schema.GroupVersionResource]etcddata.StorageD
 // TODO fix for real GVK.
 var kindWhiteList = sets.NewString(
 	"ImageStreamTag",
+	"ImageTag",
 	"UserIdentityMapping",
 	// these are now served using CRDs
 	"ClusterResourceQuota",


p0lyn0mial:
it doesn't affect only metal, it affects `API data in etcd should be stored at the correct location and version for all resources` test. 

The new resource was added in https://github.com/openshift/openshift-apiserver/pull/51 and the test wasn't updated.

The test only runs as a part of a "serial test suite" for example `e2e-aws-serial`

p0lyn0mial:
Why https://github.com/openshift/openshift-apiserver/pull/51 didn't run `e2e-aws-serial`?

bertinatto:
This needs to merged manually, right? Otherwise `e2e-aws-serial` will keep failing?

p0lyn0mial:
@mfojtik it looks like the stub is broken - https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24458/pull-ci-openshift-origin-master-e2e-aws-serial/12341

I think it should be ``{"metadata": {"name": "itag1"}, "spec": {"name": "tag"}, "status": {"tag":"itag1"}}``

p0lyn0mial:
> This needs to merged manually, right? Otherwise e2e-aws-serial will keep failing?

No, this pull should fix `e2e-aws-serial` (unless it has more issues)

smarterclayton:
lgtm as well

bparees:
> Why openshift/openshift-apiserver#51 didn't run e2e-aws-serial?

e2e-aws-serial isn't enabled on openshift-apiserver.  sounds like it should be added?

p0lyn0mial:
> e2e-aws-serial isn't enabled on openshift-apiserver. sounds like it should be added?

Yes, I'm going to add it - created a placeholder - https://bugzilla.redhat.com/show_bug.cgi?id=1795194 for it.

smarterclayton:
Actually, changing my LGTM.  ImageTags are virtual, and won't be in etcd.  They're basically the same as image stream tags.

mfojtik:
@smarterclayton alright, updated.

marun:
This can't merge due to the failing oauth check in e2e-aws-serial resulting from incompatibility with bound tokens.  Options to merge include:

1. force merging this PR once all tests but the oauth check are passing
2. force merging the fix for auth compatibility: https://github.com/openshift/origin/pull/24461
3. reverting bound tokens: https://github.com/openshift/cluster-kube-apiserver-operator/pull/734

I think 1 is suggested, and the decision whether to fix or revert bound token support can be handled separately.

smarterclayton:
Force merging, test passes

PR #24461, opened 2020-01-27
UPSTREAM: <carry>: oauth-authn: add implicit audience support

sttts:
The TokenReview REST implementation expects that the authn chain returns audiences in the authn response. Authenticators that don't support audiences must return the default slice of audiences, e.g. the legacy service account authenticator does that. This PR adds that to our custom oauth authenticator.

Folliowing https://github.com/kubernetes/kubernetes/blob/master/pkg/serviceaccount/jwt.go#L283.

Compare https://github.com/kubernetes/kubernetes/pull/69582/files#r225535241.

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/tokenauthenticator.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/tokenauthenticator.go
@@ -3,6 +3,7 @@ package oauth
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kauthenticator "k8s.io/apiserver/pkg/authentication/authenticator"
@@ -16,18 +17,20 @@ import (
 var errLookup = errors.New("token lookup failed")
 
 type tokenAuthenticator struct {
-	tokens      oauthclient.OAuthAccessTokenInterface
-	users       userclient.UserInterface
-	groupMapper UserToGroupMapper
-	validators  OAuthTokenValidator
+	tokens       oauthclient.OAuthAccessTokenInterface
+	users        userclient.UserInterface
+	groupMapper  UserToGroupMapper
+	validators   OAuthTokenValidator
+	implicitAuds kauthenticator.Audiences
 }
 
-func NewTokenAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, users userclient.UserInterface, groupMapper UserToGroupMapper, validators ...OAuthTokenValidator) kauthenticator.Token {
+func NewTokenAuthenticator(tokens oauthclient.OAuthAccessTokenInterface, users userclient.UserInterface, groupMapper UserToGroupMapper, implicitAuds kauthenticator.Audiences, validators ...OAuthTokenValidator) kauthenticator.Token {
 	return &tokenAuthenticator{
-		tokens:      tokens,
-		users:       users,
-		groupMapper: groupMapper,
-		validators:  OAuthTokenValidators(validators),
+		tokens:       tokens,
+		users:        users,
+		groupMapper:  groupMapper,
+		validators:   OAuthTokenValidators(validators),
+		implicitAuds: implicitAuds,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


deads2k:
I have to get to a computer to see, but this looks plausible.

/Approve

sttts:
Fixed the unit tests using the changed constructor.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/tokenauthenticator.go
```diff
@@ -55,6 +58,18 @@ func (a *tokenAuthenticator) AuthenticateToken(ctx context.Context, name string)
...
+	tokenAudiences := a.implicitAuds
+	requestedAudiences, ok := kauthenticator.AudiencesFrom(ctx)
+	if !ok {
+		// default to apiserver audiences
+		requestedAudiences = a.implicitAuds
+	}
+
+	auds := kauthenticator.Audiences(tokenAudiences).Intersect(requestedAudiences)
+	if len(auds) == 0 && len(a.implicitAuds) != 0 {
+		return nil, false, fmt.Errorf("token audiences %q is invalid for the target audiences %q", tokenAudiences, requestedAudiences)
```
deads2k: I'll allow this because I think it will never happen, but the oidc example suggests `return nil, false, nil`

deads2k:
/lgtm
/hold

holding on proof that the serial test is improve
/retest

vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/tokenauthenticator.go
sttts: It's copied one-to-one from the service account authenticator.

sttts:
GCP infra flakes

marun:
Serial test failure does not appear related to bound tokens:

```
fail [github.com/openshift/origin/test/extended/etcd/etcd_test_runner.go:72]: test failed:
no test data for image.openshift.io/v1, Kind=ImageTag.  Please add a test for your new type to etcdStorageData.
etcd data does not match the types we saw:
in etcd data but not seen:
[]
seen but not in etcd data:
[
	image.openshift.io/v1, Resource=imagetags]
```

marun:
- The only failures in `e2e-aws-serial` appears to be fixed by https://github.com/openshift/origin/pull/24458
- No instances in `e2e-aws-serial` kube-apiserver logs of the error message reported in the bz:

`storage.go:100] error validating audience. want=["auth.openshift.io"] got=[]`

vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/tokenauthenticator.go
```diff
@@ -55,6 +58,18 @@ func (a *tokenAuthenticator) AuthenticateToken(ctx context.Context, name string)
 		groupNames = append(groupNames, group.Name)
 	}
 
+	tokenAudiences := a.implicitAuds
```
marun: (No action required) Consider simplifying:

```golang
    // Default to implicit audiences
    audiences := a.implicitAuds

    if len(a.implicitAuds) > 0 {
        // Only validate audiences when both implicit audiences and context-provided
        // audiences are present.

        requestedAudiences, ok := kauthenticator.AudiencesFrom(ctx)
        if ok {
            // Valid audiences for the token consist of the non-empty intersection
            // between the implicit and requested audiences.
            audiences = kauthenticator.Audiences(a.implicitAuds).Intersect(requestedAudiences)
            if len(audiences) == 0 {
                return nil, false, fmt.Errorf("token audiences %q is invalid for the target audiences %q", a.implicitAuds, requestedAudiences)
            }
        }
    }
```

marun:
Now that #24458 was manually merged e2e-aws-serial should be capable of passing.

sttts:
https://github.com/openshift/origin/pull/24458 merged to fix the imagetags issue.

Rebased.

sttts:
`error validating audience. want=["auth.openshift.io"] got=[]` is gone from the logs. So this PR fixes the TokenReview issue.

vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/tokenauthenticator.go
sttts: wanted to keep the structure of the service account authenticator

sttts:
/retest

GCP flakes

vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/tokenauthenticator.go
stlaz: +1 to less nesting however visual-only that might appear

stlaz:
/hold cancel
the serial test now passes

mfojtik:
force merging this as it make the queue stable again.

PR #24462, opened 2020-01-27
Bug 1761930: Fix ClusterLogging curator diagnostic check

jcantrill:
Fix cluster logging diagnostic cronjob curator check 

https://bugzilla.redhat.com/show_bug.cgi?id=1761930


bparees:
overriding the BZ, this function has no equivalent in 4.x

PR #24463, opened 2020-01-27
test: Set appropriate cloud defaults even when not specified on tests

smarterclayton:
The --provider argument for openshift-tests accepts a string (the type)
which triggers discovery, or JSON, which previously meant "use exactly
these values". However, we have more tests which depend on logic from
implicit fields (multimaster, region, zone, and number of nodes) that
we don't want to have accidentally off. This was recently triggered
when refactoring the methods in that numNodes was lost in CI, causing
certain networking tests not to run.

This commit changes openshift-tests run/run-upgrade to attempt to fill
all fields of the config even when JSON is passed to --provider by
invoking the cloud config load method, then updating provider for
the subsequent exec'd test command so that it is accurate. The run-test
command does not perform discovery, which forces the suite runner to
be accurate.

If correct, the granular network tests will not be in the skip list.

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -111,7 +111,12 @@ func newRunCommand() *cobra.Command {
 		SilenceErrors: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return mirrorToFile(opt, func() error {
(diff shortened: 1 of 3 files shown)


smarterclayton:
Because we lost test coverage when the previous change merged and this restores several fairly important tests for networking, force merging to prevent regressions.

PR #24469, opened 2020-01-28
Bug 1795705: increase error logging around apiserver binary build flow

gabemontero:
@openshift/openshift-team-developer-experience FYI

/assign @bparees

### DIFF
--- a/pkg/build/apiserver/registry/buildconfiginstantiate/rest.go
+++ b/pkg/build/apiserver/registry/buildconfiginstantiate/rest.go
@@ -206,7 +206,7 @@ func (h *binaryInstantiateHandler) handle(r io.Reader) (runtime.Object, error) {
 					}
 				}
 			}
-			glog.V(2).Infof("failed to instantiate: %#v", request)
+			glog.V(2).Infof("failed to instantiate: %#v with error %#v", request, err)
 			return false, err
 		}
 		build = result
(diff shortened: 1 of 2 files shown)


gabemontero:
oh, should this be a straight openshift/ose patch @bparees, or should we in fact start with openshift/origin?

bparees:
> oh, should this be a straight openshift/ose patch @bparees, or should we in fact start with openshift/origin?

Ultimately we're going to be building a one-off locally, so where the PR is created doesn't really matter right now.

Beyond building the patch (which i guess should be built off ose, not origin), we'll want to deliver these changes as a normal "bug fix" meaning:  master, then cherry-pick to 3.11 (and we can debate if we really need to do it in 4.4/4.3/etc... i'm ok skipping those).

pkg/build/apiserver/registry/buildconfiginstantiate/rest.go
```diff
@@ -206,7 +206,7 @@ func (h *binaryInstantiateHandler) handle(r io.Reader) (runtime.Object, error) {
 					}
 				}
 			}
-			glog.V(2).Infof("failed to instantiate: %#v", request)
+			glog.V(2).Infof("failed to instantiate: %#v with error %#v", request, err)
```
bparees: i don't believe the `#` is adding anything useful here or in the other spots, probably just making the output less readable.

bparees: (the # on the err format.  the one on the request is useful since it's a struct)

gabemontero: so just #v for the err then ?

bparees:
```suggestion
			glog.V(2).Infof("failed to instantiate: %#v with error %v", request, err)
```

gabemontero: sorry I meant %v which is what your suggestion does 

I'll just include it with what I move to openshift/ose

gabemontero:
closing this out

have changes in branch https://github.com/gabemontero/ose/tree/bin-build-debuggability

awaiting guidance on building image that SRE team will apply to dedicated cluster

ultimately, per @bparees prior note will propagate any changes through the patch manager process, starting with master branch

PR #24471, opened 2020-01-28
Bug 1796715: Add a service upgrade test that verifies availability

smarterclayton:
The upstream upgrade test doesn't correctly test for disruption of the backend during upgrade - it only tests that the service is not available for more than 2 minutes at a time.

This is obviously not a good test of whether a service maintains availability during upgrade. Take the core structure and use the monitor and polling behavior of the apiserver which is much more effective.

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -40,7 +40,7 @@ func Start(ctx context.Context) (*Monitor, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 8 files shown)


smarterclayton:
@ironcladlou FYI this is the test

We can also hang route detection on this easily (basically just create a route, get the DNS name, and then verify it).

smarterclayton:
Merging so we can begin back porting, tests allow a higher tolerance than we want, but we need the gates on to prevent future regressions

PR #24473, opened 2020-01-29
Bug 1796017: UPSTREAM: 79160: e2e Flake: Pass updated deployment for comparison of replica count

soltysh:
/assign @tnozicka


soltysh:
This is not relevant anymore.
/close

PR #24477, opened 2020-01-29
Bug 1796234: Move the oc explain tests for networking CRDs to a new test

abhat:
Networking CRDs such as `clusternetworks`, `hostsubnets`, `egressnetworkpolicies` etc. don't apply to `OVNKubernetes`, so the `oc explain` tests need to be split into two. 

This PR adds a new test so that the verification of `oc explain` output for these CRDs can be skipped for `OVNKubernetes`.


test/extended/util/annotate/rules.go
```diff
@@ -209,6 +209,7 @@ var (
 			`\[sig-network\] Networking Granular Checks: Services should function for endpoint-Service`,
 			// https://github.com/ovn-org/ovn-kubernetes/issues/928
 			`\[sig-network\] Services should be rejected when no endpoints exist`,
+			`\[cli\] Oc explain networking types should contain proper fields description for special networking types`,
```
vrutkovs: Still not skipped in OVN test. Perhaps due to a typo: `oc` needs to be lowercase?

test/extended/cli/explain.go
```diff
@@ -315,6 +315,9 @@ var (
 			field:   "consoleyamlsamples.spec",
 			pattern: `DESCRIPTION\:.*`,
 		},
+	}
+
+	specialNetworkingTypes = []explainExceptions{
```
vrutkovs: Not sure if the test stops on the first instance it finds - or OVN doesn't have any of these?

abhat: ovn doesn't have any of these.

test/extended/util/annotate/rules.go
abhat: I actually was following the perceived formatting rules and thought that it needs to be `Oc` instead of `oc`. @smarterclayton wdyt?

dcbw: @abhat IIRC they do need to match case; it's just a regex against the test name. So the exclusion explicitly handles case, it needs to match the test Describe.

abhat: pushed a fix to match the case.

vrutkovs:
seems to be flakes in gcp test
/retest

PR #24482, opened 2020-01-30
Bug 1796535: Handle SDN CRDs as regular crd types without special rules

abhat:
Previously, it was believed that the sdn crds need special exception
handling. But it turns out that the crds were missing the preserveUnknownFields
flag. Adding that flag back obviates the need for special handling.


test/extended/cli/explain.go
```diff
@@ -375,11 +359,11 @@ var _ = g.Describe("[cli] oc explain networking types", func() {
 
 	oc := exutil.NewCLI("oc-explain", exutil.KubeConfigPath())
 
-	g.It("should contain proper fields description for special networking types", func() {
-		for _, st := range specialNetworkingTypes {
-			e2e.Logf("Checking %s, Field=%s...", st.gv, st.field)
-			o.Expect(verifyExplain(oc, nil, schema.GroupVersionResource{},
-				st.pattern, st.field, fmt.Sprintf("--api-version=%s", st.gv))).NotTo(o.HaveOccurred())
+	g.It("should contain proper spec+status for networking CRDs", func() {
```
soltysh: If networking is now following the usual types, I'd expect it to be part of `crdTypes` as the others are. No exceptions.

abhat: Yes, but we need to split the test out since these crds are applicable only for Openshift SDN and not OVN.

soltysh: :+1:

abhat:
@soltysh can you approve please?

abhat:
Upon closer examination, some of the crds have spec, but no status. It's best to leave them with special handling. Closing the PR.

PR #24487, opened 2020-01-31
Bug 1797045: [3.11] Clean up stale egress IP iptables rules on startup

danwinship:
Backport of https://github.com/openshift/sdn/pull/88

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -97,6 +97,8 @@ func (eip *egressIPWatcher) Synced() {
 			}
 		}
 	}
+
+	eip.iptables.SyncEgressIPRules()
 }
 
 func egressIPLabel(link netlink.Link) (string, error) {
--- a/pkg/network/node/iptables.go
+++ b/pkg/network/node/iptables.go
@@ -3,7 +3,10 @@
 package node
 
 import (
+	"bytes"
 	"fmt"
+	"regexp"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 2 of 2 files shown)


danwinship:
/assign @knobunc @juanluisvaladas 
(Ben reviewed the master PR, openshift/sdn#88, Juan reviewed the 4.3 PR, openshift/sdn#100.)

pkg/network/node/networkpolicy_test.go
```diff
@@ -550,7 +550,7 @@ func _TestNetworkPolicyCache(t *testing.T) {
 		name := fmt.Sprintf("namespace-%d", vnid)
 		addNamespace(np, name, vnid, map[string]string{
 			"pod.network.openshift.io/legacy-netid": fmt.Sprintf("%d", vnid),
-			"name": name,
+			"name":                                  name,
```
juanluisvaladas: Looks like there is a formatting issue here:

pkg/network/node/networkpolicy_test.go
```diff
@@ -638,7 +638,7 @@ func _TestNetworkPolicyCache(t *testing.T) {
...
```
juanluisvaladas: same

pkg/network/node/networkpolicy_test.go
```diff
@@ -647,7 +647,7 @@ func _TestNetworkPolicyCache(t *testing.T) {
...
```
juanluisvaladas: same

danwinship:
> Looks like there are a few unintended changes in networkpolicy_test.go, I assume these were commited accidentally.

ah, no, it seems that they're necessary to make test-verify pass. I'm not sure how the branch ended up in the current state...

anyway, repushed as two separate commits

juanluisvaladas:
You're right :-|
/lgtm
/retest

danwinship:
> ah, no, it seems that they're necessary to make test-verify pass. I'm not sure how the branch ended up in the current state...

ok, I think I was comparing the wrong test results vs pushes or something, because it seems like actually that part of the patch was *breaking* test-verify... anyway, reverted now

danwinship:
@juanluisvaladas this seems to have passed now. can you re-lgtm?

PR #24492, opened 2020-02-03
Bug 1768756: Run Prometheus specific tests cases at the end of the test run

lilic:
Testing, do not merge.

/shrug
/hold


pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -214,6 +214,10 @@ func (opt *Options) Run(args []string) error {
 		return strings.Contains(t.name, "[Smoke]")
 	})
 
+	water, normal := splitTests(normal, func(t *testCase) bool {
+		return strings.Contains(t.name, "[Water]")
```
smarterclayton: Water is probably too obscure (even smoke was pushing it).  Probably better to use [Pre] and [Post] or [Early] and [Late]

LiliC: Yes yes, sorry hence the WIP, just wanted an easier way to find the job :D 

Can also change the Smoke to Pre and Change to Post, sgty?

LiliC: @smarterclayton Done, PTAL, thanks!

deads2k:
I haven't looked at the implementation, but I'm strongly in favor of this idea if it is judiciously used.  I'd like to see guidance that only api-approvers can merge PRs that assert, "I must run late" because otherwise we'll end up with everyone wanting it.

test/extended/prometheus/prometheus.go
```diff
@@ -33,6 +33,53 @@ import (
...
+	var (
+		oc = exutil.NewCLIWithoutNamespace("prometheus")
+
+		url, bearerToken string
+	)
+	g.BeforeEach(func() {
+		var ok bool
+		url, bearerToken, ok = locatePrometheus(oc)
+		if !ok {
+			e2e.Skipf("Prometheus could not be located on this cluster, skipping prometheus test")
```
smarterclayton: Prometheus missing is an error condition, my PR changed this to Failf and this should too.

LiliC:
> that only api-approvers can merge PRs that assert, "I must run late" because otherwise we'll end up with everyone wanting it.

Whenever you regenerate the test file (test/extended/util/annotate/generated/zz_generated.annotations.go) you need to get approval on the origin PR from Clayton anyways so this won't be a problem.

> I haven't looked at the implementation,

My main concern with this actual implementation is that some tests might run longer than the alerts test case (or any other "Late" tests) that might potentially cause some flakes if those vary in length. If anyone has any better ideas where we could move the trigger of the "Late" tests I am open to suggestions.

test/extended/prometheus/prometheus.go
LiliC: Makes sense, done.

smarterclayton:
Hrm, the count over 2h - shouldn’t that be “firing now”?

LiliC:
@smarterclayton 
No, we want to catch any alerts firing in the duration of the cluster run. If we get alerts and they are not valid:
- the alerting rule should be fixed as its maybe too sensitive or just wrong query. As we want to be shipping the right alerts.
- if any are really the not valid we can always add them to the exclusion list

I think this is a way of catching flaky alerts now so we don’t have flaky tests later.

LiliC:
We have two failures, both with alerts firing in the cluster, seen those constantly now in the few runs:
`FailingOperator`, `NoOperatorGroup`both from OLM and two times KubeAPILatencyHigh in the https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24492/pull-ci-openshift-origin-master-e2e-gcp/5771

and for second one twice `KubeAPILatencyHigh`
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24492/pull-ci-openshift-origin-master-e2e-aws-fips/1274

I guess we have to fix those alerts before merging this PR.

LiliC:
/retest

Just want to make sure two runs have the same expected alerts firing.

smarterclayton:
I'm confused why we need `[Late]` if you already have to fix flaky alerts caused by tests.  The current flake is a test causing an alert - how are you preventing that with this change?

LiliC:
If we have valid reasons to have tests cause alerts, those alerts can be added to the list of those we expect to fire, not sure that should be the case.

Point of it is to detect all alerts firing so we are consistent with catching alerts, that can be only done if we evaluate alerts that fired throughout the cluster run.

smarterclayton:
https://github.com/openshift/origin/pull/24500#issuecomment-583101783 - the FailingOperator alert shouldn't be info, which means the moment that is fixed this starts flaking again because it's firing.  Who is fixing that?

LiliC:
@smarterclayton 
> this starts flaking again because it's firing. Who is fixing that?

If this is expected to fail because of tests we add it to the allowed alerts and it won't flake as it will be ignored.

LiliC:
Again tests should not influence alerts, if they do we should create a silence for it, during the duration. 

How do we handle this for the CVO Operator Status, I assume we have a similar check that all operators are not degraded at the end of the test run? Should not things we detect in alerts set Status to degraded/failed as well? Just curious, but maybe my logic is wrong here. :)

LiliC:
We should consider adding the Early alert check as well, to make sure absolutely *no* alerts are firing ever, as long as we add a Late one as well that makes sure after a cluster has been running no alerts are firing unless expected.

Let me know wdyt.

LiliC:
/retest

some failed builds

smarterclayton:
> Again tests should not influence alerts, if they do we should create a silence for it, during the duration.

Ok, but then we need to add that silence - can you work with the operator team to get that fixed?

> How do we handle this for the CVO Operator Status, I assume we have a similar check that all operators are not degraded at the end of the test run? Should not things we detect in alerts set Status to degraded/failed as well? Just curious, but maybe my logic is wrong here. :)

No operator should go degraded during a parallel e2e run.  Operators will go degraded during serial, disruptive, and a upgrade runs.  We run the full parallel e2e suite after some of the release periodic upgrades, so it's likely that this is going to start failing on those runs.

smarterclayton:
I don't want to derail this PR (I agree in general with the need) but I want to clear the flake from our builds.  For now, I'd like to merge #24499 and this can be based on top of it.

LiliC:
@smarterclayton okay, go ahead :shipit: , I think having both Early and Late alerts is a good idea.

- will rebase next week
-exclude the flakes from here, open bugzilla and work with the teams to create silences
- we will need to query alertmanager directly instead of query Prometheus ALERTS to be able to take advantage of the silence, but I think its the right approach.

smarterclayton:
> we will need to query alertmanager directly instead of query Prometheus ALERTS to be able to take advantage of the silence, but I think its the right approach.

I don't know that silences are going to work on all tests - for instance serial tests upstream may not be changeable easily and the fixes would have to be carried forever.  I think it's possible that checking for ANY alerts late may be too difficult to do in all test scenarios, and we might need to have a more select set.

LiliC:
> level=error msg="Error: NoSuchBucket: The specified bucket does not exist"
AWS  🤷‍♀ 
/test e2e-aws-serial

LiliC:
/test e2e-gcp-upgrade
/test e2e-cmd

seems like random failures, the aws is a valid alert one.
>             s: "promQL query: count_over_time(ALERTS{alertname!~\"Watchdog|AlertmanagerReceiversNotConfigured\",alertstate=\"firing\"}[2h]) >= 1 had reported incorrect results:\n[{\"metric\":{\"alertname\":\"KubeAPILatencyHigh\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"component\":\"apiserver\",\"endpoint\":\"https\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"events\",\"scope\":\"namespace\",\"service\":\"kubernetes\",\"severity\":\"warning\",\"verb\":\"DELETECOLLECTION\",\"version\":\"v1\"},\"value\":[1581353714.323,\"3\"]}]",

Will have a look at where the KubeAPILatencyHigh might be coming from. Retesting to hopefully get the same results. Seems like the OLM alerts were fixed 🤔  

/test e2e-aws-fips

LiliC:
The recently merged PR that introduced alert `UsingDeprecatedAPIExtensionsV1Beta1 ` is now firing in the "Late" test and blocking this PR to be merged. The test failure is valid, and the "Early" Alert test check did not catch that, which is why we need the Late one as well. cc @smarterclayton 

We should not have merged that PR as per my comment https://github.com/openshift/cluster-kube-apiserver-operator/pull/742#issuecomment-584089082 cc @soltysh can we at least exclude the openshift namespace from the alerting rule and add it instead to the e2e tests as a recording rule test case as suggested? 

>             s: "promQL query: count_over_time(ALERTS{alertname!~\"Watchdog|AlertmanagerReceiversNotConfigured\",alertstate=\"firing\"}[2h]) >= 1 had reported incorrect results:\n[{\"metric\":{\"alertname\":\"KubeAPILatencyHigh\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"component\":\"apiserver\",\"endpoint\":\"https\",\"group\":\"events.k8s.io\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"events\",\"scope\":\"namespace\",\"service\":\"kubernetes\",\"severity\":\"warning\",\"verb\":\"DELETECOLLECTION\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"3\"]},{\"metric\":{\"alertname\":\"KubeAPILatencyHigh\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"component\":\"apiserver\",\"endpoint\":\"https\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"events\",\"scope\":\"namespace\",\"service\":\"kubernetes\",\"severity\":\"warning\",\"verb\":\"DELETECOLLECTION\",\"version\":\"v1\"},\"value\":[1581414633.285,\"3\"]},{\"metric\":{\"alertname\":\"UsingDeprecatedAPIExtensionsV1Beta1\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"client\":\"cluster-policy-controller/v0.0.0 (linux/amd64) kubernetes/$Format\",\"code\":\"0\",\"component\":\"apiserver\",\"contentType\":\"application/vnd.kubernetes.protobuf;stream=watch\",\"endpoint\":\"https\",\"group\":\"extensions\",\"instance\":\"10.0.159.45:6443\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"ingresses\",\"scope\":\"cluster\",\"service\":\"kubernetes\",\"severity\":\"info\",\"verb\":\"WATCH\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"45\"]},{\"metric\":{\"alertname\":\"UsingDeprecatedAPIExtensionsV1Beta1\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"client\":\"cluster-policy-controller/v0.0.0 (linux/amd64) kubernetes/$Format\",\"code\":\"200\",\"component\":\"apiserver\",\"contentType\":\"application/vnd.kubernetes.protobuf\",\"endpoint\":\"https\",\"group\":\"extensions\",\"instance\":\"10.0.159.45:6443\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"ingresses\",\"scope\":\"cluster\",\"service\":\"kubernetes\",\"severity\":\"info\",\"verb\":\"LIST\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"63\"]},{\"metric\":{\"alertname\":\"UsingDeprecatedAPIExtensionsV1Beta1\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"client\":\"kube-controller-manager/v1.17.1 (linux/amd64) kubernetes/3351d34/kube-controller-manager\",\"code\":\"0\",\"component\":\"apiserver\",\"contentType\":\"application/vnd.kubernetes.protobuf;stream=watch\",\"endpoint\":\"https\",\"group\":\"extensions\",\"instance\":\"10.0.132.104:6443\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"ingresses\",\"scope\":\"cluster\",\"service\":\"kubernetes\",\"severity\":\"info\",\"verb\":\"WATCH\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"55\"]},{\"metric\":{\"alertname\":\"UsingDeprecatedAPIExtensionsV1Beta1\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"client\":\"kube-controller-manager/v1.17.1 (linux/amd64) kubernetes/3351d34/system:serviceaccount:kube-system:namespace-controller\",\"code\":\"200\",\"component\":\"apiserver\",\"contentType\":\"application/vnd.kubernetes.protobuf\",\"endpoint\":\"https\",\"group\":\"extensions\",\"instance\":\"10.0.132.104:6443\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"ingresses\",\"scope\":\"namespace\",\"service\":\"kubernetes\",\"severity\":\"info\",\"verb\":\"DELETECOLLECTION\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"54\"]},{\"metric\":{\"alertname\":\"UsingDeprecatedAPIExtensionsV1Beta1\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"client\":\"kube-controller-manager/v1.17.1 (linux/amd64) kubernetes/3351d34/system:serviceaccount:kube-system:namespace-controller\",\"code\":\"200\",\"component\":\"apiserver\",\"contentType\":\"application/vnd.kubernetes.protobuf\",\"endpoint\":\"https\",\"group\":\"extensions\",\"instance\":\"10.0.132.104:6443\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"ingresses\",\"scope\":\"namespace\",\"service\":\"kubernetes\",\"severity\":\"info\",\"verb\":\"LIST\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"54\"]},{\"metric\":{\"alertname\":\"UsingDeprecatedAPIExtensionsV1Beta1\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"client\":\"kube-state-metrics/v1.9.4 (linux/amd64) kube-state-metrics/29872a7\",\"code\":\"0\",\"component\":\"apiserver\",\"contentType\":\"application/vnd.kubernetes.protobuf;stream=watch\",\"endpoint\":\"https\",\"group\":\"extensions\",\"instance\":\"10.0.141.23:6443\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"ingresses\",\"scope\":\"cluster\",\"service\":\"kubernetes\",\"severity\":\"info\",\"verb\":\"WATCH\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"44\"]},{\"metric\":{\"alertname\":\"UsingDeprecatedAPIExtensionsV1Beta1\",\"alertstate\":\"firing\",\"apiserver\":\"kube-apiserver\",\"client\":\"openshift-controller-manager/v0.0.0 (linux/amd64) kubernetes/$Format\",\"code\":\"0\",\"component\":\"apiserver\",\"contentType\":\"application/vnd.kubernetes.protobuf;stream=watch\",\"endpoint\":\"https\",\"group\":\"extensions\",\"instance\":\"10.0.141.23:6443\",\"job\":\"apiserver\",\"namespace\":\"default\",\"resource\":\"ingresses\",\"scope\":\"cluster\",\"service\":\"kubernetes\",\"severity\":\"info\",\"verb\":\"WATCH\",\"version\":\"v1beta1\"},\"value\":[1581414633.285,\"40\"]}]",

soltysh:
> We should not have merged that PR that introduced the alert as per my comment [openshift/cluster-kube-apiserver-operator#742 (comment)](https://github.com/openshift/cluster-kube-apiserver-operator/pull/742#issuecomment-584089082) cc @soltysh can we at least exclude the openshift-* namespace from the alerting rule and add it instead to the e2e tests as a recording rule test case as suggested?

Right, sorry for that, I've left that other PR to ensure it passes and that ended being tagged and merged. The fix excluding openshift components from that alert is in https://github.com/openshift/cluster-kube-apiserver-operator/pull/762

LiliC:
/retest

random failures unrelated

lilic:
again unrelated :(
> could not resolve inputs: could not determine inputs for step [input:base]: could not resolve base image: imagestreamtags.image.openshift.io "4.4:base" is forbidden: User "system:anonymous" cannot get imagestreamtags.image.openshift.io in the namespace "ocp": no RBAC policy matched

/test e2e-aws-serial

lilic:
> could not resolve inputs: could not determine inputs for step [input:base]: could not resolve base image: imagestreamtags.image.openshift.io "4.4:base" is forbidden: User "system:anonymous" cannot get imagestreamtags.image.openshift.io in the namespace "ocp": no RBAC policy matched


/test e2e-aws-fips

lilic:
/retest

Should be fixed now

PR #24495, opened 2020-02-03
DO NOT MERGE: Add fake metrics report for configured signals

smarterclayton:
We aren't reporting any eviction metrics, try priming the counter
with zero for each threshold type just to see.


smarterclayton:
Confirmed the metric is registered but is just empty.

PR #24496, opened 2020-02-04
Bug 1798450: makes unavailableGauge metric to always reflect the current state of a service

p0lyn0mial:


p0lyn0mial:
fixes https://github.com/openshift/cluster-kube-apiserver-operator/pull/746

PR #24503, opened 2020-02-05
Bug 1794885: UPSTREAM: <carry>: bootstrap user - make tokens have implicit audiences

stlaz:
cc @sttts @marun

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/bootstrapauthenticator.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/authentication/oauth/bootstrapauthenticator.go
@@ -2,6 +2,7 @@ package oauth
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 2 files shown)


marun:
@stlaz @sttts Why does this code need to be duplicated from https://github.com/openshift/origin/pull/24461 ? Seems like an opportunity to factor out the common code to avoid this class of error in the future.

sttts:
It's actually copied from upstream (service account authn). If we want to factor it out, then in upstream.

marun:
> It's actually copied from upstream (service account authn). If we want to factor it out, then in upstream.

In an ideal world, yes, the openshift patches for tokenauthenticator and bootstrapauthenticator would derive from better-factored upstream code. I'm not sure why it would be necessary to wait for that, though, to remove the potential source of errors represented by tokenauthenticator and bootstrapauthenticator being independently implemented given that their functionality is largely common.

stlaz:
/retest
looks like KAS couldn't reach etcd in the upgrade job, not good

PR #24505, opened 2020-02-05
Bug 1746149: test if oc tag can work without imagestreamtags

dmage:
The test should be red until https://github.com/openshift/oc/pull/275 is merged.


adambkaplan:
/hold cancel

oc pr has merged

PR #24513, opened 2020-02-06
Bug 1767523: UPSTREAM: 84614: kubelet: add certificate rotation error metric

rphillips:
Adds a certificate rotation error metric to the kubelet.

Upstream ref: https://github.com/kubernetes/kubernetes/pull/84614


rphillips:
cc @sjenning

PR #24515, opened 2020-02-07
[release-4.1] Bug 1772593: Make DNS querying more efficient by querying once per dns name

juanluisvaladas:
Refactor the datastructures to query DNS names more efficiently as
the same DNS name may appear across multiple policies.

### DIFF
--- a/pkg/network/common/dns.go
+++ b/pkg/network/common/dns.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/miekg/dns"
 
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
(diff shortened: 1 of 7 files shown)


juanluisvaladas:
/retitle  [release-4.1] Bug 1772593: Make DNS querying more efficient by querying once per dns name

bparees:
At this point we are only accepting critical bug fixes for 4.1. This does not seem to meet that target, do you want to close it (and the BZ) or make a case for why it is critical?

juanluisvaladas:
I don't think anyone cares about 4.1, but it's important for 3.11: https://bugzilla.redhat.com/show_bug.cgi?id=1684079
And we can't introduce regressions.

bparees:
> I don't think anyone cares about 4.1, but it's important for 3.11: https://bugzilla.redhat.com/show_bug.cgi?id=1684079
And we can't introduce regressions.

i think 3.11 is a unique case.  No one is going to upgrade from 3.11 to 4.1 (they'd go directly to the latest 4.x).  So i do not think we require 3.11 fixes to exist in 4.1.  I'm not sure precisely what logic the bot is enforcing on 3.11 branches, but if it is requiring a 4.1.z BZ, we need to make it more flexible (and override it for now).

cc @eparis

juanluisvaladas:
@danwinship @knobunc ^^^

sdodson:
/close
4.1 is in a phase of its lifecycle that this fix is not warranted there. 3.11 is as @bparees points out at a different phase of its lifecycle and a bit of a special case.

PR #24530, opened 2020-02-12
Bug 1801095: Add e2e image resolve test for Deployment, DaemonSet adn StatefulSet

tnozicka:
These were not covered before.

Requires:
 - [x] https://github.com/openshift/apiserver-library-go/pull/24

/cc @soltysh @adambkaplan @smarterclayton 

~/hold 
(I'll like to sync merging it with our decision on fixing `oc set image-lookup` which sets the annotation only for the template.)~


soltysh:
> (I'll like to sync merging it with our decision on fixing `oc set image-lookup` which sets the annotation only for the template.)

All of our higher-level resources (those that embed PodTemplate) set it on PodTemplate, so it's consistent, I don't think we need to change it.

tnozicka:
>(I'll like to sync merging it with our decision on fixing oc set image-lookup which sets the annotation only for the template.)

> All of our higher-level resources (those that embed PodTemplate) set it on PodTemplate, so it's consistent, I don't think we need to change it.

Mutating admission that changes only a RS and not the parent resource template (like Deployment) is gonna set any workload on fire (hot looping) as they are based on comparing the template with the RS.

test/extended/images/resolve.go
```diff
@@ -128,6 +127,83 @@ var _ = g.Describe("[Feature:ImageLookup][registry][Conformance] Image policy",
...
+						TerminationGracePeriodSeconds: &one,
+						Containers: []kapiv1.Container{
+							{Name: "resolve", Image: "busybox:latest", Command: []string{"/bin/true"}},
+						},
+					},
+				},
+			},
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(statefulset.Spec.Template.Spec.Containers[0].Image).To(o.Equal(tag.Image.DockerImageReference))
```
soltysh: How about also adding CronJob and job here as well?

tnozicka: I'll add them too

test/extended/images/resolve.go
```diff
@@ -128,6 +132,325 @@ var _ = g.Describe("[Feature:ImageLookup][registry][Conformance] Image policy",
...
+
+		g.By("auto replacing local references on Deployments when the annotation is added")
+		if deployment.Annotations == nil {
+			deployment.Annotations = map[string]string{}
+		}
+		deployment.Annotations["alpha.image.policy.openshift.io/resolve-names"] = "*"
+		deployment, err = oc.KubeClient().AppsV1().Deployments(deployment.Namespace).Patch(deployment.Name, types.StrategicMergePatchType, []byte(`{"metadata": {"annotations": {"alpha.image.policy.openshift.io/resolve-names": "*"}}}`))
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(deployment.ObjectMeta.Annotations["alpha.image.policy.openshift.io/resolve-names"]).To(o.Equal("*"))
+		o.Expect(deployment.Spec.Template.Spec.Containers[0].Image).To(o.Equal(tag.Image.DockerImageReference))
```
tnozicka: @soltysh this will fail and prove https://github.com/openshift/origin/blob/8fd3eb390bdd2e1c6057fda7a3ae82c2d2fb2c6d/vendor/github.com/openshift/apiserver-library-go/pkg/admission/imagepolicy/imagereferencemutators/pods.go#L207 is broken because it doesn't respect changing the annotation which is the real reason the deployment in https://bugzilla.redhat.com/show_bug.cgi?id=1801095 never gets the image replaced and hotloops - because it sets the annotation only after creation.

I'll fix the admission tomorrow and vendor it here.

test/extended/images/resolve.go
```diff
@@ -128,6 +132,325 @@ var _ = g.Describe("[Feature:ImageLookup][registry][Conformance] Image policy",
...
+									{Name: "resolve", Image: "busybox:latest", Command: []string{"/bin/true"}},
+								},
+							},
+						},
+					},
+				},
+			},
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		// FIXME: this needs actually fixing the admission first
```
tnozicka: v1beta1.cronjobs are missing in the switch in admission

soltysh: Right, that's why we need to the bump from https://github.com/openshift/apiserver-library-go/pull/23

tnozicka: yep, I have reinstantiated the test after adding the fake bump

test/extended/images/resolve.go
tnozicka: https://github.com/openshift/apiserver-library-go/pull/24

tnozicka:
infra flakes :(
/retest

tnozicka:
iam policy flake
/retest

tnozicka:
https://github.com/openshift/apiserver-library-go/pull/24 merged

tnozicka:
/hold cancel
no `oc set image-lookup` changes seems to be needed as the annotation works the same on the toplevel object if it is put in the object meta or into the template

tnozicka:
IAM flake
/retest

PR #19742, opened 2018-05-16
UPSTREAM: 63926: Avoid unnecessary calls to the cloud provider

Miciah:
If the service controller is processing a service that does not have type LoadBalancer and does not have any external load balancer recorded in its status, do not call the cloud provider to check for an associated load balancer.

This commit fixes bug 1571940.

https://bugzilla.redhat.com/show_bug.cgi?id=1571940

---

Note for backporting: `createLoadBalancersIfNeeded` in 3.9 returns two arguments, so the `return nil` will need to be amended as `return nil, notRetryable` in any backport to 3.9.

PR #24532, opened 2020-02-13
Bug 1798282: DROP: Avoid unnecessary calls to the cloud provider

Miciah:
Drop UPSTREAM: 63926, which causes deletion of "LoadBalancer"-type services to get stuck.

#19742 added the carry in order to reduce unnecessary cloud-provider API calls, but it is no longer needed and causes problems with recent changes to the clean-up logic for services with type "LoadBalancer".

At the time the PR was written, the service controller added every newly created service to its work queue, which was also used for update and delete events.  When a worker for this queue subsequently processed the service, it would check if the service needed an external load-balancer, and if not, it would then check if a load balancer existed for the service (in case the service had been added to the queue by an update that changed its type from "LoadBalancer", in which case any previously provisioned load-balancer would need to be deleted).  To perform this check, the worker would make a `GetLoadBalancer` API call using the cloud provider.

#19742 added a check to skip the `GetLoadBalancer` call if the service's status indicated that it had no associated load-balancer, which should always be the case for a newly created service if its type is not "LoadBalancer".

Later, upstream commit https://github.com/kubernetes/kubernetes/pull/78262/commits/aa3f81d657c1d4b7c789352af7179c08508adc3c modified the service controller's logic to add a newly created service to the work queue only if its type were "LoadBalancer", thereby obviating the need for the check that #19742 had added. The upstream commit also modified the service controller's load-balancer clean-up logic, and the check that #19742 added breaks this new logic.

* `vendor/k8s.io/kubernetes/pkg/controller/service/controller.go` (`syncLoadBalancerIfNeeded`): Delete check for empty load balancer status.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/controller/service/controller.go
+++ b/vendor/k8s.io/kubernetes/pkg/controller/service/controller.go
@@ -42,7 +42,6 @@ import (
 	servicehelper "k8s.io/cloud-provider/service/helpers"
 	"k8s.io/component-base/metrics/prometheus/ratelimiter"
 	"k8s.io/klog"
-	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


Miciah:
In the e2e-gcp-upgrade CI job, both the control-plane-upgrade test and the k8s-service-upgrade test failed.  The control plane uses a load balancer that is not managed by the service controller, so this change cannot be responsible for that failure; I suspect the other failure is a flake.

The other jobs failed before the cluster was up.  In e2e-cmd, the installer failed with the following:

    level=error msg="Error: Error applying IAM policy to project \"openshift-gce-devel-ci\": Too many conflicts.  Latest error: Error setting IAM policy for project \"openshift-gce-devel-ci\": googleapi: Error 409: There were concurrent policy changes. Please retry the whole read-modify-write with exponential backoff., aborted"

In e2e-aws-fips, the job filed with the following:

    could not resolve inputs: could not determine inputs for step [input:machine-os-content-base]: could not resolve base image: imagestreamtags.image.openshift.io "4.4:machine-os-content" is forbidden: User "system:anonymous" cannot get imagestreamtags.image.openshift.io in the namespace "ocp": no RBAC policy matched

/retest

PR #24537, opened 2020-02-13
Bug 1802648: Prefer apps version over deprecated extensions

soltysh:
/assign @deads2k
/hold
for upstream approval


soltysh:
We're not going to fix it.

PR #24540, opened 2020-02-14
Bug 1803163: UPSTREAM: 65357: Allow more fields at root of CRD schema if status is enabled

munnerz:
This PR is a cherrypick from the upstream PR: https://github.com/kubernetes/kubernetes/pull/65357

We're specifically requesting this change be cherrypicked back to release-3.11 as it currently prevents a single set of CRD resources being compatible with OpenShift 3.11 whilst also being considered 'structural' in OpenShift 4.x (and newer Kubernetes versions).

Without this patch, users will see errors like:

```
Error from server (Invalid): error when creating "cert-manager-openshift-v0.13.0.yaml": CustomResourceDefinition.apiextensions.k8s.io "certificaterequests.cert-manager.io" is invalid: spec.validation.openAPIV3Schema: Invalid value: apiextensions.JSONSchemaProps{ID:""

...

ensions.JSONSchemaDefinitions(nil), ExternalDocs:(*apiextensions.ExternalDocumentation)(nil), Example:(*apiextensions.JSON)(nil)}: must only have "properties", "required" or "description" at the root if the status subresource is enabled
```

when installing CRDs that enable the `status` resource that (correctly) specify `type: object` at the root of their CRD schema.

See also: https://github.com/jetstack/cert-manager/issues/2200

This patch applied clean from the upstream PR. This is my first contribution to `origin` so please let me know if I need to make any changes to how I've submitted this 😄 

/assign @sttts

### DIFF
--- a/vendor/k8s.io/kubernetes/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go
+++ b/vendor/k8s.io/kubernetes/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go
@@ -291,7 +291,8 @@ func ValidateCustomResourceDefinitionValidation(customResourceValidation *apiext
 	}
 
 	if schema := customResourceValidation.OpenAPIV3Schema; schema != nil {
-		// if subresources are enabled, only "properties", "required" and "description" are allowed inside the root schema
+		// if the status subresource is enabled, only certain fields are allowed inside the root schema.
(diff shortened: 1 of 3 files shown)


munnerz:
I created a Bugzilla bug against 3.11, and then created another against 4.4 (?) and set it as a dependent of the first Bugzilla bug as per @sttts's help. I cannot set the `State` of either ticket to anything other than CLOSED or NEW so I think that's all I can do here 😅 if there's any additional docs on how this all works that someone can point me to that'd be great 💯

PR #24543, opened 2020-02-14
Bug 1803083: 4.2: UPSTREAM: 87673: blank out value for unbounded client label

sttts:


mfojtik:
next patch manager: this need to merge next week. ping @sttts if you need details.

bparees:
the bug hierarchy on this is not correct.

There should be a 4.2 bug, blocked by a 4.3 bug, blocked by a 4.4 bug.

sttts:
@bparees really? Is that new and required?

bparees:
> @bparees really? Is that new and required?

No it's not new.  If you don't do it in that order, then you could:

1) merge/verify the 4.4 BZ fix
2) merge/verify the 4.2 BZ fix
3) fail to merge/verify the 4.3 fix (possibly not even have a 4.3 BZ tracking the issue)

and now when a customer upgrades from 4.2 to 4.3, they will experience a regression.

We avoid that by ensuring that fixes are merged/verified from the current release first, working backwards to the oldest z-stream.

PR #24545, opened 2020-02-14
Bug 1803090: 3.11: UPSTREAM: 87673: blank out value for unbounded client label

sttts:


sdodson:
@sttts @mfojtik Do we want this? If so lets override the bugzilla/valid-bug labels. There are no future 4.1 releases scheduled and we should probably just close the 4.1 BZ and PR.

sttts:
Fine with me.

/close

sttts:
@sdodson this is 3.11, not 4.1. Please take another look.

sdodson:
> @sdodson this is 3.11, not 4.1. Please take another look.

Yeah that's what I meant. Since 4.1 is EOL'd that PR should be closed, but this PR shouldn't be constrained by valid-bug at least in so much as it depends on a 4.1 bug.

/test e2e-aws

sdodson:
e2e-gcp is broken right now, https://bugzilla.redhat.com/show_bug.cgi?id=1828484
/override ci/prow/e2e-gcp

PR #24547, opened 2020-02-14
Bug 1803190: test/extended/prometheus: temporarily disable etcdInsufficientMembers

hexfusion:
In order to merge the `cluster-etcd-operator`, we need to disable `etcdInsufficientMembers` as we pivot from the old etcd-member static pod. This will be followed by the removal of the etcd-member spec from MCO and reenabling this alert.


hexfusion:
@lilic thanks updated

deads2k:
Terraform
/retest

PR #24555, opened 2020-02-17
Eliminate use of tags to determine OS_GIT_VERSION

jupierce:
The reliance on tags to determine OS_GIT_VERSION means that
origin/ose needed to be treated as snowflakes. To allow
us to use pure fast-forwading, version information should
be supplied at build time instead of being detected through
tags.


jupierce:
This was addressed by avoiding the use of the custom tito logic entirely.

PR #24562, opened 2020-02-18
Bug 1804429: Revert: test/extended/prometheus: temporarily disable etcdInsufficientMembers"

hexfusion:
As promised we are returning the alert.

Reverts openshift/origin#24547


hexfusion:
/bugzilla refresh

cc @deads2k @lilic

PR #24564, opened 2020-02-19
kubelet: Make default journal format precise

smarterclayton:
short-precise gives us nanoseconds which makes sorting accurate.


smarterclayton:
4.3 bug is 1804516 and 4.4 bug is 1804514

sjenning:
fyi @rphillips

PR #24565, opened 2020-02-19
test/extended/prometheus: Catch when we exceed the limit of series sent

lilic:
We want to know when we go over the limit of series we send via telemetry. This helps us control the cardinality of the series that are sent. This was designed in mind to also work when OLM Red Hat operators have a job where they install all the red hat operators as that is the only time we can test how many series they produce.

Note: The actual number is up for discussion. Maybe we should even say 400 so we know when we are approaching a limit of 500 we set to not go over.


lilic:
/hold

until I test with lower limit to make sure it works as expected.

lilic:
Failed test confirmed: :) 

> [Feature:Prometheus][Late] Alerts [Top Level] [Feature:Prometheus][Late] Alerts shouldn't exeeced the 500 series limit of total series sent via telemetry from each cluster [Suite:openshift/conformance/parallel] expand_less	1m31s
fail [github.com/openshift/origin/test/extended/prometheus/prometheus_builds.go:163]: Expected
    <map[string]error | len:1>: {
        "max_over_time(cluster:telemetry_selected_series:count[2h]) >= 10": {
            s: "promQL query: max_over_time(cluster:telemetry_selected_series:count[2h]) >= 10 had reported incorrect results:\n[{\"metric\":{},\"value\":[1582123651.357,\"277\"]}]",
        },
    }
to be empty

lilic:
/retest

Seems like flakes.

lilic:
/hold cancel

I would like to discuss the number of series we would want to cap to here, I think when we hit 400 we should start looking at things, not at 500. We already noticed a problem on clusters with large amount of nodes. Currently we are at around 200ish total number of series we send on a small cluster.

@openshift/openshift-team-monitoring This should be ready for review. Also @smarterclayton need a review from you because of test/extended/util/annotate/ file that this PR touches.

s-urbaniak:
looking good so far, in a vanilla openshift cluster (3 masters, 3 workers) i get ~260 series via telemetry, so 500 seems to have safe margin here for e2e.

image

test/extended/prometheus/prometheus.go
```diff
@@ -76,6 +78,27 @@ var _ = g.Describe("[Feature:Prometheus][Late] Alerts", func() {
 
 		e2e.Logf("Watchdog alert is firing")
 	})
+
+	g.It("shouldn't exeeced the 500 series limit of total series sent via telemetry from each cluster", func() {
```
simonpasquier: s/exeeced/exceed/

test/extended/prometheus/prometheus.go
```diff
@@ -76,6 +78,27 @@ var _ = g.Describe("[Feature:Prometheus][Late] Alerts", func() {
...
+		defer func() { oc.AdminKubeClient().CoreV1().Pods(ns).Delete(execPod.Name, metav1.NewDeleteOptions(1)) }()
+
+		tests := map[string]bool{
+			// We want to limit the number of total series sent, the cluster:telemetry_selected_series:count
+			// rule contains the count of the all the series that are sent via telemetry.
+			`max_over_time(cluster:telemetry_selected_series:count[2h]) >= 500`: false,
+		}
+		runQueries(tests, oc, ns, execPod.Name, url, bearerToken)
+
+		e2e.Logf("Total number of series sent via telemetry is below the limit")
```
simonpasquier: could it be useful to log the actual number?

simonpasquier: scratch that, it would require another query...

lilic: Yeah its easy to look at the Prometheus dump in the CI artefacts afterwards to find the actual number I think.

test/extended/prometheus/prometheus.go
lilic: Spelled it correctly in the commit lol 🤦‍♀

lilic:
Done @simonpasquier PTAL, thanks!

lilic:
/retest 

CI 🤷‍♀

lilic:
/retest

Prow was broken....

test/extended/prometheus/prometheus.go
simonpasquier: :+1:

lilic:
meh

/retest

lilic:
/retest

Lets see if CI is feeling better today.

lilic:
/retest

flakes again...

/assign @derekwaynecarr

lilic:
Ping again @derekwaynecarr or @smarterclayton PTAL you are the only two in the OWNERS file, tests passed it has an lgtm, just needs approve. Thanks!

PR #24567, opened 2020-02-19
Bug 1804843: Useless Test Change

eparis:


eparis:
I just set the 4.3.z BZ VERIFIED so
/bugzilla refresh

eparis:
The parent bug is now VERIFIED but target release 4.4.0 so
/bugzilla refresh

eparis:
and that just wrong :-(

AlexNPavel:
(Testing a new bugzilla command that was just added)
/bugzilla assign-qa

AlexNPavel:
Looks like it's fixed

PR #24568, opened 2020-02-19
Bug 1802687: UPSTREAM: 88251: Partially fix incorrect configuration of kubepods.slice unit by kubelet

rphillips:
Fixes node allocations on the kubepods.slice on Kubelet startup.

ref: https://github.com/kubernetes/kubernetes/pull/88251
ref: https://github.com/kubernetes/kubernetes/issues/88197

/cc @sjenning @smarterclayton


rphillips:
upstream patch should merge shortly... holding until that happens.

/hold

PR #24570, opened 2020-02-20
test: Allow sequential upgrades to be passed to upgrade test

smarterclayton:
Upgrade test now accepts comma-delimited images or versions in
`openshift-tests run-upgrade --image A,B`. This simplifies bulk testing of
upgrades and will allow us to test existing upgrade paths.

This will allow us to test upgrades over multiple releases more easily
and allows us to avoid testing at ecah upgrade interval.

Will be used in 4.4 to test multiple input upgrades.


smarterclayton:
4.4 Bug is 1805457 and 4.3 is 1805459.  I'll have to cherry-pick a few changes into it.

PR #24574, opened 2020-02-20
[WIP]Image prune test enablement

wzheng1:
This is for https://issues.redhat.com/browse/DEVEXP-434 and still working in progress.

### DIFF
--- a/test/extended/images/hardprune.go
+++ b/test/extended/images/hardprune.go
@@ -364,14 +364,14 @@ func (ba byAgeDesc) Less(i, j int) bool {
 
 // GetRegistryPod returns the youngest registry pod deployed.
 func GetRegistryPod(podsGetter kcoreclient.PodsGetter) (*kapiv1.Pod, error) {
-	podList, err := podsGetter.Pods(metav1.NamespaceDefault).List(context.Background(), metav1.ListOptions{
-		LabelSelector: labels.SelectorFromSet(labels.Set{"deploymentconfig": "docker-registry"}).String(),
+	podList, err := podsGetter.Pods(RegistryOperatorDeploymentNamespace).List(context.Background(), metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{"docker-registry": "default"}).String(),
 	})
 	if err != nil {
 		return nil, err
 	}
 	if len(podList.Items) == 0 {
-		return nil, fmt.Errorf("failed to find any docker-registry pod")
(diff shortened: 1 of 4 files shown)


vendor/github.com/onsi/gomega/go.mod
```diff
@@ -13,3 +13,5 @@ require (
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	gopkg.in/yaml.v2 v2.2.1
 )
+
+go 1.13
```
adambkaplan, line 17: @wzheng1 please revert this (not clear why it is here).

test/extended/images/hardprune.go
```diff
@@ -23,323 +23,163 @@ import (
...
+	//imagesutil "github.com/openshift/origin/test/extended/images"
+
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
 var _ = g.Describe("[sig-imageregistry][Feature:ImagePrune][Serial][Suite:openshift/registry/serial][Local] Image hard prune", func() {
 	defer g.GinkgoRecover()
 	var oc = exutil.NewCLI("prune-images")
-	var originalAcceptSchema2 *bool
-	ctx := context.Background()
```
adambkaplan, line 33: For a first attempt at restoring these tests, I would not worry about hard pruning. Eventually we'll want this test back, but it isn't as pressing as ensuring the core pruning functions work.

@dmage would you agree?

dmage: Yes, I agree.

I'm fine to slightly update this test even if it stays Skipped if that helps to remove legacy code. We'll never have the docker socket anyway.

wzheng1: OK, I will not spend too much effort on hard pruning now. Should I revert my changes?

dmage: Let's focus on `oc adm prune images` and remove hard pruner changes from this PR. Keep them for a follow-up PR.

btw, you shouldn't have back merges in your PR, so you'll need to squash commits.

vendor/github.com/onsi/gomega/go.mod
wzheng1: Have reverted.

test/extended/images/hardprune.go
wzheng1: OK, I have restored most part of hard prune, just leave some basic functions change which maybe used by prune.go.

PR #24584, opened 2020-02-21
test: Set upgrade service LB timeouts for graceful shutdown

smarterclayton:
Each platform has different settings. Make the different platform
configurations equal where possible and tolerate the longest
possible duration.


smarterclayton:
After making these changes, only existing connections to the pod are impacted during SDN upgrade, not new connections.  It appears that node drain is still causing disruption to workers even with these settings.

PR #24585, opened 2020-02-21
Bug 1765220: Fix pod startup reattach/kill loop

abhat:
We should kill every (non-hostNetwork) sandbox that we can't reattach
to OVS, but previously we were only killing "every (non-hostNetwork)
sandbox that we knew was previously attached to OVS that we couldn't
reattach to OVS".

This is a backport of https://github.com/openshift/sdn/pull/58 and https://github.com/openshift/sdn/pull/52 for 3.11.

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -99,7 +99,9 @@ type OsdnNode struct {
 	oc                 *ovsController
 	networkInfo        *common.NetworkInfo
 	podManager         *podManager
+	clusterCIDRs       []string
 	localSubnetCIDR    string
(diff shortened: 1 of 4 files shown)


danwinship:
/assign
please fix the commit messages to have the original commit messages. (You can use `git cherry-pick -x` if you want it to add a reference to the original commit in the commit message)

also, please link to the original PR(s) that this is a backport of from the initial github comment

abhat:
This is a backport of https://github.com/openshift/sdn/pull/58 and https://github.com/openshift/sdn/pull/52 for 3.11.

PR #24593, opened 2020-02-24
test: Allow more control plane disruption for multi upgrade releases

smarterclayton:
When running sequential updates, the control plane appears to take
more disruption than can be currently tolerated. Bump the limit for
now until we identify the root cause of the disruption.


smarterclayton:
Merging so we can begin the backport for multistage upgrades

PR #24594, opened 2020-02-24
Bug 1799491: Don't instrument etcd operator

kevinrizza:
An OLM extended test currently installs the community etcd operator then
applies an `etcdCluster` cr. This test is somewhat orthogonal to testing
OLM itself, and it is currently disabled in Azure environments due to a
performance issue with the community etcd operator and azure clusters.

This commit updates the test. It still installs the etcd operator, but
merely checks to see that OLM was able to resolve the operator install
then returns successfully. Additionally, it reenables the test in Azure
environments.

https://bugzilla.redhat.com/show_bug.cgi?id=1799491


jmrodri:
This is blocking my 4.5 service catalog removal feature one of the tests that's failing. https://github.com/openshift/cluster-svcat-apiserver-operator/pull/74

PR #24595, opened 2020-02-24
Bug 1806698: Limit-based eviction should report metrics

smarterclayton:
Limit-based eviction has the same outcome as a capacity-based eviction
but was not being reported to the kubelet_evictions metric, which makes
it harder to alert on. Report those evictions to the metric with
appropriate signals.


smarterclayton:
4.4 is bug 1806699

PR #24597, opened 2020-02-25
Bug 1805821: change restart policy of command pod to RestartPolicyOnFailure

vareti:
This PR aims to cut down the CI flakes for the LDAP test.

There are a good number of failures because of connection failures due to race conditions between the LDAP client and the LDAP server. Using `RestartPolicyOnFailure` instead of `RestartPolicyNever` will make sure the pod command completes by restarting the pod in case of transient failures.

The same ldap container image is being used for both server and client. The client implementation is not under our control for the pod to be resilient to transient failures. To cover these failures restarting pod on failure seems to be a good option.

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -1648,7 +1648,8 @@ func RunOneShotCommandPod(
 		}
 
 		if podHasErrored(cmdPod) {
-			return true, fmt.Errorf("the pod errored trying to run the command")
+			e2e.Logf("pod %q errored trying to run the command: %v", pod.Name, err)
+			return false, nil
 		}
 		return podHasCompleted(cmdPod), nil
 	})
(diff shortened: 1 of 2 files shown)


vareti:
doing a retest to see if we get ldap failures

/test e2e-gcp

vareti:
build failure

/test e2e-gcp

test/extended/util/framework.go
```diff
@@ -1648,7 +1648,8 @@ func RunOneShotCommandPod(
 		}
 
 		if podHasErrored(cmdPod) {
-			return true, fmt.Errorf("the pod errored trying to run the command")
```
stlaz, line 1651: nit: `pod %q errored...`

stlaz:
/approve
(not sure I can do that)

test/extended/util/framework.go
stlaz, line 1651: nit: `pod %q errored...`

vareti: TIL about `%q` :+1:

PR #24601, opened 2020-02-25
Bug 1711674: Add pv mount failure event

gnufied:
Add a even to PV when filesystem mount fails because of mismatch between pv's filesystem and actual filesystem on disk.

This will help us in alerting about problems like - https://bugzilla.redhat.com/show_bug.cgi?id=1711674

/sig storage
cc @jsafrane @knobunc


jsafrane:
/retitle Bug 1711674: Add pv mount failure event

jsafrane:
@gnufied, was this fixed by 1.18 rebase? I can see `checkForFailedMount` in current origin/master: https://github.com/openshift/origin/blob/16abec0d471f3c40e04622210edba33d43f21704/vendor/k8s.io/kubernetes/pkg/volume/util/operationexecutor/operation_generator.go#L696-L706

gnufied:
Yes. we can close this. This should be fixed by rebase.

PR #24614, opened 2020-03-02
Bug 1807990: UPSTREAM: 88663: update kube-controller-manager and kube-scheduler to match kube-apiserver defaults

soltysh:
Clean pick with the exception that the variable in `vendor/k8s.io/kubernetes/cmd/kube-scheduler/app/options/options.go` changed from `versionedCfg` to `cfgv1alpha1`. 

/assign @deads2k


deads2k:
upstream merged

PR #24616, opened 2020-03-02
Bug 1760807: Add check for PodAffinity value validity in Deployments

damemi:
This attempts to strengthen Deployment validation by checking that pod affinity terms contain valid values, to prevent unexpected behavior/errors later.

### DIFF
--- a/vendor/k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
+++ b/vendor/k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
@@ -49,6 +49,11 @@ func ValidateLabelSelectorRequirement(sr metav1.LabelSelectorRequirement, fldPat
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("operator"), sr.Operator, "not a valid selector operator"))
 	}
 	allErrs = append(allErrs, ValidateLabelName(sr.Key, fldPath.Child("key"))...)
+	for _, v := range sr.Values {
+		for _, msg := range validation.IsValidLabelValue(v) {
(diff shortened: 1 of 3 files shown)


damemi:
/hold
for review and testing

damemi:
Note: I haven't been able to validate this in my local system as I'm having difficulty getting a custom 3.11 image built and spun up. Any help verifying this before merging would be great

vendor/k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/validation/validation.go
```diff
@@ -49,6 +49,11 @@ func ValidateLabelSelectorRequirement(sr metav1.LabelSelectorRequirement, fldPat
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("operator"), sr.Operator, "not a valid selector operator"))
 	}
 	allErrs = append(allErrs, ValidateLabelName(sr.Key, fldPath.Child("key"))...)
+	for _, v := range sr.Values {
```
soltysh, line 52: I don't think this is the right approach, since this might introduce inconsistencies with newer versions.

damemi: Looking at this further, I think this is the error and it needs to be ported to all available versions (see https://github.com/kubernetes/kubernetes/pull/92714). There is currently no guaranteed handling of an invalid selector and this should be caught at creation time

damemi:
/close
https://github.com/openshift/origin/pull/25350 is probably a better implementation of this that won't break on resource updates (see upstream discussion in https://github.com/kubernetes/kubernetes/pull/92714)

PR #24622, opened 2020-03-03
Bug 1809709: Don't send no-op pod status patches

smarterclayton:
The kubelet was sending no-op patches unncessarily about half the
time. This dramatically reduces the amount of API requests large
clusters send with some CPU reduction.


smarterclayton:
4.4 bug is 1809712

rphillips:
`generic build failure`
/retest

PR #24624, opened 2020-03-03
[4.3] Bug 1808548: UPSTREAM: 88146: remove duplicate pv delete commands

gnufied:
https://bugzilla.redhat.com/show_bug.cgi?id=1808548

/sig storage


jsafrane:
/retitle [4.3] Bug 1808548: UPSTREAM: 88146: remove duplicate pv delete commands

jsafrane:
/bugzilla refresh
just in case ON_QA is OK...

PR #24625, opened 2020-03-03
[4.4] Bug 1796660: UPSTREAM: 88146: remove duplicate pv delete calls

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1796660

/sig storage


jsafrane:
/retitle [4.4] Bug 1796660: UPSTREAM: 88146: remove duplicate pv delete calls

PR #24629, opened 2020-03-04
Bug 1810008: Support TLS Server Name overrides in kubeconfig file

tnozicka:
Backport of https://github.com/kubernetes/kubernetes/pull/88769

We need to reliably plumb TLS SNI name into all clients in CKASO, CKCMO, CKSO. This allows to do it cleanly and directly in kubeconfig which makes sure it is plumbed into all clients even those hidden like getting CA configmap for extended auth or leader election.

/cc @soltysh @deads2k


deads2k:
/hold

Not like this. Do not couple our operators to origin.  Instead, you can create a different fork of client-go if you wish for those operators.

/close

PR #24630, opened 2020-03-04
[release-4.4] Bug 1771410: Support TLS Server Name overrides in kubeconfig file

tnozicka:
backporting https://github.com/openshift/origin/pull/24629

/cc @soltysh @deads2k


tnozicka:
/close
we go for fork

PR #24632, opened 2020-03-04
test: Add a delay between successive retries for service reachability check and wait for n consecutive successes.

abhat:
This change adds a `framework.Poll` delay between successive retries for reachability check for the test service. This will make sure the SLB is stable before we start disruption.

### DIFF
--- a/test/e2e/upgrade/service/service.go
+++ b/test/e2e/upgrade/service/service.go
@@ -96,9 +96,17 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
 	// Hit it once before considering ourselves ready
 	ginkgo.By("hitting pods through the service's LoadBalancer")
 	timeout := service.LoadBalancerLagTimeoutAWS
-	// require five passing requests to continue (in case the SLB becomes available and then degrades)
-	for i := 0; i < 5; i++ {
-		service.TestReachableHTTP(tcpIngressIP, svcPort, timeout)
+	// require thirty seconds of passing requests to continue (in case the SLB becomes available and then degrades)
+	for successCount := 0; successCount < 30; successCount++ {
+		// sleep for 1s between successive tries.
+		time.Sleep(1 * time.Second)
+		func() {
+			defer func() {
+				recover()
+				successCount = 0
+			}()
+			service.TestReachableHTTP(tcpIngressIP, svcPort, timeout)
+		}()
 	}
 
 	t.jig = jig
--- a/test/e2e/upgrade/service/service.go
+++ b/test/e2e/upgrade/service/service.go
@@ -11,17 +11,18 @@ import (
 
 	"github.com/onsi/ginkgo"
 
+	"github.com/openshift/origin/pkg/monitor"
+	"github.com/openshift/origin/test/extended/util/disruption"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/events"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2enetwork "k8s.io/kubernetes/test/e2e/framework/network"
 	"k8s.io/kubernetes/test/e2e/framework/service"
 	"k8s.io/kubernetes/test/e2e/upgrades"
-
-	"github.com/openshift/origin/pkg/monitor"
-	"github.com/openshift/origin/test/extended/util/disruption"
 )
 
 // UpgradeTest tests that a service is available before, during, and
@@ -97,17 +98,7 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
 	ginkgo.By("hitting pods through the service's LoadBalancer")
 	timeout := service.LoadBalancerLagTimeoutAWS
 	// require thirty seconds of passing requests to continue (in case the SLB becomes available and then degrades)
-	for successCount := 0; successCount < 30; successCount++ {
-		// sleep for 1s between successive tries.
-		time.Sleep(1 * time.Second)
-		func() {
-			defer func() {
-				recover()
-				successCount = 0
-			}()
-			service.TestReachableHTTP(tcpIngressIP, svcPort, timeout)
-		}()
-	}
+	TestReachableHTTPWithMinSuccessCount(tcpIngressIP, svcPort, 30, timeout)
 
 	t.jig = jig
 	t.tcpService = tcpService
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade/service/service.go
```diff
@@ -96,9 +96,17 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
...
+	// require thirty seconds of passing requests to continue (in case the SLB becomes available and then degrades)
+	for successCount := 0; successCount < 30; successCount++ {
+		// sleep for 1s between successive tries.
+		time.Sleep(1 * time.Second)
+		func() {
+			defer func() {
+				recover()
+				successCount = 0
+			}()
+			service.TestReachableHTTP(tcpIngressIP, svcPort, timeout)
```
smarterclayton: Unfortunately not the same, the loop I was talking about hoisting was TestReachableHTTPWithRetriableErrorCodes

```
func TestReachableHTTPWithRetriableErrorCodes(host string, port int, retriableErrCodes []int, timeout time.Duration) {
	pollfn := func() (bool, error) {
		result := e2enetwork.PokeHTTP(host, port, "/echo?msg=hello",
			&e2enetwork.HTTPPokeParams{
				BodyContains:   "hello",
				RetriableCodes: retriableErrCodes,
			})
		if result.Status == e2enetwork.HTTPSuccess {
			return true, nil
		}
		return false, nil // caller can retry
	}

	if err := wait.PollImmediate(framework.Poll, timeout, pollfn); err != nil {
		if err == wait.ErrWaitTimeout {
			framework.Failf("Could not reach HTTP service through %v:%v after %v", host, port, timeout)
		} else {
			framework.Failf("Failed to reach HTTP service through %v:%v: %v", host, port, err)
		}
	}
}
```

smarterclayton: just take all of this code and move it up into a function in this, and call it TestReachableWithMinimumSuccess

abhat: I didn't want to duplicate code especially since this is vendored from k/k. Do you think the above loop will work?

smarterclayton: I want to duplicate code.  Kube code isn't special.  We already split half the e2es because the upstream code is unreliable and not fit for purpose.

abhat: Ok, I will create a separate function like `TestReachableWithMinimumSuccess` and use that.

test/e2e/upgrade/service/service.go
```diff
@@ -259,3 +258,31 @@ func startEndpointMonitoring(ctx context.Context, m *monitor.Monitor, svc *v1.Se
...
+				BodyContains:   "hello",
+				RetriableCodes: []int{},
+			})
+		if result.Status == e2enetwork.HTTPSuccess {
+			return true, nil
+		}
+		return false, nil // caller can retry
+	}
+
+	for retryCount := 0; retryCount < successCount; retryCount++ {
```
smarterclayton: I was expecting this to be slightly different: no loop here, and

```
pollFn := ... {
  PokeHTTP
  if successful {
    consecutiveSuccess++
    return consecutiveSuccess > successCount, false
  }
  consecutiveSuccess = 0
  return false, nil
}

err := wait.PollImmediate(...)
expectNoFailure(err)
```

smarterclayton: We want to get X consecutive successes, and we want timeouts to cause us to exit with an error (the test should fail if we can't get X consecutive before timeout.

abhat: if we just try `n` times, that's not going to be sufficient. it has to be with some `sleep` in between. Also, the timeOut passed to the wait function is with a single request in mind. If we are going to let that govern the `n` successes, it ought to be accounting for those many requests. The current logic let's the timeout stay as is, and wraps around each request as a whole.

smarterclayton: What I suggested would be part of the poll loop, so even a success would wait framework.Poll in between.

smarterclayton: let me pull this up in a single example:

```
var conditionSuccess int
err := wait.PollImmediate(..., func() (bool, error) {
  result := e2enetwork.PokeHTTP(host, port, "/echo?msg=hello",
    &e2enetwork.HTTPPokeParams{
      BodyContains:   "hello",
      RetriableCodes: []int{},
    })
  if result.Status == e2enetwork.HTTPSuccess {
    conditionSuccess++
    return conditionSuccess >= successCount, nil
  }
  conditionSuccess = 0
  return false, nil
})
expectNoError(err)
```

smarterclayton: This requires successCount consecutive successes, each spaced by framework.Poll, within duration.

test/e2e/upgrade/service/service.go
```diff
@@ -261,28 +261,28 @@ func locateService(svc *v1.Service) string {
...
-	for retryCount := 0; retryCount < successCount; retryCount++ {
-		if err := wait.PollImmediate(framework.Poll, timeout, pollfn); err != nil {
-			if err == wait.ErrWaitTimeout {
-				framework.Logf("Could not reach HTTP service through %v:%v after %v", host, port, timeout)
-			} else {
-				framework.Logf("Failed to reach HTTP service through %v:%v: %v", host, port, err)
-			}
-			retryCount = 0
+	if err := wait.PollImmediate(framework.Poll, timeout, pollfn); err != nil {
+		if err == wait.ErrWaitTimeout {
```
smarterclayton: You need to error if poll immediate fails.  That's why my previous comment inlined the poll function (which is weird anyway).

test/e2e/upgrade/service/service.go
```diff
@@ -261,28 +261,28 @@ func locateService(svc *v1.Service) string {
...
 		result := e2enetwork.PokeHTTP(host, port, "/echo?msg=hello",
 			&e2enetwork.HTTPPokeParams{
 				BodyContains:   "hello",
 				RetriableCodes: []int{},
 			})
 		if result.Status == e2enetwork.HTTPSuccess {
-			return true, nil
+			consecutiveSuccessCnt++
+			return consecutiveSuccessCnt >= successCount, nil
+		} else {
```
smarterclayton: don't do else's after an if with a return

test/e2e/upgrade/service/service.go
abhat: I thought `framework.Failf` will cause panic and the test to fail?

test/e2e/upgrade/service/service.go
abhat: What is the motivation to inline the pollfn? Is there a gotcha here that `err` will get over-written if the poll func is not inline?

smarterclayton:
LGTM, go ahead and squash

abhat:
@smarterclayton can I get an lgtm on this?

abhat:
@smarterclayton this looks good to go. e2e-aws is passing now.

smarterclayton:
Running two AWS upgrade jobs

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade/21278
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade/21279

to verify this fixes

PR #24637, opened 2020-03-05
BUG 1810446: UPSTREAM: 87591: set nil cache entry based on old cache

enxebre:
1.17 cherry-pick of https://github.com/kubernetes/kubernetes/pull/87531

Set cache entries based on old cache to prevent aggressive cloud provider calls during scale up and scale down operations.

Part of kubernetes-sigs/cloud-provider-azure#247

Note that although this is sane to back port I don't believe if should have a big impact as we don't use VMSS


enxebre:
cc @jim-minter

PR #24640, opened 2020-03-05
Release 3.9

jihongbo:
wordpress  examples


adambkaplan:
@jihongbo what are you trying to restore into origin?

Most openshift code exists in other repositories. Wordpress have been dropped, but there are plenty of imagestreams and templates in [openshift/library](https://github.com/openshift/library). These are then added to OpenShift via the [cluster-samples-operator](https://github.com/openshift/cluster-samples-operator).

PR #24641, opened 2020-03-05
Bug 1808568: must-gather: when openshift-apiserver restarts, ignore wrong prefix

sallyom:
/assign @soltysh 
/cc @smarterclayton 

example failure logs where this flake occurs:
```
572 ift:discovery\" to Group \"system:authenticated\""}}
573 {"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Metadata","auditID":"2d7c58d5-f230-4097-bd5e-620e9ca3ec24","stage":"ResponseComplete","requestUR      I":"/openapi/v2","verb":"get","user":{"username":"system:aggregator","groups":["system:authenticated"]},"sourceIPs":["10.128.0.1"],"responseStatus":{"me      tadata":{},"code":304},"requestReceivedTimestamp":"2020-03-04T19:04:57.786208Z","stageTimestamp":"2020-03-04T19:04:57.786270Z","annotations":{"authoriza      tion.k8s.io/decision":"allow","authorization.k8s.io/reason":"RBAC: allowed by ClusterRoleBinding \"system:openshift:discovery\" of ClusterRole \"system:      openshift:discovery\" to Group \"system:authenticated\""}}
```

### DIFF
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -125,6 +125,7 @@ var _ = g.Describe("[cli] oc adm must-gather", func() {
 		// make sure we do not log OAuth tokens
 		for _, auditDirectory := range auditDirectories {
 			eventsChecked := 0
+			wrongPrefix := 0
 			err := filepath.Walk(path.Join(auditDirectory...), func(path string, info os.FileInfo, err error) error {
 				g.By(path)
 				o.Expect(err).ToNot(o.HaveOccurred())
@@ -147,7 +148,9 @@ var _ = g.Describe("[cli] oc adm must-gather", func() {
 					if !strings.HasSuffix(text, "}") {
 						continue // ignore truncated data
 					}
-					o.Expect(text).To(o.HavePrefix(`{"kind":"Event",`))
+					if !strings.HasPrefix(text, `{"kind":"Event",`) {
+						wrongPrefix++
+					}
 					for _, token := range []string{"oauthaccesstokens", "oauthauthorizetokens", tokenName} {
 						o.Expect(text).ToNot(o.ContainSubstring(token))
 					}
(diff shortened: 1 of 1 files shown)


deads2k:
/hold

This is a symptom of the openshift-apiserver not correctly flushing its audit log to disk or must-gather not correctly retrieving content.  Either way that this happened, do we know why it suddenly became a problem?

soltysh:
> This is a symptom of the openshift-apiserver not correctly flushing its audit log to disk or must-gather not correctly retrieving content. Either way that this happened, do we know why it suddenly became a problem?

From my investigation this was happening at the time when openshift-apiserver was restarting, but back then it wasn't such a big issue. Not sure why this e2e flake bubbled up recently this high. I agree that we should investigate the root cause, but at the same time, I'd prefer make this test less strict. We can make a separate BZ about openshift-apiserver that will need further investigation and make the test more resilient in parallel. @deads2k objections?
@sallyom can you make a separate BZ for that audit issue?

sallyom:
I've opened https://bugzilla.redhat.com/show_bug.cgi?id=1811737  to track root cause

deads2k:
> From my investigation this was happening at the time when openshift-apiserver was restarting, but back then it wasn't such a big issue. Not sure why this e2e flake bubbled up recently this high. I agree that we should investigate the root cause, but at the same time, I'd prefer make this test less strict. We can make a separate BZ about openshift-apiserver that will need further investigation and make the test more resilient in parallel. @deads2k objections?

Yeah, this exposed a problem a critical data corruption problem with the audit log that we fixed with https://github.com/openshift/cluster-openshift-apiserver-operator/pull/331 . Have you see it since then?

sallyom:
this is no longer required, AFAICT, since https://github.com/openshift/cluster-openshift-apiserver-operator/pull/331 resolved root cause

PR #24651, opened 2020-03-06
e2e-disruptive: add test for deleting and recreating a master

alaypatel07:
This PR adds etcd member health check on top of #24643 
The flow is:

1. select a master and a worker node to delete force shutdown
2. set up machine-health-checks for both master and worker nodes, 
so after shutdown the machines get deleted.
2. force shutdown the nodes.
3. remove etcd member corresponding to master node shutdown.
4. wait for the machines to be deleted
5. recreate master machine
6. wait for both master machine and worker machine to come back
7. while all of the above is happening, make sure kube API and 
Openshift API is up.

The work is being blocked by the following potential theory:

the MHC's are for 5 minutes. After 5 mins of shutting down master
 node will be deleted, but the node will wait for 5 more mins for draining
before being deleted. The step 7 of making sure that kube API is healthy, 
times out at 7mins if it does not find all the three master nodes. This overlap
of timing makes the test to fail locally


test/extended/dr/machine_recover.go
```diff
@@ -155,7 +175,102 @@ var _ = g.Describe("[Feature:DisasterRecovery][Disruptive]", func() {
...
+				framework.Logf("Verify recreated master has healthy etcd endpoint")
+				healthyEtcds = getHealthyEtcds(certsDir, masters)
+				o.Expect(len(healthyEtcds)).To(o.BeNumerically("==", len(masters)))
+
 			})
 	},
 	)
 })
+
+func getHealthyEtcds(certsDir string, masters []*corev1.Node) []*etcdserverpb.Member {
```
alaypatel07: IMO these helper functions should be in common.go, will add a TODO

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,276 @@
...
+
+	err = ioutil.WriteFile(path.Join(dir, "tls.key"), []byte(clientKey), 0600)
+	o.Expect(err).NotTo(o.HaveOccurred())
+}
+
+func getEtcdEndpoints(masters []*corev1.Node) []string {
+	endpoints := []string{}
+	for _, m := range masters {
+		for _, addr := range m.Status.Addresses {
+			if addr.Type == corev1.NodeInternalIP {
```
alaypatel07: From where tests are run do we have access to the node network? If not we would have to ssh into the master and run etcd health check

alaypatel07: I see `expectSSH` function using node's `ExternalIP` to ssh and if not available, using the `InternalIP`. Can we use the same strategy for a non-ssh port i. e 2379?

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	cli, err := clientv3.New(*cfg)
+	if err != nil {
+		return nil, err
+	}
+	return cli, err
+}
+
+func dumpEtcdCertsOnDisk(oc *exutil.CLI, dir string) {
```
alaypatel07: is there an easier way to get the certs on disk? the operator pod mounts the confimap and secret, is that option available for where this test is run?

test/extended/dr/machine_recover.go
smarterclayton: Note that we don't want big bundles of helpers - there's not a ton of overlap between this method and the other methods, so placing helpers specific to a specific location is ok for now and refactoring later is fine.

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,276 @@
...
+				masters, workers := clusterNodes(oc)
+				o.Expect(len(masters)).To(o.BeNumerically(">=", 3))
+				o.Expect(len(workers)).To(o.BeNumerically(">=", 2))
+
+				framework.Logf("Verify etcd endpoints are healthy")
+				certsDir := "/tmp/etcd-client-certs/"
+				dumpEtcdCertsOnDisk(oc, certsDir)
+				defer os.RemoveAll(certsDir)
+				healthyEtcds := getHealthyEtcds(certsDir, masters)
+				o.Expect(len(healthyEtcds)).To(o.BeNumerically("==", len(masters)))
```
smarterclayton: Might be better to gate this on 3 (we expect a quorum before we start) rather than number of masters, since number of masters might be 4.

smarterclayton: No unhealthy etcd, must have at least 3, or something.

alaypatel07: good idea will modify to check:
1. 0 unhealthy before starting
2. 1 unhealthy after the master is destroyed
3. 0 unhealthy after master recreated

alaypatel07: modified

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,309 @@
...
+	},
+	)
+})
+
+func createMachineHealthCheckForRole(oc *exutil.CLI, role string) {
+	mhc := oc.DynamicClient().Resource(schema.GroupVersionResource{
+		Group:    "machine.openshift.io",
+		Version:  "v1beta1",
+		Resource: "machinehealthchecks",
+	}).Namespace("openshift-machine-api")
```
alaypatel07: not sure what namespace the health checks should be created in

alaypatel07:
/retest

ugh typo in previous comment :/

test/extended/dr/machine_recover.go
smarterclayton: this is fine for now.  Remember to delete them after the test.

smarterclayton:
in cmd/openshift-tests/e2e.go replace this line (just for testing this PR)

			return strings.Contains(name, "[Disruptive]") && strings.Contains(name, "[Feature:EtcdRecovery]")


with 

			return strings.Contains(name, "[Disruptive]") && strings.Contains(name, "master and worker failure and recover with machine health checks")

Then we can run /test e2e-aws-disruptive on this PR

test/extended/dr/machine_recover.go
alaypatel07: added the logic to delete after the test, thanks for spotting

alaypatel07:
```
Mar 15 08:08:54.615: INFO: Verify etcd endpoints are healthy
I0315 08:08:54.675272     880 client.go:361] parsed scheme: "endpoint"
I0315 08:08:54.675377     880 endpoint.go:68] ccResolverWrapper: sending new addresses to cc: [{https://10.0.135.104:2379 0  <nil>} {https://10.0.140.66:2379 0  <nil>} {https://10.0.151.50:2379 0  <nil>}]
W0315 08:09:09.675643     880 clientconn.go:1120] grpc: addrConn.createTransport failed to connect to {https://10.0.140.66:2379 0  <nil>}. Err :connection error: desc = "transport: Error while dialing dial tcp 10.0.140.66:2379: operation was canceled". Reconnecting...
W0315 08:09:09.675643     880 clientconn.go:1120] grpc: addrConn.createTransport failed to connect to {https://10.0.135.104:2379 0  <nil>}. Err :connection error: desc = "transport: Error while dialing dial tcp 10.0.135.104:2379: operation was canceled". Reconnecting...
W0315 08:09:09.675734     880 clientconn.go:1120] grpc: addrConn.createTransport failed to connect to {https://10.0.151.50:2379 0  <nil>}. Err :connection error: desc = "transport: Error while dialing dial tcp 10.0.151.50:2379: operation was canceled". Reconnecting...
```

need a better etcd health check

cmd/openshift-tests/e2e.go
```diff
@@ -50,7 +50,7 @@ var staticSuites = []*ginkgo.TestSuite{
 		The disruptive test suite.
 		`),
 		Matches: func(name string) bool {
-			return strings.Contains(name, "[Disruptive]") && strings.Contains(name, "[Feature:EtcdRecovery]")
+			return strings.Contains(name, "[Disruptive]") && strings.Contains(name, "master and worker failure and recover with machine health checks")
```
smarterclayton: Now that we're getting close, this should be the new set of tests we want to run:

```
return strings.Contains(name, "[Feature:EtcdRecovery]") || strings.Contains(name, "[Feature:NodeRecovery]")
```

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,324 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/upgrades"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/disruption"
+	"github.com/openshift/origin/test/extended/util/disruption/controlplane"
+	"github.com/openshift/origin/test/extended/util/disruption/frontends"
+)
+
+var _ = g.Describe("[Feature:DisasterRecovery][Disruptive]", func() {
```
smarterclayton: Put "[sig-cluster-lifecycle]" at the beginning of the describe

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,324 @@
...
+	f := framework.NewDefaultFramework("machine-recovery")
+	f.SkipNamespaceCreation = true
+	f.SkipPrivilegedPSPBinding = true
+
+	oc := exutil.NewCLIWithoutNamespace("machine-recovery")
+
+	g.It("[dr-quorum-restore] Cluster should survive master and worker failure and recover with machine health checks", func() {
```
smarterclayton: Change `[dr-quorum-restore]` to be `[Feature:NodeRecovery]`

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,324 @@
...
+				certsDir := "/tmp/etcd-client-certs/"
+				dumpEtcdCertsOnDisk(oc, certsDir)
+				defer os.RemoveAll(certsDir)
+				//unhealthyEtcds := getUnhealthyEtcds(certsDir, masters)
+				//o.Expect(len(unhealthyEtcds)).To(o.BeNumerically("==", 0))
+
+				createMachineHealthCheckForRole(oc, "master")
+				defer deleteMachineCheckForRole(oc, "master")
+				createMachineHealthCheckForRole(oc, "worker")
+				defer deleteMachineCheckForRole(oc, "master")
```
smarterclayton: copy and paste error, you didn't delete the worker

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,324 @@
...
+	}).Namespace("openshift-machine-api")
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "machine.openshift.io",
+		Version: "v1beta1",
+		Kind:    "machinehealthchecks",
+	})
+	err := unstructured.SetNestedField(u.Object, role, "spec", "selector", "machine.openshift.io/cluster-api-machine-role")
+	o.Expect(err).ToNot(o.HaveOccurred())
+	err = unstructured.SetNestedField(u.Object, role, "metadata", "name", "test-health-check"+role)
```
smarterclayton: `"e2e-health-check-" + role` instead, and same up in delete

smarterclayton:
You need to regenerate the test bindata (after you address the remaining comments)

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,324 @@
...
+				}).Namespace("openshift-machine-api")
+
+				framework.Logf("Verify SSH is available before restart")
+				masters, workers := clusterNodes(oc)
+				o.Expect(len(masters)).To(o.BeNumerically(">=", 3))
+				o.Expect(len(workers)).To(o.BeNumerically(">=", 2))
+
+				framework.Logf("Verify etcd endpoints are healthy")
+				certsDir := "/tmp/etcd-client-certs/"
+				dumpEtcdCertsOnDisk(oc, certsDir)
```
smarterclayton: comment this out for now

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,330 @@
...
+	}).Namespace("openshift-machine-api")
+	err = mhc.Delete("test-health-check"+role, &metav1.DeleteOptions{})
+	o.Expect(err).ToNot(o.HaveOccurred())
+}
+
+func createMachineHealthCheckForRole(oc *exutil.CLI, role string) {
+	config, err := framework.LoadConfig()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	dynamicClient := dynamic.NewForConfigOrDie(config)
+	mhc := dynamicClient.Resource(schema.GroupVersionResource{
```
alaypatel07: not sure if this is right or I should be doing.. `oc.AsAdmin().DynamicClient().Resource(schema.GroupVersionResource{`

test/extended/dr/machine_recover.go
alaypatel07: we don't dropped

test/extended/dr/machine_recover.go
smarterclayton: The kube client from framework is always assumed to be admin

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,330 @@
...
+		Kind:    "machinehealthchecks",
+	})
+	err = unstructured.SetNestedField(u.Object, role, "spec", "selector", "machine.openshift.io/cluster-api-machine-role")
+	o.Expect(err).ToNot(o.HaveOccurred())
+	err = unstructured.SetNestedField(u.Object, role, "metadata", "name", "e2e-health-check-"+role)
+	o.Expect(err).ToNot(o.HaveOccurred())
+	err = unstructured.SetNestedField(u.Object, []interface{}{
+		map[string]string{
+			"type":    "Ready",
+			"timeout": "5m",
```
smarterclayton: Hrm, I don’t think the test is doing what we expect.  It takes 2m to run but the timeout is longer.  Can you run the test directly against a cluster and verify the expected flows occur?

alaypatel07: @smarterclayton I think we never waited for the worker node to get created after 5 min timeout. And we create the master machine in the test itself which was the reason for smaller time window. I added the code[1] to wait for worker node, hopefully, that would give test times closer to, if not more than, 5mins

1. https://github.com/openshift/origin/pull/24651/files#diff-a7ee9c3c62d07e787a70359b84c68fe7R184

alaypatel07: > 

[sig-cluster-lifecycle][Feature:DisasterRecovery][Disruptive] [Feature:NodeRecovery] Cluster should survive master and worker failure and recover with machine health checks [Serial] [Suite:openshift] | 5m19s
-- | --

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24651/pull-ci-openshift-origin-master-e2e-aws-disruptive/55

@smarterclayton PTAL, I can still run locally if need be, thanks.

smarterclayton: I think that’s what I expect.  Before we merge I want to merge a fix that ensures logs are preserved when we call openshift tests twice in a row (caught that debugging).

alaypatel07: sounds good, thanks

alaypatel07:
/test e2e-aws-disruptive

going to pull the kubeconfig and inspect 🔍

test/extended/dr/machine_recover.go
```diff
@@ -0,0 +1,388 @@
...
+						vanishedMachines.Delete(machine.GetName())
+					}
+					if vanishedMachines.Len() != len(targetMachineNames) {
+						framework.Logf("Machines waiting to go be deleted: %v", vanishedMachines.List())
+						return false, nil
+					}
+					return true, nil
+				})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
```
alaypatel07, line 161: at this point, we need to rsh into one of the remaining etcd pods and remove the member. CEO does not currently have an automated way of scaling down. We removed the feature in 4.4 to reduce the scope during rework.

smarterclayton:
Replaced by #25069

PR #24661, opened 2020-03-09
Bug 1811726: Make parallelism configurable

jsafrane:
`openshift-tests` runs several tests in parallel (in a parallel suite). It is currently not possible to set the parallelism.

During CSI certification (included in openshift-tests binary) some users complain that the parallelism is too high for their small test clusters / storage backends / CI.


smarterclayton:
/hold

Need a "why" in the PR and commit description why you need this.

jsafrane:
@smarterclayton, copied "why" from BZ.

cmd/openshift-tests/openshift-tests.go
```diff
@@ -278,4 +278,5 @@ func bindOptions(opt *testginkgo.Options, flags *pflag.FlagSet) {
 	flags.IntVar(&opt.Count, "count", opt.Count, "Run each test a specified number of times. Defaults to 1 or the suite's preferred value.")
 	flags.DurationVar(&opt.Timeout, "timeout", opt.Timeout, "Set the maximum time a test can run before being aborted. This is read from the suite by default, but will be 10 minutes otherwise.")
 	flags.BoolVar(&opt.IncludeSuccessOutput, "include-success", opt.IncludeSuccessOutput, "Print output from successful tests.")
+	flags.IntVar(&opt.Parallelism, "parallelism", opt.Parallelism, "Number of tests running in parallel. 0 defaults to test suite recommended value, which is different in each suite.")
```
smarterclayton: `--max-parallel-tests` to be consistent with other places where we limit overall throughput.

jsafrane: Fixed, PTAL

PR #24673, opened 2020-03-11
Bug 1765220: Use sandbox id instead of getting it using pod key

abhat:
This PR addresses the use of an incorrect key as the sandbox ID. Previous fix #24585 introduced a bug where we were using the `getPodKey` API to index the sandbox. This change fixes that to use the `Id` field of the sandbox instead.


danwinship:
>  _No description provided._

:unamused:

abhat:
> > _No description provided._
> 
> 😒

Fixed. Isn't the commit message of the first commit supposed to be the default description?

danwinship:
If there's only one commit then the body of that commit message will be used as the default description. But your commit message has no body, so there was nothing to use.

abhat:
> If there's only one commit then the body of that commit message will be used as the default description. But your commit message has no body, so there was nothing to use.

Yep, makes sense. Thanks. Can I get an lgtm on this?

PR #24674, opened 2020-03-11
test/extended/router: Add http/2 conformance test 

frobware:
Add a new test that validates the router for http2 conformance; conformance is validated by running the h2spec[1] tool.

In OpenShift 4.5 we are using haproxy-2.0. This is currently known to not pass the h2spec tests in its entirety (~5 tests out of a possible 146 fail, sporadically). There is the upstream issue https://github.com/haproxy/haproxy/issues/471 that details why some of these tests fail. Note: in later versions of haproxy (2.2-dev) the failure rate is now between 0 and 1 (again, a sporadic timing failure).

To mitigate flaking in CI I have allowed the test to run (from a developer machine) for a number of iterations (150) to aggregate the set of known failures for OpenShift 4.5 / haproxy-2.0.13. This list of known flakes will not cause the test to fail, or CI to fail.

As we take newer versions of haproxy we should re-run the sampling/aggregation to ensure that the set of flakes is a) still valid and/or b) to discover any new flakes.

TODO: Work out where / how we integrate the image `docker.io/summerwind/h2spec:2.4.0`. In this test we currently pull it from `docker.io`.

[1] https://github.com/summerwind/h2spec


frobware:
cc @ironcladlou @Miciah @danehans

frobware:
By default we currently run 10 iterations of h2spec.

The runtime for this is ~1m 30s - don't think that qualifies as `[Slow]`, just want to highlight the cost of running h2spec more than once.

ironcladlou:
I'm kind of confused about the test topology here... looks like there's a bare pod running in-cluster using the h2spec image, and then a command that enters the pod namespace to run the test, but how does that exercise http/2 through ingress to an endpoint behind a route? I expected:


```
CI cluster (hosts test process) | Test cluster (target of tests)
client -----------------------> | haproxy <---> pod
```

And using three routes: passthrough, edge, reecrypt.

Am I misunderstanding?

Exercising connectivity through the edge may not be strictly necessary, but I'm not sure how we get around needing to test through the router and routes...

ironcladlou:
My understanding of the goal:

_Verify that end-users can deploy conformant HTTP/2 applications exposed by routes on OpenShift._

Which would be implemented by deploying some echo server in a pod, exposing it through a variety of routes, and then running the h2spec conformance tests against _the application_ through the routes — _not_ against haproxy itself (as testing haproxy itself for compliance using some canonical configuration is orthogonal and separately important).

@openshift/sig-network-edge @smarterclayton @knobunc am I confused?

ironcladlou:
If I got the goal right, maybe I misunderstand the role of h2spec in that context and some other testing methodology would be more appropriate (e.g. something like gRPC test?)

frobware:
> Which would be implemented by deploying some echo server in a pod, exposing it through a variety of routes, and then running the h2spec conformance tests against _the application_

So, maybe my wires are crossed here too.

Do we not want to validate haproxy itself?

It was my understanding that we would also validate http/2 conformance for _applications_ by using gRPC client/server (where server==pod behind a route).

frobware:
And... for h2pec it's not actually or necessarily about haproxy itself though. If we switched to something else then wouldn't testing the new thing for its h2spec conformance be something we would want to do?

ironcladlou:
Testing haproxy (aka "the ingress controller implementation") itself seems necessary, otherwise we can't extend http/2 support to client applications being proxied... so that seems important/necessary.

Separately, and building on the assumption that the proxy itself supports http/2, it seems useful to prove that end users can deploy applications that rely on http/2. Whether h2spec is appropriate for the _application_ test, I'm not clear yet

frobware:
I can make this PR also test h2spec invocations against 3 route types: edge, pass-through and re-encrypt. We can also test against haproxy itself it that still makes sense.

frobware:
Running h2spec terminating at the edge against a service that is h2c-based (http/2 cleartext) throws up the same issues [1] I had with the haproxy template when trying gRPC so arguably h2spec still has a lot of value here. I will follow-up with a separate PR for that.

[1] https://github.com/frobware/haproxy-hacks/blob/master/grpc-demo/0001-h2-without-tls-annotation.patch

frobware:
Just recapping based on some design discussion today:
- first iteration is to make passthrough work
- second is to terminate at edge
- third is re-encrypt

The goal is to not wait for all of these to be running/flake-free before we merge but to get some exposure to running the tests ASAP. 

If we could make the BE server be haproxy itself (preferably same version we use) we would implicitly test haproxy for h2spec conformance. Which is both neat and a litttle _meta_.

frobware:
@ironcladlou reworked this to test via routes as you suggested

- "edge", tests skipped because it needs https://github.com/openshift/router/pull/104
- "passthrough", tests skipped as the nginx config throws more errors than I'm comfortable with
- "reencrypt", enabled

Each route type has a list of known failures.

For "reencrypt" we get the following results:

```
146 tests, 145 passed, 0 skipped, 1 failed
```

and the test that fails is captured as a known failure which will prevent the overall e2e from failing.

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+	})
+
+	g.By(fmt.Sprintf("creating routes from a config file %q", configPath))
+
+	g.Describe("The HAProxy router", func() {
+		g.It("should pass the h2spec conformance tests", func() {
+			var err error
+
+			_, err = exutil.WaitForRouterServiceIP(oc)
```
ironcladlou: Not sure if this check matters since you're waiting for route status (implying router processed the route) and since the tests won't have started without a ready cluster anyway

frobware: Ack. Will drop this.

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,289 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			err = oc.Run("create").Args("-f", configPath).Execute()
+			o.Expect(err).NotTo(o.HaveOccurred())
```
ironcladlou: Are you tearing down the test namespace anywhere?

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,289 @@
...
+
+	e2e.Logf("[%s] Sampling completed: %v test cases failed", t.routeType, len(failuresByTestCaseID))
+	printFailures("", failuresByTestCaseID)
+}
+
+func getHostnameForRoute(oc *exutil.CLI, routeName string) string {
+	var hostname string
+	ns := oc.KubeFramework().Namespace.Name
+	err := wait.Poll(time.Second, changeTimeoutSeconds*time.Second, func() (bool, error) {
+		e2e.Logf("Getting hostname for route %q", routeName)
```
ironcladlou: This could be spammy... consider only logging on failure?

test/extended/router/h2spec.go
frobware: I believe so:

```console
STEP: [reencrypt] Fetching "/tmp/runConformanceTests912928811" from the container
Mar 18 12:30:07.021: INFO: Running 'oc --namespace=e2e-test-router-h2spec-8dl5l --kubeconfig=/tmp/configfile241935612 rsync h2spec:/tmp/runConformanceTests912928811 /tmp --strategy=tar'
runConformanceTests912928811
STEP: [reencrypt] Reading "/tmp/runConformanceTests912928811"
STEP: [reencrypt] Decoding "/tmp/runConformanceTests912928811"
STEP: Analyzing results
Mar 18 12:30:08.198: INFO: [reencrypt] TestCase ID: "http2/4.2.3" is a known failure; ignoring
[AfterEach] [sig-network-edge][Conformance][Area:Networking][Feature:Router]
  /home/aim/go-projects/origin/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/test/extended/util/client.go:101
Mar 18 12:30:08.297: INFO: Deleted {user.openshift.io/v1, Resource=users  e2e-test-router-h2spec-8dl5l-user}, err: <nil>
Mar 18 12:30:08.394: INFO: Deleted {oauth.openshift.io/v1, Resource=oauthclients  e2e-client-e2e-test-router-h2spec-8dl5l}, err: <nil>
Mar 18 12:30:08.489: INFO: Deleted {oauth.openshift.io/v1, Resource=oauthaccesstokens  yG_tgwKRS1SnSLOXccW6VAAAAAAAAAAA}, err: <nil>
[AfterEach] [sig-network-edge][Conformance][Area:Networking][Feature:Router]
  /home/aim/go-projects/origin/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/vendor/k8s.io/kubernetes/test/e2e/framework/framework.go:155
Mar 18 12:30:08.489: INFO: Waiting up to 7m0s for all (but 100) nodes to be ready
STEP: Destroying namespace "e2e-test-router-h2spec-8dl5l" for this suite.
[AfterEach] [sig-network-edge][Conformance][Area:Networking][Feature:Router]
  /home/aim/go-projects/origin/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/test/extended/router/h2spec.go:52
Mar 18 12:30:08.856: INFO: Running AfterSuite actions on all nodes
Mar 18 12:30:08.856: INFO: Running AfterSuite actions on node 1

aim@spicy:~/go-projects/origin/src/github.com/openshift/origin
$ oc get namespaces | grep h2spec
e2e-test-router-h2spec-8dl5l                            Terminating   40s
```

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,289 @@
...
+				// iterations to identify flaking
+				// tests. This should only be set to
+				// true in development (i.e., not CI).
+				//
+				// This should be run every time we
+				// consume any new version of haproxy
+				// be that a major, minor or a micro
+				// update to continuously validate the
+				// set of test case IDs that fail.
+				if sampleForFailures := false; sampleForFailures {
```
ironcladlou: Maybe add an env variable? Not sure if there's a better way to inject args into the test suite, but I think if this is useful for development I'd like a way to set it without compiling

frobware: Yep, good idea. However, .... I think it's also likely if you're in this mode you'll be updating the list of known failures. But will add env variable.

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	return failures
+}
+
+func runConformanceTests(oc *exutil.CLI, t h2specRouteTypeTest) []*h2spec.JUnitTestSuite {
+	podName := "h2spec"
+	e2e.ExpectNoError(oc.KubeFramework().WaitForPodRunning(podName))
+
+	outputFile, err := ioutil.TempFile("/tmp", "runConformanceTests")
```
ironcladlou: Not sure if there are any gotchas writing a temp file in this context, might want to ask around. I would at least suggest `ioutil.TempFile("", "runConformanceTests")` for portability (which will use the system's default tmp dir)

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,289 @@
...
+		// testsuite in the results.
+		e2e.Logf("[%s] h2spec: error=\n%s", t.routeType, h2specErr)
+	}
+	if h2specErr != nil && h2specErr.Error() != h2specOutput {
+		// Also log any output if different.
+		e2e.Logf("[%s] h2spec: output=\n%s", t.routeType, h2specOutput)
+	}
+
+	g.By(fmt.Sprintf("[%s] Fetching %q from the container", t.routeType, outputFile.Name()))
+	rsyncErr := oc.Run("rsync").Args(podName+":"+outputFile.Name(), filepath.Dir(outputFile.Name()), "--strategy=tar").Execute()
```
ironcladlou: If you were having some stability issue with rsync, what about using `oc cp`?

test/extended/router/h2spec.go
frobware: w.r.t environment name:

- ROUTER_H2SPEC_SAMPLE_MODE
- other preferences?

test/extended/router/h2spec.go
frobware: Thanks - will try that in sample mode.

test/extended/router/h2spec.go
frobware: Addressed in https://github.com/openshift/origin/pull/24674/commits/2ecf50fbc41867d1be158f77fb4167d15c083b2f

frobware:
@ironcladlou addressed your feedback in https://github.com/openshift/origin/pull/24674/commits/2ecf50fbc41867d1be158f77fb4167d15c083b2f. Thanks!

test/extended/router/h2spec.go
frobware: I think this is using rsync/tar under the hood because I see: 

```
STEP: [reencrypt] Copying h2spec:"/tmp/runConformanceTests513746697" "/tmp/runConformanceTests513746697"
Mar 18 13:50:10.115: INFO: Running 'oc --namespace=e2e-test-router-h2spec-6ds6p --kubeconfig=/tmp/configfile831321424 cp h2spec:/tmp/runConformanceTests513746697 /tmp/runConformanceTests513746697'
tar: Removing leading `/' from member names
```

I switched to `oc cp` anyway.

test/extended/router/h2spec.go
ironcladlou: If you're not doing it in this test code, what's deleting the ns?!

Usually we must clean up after ourselves in these tests.

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,286 @@
...
+
+const h2specTimeout = 5 * time.Second
+
+type h2specFailingTest struct {
+	TestCase   *h2spec.JUnitTestCase
+	TestNumber int
+}
+
+type h2specRouteTypeTest struct {
+	routeType     string
```
ironcladlou: instead of `string`, how about [TLSTerminationType ](https://github.com/openshift/api/blob/master/route/v1/types.go#L236)?

test/extended/router/h2spec.go
```diff
@@ -0,0 +1,286 @@
...
+				// be that a major, minor or a micro
+				// update to continuously validate the
+				// set of test case IDs that fail.
+				if _, sampleForFailures := os.LookupEnv("ROUTER_H2SPEC_SAMPLE_MODE"); sampleForFailures {
+					runConformanceTestsAndLogAggregateFailures(oc, t, 150)
+					continue
+				}
+
+				testSuites := runConformanceTests(oc, t)
+				o.Expect(len(testSuites)).ShouldNot(o.BeZero(), "expected > 0 testSuites")
```
ironcladlou: Ginkgo-golf:

```
o.Expect(testSuites).ShouldNot(o.BeEmpty())
```

test/extended/testdata/router/router-h2spec.yaml
```diff
@@ -0,0 +1,165 @@
+apiVersion: v1
```
ironcladlou: I guess we should talk about nginx... if it's possible to re-do with haproxy, benefits include:

1. We can test the same haproxy image we're deploying for ingress
2. No concerns about using an image we're not shipping
3. ???

Downsides include:

1. You already did it in nginx

frobware: I think it makes sense to go for haproxy - I just wanted the basic infrastructure up as a PR so we could discuss it.

frobware: I actually considered an axis of `haproxy, nginx` which is why I named these as such. Perhaps that's overkill. Thoughts?

The test would effectively then be:

    for backed in "nginx", "haproxy", do...

test/extended/router/h2spec.go
frobware: Addressed in https://github.com/openshift/origin/pull/24674/commits/798acd38e4abc3439c5e905a8a090afd838ec3b5

test/extended/router/h2spec.go
frobware: Addressed in https://github.com/openshift/origin/pull/24674/commits/798acd38e4abc3439c5e905a8a090afd838ec3b5

test/extended/testdata/router/router-h2spec.yaml
frobware: Switched to just haproxy. Squashed commits.

frobware:
For reference some upstream effort to also run h2spec in haproxy CI: https://github.com/chipitsine/haproxy/commit/8c90ea82fd32c0ca9bd3df0ae7d9361525eda590

frobware:
@ironcladlou this is good for another review; switched from nginx to haproxy for the backend testing. It also uses the same image as we do for the frontend.

frobware:
> It also uses the same image as we do for the frontend.

```console
--> Deploying template "e2e-test-router-h2spec-fqdqj/" for "/tmp/fixture-testdata-dir616346082/test/extended/testdata/router/router-h2spec.yaml" to project e2e-test-router-h2spec-fqdqj

     * With parameters:
        * HAPROXY_IMAGE=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:559e38c1171467ee375f2ea873495624920accf3ae0ff4b99cae98964e708897
```

test/extended/testdata/router/router-h2spec.yaml
```diff
@@ -0,0 +1,287 @@
...
+- apiVersion: v1
+  kind: Pod
+  metadata:
+    name: h2spec
+    labels:
+      app: h2spec
+  spec:
+    containers:
+    - name: h2spec
+      image: docker.io/summerwind/h2spec:2.4.0
```
ironcladlou: Is coming up with a non-dockerhub image here the last remaining issue?

frobware: Good point. Does it mean that other images we pull in tests (e.g., nginx) come from another registry?

frobware: The consensus on Slack was that "in general it’s fine to use an image to get the first 80% of verification".

ironcladlou:
/approve

@openshift/sig-network-edge @smarterclayton before I pull the trigger on this one, something I forgot to bring up earlier: is this appropriate to remain in the origin repo, or should it be in the ingress-operator e2e?

ironcladlou:
/lgtm

If we need to move it, we will.

PR #24678, opened 2020-03-12
relax must-gather e2e test - audit log removed

sanchezl:
In preparation for https://github.com/openshift/must-gather/pull/145


deads2k:
prefer the approach in https://github.com/openshift/origin/pull/24680

sanchezl:
Let's go with https://github.com/openshift/origin/pull/24680 instead.

PR #24681, opened 2020-03-12
Bug 1811695: Add missing optional tags

soltysh:
/assign @sttts


soltysh:
Not needed, openshift/api was bumped in https://github.com/openshift/origin/pull/24720.

PR #24682, opened 2020-03-12
test: Update the ginkgo annotator to avoid double naming

smarterclayton:
We were including `[Top Level]` and the parent nodes twice in output, which
made all test names substantially longer. Pick up the improved ginkgo
interface and keep the parent node name separate.

Before:

> [Feature:UserAPI] [Top Level] [Feature:UserAPI] users can manipulate groups [Suite:openshift/conformance/parallel]

After:

> [Feature:UserAPI] users can manipulate groups [Suite:openshift/conformance/parallel]


smarterclayton:
Rebased and reapplied label.  Failures are not specific to this Pr

PR #24683, opened 2020-03-13
test: Ensure all tests have an appropriate sig prefix

smarterclayton:
SIG names were chosen to match the upstream (apps is workloads).
Where no upstream exists new ones were defined (etcd|devex).


cmd/openshift-tests/e2e.go
```diff
@@ -93,7 +93,7 @@ var staticSuites = []*ginkgo.TestSuite{
 		Tests that exercise language and tooling images shipped as part of OpenShift.
 		`),
 		Matches: func(name string) bool {
-			return strings.Contains(name, "[image_ecosystem]") && !strings.Contains(name, "[Local]")
+			return strings.Contains(name, "[Feature:ImageEcosystem]") && !strings.Contains(name, "[Local]")
```
bparees: sig-samples ?
sig-imagesamples ?
sig-imageecosystem ?

bparees:
do we have a canonical list of sigs -> team mappings somewhere?

also i would have voted for things like "sig-builds" instead of "sig-devex".  makes it easier to move components between teams if we were so inclined.

smarterclayton:
I will accept a follow on PR

PR #24690, opened 2020-03-13
Bug 1813428: Restore globally-writeable /etc/passwd in tests image

smarterclayton:
This reverts commit 33a6115867f59fc479dba6ca009de3492632c114.

This is required for our CI infrastructure to use SSH correctly
until we have completely moved CI to 4.x infrastructure (where
/etc/passwd becomes dynamic). This broke the disruptive test
suite and etcd recovery, and blocks a number of other tests being
added.

The issue is not a CVE.


smarterclayton:
4.4 bug is 1813429

smarterclayton:
disruptive test executed again

wking:
Can't folks who need to SSH run a custom test image that allows this, without us needing a writable `/etc/passwd` in the test image that gets shipped in the release?

PR #24698, opened 2020-03-16
test: All tests must have a sig, prevent double sig auto assignement

smarterclayton:
The [sig-xxx] label is now required on all tests, including upstream. Add
logic to fail update-generated-bindata if a test doesn't have the label,
and add a special case exception during label assignment to prevent assign
a sig-label twice in rules.

/assign @bparees


smarterclayton:
/retest

unrelated timeout importing from docker hub taking longer than 1m (should be moved to quay)

test/extended/util/annotate/annotate.go
```diff
@@ -35,6 +37,28 @@ func main() {
...
+	// If you get this error, you should add the [sig-X] tag to your test (if its
+	// in origin) or if it is upstream add a new rule to rules.go that assigns
+	// the test in question to the right sig.
+	//
+	// Upstream sigs map to teams (if you have representation on that sig, you
+	//   own those tests in origin)
+	// Downstream sigs: sig-imageregistry, sig-builds, sig-devex
+	var errors []string
+	for from, to := range generator.output {
+		if !reHasSig.MatchString(from) && !reHasSig.MatchString(to) {
```
bparees: why do we care about the from?

smarterclayton: `from` contains strings that `to` doesn't.  The previous bug fixed the issue where we were doubling up the first part of the from, now we can't just search to (that's everything on the test and that we added), we have to search from (includes things we won't change)

bparees: ok

bparees: i was assuming to would be the full new string, not just the additions

PR #24700, opened 2020-03-16
Bug 1813421: e2e well-known: use proxy if in a proxied environment

stlaz:


stlaz:
tested manually in a proxied cluster

PR #24708, opened 2020-03-17
Bug 1814357: Fix nil reference when accessing un-migrated platform status

ironcladlou:
Fix nil reference crashes when accessing an infrastructure status field that
hasn't been migrated.

### DIFF
--- a/test/extended/router/headers.go
+++ b/test/extended/router/headers.go
@@ -44,9 +44,14 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 		g.It("should set Forwarded headers appropriately", func() {
 			o.Expect(infra).NotTo(o.BeNil())
 
-			if !(infra.Status.PlatformStatus.Type == configv1.AWSPlatformType ||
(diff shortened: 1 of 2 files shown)


test/extended/router/headers.go
```diff
@@ -44,9 +44,14 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
...
-				infra.Status.PlatformStatus.Type == configv1.AzurePlatformType ||
-				infra.Status.PlatformStatus.Type == configv1.GCPPlatformType) {
+			platformType := infra.Status.Platform
+			if infra.Status.PlatformStatus != nil {
+				platformType = infra.Status.PlatformStatus.Type
+			}
+			switch platformType {
+			case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.GCPPlatformType:
+				// supported
+			default:
```
frobware: Isn't the logging line below still a nil deref?

ironcladlou: Yep, good catch. Opened https://github.com/openshift/origin/pull/24718

PR #24709, opened 2020-03-17
Bug 1806594: Multi-AZ test should only check nodes the pods are scheduled to

smarterclayton:
The test assumes that all nodes are schedulable when calculating nodes,
but masters are not and nodes in many e2e runs are only in two zones.
The e2e suite needs to be fixed to take selectors for nodes that workloads
can schedule to by default and then the test should use only those nodes
to get zones, but that is a much more invasive change.

The minimum workaround is to only verify spreading across the nodes we
are actually scheduled to and error if there is only one node scheduled
to (a multi-az test should fail if we are in a single AZ).

This flakes 1/4 times on AWS because we run 2 zones.

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/scheduling/ubernetes_lite.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/scheduling/ubernetes_lite.go
@@ -19,6 +19,7 @@ package scheduling
 import (
 	"fmt"
 	"math"
+	"time"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


smarterclayton:
4.4 bug is 1814360 and 4.3 bug is 1814363

smarterclayton:
/hold

I may have misdiagnosed the particular failure here in GCP.  The general bug is still a bug (masters and workers don't have to match) but this may not fix the GCP flake correctly.

soltysh:
Can you make sure we don't call that upstream commit with number pointing to an issue, it will be confusing during the next rebase. 
/cc @marun 
since you're doing the current rebase

smarterclayton:
There is no upstream fix yet, so I'd rather reference the issue where you can find the bug than nothing.

marun:
@smarterclayton I've picked the main commit for the rebase PR to try to get the multi-AZ test passing. Will revisit before merge.

soltysh:
@smarterclayton or @damemi can we get some resolution on this one?

PR #24711, opened 2020-03-17
test: Add a test that verifies no terminating namespaces or pods

smarterclayton:
At the end of an e2e run any terminating pods for more than 2m is
a product bug (either in cleanup of test, or cleanup of infra, or
in the kubelet or container runtimes). Set a relatively low
tolerance and error out.


smarterclayton:
/hold

for results

rphillips:
e2e-gcp appears to have unrelated errors... 

/retest

smarterclayton:
/retest

I believe volume provisioning and CSI tests are the main break here

PR #24712, opened 2020-03-17
[release-4.3] Bug 1811200: /readyz should start returning failure on shutdown initiation

tkashem:
Currently, /readyz starts reporting failure after ShutdownDelayDuration
elapses. We expect /readyz to start returning failure as soon as
shutdown is initiated. This gives the load balancer a window defined by
ShutdownDelayDuration to detect that /readyz is red and stop sending
traffic to this server.


tkashem:
Closed https://github.com/openshift/origin/pull/24657 and opened this PR manually. Needed to include the following upstream commit into the cherry-pick PR but was not possible to do so.
`UPSTREAM 89164: Fix data race issue in unit test`

PR #24721, opened 2020-03-19
Bug 1815050: Fix nil reference when accessing un-migrated platform status

ironcladlou:
Fix nil reference crashes when accessing an infrastructure status field that
hasn't been migrated.

Combined backport of https://github.com/openshift/origin/pull/24708 and https://github.com/openshift/origin/pull/24718.


ironcladlou:
@knobunc PTAL

PR #24723, opened 2020-03-19
Bug 1790825: fix test flake in operators test

ecordell:
When the namespace is deleted, the operatorgroup can be removed
before the csv, which can cause the csv to flip to failed.

there is an alert that fires when the csv flips to failed, which
can cause the monitoring test ensuring no alerts fired to fail


ecordell:
/retest

status seems wonky - I think g2p-upgrade is the only real failure.

PR #24724, opened 2020-03-19
Bug 1814460: test: only run the networking "oc explain" tests under openshift-sdn

danwinship:
The test that "oc explain" works for the openshift-sdn types was being skipped under ovn-kubernetes, but this got broken when the test was renamed for the new "sig" stuff.

But it shouldn't have been "skipped under ovn-kubernetes" anyway, it should be "skipped under everything except openshift-sdn". There's no way to say that in the skip lists, but we use this helper function in some other places as well.

/cc @abhat @dcbw


danwinship:
/test e2e-aws-ovn

/hold
until the results are confirmed

test/extended/cli/explain.go
```diff
@@ -372,11 +373,13 @@ var _ = g.Describe("[sig-cli] oc explain networking types", func() {
...
-			o.Expect(verifyExplain(oc, nil, schema.GroupVersionResource{},
-				st.pattern, st.field, fmt.Sprintf("--api-version=%s", st.gv))).NotTo(o.HaveOccurred())
-		}
+		networking.InOpenShiftSDNContext(func() {
+			for _, st := range specialNetworkingTypes {
+				e2e.Logf("Checking %s, Field=%s...", st.gv, st.field)
+				o.Expect(verifyExplain(oc, nil, schema.GroupVersionResource{},
+					st.pattern, st.field, fmt.Sprintf("--api-version=%s", st.gv))).NotTo(o.HaveOccurred())
+			}
+		})
```
smarterclayton: Please add the correct explain rules for OVN types.  All product CRDs are required to have schemas.

abhat: Each type has it's own idiosyncrasies. They don't have a Spec and Status as a rule. Whatever we write will be very specific.

danwinship: ovn-kubernetes does not create any CRDs. (It stores its private information in the OVN databases rather than in kubernetes.) If we do create CRDs for it in the future we'll create them "correctly" (spec and status, etc) so they don't need to be tested as exceptions.

dcbw: @smarterclayton ^^^

dcbw:
@danwinship 
```
./test/extended/cli/explain.go
```

dcbw:
Also, did it really stop running the tests? e2e-aws-ovn has:

```
Failing tests:

[sig-cli] oc explain networking types should contain proper fields description for special networking types [Suite:openshift/conformance/parallel]
```

abhat:
It will not stop running the test @dcbw. But it will not execute a code body in the non-sdn contexts afaict.

dcbw:
@danwinship `panic: unable to find test [Top Level] [sig-cli] oc explain networking types when using openshift-sdn should contain proper fields description for special networking types`

test/extended/cli/explain.go
smarterclayton: That's all I meant.  While we're in here fixing it, run the test for each type and say "doesn't need explain" or "here are the exceptions and the bugs to fix".  If a particular network doesn't have CRDs, say so in code.

danwinship: `network.operator.openshift.io` was the only networking-related type listed-but-commented-out in explain.go, and I've tested it and uncommented it.

So, you mean you want every CRD type that the CNO ever creates, directly or indirectly, to be added to this list, not just CRDs for types in `openshift/api`? I don't have time to do that now but I can create a jira card about it...

danwinship:
blah, e2e-gcp is on fire, but anyway, it passed the networking oc explain test, and e2e-aws-ovn skipped it.
/hold cancel

test/extended/cli/explain.go
dcbw: Yeah, that really sounds like a completely separate PR as it won't solve the current problem of unbreaking the CI tests that the switch to SIG-prefixed tests broke.

dcbw:
Consistent CI failure of `[sig-builds][Feature:Builds][Slow] Capabilities should be dropped for s2i builders  s2i build with a rootable builder should not be able to switch to root with an assemble script [Suite:openshift]`

```
failure: repodata/repomd.xml from cbs.centos.org_repos_sclo7-rh-ruby25-rh-candidate_x86_64_os_: [Errno 256] No more mirrors to try.
https://cbs.centos.org/repos/sclo7-rh-ruby25-rh-candidate/x86_64/os/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
subprocess exited with status 1
subprocess exited with status 1
error: build error: error building at STEP "RUN yum -y install expect": exit status 1
```

test/extended/cli/explain.go
smarterclayton: Yes, you can fix that separately.

CRDS must have schemas.  This is a product requirement.  We can't give customers an empty JSON object and tell them good luck.  It definitely needs a card, and it needs priority, and if you need help, just tell me and I'll get some folks reassigned from other teams to help.

smarterclayton:
/lgtm

We'll follow up on the missing CRD schema for network types separately.

smarterclayton:
Green boxing to unblock the team (all the impacted suites have passed and manually verified that openshift-sdn version was still running)

test/extended/cli/explain.go
danwinship: filed https://issues.redhat.com/browse/SDN-873

PR #24725, opened 2020-03-19
bug 1813191: Fix ipfailover nopreempt & notification problems

pecameron:
https://bugzilla.redhat.com/show_bug.cgi?id=1813191

Signed-off-by: Phil Cameron <pcameron@redhat.com>

### DIFF
--- a/images/ipfailover/keepalived/lib/config-generators.sh
+++ b/images/ipfailover/keepalived/lib/config-generators.sh
@@ -190,6 +190,11 @@ function generate_vrrpd_instance_config() {
 
   [ "${instancetype}" = "master" ] && initialstate="state MASTER"
 
+  if [[ ${preempt} == "nopreempt" ]] ; then
+	initialstate="state BACKUP"
+	priority=1
+  fi
+
   local instance_name ; instance_name=$(generate_vrrp_instance_name "${servicename}" "${iid}")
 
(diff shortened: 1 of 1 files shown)


pecameron:
@knobunc PTAL
This can only be fixed in 3.11, file doesn't exist in master.

images/ipfailover/keepalived/lib/config-generators.sh
```diff
@@ -190,13 +190,19 @@ function generate_vrrpd_instance_config() {
...
+  fi
+
   local instance_name ; instance_name=$(generate_vrrp_instance_name "${servicename}" "${iid}")
 
   local auth_section ; auth_section=$(generate_authentication_info "${servicename}")
   local vip_section ; vip_section=$(generate_vip_section "${vips}" "${interface}")
   # Emit instance
   echo "
 vrrp_instance ${instance_name} {
+   smtp_alert
```
sseanlee: how about providing a parameter to enable/disable notifications, since not everyone wants to receive notifications?

pecameron: Changed it to set when SMTP_SERVER is provided.

pecameron:
@sseanlee @knobunc @ironcladlou @smarterclayton PTAL

PR #24736, opened 2020-03-20
WIP: drain: if a namespace is draining then do not error

rphillips:


sjenning:
fate of this PR is tied to https://github.com/kubernetes/kubernetes/pull/89314

PR #24746, opened 2020-03-22
Bug 1815944: Pick up Ginkgo test performance fix to reduce e2e CPU usage

smarterclayton:
Picks up a fix to Ginkgo that substantially reduces amount of CPU used
in our CI infrastructure to run e2e tests.


smarterclayton:
/retest

Not sure if it is reliable

PR #24750, opened 2020-03-23
Bug 1816302: UPSTREAM: 88505: Honor status.podIP over status.podIPs, node.spec.podCIDR over node.spec.podCIDRs

danwinship:
`pod.status.podIPs` was introduced in kube 1.16 with the requirement that if `podIP` and `podIPs` are both set, they must match. This means that when a kube 1.14 / OCP 4.2 kubelet tries to change the `podIP` of a pod on a kube 1.16 / OCP 4.3 apiserver, it will fail because it didn't also try to change `podIPs`:

```
Mar 23 18:04:42 ip-10-0-163-3 hyperkube[1938]: W0323 18:04:42.769863    1938
status_manager.go:519] Failed to update status for pod "kube-apiserver-operator-
5d7c58bbb4-27mww_openshift-kube-apiserver-operator(93729130-6d20-11ea-
a89d-0a193a986132)": failed to patch status "..." for pod
"openshift-kube-apiserver-operator"/"kube-apiserver-operator-5d7c58bbb4-27mww":
conversion Error: v1.PodIP(10.129.0.65) != v1.PodIPs[0](10.129.0.41)
```

eg, specifically, if during a 4.2 to 4.3 upgrade, the apiserver pods are updated to 4.3, and then another pod gets restarted on a node that is still running a 4.2 kubelet, and upon being restarted it has a different IP then it did before, then it will get stuck.

This seems like something we should have hit sooner and I'm not sure why we didn't. (Possibly we used to end up always reassigning the pod its old IP again in this case, but some recent sdn/multus fix changed that?)

Upstream discussion: https://github.com/kubernetes/kubernetes/pull/88505

This will need to be backported to 4.4 and 4.3.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/apis/core/v1/conversion.go
+++ b/vendor/k8s.io/kubernetes/pkg/apis/core/v1/conversion.go
@@ -253,9 +253,13 @@ func Convert_v1_PodStatus_To_core_PodStatus(in *v1.PodStatus, out *core.PodStatu
 		return err
 	}
 
-	// If both fields (v1.PodIPs and v1.PodIP) are provided, then test v1.PodIP == v1.PodIPs[0]
+	// If both fields (v1.PodIPs and v1.PodIP) are provided and differ, then PodIP is authoritative for compatibility with older kubelets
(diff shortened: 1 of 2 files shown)


deads2k:
/lgtm

but I prefer bumping all of kube together for this.  We have two CVEs to grab too.

PR #24679, opened 2020-03-12
Bug 1826807: Add tests for topology manager in single and multisocket configurations

fromanirh:
This PR adds the initial basic topology manager tests.

We cover all the testcases kubernetes has on its e2e tests, and all the known issues origin had reported against the topology manager.
The reason for the overlap is because the e2e kubernetes node tests do not run against a full cluster, but on a reduced environment (kubelet+apiserver only, for example).
Furthermore, we care about some test scenarios that aren't very relevant for kubernetes. The path forward is of course to move the tests upstream were feasible, but this testsuite is meant to stay.

The tests by default skip if the cluster preconditions aren't met, so if topology manager and all its dependencies are not enabled, or configured with a policy which is not `single-numa-node`. This is because this policy is the only one which allow to unambiguosly check the correctness of the resource alignment.

The tests are configurable with few environement variables, documented in the README. This is unfortunate, but needed by the complexity of the scenarios the test need to check, and which we cannot reliably autodetect.

The tests require an environment with a SRIOV device enabled to properly exercise the alignment. Should the environment not have such a device, the test will skip.
This is because SRIOV and GPU devices are the only devices supported by the devicemanager at this time, and SRIOV device are the easiest and the cheapest to obtain.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #24755, opened 2020-03-24
add fixture and initial extended tests for topology manager

fromanirh:
This PR adds the initial basic topology manager tests, requesting only cpu cores and checking their alignment.

These tests assume nothing about the cluster state. We add a test fixture to setup a cluster with topology manager and restore it to its pristine state.
Because of its complexity, the fixture is configurable using env vars to skip the setup or the teardown stage, or both. This is useful when we want to continuously run the tests against a cluster.

Signed-off-by: Francesco Romani fromani@redhat.com


fromanirh:
This PR extends https://github.com/openshift/origin/pull/24679 adding the test fixture.
Setting up the cluster turned out to be a complex business, and I'm not 100% sure this is the best approach. Hence, I kept the PRs separated.

Reviewers please suggest which approach is better: with or without the fixture.

fromanirh:
the failure in  `ci/prow/e2e-aws-serial` is real and it is caused by timeout while applying the configuration required by Topology Manager.The timeout is 10m and it is already pretty long. The problem is that we need
1. to update a feature gate
2. wait for it to be updated
3. to label the workers MCP
4. to wait for it to be updated
5. to add the kubelet config settings to the workers MCP
6. to wait for the MCP workers to be updated *and restarted* <- this is what didn't complete in 10m

it seems this process is too slow to be completed timely in the aws-serial lane.
Besides increasing the timeout, do we have any more options?

fromanirh:
I'm keeping this PR open for reference purposes, but I believe we should move forward with https://github.com/openshift/origin/pull/24679 first - this PR builds on top of this anyway.

fromanirh:
It is still not clear if the testsuite should change the cluster configuration.
Closing to reduce the clutter, will restore later should we agree this is needed.

PR #24760, opened 2020-03-25
Make new-app related tests more resilient to changes to deployment by default

soltysh:
https://github.com/openshift/oc/pull/355 is switching `oc new-app` to produce deployments instead of deployment configs by default. There will be a new flag `--as-deployment-config` to provide backwards compatiblity. 
This change requires our tests to be more resilient expecting the outcome. This PR either changes tests to explicitly create DC where possible or expects either deployment or DC. 

/assign @mfojtik

### DIFF
--- a/test/extended/builds/new_app.go
+++ b/test/extended/builds/new_app.go
@@ -4,6 +4,10 @@ import (
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 
+	kapierrs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kdeployutil "k8s.io/kubernetes/test/e2e/framework/deployment"
+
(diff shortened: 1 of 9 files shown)


test/extended/builds/new_app.go
```diff
@@ -62,8 +66,16 @@ var _ = g.Describe("[sig-builds][Feature:Builds] oc new-app", func() {
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			g.By("waiting for the deployment to complete")
-			err = exutil.WaitForDeploymentConfig(oc.KubeClient(), oc.AppsClient().AppsV1(), oc.Namespace(), a58, 1, true, oc)
-			o.Expect(err).NotTo(o.HaveOccurred())
+			deploy, err := oc.KubeClient().AppsV1().Deployments(oc.Namespace()).Get(a58, metav1.GetOptions{})
+			if kapierrs.IsNotFound(err) {
```
mfojtik: how about other errors?

PR #24762, opened 2020-03-25
Bug 1817039: [4.4] UPSTREAM: 88506: Honor status.podIP over status.podIPs, node.spec.podCIDR over node.spec.podCIDRs

danwinship:
Backport of #24750, but via cherry-picking the upstream backport rather than the origin master PR

/assign @deads2k


jwforres:
Overriding BZ requirement since this bug manifests specifically in the 4.2 -> 4.3 upgrades

PR #24763, opened 2020-03-25
Bug 1817040: [4.3] UPSTREAM: 88507: Honor status.podIP over status.podIPs, node.spec.podCIDR over node.spec.podCIDRs

danwinship:
Backport of #24750, but via cherry-picking the upstream backport rather than the origin master PR

/assign @deads2k


sdodson:
Because this can only effectively be tested in 4.3 I've overriden the `bugzilla/valid-bug` labels
#24762 has merged so we're not at risk of regressing at a code level

PR #24764, opened 2020-03-25
Bug 1802311: e2e: Disable router header test on OVN

Miciah:
Ignore the address in the `X-Forwarded-For:` header when the cluster is using the `OVNKubernetes` network plugin.

* `test/extended/router/headers.go`: Get the `networks.config/cluster` object. If it indicates that the `OVNKubernetes` plugin is used, ignore the client IP address in the echo server's response.

----

This is temporary measure to get the test passing until we figure out a fix.

@ironcladlou, @abhat

### DIFF
--- a/test/extended/router/headers.go
+++ b/test/extended/router/headers.go
@@ -28,11 +28,14 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 		routerIP  string
 		metricsIP string
 		infra     *configv1.Infrastructure
+		network   *configv1.Network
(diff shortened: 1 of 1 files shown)


smarterclayton:
Link the bug tracking the fix to here

danehans:
4.4 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1820274
4.3 BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1820277

PR #24770, opened 2020-03-26
Show test binary version

jsafrane:
Each test output should start with version of the test binary. This is useful when using openshift-tests as CSI driver certification suite - the test results should show what version of tests was run.

On related note, why standard klog arguments are not exposed?

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -19,6 +20,7 @@ import (
 	"github.com/openshift/library-go/pkg/serviceability"
 	"github.com/openshift/origin/pkg/monitor"
 	testginkgo "github.com/openshift/origin/pkg/test/ginkgo"
+	"github.com/openshift/origin/pkg/version"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
@@ -111,6 +113,9 @@ func newRunCommand() *cobra.Command {
 		SilenceErrors: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return mirrorToFile(opt, func() error {
+				if !opt.DryRun {
+					fmt.Fprintf(os.Stderr, "%s version: %s\n", filepath.Base(os.Args[0]), version.Get().String())
+				}
 				config, err := decodeProvider(opt.Provider, opt.DryRun, true)
 				if err != nil {
 					return err


test/extended/util/test.go
```diff
@@ -91,7 +91,7 @@ func InitTest(dryRun bool) error {
 	// Ensure that Kube tests run privileged (like they do upstream)
 	TestContext.CreateTestingNS = createTestingNS
 
-	klog.V(2).Infof("Extended test version %s", version.Get().String())
+	klog.Infof("Extended test version %s", version.Get().String())
```
smarterclayton: This isn't the right place to print this.  It should only be printed in run suite, and we should print it with `fmt.Printf()`.

jsafrane: Reworked to `openshift-tests run` only and sent to stderr.

cmd/openshift-tests/openshift-tests.go
```diff
@@ -111,6 +112,7 @@ func newRunCommand() *cobra.Command {
 		SilenceErrors: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return mirrorToFile(opt, func() error {
+				fmt.Fprintln(os.Stderr, "openshift-test version:", version.Get().String())
```
smarterclayton: One more, sorry.  Exclude `dry-run`.

jsafrane: with `--dry-run`, I get this on stderr anyway:
```
I0326 18:32:47.828294  369176 test_context.go:419] Tolerating taints "node-role.kubernetes.io/master" when considering if nodes are ready
```

jsafrane: Fixed, still I think it's quite useless.

cmd/openshift-tests/openshift-tests.go
```diff
@@ -111,6 +112,9 @@ func newRunCommand() *cobra.Command {
 		SilenceErrors: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return mirrorToFile(opt, func() error {
+				if !opt.DryRun {
+					fmt.Fprintln(os.Stderr, "openshift-test version:", version.Get().String())
```
smarterclayton: space between ':' and version.  Then lgtm

smarterclayton: `openshift-test version: x.y.z`

smarterclayton: Actually... should be `fmt.Fprintf(os.Stderr, "%s version: %s\n", filepath.Base(os.Args[0]), version.Get().String())`

jsafrane: Space is added by `Fprintln`:
```
openshift-test version: v4.1.0+d56e7d7-2472-dirty
```

jsafrane: > Spaces are always added between operands and a newline is appended.

jsafrane: Fixed to your proposal. Now it must be perfect

PR #24771, opened 2020-03-26
Add test version into JUnit report

jsafrane:
So we know what version of openshift-tests was running.

This is useful for us when receiving CSI driver test reports to check CSI driver vendors used the right version.

/assign @smarterclayton


smarterclayton:
/lgtm

Thanks

PR #24774, opened 2020-03-26
Revert "tests: disable ResourceQuota e2e test for life of a secret"

mfojtik:
We believe that openshift/openshift-controller-manager#72 and it's parents is the likely cause of the issue, that is reverted in openshift/openshift-controller-manager#80


smarterclayton:
As soon as this goes green I'm going to merge it. I wish I had `test e2e PR1,PR2`, time to go finish writing that.

PR #24778, opened 2020-03-27
hunting ResourceQuota and capture the life of a secret

p0lyn0mial:


p0lyn0mial:
@mfojtik @sttts 

I think that the way the test counts the expected number of secrets is not deterministic. Especially under moderate load. 

I wrote a script [rage.sh](https://github.com/openshift/origin/pull/24778/commits/bed0aa3cae63bcecebc8f7b7e209835c27115866) that runs the test **50** times in parallel on my local machine. The output (the [log](https://github.com/openshift/origin/pull/24778/commits/b2477b83b28b61af7506cb374ee245809533b5b8) file ) is attached as a commit.

It shows that the number of expected vs actual secrets diverges quickly. For example "**expected 6, actual 7"**, "**expected 6, actual 8**", "**expected 3, actual 6**".

It's worth noting that the default number of secrets in **4.5** cluster is **9**.

The secrets are generated by various controllers I suspect they are simply throttled. To properly fix the test I think we would need a way of knowing a namespace has been fully initialized and it's ready to be consumed.

Additionally, we could inspect the individual controllers to see if there are no glaring errors that would slow them down, for example, https://github.com/openshift/openshift-controller-manager/blob/master/pkg/serviceaccounts/controllers/create_dockercfg_secrets.go#L68

p0lyn0mial:
spoke with @mfojtik, it seems that this PR https://github.com/openshift/openshift-controller-manager/pull/72/files#diff-bc9c443ebaeefc1c4319cb1018e62b8cR17 increased the number of times the test fails substantially.

PR #24786, opened 2020-03-27
Bug 1828427: Add CI test to check for critical alerts post upgrade success

jottofar:


test/e2e/upgrade/alert/alert.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// Block until upgrade is done
+	g.By("Waiting for upgrade to finish before checking for critical alerts")
+	<-done
+
+	_, cancel := context.WithCancel(context.Background())
+	start := time.Now()
+
+	g.By("Waiting before checking for critical alerts")
+	time.Sleep(alertCheckSleep)
```
p0lyn0mial, line 76: out of curiosity, why do you have to sleep here?

jottofar: This change is to satisfy [OTA-141](https://issues.redhat.com/browse/OTA-141) whereby we want to check if critical alerts are present on cluster post-upgrade after some period. So we want to give the cluster a chance to settle/run steady state after the upgrade and then check for any critical alerts.

wking: Would be good to capture this motivation^ in the commit message.

test/e2e/upgrade/alert/alert.go
```diff
@@ -0,0 +1,227 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// Additonal delay after upgrade completion
+	g.By("Waiting before checking for critical alerts")
+	time.Sleep(alertCheckSleep)
+	cancel()
+
+	if len(os.Getenv("TEST_UNSUPPORTED_ALLOW_VERSION_SKEW")) > 0 {
+		e2eskipper.Skipf("Test is disabled to allow cluster components to have different versions, and skewed versions trigger multiple other alerts")
```
wking, line 92: Instead of duplicating logic like this for both the update and non-update suites, can we factor the logic out into helpers that are shared between the two consumers?

wking:
Not sure quite where it would fit in, but I think we want to include [the `should have a Watchdog alert in firing state the entire cluster run` check][1] as well.  Otherwise we risk false-negatives if Prometheus just falls over on us.

[1]: https://github.com/openshift/origin/blob/11c2afd825914e0968372b86e549d7e2061cf669/test/extended/prometheus/prometheus.go#L70

test/e2e/upgrade/alert/alert.go
jottofar: This commit was not to specific to this comment. The reason I have so many commits is that I was doing all my testing through CI (becuase I couldn't get the upgrade e2e test to run myself) so Iwas hacking around, testing, developing, etc. The sleep was always going to be there as part of the overall design of the product

test/e2e/upgrade/alert/alert.go
jottofar: Yes, agree. I was waiting until we sorted out "our side" before coming up with reusables and I also wasn't sure if we wanted to head down that path or just get our stuff out there. Now I know :)

jottofar:
> Not sure quite where it would fit in, but I think we want to include [the `should have a Watchdog alert in firing state the entire cluster run` check](https://github.com/openshift/origin/blob/11c2afd825914e0968372b86e549d7e2061cf669/test/extended/prometheus/prometheus.go#L70) as well. Otherwise we risk false-negatives if Prometheus just falls over on us.

Yep, I'll add that back. I removed it as part of my testing and experimenting and meant to add it back.

jottofar:
> Not sure quite where it would fit in, but I think we want to include [the `should have a Watchdog alert in firing state the entire cluster run` check](https://github.com/openshift/origin/blob/11c2afd825914e0968372b86e549d7e2061cf669/test/extended/prometheus/prometheus.go#L70) as well. Otherwise we risk false-negatives if Prometheus just falls over on us.

@wking I pushed up a change that checks for Watchdog if no critical alerts are found to be firing.

My concern though is that after 10-20 runs I have not had the situation where there has not been a crtical alert when checking the past 1 minute. Most recently have been getting alert KubeAPIErrorBudgetBurn. The result being of course that the test fails.

test/e2e/upgrade/alert/alert.go
wking: > The sleep was always going to be there as part of the overall design of the product.

OTA-141 is not public (at the moment).  And even for folks who do have access, it's a separate click out of `git blame ...` beyond the commit that landed the code they're wondering about.  So it's nice to outline the design in the commit that lands the implementation so it's conveniently close for anyone poking about in the public Git repo.

wking:
> Most recently have been getting alert KubeAPIErrorBudgetBurn.

This is [rhbz#1821661](https://bugzilla.redhat.com/show_bug.cgi?id=1821661), and I'm fine with this landing and blocking further 4.5 CI until that bug gets sorted, or idling this PR until that bug gets sorted.  I don't think it's worth adding special workarounds so we can land this PR while that bug is still unfixed without failing tests, because this is exactly the sort of issue that this test extension is designed to detect.

test/extended/util/prometheus/helpers.go
```diff
@@ -0,0 +1,94 @@
...
+	kapierrs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/kubernetes/pkg/client/conditions"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// PrometheusResponse is used to contain prometheus query results
+type PrometheusResponse struct {
```
smarterclayton, line 25: Did this package get deleted or moved?

smarterclayton: We used to have a package exactly like this.

test/extended/util/prometheus/helpers.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	_, err = watchtools.UntilWithoutRetry(ctx, w, conditions.ServiceAccountHasSecrets)
+	return err
+}
+
+// LocatePrometheus returns information used to make http request to Prometheus
+func LocatePrometheus(oc *exutil.CLI) (url, bearerToken string, ok bool) {
```
smarterclayton, line 65: We might just want to move this into the global helper package and thne clean up all references

test/extended/util/client.go
```diff
@@ -123,6 +131,15 @@ func NewCLIWithoutNamespace(project string) *CLI {
 	return cli
 }
 
+// NewCLIWithoutNamespacaInside initializes the CLI and Kube framework helpers
+// without a namespace. It can be called inside of a Ginkgo .It()
+// function. Use SetupProject() to create a project for this namespace.
+func NewCLIWithoutNamespaceInside(project string) *CLI {
```
smarterclayton, line 137: You can't create frameworks inside of a function, that's not what frameworks are for.  You should never have more than one framework per test

test/e2e/upgrade/alert/alert.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	o.Expect(errors).To(o.BeEmpty())
+}
+
+// Teardown cleans up any remaining resources.
+func (t *UpgradeTest) Teardown(f *framework.Framework) {
+	// rely on the namespace deletion to clean up everything
+}
+
+func runQuery(f *framework.Framework, query string, expected bool, oc *exutil.CLI, ns, execPodName, baseURL, bearerToken string) map[string]error {
```
smarterclayton, line 118: This method is super common, I would go find one of these and move it into a separate package that everyone shares

test/e2e/upgrade/alert/alert.go
```diff
@@ -0,0 +1,167 @@
...
+func (UpgradeTest) DisplayName() string {
+	return "Check if critical alerts are firing after upgrade success"
+}
+
+// Setup creates parameters to query Prometheus
+func (t *UpgradeTest) Setup(f *framework.Framework) {
+	g.By("Setting up post-upgrade alert test")
+
+	// Must use version that's can run within Ginkgo It
+	oc := exutil.NewCLIWithoutNamespaceInside("prometheus")
```
smarterclayton, line 54: You can't do this, you need to use the framework you get passed.

test/e2e/upgrade/alert/alert.go
```diff
@@ -0,0 +1,167 @@
...
+
+	var ok bool
+	url, bearerToken, ok := helper.LocatePrometheus(oc)
```
smarterclayton, line 57: I would reduce this to "ExpectPrometheus" and have that be a helper, it shouldn't require anything except a framework.

test/extended/util/client.go
jottofar: Does this address the concern?

`
func NewCLIWithFramework(kubeFramework *framework.Framework) *CLI {
        cli := &CLI{
                kubeFramework:   kubeFramework,
                username:        "admin",
                execPath:        "oc",
                adminConfigPath: KubeConfigPath(),
        }
        return cli
}
...
// NewCLIWithoutNamespacWithFramework initializes the CLI using the provided Kube
// framework without a namespace. It can be called inside of a Ginkgo .It()
// function.
func NewCLIWithoutNamespaceWithFramework(kubeFramework *framework.Framework) *CLI {
        cli := NewCLIWithFramework(kubeFramework)
        cli.kubeFramework.SkipNamespaceCreation = true
        return cli
}`

smarterclayton: No, just don't use this path.  Yeah, new with framework is fine.  Don't do two variants, just straight up take a framework.

test/extended/prometheus/prometheus_builds.go
```diff
@@ -61,12 +53,12 @@ var _ = g.Describe("[sig-instrumentation][sig-builds][Feature:Builds] Prometheus
 			// allow for some retry, a la prometheus.go and its initial hitting of the metrics endpoint after
 			// instantiating prometheus tempalte
```
wking, line 54: nit: reach over and fix this pre-existing "tempalte" -> "template" here?

test/extended/prometheus/prometheus_builds.go
```diff
@@ -61,12 +53,12 @@ var _ = g.Describe("[sig-instrumentation][sig-builds][Feature:Builds] Prometheus
...
 			var err error
-			for i := 0; i < maxPrometheusQueryAttempts; i++ {
+			for i := 0; i < helper.MaxPrometheusQueryAttempts; i++ {
```
wking, line 56: nit: By making this 403 polling a `helper` function, you could make `maxPrometheusQueryAttempts` and `prometheusQueryRetrySleep` internal to the `helper` package.  Not a big deal, but I think that would make for a slightly cleaner API.

wking:
[e2e-gcp-upgrade][1] is failing on this test and `KubeAPIErrorBudgetBurn`.  No need to keep kicking it, although we should give the API folks some heads up that it will be fatal after this PR lands.  If we want to buy them more time while guarding against backsliding on other alerts, we could temporarily append them to the `Watchdog|AlertmanagerReceiversNotConfigured|KubeAPILatencyHigh` exceptions.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24786/pull-ci-openshift-origin-master-e2e-gcp-upgrade/3881

test/e2e/upgrade/alert/alert.go
```diff
@@ -83,8 +83,9 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 	// period by verifying Watchdog alert has been in firing state
 	watchdogQuery := fmt.Sprintf(`count_over_time(ALERTS{alertstate="firing",alertname="Watchdog", severity="none"}[%dm])`, alertCheckSleepMinutes)
 
-	// Query to check for any critical severity alerts that have occurred within the last alertPeriodCheckMinutes
-	criticalAlertQuery := fmt.Sprintf(`count_over_time(ALERTS{alertname!~"Watchdog|AlertmanagerReceiversNotConfigured|KubeAPILatencyHigh",alertstate="firing",severity="critical"}[%dm]) >= 1`, alertPeriodCheckMinutes)
+	// Query to check for any critical severity alerts that have occurred within the last alertPeriodCheckMinutes.
```
smarterclayton, line 86: when you squash can you keep this in its own commit but put the other changes in this file into the other commit?  Basically 2 - one for this, one for your test?

wking:
[e2e-cmd][1]:

```
fail [github.com/openshift/origin/test/extended/cmd/cmd.go:118]: Expected
    <[]error | len:1, cap:1>: [
        {
            s: "error waiting for the pod 'test-cmd' to complete: the pod errored trying to run the command",
        },
    ]
to have length 0
...
failed: (1m3s) 2020-04-24T21:57:17 "[sig-cli][Feature:LegacyCommandTests][Disruptive][Serial] test-cmd: test/cmd/admin.sh [Suite:openshift]"
```

Dunno what that's about.

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24786/pull-ci-openshift-origin-master-e2e-cmd/5284

test/e2e/upgrade/alert/alert.go
jottofar: Done

PR #24794, opened 2020-03-30
Split verify-commits

tnozicka:
The check is meant to be overridden for rebases, it's easier and more reliable to do in a separate job. Needs to go in 2 PR so nothing sneaks in between until the CI is switched over and we can't merge a new CI check before it is in the repo.

/cc @sttts @marun

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,7 @@ check: | build verify
 #
 # Example:
 #   make verify
+# TODO: remove verify-upstream-commits when CI is switched over
 verify: build
 	# build-tests task has been disabled until we can determine why memory usage is so high
 	{ \
@@ -96,6 +97,9 @@ verify: build
 	}
 .PHONY: verify
 
+verify-commits:
+	hack/verify-upstream-commits.sh
+.PHONY: verify-commits
 
 # Update all generated artifacts.
 #


mfojtik:
nack, please don't add new jobs and don't make CI more complex. the rebase case can be handled with one override command and it is one-off

tnozicka:
On the rebase flow this specific check needs to be overridden. Having it's own job makes it safer as all regular verify checks still apply.

tnozicka:
the split also makes verify rehearsals succeed as the commit checker has commit id issues there https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/7992/rehearse-7992-pull-ci-openshift-origin-master-verify/3

marun:
I think it's a good idea to minimize the scope of override. Unless the approver of a rebase PR is manually verifying that `make verify` passes for everything but the commit checker, accidents are likely to happen. All the more so if novices like me are doing the rebasing!

marun:
This may be worth merging to the active release-x.x branches, but master would no longer benefit from it. Note that openshift/kubernetes, inspired by this PR, does run a separate verify-commits job to simplify rebase review.

PR #24802, opened 2020-03-31
Update test data to pass validation

soltysh:
This PR is cleaning our testdata to be able to turn on client-side validation in https://github.com/openshift/oc/pull/359
/assign @sttts


smarterclayton:
/hold

Why would these objects have to change?  If validation is correct, then objects that are valid on the server are valid on the client?

sttts:
> Why would these objects have to change? If validation is correct, then objects that are valid on the server are valid on the client?

+1 – see aos-devel thread. Our spec is still too strong.

PR #24805, opened 2020-03-31
Revert "Bug 1797788: add openshift-tests as a future delegation binary"

smarterclayton:
Reverts openshift/origin#24560


smarterclayton:
Covered in #24808

PR #24808, opened 2020-03-31
Revert "Bug 1797788: add openshift-tests as a future delegation binary"

smarterclayton:


smarterclayton:
I'm going to revert this for now while we sort out options - I had some larger concerns that there may be an alternative direction on.

smarterclayton:
Master is flaky due to dockercfg secret, direct merging to get tests in better shape.  Passed all tests and batch tests at least once.

PR #24811, opened 2020-03-31
monitor: Report node when events include it

smarterclayton:
Also tone down the error state when a pod on a nade moves back to
pending (such as when it is restarted on reboot).


smarterclayton:
Brings more standardization and faster scanning to event log, captures ready transformation, adds node info to many more event log lines

PR #24815, opened 2020-04-01
integration: fix TestOadmPodNetwork for CI changes

danwinship:
In the new CI cluster, test-integration gets run in a pod in a cluster
using the default OCP IP ranges, which broke TestOadmPodNetwork
because that test brings up enough of openshift-sdn that it complains
when it sees that eth0 has an IP inside the range it was planning to
use.

The test doesn't actually allocate any IPs, so it doesn't matter what
ranges the master is configured to use, so just set it to something
invalid so it's guaranteed to not overlap no matter what.

### DIFF
--- a/test/integration/sdn_test.go
+++ b/test/integration/sdn_test.go
@@ -10,6 +10,7 @@ import (
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 	restclient "k8s.io/client-go/rest"
 
+	"github.com/openshift/origin/pkg/cmd/server/apis/config"
 	"github.com/openshift/origin/pkg/network"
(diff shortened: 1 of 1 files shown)


danwinship:
/assign @smarterclayton 
can you just merge this? It unbreaks 3.11-integration. I don't feel like dancing with the bugzilla bot.

/cc @stevekuznetsov

danwinship:
/assign @dcbw @knobunc @squeed 
just realize the bot doesn't actually care if this has a bug associated with it. yay

danwinship:
/retest
for e2e-gcp and images, but unit is going to fail again (#24874) so this needs to be force-merged

PR #24819, opened 2020-04-02
Bug 1740604: UPSTREAM 89300: Expand the PVC validation messages

huffmanca:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
When making modifications to an immutable field we don't print the field that's attempted to be edited, only the included message. This PR makes the messages more user-friendly. 

**Which issue(s) this PR fixes**:
Bug 1740604


jsafrane:
/retitle Bug 1740604: UPSTREAM 89300: Expand the PVC validation messages

smarterclayton:
Retag once master is merging - retries not helping

jsafrane:
/lgtm
aws-serial is green

PR #24824, opened 2020-04-03
UPSTREAM: 89592: log previous pod logs if pod was removed

dcbw:
`Apr  2 20:19:21.689: INFO: unable to fetch logs for pods: paused-1-lzxdz[e2e-test-cli-deployment-79fhw].container[myapp].error=the server could not find the requested resource (get pods paused-1-lzxdz)`

Companion to https://github.com/openshift/origin/pull/24799 except this part of the code is not upstream so must be carried.

@danwinship @knobunc

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/framework/pod/resource.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/framework/pod/resource.go
@@ -27,6 +27,7 @@ import (
 	"github.com/onsi/gomega"
 
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


sjenning:
@dcbw is there any reason this could not be accepted upstream as well so we don't have to carry?

dcbw:
> @dcbw is there any reason this could not be accepted upstream as well so we don't have to carry?

@sjenning also the code I'm touching here isn't upstream anyway; it's already <carry> code.

dcbw:
@danwinship @sjenning PTAL thanks!

danwinship:
/lgtm
but honestly we should just rip out that code. I asked about this upstream and you claimed you really did care about the pod logs, but now that I have more context, I realize you don't, you just want the test to fail with the right error message instead of the wrong one

PR #24840, opened 2020-04-06
Add ci-monitor-operator to openshift-tests

damemi:


damemi:
/cc @mfojtik 
Including the command in this way is giving me some weird deps changes that I think are failing CI (like removing a bunch of k8s repos). I also tried including the `RunOperator` command directly but got the same results. I'm wondering if we should be trying to import an operator into origin that depends on library-go, wdyt?

mfojtik:
@damemi i would try to copy the repo into origin, wdyt?

PR #24845, opened 2020-04-07
openshift-tests: add run-resourcewatch command

mfojtik:
This command will start watching changes in `*.config.openshift.io` resources (including all cluster operators) and record all changes into local Git repository as a "commit".

Usage:

```
$ openshift-tests run-resourcewatch
```

If running inside the pod, it will use the SA token of the pod.

Additionally, it will produce Prometheus metrics that will contain timeline of changes in clusteroperator resource status:

```
# HELP openshift_ci_monitor_operator_cluster_operator_status [ALPHA] A metric that tracks individual cluster operator status.
# TYPE openshift_ci_monitor_operator_cluster_operator_status gauge
openshift_ci_monitor_operator_cluster_operator_status{condition="Available",name="authentication",status="True"} 1.586055762e+09
openshift_ci_monitor_operator_cluster_operator_status{condition="Available",name="cloud-credential",status="True"} 1.586054345e+09
```

Follow ups:

* Wire this into jobs that run openshift-tests (e2e-aws, etc..)
* Make sure we archive the Git repository and copy it into artifacts
* Figure out if we can make service monitor and the aggregated prometheus to scrape the metrics from this operator.

More details: https://github.com/mfojtik/ci-monitor-operator


glide.lock
```diff
@@ -2213,7 +2312,7 @@ imports:
   - pkg/apis/deviceplugin/v1beta1
   - pkg/apis/pluginregistration/v1
 - name: k8s.io/kubernetes
-  version: 5227b6f968eb46decfbb2d42e9827448a9b3910d
+  version: 3ee09fda8c72ef2489803eb69faec1e1f497302d
```
tnozicka, line 2315: you can't do that except for rebases

tnozicka: To be more precise the vendor changes should be empty if publishing bot works. I am not sure if the bot also re-bumps origin glide to published levels, so this isolated change (without the vendor/k8s.io changes) might be fine.

deads2k:
yeah, you changed vendored k8s by accident. don't mess with that.

mfojtik:
@deads2k publishing bot was broken, i'm revendoring now.

mfojtik:
@deads2k @damemi this is ready for review

pkg/monitor/resourcewatch/controller/configmonitor/crd_controller.go
```diff
@@ -0,0 +1,163 @@
...
+				gvk := schema.GroupVersionKind{
+					Group:   gv.Group,
+					Version: gv.Version,
+					Kind:    crd.Spec.Names.Kind,
+				}
+				if currentKinds.Has(gvk.Kind) {
+					continue
+				}
+				currentKinds.Insert(gvk.Kind)
+				currentConfigResources = append(currentConfigResources, gvk)
```
damemi, line 86: nit: you could just condense these into a `if !currentKinds.Has(gvk.Kind)` block and drop the `continue`

damemi:
/retitle Bug 1842631: openshift-tests: add run-resourcewatch command

PR #24851, opened 2020-04-09
Write an event when the server is ready to serve

tkashem:
Write an event to indicate that the kube-apiserver has bound to a port and is ready to serve request(s). This will come in handy to debug issues n perf testing, we can time when the apiserver stops serving and comes back up again and ready to serve during a roll out.

### DIFF
--- a/vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
+++ b/vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
@@ -395,6 +395,8 @@ func (s preparedGenericAPIServer) NonBlockingRun(stopCh <-chan struct{}) error {
 			close(auditStopCh)
 			return err
 		}
+
+		s.Eventf(corev1.EventTypeNormal, "ReadyToServeRequest", "The server has started and is ready to serve request(s)")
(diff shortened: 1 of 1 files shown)


sjenning:
We typically to not make direct changes to the vendor tree that are not upstream.

Additionally, I think there is a `/healthz` endpoint on the apiserver for this.

/cc @sttts @deads2k

tkashem:
> We typically to not make direct changes to the vendor tree that are not upstream.
> 
> Additionally, I think there is a `/healthz` endpoint on the apiserver for this.

@sjenning I usually make changes directly to the vendor folder to deploy a custom kube-apiserver on my cluster.

This is not in upstream, we (as in OpenShift) add these events when the apisever rolls out. 
May be https://github.com/openshift/kubernetes/tree/master/staging/src/k8s.io/apiserver is where I should add it if we decide this event is useful.

sttts:
> We typically to not make direct changes to the vendor tree that are not upstream.
> 
> Additionally, I think there is a `/healthz` endpoint on the apiserver for this.
> 
> /cc @sttts @deads2k

We have extended our events already considerably with a carry patch. Upstream does not care much about graceful rolling upgrades, we do.

But following @sjenning's suggestions, isn't there an even already created by the kubelet when a pod become healthy?

PR #24853, opened 2020-04-09
tests: skip secured metrics test for ovn-kubernetes

dcbw:
To deliver bits to somebody who really needs them, we need to
skip this test for now. Issue filed to fix it, and there's already a
PR in progress but it's more complicated for network plugins to do
this than other components.

https://issues.redhat.com/browse/SDN-912

@eparis @knobunc @fepan


dcbw:
The specific test ovnkube fails is "some services expose metrics over insecure channel" but I'm not sure how to only skip that one test.

knobunc:
@dcbw is this the right way -- https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L253-L268 ?

dcbw:
Closing in preference to https://github.com/openshift/origin/pull/24857

PR #24868, opened 2020-04-13
Bug 1731263: Disable failing preemption e2es

damemi:
Main BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1731263. BZ contains discussion on failure causes, usually that the "preemptor" pod does not land on the expected "victim" pod's node, and so it does not evict the pod we expect it to.

These tests have been disabled for several releases, and were recently re-enabled by https://github.com/openshift/origin/commit/5f52d7e1d9007956a243c7a84d3a9774f230270e#diff-6ba77494282f6e840f44b01ce97335afL88

After being reenabled, they are still failing: https://testgrid.k8s.io/redhat-openshift-ocp-release-4.5-informing#release-openshift-ocp-installer-e2e-azure-serial-4.5

These have had the following attempted fixes:
* adjust usage calculations to be more accurate: https://github.com/openshift/origin/pull/23645
* (upstream) up-port(?) of the above changes: https://github.com/kubernetes/kubernetes/pull/82350
(among others that have been scrapped or overwritten by the above)

There was an upstream issue for some of these tests here https://github.com/kubernetes/kubernetes/issues/88441. They do not currently flake in upstream CI runs

There are some comments in this PR explaining how the test currently assigns a pod to a node with affinity: https://github.com/kubernetes/kubernetes/pull/90118. I believe my attempt to fix this by manually setting the `NodeName` won't actually work because that bypasses scheduling, and so preemption does not run.


damemi:
@soltysh what are you referring to? there doesn't appear to be any major problems with the test upstream, and I think the issue I linked was actually caused by us trying to patch up the test to work for our clusters

soltysh:
/retitle Bug 1731263: Disable failing preemption e2es

PR #24869, opened 2020-04-13
update Makefile to set GO11MODULE=off and generate test/extended

sallyom:
Until/unless origin makes switch to gomod, need to set GO111MODULE=off in build scripts.

### DIFF
--- a/hack/update-generated-bindata.sh
+++ b/hack/update-generated-bindata.sh
@@ -2,6 +2,8 @@
 STARTTIME=$(date +%s)
 source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
+os::golang::ensure_gomod_off
+
 go generate ./test/extended
 
 os::build::setup_env
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -20,6 +20,8 @@ if [[ -z "$@" ]]; then
     vendor/k8s.io/kubernetes/cmd/kubelet)
 fi
 
+os::golang::ensure_gomod_off
+
 OS_BUILD_PLATFORMS=("${OS_BUILD_PLATFORMS[@]:-${platform}}")
 os::build::build_binaries "${build_targets[@]}"
 os::build::place_bins "${build_targets[@]}"
--- a/hack/lib/util/golang.sh
+++ b/hack/lib/util/golang.sh
@@ -34,3 +34,9 @@ function os::golang::verify_glide_version() {
 
 }
 readonly -f os::golang::verify_glide_version
+
+# os::golang::ensure_gomod_off ensures gomod is off
+function os::golang::ensure_gomod_off() {
+    export GO111MODULE=off
+}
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -56,6 +57,7 @@ build-tests: build-extended-test
 .PHONY: build-tests
 
 build-extended-test:
+	go generate ./test/extended/
```
smarterclayton, line 60: This is covered by hack/update-generated-bindata.sh, you should be calling that.  And we generally don't call update bindata automatically.

Makefile
```diff
@@ -12,6 +12,7 @@ OUT_DIR = _output
 
 export GOFLAGS
 export TESTFLAGS
+export GO111MODULE=off
```
smarterclayton, line 15: What's the reason for this?  Go 1.14?

Makefile
sallyom: it took me more than a few minutes to figure out the 'test not found' panic after adding a new test/extended/file.  I didn't see any documentation in the test/extended/README about generating `test/extended/util/annotate/generated/zz_generated.annotations.go`.  I can add that documentation instead, then.

Makefile
sallyom: go 1.13 has GO111MODULE=on by default AFAIK. Build fails unless you set GO111MODULE=off

Makefile
sallyom: I know it's also in hack/update-generated-bindata.sh, but that script also fails unless I add 
`export GO111MODULE=off`

Makefile
smarterclayton: Then it should go into the library helper for build to set that (same place we set up go env) - hack/lib/util/golang.sh

hack/update-generated-bindata.sh
```diff
@@ -2,6 +2,7 @@
 STARTTIME=$(date +%s)
 source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
+export GO111MODULE=off
```
smarterclayton, line 5: go generate fails if it can't load modules?  If i run the current script on a go 1.13 system it works - why would this be needed?

sallyom:
```console
$ go version
go version go1.14.1 linux/amd64
$ ./hack/update-generated-bindata.sh 
go: finding module for package github.com/openshift/origin/test/extended/authorization
go: finding module for package github.com/openshift/origin/test/extended/apiserver
---
[ERROR] hack/update-generated-bindata.sh:6: `go generate ./test/extended` exited with status 1.
```

Makefile
sallyom: ok, updated

sallyom:
Upon further review, I had a rogue `go.mod` file in my home dir that was causing go to _always_ run with go mod on.  After deleting this rogue go.mod, go properly detects there is no go.mod and does the right thing. Closing this.

PR #24871, opened 2020-04-14
Bug 1820731  Can't show the message in time

emmajiafan:
Remove the test of status.currentPhase.phase.message since the message can't be shown in time , keep the status.name test .


emmajiafan:
@kevinrizza , I fix this bug by removing the check of "status.currentPhase.phase.message" since it can't show in time. Could you help take a look at it please? Thank you!

emmajiafan:
close this pr since the fix should merge to 4.4

PR #24875, opened 2020-04-14
add extended clusteroperators delete operand namespace recover test

sallyom:
e2e to delete clusteroperators' operand ns, ensure cluster recovery,
that is, all COs are Available=True Degraded=False Progressing=False,
except "kube-apiserver", deemed healthy with Progressing=True to meet e2e time limit.

This test deletes operand namespaces and waits for each to be terminated before deleting another. The recovery happens in parallel. This test is more to check that individual cluster operators recover from operand namespace deletion, rather than recovery from deletion of _all_ operand namespaces at once. 
 
This test is necessary because deletion of an operand namespace has an effect on another
clusteroperator's health.  For example, deleting openshift-service-ca namespace causes openshift-console
to progress, degrade (if service-ca not recovered).  kube-controller-manager co also affects other COs.

### DIFF
--- a/cmd/openshift-tests/e2e.go
+++ b/cmd/openshift-tests/e2e.go
@@ -103,7 +103,7 @@ var staticSuites = testSuites{
 				if strings.Contains(name, "Cluster should survive master and worker failure and recover with machine health checks") {
 					return false
 				}
-				return strings.Contains(name, "[Feature:EtcdRecovery]") || strings.Contains(name, "[Feature:NodeRecovery]") || isStandardEarlyTest(name)
+				return strings.Contains(name, "[Feature:EtcdRecovery]") || strings.Contains(name, "[Feature:NodeRecovery]") || strings.Contains(name, "[Feature:ClusterOperatorRecovery]") || isStandardEarlyTest(name)
 
 			},
 			// Duration of the quorum restore test exceeds 60 minutes.
--- a/test/extended/dr/operand_ns_recover.go
+++ b/test/extended/dr/operand_ns_recover.go
@@ -1,0 +1,159 @@
+package dr
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/kube-openapi/pkg/util/sets"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	config "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+)
+
+var _ = g.Describe("[sig-arch][Disruptive] Managed cluster should recover", func() {
+	defer g.GinkgoRecover()
+	// Delete clusteroperator-owned operand ns, ensure cluster recovery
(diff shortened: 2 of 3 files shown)


test/extended/operators/clusteroperators_operand_ns.go
```diff
@@ -0,0 +1,160 @@
...
+		hasOperandNS := []string{
+			"dns",
+			"etcd",
+			"kube-scheduler",
+			"kube-controller-manager",
+			"kube-apiserver",
+			"service-ca",
+			"openshift-apiserver",
+			"network",
+			// these use CVO to delegate operand namespace, ns can take several minutes to recreate
```
sallyom, line 40: or rather, delegate ns creation to cvo - will fix grammar

sallyom:
seems when i add [Slow] to a test, that test gets excluded from all suites. :woman_shrugging:

sallyom:
note: expect this to fail until the PRs listed above merge

test/extended/operators/recover.go
```diff
@@ -93,7 +94,129 @@ var _ = g.Describe("[sig-arch] Managed cluster should recover", func() {
...
+			buf := &bytes.Buffer{}
+			w := tabwriter.NewWriter(buf, 0, 1, 1, ' ', 0)
+			for _, operator := range currentOperators.Items {
+				fmt.Fprintf(w, "%s\t%s\t%s\n", operator.Name, conditionStatus(&operator, config.OperatorAvailable), conditionStatus(&operator, config.OperatorDegraded))
+			}
+			w.Flush()
+			o.Expect(err).ToNot(o.HaveOccurred(), fmt.Sprintf("Operators never became available: %s\n%s", strings.Join(operatorNames(operators.Items).Difference(operatorNames(healthy)).List(), ", "), buf.String()))
+		}
+	})
+	g.It("when operator-owned operand namespace is deleted [Disruptive]", func() {
```
soltysh, line 123: Tags should be at the top next to `g.Describe`, also add there one more unique tag so we can ensure we select only these in that separate test-suite we want to add. Something like `[OperatorRecovery]`, for example.

sallyom: I was following suit w/ the other test-case in this file: https://github.com/openshift/origin/blob/0fae074b297b0d7009fdd2cea2c37ec99295af16/test/extended/operators/recover.go#L28 - I'll move the [Disruptive] for _both_ above, in the `g.Describe` though.

sallyom:
Added a commit to delete 3 (random) operand ns at a time, then wait for recovery.  I was thinking this might make the test faster, but it does not.  Bulk deleting all then wait for 1 recovery takes ~11min.  Deleting 3 at a time then waiting for recovery after each group of 3 takes _a lot_ longer.

sallyom:
these are failing, and I _think_ the var `TEST_LIMIT_START_TIME="$(date +%s)` is set for the rehearse job, that means the conformance tests should not fail due to conditions during disruptive test run. 
```
Failing tests:
[sig-instrumentation] Prometheus when installed on the cluster should have important platform topology metrics [Suite:openshift/conformance/parallel]
[sig-instrumentation] Prometheus when installed on the cluster should provide ingress metrics [Suite:openshift/conformance/parallel]
[sig-instrumentation] Prometheus when installed on the cluster should start and expose a secured proxy and unsecured metrics [Suite:openshift/conformance/parallel]
[sig-instrumentation] Prometheus when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured [Early] [Suite:openshift/conformance/parallel]
```

sallyom:
this e2e will be added to disruptive suite
#25774 fixes the etcd quorum restore disruptive test so rather than add a commit to skip that this can wait for #25774 to merge

PR #24883, opened 2020-04-16
clarify test failure message

bparees:


adambkaplan:
/hold

The tests are ultimately flaking due to https://bugzilla.redhat.com/show_bug.cgi?id=1823949. Based on the current thread this fix won't land in the immediate future. IMO the test should only check that the leader has the internal registry hostname  - we don't need to wait for everything to be rolled out.

bparees:
> Based on the current thread this fix won't land in the immediate future. 

it better land in the immediate future, it's a major product bug.  We can't ship 4.4 w/ it.  And we can't tolerate it in 4.5 for any length of time.

> IMO the test should only check that the leader has the internal registry hostname - we don't need to wait for everything to be rolled out.

while it could do that, there is no reason under normal circumstances that the daemonset should not be successfully rolled out.  Since we didn't seem to catch this elsewhere, if we're going to drop that check here, we need to ensure there is a test somewhere else that would catch cases like this.  Personally i'd just leave it as is.

PR #24889, opened 2020-04-18
Bug 1826499: IngressClass error test is flaky

smarterclayton:
The test races with the admission cache and is failing 5-7% of the
time in e2e runs.

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/network/ingressclass.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/network/ingressclass.go
@@ -19,10 +19,12 @@ package network
 import (
 	"context"
 	"strings"
+	"time"
 
 	networkingv1beta1 "k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


ironcladlou:
I actually suspect the admission controller itself is racy — if so, this change might be hiding the root problem, but I could be wrong

smarterclayton:
Admission is completely racy.  An admission controller is an observer of cluster state and is independent across all three masters.  You may not assume admission happens across all masters at some point in time.

ironcladlou:
What I mean is, in this specific case, my theory is a stale cache means the admission plugin erroneously admits the Ingress (i.e. were the cache to be fresh, the plugin would have rejected), and the system takes no corrective action to retroactively reject or repair the erroneously admitted Ingress.

smarterclayton:
That is how admission works.  The same applies the other way - that's also how other admission controls work that check RBAC.

ironcladlou:
>That is how admission works. The same applies the other way - that's also how other admission controls work that check RBAC.

Hrm — what exactly does admission set out to guarantee, then?

ironcladlou:
The Ingress admission code seems to be declaring some invariant ("there shall exist only one 'default' IngressClass") but it's not clear to me that's actually enforced. Is that actually an invariant that needs enforcing? If so, how?

smarterclayton:
If you create two, then some O(seconds) later you can't create new ingress.

smarterclayton:
Admission in the presence of config or input changes has always behaved like this

ironcladlou:
I believe you — it's just not something I'm deeply familiar with. Thanks for the info!

PR #24900, opened 2020-04-20
Bug 1823494: Storage tests should more carefully select zones for testing

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1823494

cc @openshift/storage

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/storage/drivers/in_tree.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/storage/drivers/in_tree.go
@@ -1316,6 +1316,7 @@ func (g *gcePdDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTestCo
 		Prefix:    "gcepd",
 		Framework: f,
 	}
+
 	if framework.NodeOSDistroIs("windows") {
 		config.ClientNodeSelection = e2epod.NodeSelection{
 			Selector: map[string]string{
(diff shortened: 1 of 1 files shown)


gnufied:
The azure failures are unrelated to this PR and issues that this PR set out to fix has been fixed - https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24900/pull-ci-openshift-origin-master-e2e-azure/46 (see no more inline volume failures!).

I think e2e-azure is not a blocking job, but since I summoned it manually, we can't merge this PR until it passes.

smarterclayton:
Looks like storage tests on azure  are failing for other reasons now after this fix:

```
Apr 22 21:33:58.269: INFO: At 2020-04-22 21:27:52 +0000 UTC - event for pvc-qvt4m: {persistentvolume-controller } ProvisioningFailed: storageclass.storage.k8s.io "e2e-volume-7559" not found
Apr 22 21:33:58.269: INFO: At 2020-04-22 21:31:20 +0000 UTC - event for azure-injector: {attachdetach-controller } FailedAttachVolume: AttachVolume.Attach failed for volume "azure-disk-hdtrl" : Retriable: false, RetryAfter: 0s, HTTPStatusCode: 409, RawError: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 409, RawError: {
  "error": {
    "code": "OperationNotAllowed",
    "message": "The maximum number of data disks allowed to be attached to a VM of this size is 4.",
    "target": "dataDisks"
  }
}
```
and then the injector pod timeout.  Are there bugs open tracking those?

smarterclayton:
Also, 4 data disks is crazy small.  Is that a bug?

gnufied:
@smarterclayton this is the table used for computing disk attach limits for Azure - https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/pkg/volume/azure_dd/azure_dd_max_disk_count.go#L23

PR #24906, opened 2020-04-21
WIP dynamic network readiness

danwinship:
WIP, for testing

Adds a feature to kubelet to soft-block pod-network pods until a (host-network) pod appears claiming to be the network plugin.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/kubelet/apis/config/types.go
+++ b/vendor/k8s.io/kubernetes/pkg/kubelet/apis/config/types.go
@@ -308,6 +308,9 @@ type KubeletConfiguration struct {
 	// These sysctls are namespaced but not allowed by default.  For example: "kernel.msg*,net.ipv4.route.min_pmtu"
 	// +optional
 	AllowedUnsafeSysctls []string
+	// If true, Kubelet will not attempt to start non-host-network pods unless there is a Running pod
+	// on the node with a "networking.k8s.io/network-plugin" label.
+	DynamicNetworkReadiness bool
 
 	/* the following fields are meant for Node Allocatable */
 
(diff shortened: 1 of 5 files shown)


danwinship:
@dcbw @dougbtv @smarterclayton @derekwaynecarr 
So this adds a `softAdmitHandler` to kubelet, to delay pod-network pods when there's no network plugin running, as vaguely discussed in http://post-office.corp.redhat.com/archives/aos-devel/2020-April/msg00115.html.

From the logs you can see that it seems to be working, but at a high level it has no real effect because the multus retry/timeout stuff was already successfully blocking all the warnings/errors/events anyway... so I don't know if this is interesting to pursue upstream?

smarterclayton:
In theory if we had to switch top level providers (instead of multus) it would be useful.  We minimized the multus unavailaibility mostly down to zero, but it could still fail in some cases.  If we had this, we might be able to tune those down to shorter increments so at least the kubelet would be able to see the failure.

Does this make the kubelet harder to understand?  Does it make multus easier to understand?

One school of thought is that you error as fast as possible and let the client smooth out the errors.  A downside is that that introduces latency and complexity to kubelet.  If multus ever needs to handle a smooth transition between two plugins, multus woul be the one who has to be aware of that, whereas if we wanted to transition from multus to something else or back, kubelet has to.  I kind of like that multus hides this for us - it means multus can deliver real value and hide complexity.

danwinship:
> Does this make the kubelet harder to understand?

I don't think so; it's a pretty small patch, using established scheduling mechanisms.

And even with the multus patch there, I guess I shouldn't have said "at a high level it has no real effect". I meant "it doesn't prevent any spurious errors that we aren't already preventing", but even with the multus patch, this patch has the side effect of providing more status: instead of just seeing that a pod is `phase: Pending`, you can see, eg:

      phase: Pending
      reason: NetworkUnavailable
      message: The pod network is currently not available; waiting for the network plugin to be deployed on this node

Likewise in the kubelet logs

danwinship:
> Does it make multus easier to understand?

and actually, yes, it could; we could totally get rid of the whole "don't create your CNI config file until you're ready to receive requests, and the delete it on exit" hack from multus, openshift-sdn, and ovn-kubernetes.

PR #24912, opened 2020-04-22
Bug 1731222: unskip template extended tests

gabemontero:


gabemontero:
8 of the 11 template extended tests passed in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24912/pull-ci-openshift-origin-master-e2e-gcp/7434

commencing triage

gabemontero:
So on `[sig-devex][Feature:Templates] templateinstance security tests should pass security tests [Suite:openshift/conformance/parallel]` the test timed out after 30 seconds waiting after a delete of the template instance  for all the finalization/gc to occur on the associated objects.

On the subsequent templateinstance dump, sure the ref object added as part of the test appears to a route in this case, is still there:

```
[]v1.TemplateInstanceObject{v1.TemplateInstanceObject{Ref:v1.ObjectReference{Kind:"Route", Namespace:"e2e-test-templates-4qgkn", Name:"route", UID:"15bef579-fe89-4304-84d2-b94f56f2e7d9", APIVersion:"route.openshift.io/v1", ResourceVersion:"", FieldPath:""}}}}}}
```
Looking at https://github.com/openshift/openshift-controller-manager/pull/6/files I don't see the route schema getting added, and it is still missing when looking at https://github.com/openshift/openshift-controller-manager/blob/master/cmd/openshift-controller-manager/main.go it is still missing

I'll still need to look at the ocm logs @adambkaplan @bparees but I think we may have at least 1 additional OCM change to complete the template changes needed once OCM was broken out.

gabemontero:
The `[sig-devex][Feature:Templates] templateinstance cross-namespace test should create and delete objects across namespaces [Suite:openshift/conformance/parallel]` failure was the same as the security one, delete did not finish in time, but with the exception

1) the associated object was a k8s `Secret` vs.  an openshift `Route`
2) the legacy scheme stuff should not matter for k8s `Secret`, right?  the finalizer currently uses" k8s.io/apimachinery/pkg/runtime/schema"

going to try and run analogous tests locally and try to zero in ... the OCM / apiserver / event logs from the test run were not helpful.

gabemontero:
I *think* the impersonation tests may be in error....or maybe now are in error based on how the underlying update in the api server works, wrt update?

They
1) get the template instance, which includes the requester
2) set additional fields in the template instance, but leave the requester set
3) update
4) get errors on the update like 

```
fail [github.com/openshift/origin/test/extended/templates/templateinstance_impersonation.go:307]: Unexpected error:
    <*errors.StatusError | 0xc001a643c0>: {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "TemplateInstance.template.openshift.io \"test\" is invalid: spec.requester.username: Forbidden: you do not have permission to set username",
            Reason: "Invalid",
            Details: {
                Name: "test",
                Group: "template.openshift.io",
                Kind: "TemplateInstance",
                UID: "",
                Causes: [
                    {
                        Type: "FieldValueForbidden",
                        Message: "Forbidden: you do not have permission to set username",
                        Field: "spec.requester.username",
                    },
                ],
                RetryAfterSeconds: 0,
            },
            Code: 422,
        },
    }
    TemplateInstance.template.openshift.io "test" is invalid: spec.requester.username: Forbidden: you do not have permission to set username
```

the update right before that tried to only update the conditions

I'm going to update the test to unset requester before this and see what happens.

gabemontero:
OK with the delete, running with trace locally, and cross referencing with the last test run, it isn't this legacy schema thing:

- in both local and CI, the template finalizer `template.openshift.io/finalizer` is removed and we are just left with the k8s `foregroundDeletion` in the finalizer array
- with trace running locally, I see both the delete of the associated obj and the templateinstance status update work successfully
- the deletion timestamp is set in both cases:  i.e. `DeletionTimestamp:(*v1.Time)(0xc0021e5860)`
- the deletion of the object(s) associated with the template (either routes or secrets) are successful.

Reading up now on k8s gc (https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#controlling-how-the-garbage-collector-deletes-dependents) ....

Also note, it *may* be that templates predate some of the GC stuff in the link I noted above enough that something it is doing is gumming up the works.  And since these tests have been disabled for a a bit, we've just missed whatever k8s rebase changed behavior such that templateinstance deletes are not fully functional.

Certainly we use finalizers in our operators.  They create objects.  And we can delete them no problem.

Moving from hacking on the test to hacking on the test and OCM.

@bparees @adambkaplan FYI

gabemontero:
ok I got a few of the delete templateinstance use cases working now with changes to the templateinstance_finalizer controller @adambkaplan @bparees 

I'll start sorting through the rest tomorrow

gabemontero:
discovery today @adambkaplan @bparees includes the fact that the `system:serviceaccount:kube-system:generic-garbage-collector` via its clusterrole `ystem:controller:generic-garbage-collector` does not pass the SAR check for `assign` to `templatesinstances` in the openshift-apiserver.

I've asked in #forum-apiserver about it

gabemontero:
> discovery today @adambkaplan @bparees includes the fact that the `system:serviceaccount:kube-system:generic-garbage-collector` via its clusterrole `system:controller:generic-garbage-collector` does not pass the SAR check for `assign` to `templatesinstances` in the openshift-apiserver at https://github.com/openshift/openshift-apiserver/blob/master/pkg/template/apiserver/registry/templateinstance/strategy.go#L148-L158.
> 
> I've asked in #forum-apiserver about it

yeah manually adding adding a clusterrole with assign for templates instances and adding it to generic garbage collection now seems to have delete passing in my local prototype for templateinstance_security.go

I'll now start pruning my other forays for debug / fixing and see if I can get a PR up either later today or Monday

gabemontero:
this PR will need https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/156 in order for the template suite to pass in e2e-gcp

also added a suite for templates to minimally assist local cluster testing

I'll remove that commit if that becomes controversial

/hold

gabemontero:
now this PR will pass template tests after https://github.com/openshift/openshift-apiserver/pull/102 merges

that PR replaces https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/156

gabemontero:
ok apiserver changed merge about 30 minutes ago ... let's give this a go

/retest

gabemontero:
IP flake on FIPS test ... retesting

/test e2e-aws-fips

gabemontero:
template e2e's passed in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24912/pull-ci-openshift-origin-master-e2e-gcp/7632 !!!

PR #24899, opened 2020-04-20
Remove skipped tests that OVN now supports

trozet:
Signed-off-by: Tim Rozet <trozet@redhat.com>

PR #24915, opened 2020-04-23
test/extended: un-skip OVNKubernetes tests that should now pass

dcbw:
@trozet @danwinship


dcbw:
have to make sure we run this against OVN tests which may not be defined yet for origin...
/hold

danwinship:
/lgtm
will need backports too

trozet:
@dcbw I already have a PR for this https://github.com/openshift/origin/pull/24899

PR #24916, opened 2020-04-23
Bug 1821352: vendor: Bump to pick up onsi/ginkgo panic reporting

smarterclayton:
A previous performance fix meant that some panic traces from tests
were not being reported fully, making it harder to debug.


soltysh:
@marun is working on updating origin to k8s 1.18.2 let's leave bumping ginko along that PR, pls. 
@marun make sure to pick the updated version of ginko from this PR.

smarterclayton:
You don't need to pick, the branch is already updated.  So it should come in.  How far away is that?

marun:
> You don't need to pick, the branch is already updated. So it should come in. How far away is that?

Home stretch: https://github.com/openshift/kubernetes/pull/122

marun:
Ginkgo was bumped to include this fix by #24936

PR #24917, opened 2020-04-24
add oauth-proxy login to kubeadmin tests

stlaz:
This mocks web-based login to oauth-server when redirected
from the oauth-proxy.

/assign @lilic


stlaz:
Credits for the oauth-server page login mocking code go to @mrogers950, I merely took it, cleaned it a bit and adapted it for this purpose

stlaz:
It's a bit hard to predict which repo can break this, but running at least this one test in authn-operator and oauth-proxy should be a must, I think it should be possible to add it in ci-operator config. I'll try to see about that.

lilic:
>, I think it should be possible to add it in ci-operator config. I'll try to see about that.

Sounds good to me, not a blocker for this right now, we can merge this as well to run for those repos just now. Thanks!

PR #24923, opened 2020-04-28
Bug 1826230: oc get --request-timeout: Timeout exceeded while reading body printed with -v=5

ingvagabund:
Expected "Timeout exceeded while reading body" error message is printed with -v=5.
Without the error message, both --watch oc get commands unexpectedly fail.


soltysh:
/retitle Bug 1826230: oc get --request-timeout: Timeout exceeded while reading body printed with -v=5

PR #24925, opened 2020-04-28
UPSTREAM: [24679]: BUG 1828923: Add tests for topology manager in single and multisocket configurations

fromanirh:
This PR adds the initial basic topology manager tests.

We cover all the testcases kubernetes has on its e2e tests, and all the known issues origin had reported against the topology manager.
The reason for the overlap is because the e2e kubernetes node tests do not run against a full cluster, but on a reduced environment (kubelet+apiserver only, for example).
Furthermore, we care about some test scenarios that aren't very relevant for kubernetes. The path forward is of course to move the tests upstream were feasible, but this testsuite is meant to stay.

The tests by default skip if the cluster preconditions aren't met, so if topology manager and all its dependencies are not enabled, or configured with a policy which is not `single-numa-node`. This is because this policy is the only one which allow to unambiguosly check the correctness of the resource alignment.

The tests are configurable with few environement variables, documented in the README. This is unfortunate, but needed by the complexity of the scenarios the test need to check, and which we cannot reliably autodetect.

The tests require an environment with a SRIOV device enabled to properly exercise the alignment. Should the environment not have such a device, the test will skip.
This is because SRIOV and GPU devices are the only devices supported by the devicemanager at this time, and SRIOV device are the easiest and the cheapest to obtain.

https://bugzilla.redhat.com/show_bug.cgi?id=1826807


fromanirh:
the test failures look suspicious. Retesting to get fresh logs.

fromanirh:
uhm, the `e2e-aws-disruptive` lane failed twice in a row because of ` fail [runtime/panic.go:679]: Test Panicked: runtime error: index out of range [1] with length 1 ` - cpuld please someone (@rphillips ? @mrunalp ?)  shed some light here ?

fromanirh:
`aws-disruptive` seems flaky

fromanirh:
I was mistaken, the `release-4.5` branch is still syncing with master, so we don't need this PR.

PR #24926, opened 2020-04-28
Bug 1823406: Upstream: 89160: Remove potentially unhealthy symlink only for dead containers

tedyu:
As the discussion over #52172 showed, there is race condition between the container log rotation and the kubelet GC which may result in the loss of symlink.

Here is how container log rotation works (see containerLogManager#rotateLatestLog):

* rename current log to rotated log file whose filename contains current timestamp (fmt.Sprintf("%s.%s", log, timestamp))
* reopen the container log
* if #2 fails, rename rotated log file back to container log

There is small but indeterministic amount of time during which log file doesn't exist (between steps #1 and #2, between #1 and #3). Hence the symlink may be deemed unhealthy during that period.

This PR resorts to runtimeService.ContainerStatus() to check whether the container corresponding to the potentially unhealthy symlink is alive or not. The symlink would only be removed for dead containers.


rphillips:
Upstream [ref](https://github.com/kubernetes/kubernetes/pull/89160/files)

rphillips:
/retitle Bug 1823406: Remove potentially unhealthy symlink only for dead containers

rphillips:
/retitle Bug 1823406: Upstream: 89160: Remove potentially unhealthy symlink only for dead containers

PR #24928, opened 2020-04-28
test/e2e/upgrade/service: Use 10m timeout for TestReachableHTTPWithMinSuccessCount

wking:
Instead of assuming that the AWS default applies to all providers, as we've done since we carried the upstream code in 9732d0124d (#24471).  This catches our copied-and-tweaked code up with upstream logic like that has switched on provider since kubernetes/kubernetes@46b89464fd (kubernetes/kubernetes#20959).

### DIFF
--- a/test/e2e/upgrade/service/service.go
+++ b/test/e2e/upgrade/service/service.go
@@ -96,7 +96,7 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
 
 	// Hit it once before considering ourselves ready
 	ginkgo.By("hitting pods through the service's LoadBalancer")
-	timeout := service.LoadBalancerLagTimeoutAWS
+	timeout := 10 * time.Minute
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade/service/service.go
```diff
@@ -96,7 +96,10 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
 
 	// Hit it once before considering ourselves ready
 	ginkgo.By("hitting pods through the service's LoadBalancer")
-	timeout := service.LoadBalancerLagTimeoutAWS
+	timeout := service.LoadBalancerLagTimeoutDefault
```
smarterclayton, line 99: i'd rather just put 10m right here.  there is no scenario where i care about what upstream's value is.

wking: > i'd rather just put 10m right here...

Done with 0fd8f40010 -> 65bd1de636.

PR #24930, opened 2020-04-29
Bug 1761043: provides a temporal fix to improve CRD publishing e2e tests in an HA setup

p0lyn0mial:

PR #24931, opened 2020-04-29
Bug 1761043: provides a temporal fix to improve CRD publishing e2e tests in an HA setup

p0lyn0mial:


soltysh:
This is not needed, we haven't cut 4.5, yet. Master is still automatically merged to 4.5, so https://github.com/openshift/origin/pull/24930 is sufficient. 
/close

PR #24932, opened 2020-04-29
adds kube publishing rules for origin-4.5-kubernetes-1.18.2

p0lyn0mial:


publishing-kube-rules.yaml
```diff
@@ -7,6 +7,145 @@ recursive-delete-patterns:
 - BUILD.bazel
 - "*/BUILD.bazel"
 rules:
+# origin-4.5-kubernetes-1.18.2
```
marun, line 10: Is it desirable to add new rules vs updating the rules for 1.18.0-rc.1?

sttts: We remove the old eventually.

sttts: ... when origin rebase merged.

PR #24944, opened 2020-05-04
Bug 1749246: UPSTREAM: Fix TestContext references in e2e skippers

damemi:
This test has recently (and somewhat suspiciously) disappeared from the testgrid top flakes. Upon further investigation, it appears that it is always skipping due to failing [this node count check](https://github.com/kubernetes/kubernetes/blob/9b44fa4a07b14b3dbc644f80e087d516677632f0/test/e2e/scheduling/priorities.go#L161) (even when run locally, which is very odd).

~~This PR removes the node check, but it would be better if we could investigate why this check is constantly skipping (did we change how we set the node count?) and if other tests that use this check are also skipping.~~

**Update:** It turns out that the cause of this is because the e2e skipper package was broken and referred to `TestContext `as an uninitialized variable local to its own package rather than `framework.TestContext` which has the node values properly set. Fixed by upstream kubernetes/kubernetes#90707 and picked here. 

This PR no longer removes the check mentioned above and now carries the changes from upstream

Some runs where this has happened:
* https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-azure-serial-4.5/1021
* https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-gcp-serial-4.5/981
* https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-metal-serial-4.5/986
* https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-openstack-serial-4.5/673
* https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-vsphere-upi-serial-4.5/974

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/framework/skipper/skipper.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/framework/skipper/skipper.go
@@ -41,9 +41,6 @@ import (
 	e2essh "k8s.io/kubernetes/test/e2e/framework/ssh"
 )
 
-// TestContext should be used by all tests to access common context data.
-var TestContext framework.TestContextType
-
 func skipInternalf(caller int, format string, args ...interface{}) {
 	msg := fmt.Sprintf(format, args...)
 	framework.Logf(msg)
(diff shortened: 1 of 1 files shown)


damemi:
Definitely want to /hold this to see how it does in CI, because removing this skip might just be re-introducing a flaky test

smarterclayton:
How can you test pod affinity unless you have 2 nodes?

smarterclayton:
Node count is a setting filled in automatically by the ci infra and passed to the openshift-tests process - if that is getting set to 0, then this will be skipped.  You need to ensure that this value is filled.

/hold

damemi:
@smarterclayton that's what we're trying to figure out, why is this value being 0 in this test. From a quick search it doesn't look like other tests with similar checks are being skipped, but I don't know the details of how this value is set to determine if it's being properly passed.

Obviously this test requires 2 nodes so I don't intend to merge this, but I wanted to see if it would even pass without it to start a debugging discussion

damemi:
To update, talked to @soltysh and @tnozicka this morning about this and they suggested checking what's returned from [`GetReadySchedulableNodes`](https://github.com/kubernetes/kubernetes/blob/3b5ea194ea757fcbccf27cec493486812454f9d9/test/e2e/framework/node/resource.go#L303-L315), which I logged in the [e2e setup suite function](https://github.com/kubernetes/kubernetes/blob/3b5ea194ea757fcbccf27cec493486812454f9d9/test/e2e/e2e.go#L265-L268).

There, it returns 3. Yet our check still fails for `0 < 2`.

It doesn't log anything in the above `framework.TestContext.CloudConfig.NumNodes == framework.DefaultNumNodes` block, meaning `framework.TestContext.CloudConfig.NumNodes` is being set to `3` somewhere. And possibly overridden during this specific scheduler e2e? I don't know

damemi:
It turns out that the cause of this is because the e2e skipper package was broken and referred to `TestContext` as an uninitialized variable local to its own package rather than `framework.TestContext` which has the node values properly set. Fixed by upstream https://github.com/kubernetes/kubernetes/pull/90707 and picked here

ingvagabund:
/hold

feel free to unhold. Just holding the PR so you can validate the number of tests ran increased

damemi:
/hold cancel
looks like the job is no longer skipped from a brief check

PR #24936, opened 2020-04-30
Rebase 1.18.2

marun:
/cc @sttts @p0lyn0mial

PR #24947, opened 2020-05-04
Bug 1731263: Balance preemption e2e nodes

damemi:
These preemption e2es attempt to create filler "victim" pods that use 60% of a node's available resources, and assumes that with all nodes full, the low priority pod will get evicted.

However, if the nodes have very different usages (say one is at 500/1500 cpu and another is at 1000/1500), then the request of 60% of the low-priority-pod's-node's usage may not fit on one node, but will fit on another and the test fails.

This PR creates balanced pods in these e2es to ensure a level field for the preemption tests.

Upstream: https://github.com/kubernetes/kubernetes/pull/90740


damemi:
So, re-enabled and with these fixes the tests fail with:
> Unschedulable 0/6 nodes are available: 2 Insufficient memory, 3 Insufficient cpu, 3 node(s) didn't match node selector.}]
But they pass locally for me, because this is exactly the case I would expect preemption to happen it means that for some reason it must not be.

Something else I see is that the calculated usage for the balanced pods differs from the usage we calculate when creating the victim pods

soltysh:
/hold 
for k8s bump, feel to drop when that lands.

damemi:
https://github.com/openshift/origin/pull/24936 has merged the 1.18.2 bump
/hold cancel

PR #24948, opened 2020-05-05
Bug 1829405: test/extended/operators/operators: Rework "start all core operators"

wking:
Drop namespace from output strings, because [`ClusterOperator` is cluster-scoped][1].

Shift the `available` variable down to the block of code that actually consumes it, and convert it from a map to a counter, because we don't care about displaying available operators.

Rename `unavailable` to the more generic `working` (to match the existing "still doing work" messages).  Working operator may be available but progressing, etc., so `unavailable` was too specific.

Remove the `break` from the wait-loop's `ClusterOperator` iteration.  The previous logic would only list the first unavailable operator.  With this commit we now list each unavailable operator while we wait.

Reformat the tab-writer block to print each operator and its worst condition.  The previous format had columns for progressing and available, but none for degraded status.  It also had a single message column, making it unclear which condition's message was being displayed.  The new format displays the full type, status, reason, and message of the worst condition for each operator.

The new code also includes the worst condition (or, if all existing conditions for an operator were expected, any missing condition types) in the failing error message.  This should be more actionable with less digging than the current messages, which just list the unsettled operator names.

[1]: https://github.com/openshift/api/blob/b98a784d8e6dc93c416b40cdf54bf3102f2b61d2/config/v1/types_cluster_operator.go#L9
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1829405#c1

### DIFF
--- a/test/extended/operators/operators.go
+++ b/test/extended/operators/operators.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -80,7 +81,6 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should", func() {
 		}
 
 		// gate on all clusteroperators being ready
-		available := make(map[string]struct{})
 		g.By(fmt.Sprintf("waiting for all cluster operators to be stable at the same time"))
 		coc := dc.Resource(schema.GroupVersionResource{Group: "config.openshift.io", Resource: "clusteroperators", Version: "v1"})
 		lastErr = nil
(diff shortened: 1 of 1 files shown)


wking:
[images]:

```
2020/05/05 17:52:55 error: unable to signal to artifacts container to terminate in pod release-latest, triggering deletion: could not run remote command: unable to upgrade connection: container not found ("artifacts")
2020/05/05 17:52:55 error: unable to retrieve artifacts from pod release-latest: could not read gzipped artifacts: unable to upgrade connection: container not found ("artifacts")
```

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24948/pull-ci-openshift-origin-master-images/12645

wking:
I'd flipped the expected sense on Degraded (fixed with a7b435d354 -> fe2a71752e), but it [shows][1] the output formatting:

```
fail [github.com/openshift/origin/test/extended/operators/operators.go:156]: May  5 22:14:53.039: Some cluster operators never became available authentication (Degraded=False AsExpected: <nil>), cloud-credential (Degraded=False NoCredentialsFailing: No credentials requests reporting errors.), ...
```

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/24948/pull-ci-openshift-origin-master-e2e-gcp/7600

wking:
Remaining CI failures look unrelated.  @bparees, want to take a look and approve if you like it?

/assign @bparees

test/extended/operators/operators.go
```diff
@@ -101,68 +101,65 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should", func() {
...
-				}
-				if condition(co, "Failing").Get("status").String() != "False" {
-					ns := co.Get("metadata.namespace").String()
-					name := co.Get("metadata.name").String()
-					unavailableNames = append(unavailableNames, fmt.Sprintf("%s/%s", ns, name))
-					unavailable = append(unavailable, co)
-					break
+				conditions, missingTypes := surprisingConditions(co)
+				if len(conditions) > 0 || len(missingTypes) > 0 {
+					working = append(working, co.Get("metadata.name").String())
```
bparees, line 108:
```suggestion
					unready = append(unready, co.Get("metadata.name").String())
```
working is a confusing name here.  i first read it as "these are the operators that are functional" not "these are the operators that are still churning"

test/extended/operators/operators.go
```diff
@@ -101,68 +101,65 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should", func() {
...
-					unavailable = append(unavailable, co)
-					break
-				}
-				if condition(co, "Failing").Get("status").String() != "False" {
-					ns := co.Get("metadata.namespace").String()
-					name := co.Get("metadata.name").String()
-					unavailableNames = append(unavailableNames, fmt.Sprintf("%s/%s", ns, name))
-					unavailable = append(unavailable, co)
-					break
+				conditions, missingTypes := surprisingConditions(co)
```
bparees, line 106:
```suggestion
				badConditions, missingConditions := surprisingConditions(co)
```

test/extended/operators/operators.go
```diff
@@ -101,68 +101,65 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should", func() {
...
 		buf := &bytes.Buffer{}
 		w := tabwriter.NewWriter(buf, 0, 4, 1, ' ', 0)
-		fmt.Fprintf(w, "NAMESPACE\tNAME\tPROGRESSING\tAVAILABLE\tVERSION\tMESSAGE\n")
+		fmt.Fprintf(w, "NAME\tTYPE\tSTATUS\tREASON\tMESSAGE\n")
 		for _, co := range lastCOs {
-			ns := co.Get("metadata.namespace").String()
 			name := co.Get("metadata.name").String()
-			if condition(co, "Available").Get("status").String() != "True" {
-				unavailable = append(unavailable, fmt.Sprintf("%s/%s", ns, name))
+			conditions, missingTypes := surprisingConditions(co)
```
bparees, line 127:
```suggestion
			badConditions, missingConditions := surprisingConditions(co)
```

test/extended/operators/operators.go
```diff
@@ -101,68 +101,65 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should", func() {
 				return false, nil
 			}
 
-			var unavailable []objx.Map
-			var unavailableNames []string
+			var working []string
```
bparees, line 104:
```suggestion
			var unready []string
```

wking:
I pushed fe2a717 -> 5819963 to fig working -> unready and use `badConditions`.  There's an example of the "failure" output  linked from [this earlier comment][1].

[1]: https://github.com/openshift/origin/pull/24948#issuecomment-624374684

test/extended/operators/operators.go
wking: Fixed in fe2a717 -> 5819963.

test/extended/operators/operators.go
wking: Fixed in fe2a717 -> 5819963.

test/extended/operators/operators.go
wking: Fixed in fe2a717 -> 5819963.

test/extended/operators/operators.go
wking: Fixed in fe2a717 -> 5819963.

test/extended/operators/operators.go
```diff
@@ -101,68 +101,65 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should", func() {
...
-			)
 		}
 		w.Flush()
 		e2e.Logf("ClusterOperators:\n%s", buf.String())
 
-		if len(unavailable) > 0 {
-			e2e.Failf("Some cluster operators never became available %s", strings.Join(unavailable, ", "))
+		if len(unready) > 0 {
+			sort.Strings(unready)
+			e2e.Failf("Some cluster operators never became ready %s", strings.Join(unready, ", "))
```
bparees, line 159:
```suggestion
			e2e.Failf("Some cluster operators never became ready: %s", strings.Join(unready, ", "))
```

test/extended/operators/operators.go
bparees: i see the change elsewhere but not on this line?

bparees:
>  There's an example of the "failure" output linked from this earlier comment.

thanks, that looks nice

test/extended/operators/operators.go
wking: Fixed with 5bbe680d26 -> d7a7ff790f.

test/extended/operators/operators.go
wking: Fixed with 5bbe680d26 -> d7a7ff790f.

PR #24954, opened 2020-05-06
Bug 1832641: UPSTREAM: 90823: Service load balancers should include unschedulable nodes

Miciah:
In 16443 a set of changes were made to enable service load balancers
for Kube 1.1 and GCP to exclude unschedulable nodes from load balancer
pools. However, Unschedulable has nothing to do with whether nodes
should host service load balancer pods - the act of preventing new
pods from landing on a node has no impact on existing workloads.

In general nodes are not special, and a user who wants to exclude
otherwise healthy nodes can use the beta service load balancer
label to exclude the node from the LB pool.

This commit removes the check for unschedulable nodes from the
LB pool.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/controller/service/controller.go
+++ b/vendor/k8s.io/kubernetes/pkg/controller/service/controller.go
@@ -604,12 +604,6 @@ func nodeSlicesEqualForLB(x, y []*v1.Node) bool {
 
 func getNodeConditionPredicate() NodeConditionPredicate {
 	return func(node *v1.Node) bool {
-		// We add the master to the node list, but its unschedulable.  So we use this to filter
-		// the master.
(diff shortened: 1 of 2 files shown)


Miciah:
/retitle Bug 1832641: UPSTREAM: 90823: Service load balancers should include unschedulable nodes

Miciah:
/retest
/assign @smarterclayton 

Clayton, do you want to go ahead and `/approve` this to ship it in 4.5?

I saw you rebased your upstream PR after I made this pick PR, but only to rebase it; the critical change to the logic in `getNodeConditionPredicate` is still valid with this pick PR.

smarterclayton:
/approve
/lgtm

This makes no sense for openshift.  At worst, we carry.

PR #24955, opened 2020-05-07
Pick Topology Manager from 4.5 (1.18.2) to 4.4 (1.17.1)

vpickard:
Topology Manager went to Beta in Kubernetes 1.18 release and is included in OCP 4.5 in this rebase:
https://github.com/openshift/kubernetes/pull/118 

This PR picks the Topology Manager component  to OCP 4.4 (1.17.1).

PR #24959, opened 2020-05-08
[release-4.4] e2e: add extended tests for topology manager

fromanirh:
The tests will skip if topology manager is not enabled
on the cluster, with single-numa node policy.
If it is enabled, will check the pod resources are usable
and properly aligned on NUMA node.

This is a superset of the e2e tests found on kubernetes.
A key difference is that these tests run against a full cluster,
while the e2e node tests runs in a reduced environment.

release-4.4 backport note
- ported back to the 1.17 client API (no `context` in calls)


fromanirh:
This PR is needed to test https://github.com/openshift/origin/pull/24955

fromanirh:
PR #24955 is not happening, so we don't need this backport anymore.

PR #24960, opened 2020-05-08
[wip] stop honoring the kube-apiserver struct config and use flag

deads2k:
cannot merge until https://github.com/openshift/cluster-kube-apiserver-operator/pull/850 lands

But this shows where the values used to come from.


deads2k:
replaced by https://github.com/openshift/kubernetes/pull/439

PR #24963, opened 2020-05-09
IPFamily fixes from upstream

danwinship:
The `IPv6DualStack` feature gate is currently unusable in OCP because of bugs in `Service.Spec.IPFamily` validation. This is now fixed upstream and it would be useful for dual-stack testing to get these fixes into origin master now rather than waiting for the 1.19 rebase

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/apis/core/validation/conditional_validation.go
+++ b/vendor/k8s.io/kubernetes/pkg/apis/core/validation/conditional_validation.go
@@ -17,14 +17,21 @@ limitations under the License.
 package validation
 
 import (
+	"fmt"
+	"net"
+	"strings"
+
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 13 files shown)


danwinship:
/retest
/assign @knobunc 
This is just pulling some 1.19 bugfixes into origin master ahead of the main rebase. Dual stack testing is not possible without these fixes.

aojea:
/retest
unrelated error
/lgtm

aojea:
@danwinship maybe we'll need this https://github.com/kubernetes/kubernetes/pull/89914/commits/e71f84b1c4581535fd7fca1bf55f0b69cdf38304 too?

danwinship:
We _eventually_ need all the fixes, and we'll get them when we rebase to 1.19. The only thing actively blocking us from doing initial testing is the validation bug. (I included the other fix just because the validation fix didn't backport cleanly without it, not because I actually care about the fix.)

PR #24967, opened 2020-05-11
UPSTREAM: <carry>: eliminate unnecessary closure in openshift configu…

deads2k:
…ration wiring

This eliminates indirection through a global and instead directly wires the globals.  It's still ugly and order dependent, but one layer of indirection is gone.


vendor/k8s.io/kubernetes/cmd/kube-apiserver/app/server.go
```diff
@@ -117,17 +117,15 @@ cluster's shared state through which all other components interact.`,
 			verflag.PrintAndExitIfRequested()
 
 			if len(s.OpenShiftConfig) > 0 {
+				// if we are running openshift, we modify the admission chain defaults accordingly
+				admissionenablement.InstallOpenShiftAdmissionPlugins(s)
```
stlaz, line 121: no action needed, just curious: the patching function assignment was noop here, just a side effect elsewhere, so is there anything to this admission plugins install move?

deads2k: > no action needed, just curious: the patching function assignment was noop here, just a side effect elsewhere, so is there anything to this admission plugins install move?

doing it early made it easier for me to read and reason about. nothing special.

sttts:
Sgtm
/approve
/assign @marun

marun:
/lgtm

@deads2k I'm assuming this can be squashed on rebase?

PR #24968, opened 2020-05-12
Bug 1827863: [Feature:Platform][Smoke] Managed cluster should ensure pods use downstream images from our release image with proper ImagePullPolicy [Suite:openshift/conformance/parallel] timing out 

Danil-Grigorev:
Increased default timeout on image-registry up to 20 min

### DIFF
--- a/cmd/openshift-tests/e2e.go
+++ b/cmd/openshift-tests/e2e.go
@@ -96,6 +96,7 @@ var staticSuites = []*ginkgo.TestSuite{
 		Matches: func(name string) bool {
 			return strings.Contains(name, "[sig-imageregistry]") && !strings.Contains(name, "[Local]")
 		},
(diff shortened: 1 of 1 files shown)


dmage:
/hold

1. timeout happens on 4.2, not master
2. even in master this test is not tagged `sig-imageregistry`, it's `sig-arch`

cmd/openshift-tests/e2e.go
```diff
@@ -96,6 +96,7 @@ var staticSuites = []*ginkgo.TestSuite{
 		Matches: func(name string) bool {
 			return strings.Contains(name, "[sig-imageregistry]") && !strings.Contains(name, "[Local]")
 		},
+		TestTimeout: 20 * time.Minute,
```
enxebre, line 99: I realised this test is not tagged as [registry] so this wont' take effect.

PR #24971, opened 2020-05-12
Bug 1763340: ingress: targetPortForService: Use port name only

Miciah:
When translating the port specified by an Ingress to the target port for a Route, use the port name rather than the number, or omit the target port if it has no name.

An Ingress resource specifies a Service and a port on that Service whereas a Route resource specifies a Service and a target port on the corresponding Endpoints resource.  The ports on the Endpoints resource and the ports on the Service resources have matching names but may have different numbers. Therefore, when mapping an Ingress and Service to a Route and a target port for the Endpoints resource, we cannot use port numbers; we must use names or omit the target port entirely.

It is safe to omit the target port if it does not have a name because a Service port must have a name if the Service has more than one port; therefore if a port is nameless, it must be the only port; and omitting the target port in the Route spec means that every port (or in this case, the only port) on the Endpoints resource is targeted.

* `pkg/route/ingress/ingress.go` (`targetPortForService`): Add an error return value.  Return an error if the Ingress references a Service or port that cannot be found.  Otherwise, return the Service port's name as the target port, or return nil if the port has no name.
(`newRouteForIngress`, `routeMatchesIngress`): Update to check the error value from `targetPortForService` and to handle a nil target port (which means the Route should have a null port specification).
* `pkg/route/ingress/ingress_test.go` (`TestController_stabilizeAfterCreate`):
Add test case with a nameless port.
(`TestController_sync`): Add a test case where the Ingress references a Service that does not exist.  Add a test case where the Ingress specifies a port by number.  Update test cases to expect port names in Route port specifications.  Update the failure messages for unexpected actions to print the actions.

Co-authored-by: derkoe <christian.koeberl@gmail.com>

### DIFF
--- a/pkg/route/controller/ingress/ingress.go
+++ b/pkg/route/controller/ingress/ingress.go
@@ -9,7 +9,7 @@ import (
 
 	"k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


Miciah:
I have opened https://bugzilla.redhat.com/show_bug.cgi?id=1835480 for the CI failure.

PR #24980, opened 2020-05-14
tests: Skip "Services should be rejected when no endpoints exist"

cgwalters:
We debugged this in the kernel, but we really can't block shipping
RHEL 8.2 on this.
https://bugzilla.redhat.com/show_bug.cgi?id=1781575

I'll set up a reminder to remove this in a bit.


cgwalters:
My first time touching this bit of code, didn't test that this successfully skips yet; opening for debate.

smarterclayton:
/hold

I don't think we should be on rhel 8.2 until this is fixed.  Do we have an ETA?

smarterclayton:
I will keep the hold on until we can articulate how this would impact normal kube behavior for a normal customer.  We see it in a number of tests.  Those are completely reasonable tests that simulate applications that start up and begin connecting to each other.  If we have passed a threshold where the rate of change (in terms of new / old services) saturates the rate limiting function, then the rate limiting function is the problem, not the use of the platform.

The closest framing I've seen on this issue (based on my limited understanding of all of the pieces) is "if you start two apps, and app 1 tries to connect to app 2 before it is ready, it is not blackholing connections for 3 minutes after app 2 is ready" seems like the scenario i would say is absolutely fundamental to a well behaved system (that is a severe issue).  If my assessment is incorrect and the system does not behave this way, please let me know.

If the first is true, the second is "in a zone reset, where 20% of machines reboot, and 20-30% of pods are restarted elsewhere, hundreds of services on a node may suddenly become unavailable - if ICMP rate limiting turns a fast retry of REFUSED into BLACKHOLE, then at the worst possible time we would be changing behavior of the system in a way that makes the diagnosis of problems worse, when we should be making it better".  Again, if I am missing a description of the problem, that's fine, but I need to know what the truth is.

danwinship:
> The closest framing I've seen on this issue (based on my limited understanding of all of the pieces) is "if you start two apps, and app 1 tries to connect to app 2 before it is ready, it blackholes connections for 3 minutes after app 2 is ready" seems like the scenario i would say is absolutely fundamental to a well behaved system not happening (i.e. that is a severe issue). If my assessment is incorrect and the system does not behave this way, please let me know.

The system does not behave that way. New connection attempts to working services will always succeed immediately, regardless of previous failed attempts. Pending connection attempts that started before the service came up might be a bit slower to complete because of exponential backoff on retries, but that would not be on the order of minutes.

smarterclayton:
> The system does not behave that way. New connection attempts to working services will always succeed immediately, regardless of previous failed attempts.

Not useful for the case I mentioned (service A is making a DB connection in a loop, so it's worst possible case).

> Pending connection attempts that started before the service came up might be a bit slower to complete because of exponential backoff on retries, but that would not be on the order of minutes.

What's the order?  Lower bound, upper bound, p95?

danwinship:
> What's the order? Lower bound, upper bound, p95?

Hard to say for sure since it depends on whether the code has any application/library-level timeouts.

If I do a raw curl, there's about a minute between the last time it tries to connect and when it finally times out.

If I try using `oc`, there's a 30-second timeout coming from somewhere (so it does TRY (wait 1s) TRY (wait 2s) TRY (wait 4s) TRY (wait 8s) TRY (wait not-quite-15s) FAIL) but I think that timeout is oc/kubectl-specific.

If I try blocking apiserver responses into a randomly-selected pod and then watching retries, it seems like any given request times out after around 15 seconds (and so the longest gap between attempt and failure is about 8s).

cgwalters:
We have 8.2 in 4.6 now.

PR #24981, opened 2020-05-14
Bug 1814282: Storage e2es leaving namespaces/pods around

gnufied:
Backport of:

1. https://github.com/kubernetes/kubernetes/pull/90773
2. https://github.com/kubernetes/kubernetes/pull/91221
3. https://github.com/kubernetes/kubernetes/pull/91689

This fixes issues that were caused by CSI driver being deleted before pod that was using could be deleted and hence such pods were left in "terminating" state.  This PR fixes that bug by causing CSI driver to be installed in different namespace. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1814282

There are still some issues around snapshot e2es leaving namespaces in terminating state. But those are not e2e bugs but are actual bugs in snapshot controller etc - this is being tracked https://bugzilla.redhat.com/show_bug.cgi?id=1848633

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/storage/drivers/csi.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/storage/drivers/csi.go
@@ -208,7 +208,6 @@ func (h *hostpathCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.Per
 	// Cleanup CSI driver and namespaces. This function needs to be idempotent and can be
 	// concurrently called from defer (or AfterEach) and AfterSuite action hooks.
 	cleanupFunc := func() {
-		framework.RemoveCleanupAction(h.cleanupHandle)
 		ginkgo.By(fmt.Sprintf("deleting the test namespace: %s", ns1))
 		// Delete the primary namespace but its okay to fail here because this namespace will
 		// also be deleted by framework.Aftereach hook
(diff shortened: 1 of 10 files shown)


gnufied:
cc @openshift/sig-storage

jsafrane:
lgtm-ish, but please squash into one and use `UPSTREAM: xxx ` commit message.

vendor/k8s.io/kubernetes/test/e2e/storage/drivers/csi.go
```diff
@@ -220,6 +219,9 @@ func (h *hostpathCSIDriver) PrepareTest(f *framework.Framework) (*testsuites.Per
 
 		ginkgo.By(fmt.Sprintf("deleting the driver namespace: %s", ns2))
 		tryFunc(deleteNamespaceFunc(f.ClientSet, ns2, framework.DefaultNamespaceDeletionTimeout))
+		// we need to remove the cleanupHandle at last so as if AfterEach and AfterSuite races with
+		// each other and AfterEach(or defer) run gets killed then it still cleanups.
+		framework.RemoveCleanupAction(h.cleanupHandle)
```
jsafrane, line 224: This change in not in the referenced UPSTREAM PR. Upstream master still has `RemoveCleanupAction` as the first call in `cleanupFunc`.

gnufied: Yes. I dropped this in upstream PR. I intend to bring it back in a new PR in upstream. This is still needed.

jsafrane: The please create a separate `UPSTREAM: <carry>` commit, so it does not get lost on the next rebase. fixing upstream first is even better.

gnufied: I updated this PR to use right commit from upstream.

gnufied:
I am planning to fix issues around snapshot in a different Bug - https://bugzilla.redhat.com/show_bug.cgi?id=1848633

This should fix issues with most of the other tests that depend on driver to be around for volume cleanup to happen. I think we should still merge this PR.

gnufied:
Failures look unrelated to this PR.

/retest

PR #24982, opened 2020-05-14
Bug 1835931: Fix panic in test code

ironcladlou:
Fix out of bounds access which causes intermittent test panics.


hexfusion:
we have more work here to fix tests but let's not panic.

/lgtm

PR #24988, opened 2020-05-15
Bug 1753706: sync node's CNI IP store to clean stale sandbox IPs

alexanderConstantinescu:
Why is this needed? 

In the unlikely event that we have severe and unknown cluster / node failures, the SDN risks not removing stale IP allocations in `/var/lib/cni/networks/openshift-sdn` and pile up garbage. Once the garbage becomes too big, we start failing with false IP saturation claims:

```
failed to run CNI IPAM ADD: failed to allocate for range 0: no IP addresses available in range set: 100.65.2.1-100.65.3.254
```

What to do? Garbage collect once per day, with a sync. It's not needed to do too often, so set it to once per day.  

For more background, see: https://bugzilla.redhat.com/show_bug.cgi?id=1753706#c42

/assign @danwinship

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -49,6 +49,7 @@ import (
 const (
 	openshiftCNIFile = "80-openshift-network.conf"
 	hostLocalDataDir = "/var/lib/cni/networks"
+	cniIPStorePath   = hostLocalDataDir + "/openshift-sdn/"
 )
 
 type osdnPolicy interface {
@@ -372,6 +373,7 @@ func (node *OsdnNode) Start() error {
 	}
 
 	go kwait.Forever(node.policy.SyncVNIDRules, time.Hour)
+	go kwait.Forever(node.syncCNIDir, time.Hour*24)
 	go kwait.Forever(func() {
 		gatherPeriodicMetrics(node.oc.ovs)
 	}, time.Minute*2)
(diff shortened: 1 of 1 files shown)


danwinship:
We used to have code to do this in openshift-sdn (in like the 3.6 timeframe? Maybe even earlier?) to work around a specific kubelet bug, but we removed it after that, because if something is happening that causes IPAM files to be left around, then it ought to be causing other failures too and GC'ing IPAM junk just papers over the larger problems.

If the problem is specifically related to reboots, I think we ought to be dropping all IPAM state after a reboot, and if we aren't, that may be the whole bug. 

@dcbw may remember more

alexanderConstantinescu:
> We used to have code to do this in openshift-sdn (in like the 3.6 timeframe? Maybe even earlier?) to work around a specific kubelet bug, but we removed it after that, because if something is happening that causes IPAM files to be left around, then it ought to be causing other failures too and GC'ing IPAM junk just papers over the larger problems.
> 
> If the problem is specifically related to reboots, I think we ought to be dropping all IPAM state after a reboot, and if we aren't, that may be the whole bug.
> 
> @dcbw may remember more

I agree, but as you can see from my cited BZ comment: some nodes do not even log the very mention of the Sandbox ID in their journal (which stretches one year back), meaning they are experience _very_ weird issues. And I am not sure we'll be able to find out why.

alexanderConstantinescu:
@dcbw can we get your input on this?

dcbw:
It would probably be worthwhile to add clearing the CNI IPAM dir for the SDN plugin to roles/openshift_node/templates/cleanup-cni.j2 in openshift-ansible.

dcbw:
The only gap is if (a) the node forgets that the container exists, eg kubelet/docker/crio backing store get cleared and the container doesn't exist anymore, and (b) the container is removed from the Kube API. In this specific case, then kubelet wouldn't call GC and the IPAM wouldn't get released. But that would usually only happen in node reboot or exceptional cases, and the IPAM clear on reboot should address that.

koep:
@dcbw @alexanderConstantinescu from what I understand, this change looks reasonable. can you advise on what the next steps are to proceed?

alexanderConstantinescu:
Closing this as this change would hide other failing issues with OCP 3.11

PR #24993, opened 2020-05-15
Bug 1833387: test/extended/operators/operators: Drop cvoWait and operatorWait

wking:
We've had these since 5ff7aed012 (#21459), but conditions like `ClusterVersion` `Progressing=True` and `ClusterOperator` `Degraded=True` are serious enough that we don't want to see them post-install.  Make it fatal to see those surprises at all, instead of giving operators a pass during an arbitrary post-install recovery window.

### DIFF
--- a/test/extended/operators/operators.go
+++ b/test/extended/operators/operators.go
@@ -1,13 +1,11 @@
 package operators
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"os"
 	"sort"
 	"strings"
-	"text/tabwriter"
 	"time"
 
 	g "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


wking:
[e2e-gcp][1]:

```
fail [github.com/openshift/origin/test/extended/operators/operators.go:94]: May 16 01:01:42.400: Some cluster operators are not ready: kube-apiserver (Progressing=True NodeInstaller: NodeInstallerProgressing: 1 nodes are at revision 2; 1 nodes are at revision 6; 1 nodes are at revision 8), kube-controller-manager (Progressing=True NodeInstaller: NodeInstallerProgressing: 2 nodes are at revision 7; 1 nodes are at revision 8), kube-scheduler (Progressing=True NodeInstaller: NodeInstallerProgressing: 1 nodes are at revision 7; 2 nodes are at revision 10)
```

So with this and #24992, we should be all set.

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/24993/pull-ci-openshift-origin-master-e2e-gcp/7714

wking:
/retest

Now that #24992 has landed.

PR #24995, opened 2020-05-18
[WIP] (test only) artificial delay adder filter

tkashem:
kube-apiserver filter that adds artificial delay to request(s) from a certain user. It will be used for testing purposes only.


tkashem:
for testing only.

/hold

PR #24998, opened 2020-05-19
Provide a basic test for secondary interface naming metrics

fedepaol:
This is a follow up to https://github.com/openshift/cluster-network-operator/pull/552 , testing for the presence of the basic metric produced by the new daemonset.

The test creates a new secondary network using the `networks.operator.openshift.io` crd and verifies the presence of the metrics for both the primary and the secondary interface.


fedepaol:
/hold
Depends on https://github.com/openshift/cluster-network-operator/pull/552

fedepaol:
> Curious does this needs to be part of the origin tests or can you just add them to the cluster-network-operator?

Yes, I was not 100% sure about this but: 
- I think I was pointed here in the past, as the right place to add new prometheus tests. 
- My reasoning is that we want to test the functionality in the platform. This is not much different as as validating the presence of `openshift_sdn_ovs_flows` or `ingress_metrics` which is being done here.
- By putting these tests in the CNO (or possibly different, in the daemon that produces them) we'll be able to gate the changes only in the single repository. Moreover, I am not seeing e2e tests at all in the CNO repo.

lilic:
> Moreover, I am not seeing e2e tests at all in the CNO repo.

:) interesting, we have e2e tests in our cluster-monitoring-operator, for basic functionality. 

I am not against adding it here just curious if we can just keep it for the component.

fedepaol:
/hold cancel
https://github.com/openshift/cluster-network-operator/pull/552 is merged

fedepaol:
@lilic feature was merged, do you mind giving a look?

PR #24999, opened 2020-05-19
e2e: topomgr: remove policy check

fromanirh:
When we introduced the topology manager extended tests, we added a check
to ensure that the topology manager was configured with the
`single-numa-node` policy. This because the `single-numa-node` policy
is the strictest one, so this is the simplest scenario on which we can
check the correct outcome of the allocation.

Checking other policies in the general case requires either
- complex testing code to properly detect various legal resource allocation
- accept the chance of having false negatives (test code misdetecting legal allocations)
- even a stricter coupling between test code and production code

However, we want to be able to check other policies.
The simplest possible step forward is to just remove the check for the
policy and to put on who sets the test up the responsability of
configuring the environment on such a way that test is relevant.
In other words, that could mean to over-provision the environment to
make sure the aligned allocations are possible with less strict
policies.

This change nothing for the current test scenarios, on which we set the
single-numa-node policy anyway, but makes it possible to use the
testsuite against clusters configured with the `best-effort` policy,
like it is the case when the `performance-addon-operator` is enabled
in the cluster.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,7 +67,6 @@ require (
 	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20200427153329-656914f816f9
 	k8s.io/kubectl v0.0.0
-	k8s.io/kubelet v0.0.0
 	k8s.io/kubernetes v1.19.0-rc.2
 	k8s.io/legacy-cloud-providers v0.0.0
 	sigs.k8s.io/yaml v1.2.0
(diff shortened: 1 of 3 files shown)


fromanirh:
/cc @vpickard @Deepthidharwar 
this is the change we discussed

fromanirh:
This PR is reviewable. The tests still work as expected against a cluster (disclosure: cluster configured with `single-numa-node` policy but for verification's sake each policy is equally good)

fromanirh:
checking why CI complains

fromanirh:
@sjenning could you please review?

fromanirh:
obsolete. Let's re-evaluate later on.

PR #25002, opened 2020-05-20
Bug 1837992: images/hyperkube: install iproute

squeed:
This is so we can access the "ss" utility, which we'd like to use during apiserver startup.

Right now we use `lsof`, which doesn't detect all sockets.


sdodson:
> /cherry-pick release-4.5

This merged prior to branching or at least while release-4.5 branch was still fast forwarding.

PR #25006, opened 2020-05-20
Bug 1836641: bump (*)

gnufied:
We need to bump k8s.io/utils package because it has a 1.18 blocker bug which was fixed - https://github.com/kubernetes/utils/pull/150


gnufied:
cc @openshift/storage

marun:
/hold

k8s.io/utils is bumped to a proper version by the 1.18.3 rebase: #25015

PR #25011, opened 2020-05-21
improve disruption "failure" messaging

bparees:
Currently certain disruption measurement tests will report a failure in
junit but not fail the actual job, if the disruption amount is within a
certain limit.  However the message reported does not make it clear:

1) the job is being passed in spite of this "failure"
2) how to distinguish between "good enough" and "correct" results


bparees:
/assign @smarterclayton 

@deads2k maybe you're involved enough in these to have an opinion here.

bparees:
@sttts thoughts?  you've been working on these tests/jobs lately.

PR #24833, opened 2020-04-04
e2e/network: skip new session affinity timeout tests for OVNKubernetes

dcbw:
Session affinity doesn't work yet in ovnkube; these tests were recently added so we need to ignore them too for now. We already have SDN-510 for implementing session affinity.

Fixes: 16abec0d471f3c

@danwinship @openshift/networking

PR #25012, opened 2020-05-21
Bug 1840707: oVirt, skip network service tests

Gal-Zaidman:
This PR skips the network service tests for ovirt.
we didn't know the reason until recently and now there is a fix which is targeted for RHEL 8.3 [1] and there are openshift bugs for tracking the fix in openshift [2][3][4].
Since this is now a very known issue and is currently failing our CI and making it hard to distinguish between a red failed CI and a false one we decided to skip those test for ovirt until the fix will land.

Tracker-Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1838751

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1794714
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1822714
[3] https://bugzilla.redhat.com/show_bug.cgi?id=1822755
[4] https://bugzilla.redhat.com/show_bug.cgi?id=1822766

Signed-off-by: Gal-Zaidman <gzaidman@redhat.com>


test/extended/util/annotate/rules.go
```diff
@@ -146,6 +146,12 @@ var (
 			`\[sig-storage\] In-tree Volumes \[Driver: azure\] \[Testpattern: Inline-volume`,
 			`\[sig-storage\] In-tree Volumes \[Driver: azure\] \[Testpattern: Pre-provisioned PV`,
 		},
+		"[Skipped:ovirt]": {
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1838751
+			`\[sig-network\] Networking Granular Checks: Services should function for endpoint-Service`,
+			`\[sig-network\] Networking Granular Checks: Services should function for node-Service`,
+			`\[sig-network\] Networking Granular Checks: Services should function for pod-Service`,
+		},
```
bparees, line 154: i don't believe this does what you want.  you need an actual change in the test to invoke skip when the ovirt platform is detected.

see that https://github.com/openshift/origin/pull/25012/files#diff-6ba77494282f6e840f44b01ce97335afR300 will not match this.

Gal-Zaidman: I think that is the wrong URL

bparees: it's right but github doesn't autoexpand for you.

line 300 of this file.

see also:
https://github.com/openshift/origin/blob/master/test/extended/oauth/requestheaders.go#L66-L68

as an example of how this is normally done.

Gal-Zaidman: Looking at other PRs such as:
- https://github.com/openshift/origin/pull/24833/files
- https://github.com/openshift/origin/pull/24797/files

Are you sure I don't need to just edit zz_generated.annotations.go ?

bparees: I am reasonably certain what you've done here will accomplish nothing other than adding a `[Skipped:ovirt]` prefix to the tests in question.

Gal-Zaidman: I think that its what needs to be done, I looked into:
https://github.com/openshift/origin/blob/master/cmd/openshift-tests/provider.go#L57-L63
and it seems that the tests will be skipped if we have [Skipped:ovirt] in them.

Plus the tests come from kubernetes so I don't think we need to send a patch there:
https://github.com/kubernetes/kubernetes/blob/2fd8debe9b913f18b20a2269117409212391ef19/test/e2e/network/networking.go#L162

bparees: Thanks, https://github.com/openshift/origin/blob/master/cmd/openshift-tests/provider.go#L57-L63 is the piece i was missing

bparees: hack/update-generated-bindata.sh should update the other file for you

bparees:
@squeed @danwinship one of you can probably approve this.  There is a bug tracking that these tests need to be re-enabled on ovirt once the rhel issues are resolved.

danwinship:
> @squeed @danwinship one of you can probably approve this. There is a bug tracking that these tests need to be re-enabled on ovirt once the rhel issues are resolved.

I can confirm that it makes sense to disable those tests on oVirt until OCP has a kernel containing a fix for the vxlan offload bug, but I don't have `/approve` power there.
/lgtm

Gal-Zaidman:
/retitle Bug 1840707: oVirt, skip network service tests

PR #25013, opened 2020-05-21
Add publishing rules for origin-4.5-kubernetes-1.18.3

marun:
/cc @sttts @deads2k @soltysh


deads2k:
publish rules only. Looks correct.

PR #25018, opened 2020-05-22
Bug 1839187: more upstream to installed template switch in image-eco (django)

gabemontero:
NOTE: there is no django template in the official OCP catalog that *DOES NOT* employ persistence.

The `oc new-app https://github.com/slcorg/django-ex` ultimately employ a non-persistent template that is in the upstream repo but is not marked for inclusion in OCP.

There was a push by PM a while back to focus on sample templates using persistence, so the lack of a non-persistent template make sense, especially with these SCL ones, who PM at the time would have focused on.

As such, as needed, test case checks will be adjusted to account for the fact that state is persisted in the PV.

### DIFF
--- a/test/extended/image_ecosystem/s2i_python.go
+++ b/test/extended/image_ecosystem/s2i_python.go
@@ -21,15 +21,15 @@ var _ = g.Describe("[sig-devex][Feature:ImageEcosystem][python][Slow] hot deploy
 	defer g.GinkgoRecover()
 
 	var (
-		oc               = exutil.NewCLI("s2i-python")
-		djangoRepository = "https://github.com/sclorg/django-ex.git"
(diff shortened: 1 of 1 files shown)


gabemontero:
OK we have a passing image eco and django test using an official (and now persistent) template with https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/25018/pull-ci-openshift-origin-master-e2e-gcp-image-ecosystem/224

/assign @adambkaplan 

FYI - this change is needed to unblock any samples PRs that need to pass image eco.

gabemontero:
@adambkaplan just in case ... I added detail in the description about persistent / non-persistent based templates and the effects on all this

gabemontero:
networking/auth/monitoring operators degraded on upgrade job

/retest

gabemontero:
OK we have all green tests

PR #25024, opened 2020-05-26
UPSTREAM: 91816: GC doesn't have to create monitors in the constructor

p0lyn0mial:
It prevents the GC from entering a slow loop (up to 1 hour) and blocking the other controllers.

Before this PR the GC used the restmapper (with the cache) in the constructor (a blocking operation) to construct its dependency graph.

In order to fill up the cache, the mapper needs to perform the full discovery which visits all API servers even if what it needs is already in the cache. The cache itself is invalidated by the KCM every 30 seconds. The GC needs to use the mapper many times (~len(deletableResources)) before it creates the graph.

When the discovery is slow (a node is down, a network is slow/broken) and takes more than 30s, creating the graph is also slow (cache invalidation) and might block the other controllers when the GC is run early in the process.

This PR moves the logic that creates the graph to the main loop (async operation) from the constructor.


p0lyn0mial:
updated to match the upstream PR.

PR #25027, opened 2020-05-26
[release 3.11]Bug 1824243: Fix egressVXLANMonitor and egressIPTracker deadlock

juanluisvaladas:
The egressIPTracker has methods that lock eit.mutex and that call
evm functions that lock evm.mutex.

The problem with this is that evm.mutex has to write to the evm.updates
channel which isn't buffered and becomes blocked until
eit.setNodeOffline, which also locks eit.mutex, is running.

This causes a deadlock. Initially I tried doing a horrible hack by
making the updates channel huge, however it wasn't enough.

Instead this fix adds a shared nodes list between eit and evm and the
evm uses the updates channel just to notify the eit that there are
updates in the shared list.

### DIFF
--- a/pkg/network/node/egressip.go
+++ b/pkg/network/node/egressip.go
@@ -62,10 +62,9 @@ func (eip *egressIPWatcher) Start(networkInformers networkinformers.SharedInform
 
 	eip.iptables = iptables
 
-	updates := make(chan *egressVXLANNode)
+	updates := make(chan struct{}, 1)
 	eip.vxlanMonitor = newEgressVXLANMonitor(eip.oc.ovs, eip.tracker, updates)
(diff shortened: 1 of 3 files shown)


juanluisvaladas:
/hold
The BZ has a few blockers on 4.2-4.4 pending to be merged. I'll cancel it as soon as I get that one merged

juanluisvaladas:
e2e-gcp seems to be haven been down for a few days so I'm not retrying it for a while

vikaslaad:
@knobunc could you please skip e2e-gcp tests ? they are broken.

knobunc:

This is failing due to a mirror problem.  We had a clean test run prior to that.  I'll work on working out the test problem separately.

PR #25031, opened 2020-05-27
Bug 1850140: tests: extended: fix TM regression test

fromanirh:
TL;DR: the topology manager code is correct, a specific test fails to capture this leading to false negatives

The existing regression test which wants to cover against regressions
on RHBZ#1813567 is actually doing the reverse check than it should.
We noticed after having the test run for few days on environment kindly
provided by nfvpe team.
The likely cause of why this mistake sneaked in is that I wrongly
configured the original environment on which I tried the test.

This patch fixes the issue and puts extra care in documenting and
checking the intent of the original bugzilla.

### DIFF
--- a/test/extended/topology_manager/resourcealign.go
+++ b/test/extended/topology_manager/resourcealign.go
@@ -260,6 +260,13 @@ var _ = g.Describe("[Serial][sig-node][Feature:TopologyManager] Configured clust
 			})
 
 			// rhbz#1813567
+			// a well-behaving TM must admit only *one* of the pods.
+			// The reference test environment is like this:
+			// 1. Two NUMA nodes
+			// 2. A SRIOV Device - with 2+ VFs - on NUMA node 1. No VFs on the other node (see below)
+			// 3. No other workload is running together with the tests (all resources available to the test)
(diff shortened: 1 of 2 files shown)


fromanirh:
verified on 4.4 (fails as expected), pending verification on 4.5

test/extended/topology_manager/resourcealign.go
```diff
@@ -260,6 +260,11 @@ var _ = g.Describe("[Serial][sig-node][Feature:TopologyManager] Configured clust
 			})
 
 			// rhbz#1813567
+			// a well-behaving TM must *NOT* admit either of the pods.
```
vpickard, line 263: Should this be, a well-behaving TM must admin only 1 of the pods?

fromanirh: Yep, bad wording. Will rephrase and fix.

fromanirh:
had a few runs against a 4.5 cluster, works as expected

fromanirh:
3rd parthy verification: https://bugzilla.redhat.com/show_bug.cgi?id=1826807

fromanirh:
ping?

rphillips:
/retitle Bug 1850140: tests: extended: fix TM regression test

PR #24974, opened 2020-05-13
Bug 1833558: update test to be more resilient for oc adm policy add-scc-to-user changes

soltysh:
/assign @stlaz

PR #25032, opened 2020-05-27
Bug 1833558: update test to be more resilient for oc adm policy add-sss-to-user changes

soltysh:
Another alternate to https://github.com/openshift/origin/pull/24974


stlaz:
/lgtm
the actual fix is supposed to come later

PR #25034, opened 2020-05-27
Bug 1824423: Revert "UPSTREAM: <drop>: Increate timeout in volume expansion test"

bertinatto:
This reverts commit 648d8f84875920df4a6509ec4c1677859117968a.

Currently the flake happens around twice a day. That's probably not caused by the patch itself, but we should drop it anyways because it hasn't improved the situation.


bertinatto:
CC @openshift/storage

bertinatto:
/assign @soltysh 
for approval

PR #25037, opened 2020-05-27
Bug 1839517: UPSTREAM: <carry>: kube-controller-manager: allow running bare kube-controller-manager

soltysh:


deads2k:
Can you go in the other direction.  If we pass no openshift data in the openshift config, let's stop using it and only pass flags in the pod

vendor/k8s.io/kubernetes/cmd/kube-controller-manager/app/patch_flags.go
```diff
@@ -90,32 +92,25 @@ func applyOpenShiftConfigControllerArgs(controllerManagerOptions *options.KubeCo
...
+	if err := applyFlags(args, cmd.Flags()); len(err) > 0 {
 		return kerrors.NewAggregate(err)
 	}
 	return nil
 }
 
 // applyFlags stores the provided arguments onto a flag set, reporting any errors
 // encountered during the process.
-func applyFlags(args map[string][]string, flags flag.NamedFlagSets) []error {
+func applyFlags(args map[string][]string, flags *pflag.FlagSet) []error {
```
sttts, line 103: why is all this still needed?

soltysh: Since we wire and pass these flag values, we can't break them. I don't want to end up in the situation where this stops working, although truth to be told, this never worked in past few releases :sweat: In 4.6 this will be entirely dropped in favor of regular flag passing.

deads2k:
good enough for half a release I guess
/lgtm

PR #25044, opened 2020-05-29
pkg/test/ginkgo: Add 'weight' property to JUnit <testcase>

wking:
This lays the groundwork for having the job-detail page rank failing-case severity.  `failure` cases are the most important, and users who are frustrated by their tests not passing should focus on resolving those issues.  `flake` cases did not sink this particular run, but should be addressed in the absence of `failure` cases to help drive down sporadic stability issues.  `informer` cases do not need to be "fixed", but might provide some context that helps understand cluster health or root causes of failure and flake cases.

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -332,8 +332,14 @@ func (opt *Options) Run(args []string) error {
 	}
 
 	if len(opt.JUnitDir) > 0 {
-		if err := writeJUnitReport("junit_e2e", "openshift-tests", tests, opt.JUnitDir, duration, opt.ErrOut, syntheticTestResults...); err != nil {
-			fmt.Fprintf(opt.Out, "error: Unable to write e2e JUnit results: %v", err)
+		data, err := renderJUnitReport("openshift-tests", tests, duration, syntheticTestResults...)
+		if err == nil {
+			path := filepath.Join(opt.JUnitDir, fmt.Sprintf("junit_e2e_%s.xml", time.Now().UTC().Format("20060102-150405")))
+			fmt.Fprintf(opt.Out, "Writing JUnit report to %s\n\n", path)
+			err = ioutil.WriteFile(path, data, 0640)
+		}
+		if err != nil {
+			fmt.Fprintf(opt.ErrOut, "error: Unable to write e2e JUnit results: %v", err)
 		}
 	}
 
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -290,6 +290,14 @@ func (opt *Options) Run(args []string) error {
 				FailureOutput: &FailureOutput{
 					Output: fmt.Sprintf("%d error level events were detected during this test run:\n\n%s", errorCount, errBuf.String()),
 				},
+				Properties: &JUnitProperties{
+					Properties: []JUnitProperty{
+						{
+							Name:  "weight",
+							Value: "informer",
+						},
+					},
+				},
 			})
 		}
 
@@ -300,7 +308,9 @@ func (opt *Options) Run(args []string) error {
 	if fail > 0 && fail <= suite.MaximumAllowedFlakes {
 		var retries []*testCase
 		for _, test := range failing {
-			retries = append(retries, test.Retry())
+			retry := test.Retry()
+			test.retries = append(test.retries, retry)
+			retries = append(retries, retry)
 			if len(retries) > suite.MaximumAllowedFlakes {
 				break
 			}
--- a/pkg/test/ginkgo/junit.go
+++ b/pkg/test/ginkgo/junit.go
@@ -2,10 +2,6 @@ package ginkgo
 
 import (
 	"encoding/xml"
-	"fmt"
-	"io"
-	"io/ioutil"
-	"path/filepath"
 	"strings"
 	"time"
 
@@ -45,8 +41,8 @@ type JUnitTestSuite struct {
 	// Duration is the time taken in seconds to run all tests in the suite
 	Duration float64 `xml:"time,attr"`
 
-	// Properties holds other properties of the test suite as a mapping of name to value
-	Properties []*TestSuiteProperty `xml:"properties,omitempty"`
+	// Properties holds other properties of the test suite as a mapping of name to value.
+	Properties *JUnitProperties `xml:"properties,omitempty"`
 
 	// TestCases are the test cases contained in the test suite
 	TestCases []*JUnitTestCase `xml:"testcase"`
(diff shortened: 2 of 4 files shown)


wking:
[Failed e2e-cmd job][1] shows this in action:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/25044/pull-ci-openshift-origin-master-e2e-cmd/5741/artifacts/e2e-cmd/junit/junit_e2e_20200529-065238.xml | sed 's/></>\n</g' | grep 'testcase.*weight'
<testcase name="[sig-cli][Feature:LegacyCommandTests][Disruptive][Serial] test-cmd: test/cmd/set-data.sh [Suite:openshift]" time="317" weight="failure">
<testcase name="[sig-cli][Feature:LegacyCommandTests][Disruptive][Serial] test-cmd: test/cmd/set-image.sh [Suite:openshift]" time="317" weight="failure">
<testcase name="Monitor cluster while tests execute" time="2374" weight="informer">
```

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/25044/pull-ci-openshift-origin-master-e2e-cmd/5741

pkg/test/ginkgo/junit.go
```diff
@@ -76,6 +76,16 @@ type JUnitTestCase struct {
 	// Duration is the time taken in seconds to run the test
 	Duration float64 `xml:"time,attr"`
 
+	// Weight is the impact of a failing test on the suite.
```
abhinavdahiya, line 79: what should consumers assume when this is empty? Or does it even make sense to define that?

wking: It's `omitempty`, so it can only be unset.  It should never be unset for any failed tests.  I guess we could call that out in the docstring.  But it gets kind of complicated if test-output consumers are themselves effectively a subsequent test suite.  Would the job-detail page distinguish between test passed/failed and an orthogonal test output strictly valid, invalid but still mostly readable, or completely invalid (e.g. nominally JUnit content that is not XML)?

deads2k:
I like the idea behind it.  I don't know the details of the xml junit production.

wking:
/assign @smarterclayton 

Looks like tge bulk of this logic is Clayton's, so he will probably know if this would somehow break a consumer.

wking:
Poking around in test-infra to see about getting this rendered, I learned that `<testcase>` [already allows for a `<properties>` child][1] and [has for years][2].  So I've pushed 6c61613c16 -> 53a68bde47 to pivot to that approach, which may mean we get rendered output in the job-detail pages with the default lens :crossed_fingers: 

[1]: https://github.com/kubernetes/test-infra/blob/2eab27db6e653d809cd48ad06a35462f1aa3cf0e/gopherage/pkg/cov/junit/xmlwriter.go#L47
[2]: https://github.com/kubernetes/test-infra/pull/9887/commits/ce50d4c65c957286756b26f8baf3c786270b8077#diff-33ff1c70f03ec46f8567ad05cacf5254R47

wking:
Turns out our existing `<properties>` approach was broken.  I've pushed a fix with 53a68bde47 -> d1f453c826 and added a basic unit test.

wking:
[e2e-gcp][1] working:

```console
 curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/25044/pull-c
i-openshift-origin-master-e2e-gcp/8184/artifacts/e2e-gcp/junit/junit_e2e_20200601-232800.xml | sed
 's/></>\n</g' | grep -2 '<property' | cut -b-70                                                
<testsuite name="openshift-tests" tests="2332" skipped="1361" failures
<properties>
<property name="TestVersion" value="v4.1.0+2b2306e-2772-dirty">
</property>
</properties>
--
<system-out>I0601 23:25:09.994212   20068 test_context.go:423] Tolerat
<properties>
<property name="weight" value="failure">
</property>
</properties>
--
<system-out>I0601 23:11:16.998641   13896 test_context.go:423] Tolerat
<properties>
<property name="weight" value="failure">
</property>
</properties>
--
<system-out>&#xA;Timeline:&#xA;&#xA;Jun 01 23:03:12.613 I ns/openshift
<properties>
<property name="weight" value="informer">
</property>
</properties>
```

But looks like we still need lens changes to render the `<property>` data.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/25044/pull-ci-openshift-origin-master-e2e-gcp/8184

wking:
#25053 should also distinguish flaky vs. failing.  But because it will not distinguish informers or fix the `<properties>` handling, I think we want both changes.

pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -290,6 +290,14 @@ func (opt *Options) Run(args []string) error {
 				FailureOutput: &FailureOutput{
 					Output: fmt.Sprintf("%d error level events were detected during this test run:\n\n%s", errorCount, errBuf.String()),
 				},
+				Properties: &JUnitProperties{
```
smarterclayton, line 293: Is this based on any standard?

smarterclayton:
So can you articulate how you are going to use this?  A test can flake because

1. the test is flaky
2. the code under test is broken
3. the code not under test is broken

How would you show things differently because of this?  We have to fix all three of those cases - the frequency is all that matters, and the correlation.  This won't tell the difference between 2 and 3.

I don't think you should be making "value judgements" in the junit about informing vs failure.  If you want to indicate that this passed during retry, that seems like the attribute you would expose, "retried=true", not "informer or failure".

pkg/test/ginkgo/cmd_runsuite.go
wking: > Is this based on any standard?

At least the "test-infra schema supports it" and "we already use it for `<testsuite>`" (which test-infra does not support) standards, as linked from the commit messages.

wking:
> So can you articulate how you are going to use this?

kubernetes/test-infra#17798

> We have to fix all three of those cases...

Agreed, but I think fixing failures is more important than zeroing informers (or we'd have made those informers fatal).

smarterclayton:
I think you should just identify tests that are retries and expose that attribute and have the the lens show that this was a retry failure vs a not retry failure.

You don't have enough context inside the test to say "this failed once, therefore it is informing".  That is only possible to do outside of the test suite.

smarterclayton:
retry=failed and retry=success as properties on the ones that were retried (or some similar attribute)

wking:
> You don't have enough context inside the test to say "this failed once, therefore it is informing". That is only possible to do outside of the test suite.

I am not saying that.  The only `informing` bit in this PR is where we are explicitly creating the informing result.

wking:
> retry=failed and retry=success as properties on the ones that were retried (or some similar attribute)

I don't think we need this.  Have you looked at #25053?

PR #25051, opened 2020-06-01
[3.11] Bug 1822491: UPSTREAM: 89794: Clean up event messages for errors

jsafrane:
Backport 89794 into 3.11.


raghavendra-talur:
@jsafrane is kube-controller-manager.log available in OCP 3.11 installation? The log messages refer to it but we weren't locate the file. If not, what is the equivalent file in OCP 3.11?

gnufied:
@raghavendra-talur controller manager runs in a pod in OCP-3.11, you can get logs from that pod.

PR #25056, opened 2020-06-03
WIP: Bug 1843183: test/e2e/upgrade: Split update testing into per-hop cases

wking:
"Cluster should remain functional during upgrade" is a big bucket, and it's hard to figure out what's going on when it fails based on just the test-case.  This commit pulls the setup steps out into their own case (which is unlikely to fail, but still).  And it puts each hop in its own case, including the source and target version.

FIXME: break out major/minor so we can aggregate in Sippy, etc.

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -111,35 +111,44 @@ func SetUpgradeAbortAt(policy string) error {
 	return fmt.Errorf("abort-at must be empty, set to 'random', or an integer in [0,100], inclusive")
 }
 
-var _ = g.Describe("[sig-arch][Feature:ClusterUpgrade]", func() {
+var _ = g.Describe("[sig-arch][Feature:ClusterUpgrade] Upgrading", func() {
 	f := framework.NewDefaultFramework("cluster-upgrade")
 	f.SkipNamespaceCreation = true
 	f.SkipPrivilegedPSPBinding = true
+	var (
+		config *rest.Config
+		client *configv1client.Clientset
+		dynamicClient dynamic.Interface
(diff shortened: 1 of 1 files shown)


smarterclayton:
/hold

I really don't see how this improves things.  It takes 5 seconds to actually spot which upgrade is failing.  Is this so you can see the summary as a test name?

test/e2e/upgrade/upgrade.go
```diff
@@ -111,35 +111,44 @@ func SetUpgradeAbortAt(policy string) error {
...
-	g.It("Cluster should remain functional during upgrade [Disruptive]", func() {
-		config, err := framework.LoadConfig()
+	var (
+		config *rest.Config
+		client *configv1client.Clientset
+		dynamicClient dynamic.Interface
+		upgCtx *upgrades.UpgradeContext
+	)
+
+	g.It("should load the test configuration", func() {
```
smarterclayton, line 125: You can't do this, it gets processed in the upgrade step.

Dynamic tests cannot be done based on cluster info.

wking:
> It takes 5 seconds to actually spot which upgrade is failing. Is this so you can see the summary as a test name?

Yup.  I don't want more bugs with [this level of initial granularity][1], and would much rather have bugs and success aggregation for `from {source} to {target} should succeed`.  Even better would be having this update-scoped test just be "did the update get through or not" with the parallel disruption constraints being independently fatal and not [feeding into this JUnit bucket][2], but I haven't figured that out yet.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1843183
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1843183#c3

PR #25071, opened 2020-06-05
Bug 1797244: test/extended/dr: update test to reflect new DR process

hexfusion:
The addition of cluster etcd operator introduced a new disaster recovery workflow that broke the disruptive tests. The previous workflow involved doing a snapshot restore of the etcd data file. The advantage to this approach is that once etcd starts the process is complete. The downside is a restore action must take place on each master node.

In 4.4+ we automate this process and require only a single invocation of the cluster-backup.sh script and cluster-restore.sh script. Given new machines the operator then scales etcd on the new machines resulting in a 3 node etcd cluster.

The downside of this process is that it is more disruptive as etcd and the rest of the control plane scales. For this reason, we needed to adjust not only some of the timings of the old test allowing for certain steps to take longer. But also to allow for transient etcd timeouts. During scaling from 1 to 2 etcd members we must lose quorum. The resulting leader election will result in client timeouts.

This is a starting place where we can begin to harden the requirements and improve. But for now lets make sure the workflow itself completes as expected.

### DIFF
--- a/test/extended/dr/quorum_restore.go
+++ b/test/extended/dr/quorum_restore.go
@@ -5,25 +5,24 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/rand"
-	"os"
 	"strings"
 	"time"
 
 	"k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/route53"
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
@@ -36,10 +35,11 @@ import (
 
 const (
 	machineAnnotationName = "machine.openshift.io/machine"
-	localEtcdSignerYaml   = "/tmp/kube-etcd-cert-signer.yaml"
 )
 
 var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
+	defer g.GinkgoRecover()
+
 	f := framework.NewDefaultFramework("disaster-recovery")
 	f.SkipNamespaceCreation = true
 	f.SkipPrivilegedPSPBinding = true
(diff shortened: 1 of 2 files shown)


hexfusion:
flake 
/test e2e-aws-disruptive

hexfusion:
> [sig-etcd][Feature:DisasterRecovery][Disruptive] [Feature:EtcdRecovery] Cluster should restore itself after quorum loss [Disabled:Broken] [Serial] [Suite:openshift]

Passed \o/

test failed because of what I thought was https://github.com/openshift/origin/pull/25087 but CI merged already

test/extended/dr/quorum_restore.go
```diff
@@ -229,20 +233,20 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
...
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 				framework.Logf("Force new revision of kube-apiserver")
-				_, err = oc.AdminOperatorClient().OperatorV1().KubeAPIServers().Patch(context.Background(), "cluster", types.MergePatchType, []byte(`{"spec": {"forceRedepl  oymentReason": "recover-kube-apiserver"}}`), metav1.PatchOptions{})
+				_, err = oc.AdminOperatorClient().OperatorV1().KubeAPIServers().Patch(context.Background(), "cluster", types.MergePatchType, []byte(`{"spec": {"forceRedeploymentReason": "recover-kube-apiserver"}}`), metav1.PatchOptions{})
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 				framework.Logf("Wait for etcd pods to become available")
-				_, err = exutil.WaitForPods(
+				_, err = waitForPodsTolerateClientTimeout(
```
smarterclayton, line 244: wait, why do you have to tolerate timeouts here?  at this point etcd should be available?

test/extended/dr/quorum_restore.go
```diff
@@ -242,7 +242,7 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 					exutil.ParseLabelsOrDie("k8s-app=etcd"),
 					exutil.CheckPodIsReady,
 					expectedNumberOfMasters,
-					20*time.Minute,
+					80*time.Minute,
```
smarterclayton: wow... why?

test/extended/dr/quorum_restore.go
```diff
@@ -147,9 +142,9 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 				}
 
 				framework.Logf("Perform etcd backup on remaining machine %s (machine %s)", survivingNodeName, survivingMachineName)
-				expectSSH("sudo -i /bin/bash -c 'rm -f /root/assets/backup/snapshot*'; sudo -i /bin/bash -x /usr/local/bin/etcd-snapshot-backup.sh /root/assets/backup", survivingNode)
-				framework.Logf("Restore etcd on remaining node %s (machine %s)", survivingNodeName, survivingMachineName)
-				expectSSH(fmt.Sprintf("sudo -i /bin/bash -c '/bin/bash -x /usr/local/bin/etcd-snapshot-restore.sh /root/assets/backup/snapshot* %s'", etcdConnectionString), survivingNode)
+				expectSSH("sudo -i /bin/bash -cx 'sudo rm -rf /home/core/backup; sudo /usr/local/bin/cluster-backup.sh ~core/backup'", survivingNode)
```
smarterclayton, line 148: why do you have to sudo when you're running in sudo?

test/extended/dr/quorum_restore.go
```diff
@@ -211,15 +206,17 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
...
 						defer func() {
 							if r := recover(); r != nil {
 								fmt.Println("Recovered from panic", r)
 							}
 						}()
 						nodes, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/master="})
 						if err != nil {
-							return false, err
+							// scale up to 2nd etcd will make this failure inevitable
+							framework.Logf("return false - err %v", err)
```
smarterclayton: can you improve this error message to say "ignoring errors while waiting for masters to recover: %v"

test/extended/dr/quorum_restore.go
hexfusion: needs tuning I am verifying those now.

test/extended/dr/quorum_restore.go
hexfusion: I am forcing a revision of both etcd and apiserver static pod resources. This causes disruption as we transfer leadership.  I will try to dial this in but it has proven a challenge. The error comes from apiserver.

test/extended/dr/quorum_restore.go
hexfusion: typo, thanks

test/extended/dr/quorum_restore.go
hexfusion: sure thing

hexfusion: hmm rebase is off here let me verify...

PR #25072, opened 2020-06-07
Bug 1844831: oVirt, skip network service tests

Gal-Zaidman:
This is a clone of PR
https://github.com/openshift/origin/pull/25012
For release 4.4

This PR skips the network service tests for ovirt.
we didn't know the reason until recently and now there is a fix which is targeted for RHEL 8.3 [1] and there are openshift bugs for tracking the fix in openshift [2][3][4].
Since this is now a very known issue and is currently failing our CI and making it hard to distinguish between a red failed CI and a false one we decided to skip those test for ovirt until the fix will land.

Tracker-Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1838751

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1794714
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1822714
[3] https://bugzilla.redhat.com/show_bug.cgi?id=1822755
[4] https://bugzilla.redhat.com/show_bug.cgi?id=1822766

Signed-off-by: Gal-Zaidman <gzaidman@redhat.com>


sdodson:
@bparees don't we have a mechanism to mark a test as broken so that it continues to run but doesn't fail the job?

bparees:
> @bparees don't we have a mechanism to mark a test as broken so that it continues to run but doesn't fail the job?

not that i'm aware of, short of explicit logic in the test itself.

sdodson:
/lgtm
Ok, lets make sure that we pull this back out once the kernel fixes land.

bparees:
created https://bugzilla.redhat.com/show_bug.cgi?id=1848570 to track re-enabling this in 4.4.z (the existing bug was for 4.5.z)

PR #24275, opened 2019-12-09
replace hyperkube binary with a bash script

p0lyn0mial:
since `hyperkube` has been removed from k/k (1.17) repository this pulls replaces it with a custom script that has similar behavior.

PR #25073, opened 2020-06-08
apiserver: verbosity 3 on termination and watch-termination binary

sttts:
This should give us insights into requests late in the termination phase. We still won't see those coming after termination. But as we wait 70s anyway to terminate, a request coming into after 20+ seconds (without our current LB configuration) is suspicious.

Note: this needs a fix to actually see logs after SIGTERM. Working on this.

### DIFF
--- a/cmd/watch-termination/main.go
+++ b/cmd/watch-termination/main.go
@@ -1,0 +1,140 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"os/signal"
+	"sync"
+	"syscall"
+	"time"
+)
+
+func main() {
+	terminationLog := flag.String("termination-log-file", "", "Write logs after SIGTERM to this file (in addition to stderr)")
+	terminationLock := flag.String("termination-touch-file", "", "Touch this file on SIGTERM and delete on termination")
+
+	flag.Parse()
+	args := flag.CommandLine.Args()
+
+	if len(args) == 0 {
+		fmt.Println("Missing command line")
+		os.Exit(1)
+	}
+
+	// use special tee-like writer when termination log is set
+	termCh := make(chan struct{})
+	var stderr io.Writer = os.Stderr
+	if len(*terminationLog) > 0 {
+		stderr = &terminationFileWriter{
+			Writer:             os.Stderr,
+			fn:                 *terminationLog,
+			startFileLoggingCh: termCh,
+		}
+	}
+
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = stderr
+
+	// forward SIGTERM and SIGINT to child
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for s := range sigCh {
+			fmt.Fprintf(stderr, "Received signal %s\n", s)
+
+			if len(*terminationLock) > 0 {
+				fmt.Fprintln(stderr, "Touching", *terminationLock)
+				if err := touch(*terminationLock); err != nil {
+					fmt.Fprintln(stderr, fmt.Errorf("error touching %s: %v", *terminationLock, err))
+					// keep going
+				}
+			}
+
+			select {
+			case <-termCh:
+			default:
+				close(termCh)
+			}
+			cmd.Process.Signal(s)
+		}
+	}()
+
+	fmt.Printf("Launching %v\n", cmd)
+	rc := 0
+	if err := cmd.Run(); err != nil {
+		if exitError, ok := err.(*exec.ExitError); !ok {
+			rc = exitError.ExitCode()
+		} else {
+			fmt.Fprintf(stderr, "Failed to launch %s: %v\n", args[0], err)
+			os.Exit(255)
+		}
+	}
+
+	// remove signal handling
+	signal.Stop(sigCh)
+	close(sigCh)
+	wg.Wait()
+
+	if len(*terminationLock) > 0 {
+		os.Remove(*terminationLock)
+	}
+
+	fmt.Fprintf(stderr, "Exit code %d\n", rc)
+	os.Exit(rc)
+}
+
+// terminationFileWriter forwards everything to the embedded writer. When
+// startFileLoggingCh is closed, everything is appended to the given file name
+// in addition.
+type terminationFileWriter struct {
+	io.Writer
+	fn                 string
+	startFileLoggingCh <-chan struct{}
+
+	f *os.File
+}
+
+func (w *terminationFileWriter) Write(bs []byte) (int, error) {
+	select {
+	case <-w.startFileLoggingCh:
+		if w.f == nil {
+			f, err := os.OpenFile(w.fn, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+			if err != nil {
+				log.Fatal(err)
+			}
+			w.f = f
+			fmt.Println("Starting logging to", w.fn)
+		}
+		if n, err := w.f.Write(bs); err != nil {
+			return n, err
+		} else if n != len(bs) {
+			return n, io.ErrShortWrite
+		}
+	default:
+	}
+
+	return w.Writer.Write(bs)
+}
+
+func touch(fn string) error {
+	_, err := os.Stat(fn)
+	if os.IsNotExist(err) {
+		file, err := os.Create(fn)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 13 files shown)


cmd/watch-termination/main.go
```diff
@@ -0,0 +1,122 @@
...
+	if err := cmd.Run(); err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			rc = exitError.ExitCode()
+		} else {
+			log.Fatal(err)
+		}
+	}
+
+	if len(*terminationLock) > 0 {
+		os.Remove(*terminationLock)
```
tnozicka, line 77: lock, so you know if the file was already created in case you are exiting early?

cmd/watch-termination/main.go
```diff
@@ -0,0 +1,122 @@
...
+			startFileLoggingCh: termCh,
+		}
+	}
+
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = stderr
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
```
tnozicka, line 44: are there more termination signals to handle? e.g. SIGABRT?

vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
```diff
@@ -327,6 +328,14 @@ func (s preparedGenericAPIServer) Run(stopCh <-chan struct{}) error {
 
 		<-stopCh
 
+		// Set verbosity to at least 3 in order to log requests.
+		if !klog.V(3) && klog.V(1) {
```
tnozicka, line 332: `!klog.V(3) && klog.V(1)` sounds logically equivalent to `klog.V(1)`

vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
```diff
@@ -327,6 +328,14 @@ func (s preparedGenericAPIServer) Run(stopCh <-chan struct{}) error {
...
+			klog.Infof("Termination has started. Logging requests from here on.")
+			fs := flag.NewFlagSet("termination", flag.ContinueOnError)
+			klog.InitFlags(fs)
```
tnozicka, line 335: is this gonna be global?

cmd/watch-termination/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if len(args) == 0 {
+		fmt.Println("Missing command line")
+		os.Exit(1)
+	}
+
+	// use special tee-like writer when termination log is set
+	termCh := make(chan struct{})
+	var stderr io.Writer = os.Stderr
+	if len(*terminationLog) > 0 {
```
p0lyn0mial, line 30: is `terminationLog` a pointer ?

cmd/watch-termination/main.go
```diff
@@ -0,0 +1,129 @@
...
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		for s := range sigCh {
+			fmt.Fprintf(stderr, "Received signal %s\n", s)
+
+			if len(*terminationLock) > 0 {
+				fmt.Fprintln(stderr, "Touching", *terminationLock)
+				if err := touch(*terminationLock); err != nil {
+					fmt.Fprintln(stderr, err)
+					os.Exit(1)
```
p0lyn0mial, line 53: will this also kill the child process (for example kube-api) ?

cmd/watch-termination/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// forward SIGTERM and SIGINT to child
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		for s := range sigCh {
+			fmt.Fprintf(stderr, "Received signal %s\n", s)
+
+			if len(*terminationLock) > 0 {
+				fmt.Fprintln(stderr, "Touching", *terminationLock)
```
p0lyn0mial, line 50: will we be creating the same file for each signal?

sttts: yes. Termination is only started once. We touch. So this idem-potent modulo time.

cmd/watch-termination/main.go
sttts: no. We might just keep going here. Only log.

sttts: done

cmd/watch-termination/main.go
sttts: why lock?

cmd/watch-termination/main.go
sttts: yes. See flag library.

cmd/watch-termination/main.go
sttts: SIGABRT is handled by Golang. You destroy debugging abilities if you trap it.

vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
sttts: no. it's 2 or 3.

sttts: No, it's 1 or 2.

vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/server/genericapiserver.go
sttts: klog is global

cmd/watch-termination/main.go
sttts: fixed

sttts:
Infra flake?
```
http://base-4-5.ocp.svc/rhel-server/Packages/i/iproute-4.11.0-25.el7_7.2.x86_64.rpm: [Errno 14] curl#6 - "Could not resolve host: base-4-5.ocp.svc; Unknown error"
```

/retest

cmd/watch-termination/main.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}()
+
+	fmt.Printf("Launching %v\n", cmd)
+	rc := 0
+	if err := cmd.Run(); err != nil {
+		if exitError, ok := err.(*exec.ExitError); !ok {
+			rc = exitError.ExitCode()
+		} else {
+			log.Fatal(err)
```
p0lyn0mial, line 76: why do we want to crash here? (I think that this will `os.Exit(255)`)

cmd/watch-termination/main.go
p0lyn0mial: nit: we could use https://golang.org/pkg/flag/#StringVar and pass the address of a local var

images/hyperkube/Dockerfile.rhel
```diff
@@ -2,10 +2,10 @@ FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
 WORKDIR /go/src/github.com/openshift/origin
 COPY . .
 RUN for p in vendor/k8s.io/kubernetes/cmd/kube-apiserver vendor/k8s.io/kubernetes/cmd/kube-controller-manager \
-    vendor/k8s.io/kubernetes/cmd/kube-scheduler vendor/k8s.io/kubernetes/cmd/kubelet; do make build WHAT=$p; done && \
+    vendor/k8s.io/kubernetes/cmd/kube-scheduler vendor/k8s.io/kubernetes/cmd/kubelet cmd/watch-termination; do make build WHAT=$p; done && \
```
p0lyn0mial, line 5: make sure you don't have to update `hyperkube` and `rpm`, see https://github.com/openshift/origin/pull/24275

sttts: We don't need this in the RPM. Do you think I have to do anything beyond this here?

sttts: From #24275 I think we are fine.

cmd/watch-termination/main.go
sttts: also fine. It's an error we will probably never see.

sttts: fixed

cmd/watch-termination/main.go
sttts: not better. This is a common Golang pattern. It's the default flag library after all ;-)

sttts:
Infra.

/retest

sttts:
Infra.

/retest

PR #25076, opened 2020-06-08
UPSTREAM: <carry>: logs a user-agent that uses /healthz endpoint

p0lyn0mial:
Allow us to see what else is using /healhtz endpoint in the system. We think that only kubelet should actually use it. Hopeful it will help us to prevent issues like https://bugzilla.redhat.com/show_bug.cgi?id=1844387

### DIFF
--- a/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/log_healthz.go
+++ b/vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/log_healthz.go
@@ -1,0 +1,29 @@
+package openshiftkubeapiserver
+
+import (
+	"k8s.io/klog"
+	"net/http"
+	"strings"
+
+	"k8s.io/apiserver/pkg/server/healthz"
+)
+
+// LogRequestToHealthz logs a user-agent that uses /healthz endpoint
+var LogRequestToHealthz healthz.HealthChecker = &logRequestToHealthz{}
+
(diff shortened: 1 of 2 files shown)


vendor/k8s.io/kubernetes/openshift-kube-apiserver/openshiftkubeapiserver/log_healthz.go
```diff
@@ -0,0 +1,29 @@
...
+
+func (l *logRequestToHealthz) Name() string {
+	return "logRequestToHealthz"
+}
+
+func (l *logRequestToHealthz) Check(r *http.Request) error {
+	userAgent := r.Header.Get("User-Agent")
+
+	// it seems that identifies itself us "kube-probe/1.18+",
+	if len(userAgent) > 0 && !strings.Contains(userAgent, "kube-probe") {
```
p0lyn0mial, line 24: we could also log when a `User-Agent` hasn't been provided

sttts: does it happen?

p0lyn0mial: good question, we could find out by logging all requests

p0lyn0mial:
closing as it seems to overlap with https://github.com/openshift/origin/pull/25199

PR #25081, opened 2020-06-09
Bug 1831908: UPSTREAM: google/cadvisor: 2575: Fix incorrect diskstats

harche:
https://github.com/google/cadvisor/pull/2575

xref : https://bugzilla.redhat.com/show_bug.cgi?id=1831908

Signed-off-by: Harshal Patil <harpatil@redhat.com>


rphillips:
/close

Looks like we need to patch into https://github.com/openshift/google-cadvisor

PR #25091, opened 2020-06-10
Bug 1845549: manually patch scc/restricted and scc/anyuid for conformamce tests

soltysh:
Since recent changes to SCC which in the long term will be managed by
kas-o oc currently modifies rolebindings when assigning SCC to or
removing from user/group. This PR modifies our conformance test such
that the OpenShift specific restrictions are removed from SCC.

/assign @mfojtik @stlaz

### DIFF
--- a/test/extended/conformance-k8s.sh
+++ b/test/extended/conformance-k8s.sh
@@ -64,8 +64,7 @@ os::log::info "Running Kubernetes conformance suite for ${version}"
 # Execute OpenShift prerequisites
 # Disable container security
 oc adm policy add-scc-to-group privileged system:authenticated system:serviceaccounts
-oc adm policy remove-scc-from-group restricted system:authenticated
-oc adm policy remove-scc-from-group anyuid system:cluster-admins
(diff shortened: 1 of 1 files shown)


test/extended/conformance-k8s.sh
```diff
@@ -64,8 +64,6 @@ os::log::info "Running Kubernetes conformance suite for ${version}"
 # Execute OpenShift prerequisites
 # Disable container security
 oc adm policy add-scc-to-group privileged system:authenticated system:serviceaccounts
```
smarterclayton, line 66: you need to add anyuid too i think.

soltysh: Yeah, tested that locally and just updated. Looking for new results.

soltysh:
This is borked due to https://github.com/openshift/origin/pull/25122/

soltysh:
The conformance is working as expected, and I talked in depth about it with @smarterclayton and @stlaz on slack
/hold cancel
/retest

mfojtik:
/override ci/prow/e2e-cmd

damn thing...

PR #24984, opened 2020-05-14
add test for oauth client secret with plus

deads2k:

PR #25107, opened 2020-06-15
e2e: Add other client-authn methods to token request test

stlaz:
This is a follow-up to https://github.com/openshift/origin/pull/24984

/assign @deads2k


deads2k:
/lgtm

forever put to rest

PR #25108, opened 2020-06-15
UPSTREAM: 91984: genericapiserver library must wait for server.Shutdown

p0lyn0mial:
the genericapiserver library must wait until server.Shutdown completes so that connections are gracefully shutdown.

previously there was a race between shutting down a process (server) and closing connections. That might have resulted in some connections not being properly closed on a TCP level.


p0lyn0mial:
requires https://github.com/openshift/origin/pull/25122 to pass

PR #25111, opened 2020-06-15
Bug 1847043: UPSTREAM: 89053: migration of re-usable package from pkg/kubelet/cm/cpumanager to pkg/kubelet/cm

cynepco3hahue:
ref: kubernetes/kubernetes#89053

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>


rphillips:
matches upstream ref: https://github.com/kubernetes/kubernetes/pull/89053/files

/lgtm

PR #25117, opened 2020-06-15
add end to end OCM workqueue metric verification for pre 1.18 workaround

gabemontero:
per ask from @adambkaplan as part of this get named workqueue metrics for OCM working on per 1.18 versions like 4.4


gabemontero:
/hold 

until https://github.com/openshift/openshift-controller-manager/pull/117 merges

test/extended/prometheus/prometheus_ocm_workqueue.go
```diff
@@ -0,0 +1,85 @@
...
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/ibmcloud"
+)
+
+var _ = g.Describe("[Feature:Prometheus][Feature:Builds] Prometheus", func() {
```
adambkaplan, line 16: Not sure if it is appropriate to add this to the `Conformance` suite. We need to make sure that we never break metrics reported by OCM. `Feature:Builds` will only run in the `e2e-*-build` suites.

test/extended/prometheus/prometheus_ocm_workqueue.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+	})
+
+	g.Describe("when installed on the cluster", func() {
+		g.It("should start and expose a secured proxy and verify OCM workqueue metrics", func() {
+
+			// https://issues.redhat.com/browse/CO-895
+			if e2e.TestContext.Provider == ibmcloud.ProviderName {
+				e2e.Skipf("Prometheus in IBM ROKS clusters does not collect metrics from the OpenShift controller manager because it lives outside of the cluster. " +
+					"The openshift_build_total metric expected by this test is reported by the OpenShift controller manager. " +
```
adambkaplan, line 44: nit - 
```suggestion
					"The metrics expected by this test are reported by the OpenShift controller manager. " +
```

test/extended/prometheus/prometheus_ocm_workqueue.go
```diff
@@ -0,0 +1,85 @@
...
+			// instantiating prometheus tempalte
+			var err error
+			for i := 0; i < maxPrometheusQueryAttempts; i++ {
+				err = expectURLStatusCodeExec(ns, execPod.Name, url, 403)
+				if err == nil {
+					break
+				}
+				time.Sleep(prometheusQueryRetrySleep)
+			}
+			o.Expect(err).NotTo(o.HaveOccurred())
```
adambkaplan, line 65: Is the above step needed to ensure that the `oauth-proxy` is up and the service account token will authenticate?

test/extended/prometheus/prometheus_ocm_workqueue.go
gabemontero: meant to switch to Conformance and forgot ... as I think you are, given the broad range of workqueues (networking, deployment configs, images stuff, builds, templates) not sure what else

and sig-instrumentation is maybe the only other choice, but that seems less right ... of course chime in if you think differently

will update to Conformance here shortly in the interim just in case

gabemontero: actually sig-instrumentation is better than [Feature:Prometheus]

will do that and conformance

gabemontero: no, sig-* are master/4.5 labels :-)

just switch builds to conformance

test/extended/prometheus/prometheus_ocm_workqueue.go
gabemontero: part of next push

test/extended/prometheus/prometheus_ocm_workqueue.go
gabemontero: the above is the endpoint/svc is up, the code below is related to the SA token

it mimics what is done in prometheus_builds.go 

prometheus.go has something similar, but without the retry on the 403 ... as I recall my testing back in 3.10 for builds seemed to hit timing windows the vanilla prometheus did not ... the diff goes back that far ... but it has been a while

gabemontero:
updates pushed @adambkaplan

gabemontero:
@adambkaplan and I talked and we are going to do this test out of band with the bug, and starting with master branch

un-associatiating with the bug

PR #25118, opened 2020-06-16
watch-termination: don't log oauth tokens

sttts:
We log requests of oauth tokens when log level is increased. But in the default case we don't log any requests. This PR restore the old (sub-optimal) situation, i.e. we are not worse than before.

### DIFF
--- a/cmd/watch-termination/main.go
+++ b/cmd/watch-termination/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


cmd/watch-termination/main.go
```diff
@@ -123,7 +139,7 @@ func (w *terminationFileWriter) Write(bs []byte) (int, error) {
 				Compress:   false,
 			}
 			w.logger = l
-			klog.Infof("Redirecting termination logs to %q", w.fn)
+			fmt.Fprintf(os.Stderr, "Copying termination logs to %q\n", w.fn)
```
sttts, line 142: with the recursive `klog.Infof` there was no logging at all.

p0lyn0mial:
requires https://github.com/openshift/origin/pull/25122

sttts:
Infra.

/retest

PR #25121, opened 2020-06-16
Bug 1847185: fix: GetLabelsForVolume panic issue for azure disk PV

enxebre:
This prevents `GetAzureDiskLabels` from panicking when `c.DisksClient` is nil.
This panicking makes the API server to crash loop for some ARO clusters as elaborated in https://bugzilla.redhat.com/show_bug.cgi?id=1847185

https://github.com/kubernetes/kubernetes/pull/92166
https://github.com/kubernetes/kubernetes/issues/92167

### DIFF
--- a/vendor/k8s.io/kubernetes/staging/src/k8s.io/legacy-cloud-providers/azure/azure_managedDiskController.go
+++ b/vendor/k8s.io/kubernetes/staging/src/k8s.io/legacy-cloud-providers/azure/azure_managedDiskController.go
@@ -340,6 +340,14 @@ func (c *Cloud) GetAzureDiskLabels(diskURI string) (map[string]string, error) {
 		return nil, err
 	}
 
+	labels := map[string]string{
+		v1.LabelZoneRegion: c.Location,
+	}
+	// no azure credential is set, return nil
+	if c.DisksClient == nil {
+		return labels, nil
+	}
+
 	// Get information of the disk.
 	ctx, cancel := getContextWithCancel()
 	defer cancel()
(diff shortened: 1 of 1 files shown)


enxebre:
cc @sttts @mfojtik @jim-minter PTAL

jim-minter:
@enxebre I'm encouraging upstream to fix this more comprehensively if it's possible.  Might be worth waiting a little to see what happens?

enxebre:
/hold
as per https://github.com/openshift/origin/pull/25121#issuecomment-644792714

enxebre:
/hold cancel
Added latest upstream changes. I'd be happy to get this as it is to fix panicking and prevent any other cluster from getting to a crash looping state and so we can get it into 4.5 before freezing and backport to 4.4 asap.

Then if ARO still needs a way to instantiate the azure client for the api server can be discussed separately. I think getting back to https://github.com/Azure/ARO-RP/pull/487/files#diff-f23154e33f71b30de1fae50fbf2b1dadL56-L60 might be an option. FWIW cloud providers are planned to go out of tree in kubernetes 1.21 / OCP 4.8.

jim-minter:
> Added latest upstream changes. I'd be happy to get this as it is to fix panicking and prevent any other cluster from getting to a crash looping state and so we can get it into 4.5 before freezing and backport to 4.4 asap.

@enxebre I agree.  We'd like to get this into 4.3 urgently.

enxebre:
@jim-minter as per @mjudeikis this has been validated to prevent ARO from panicking.
@sttts @mfojtik can we get labels here?

mfojtik:
merging with button, the queue is hosed and this is critical fix needed in ARO 4.5.

mfojtik:
@eparis @jwforres @stevekuznetsov 3 days of retesting this simple fix that is actually causing 40% of our ARO cluster to be broken... i would like to know how we can mitigate this in future.

PR #25135, opened 2020-06-17
Bug 1809195: test/extended/prometheus: Require HTTPS for the CVO

wking:
Now that the CVO pull request has landed, require HTTPS going forward (firming up the flexible behavior from #25134).

/hold

Until openshift/cluster-version-operator#385 actually lands.


wking:
/hold cancel

CVO PR merged

PR #25141, opened 2020-06-17
Bug 1849175: UPSTREAM: 91569: kubelet: block non-forwarded packets from crossing the localhost boundary

joelsmith:
See https://github.com/kubernetes/kubernetes/pull/91569


rphillips:
/retitle Bug 1843358: 91569: kubelet: block non-forwarded packets from crossing the localhost boundary

rphillips:
/retitle Bug 1843358: UPSTREAM: 91569: kubelet: block non-forwarded packets from crossing the localhost boundary

PR #25144, opened 2020-06-17
Bug 1845773: [release-4.3] kubelet: block non-forwarded packets from crossing the localhost boundary

joelsmith:
This is a release-4.3 cherry-pick of https://github.com/openshift/origin/pull/25141


bparees:
bot can't handle the BZ deps on this because it's a private bug, please manually apply the label once the 4.4.z bug is verified.

PR #25156, opened 2020-06-18
Bug 1826183: Verify builds use custom CA bundle

adambkaplan:
Add an e2e test to verify that builds use a custom CA bundle if one is added to the cluster
proxy config.

### DIFF
--- a/test/extended/builds/cluster_config.go
+++ b/test/extended/builds/cluster_config.go
@@ -3,6 +3,7 @@ package builds
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"time"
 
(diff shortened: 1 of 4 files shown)


adambkaplan:
https://github.com/openshift/builder/pull/158 and https://github.com/openshift/openshift-controller-manager/pull/119 should merge in order for the test to pass. If it passes now, we have a problem.

adambkaplan:
/close

This test is inherently flaky because changing the PKI causes node drain events. Even running the test serially, the build pod gets deleted more than 50% of the time. This may be best verified manually by QE.

PR #25160, opened 2020-06-18
images/os/Dockerfile: use a recent Fedora

dcbw:


marun:
@dcbw Considering the failed jobs and the [ci config for this branch](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L1%200), maybe bumping the fedora version requires coordinating a ci change?

cc: @smarterclayton

dcbw:
@marun yeah, I'm just poking around here and seeing what works/doesn't. I should WIP this one.

PR #25169, opened 2020-06-19
Move admin.sh to a proper e2e

soltysh:
This is slowly switching test-cmd bits into proper e2e, I've started with admin.sh

/assign @smarterclayton @mfojtik 
I need to sync with you if we want to keep them separate test group, or it's ok to add them to overall e2e. The time needed to run it is less than a minute, so it's almost negligible but we'll have additional 50-ish files to be switched, so I guess overall it might be 15-20-ish minutes. 

/assign @stlaz @sallyom 
for actual checks if the tests make sense, I've managed to bring some back, but I also skipped those that are testing deprecated functionality (`oc adm ca`), which are tested elsewhere (`oc adm prune images`), or have been duplicates. I'll add comments below explaining those.


soltysh:
/hold
for comments, feedback & a few test runs

test/extended/cli/admin.go
```diff
@@ -0,0 +1,517 @@
...
+		o.Expect(out).To(o.MatchRegexp(`default/busybox\W+653\.4KiB\W+1\W+1`))
+
+		oc.Run("delete").Args("-f", stableBusyboxPath).Execute()
+	})
+
+	// TODO (soltysh): sync with Standa and figure out if we can get these
+	// enabled back, they were all disabled in admin.sh:
+
+	// 	os::test::junit::declare_suite_start "cmd/admin/rolebinding-allowed"
+	// # Admin can bind local roles without cluster-admin permissions
```
stlaz, line 439: You could create a synthetic user by creating a `User` and `OAuthAccessToken` objects, in which case you would use `oc login -t <token>` to log in as that user.

stlaz: I think the non-existence of the `local-admin` user is probably the only reason these tests were disable, right?

test/extended/testdata/cmd/test/cmd/admin.sh
```diff
@@ -1,373 +0,0 @@
...
-#os::cmd::expect_success "oc get clusterrolebinding/basic-users2"
-#os::cmd::expect_success "oc adm prune auth clusterrole/basic-user2"
-#os::cmd::try_until_failure "oc get clusterrolebinding/basic-users2"
-#os::cmd::expect_success "oc delete clusterrole/basic-user2"
-#os::cmd::expect_success "oc policy add-role-to-user edit foo"
-#os::cmd::expect_success "oc get rolebinding/edit"
-#os::cmd::expect_success "oc adm prune auth clusterrole/edit"
-#os::cmd::try_until_failure "oc get rolebinding/edit"
-#os::cmd::expect_success "oc delete clusterrole/edit"
-#os::cmd::expect_success 'oc auth reconcile --remove-extra-permissions --remove-extra-subjects -f "${BASE_RBAC_DATA}"'
```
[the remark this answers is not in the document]
stlaz: Thanks for bringing these back. I am that much familiar with `oc auth reconcile`, is there a specific reason for it not to be favored?

stlaz:
```
[INFO] hack/verify-generated-versions.sh exited with code 0 after 00h 00m 07s
./test/extended/util/client.go
[FATAL] Try running 'gofmt -s -d [path]'
[FATAL] Or autocorrect with 'hack/verify-gofmt.sh | xargs -n 1 gofmt -s -w'
[ERROR] hack/verify-gofmt.sh exited with code 1 after 00h 00m 01s
```

test/extended/cli/admin.go
soltysh: Unfortunately, the token usage is not supported, yet. See my other comment earlier in the file. I will work on getting it work, though. But it might be a bit later.

test/extended/testdata/cmd/test/cmd/admin.sh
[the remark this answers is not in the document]
soltysh: reconcile re-creates the roles and cluster roles to its default state, iirc it's done currently on every apiserver restart. Most importantly, though, you need that file and with the current proliferation I doubt we have a single file any more.

test/extended/cli/admin.go
```diff
@@ -0,0 +1,517 @@
...
+		// Test deleting and recreating a project
+		o.Expect(oc.Run("adm", "new-project").Args("recreated-project", "--admin=createuser1").Execute()).To(o.Succeed())
+		o.Expect(oc.Run("delete").Args("project", "recreated-project").Execute()).To(o.Succeed())
+		err := wait.Poll(cliInterval, cliTimeout, func() (bool, error) {
+			err := ocns.Run("get").Args("project/recreated-project").Execute()
+			return err == nil, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Expect(oc.Run("adm", "new-project").Args("recreated-project", "--admin=createuser2").Execute()).To(o.Succeed())
```
stlaz, line 336: The recreation fails as the previous project was still present:
https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/25169/pull-ci-openshift-origin-master-e2e-gcp/1275050036434046976

soltysh: I'll have to add the unique prefix, probably for the best.

soltysh:
@stlaz this is ready for final reviews, I've added generated names everywhere it seemed reasonable

test/extended/cli/admin.go
stlaz: I wonder whether just adding `--wait` would be enough here as in the failing test, the project was already being terminated, but wasn't done yet

soltysh:
@stlaz @sallyom ready for final review

test/extended/util/client.go
```diff
@@ -517,9 +525,13 @@ func (c *CLI) Run(commands ...string) *CLI {
...
-		globalArgs: append([]string{
-			fmt.Sprintf("--kubeconfig=%s", c.configPath),
-		}, commands...),
+		globalArgs:      commands,
+	}
+	if len(c.configPath) > 0 {
+		nc.globalArgs = append([]string{fmt.Sprintf("--kubeconfig=%s", c.configPath)}, nc.globalArgs...)
+	}
+	if len(c.configPath) == 0 && len(c.token) > 0 {
+		nc.globalArgs = append([]string{fmt.Sprintf("--token=%s", c.token)}, nc.globalArgs...)
```
stlaz, line 533: don't you still need kubeconfig to know which cluster you're working with when you're using a token?

PR #25179, opened 2020-06-21
Bug 1850109: remove problematic upstream client plugin test

gabemontero:
@adambkaplan @akram @waveywaves fyi

PR #25181, opened 2020-06-21
[4.5] disable broken jenkins test

bparees:
this test consistently fails because the pipeline references a second deployment which does not exist.

### DIFF
--- a/test/extended/builds/pipeline_origin_bld.go
+++ b/test/extended/builds/pipeline_origin_bld.go
@@ -217,6 +217,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds][sig-devex][Feature:Jenkins][Slo
 	g.Context("jenkins-client-plugin tests", func() {
 
 		g.It("using the ephemeral template", func() {
+			g.Skip("skipping because second deployment does not exist")
 			defer cleanup(jenkinsEphemeralTemplatePath)
 			setupJenkins(jenkinsEphemeralTemplatePath)
 


bparees:
@akram @gabemontero 4.5 backport to unblock 4.5 merges.

gabemontero:
@bparees yep saw that as well this weekend and have https://github.com/openshift/origin/pull/25179 up .. though only in master for now ... I tagged most everyone but you in the recent PR comments.

reader's digest version:  it got fixed a week or so ago and then rebroke over the last few days

my PR removes the problematic scneario vs. disables the whole test, but I'm fine with what you have done going in followed by my PR 

I'm also working on updates to the jenkins client plugin to move the test removed from my PR into that repo via a operator golang styled e2e ... this should resolve that old problem we've had if you recall with testing the master branch of that test file vs. and PR version of it.

I'm testing that now and will tag you in the client plugin PR when it is up.

bparees:
failures not related to this change and this is deadlocked w/ the DR etcd test fix, so manually merging this one.

akram:
@bparees and @gabemontero , I saw this also. We also have them failing in 4.2 for which I proposed a first fix changing the the origin-cli that we depend on https://github.com/openshift/jenkins/pull/1089 , because, there was a regression on `oc run`  command that now creates pods instead of deployments.
This fixed 2 e2e over the 4 failing. But, there is still something remaining. I put them on hold then.

skipping these tests for now would help

PR #25185, opened 2020-06-23
only patch authenticators when there's no webhook authenticators

stlaz:
As designed in: https://github.com/openshift/enhancements/pull/142

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/kubeapiserver/authenticator/config.go
+++ b/vendor/k8s.io/kubernetes/pkg/kubeapiserver/authenticator/config.go
@@ -189,13 +189,14 @@ func (config Config) New() (authenticator.Request, *spec.SecurityDefinitions, er
 		}
 
 		tokenAuthenticators = append(tokenAuthenticators, webhookTokenAuth)
+	} else {
+		// openshift gets injected here as a separate token authenticator because we also add a special group to our oauth authorized
+		// tokens that allows us to recognize human users differently than machine users.  The openAPI is always correct because we always
+		// configuration service account tokens, so we just have to create and add another authenticator.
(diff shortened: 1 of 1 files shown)


vendor/k8s.io/kubernetes/pkg/kubeapiserver/authenticator/config.go
```diff
@@ -189,14 +189,15 @@ func (config Config) New() (authenticator.Request, *spec.SecurityDefinitions, er
 		}
 
 		tokenAuthenticators = append(tokenAuthenticators, webhookTokenAuth)
+	} else {
+		// openshift gets injected here as a separate token authenticator because we also add a special group to our oauth authorized
+		// tokens that allows us to recognize human users differently than machine users.  The openAPI is always correct because we always
+		// configuration service account tokens, so we just have to create and add another authenticator.
+		// TODO make this a webhook authenticator and remove this patch.
```
mfojtik, line 196: please open a BZ

stlaz: Added enhancement PR link instead

marun:
/lgtm

@stlaz Can I squash this commit with the one introducing the patch during the rebase?

stlaz:
> Can I squash this commit with the one introducing the patch during the rebase?

sure, I would even advise to do that :)

vendor/k8s.io/kubernetes/pkg/kubeapiserver/authenticator/config.go
sttts: This is part of our move to webhooks in general as described in the enhancement. No need to track that in BZ.

PR #25188, opened 2020-06-23
Bug: Cockroach  2.1.11 operator was removed

kevinrizza:
This commit updates an olm e2e test that ensures operator subscription
works. The test currently relies on an external repository of operator
metadata to generate that subscription. An update removed the 2.1.11
version from the current update graph, so this test is broken. As a
temporary solution, updating the test to point to the latest available
version. In the future, this test should either ship its own manifests
(so that it is not dependent on an external content souce) or be
removed entirely.


kevinrizza:
closing, resolved this for now by reaching out to the upstream owners for this content.

PR #25191, opened 2020-06-23
Bug 1844483: poll on deleted obj check; give controller finalizer deletes time to complete

gabemontero:
/assign @adambkaplan 

/hold

feel free to lgtm if you are happy @adambkaplan 

I plan on running the e2e several times to at least get a warm fuzzy on addressing the timing window I'm speculating about

### DIFF
--- a/test/extended/templates/templateinstance_cross_namespace.go
+++ b/test/extended/templates/templateinstance_cross_namespace.go
@@ -2,6 +2,7 @@ package templates
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"time"
@@ -110,12 +111,18 @@ var _ = g.Describe("[sig-devex][Feature:Templates] templateinstance cross-namesp
 		})
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		framework.Logf("Template Instance object: %#v", templateinstance)
-		_, err = cli.KubeClient().CoreV1().Secrets(cli.Namespace()).Get(context.Background(), "secret1", metav1.GetOptions{})
-		o.Expect(err).NotTo(o.HaveOccurred())
+		tiJSON, _ := json.MarshalIndent(templateinstance, "", "    ")
+		framework.Logf("Template Instance object : %s", string(tiJSON))
 
-		_, err = cli.KubeClient().CoreV1().Secrets(cli2.Namespace()).Get(context.Background(), "secret2", metav1.GetOptions{})
+		s1, err := cli.KubeClient().CoreV1().Secrets(cli.Namespace()).Get(context.Background(), "secret1", metav1.GetOptions{})
 		o.Expect(err).NotTo(o.HaveOccurred())
+		s1JSON, _ := json.MarshalIndent(s1, "", "    ")
+		framework.Logf("secret1: %s", string(s1JSON))
+
+		s2, err := cli.KubeClient().CoreV1().Secrets(cli2.Namespace()).Get(context.Background(), "secret2", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		s2JSON, _ := json.MarshalIndent(s2, "", "    ")
+		framework.Logf("secret2: %s", string(s2JSON))
 
 		g.By("deleting the templateinstance")
 		foreground := metav1.DeletePropagationForeground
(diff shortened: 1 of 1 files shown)


gabemontero:
forgot a key point ... in theory, with "foreground" gc, I would think we would not need this retry

so not hitting the issue implies to me a disconnect perhaps wrt "what foreground gc means" ?

gabemontero:
I'm also adding more bread crumbs in line with the test to sort out owner refs, finalizers, etc.

Will post comment when those are pushed @adambkaplan

gabemontero:
> forgot a key point ... in theory, with "foreground" gc, I would think we would not need this retry
> 
> so not hitting the issue implies to me a disconnect perhaps wrt "what foreground gc means" ?

forgot templates has a finalizer and its own deletion of dependents in its controller ... so timing gaps could come into play

gabemontero:
ok @adambkaplan finally ready for review, and I believe this is the real fix, once I got unconfused around use of k8s foreground vs. template's use of a finalizer 

also note, the template instance controller submits events if its delete attempts fail .... looking at several failures from search-ci, I'm not seeing any events in the OCM namespace

pretty convinced now this is timing window thing from when deletes are issued and when etcd processes them (where order is not guaranteed given etcd's QofS as a non-transactional KV store.

gabemontero:
install went degraded never completed

/test e2e-gcp

gabemontero:
passed:  `[sig-devex][Feature:Templates]  templateinstance cross-namespace test should create and delete objects  across namespaces [Suite:openshift/conformance/parallel] | 38s`

/test e2e-gcp

gabemontero:
Another passing result:  `[sig-devex][Feature:Templates]  templateinstance cross-namespace test should create and delete objects  across namespaces [Suite:openshift/conformance/parallel] | 38s`

/retest

/test e2e-gcp

test/extended/templates/templateinstance_cross_namespace.go
```diff
@@ -124,19 +131,53 @@ var _ = g.Describe("[sig-devex][Feature:Templates] templateinstance cross-namesp
...
 			return false, err
 		})
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		_, err = cli.KubeClient().CoreV1().Secrets(cli.Namespace()).Get(context.Background(), "secret1", metav1.GetOptions{})
-		o.Expect(kerrors.IsNotFound(err)).To(o.BeTrue())
+		// poll in case the dependent objects deletions in the template instance controller finalizer complete with
+		// unexpected timings compared to the template instance itself; as you'll see in the debub prints above,
+		// there is not use of `ownerReferences` with templates ... we delete dependent objects, much like we delete
+		// operands today from our CVO operator finalizers
```
adambkaplan, line 163: nit - this comment could be more succinct.
```suggestion
		// Template instances use a finalizer to delete objects, rather than owner references. As a result, objects
		// created by the template can be deleted after the TemplateInstance itself is deleted. 
```

test/extended/templates/templateinstance_cross_namespace.go
```diff
@@ -124,19 +131,53 @@ var _ = g.Describe("[sig-devex][Feature:Templates] templateinstance cross-namesp
...
-		_, err = cli.KubeClient().CoreV1().Secrets(cli2.Namespace()).Get(context.Background(), "secret2", metav1.GetOptions{})
-		o.Expect(kerrors.IsNotFound(err)).To(o.BeTrue())
+		err = wait.Poll(100*time.Millisecond, 30*time.Second, func() (bool, error) {
+			_, err = cli.KubeClient().CoreV1().Secrets(cli.Namespace()).Get(context.Background(), "secret2", metav1.GetOptions{})
+			if kerrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
```
adambkaplan, line 180: Can't we do one poll operation, rather than two? I'd be OK increasing the timeout to 1 minute in that case.

test/extended/templates/templateinstance_cross_namespace.go
```diff
@@ -124,19 +131,53 @@ var _ = g.Describe("[sig-devex][Feature:Templates] templateinstance cross-namesp
...
 		err = wait.Poll(100*time.Millisecond, 30*time.Second, func() (bool, error) {
-			_, err = cli.TemplateClient().TemplateV1().TemplateInstances(cli.Namespace()).Get(context.Background(), templateinstance.Name, metav1.GetOptions{})
+			t, err := cli.TemplateClient().TemplateV1().TemplateInstances(cli.Namespace()).Get(context.Background(), templateinstance.Name, metav1.GetOptions{})
 			if kerrors.IsNotFound(err) {
 				return true, nil
 			}
+
+			tiJSON, _ := json.MarshalIndent(t, "", "    ")
+			framework.Logf("Template Instance object during deletion : %s", string(tiJSON))
+			s1, e1 := cli.KubeClient().CoreV1().Secrets(cli.Namespace()).Get(context.Background(), "secret1", metav1.GetOptions{})
```
adambkaplan, line 141: I would add a comment here that `IsNotFound` errors for the secrets are to be expected during this phase of the test.

test/extended/templates/templateinstance_cross_namespace.go
gabemontero: I think we can keep it at 30 but include both  in the poll ... will update thanks

test/extended/templates/templateinstance_cross_namespace.go
gabemontero: will add

gabemontero:
PR updated @adambkaplan PTAL

gabemontero:
another passing template suite run ... starting another

/test e2e-gcp

/retest 

for the non-related flakes

gabemontero:
even amongst the cluster degradation any many failures in the last e2e-gcp run our test in question passed

/retest

gabemontero:
rpm flake on latest e2e-gcp

/retest

gabemontero:
src build pod was killed almost immediately by the kubelet on e2e-gcp 

```

Normal | Started | 2h22m | kubelet | Started container
-- | -- | -- | -- | --
Normal | Killing | 2h22m | kubelet | Killing container with id docker://sidecar:Need to kill Pod

```

/test e2e-gcp

gabemontero:
upgrade full of static pod degraded errors

/retest

test/extended/templates/templateinstance_cross_namespace.go
```diff
@@ -124,19 +131,49 @@ var _ = g.Describe("[sig-devex][Feature:Templates] templateinstance cross-namesp
...
+		// Template instances use a finalizer to delete objects, rather than owner references. As a result, objects
+		// created by the template can be deleted after the TemplateInstance itself is deleted.
+		err = wait.Poll(100*time.Millisecond, 30*time.Second, func() (bool, error) {
+			_, e1 := cli.KubeClient().CoreV1().Secrets(cli.Namespace()).Get(context.Background(), "secret1", metav1.GetOptions{})
+			_, e2 := cli.KubeClient().CoreV1().Secrets(cli.Namespace()).Get(context.Background(), "secret2", metav1.GetOptions{})
+			if kerrors.IsNotFound(e1) && kerrors.IsNotFound(e2) {
+				return true, nil
+			}
+			return false, err
+		})
```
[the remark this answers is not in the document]
gabemontero: ah good catch thanks

gabemontero:
I'll squash / gofmt as needed your suggestion in a sec @adambkaplan

gabemontero:
update pushed @adambkaplan

gabemontero:
ci thrashing with serial job

hitting pause

/hold

gabemontero:
Opened https://bugzilla.redhat.com/show_bug.cgi?id=1851623 for apparent/consistent topology manager failure

gabemontero:
disabling of topology manager test has merged ... let's try again

/retest
/hold cancel

PR #25192, opened 2020-06-23
watch-termination: create NonGracefulTermination event on next launch

sttts:
When a termination lock file is found while starting up, an NonGracefulTermination event is created
using the passed kubeconfig, waiting in the background until the kubeconfig starts working (e.g. until
the kube-apiserver launches).

### DIFF
--- a/cmd/watch-termination/main.go
+++ b/cmd/watch-termination/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io"
@@ -14,6 +15,12 @@ import (
 	"time"
 
 	"gopkg.in/natefinch/lumberjack.v2"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog"
 )
 
@@ -24,6 +31,7 @@ func main() {
 func run() int {
 	terminationLog := flag.String("termination-log-file", "", "Write logs after SIGTERM to this file (in addition to stderr)")
 	terminationLock := flag.String("termination-touch-file", "", "Touch this file on SIGTERM and delete on termination")
+	kubeconfigPath := flag.String("kubeconfig", "", "Optional kubeconfig used to create events")
 
 	klog.InitFlags(nil)
 	flag.Set("v", "9")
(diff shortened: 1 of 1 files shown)


cmd/watch-termination/main.go
```diff
@@ -53,6 +65,54 @@ func main() {
 		klog.SetOutputBySeverity("INFO", stderr)
 	}
 
+	var client kubernetes.Interface
+	if len(*kubeconfigPath) > 0 {
```
mfojtik, line 69: check for nil ?

cmd/watch-termination/main.go
```diff
@@ -53,6 +65,54 @@ func main() {
...
+		loader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: *kubeconfigPath}, nil)
+		if cfg, err := loader.ClientConfig(); err != nil {
+			klog.Errorf("failed to load kubeconfig %q: %v", *kubeconfigPath, err)
```
mfojtik, line 72: klog.Fatalf() ?

cmd/watch-termination/main.go
```diff
@@ -53,6 +65,54 @@ func main() {
...
+			return 1
+		} else {
+			client = kubernetes.NewForConfigOrDie(cfg)
+		}
+	}
+
+	// touch file early. If the file is not removed on termination, we are not
+	// terminating cleanly via SIGTERM.
+	if len(*terminationLock) > 0 {
```
mfojtik, line 81: nit: check for nil

cmd/watch-termination/main.go
```diff
@@ -53,6 +65,54 @@ func main() {
...
+		ref, err := eventReference()
+		if err != nil {
+			klog.Errorf("failed to get event target: %v", err)
```
mfojtik, line 84: we continue without reference?

cmd/watch-termination/main.go
```diff
@@ -53,6 +65,54 @@ func main() {
...
+		}
+
+		if st, err := os.Stat(*terminationLock); err == nil {
+			klog.Warningf("Previous pod did not terminate gracefully: %s", st.ModTime().String())
+			if client != nil {
+				go wait.PollUntil(5*time.Second, func() (bool, error) {
+					if err := Eventf(client.CoreV1().Events(ref.Namespace), *ref, corev1.EventTypeWarning, "NonGracefulTermination", "Previous pod did not terminate gracefully: %s", st.ModTime().String()); err != nil {
```
mfojtik, line 91: if ref == nil, this will panic?

cmd/watch-termination/main.go
sttts: no. It can never be nil by definition of the flag library.

cmd/watch-termination/main.go
```diff
@@ -166,3 +213,64 @@ func touch(fn string) error {
 	currentTime := time.Now().Local()
 	return os.Chtimes(fn, currentTime, currentTime)
 }
+
+func Eventf(client corev1client.EventInterface, ref corev1.ObjectReference, eventType, reason, messageFmt string, args ...interface{}) error {
```
mfojtik, line 217: why exported?

cmd/watch-termination/main.go
sttts: no need for backtrace. This is a main func.

cmd/watch-termination/main.go
sttts: nope. same reason

cmd/watch-termination/main.go
```diff
@@ -53,6 +65,54 @@ func main() {
...
+	if len(*terminationLock) > 0 {
+		ref, err := eventReference()
+		if err != nil {
+			klog.Errorf("failed to get event target: %v", err)
+		}
+
+		if st, err := os.Stat(*terminationLock); err == nil {
+			klog.Warningf("Previous pod did not terminate gracefully: %s", st.ModTime().String())
+			if client != nil {
+				go wait.PollUntil(5*time.Second, func() (bool, error) {
```
mfojtik, line 90: i would retry longer

cmd/watch-termination/main.go
sttts: fixed

cmd/watch-termination/main.go
mfojtik: if it is a pointer it can be nil :-) I don't trust godoc.

sttts: It is the contract from the golang standard library. It's only a pointer because it is a reference to the actual value of that flag. Unfortunately, Golang has no concept of a non-nil pointer (i.e. a reference).

cmd/watch-termination/main.go
sttts: fixed

cmd/watch-termination/main.go
sttts: fixed

cmd/watch-termination/main.go
sttts: it retries infinitely, but only every 5s.

PR #25194, opened 2020-06-23
UPSTREAM: <carry>: Bug 1850149: Include / prefix in the instance ID output

Fedosin:
When we want to read instance ID from the metadata service, cloud provider doesn't include "/" prefix, which is required for successful parsing of provider ID later.
This commit adds the missing "/" prefix to the output.

### DIFF
--- a/deps.diff
+++ b/deps.diff
@@ -1,0 +1,5 @@
+diff --no-dereference -N -r current/vendor/k8s.io/legacy-cloud-providers/openstack/openstack_instances.go updated/vendor/k8s.io/legacy-cloud-providers/openstack/openstack_instances.go
+260c260
+< 		return "/" + md.UUID, nil
+---
+> 		return md.UUID, nil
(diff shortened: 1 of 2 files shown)


mandre:
/hold

I'd like to know what is the impact of this change. Without CI that will be impossible to tell.
Granted, this is supposed to fix https://bugzilla.redhat.com/show_bug.cgi?id=1850149 but this might be breaking other places, literally anything that was calling the `InstanceID` method.

jsafrane:
Why is this not fixed upstream?
/hold

Fedosin:
@jsafrane hi! this happens in the patch that hasn't been merged in upstream yet: https://github.com/kubernetes/kubernetes/pull/89885/files#diff-8b06269a7d6e10f5ed131fbb0b4082e7R224

sdodson:
@Fedosin Can we bump upstream along?

Fedosin:
Fixed by: https://github.com/openshift/kubernetes/pull/343

PR #25115, opened 2020-06-15
Measure container restarts over an interval instead of absolute

smarterclayton:
We run these tests after disruptive events where restarts are
necessary. Instead of looking at absolute restart count, capture
the number of initial restarts and then compare subsequent runs
to ensure the count is smaller. We double the time interval and
halve the number of failing restarts in order to ensure we capture
slow restart failures (full backoff is 5m, but we can catch slow
but not limited backoff).

PR #25202, opened 2020-06-24
Bug 1850719: Measure container restarts over an interval instead of absolute

hexfusion:
Manual cherry-pick of https://github.com/openshift/origin/pull/25115 into 4.4


hexfusion:
passes locally..
/test verify

hexfusion:
cc @smarterclayton @bparees 

unblocking disruptive test for 4.4

bparees:
i fixed up the BZ deps but i think this will still be stuck on the 4.5 BZ verification:

/bugzilla refresh

bparees:
@hexfusion if this test is failing in 4.4, the severity should be higher.  if it's not failing in 4.4, then you're probably going to get pushback from the patch manager about the need to backport this.

/lgtm
/approve

PR #25203, opened 2020-06-24
Pick scheduler v1beta1 componentconfig API

damemi:
This is attempting to pull the changes from https://github.com/kubernetes/kubernetes/commit/852442c0ff820823d21b1afa2b5b5e3896947ce3 to make the v1beta1 scheduler config API available in order to merge https://github.com/openshift/cluster-kube-scheduler-operator/pull/255, which is a blocker


damemi:
I've tested this locally and it works with our existing KSO, as well as with the changes in https://github.com/openshift/cluster-kube-scheduler-operator/pull/255. However this also requires the changes from https://github.com/kubernetes/kubernetes/pull/92531 (added in a new commit here) in order to pass ComponentConfig settings along with the Policy API. Without those changes, we will either a) lose support for the Beta features in CC or b) lose support for the Policy config, because they cannot currently be configured side-by-side.

damemi:
/priority important-soon

not having this api available will block the 1.19 rebase for the scheduler

PR #25204, opened 2020-06-24
Bug 1828477: Add CI test to check for critical alerts post upgrade

jottofar:
cherry-pick https://github.com/openshift/origin/pull/24786/commits/3b8cb3ca9b57e17980aa321bb8402ac9c144a17e

After an upgrade is successful, there should not be any critical
alerts on the cluster. Post upgrade success, this change delays for
alertCheckSleepMinutes and then checks if any critical alerts occurred
within the last alertPeriodCheckMinutes and if so the test fails.

After a successful check that confirms no critical alerts have occurred
post cluster upgrade, this change will perform another Prometheus
query checking for the occurrance of Watchdog alerts thereby verifying
Prometheus is indeed up and running correctly.

Prometheus helper package created to contain common code used to query
Prometheus. The code being moved to the helper package is currently
used by the upgrade Alert test and by the Prometheus extended test
suite. The upgrade Alert test has been modified to use the helper
package.

Reworked to maintain a single framework per test

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -1,0 +1,101 @@
+package alert
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	g "github.com/onsi/ginkgo"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	helper "github.com/openshift/origin/test/extended/util/prometheus"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


wking:
In case it helps with review, [here](https://github.com/openshift/origin/files/4828200/backport.patch.txt) is a comparison of this backport with the original.  Or generate it yourself with:

```console
$ diff -U1 --label master --label backport <(git log -1 -p --format="%B" origin/pr/24786^) <(git log -1 -p --format="%B" origin/pr/25204)
```

PR #25231, opened 2020-07-01
Bug 1851623: e2e: extended: re-enable topology manager tests

fromanirh:
The e2e Topology Manager tests require a considerable amount of hardware
resource to run, in terms of cores and possibly extra devices (SRIOV).
This is done on purpose, to try to reproduce production environment as
close as possible.

For this reason the e2e tests have some safety check: each of them
runs a check to verify if any worker node have enough resources to possibly
run the test, skipping (or failing depending on the value of an
environment variable, defaulting on skip) to avoid false negatives.

Unfortunately this was true only for the  e2e tests meant to exercise
the Topology Manager, and not for the non-regression tests we added.

Non-regression tests didn't have this check because the assumption was
that they require so little resources that we deemed this check
unnecessary.

Turns out the assumption was wrong, as demonstrated in bug
https://bugzilla.redhat.com/show_bug.cgi?id=1851623

To fix this behaviour this change adds the preliminary resource check
to all the topology manager tests, including the trivial non-regression
ones.


adambkaplan:
@fromanirh please include the BZ in the title of this PR as `Bug xxxx: <title>`. This will trigger our automation to link this PR to the issue.

Looking at the BZ, it appears that the standard worker node for the `aws-serial` suite may have changed to only have 2 CPU cores.

fromanirh:
> @fromanirh please include the BZ in the title of this PR as `Bug xxxx: <title>`. This will trigger our automation to link this PR to the issue.
> 
> Looking at the BZ, it appears that the standard worker node for the `aws-serial` suite may have changed to only have 2 CPU cores.

Sure, renamed.

Regarding CI, the `aws-serial` wants to run the serial test suite against the default aws flavour, which is either `m4.large` or `m5.large`. Per https://aws.amazon.com/ec2/instance-types/ this indeed means 2 vcpus only.
I'm not clear why the test run before (please note test requirement didn't change since the initial merge) or how the worker config did look like before, but the current behaviour is AFAICT this is the intended behaviour, so those are the resources the e2e tests will have to work with.

This PR wants to make sure the e2e Topology Manager test don't raise false negatives. The tests they will just skip if the worker resources aren't enough. Of course this gives us no real signal from the `aws-serial` lane for topology manager. We will engage to see how we can create another lane to test Topology Manager, we started this conversation some time ago but it didn't progress.

adambkaplan:
/approve

This PR is restoring the single-core test in `aws-serial`, which is better than no signal.

PR #25233, opened 2020-07-01
test/extended/router: Expect 5 OKs for router ready

Miciah:
Change `waitForRouterOKResponseExec` to expect to receive 5 consecutive HTTP 200 OK responses from the router before reporting that it is OK.

Some tests use the `waitForRouterOKResponseExec` function to wait for specific routes to be active, and these tests can fail if the function reports a false positive, meaning the function reports a route is OK when a subsequent request for that route results in an HTTP 503 response.

This change is needed to avoid these false positives when the host to which `waitForRouterOKResponseExec` sends the request is actually a load balancer in front of two or more router pods.  In this scenario, the LB might forward the first request to a router pod that returned a 200 response, and `waitForRouterOKResponseExec` would then report that the router were ready.  However, a subsequent request to the LB might be forwarded to a different router pod that would return a 503 response.

* `test/extended/router/scoped.go` (`waitForRouterOKResponseExec`): Count HTTP 200 OK responses and only report success after receiving 5 consecutive such responses.

### DIFF
--- a/test/extended/router/scoped.go
+++ b/test/extended/router/scoped.go
@@ -242,13 +242,17 @@ func waitForRouterOKResponseExec(ns, execPodName, url, host string, timeoutSecon
 	cmd := fmt.Sprintf(`
 		set -e
 		STOP=$(($(date '+%%s') + %d))
+		ok_responses=0
 		while [ $(date '+%%s') -lt $STOP ]; do
 			rc=0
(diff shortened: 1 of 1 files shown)


Miciah:
/close
Superseded by #25835.

smarterclayton:
Lol great minds.  I didn't realize there was another one (but I remember us discussing this at some point a long time ago)

PR #25236, opened 2020-07-02
Bug 1873961: e2e/apiserver: check for LateConnections+NonGracefulTermination events

sttts:
These are produced by kube-apiserver when
- non-localhost clients connect in the last 20% of the termination period. This is a strong hint that the LBs are misconfigured or misbehaving => LateConnections
- the kube-apiserver is not cleanly terminating, but e.g. is receiving a SIGKILL instead of an early SIGTERM => NonGracefulTermination

### DIFF
--- a/test/extended/apiserver/graceful_termination.go
+++ b/test/extended/apiserver/graceful_termination.go
@@ -1,0 +1,63 @@
+package apiserver
+
+import (
+	"context"
+	"fmt"
+
+	g "github.com/onsi/ginkgo"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Late]", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLI("terminating-kube-apiserver")
+
+	g.It("kubelet terminates kube-apiserver gracefully", func() {
+		t := g.GinkgoT()
+
+		client, err := kubernetes.NewForConfig(oc.AdminConfig())
+		if err != nil {
+			g.Fail(fmt.Sprintf("Unexpected error: %v", err))
+		}
+
+		evs, err := client.CoreV1().Events("openshift-kube-apiserver").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			g.Fail(fmt.Sprintf("Unexpected error: %v", err))
+		}
+
+		for _, ev := range evs.Items {
+			if ev.Reason != "NonGracefulTermination" {
+				continue
+			}
+
+			t.Errorf("kube-apiserver reports a non-graceful termination: %#v. Probably kubelet or CRI-O is not giving the time to cleanly shut down. This can lead to connection refused and network I/O timeout errors in other components.", ev)
+		}
+	})
+
+	g.It("API LBs follow /readyz of kube-apiserver and stop sending requests", func() {
+		t := g.GinkgoT()
+
+		client, err := kubernetes.NewForConfig(oc.AdminConfig())
+		if err != nil {
+			g.Fail(fmt.Sprintf("Unexpected error: %v", err))
+		}
+
+		evs, err := client.CoreV1().Events("openshift-kube-apiserver").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			g.Fail(fmt.Sprintf("Unexpected error: %v", err))
+		}
+
+		for _, ev := range evs.Items {
+			if ev.Reason != "LateConnections" {
+				continue
+			}
+
+			t.Errorf("API LBs send requests to kube-apiserver far too late in termination process, probably due to broken LB configuration: %#v. This can lead to connection refused and network I/O timeout errors in other components.", ev)
+		}
+	})
+})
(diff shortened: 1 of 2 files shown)


smarterclayton:
If this is a platform semantic (no late events) we now have an event invariant at a suite level may be appropriate.

sttts:
> If this is a platform semantic (no late events) we now have an event invariant at a suite level may be appropriate.

Any link what an event invariant at a suite level means? Where are those defined?

test/extended/util/disruption/controlplane/termination.go
```diff
@@ -0,0 +1,120 @@
...
+		<-done
+	}()
+
+	m := EventMonitor{
+		Namespace: "openshift-kube-apiserver",
+		Filter: func(ev *corev1.Event) bool {
+			return ev.Reason == "LateConnections"
+		},
+		Handler: func(f *framework.Framework, ev *corev1.Event) {
+			framework.Failf("kube-apiserver got connections late during termination, probably due to broken LB configuration: %#v", ev)
```
jsafrane, line 78: It could be worth adding what consequences it could have on running tests (connection refused?) and what to do with this failure (Assign to API server? Networking? Blame the cloud and ignore?)

smarterclayton:
If your invariant holds for all possible tests (upgrade, disruption, etc), then we would have it be generic.  If you think there are some disruption tests where it would be expected for this to fire normally, we should know that now before we use something generic.

There's a new post suite invariant checker that says "if an event of this type occurs, fail the test".  It's "AdditionalJUnitFn" on the test suite.  We will be using it for the "mirrored offlien images", so we may need to reap it.

test/extended/util/disruption/controlplane/termination.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (m *EventMonitor) Run(ctx context.Context, f *framework.Framework) {
+	config, err := framework.LoadConfig()
+	framework.ExpectNoError(err)
+
+	c, err := kubernetes.NewForConfig(config)
+	framework.ExpectNoError(err)
+
+	lw := &cache.ListWatch{
```
smarterclayton, line 99: Is there really an advantage over this with just running "get all events, filter to see if they match" at the end of the test (which is both simpler and can be easily copied into a `[Late]` test)?  Upgrades take less than 2hr normally, so events aren't going to TTL.  If we lose the DB, we have bigger problems.  The events won't be in namespaces that get deleted.

test/extended/util/disruption/controlplane/termination.go
```diff
@@ -0,0 +1,120 @@
+package controlplane
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+
```
deads2k, line 7: heh. I know I don't like commenting on imports, but come on man.... :)

test/extended/util/disruption/controlplane/termination.go
```diff
@@ -0,0 +1,120 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/upgrades"
+)
+
+// GracefulTerminationTest watches events of non-graceful terminations of kube-apiserver.
+type GracefulTerminationTest struct{}
+
+func (GracefulTerminationTest) Name() string { return "kube-apiserver-terminates-gracefully" }
+func (GracefulTerminationTest) DisplayName() string {
+	return "Kubernetes API server terminates gracefully"
```
deads2k, line 31: `[sig-something]` prefix.  Decide if you think it's kube-apiserver or network.

test/extended/util/disruption/controlplane/termination.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// LateConnectionsTest watches events late connections to the kube-apiserver during termination. These usually
+// mean that the load balancer do not configure their health probes correctly, by calling /readyz of the API
+// server (not /healthz !)
+type LateConnectionsTest struct{}
+
+func (LateConnectionsTest) Name() string { return "kube-apiserver-no-late-connections" }
+func (LateConnectionsTest) DisplayName() string {
+	return "LBs stop connecting to kube-apiserver during termination in time"
```
deads2k, line 61: `[sig-something]` here too

test/extended/util/disruption/controlplane/termination.go
sttts: wow, Goland likes empty lines.

test/extended/util/disruption/controlplane/termination.go
sttts: Unfortunately this can have many reasons, from ourselves, to node, to cri-o.

test/extended/util/disruption/controlplane/termination.go
sttts: :+1:

added

test/extended/util/disruption/controlplane/termination.go
sttts: Moved into `[Late]` test.

sttts:
> If this is a platform semantic (no late events) we now have an event invariant at a suite level may be appropriate.

It is an invariant for non-node-disruptive situation. If a node is switched off hard, the invariant will break. I don't know whether we have such a test.

sttts:
```
Operator degraded (IngressControllersDegraded): Some ingresscontrollers are degraded: default | Operator degraded 
```

blocking all e2e tests. cc @mfojtik @deads2k @ TRT

/retest

sttts:
Ingress 😞

/retest

sttts:
Ingress, again.

/retest

PR #25241, opened 2020-07-06
upgrade tests: extends StartOpenShiftAPIMonitoring to monitor PATCH requests

p0lyn0mial:
It seems that retries are a popular way of increasing the high availability of the system in the face of transient and random errors. However, knowing which requests are worth retiring is hard.  Some requests are simply not safe because they might have side effects, others might be retried by different components (client vs server).

In order to understand the impact on availability under a disruption, we decided to monitor `PATCH` requests to the server to see if/how they differ from `GET` requests.

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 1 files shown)


smarterclayton:
/hold

I'm confused about the why here, can we describe?

smarterclayton:
Please add a lot more justification to the commit message and PR description :) so I can review

deads2k:
@p0lyn0mial more concretely, I suspect a new test with a new tolerance will be more welcome.

@smarterclayton patches are not guaranteed indepotent because some patches can add elements to arrays, meaning you could add more elements than expected.  Therefore they will never generically be retried.

smarterclayton:
Sure, but you can patch a field instead of an array.

smarterclayton:
All patches are retried and should be retried that have preconditions where the server knows that you failed.

smarterclayton:
And yes, adding a new check type is better.

p0lyn0mial:
@smarterclayton sorry, the commit message and PR description updated. PTAL.

pkg/monitor/api.go
```diff
@@ -91,16 +92,47 @@ func StartKubeAPIMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest
...
+
+func StartOpenShiftAPIMonitoringWithGET(ctx context.Context, m *Monitor, clusterConfig *rest.Config, timeout time.Duration) error {
 	pollingConfig := *clusterConfig
 	pollingConfig.Timeout = timeout
 	pollingClient, err := clientimagev1.NewForConfig(&pollingConfig)
 	if err != nil {
 		return err
 	}
 
+	return startOpenShiftAPIMonitoringFor(ctx, m, "GET", func() error {
```
deads2k, line 112: I think we'll need to split for another case.  This test is conflating
1. openshift-apiserver is running
2. kube-apiserver can reach openshift-apiserver over network
3. openshift-apiserver can contact etcd

Each different case requires different debugging and implies different problems and inspection by different teams.  We can at least distinguish between 2 and 3 from outside the cluster and we should attempt to do so.  You can do this by hitting the discovery endpoint with GET and PATCH in addition to hitting am imagestream.

deads2k:
looks like a start in the right direction to me.

/lgtm

p0lyn0mial:
@sttts FYI

PR #25250, opened 2020-07-07
Bug 1854311: 4.5: UPSTREAM: 91307: CSI: Modify VolumeAttachment check to use Informer/Cache

jsafrane:
Backported version of https://github.com/openshift/origin/pull/25109
@openshift/storage

PR #25109, opened 2020-06-15
Bug 1839933: UPSTREAM: 91307: CSI: Modify VolumeAttachment check to use Informer/Cache

jsafrane:
Fix slow attach in OCP >= 4.4.

https://bugzilla.redhat.com/show_bug.cgi?id=1839933
https://github.com/kubernetes/kubernetes/pull/91307
cc @openshift/storage

PR #25251, opened 2020-07-07
Bug 1854312: 4.4: UPSTREAM: 92014: CSI: Modify VolumeAttachment check to use Informer/cache

jsafrane:
This is 4.4 variant of https://github.com/openshift/origin/pull/25109 and https://github.com/openshift/origin/pull/25250. This time from upstream 1.17 branch, https://github.com/kubernetes/kubernetes/pull/92014


jsafrane:
cc @openshift/storage

humblec:
lgtm from end.

PR #25257, opened 2020-07-08
switches origin to 4.6 kubernetes 1.18.3

p0lyn0mial:

PR #24607, opened 2020-02-27
UPSTREAM: 88120: add dynamic certificate reloading for kube aggregator

p0lyn0mial:
not clean, needs review.

based on https://github.com/openshift/kubernetes/commit/adfea731ab01ae9552b8aa309d6bf9f1c97656fc#diff-40fbc204fb430ee43a18f381a4e2182e

PR #25258, opened 2020-07-08
wires the egress proxy to the kube-aggregator 

p0lyn0mial:
brings back the egress proxy configuration to the kube-aggregator lost during previous backporting.

it requires backporting at least to 4.5 and 4.4


p0lyn0mial:
/hold

until https://github.com/openshift/origin/pull/25257 lands

p0lyn0mial:
I think we lost it here https://github.com/openshift/origin/pull/24607

PR #25045, opened 2020-05-29
Bug 1835371: router/metrics: Fix haproxy_server_max_sessions test

frobware:
This assertion currently flakes a lot. (I believe) the flake occurs
when the router reloads and the GET for the latest metrics reads
zero's because there has been no updates.

It may be better to mark this test as serial. This commit retries the
assertion as opposed to dealing with it as a one-shot.

The expectation is that haxproxy_server_max_sessions needs to be
greater than zero over two successive scrapes once the first scrape
shows an increase. However, this assertion can still fail depending on
when the metrics are refreshed and when the router reloads, which is
every 5s. In my testing I have seen this to be more resilient but
there are periods when running the full openshift/conformance tests
that this continues to fail.

PR #21957, opened 2019-02-04
UPSTREAM: 73404: Kubelet: Fix volumemanager test race

wongma7:
this will fix flakes like https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/batch/pull-ci-openshift-origin-master-unit/2571#githubcomopenshiftoriginvendork8siokubernetespkgkubeletvolumemanagerreconciler-test_run_positive_volumemountcontrollerattachenabledrace

@tsmetana

PR #25270, opened 2020-07-10
FIx CI for 3.11

soltysh:
/hold

### DIFF
--- a/test/extended/router/metrics.go
+++ b/test/extended/router/metrics.go
@@ -209,7 +209,7 @@ var _ = g.Describe("[Conformance][Area:Networking][Feature:Router]", func() {
 				}
 
 				maxSessionsInvariant := func(v []float64) bool {
-					return len(v) >= 2 && v[0] > 0 && v[1] > 0
+					return len(v) == 1 && v[0] > 0
 				}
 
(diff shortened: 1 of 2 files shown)


sdodson:
Hmm, nevermind, I went looking for 3.11 PRs failing on e2e-gcp to test the fix in https://github.com/openshift/openshift-ansible/pull/12203
But this is not a good PR to test since this actually attempts to change things such that the test works.

soltysh:
Yeah, I used it to proof the initial problem @sdodson. Let's see if this now works w/o the hacks.

soltysh:
It looks like the only failure in this PR was caused by HAProxy test, which was fixed in https://github.com/openshift/origin/pull/25045 I've picked that here and let's see where we get with it.
/cc @frobware 
since you authored the original PR

frobware:
> It looks like the only failure in this PR was caused by HAProxy test, which was fixed in #25045 I've picked that here and let's see where we get with it.
> /cc @frobware
> since you authored the original PR

LGTM, though not sure I should be approving my commits...

soltysh:
Yeah, that particular unit flaked a lot and was fixed in https://github.com/openshift/origin/pull/21957 so I'm picking that one too here.

PR #25276, opened 2020-07-13
Bug 1874079: Do not harcode arch for fedora container image

### LINKED ISSUE #25275: [sig-cluster-lifecycle] uses hardcoded arch and fails in non-x86 deployments
https://github.com/openshift/origin/blob/master/test/extended/csrapprover/csrapprover.go#L48 harcodes the architecture of the container image, causing `exec format error` in non-x86 deployments. This was introduced in commit 0dc769a1166cdda413f92b2eacafedad94b375c8.

##### Version
I've first observed this error in 4.5 [1] but it should affect all versions where the above commit is present.

[1] https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/9039/rehearse-9039-release-openshift-origin-installer-e2e-remote-libvirt-s390x-4.5/1282676041990541312 

##### Steps To Reproduce
1. Run the sig-cluster-lifecycle test in any non-x86_64 architecture.

##### Current Result
The test fails because trying to run the container results in `exec format error`

##### Expected Result
The respective image for the architecture is used instead of the hardcoded one.

r4f4:
This test was failing on non-x86 images because of the hardcoded
architecture

Fixes #25275 

### DIFF
--- a/test/extended/csrapprover/csrapprover.go
+++ b/test/extended/csrapprover/csrapprover.go
@@ -45,7 +45,7 @@ var _ = g.Describe("[sig-cluster-lifecycle]", func() {
 		// the /config/master API port+endpoint is only visible from inside the cluster
 		// (-> we need to create a pod to try to reach it) and contains the token
 		// of the node-bootstrapper SA, so no random pods should be able to see it
-		pod, err := exutil.NewPodExecutor(oc, "get-bootstrap-creds", "quay.io/fedora/fedora:32-x86_64")
+		pod, err := exutil.NewPodExecutor(oc, "get-bootstrap-creds", "registry.fedoraproject.org/fedora:32")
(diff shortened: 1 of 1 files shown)


jaypoulz:
This may need to be backported to older branches once landed.

jaypoulz:
@ironcladlou can we get an approval for this - this is currently breaking 4 out of 6 multi-arch jobs

r4f4:
Another option here, if using the fedora registry is not an option, is to get whoever created `quay.io/fedora/` do add a manifest list.

yselkowitz:
@r4f4 a 4.6 BZ is required to get this in now, can you file it and retitle this PR accordingly?

r4f4:
The registry has been changed to `docker.io/fedora:32`, making this PR obsolete.

PR #25277, opened 2020-07-14
patch net/http2 library to fix the graceful shutdown of HTTP2 connections

p0lyn0mial:
this patch fixes `net/http2` library to properly manage the connection state,
this allow to proper shutdown sequence to occur.
as a result `Server.Shutdown` will wait for the active connections to finish before shutting down the HTTP server

`4d8de76-1.13.patch` was prepared based on https://go-review.googlesource.com/c/go/+/240278
the original patch targets the master branch and didn't apply to golang 1.13.

### DIFF
--- a/images/hyperkube/Dockerfile.rhel
+++ b/images/hyperkube/Dockerfile.rhel
@@ -1,6 +1,10 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
 WORKDIR /go/src/github.com/openshift/origin
 COPY . .
+# apply a patch to net/http2 library that fixes the graceful termination of active connections, more details at: https://github.com/golang/go/issues/39776
+# this patch was prepared based on https://go-review.googlesource.com/c/go/+/240278/. The original patch is targeting the master branch and didn't apply to 1.13 version
+RUN patch -d /usr/local/go -p1 < images/hyperkube/4d8de76-1.13.patch
+RUN go test -v --race images/hyperkube/gc_test.go
 RUN for p in vendor/k8s.io/kubernetes/cmd/kube-apiserver vendor/k8s.io/kubernetes/cmd/kube-controller-manager \
     vendor/k8s.io/kubernetes/cmd/kube-scheduler vendor/k8s.io/kubernetes/cmd/kubelet cmd/watch-termination; do make build WHAT=$p; done && \
     mkdir -p /tmp/build && \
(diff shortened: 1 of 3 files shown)


sttts:
Sgtm.

images/hyperkube/Dockerfile.rhel
```diff
@@ -1,6 +1,10 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
 WORKDIR /go/src/github.com/openshift/origin
 COPY . .
+# apply a patch to net/http2 library that fixes the graceful termination of active connections, more details at: https://github.com/golang/go/issues/39776
+# this patch was prepared based on https://go-review.googlesource.com/c/go/+/240278/. The original patch is targeting the master branch and didn't apply to 1.13 version
+RUN patch -d /usr/local/go -p1 < images/hyperkube/4d8de76-1.13.patch
+RUN go test images/hyperkube/gc_test.go
```
tnozicka, line 7: Would it be better to locate the file in go sources (pkg/ or test/) then put it into images?

Also why does it need to be run separately and not in unit job?

p0lyn0mial: My intention was to add it close to the patch. The patch and the tests are temporary.
The patch can be removed when the new version of `go` will be released. Once we have that then I am going to add an integration test for testing graceful shutdown to k/k.

sttts: IMO having it here makes sense. This is really a smoke test for the build image. It is not part of our product.

sttts: doesn't it have to be ./images/hyperkube.... ?

images/hyperkube/Dockerfile.rhel
```diff
@@ -1,6 +1,10 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
 WORKDIR /go/src/github.com/openshift/origin
 COPY . .
+# apply a patch to net/http2 library that fixes the graceful termination of active connections, more details at: https://github.com/golang/go/issues/39776
+# this patch was prepared based on https://go-review.googlesource.com/c/go/+/240278/. The original patch is targeting the master branch and didn't apply to 1.13 version
+RUN patch -d /usr/local/go -p1 < images/hyperkube/4d8de76-1.13.patch
```
tnozicka, line 6: Patching is this way seem a bit fragile given the builder images are not .z versioned. I personally think it would be better to fork the net package, patch it, vendor it back and add a simple patch for the server creation.

sttts: > the builder images are not .z versioned

This means what?

PR #25282, opened 2020-07-15
e2e/mustgather: ignore non audit files in /var/log/*-apiserver dir

sttts:
We also have termination logs in the /var/log/*-apiserver directories. These don't have events inside, obviously.

Fixes e2e tests in https://github.com/openshift/cluster-kube-apiserver-operator/pull/876.

### DIFF
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -146,6 +146,21 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 					return nil
 				}
 
+				// TODO: remove when https://github.com/openshift/must-gather/pull/162 has merged
+				if strings.HasSuffix(path, ".terminating.gz") {
+					return nil
+				}
+
+				if (strings.Contains(path, "-termination-") && strings.HasSuffix(path, ".log.gz")) || strings.HasSuffix(path, "termination.log.gz") {
+					// these are expected, but have unstructured log format
+					return nil
+				}
+
+				isAuditFile := (strings.Contains(path, "-audit-") && strings.HasSuffix(path, ".log.gz")) || strings.HasSuffix(path, "audit.log.gz")
+				o.Expect(isAuditFile).To(o.BeTrue())
+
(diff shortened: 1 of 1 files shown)


test/extended/cli/mustgather.go
```diff
@@ -146,6 +146,19 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 					return nil
 				}
 
+				// TODO: remove when https://github.com/openshift/must-gather/pull/162 has merged
+				if !strings.HasSuffix(path, ".terminating.gz") {
```
p0lyn0mial, line 150: it should resolve to `true` right?

test/extended/cli/mustgather.go
```diff
@@ -146,6 +146,19 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+					return nil
+				}
+
+				if !strings.Contains(path, "-termination-") && !strings.HasSuffix(path, "termination.log.gz") {
```
p0lyn0mial, line 154: same here, right ?

test/extended/cli/mustgather.go
```diff
@@ -146,6 +146,19 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+					// these are expected, but have unstructured log format
+					return nil
+				}
+
+				isAuditFile := strings.Contains(path, "-audit-") || strings.HasSuffix(path, "audit.log.gz")
```
p0lyn0mial, line 159: at this point, we expect only `audit` compliant files.

test/extended/cli/mustgather.go
sttts: sure

sttts: fixed

test/extended/cli/mustgather.go
sttts: fixed

test/extended/cli/mustgather.go
sttts: added comment

PR #24680, opened 2020-03-12
test/extended/cli/mustgather: Separate gather_audit_logs test

wking:
openshift/must-gather#143 is removing these from the default gather, because they're mostly useful for internal debugging, less useful in end-user bug reports, and can run to hundreds of megabytes.  But we still want to ensure that they work as expected when they are explicitly requested.  This commit pulls the audit-log checks out of the test-case for the generic invocation.  And it adds a new test case with those checks after an explict `gather_audit_logs` request.

CC @deads2k, @sanchezl, @sferich888, @soltysh

[1]: https://github.com/openshift/must-gather/pull/143

PR #25283, opened 2020-07-15
test/extended/cli/mustgather: Separate gather_audit_logs test

soltysh:
This is slightly improved and rebased version of #24680 

The main issue with the original was that it might happen that audit files can be empty, in which case `gzip.NewReader(file)` will return `EOF`. The second commit in this PR accounts for that and ignores empty audit files. 

/assign @deads2k @wking

### DIFF
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -31,9 +31,16 @@ import (
 var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLI("oc-adm-must-gather").AsAdmin()
+
+	g.JustBeforeEach(func() {
+		// wait for the default service account to be avaiable
+		err := exutil.WaitForServiceAccount(oc.KubeClient().CoreV1().ServiceAccounts(oc.Namespace()), "default")
+		o.Expect(err).NotTo(o.HaveOccurred())
+	})
+
 	g.It("runs successfully", func() {
 		tempDir, err := ioutil.TempDir("", "test.oc-adm-must-gather.")
-		o.Expect(err).ToNot(o.HaveOccurred())
+		o.Expect(err).NotTo(o.HaveOccurred())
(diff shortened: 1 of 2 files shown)


test/extended/cli/mustgather.go
```diff
@@ -179,51 +229,46 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
 				return nil
 			})
-			o.Expect(err).ToNot(o.HaveOccurred())
+			o.Expect(err).NotTo(o.HaveOccurred())
 
 			// On IBM ROKS, events will not be part of the output, since audit logs do not include
 			// control plane logs.
 			if e2e.TestContext.Provider != ibmcloud.ProviderName {
-				o.Expect(eventsChecked).To(o.BeNumerically(">", 10000))
+				o.Expect(eventsChecked).To(o.BeNumerically(">", 1000))
```
soltysh, line 248: I'm lowering this to 1000, 10k seems like a lot, and it might depend when in the tests it gets run. I have a cluster that after half a day has about 7k events. so this 1k is a good minimum to start with.

wking:
Anything that gets these decoupled looks good to me :)

test/extended/cli/mustgather.go
```diff
@@ -165,11 +206,20 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
-				o.Expect(err).ToNot(o.HaveOccurred())
+				o.Expect(err).NotTo(o.HaveOccurred())
 				defer file.Close()
 
+				fi, err := file.Stat()
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				// it will happen that the audit files are sometimes empty, we can
+				// safely ignore these files since they don't provide valuable information
+				if fi.Size() == 0 {
```
deads2k, line 217: make sure at least one has content.

deads2k:
looks plausible.  Found the number of items check

/lgtm

PR #25285, opened 2020-07-15
Bug 1857082: Include pod /etc/hosts in ephemeral storage calculation for eviction

joelsmith:
Backport of https://github.com/kubernetes/kubernetes/issues/93032

PR #25288, opened 2020-07-15
Bug 1857080: Include pod /etc/hosts in ephemeral storage calculation for eviction

joelsmith:
Cherry-pick of #25285


sdodson:
The upstream change has been merged, overriding bugzilla/valid-bug since this already had cherry-pick-approved.

PR #25290, opened 2020-07-16
UPSTREAM: <carry>: apiserver: skip local IPs and probes for LateConnections – fix race

sttts:
In unit tests we create multiple apiservers in parallel, leading to a race writing the global variable.


sttts:
Infra.

/retest

PR #25291, opened 2020-07-16
add test confirming service network functions from openshift-apiserver pod

deads2k:
This adds checks to be sure we don't experience outages on the service network. We noticed that there were "connection refused" errors from the openshift-apiserver to the kube-apiserver *only* through the service network.  The direct access via the node IP (kube-apiserver are host-network) continued without any failures.

/hold

Need to hold until the openshift-apiserver network checks land. The choice of Fail versus Flake will be determined based on failure rates.


test/extended/networking/service_network_consistent.go
```diff
@@ -0,0 +1,60 @@
...
+
+	g.It("shouldn't report outage to kubernetes-service", func() {
+		ctx := context.TODO()
+		endpointCheckClient := v1alpha1.NewForConfigOrDie(oc.AdminConfig())
+		connectivityChecks, err := endpointCheckClient.PodNetworkConnectivityChecks("openshift-apiserver").List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		failures := []string{}
+		for _, check := range connectivityChecks.Items {
+			if !strings.Contains(check.Name, "kube-service") {
```
sanchezl, line 29:
```suggestion
			if !strings.Contains(check.Name, "kubernetes-service") {
```

test/extended/networking/service_network_consistent.go
```diff
@@ -0,0 +1,60 @@
...
+				continue
+			}
+			for _, outage := range check.Status.Outages {
+				failures = append(failures, fmt.Sprintf("%#v", outage))
+			}
+		}
+		if len(failures) > 0 {
+			g.Fail(fmt.Sprintf("the KUBERNETES_SERVICE_HOST:KUBERNETES_SERVICE_PORT was inaccessible via the service network IP (compare against kube-apiserver direct endpoint access):\n%v", strings.Join(failures, "\n")))
```
danwinship, line 37: `KUBERNETES_SERVICE_HOST` *is* `the service network IP`. (ie, this is basically saying "172.30.0.1 was inaccessible via 172.30.0.1").

Also, a random developer looking at test failures will have no idea what "(compare against kube-apiserver direct endpoint access)" is supposed to mean. Is that an instruction to the reader? If so, where is the reader supposed to get that other information, and why can't the test case make the comparison itself?

Also, you can use `g.Errorf(...)` instead of `g.Fail(fmt.Sprintf(...))`

test/extended/networking/service_network_consistent.go
```diff
@@ -0,0 +1,60 @@
...
+		for _, check := range connectivityChecks.Items {
+			if !strings.Contains(check.Name, "kube-apiserver-service") {
+				continue
+			}
+			for _, outage := range check.Status.Outages {
+				failures = append(failures, fmt.Sprintf("%#v", outage))
+			}
+		}
+		if len(failures) > 0 {
+			g.Fail(fmt.Sprintf("the `oc -n openshift-kube-apiserver get services/apiserver` was inaccessible via the service network IP (compare against kube-apiserver direct endpoint access):\n%v", strings.Join(failures, "\n")))
```
danwinship, line 57: It would not be clear to me from the output of these two tests what the difference is between them is (what does `-n openshift-kube-apiserver services/apiserver` point to, if not the kube apiserver?)

deads2k: One is maintained by the kube-apiserver directly. The kube-apiserver directly writes into the endpoints resource.  This is 172.30.0.1.

The other is used by the service monitor.  It is a real service maintained by the service/endpoints controller.  This has a different IP.

test/extended/networking/service_network_consistent.go
deads2k: > Also, a random developer looking at test failures will have no idea what "(compare against kube-apiserver direct endpoint access)" is supposed to mean. Is that an instruction to the reader? If so, where is the reader supposed to get that other information, and why can't the test case make the comparison itself?

I'll see about adding more information. Basically we are seeing cases today where direct access to a kube-apiserver via the node IP works fine.  But access via 172.30.0.1 failed.  So we know the kube-apiservers (all of them) are accepting connections, but 172.30.0.1 shows repeated "connection refused"

deads2k: so I want to be sure no one bounces these as "well the kube-apiserver is returning connection refused".  In every case we've seen so far, the kube-apiserver is functioning, but cannot be accessed via the service network on a node that is reporting the network as ready.

deads2k:
now has checks on the actual kube-apiserver endpoints to avoid false positives.

/hold cancel

deads2k:
The test ran.

/test all

deads2k:
/test all

now that Luis fixed the names

PR #25224, opened 2020-06-30
Disable hyperkube build and switch to go modules

marun:
**TODO before merge of this PR**
- [x] Merge https://github.com/openshift/release/pull/9989
- [x] Wait for publishing bot to catch up
- [x] Update vendoring to point to the latest published code in openshift/kubernetes
- [x] Update the target branch for this PR from `marun-testing` to `master`

PR #25294, opened 2020-07-17
Update test's zz_generated.annotations.go

soltysh:
Looks like this was missed in https://github.com/openshift/origin/pull/25224


soltysh:
This was already handled.
/close

PR #25295, opened 2020-07-19
add a security test to verify capabilities

yuvalk:
This will test the restricted scc is really restricted,
including only
0x000000000000051b=cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_setpcap,cap_net_bind_service

Signed-off-by: Yuval Kashtan <yuvalkashtan@gmail.com>

### DIFF
--- a/test/extended/security/scc.go
+++ b/test/extended/security/scc.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 
 	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
 	securityv1 "github.com/openshift/api/security/v1"
 	securityv1client "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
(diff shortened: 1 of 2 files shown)


smarterclayton:
Is there a human readable diff for this?  Would be better to print that if it exists so that someone who isn't familiar with the low level linux stuff can still triage.

yuvalk:
> Is there a human readable diff for this? Would be better to print that if it exists so that someone who isn't familiar with the low level linux stuff can still triage.

yeah,
good idea,
you mean beside what I provided in the commit msg I guess

I will add a call to capsh --decode on both values, to get that

umohnani8:
This should pass after https://github.com/openshift/machine-config-operator/pull/2051 gets in

umohnani8:
@smarterclayton @mrunalp this is ready to be merged, PTAL

PR #25297, opened 2020-07-20
Print error when it happens

dmage:


test/extended/util/framework.go
```diff
@@ -1654,18 +1654,18 @@ func RunOneShotCommandPod(
...
 		if getErr != nil {
-			e2e.Logf("failed to get pod %q: %v", pod.Name, err)
+			e2e.Logf("failed to get pod %q: %v", pod.Name, getErr)
 			return false, nil
 		}
 
 		if podHasErrored(cmdPod) {
-			e2e.Logf("pod %q errored trying to run the command: %v", pod.Name, err)
+			e2e.Logf("pod %q errored trying to run the command: %#+v", pod.Name, pod.Status.ContainerStatuses[0].State.Terminated)
 			return false, nil
```
dmage, line 1663: Shouldn't it return an error in this case?

dmage: The pod uses RestartPolicyOnFailure, so probably this line is ok

PR #25302, opened 2020-07-21
split upgrade into different tests so we get clear failure signals

deads2k:
/hold

this is only even close to working if the test is serial

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	o "github.com/onsi/gomega"
+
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade/upgrade.go
```diff
@@ -116,7 +118,7 @@ var _ = g.Describe("[sig-arch][Feature:ClusterUpgrade]", func() {
 	f.SkipNamespaceCreation = true
 	f.SkipPrivilegedPSPBinding = true
 
-	g.It("Cluster should remain functional during upgrade [Disruptive]", func() {
+	g.When("Clusters upgrade [Disruptive]", func() {
 		config, err := framework.LoadConfig()
```
deads2k, line 122: panic here.  ah ginkgo, what a mess.

PR #25303, opened 2020-07-22
hack: make update and make verify check for test name generation

smarterclayton:
When bindata was removed, the verification protection for test names
was dropped which would allow someone to merge changes that don't
match the names.

This restores update-generated/verify-generated as generic scripts
for anything not covered by bindata. bindata continues to be referenced
by the makefile.


smarterclayton:
@marun re our discussion yesterday

marun:
@smarterclayton @tnozicka Do we need more bespoke tooling, or could it be made general purpose and incorporated into build-machinery-go?

smarterclayton:
I don't see a need to build generic mechanisms for one thing.  Don't abstract things that don't need to be abstracted.

smarterclayton:
FIPS and cmd tests are broken on this repo (post move?).  Force merging to prevent regressions from slipping in.

cmd has debugging info added in #25307 which seems to indicate project deletion may be failing.

PR #25304, opened 2020-07-22
test: Allow test suites to define additional checks for suite failure

smarterclayton:
Future changes will test invariants reported by events, such as whether
images were pulled from outside of the current code base. Also updates
the junit reporting for the "error level events" to be reported as a
flake, and corrects the duration to include "late" tests.


smarterclayton:
FIPS and cmd are broken (due to move to o/k?)

Merging to get naming aligned and ownership for tests.

PR #25307, opened 2020-07-22
test: Report an error message when a pod fails in a cmd test

smarterclayton:
The logic for testing for pod failure lost the reason why, so no
debug output was possible. Instead, update the one shot pod to
capture fallback logs and then have the executor report that as
the error message.

E.g.

    Jul 22 20:13:43.152: INFO: pod "test-cmd" errored trying to run the command: <nil>

because err was nil in the context it was invoked.

Noticed in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25304/pull-ci-openshift-origin-master-e2e-cmd/1286015364055961600

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


smarterclayton:
Force merging so we can see the errors

PR #25308, opened 2020-07-23
etcd-storage-test: apply 1.19 diff

sttts:
With hyperkube built from openshift/kubernetes this test must work with current version and rebase version. For that we check the server version and apply the 1.19 diff:

```
git diff origin/release-1.18 origin/release-1.19 test/integration/etcd/data.go
```


mfojtik:
Merging with button as this is blocking 1.19 rebase and was tested manually using local cluster.

PR #25321, opened 2020-07-24
UPSTREAM: <drop>: temporarily skip kubectl tests which change in the k8s bump

soltysh:
/assign @damemi


soltysh:
This will conflict with k8s bump, we'll wait for that to land first.
/close

PR #25328, opened 2020-07-27
Fix "should provide named network metrics" test.

fedepaol:
The network name is now produced as the composition of namespace and the network name.
This is introduced by a change in multus https://github.com/openshift/multus-cni/commit/e7c8977423a6c221074f173a497a5de0aaa98649
so the change in the test reflects the change in the way the interface name is added to
the pod.


cgwalters:
Looks sane to me...
/lgtm

smarterclayton:
Force merging due to FIPS failures.

PR #25331, opened 2020-07-28
Bug 1860837: Re-enable prometheus named network metrics test.

fedepaol:
Now that https://github.com/openshift/origin/pull/25328 is merged, the test should be fixed.
We can remove the skip and re-enable the test.


fedepaol:
/hold 
Want to leave the test disabled to avoid introducing other variables. Nightlies are being kept back by this failing test.

PR #25314, opened 2020-07-23
Bump to kube 1.19-rc.2

marun:

PR #25335, opened 2020-07-28
bug 1861189: UPSTREAM: <drop>: make pdb tests pass reliably

deads2k:
pdbs were fixed to be more tolerant of certain classes of delete. This means old tests and new cluster break because the delete isn't constrained.  We need https://github.com/kubernetes/kubernetes/pull/92991 . I pulled the entire file in to be sure.

This will go away once we revendor, but the level in o/k is right.  If pdb tests pass, we should green button this.

### DIFF
--- a/vendor/k8s.io/kubernetes/test/e2e/apps/disruption.go
+++ b/vendor/k8s.io/kubernetes/test/e2e/apps/disruption.go
@@ -131,10 +131,12 @@ var _ = SIGDescribe("DisruptionController", func() {
 		framework.ExpectHaveKey(updated.Status.DisruptedPods, pod.Name, "Expecting the DisruptedPods have %s", pod.Name)
 
 		ginkgo.By("Patching PodDisruptionBudget status")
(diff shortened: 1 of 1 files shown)


marun:
@deads2k This will need to be proposed to openshift/kubernetes instead.

marun:
/hold 

There are more than one test failures blocking the merging of this change. Please prioritize merging https://github.com/openshift/origin/pull/25314 instead, which fixes and skips a number of issues introduced by the 1.19 rebase. This PR will then need to unskip.

deads2k:
this fixes the must-gather problem.  TRT is prioritizing unblocking the payload promotion and this resolves one of two issues.  https://github.com/openshift/origin/pull/25335 / https://github.com/openshift/origin/pull/25336 resolves the other.  We will merge these

sttts:
This broke origin merges 🎉 Do not button merge ignoring verify scripts 😞

bparees:
> This broke origin merges tada Do not button merge ignoring verify scripts

this was a deliberate choice as it unblocked more than it broke.

PR #25336, opened 2020-07-28
bug 1861201: make must-gather test resilient to failures and disk timing

deads2k:
Apparently it takes a while for logs to appear, so we retry on a loop to avoid that.
The openshift-apiserver will have many fewer requests than the kube-apiserver, so have two different counts.
Update the error message to more clearly indicate which audit log is broken for better debugging in the future.

### DIFF
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -15,17 +15,15 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	oauthv1 "github.com/openshift/api/oauth/v1"
 	"github.com/openshift/client-go/image/clientset/versioned"
 	oauthv1client "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
 	"github.com/openshift/library-go/pkg/image/imageutil"
-	e2e "k8s.io/kubernetes/test/e2e/framework"
-
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/ibmcloud"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
 )
 
 var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
@@ -126,6 +124,11 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 	})
 
 	g.It("runs successfully for audit logs", func() {
+		// On IBM ROKS, events will not be part of the output, since audit logs do not include control plane logs.
+		if e2e.TestContext.Provider == ibmcloud.ProviderName {
+			g.Skip("ROKs doesn't have audit logs")
+		}
+
 		// makes some tokens that should not show in the audit logs
 		const tokenName = "must-gather-audit-logs-token-plus-some-padding-here-to-make-the-limit"
 		oauthClient := oauthv1client.NewForConfigOrDie(oc.AdminConfig())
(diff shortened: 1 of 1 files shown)


test/extended/cli/mustgather.go
```diff
@@ -126,6 +124,11 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 	})
 
 	g.It("runs successfully for audit logs", func() {
+		// On IBM ROKS, events will not be part of the output, since audit logs do not include control plane logs.
```
smarterclayton, line 127: Why is this new here?

test/extended/cli/mustgather.go
```diff
@@ -173,81 +173,94 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+				o.Expect(err).NotTo(o.HaveOccurred())
 
-			// On IBM ROKS, events will not be part of the output, since audit logs do not include
-			// control plane logs.
-			if e2e.TestContext.Provider != ibmcloud.ProviderName {
-				o.Expect(eventsChecked).To(o.BeNumerically(">", 1000))
+				if eventsChecked <= expectedNumberOfAuditEntries {
+					lastErr = fmt.Errorf("expected %d audit events for %q, but only got %d", expectedNumberOfAuditEntries, auditDirectory, eventsChecked)
+					return false, nil
+				} else {
```
smarterclayton, line 254: nit, kill this else

test/extended/cli/mustgather.go
```diff
@@ -173,81 +173,94 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+					o.Expect(err).NotTo(o.HaveOccurred())
 
-				scanner := bufio.NewScanner(gzipReader)
-				for scanner.Scan() {
-					text := scanner.Text()
-					if !strings.HasSuffix(text, "}") {
-						continue // ignore truncated data
+					// it will happen that the audit files are sometimes empty, we can
+					// safely ignore these files since they don't provide valuable information
+					if fi.Size() == 0 {
```
smarterclayton, line 223: because no events within that window?

smarterclayton:
Question and nit, otherwise looks fine.

test/extended/cli/mustgather.go
deads2k: > Why is this new here?

it was down below, but it nerfed the test, so this just nerfs it earlier.

test/extended/cli/mustgather.go
deads2k: > because no events within that window?

pre-existing. I don't know yet.  I think it's related to the whacky code from before that waited 10 seconds

smarterclayton: oh wait, it's because their control plane is unreachable from pods, so audit logs are unreachable

test/extended/cli/mustgather.go
```diff
@@ -173,81 +173,96 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+				// reset lastErr if we succeeded.
+				lastErr = nil
 			}
-		}
+
+			// if we get here, it means both directories checked out ok
+			return true, nil
+		})
+		o.Expect(lastErr).NotTo(o.HaveOccurred()) // print the last error first if we have one
+		o.Expect(err).NotTo(o.HaveOccurred())     // otherwise be sure we fail on the timeout if it happened
```
wking, line 265: "context expired" or whatever we get here is probably not going to be all that useful.  Can we include "no audit directories found" or some such in this error message?

marun:
/hold 

There are more than one test failures blocking the merging of this change. Please prioritize merging https://github.com/openshift/origin/pull/25314 instead, which fixes and skips a number of issues introduced by the 1.19 rebase. This PR will then need to unskip.

deads2k:
this fixes the must-gather problem.  TRT is prioritizing unblocking the payload promotion and this resolves one of two issues.  https://github.com/openshift/origin/pull/25335 / https://github.com/openshift/origin/pull/25336 resolves the other.  We will merge these

PR #25338, opened 2020-07-29
Bug 1860205: 4.4: UPSTREAM: 88684: Use MountSensitive in Azure and CephFS

tsmetana:


tsmetana:
We have reconsidered the decision to fix the bug in question: the security impact is marginal and not worth forking the k8s.io/utils library in 4.4. Closing.

PR #25350, opened 2020-07-30
[release-3.11] Bug 1760807: UPSTREAM: <carry>: Enforce label selector validation in scheduler podAffinity

damemi:
This is a non-breaking version of https://github.com/openshift/origin/pull/24616

The goal of this is to validate Pod affinitys *before* the pod is scheduled and reject the pod if there is an invalid value. Without this check, the invalid pod will be scheduled and *all subsequent pods* will fail to schedule when attempting to parse the invalid value.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/scheduler/algorithm/predicates/predicates.go
+++ b/vendor/k8s.io/kubernetes/pkg/scheduler/algorithm/predicates/predicates.go
@@ -1181,10 +1181,36 @@ func NewPodAffinityPredicate(info NodeInfo, podLister algorithm.PodLister) algor
 	return checker.InterPodAffinityMatches
 }
 
+func validatePodAffinityValues(pod *v1.Pod) error {
+	if pod.Spec.Affinity == nil {
+		return nil
+	}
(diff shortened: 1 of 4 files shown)


damemi:
/hold
need to test this in a 3.11 cluster

damemi:
/hold cancel
I didn't get a 3.11 cluster to manually test this, but did add unit tests which actually showed this wasn't originally always catching the error. Refactored it and it's not elegant, but it does get the job done and catches the error appropriately.

@soltysh could you ptal?

damemi:
```
FAIL: github.com/openshift/origin/pkg/build/builder TestDockerfilePath 20ms
FAIL: github.com/openshift/origin/pkg/build/builder TestDockerfileFromScratch 10ms
FAIL: github.com/openshift/origin/pkg/build/builder TestDockerBuildError 0s
FAIL: github.com/openshift/origin/pkg/build/builder TestPushError 0s
FAIL: github.com/openshift/origin/pkg/build/builder TestGetStrategyError 0s 
```

Don't think these are related to my changes at all
/retest

PR #25352, opened 2020-07-31
Bug 1862322: e2e/extended: disable OLM 'can subscribe to the operator' test

sttts:
After it even fails in https://github.com/openshift/origin/pull/25351, let's disable this test.

PR #25351, opened 2020-07-30
Update another OLM e2e to include quotes on substring check

damemi:
Similar to PR #25320, this changes allows this test to pass without quotes
(for pre-1.19) and with quotes (1.19+).

PR #25353, opened 2020-07-31
Bug 1857928: Re-enable olm tests

ecordell:
This reverts https://github.com/openshift/origin/pull/25352

Apparently the operator used for testing (amq) removed their stable channel, which has since been restored.

### DIFF
--- a/test/extended/operators/olm.go
+++ b/test/extended/operators/olm.go
@@ -2,6 +2,7 @@ package operators
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


ecordell:
/hold

looks like one of the test expectations is off in fips.

test/extended/operators/olm.go
```diff
@@ -172,14 +173,30 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
...
+				Status: ogStatus{
+					Namespaces: []string{},
+				},
+			}
+			if err := json.Unmarshal([]byte(output), &parsed); err != nil {
+				e2e.Logf("Failed to parse operatorgroup, error:%v", err)
+				return []string{""}
+			}
+			return parsed.Status.Namespaces
+		}, 5*time.Minute, time.Second).Should(o.Equal([]string{""}))
```
soltysh, line 199: The solution from https://github.com/openshift/origin/pull/25351/files seems like a much better alternative.

soltysh (changes requested):

test/extended/operators/olm.go
ecordell: That solution will hide bugs where the actual api response is incorrect.

hasbro17:
@ecordell This needs a rebase.

ecordell:
I should have added some context to this PR - 

The index image referenced in this test uses the 4.5 image as a base. The resolver in 4.6 doesn't work with older 4.5 index bases, so this test will not pass until the 4.5.9 base image is used in the 4.6 index.

ecordell:
/hold cancel

The 4.6 images have been updated, this test should now pass.

PR #25355, opened 2020-07-31
test: Fix inconsistencies in path reporting between Go versions

smarterclayton:
Introduction of go.mod changed the file paths reported for reflection
and when we have differening build environments we might fail to
correctly apply categorization heuristics (for instance, -mod vendor
may not report the same path in both places).

Instead of doing simple string comparison, make module checks more
precise by allowing the following forms:

* base_path/full_module/package
* base_path/full_module@/package
* base_path/full_module@version/package

### DIFF
--- a/test/extended/util/annotate/annotate.go
+++ b/test/extended/util/annotate/annotate.go
@@ -227,14 +227,26 @@ func (r *ginkgoTestRenamer) generateRename(name, parentName string, node types.T
 			name += " [Suite:openshift/conformance/parallel]"
 		}
 	}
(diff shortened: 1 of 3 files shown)


smarterclayton:
https://github.com/openshift/origin/pull/25358 passed GCP, and that is the closest way I know of to emulate ART at this point.  So given that, I will force merge this on green of serial and we will hope that it covers the as yet unknown difference between CI and ART that led to failures (but which we are hoping is 1.13 go + mod vs 1.14 go + mod)

PR #25356, opened 2020-07-31
ovnkube: skip Conntrack test that's flaky upstream

dcbw:
@danwinship @trozet @knobunc


knobunc:
Tracked by https://bugzilla.redhat.com/show_bug.cgi?id=1862557

knobunc:
@sttts I have been educated by @stevekuznetsov .  override does not do what you think it does... It will allow this PR to get into the merge queue without that test suite passing.  BUT it will not override it when the PR is tested for merging.

If you have permission to force merge this one, can you please do so?  I asked @eparis and @smarterclayton to do so, but they have not replied.  This is breaking other repos.

eparis:
it won't make things worse....

PR #25362, opened 2020-07-31
Add a new FIPS test

cgwalters:
See discussion in
https://github.com/openshift/release/pull/10488
and https://bugzilla.redhat.com/show_bug.cgi?id=1861095

This test replaces the bash code in the release repo
with a more proper test here.

While here I noticed that the topology tests had some code
that reused the MCD as a handy privileged pod; extract
that to the toplevel utils and use both here and there.

### DIFF
--- a/test/extended/security/fips.go
+++ b/test/extended/security/fips.go
@@ -1,0 +1,87 @@
+package security
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	installConfigName = "cluster-config-v1"
+	fipsFile          = "/proc/sys/crypto/fips_enabled"
+)
+
+// installConfig The subset of openshift-install's InstallConfig we parse for this test
+type installConfig struct {
+	FIPS bool `json:"fips,omitempty"`
+}
+
+func installConfigFromCluster(client clientcorev1.ConfigMapsGetter) (*installConfig, error) {
+	cm, err := client.ConfigMaps("kube-system").Get(context.Background(), installConfigName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	data, ok := cm.Data["install-config"]
+	if !ok {
+		return nil, fmt.Errorf("No install-config found in kube-system/%s", installConfigName)
+	}
+	config := &installConfig{}
+	if err := yaml.Unmarshal([]byte(data), config); err != nil {
+		return nil, err
+	}
+	return config, nil
+}
+
+func validateFIPSOnNode(oc *exutil.CLI, fipsExpected bool, node *corev1.Node) error {
+	command := []string{"cat", fipsFile}
+	out, err := exutil.ExecCommandOnMachineConfigDaemon(oc.AdminKubeClient(), oc, node, command)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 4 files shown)


cgwalters:
Tested this passes in non-FIPS, going to double check the FIPS case now too.

cgwalters:
Hmm indeed something is going wrong in fips; in [this e2e-aws-fips run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25362/pull-ci-openshift-origin-master-e2e-aws-fips/1289323031742648320) the install config has `fips: true` but it's not set in the machineconfig.  

It is working for me in a local run.

But this would be much nicer to debug if we weren't dealing with the shell test failure.  (That's https://github.com/openshift/release/pull/10488 - the problem isn't obvious to me and to be honest felt easier to rewrite than to debug and fix; bash is easy to write, but fragile in failure cases and hard to debug/maintain)

smarterclayton:
The problem with this test is that you need to know that FIPS is on.  The point of the other check is a) to check that the cluster ack'd that FIPS was on, and then b) to run a test suite that verifies normal function.  Since FIPS is not always on, this test doesn't replace the check in the ci code (which inits FIPS and then verifies a user observable side effect, then runs one or more suites).  We don't want to have a FIPS version of each suite, so we would probably need to have a "FIPS suite" that has 1-N tests that check FIPS explicitly, then in CI testing run two suites.

However, if you do that, you have to either backport that to all previous releases, or make the version check in the CI code do one of two things (if the suite is available from openshift-tests, run that, otherwise run the script for legacy behavior).  It's probably easier to just fix the CI check for the new code to start with so that we can get back to green.

cgwalters:
By the nature of Ignition, we enable FIPS from the very first thing.  Our whole architecture is designed around this "system is in desired state" by the time we get to the real root.  Specifically for FIPS in RHCOS we actually implement an extra reboot if we detect FIPS to enter that state before we do anything else. It's not like Ansible where there would be some portion of a playbook run after ssh was already up and the cluster might be running.

Hence I don't think "check FIPS, then run test suite" is necessary.

cgwalters:
> The problem with this test is that you need to know that FIPS is on

We have that data in the install config.

Now I could imagine something where we truly *triple* check by having the tests here more explicitly support the `CLUSTER_VARIANT` environment variable from the release repo and validate that if that env is set and contains `fips` it matches the install config.

(It'd be a triple check because remember FIPS is also already today validated in the MCO)

But overall this test is small and straightforward, extremely low cost - any reason not to merge?

(That said it clearly would be good to get e2e-aws-fips passing reliably again before merging)
/test e2e-aws-fips

cgwalters:
OK now that FIPS is green I'd like to get this in.

test/extended/security/fips.go
```diff
@@ -0,0 +1,87 @@
...
+	return nil
+}
+
+var _ = g.Describe("[sig-arch] [Conformance] FIPS", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("fips")
+
+	g.It("TestFIPS", func() {
+		clusterAdminKubeClientset := oc.AdminKubeClient()
+		installConfig, err := installConfigFromCluster(clusterAdminKubeClientset.CoreV1())
```
abhinavdahiya, line 66: don't read the install config to get FIPS enabled/disabled.

Use the machine config pool to figure out if FIPS is enabled using the rendered machine config and then use the target label from machine config pool to create a list of nodes to validate if the expected matched.

cgwalters: The thing is - that's what the MCO is already doing today.  See previous discussion in https://github.com/openshift/release/pull/10488

The intent of this test is to be a "triple check": if say there was a bug in the MCO that failed to get fips into the rendered machineconfig, this test would catch it.

cgwalters:
Rebased :surfer:  I still think this is a good idea, and this also contains some code I'd like to use in other tests.

cgwalters:
Trying to run `./hack/update-generated.sh` here changes all the tests to drop the `[Suite:...]`?

cgwalters:
OK @smarterclayton now that 4.7 is open I'd like to get this in, any other comments?

PR #25364, opened 2020-08-01
skip openshift-marketplace namespace when checking that images come from the release

ecordell:
openshift-marketplace intentionally pulls images from outside of the
release


bparees:
fips is perm-failing and this doesn't impact it, manually merging.

PR #25366, opened 2020-08-03
[release-4.5] Fix commit checker

tnozicka:
If we are not able to verify a commit we must fail, not silently show an error message and claim it's ok.

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/25326/pull-ci-openshift-origin-release-4.4-verify/1287675728137031680/artifacts/verify/logs/raw_test_output.log
```
=== BEGIN TEST SUITE github.com/openshift/origin/test/verify/upstream-commits ===
=== BEGIN TEST CASE ===
hack/verify-upstream-commits.sh:19: executing 'commitchecker' expecting success
SUCCESS after 0.016s: hack/verify-upstream-commits.sh:19: executing 'commitchecker' expecting success
Standard error from the command:
WARNING: one of the provided commits does not exist, not a true branch
=== END TEST CASE ===
=== END TEST SUITE ===
```

The root cause is fixed on release-4.5, the <=4.4 PR also pick https://github.com/openshift/origin/pull/24722

/cc @sttts

### DIFF
--- a/hack/verify-upstream-commits.sh
+++ b/hack/verify-upstream-commits.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+set -euxo pipefail
+
 source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
 function cleanup() {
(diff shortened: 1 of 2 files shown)


sttts:
/override ci/prow/e2e-cmd

This is a crucial fix for the verify job. e2e tests don't matter. If we don't merge it quickly, we risk inconsistencies.

sdodson:
Overrode bugzilla/valid-bug and bugzilla/severity-urgent so that these go to the front of the queue when the queue opens.

PR #25370, opened 2020-08-03
Bug 1861215: test: extended: unskip secret test

sjenning:
/cc @marun 

Now that https://github.com/openshift/kubernetes/pull/299 has merged, bump kube dep and unskip the test


sjenning:
fyi @ravisantoshgudimetla

PR #25378, opened 2020-08-05
Bug 1866411: Add --node-name and --node-ip flags to sdn

juanluisvaladas:
Currently openshift sdn tries to figure out the node name from the
hostname, and the IP address through a DNS query of that hostname.

With this code we're allowing to pass a flag to override both values.

After this is done we'll modify openshift-ansible to fill these flags
with environment variables that read fieldRefs.

### DIFF
--- a/pkg/cmd/server/kubernetes/network/sdn_linux.go
+++ b/pkg/cmd/server/kubernetes/network/sdn_linux.go
@@ -54,8 +54,8 @@ func NewSDNInterfaces(options configapi.NodeConfig, networkClient networkclient.
 
 	node, err := sdnnode.New(&sdnnode.OsdnNodeConfig{
 		PluginName:         options.NetworkConfig.NetworkPluginName,
-		Hostname:           options.NodeName,
-		SelfIP:             options.NodeIP,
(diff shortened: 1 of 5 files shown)


juanluisvaladas:
This a backport of https://github.com/openshift/sdn/pull/120 . There are a lot of differences between the bootstrap of ocp 3 and 4, so instead of doing cherry pick this is rewritten from scratch.

juanluisvaladas:
/retest
@squeed can you PTAL

juanluisvaladas:
/retest
@squeed @danwinship PTAL.

danwinship:
/hold

This is neither correct nor (I think) necessary in 3.11; you should be able to override the detected node IP via `node-config.yaml`, which can be set via an ansible variable. And for the same reason, we _can't_ override the detected node IP value with the value passed in from kubelet, because some users are currently overriding it via `node-config.yaml` and need that override to keep working.

(In 4.x there was no way to override the autodetected node IP, so switching to using the kubelet-provided IP was just fixing our autodetection to be more correct. But in 3.x, anything goes...)

PR #25385, opened 2020-08-06
Bug 1807714: oc set volume use 'filepath' not 'path' to manipulate filepaths

sallyom:
from go docs:
```
from "path" overview:
Package path implements utility routines for manipulating slash-separated paths.
The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the path/filepath package.
```
```
"path/filepath" overview:
Package filepath implements utility routines for manipulating filename paths in a way compatible with the target operating system-defined file paths.
The filepath package uses either forward slashes or backslashes, depending on the operating system. To process paths such as URLs that always use forward slashes regardless of the operating system, see the path package.
```

### DIFF
--- a/pkg/oc/cli/set/volume.go
+++ b/pkg/oc/cli/set/volume.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -667,7 +667,7 @@ func (o *VolumeOptions) setVolumeMount(spec *kapi.PodSpec, info *resource.Info) 
 
 	for _, c := range containers {
 		for _, m := range c.VolumeMounts {
-			if path.Clean(m.MountPath) == path.Clean(opts.MountPath) && m.Name != o.Name {
+			if filepath.Clean(m.MountPath) == filepath.Clean(opts.MountPath) && m.Name != o.Name {
 				return fmt.Errorf("volume mount '%s' already exists for container '%s'", opts.MountPath, c.Name)
 			}
 		}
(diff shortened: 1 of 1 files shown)


pkg/oc/cli/set/volume.go
```diff
@@ -667,7 +667,7 @@ func (o *VolumeOptions) setVolumeMount(spec *kapi.PodSpec, info *resource.Info)
 
 	for _, c := range containers {
 		for _, m := range c.VolumeMounts {
-			if path.Clean(m.MountPath) == path.Clean(opts.MountPath) && m.Name != o.Name {
+			if filepath.Clean(m.MountPath) == filepath.Clean(opts.MountPath) && m.Name != o.Name {
```
soltysh, line 670: Hmm.... that's not quite correct, b/c in containers we always want to use forward slashes, that's why `path` was being used here. Have you got a chance to verify this on a windows machine?

sallyom: hm, i see.  I'll set up Win for testing, I started doing that but then this looked straight forward so was hoping to punt to QE - I'm unfamiliar w/ Win actually and don't have a development env set up in Win - only my kid's computer. I'll set it up, though.

sallyom: @soltysh I was able to test this in Windows :) while the path->path/filepath change results in no undesired prefix, it does end up with backwards rather than forward slash in the mountPath.  See here: https://bugzilla.redhat.com/show_bug.cgi?id=1807714#c10

sallyom:
surprise - when I tried to reproduce the bug in a 3.11 cluster with oc 3.11, I could not reproduce.  I've updated the bug and am closing this PR.

PR #25386, opened 2020-08-06
Bug 1826094: Add retry options to Prometheus helper curl command

sgreene570:
Adds `--retry 15 --max-time 2 --retry-delay 1` to the Prometheus query curl call in `test/extended/util/prometheus/helpers.go`. This is to prevent the test from failing during the `e2e-aws-scaleup-rhel7` job, where DNS pods scheduled on newly spun-up rhel7 worker nodes may not immediately be responsive. As seen in the linked BZ, the first test in `test/extended/prometheus/prometheus.go` is hitting DNS errors during the `e2e-aws-scaleup-rhel7` job while the remainder of the prometheus tests are not, despite using the same curl code. Hence why adding the retry options will be beneficial in reducing test failures. 

/assign @Miciah 
/cc @danehans @frobware


Miciah:
/lgtm
/assign @knobunc
to `/approve`.

PR #25389, opened 2020-08-07
adds a new monitoring test for checking OAuth API server

p0lyn0mial:
We created a new server (OAuth API server) that is responsible for managing `oauth.openshift.io` and `user.openshift.io`. Before these groups were managed by the OpenShift API server.

This PR adds a new monitoring test for checking the availability of the new server.

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -20,6 +20,7 @@ import (
 
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 	clientimagev1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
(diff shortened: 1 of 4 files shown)


pkg/monitor/api.go
```diff
@@ -128,6 +132,44 @@ func StartOpenShiftAPIMonitoring(ctx context.Context, m *Monitor, clusterConfig
 	return nil
 }
 
+func StartOAuthAPIMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.Config, timeout time.Duration) error {
```
sttts, line 135: 1:1 copy of the openshift-apiserver one, just with another resource?

p0lyn0mial: yes :)

PR #25391, opened 2020-08-07
add a test to ensure that the kubelet is always able to mount secrets…

deads2k:
… and configmaps

At least one cause of crashlooping pods in core namespaces is a failure to start.  The container gets stuck in container creating.

This appears to be the result of the kubelet being unable to get the secret: https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-gcp-ovn-4.6/1291710556327120896

This creates a `[late]` test, which checks events in the remaining namespaces to see if we have hit this problem.

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -27,6 +27,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/origin/test/e2e/upgrade/alert"
+	"github.com/openshift/origin/test/extended/kubelet"
 	// "github.com/openshift/origin/test/e2e/upgrade/service"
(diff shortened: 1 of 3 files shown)


smarterclayton:
looks fine to me

deads2k:
The code that produces the event is here: https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/kubelet/util/manager/watch_based_manager.go#L189-L191

I think it gets retried, but we shouldn't have this problem.

sjenning:
I'd be interested to see the result of this across a lot of test runs as well.  There might be a bit of noise i.e. this condition might happen transiently but expectedly in some situations.

deads2k:
> I'd be interested to see the result of this across a lot of test runs as well. There might be a bit of noise i.e. this condition might happen transiently but expectedly in some situations.

Under what conditions do you expect the kubelet GET to take over a second?

PR #25395, opened 2020-08-10
tests: reduce flake rates when apiserver is 503

mfojtik:
This introduce `withRetryOnTemporaryErr` for extended test client that retry when API server return temporary error that are not being retried by client-go (or increase the interval we retry).

### DIFF
--- a/test/extended/util/client.go
+++ b/test/extended/util/client.go
@@ -17,9 +17,6 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/dynamic"
-
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 	"github.com/pborman/uuid"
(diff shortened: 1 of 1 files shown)


test/extended/util/client.go
```diff
@@ -702,18 +703,54 @@ func (c *CLI) AddResourceToDelete(resource schema.GroupVersionResource, metadata
...
+	}
+	if err := kretry.OnError(kretry.DefaultRetry, retriableAPIErrors, func() error {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("context cancelled")
+		default:
+		}
+		return fn()
+	}); err != nil {
+		FatalErr(err)
```
sttts, line 739: how fatal  is this? Will be get proper error messages?

test/extended/util/client.go
```diff
@@ -702,18 +703,54 @@ func (c *CLI) AddResourceToDelete(resource schema.GroupVersionResource, metadata
...
+	retriableAPIErrors := func(err error) bool {
+		return apierrors.IsServiceUnavailable(err) || // in case API server is rolling out
+			apierrors.IsServerTimeout(err) ||
+			apierrors.IsUnexpectedServerError(err) ||
+			netutil.IsConnectionRefused(err) ||
+			netutil.IsConnectionReset(err) ||
+			netutil.IsNoRoutesError(err) ||
+			netutil.IsProbableEOF(err)
+	}
+	if err := kretry.OnError(kretry.DefaultRetry, retriableAPIErrors, func() error {
```
sttts, line 731: isn't there a retry func with context already?

test/extended/util/client.go
mfojtik: stacktrace, original code did this

test/extended/util/client.go
mfojtik: in our library-go, wanted to avoid unnecessary import

test/extended/util/client.go
sttts: k

test/extended/util/client.go
sttts: don't we have library-go here?

deads2k:
why do we think we want to tolerate a 503 instead of trying to track down what is causing it to be unavailable?

deads2k:
/hold

This is working around an azure specific problem.  We should not do this.

PR #25398, opened 2020-08-10
provide operator upgrade junit results

deads2k:
This should provide per operator information about timing and failing for upgrade tests.


deads2k:
using https://github.com/openshift/origin/pull/25399 instead

PR #25399, opened 2020-08-10
test: Upgrade test should log more granular failures

smarterclayton:
If CVO failed to acknowledge upgrade within the time limit, display
a specific error. If we failed to upgrade, give each failing operator
their own failed junit result including the condition and message.

### DIFF
--- a/test/e2e/upgrade/monitor.go
+++ b/test/e2e/upgrade/monitor.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"strings"
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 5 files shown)


deads2k:
/lgtm
/hold

I want to confirm the junit looks right before releasing the hold.  This approach doesn't give how the upgrades took, but the categorization is valuable as a first step.

deads2k:
level=fatal msg="failed to fetch Cluster: failed to fetch dependency of \"Cluster\": failed to generate asset \"Platform Quota Check\": error(MissingQuota): compute.googleapis.com/cpus is not available in us-east1 because The required number of resources (24) is more than remaining quota of 21, iam.googleapis.com/quota/service-account-count is not available in global because The required number of resources (5) is more than remaining quota of 0"

/retest

smarterclayton:
Force merging since this has passed multiple times and the data will help David.

bparees:
:fireworks: :champagne:

PR #25400, opened 2020-08-10
Bug 1867800: Make EBS attachDisk idempotent

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1867800

@openshift/sig-storage


ashcrow:
@gnufied I don't see a PR in the main branch for this. Is it safe to say that the change will still be applied but via kubernetes rebases?

gnufied:
@ashcrow yes the change to master was brought via k/k rebase of 1.19.

ashcrow:
Note to next patch manager: Label removed due to missing approved.

PR #25401, opened 2020-08-10
Bug 1867803: Make EBS AttachDisk idempotent

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1867803


gnufied:
cc @openshift/sig-storage

PR #25409, opened 2020-08-12
split the crashlooping pod test to more precisely tag recycler pods

deads2k:
it looks like we have a significant number of recycler pod problems compared against "other" pods.  Both are problematic, but recycler pods are less bad and specifically storage related. This refactors the test and allows filtering.

Review with whitespace off

### DIFF
--- a/test/extended/operators/cluster.go
+++ b/test/extended/operators/cluster.go
@@ -16,105 +16,140 @@ import (
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 )
 
+var _ = g.Describe("[sig-storage] Managed cluster should", func() {
+	defer g.GinkgoRecover()
+
+	g.It("have no crashlooping recycler pods over four minutes", func() {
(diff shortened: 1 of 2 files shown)


test/extended/operators/cluster.go
```diff
@@ -16,105 +16,140 @@ import (
...
+	return strings.HasPrefix(pod.Namespace, "openshift-") || strings.HasPrefix(pod.Namespace, "kube-")
+}
+
+func not(filterFn podFilter) podFilter {
+	return func(pod *corev1.Pod) bool {
+		return !filterFn(pod)
+	}
+}
+
+func recyclerPod(pod *corev1.Pod) bool {
```
mfojtik, line 39: nit: `isRecyclerPod`

mfojtik: actually nvmd, usage reads fine

deads2k:
refining signal for e2e failures.  passed in the job that was impacted.

PR #25410, opened 2020-08-12
Bug 1868436: test/extended/dr: fix quorum-guard namespace

hexfusion:
quorum-guard switched namespaces in 4.6 this PR updates that.


hexfusion:
disruptive passed

PR #25411, opened 2020-08-12
allow longer upgrade times to run tests, but continue to fail junit at 75 minutes

deads2k:
We think we're a few minutes slow on aws.  This will let us work it out, while still having a clear indication of how slow we are.


deads2k:
only touches upgrade tests, they pass.  merging to improve aws signal

PR #25412, opened 2020-08-13
Remove marketplace test as OperatorSource is deprecated

dinhxuanvu:
The remaining test case is for OperatorSource which is deprecated
and removed from marketplace. This test suite is no longer needed.

Signed-off-by: Vu Dinh <vdinh@redhat.com>


dinhxuanvu:
Closed as other PR has fixed this already #25413

PR #25416, opened 2020-08-17
Bug 1865441: UPSTREAM: 286: Bump github.com/google/certificate-transparency-go to v1.0.20 to compile with golang >= 1.11

eparis:
This does 2 and 1/2 things which might allow 3.x to build with a newer golang. It does not actually try to build and use a newer golang.

1. bump the github.com/google/certificate-transparency-go dep since golang change something from taking a uintptr to taking a syscall.Pointer
2. use a port number instead of a port name in a couple of tests since newer golang validates that ports are numbers

and a 1/2:  use a github based import for bitbucket.org/ww/goautneg since the old one was mercurial and bitbucket dropped support for hg. hg and git have different shas, but the new shas is the exact same bits as the old sha. The github version could be bumped to something newer, but I didn't because it wasn't needed.


sosiouxme:
@eparis is this going ahead or should we make it an epic?

bparees:
/override extended_builds
/override extended_image_ecosystem

I reviewed the failures in these jobs and most of them are because of changes in the external images (tests need to be modified to change ruby deps or expected output).  One (`[Feature:Builds][Slow] can use build secrets build with secrets and configMaps should contain secrets during the source strategy build [Suite:openshift]`) is slightly concerning and i've raised it to the build-api team, but i don't have reason to believe this change is causing it as the same test was failing in this job from another PR:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25383/test_pull_request_origin_extended_builds_311/130

PR #25417, opened 2020-08-17
finish splitting each part of the upgrade into distinct junit

deads2k:
This gives distinct junit for each main failure block.  Once this merges (or once this is tagged), we can update sippy to ignore the "something failed" to reduce noise.

I recommend reviewing without whitespace diff: https://github.com/openshift/origin/pull/25417/files?w=1


deads2k:

updated test only this doesn't impact fips

deads2k:
only https://github.com/openshift/origin/pull/25386 merged since the last set of green tests.  Merging to help signal.

PR #25419, opened 2020-08-18
test: extended: add test checking for mismatch between Machines and Nodes

sjenning:
@deads2k @smarterclayton 

xref https://bugzilla.redhat.com/show_bug.cgi?id=1869654

### DIFF
--- a/test/extended/machines/cluster.go
+++ b/test/extended/machines/cluster.go
@@ -1,0 +1,49 @@
+package operators
+
+import (
+	"context"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	"github.com/stretchr/objx"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+
+	"github.com/openshift/origin/test/extended/util/ibmcloud"
+)
+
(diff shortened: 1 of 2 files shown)


deads2k:
/lgtm
/hold

holding to see if this actually fails in failed fips run as expected.  If so, I plan to override to merge so we can get the clear signal on fips for next time.  Thanks @sjenning

cgwalters:
/approve
Looks sane to me, didn't review in detail though.

deads2k:
fips is expected fail (and it did).  

This only adds a test that doesn't get added to serial overriding to improve the test coverage

sdodson:
This fails on e2e-metal
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4066/pull-ci-openshift-installer-master-e2e-metal/1295843694116605952

cgwalters:
Hmm right, we can only use this test in IPI/machineAPI managed scenarios.  Perhaps the simplest is to skip the test if the number of machines is zero?

sdodson:
Yeah that seems reasonable. Just curious if these presubmits even exist.

/test e2e-metal
/test e2e-gcp-upi

test/extended/machines/cluster.go
```diff
@@ -0,0 +1,49 @@
...
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+
+	"github.com/openshift/origin/test/extended/util/ibmcloud"
+)
+
+var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should", func() {
+	defer g.GinkgoRecover()
+
+	g.It("have same number of Machines and Nodes", func() {
+		if e2e.TestContext.Provider == ibmcloud.ProviderName {
```
cgwalters, line 23: This is just a special case of the "UPI/no-machineAPI" scenario that also applies in e.g. AWS UPI, so should be replaced by whatever check we use for that.

PR #25420, opened 2020-08-18
only produce error events for system pods, not e2e pods

deads2k:
This will prune the list of hundreds of failing events on the monitor down to something managable.  Not all e2e pods are even expected to run, so having some fail isn't something to fail or flake on.

### DIFF
--- a/pkg/monitor/pod.go
+++ b/pkg/monitor/pod.go
@@ -73,6 +73,14 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
 	podChangeFns := []func(pod, oldPod *corev1.Pod) []Condition{
 		// check phase transitions
 		func(pod, oldPod *corev1.Pod) []Condition {
+			podErrorLevel := Error
+			isSystemPod := filterToSystemNamespaces(pod)
+			if !isSystemPod {
+				// we're interested in non-system pods, but we don't want to fail our monitor on them. Not all pods
+				// in an e2e run are even expected to run, so we basically always have these events.
+				podErrorLevel = Warning
+			}
+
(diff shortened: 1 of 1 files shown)


pkg/monitor/pod.go
```diff
@@ -73,6 +73,14 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
 	podChangeFns := []func(pod, oldPod *corev1.Pod) []Condition{
 		// check phase transitions
 		func(pod, oldPod *corev1.Pod) []Condition {
+			podErrorLevel := Error
+			isSystemPod := filterToSystemNamespaces(pod)
```
bparees, line 77: how is this defined?  openshift-*?

bparees: ah, found it.

deads2k:
huh, I hope those weren't me

/test all

smarterclayton:
The problem here is that when low level stuff dies, it dies and leaves events that are useful and searchable in this list.  Ie any of the “container runtime failed to start” or “networking choked”.  Right now those are included because we had 2-3 back to back that showed them.

It might be better to prune a specific list of predictable events.  Can we try to identify a list of those first and see the number?

/hold

smarterclayton:
Another option is to dump them in another junit (split in the monitor based on namespace starting with e2e).

deads2k:
> The problem here is that when low level stuff dies, it dies and leaves events that are useful and searchable in this list. Ie any of the “container runtime failed to start” or “networking choked”. Right now those are included because we had 2-3 back to back that showed them.
> 
> It might be better to prune a specific list of predictable events. Can we try to identify a list of those first and see the number?
> 
> /hold

This PR still allows them to come out. They come out as warnings.

PR #25422, opened 2020-08-19
[release-3.11] Bug 1870064: Port stuck open when ep deleted before svc in unidling mode

tssurya:
When the ep is deleted before the svc in unidling proxy mode, the port
does not get deleted and remains stuck. This patch adds a new variable
pendingDeletion to hybrid proxier which ensures that the svc is removed
from usingUserspace and switchedToUserspace only after the deletion is
triggered from both OnServiceDelete and OnEndpointsDelete. This way the
order of whether the ep was deleted before svc or vice versa doesn't
matter anymore and the port gets safely removed.

Signed-off-by: Surya Seetharaman <suryaseetharaman.9@gmail.com>

### DIFF
--- a/pkg/proxy/hybrid/proxy.go
+++ b/pkg/proxy/hybrid/proxy.go
@@ -53,6 +53,10 @@ type HybridProxier struct {
 
 	// A gate that prevents iptables from being exhausted by proxy calls.
 	syncRunner *async.BoundedFrequencyRunner
+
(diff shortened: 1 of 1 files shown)


tssurya:
Backport of https://github.com/openshift/sdn/pull/172
So this needs to wait till the main PR gets merged.
/hold

PR #25423, opened 2020-08-19
bug 1870247: re-establish etcd connection if required for etcd test

deads2k:
The port-forward connection is being broken.  This can happen on kube-apiserver rollouts in general, which we want to tolerate, so this updates the code to handle the case of a broken connection during a test.

This happens to be the #1 serial failure on openstack.

### DIFF
--- a/test/extended/etcd/etcd_storage_path.go
+++ b/test/extended/etcd/etcd_storage_path.go
@@ -218,7 +218,7 @@ func (t *helperT) done() {
 // It will start failing when a new type is added to ensure that all future types are added to this test.
 // It will also fail when a type gets moved to a different location. Be very careful in this situation because
 // it essentially means that you will be break old clusters unless you create some migration path for the old data.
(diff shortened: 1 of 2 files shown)


test/extended/etcd/etcd_storage_path.go
```diff
@@ -434,9 +434,24 @@ func testEtcd3StoragePath(t g.GinkgoTInterface, kubeConfig *restclient.Config, e
...
+				if err != nil {
+					lastErr = err
+					continue
+				}
+				output, err = getFromEtcd(etcdClient3, testData.ExpectedEtcdPath)
+				if err != nil {
+					lastErr = err
+					continue
+				}
+				break
```
sttts, line 451: erase lastErr

deads2k:
/override ci/prow/e2e-cmd

only touches `[sig-api-machinery] API data in etcd should be stored at the correct location and version for all resources [Serial] [Suite:openshift/conformance/serial]` and https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25423/pull-ci-openshift-origin-master-e2e-aws-serial/1296129390282805248 shows that test passed

PR #25424, opened 2020-08-19
[release-3.11] Bug 1854434: UPSTREAM: 94112: Improve diff check in endpoints controller syncService

damemi:
The EndpointsController compares labels and subsets to determine if an update is needed. It sorts the calculated subsets on a service but does not sort the subsets it compares to. 

In addition, we compare the *Endpoint's* labels to the *Service's* labels, and while the Endpoint may have a `headless` label, the service does not and also leads to false no-op update calls.

This leads to *severe* performance issues in large clusters, with unimportant updates being called on as many as 2/3 (or more) of the services in the cluster at every informer resync. On clusters with thousands of services, this causes significant degradation.

From what we can tell, this issue exists at least as far back as 1.11 (likely earlier) and should be backported.

Before change:
```
$ oc get services --all-namespaces | wc -l
      70
$ oc logs pod/kube-controller-manager-ci-ln-j9ksx82-f76d1-x65w5-master-1 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "About to update endpoints"
64

$ oc logs pod/kube-controller-manager-ci-ln-j9ksx82-f76d1-x65w5-master-1 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "Update endpoints for"
22
```

These numbers correlate to how many total services are being checked vs how many are being updated. As you can see, about 1/3 of the services in the cluster are being updated without any changes happening.

After change:

```
$ oc logs pod/kube-controller-manager-ci-ln-5kyrnwk-f76d1-5qbkz-master-2 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "About to update"

64
$ oc logs pod/kube-controller-manager-ci-ln-ihf6wv2-f76d1-ch2s6-master-1 -n openshift-kube-controller-manager -c kube-controller-manager | grep endpoints_controller | grep -c "Update endpoints for"
1
```

See upstream: https://github.com/kubernetes/kubernetes/pull/94112
/priority critical-urgent
/kind bug

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/controller/endpoint/endpoints_controller.go
+++ b/vendor/k8s.io/kubernetes/pkg/controller/endpoint/endpoints_controller.go
@@ -497,9 +497,13 @@ func (e *EndpointController) syncService(key string) error {
 
 	createEndpoints := len(currentEndpoints.ResourceVersion) == 0
 
+	// Compare the sorted subsets and labels
+	// Remove the HeadlessService label from the endpoints if it exists, as this won't be set on the service itself
+	// and will cause a false negative in this diff check
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1854434
 	if !createEndpoints &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Subsets, subsets) &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Labels, service.Labels) {
+		apiequality.Semantic.DeepEqual(endpoints.RepackSubsets(currentEndpoints.Subsets), subsets) &&
+		apiequality.Semantic.DeepEqual(utillabels.CloneAndRemoveLabel(currentEndpoints.Labels, v1.IsHeadlessService), service.Labels) {
 		glog.V(5).Infof("endpoints are equal for %s/%s, skipping update", service.Namespace, service.Name)
 		return nil
 	}


damemi:
@tsmetana

vendor/k8s.io/kubernetes/pkg/controller/endpoint/endpoints_controller.go
```diff
@@ -497,9 +497,13 @@ func (e *EndpointController) syncService(key string) error {
...
 
+	// Compare the sorted subsets and labels
+	// Remove the HeadlessService label from the endpoints if it exists, as this won't be set on the service itself
+	// and will cause a false negative in this diff check
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1854434
 	if !createEndpoints &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Subsets, subsets) &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Labels, service.Labels) {
+		apiequality.Semantic.DeepEqual(endpoints.RepackSubsets(currentEndpoints.Subsets), subsets) &&
+		apiequality.Semantic.DeepEqual(utillabels.CloneAndRemoveLabel(currentEndpoints.Labels, v1.IsHeadlessService), service.Labels) {
```
soltysh, line 506: Missing `utillabels` import and `IsHeadlessService` is not present in 1.11, so this entire line is not needed for 3.11.

vendor/k8s.io/kubernetes/pkg/controller/endpoint/endpoints_controller.go
```diff
@@ -497,9 +497,13 @@ func (e *EndpointController) syncService(key string) error {
...
 	createEndpoints := len(currentEndpoints.ResourceVersion) == 0
 
+	// Compare the sorted subsets and labels
+	// Remove the HeadlessService label from the endpoints if it exists, as this won't be set on the service itself
+	// and will cause a false negative in this diff check
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1854434
 	if !createEndpoints &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Subsets, subsets) &&
-		apiequality.Semantic.DeepEqual(currentEndpoints.Labels, service.Labels) {
+		apiequality.Semantic.DeepEqual(endpoints.RepackSubsets(currentEndpoints.Subsets), subsets) &&
```
soltysh, line 505: Why `RepackSubsets`, what was the problem with sorting function?

soltysh (changes requested):
/assign

PR #25425, opened 2020-08-20
Bug 1870345: test: extended: only run new machines test against platforms that have machines

sjenning:
@deads2k @sdodson 

https://bugzilla.redhat.com/show_bug.cgi?id=1870345

### DIFF
--- a/test/extended/machines/cluster.go
+++ b/test/extended/machines/cluster.go
@@ -12,16 +12,17 @@ import (
 	"k8s.io/client-go/dynamic"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
-
-	"github.com/openshift/origin/test/extended/util/ibmcloud"
 )
 
 var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should", func() {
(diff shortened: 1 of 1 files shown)


mfojtik:
This is blocking metal tests to run, so I will button merge this when the e2e-upgrade yields green.

@sdodson @sjenning any objections?

test/extended/machines/cluster.go
```diff
@@ -12,16 +12,17 @@ import (
...
-	"github.com/openshift/origin/test/extended/util/ibmcloud"
 )
 
 var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should", func() {
 	defer g.GinkgoRecover()
 
 	g.It("have same number of Machines and Nodes", func() {
-		if e2e.TestContext.Provider == ibmcloud.ProviderName {
-			e2eskipper.Skipf("IBM Cloud clusters do not contain machineset resources")
+		if e2e.TestContext.Provider != "aws" &&
```
cgwalters, line 21: As I said in a comment in the previous PR, we need to support aws/gcp/etc UPI as well, and I am pretty sure those also don't have machine objects.

So the test should be "is machineAPI enabled"...OK I wrote up https://github.com/openshift/origin/pull/25428

PR #25427, opened 2020-08-20
BUG 1863009: vSphere provision failure on ocp46

Danil-Grigorev:
Refactor locks on registeredNodesLocks

All internal calls in these methods were already using atomic operations such as `addNode` or `GetNode`.
This lock instead was volume creation and destroying the ability to provision volume with `thin` SC.

Now the node map will be copied, and all group operations will be executed on a copy, leaving lock for others to use.

Cherry-pick https://github.com/kubernetes/kubernetes/pull/93971

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -2669,7 +2669,7 @@ k8s.io/kubernetes/third_party/forked/gonum/graph
 k8s.io/kubernetes/third_party/forked/gonum/graph/internal/linear
 k8s.io/kubernetes/third_party/forked/gonum/graph/simple
 k8s.io/kubernetes/third_party/forked/gonum/graph/traverse
-# k8s.io/legacy-cloud-providers v0.0.0 => github.com/openshift/kubernetes/staging/src/k8s.io/legacy-cloud-providers v0.0.0-20200803060402-d32435439579
+# k8s.io/legacy-cloud-providers v0.0.0 => github.com/openshift/kubernetes/staging/src/k8s.io/legacy-cloud-providers v0.0.0-20200826132615-f71a7ab366cf
 k8s.io/legacy-cloud-providers/aws
 k8s.io/legacy-cloud-providers/azure
 k8s.io/legacy-cloud-providers/azure/auth
--- a/go.mod
+++ b/go.mod
@@ -439,7 +439,7 @@ replace (
 	k8s.io/kubectl => k8s.io/kubectl v0.19.0-rc.2
 	k8s.io/kubelet => k8s.io/kubelet v0.19.0-rc.2
 	k8s.io/kubernetes => github.com/openshift/kubernetes v1.20.0-alpha.0.0.20200803060402-d32435439579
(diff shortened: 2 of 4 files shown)


Danil-Grigorev:
/hold
ci/prow/verify-deps is failing - need to pull the vendor update after merging https://github.com/openshift/kubernetes/pull/319

gnufied:
@Danil-Grigorev I think first you need to cherry-pick this into `openshift/kubernetes` and then origin repo has to be updated.

Danil-Grigorev:
@jcpowermac @sttts @gnufied PTAL. https://github.com/openshift/kubernetes/pull/319 was merged, this PR revendors origin with the fix.

sttts:
/hold

Please rename commit to "bump(k8s.io/kubernetes)".

sttts:
> Please rename commit to "bump(k8s.io/kubernetes)".

We used to have a commit checker. Looks like it's gone. cc @marun @tnozicka

vendor/modules.txt
```diff
@@ -2669,7 +2669,7 @@ k8s.io/kubernetes/third_party/forked/gonum/graph
 k8s.io/kubernetes/third_party/forked/gonum/graph/internal/linear
 k8s.io/kubernetes/third_party/forked/gonum/graph/simple
 k8s.io/kubernetes/third_party/forked/gonum/graph/traverse
-# k8s.io/legacy-cloud-providers v0.0.0 => github.com/openshift/kubernetes/staging/src/k8s.io/legacy-cloud-providers v0.0.0-20200803060402-d32435439579
+# k8s.io/legacy-cloud-providers v0.0.0 => github.com/openshift/kubernetes/staging/src/k8s.io/legacy-cloud-providers v0.0.0-20200826132615-f71a7ab366cf
```
sttts, line 2672: why is this commented out?

marun:
> > Please rename commit to "bump(k8s.io/kubernetes)".
> 
> We used to have a commit checker. Looks like it's gone. cc @marun @tnozicka

Previous release branches (release-4.5 and below) retain the commit checker. origin master (and release-4.6 and above) are effectively the same as a branch in any of the repos we maintain - no changes to `vendor` allowed` - so I removed the commit checker. It was added to openshift/kubernetes to ensure commits outside of rebases are prefixed with `UPSTREAM`.

If you feel strongly that we should force commit message format for vendoring changes, maybe that should be added as a build-machinery-go change and forced on all repos that use it vs just being a one-off for newer origin branches?

vendor/modules.txt
Danil-Grigorev: It is just a comment, identifying for a visual review where the packaged below this line comes from. This is placed automatically by `go mod vendor`. The actual information about the used packages is contained in the `go.sum`.

Danil-Grigorev:
@stts Commit message is now `Bug 1863009: bump(k8s.io/legacy-cloud-providers)`, hope this is looking as expected. PTAL

PR #25430, opened 2020-08-20
WIP: Reduce tolerance for etcd leader elections

ironcladlou:


test/extended/etcd/leader_changes.go
```diff
@@ -12,15 +12,15 @@ import (
...
-var _ = g.Describe("[sig-etcd] etcd", func() {
+var _ = g.Describe("[sig-etcd][Late] etcd", func() {
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLI("etcd-leader-change").AsAdmin()
 	g.It("leader changes are not excessive", func() {
 		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
 		o.Expect(err).ToNot(o.HaveOccurred())
 		g.By("Examining the rate of increase in the number of etcd leadership changes for last five minutes")
-		result, _, err := prometheus.Query(context.Background(), "increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[15m:1m])", time.Now())
+		result, _, err := prometheus.Query(context.Background(), "max by (job) (etcd_server_leader_changes_seen_total)", time.Now())
```
wking, line 22: I don't understand the `by (job)` bit.  @sanchezl , can you explain why you added that in a8c76b27bbc (#24291).

wking: Ah, sounds like the `by (job)` suggestion was [originally][1] from @smarterclayton.  I'm still not clear on if/why it's needed.

[1]: https://github.com/openshift/origin/pull/24291#discussion_r359007429

ironcladlou: Is it because each instance keeps its own accounting here and they often diverge? Thus we want the max as reported by all instances. cc @hexfusion

ironcladlou:
Replacing this with https://github.com/openshift/origin/pull/25454

/close

PR #25435, opened 2020-08-21
Bug 1766287: Add CI Dockerfile for hello-openshift

yselkowitz:
hello-openshift is (in)famously used throughout openshift as both an example and test image, but it's never been built or shipped by ART and it isn't multi-arch. This is the first step to bring it into the release.


sgreene570:
```
could not wait for build: the build src failed after 8s with reason PullBuilderImageFailed: Failed pulling builder image.

Pulling image docker-registry.default.svc:5000/ci-op-541dk...16d4fdaa913159315cf7d381e153339da2043ff241b4baff4f50f74 ...
error: build error: failed to pull image: unauthorized: authentication required
```
/retest

sdodson:
Can you please fill out the description of this PR? What uses this? What purpose does this new image serve?

yselkowitz:
hello-openshift is (in)famously used throughout openshift as both an example and test image, but it's never been built or shipped by ART and it isn't multi-arch.  This is the first step to bring it into the release.

PR #25439, opened 2020-08-24
Add CSI-certified driver capabilities reporting tool

arnongilboa:
We add a tool (csi_cert_storage_caps) for listing storage capabilities of a CSI certified driver, mostly but not only for CDI purposes. The tool first looks at the CSI test results junit_e2e.xml and warns if there are failed test cases (listing them if -f flag is passed). Next, it parses the Driver manifest.yaml and lists driver capabilities.

Since the tool (as a stand-alone binary) is using the input (manifest.yaml) and output (results  junit_e2e.xml) of the CDI certification tests we think the natural way for deploying it is as part of CSI certification testing container.

We currently run the CSI tests as following:
```
$ podman run -v `pwd`:/data:z --rm -it $CONTAINER sh -c "KUBECONFIG=/data/kubeconfig.yaml TEST_CSI_DRIVER_FILES=/data/manifest.yaml /usr/bin/openshift-tests run openshift/csi --junit-dir /data/results"
```
And locally run the reporting tool:
```
$ csi_cert_storage_caps 
Usage of csi_cert_storage_caps:
  -f	Lists the failed CSI tests
  -m string
    	Driver manifest.yaml (default "manifest.yaml")
  -r string
    	CSI test results junit_e2e.xml

$ csi_cert_storage_caps -m manifest.yaml -r results/junit_e2e_20200728-160628.xml
CSI test results are all successful
Test Suite Name: openshift-tests | Tests: 151 | Failures: 0 | Errors: 0 | Time: 917
---
Driver short name:                  rook-ceph
Driver name:                        rook-ceph.rbd.csi.ceph.com
Storage class:                      storageclass.yaml
Raw block VM disks supported:       true
Live migration supported:           true
Snapshotting tested:                true
VM snapshots supported:             true
Storage-assisted cloning supported: true
```
So with this patch, users will be able to run the reporting tool from the container, similarly to the tests:
```
$ podman run -v `pwd`:/data:z --rm -it $CONTAINER sh -c "/usr/bin/csi_cert_storage_caps -m /data/manifest.yaml -r /data/results/junit_e2e_20200728-160628.xml"
```

### DIFF
--- a/cmd/csi_cert_storage_caps/csi_cert_storage_caps.go
+++ b/cmd/csi_cert_storage_caps/csi_cert_storage_caps.go
@@ -1,0 +1,158 @@
+package main
+
+import (
+	"encoding/xml"
+	"flag"
+	"fmt"
+	"gopkg.in/yaml.v2"
+	"io/ioutil"
+	"os"
+)
+
+// Manifest yaml structs
+
+type StorageClass struct {
+	FromFile string `yaml:"FromFile"`
+}
+
+type SnapshotClass struct {
+	FromName bool `yaml:"FromName"`
+}
+
+type Capabilities struct {
+	Block    bool `yaml:"block"`
+	RWX      bool `yaml:"RWX"`
+	Snapshot bool `yaml:"snapshotDataSource"`
+}
+
+type DriverInfo struct {
+	Name         string `yaml:"Name"`
+	Capabilities `yaml:"Capabilities"`
+}
+
+type YamlManifest struct {
+	ShortName     string `yaml:"ShortName"`
+	StorageClass  `yaml:"StorageClass"`
+	SnapshotClass `yaml:"SnapshotClass"`
+	DriverInfo    `yaml:"DriverInfo"`
+}
+
+// Test results xml structs
+
+type TestSuite struct {
+	XMLName   xml.Name   `xml:"testsuite"`
+	Name      string     `xml:"name,attr"`
+	Tests     uint       `xml:"tests,attr"`
+	Failures  uint       `xml:"failures,attr"`
+	Errors    uint       `xml:"errors,attr"`
+	Time      float32    `xml:"time,attr"`
+	TestCases []TestCase `xml:"testcase"`
+}
+
+type TestCase struct {
+	XMLName   xml.Name `xml:"testcase"`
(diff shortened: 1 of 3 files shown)


arnongilboa:
@mfojtik, I see nothing related to the PR in the 3 failing ci jobs, and they fail again after /retest.
Can you please give it a look?

smarterclayton:
Is there an epic / story covering this?  This is pretty specific and generally "parsing junit results" isn't something we do in this repo.  If we want to report this data whenever someone runs the tests, we might want to explicitly merge this with the execution of the tests.  Is the manifest.yaml the same thing I need to pass to run the CSI tests?

aglitke:
Hi @smarterclayton, thanks for taking a look.  CNV-5018 describes the rationale behind this tool.  KubeVirt is particularly sensitive to the capabilities of a storage driver.  For example, our ability to live migrate VMs depends on storage driver support for the ReadWriteMany access mode.  The goal of this tool is to codify these dependencies so that a an integrator will have a good idea of what to expect.  The main logic maps the manifest.yaml (which is already required to run the tests and is basically a declaration of the driver capabilities) onto a set of KubeVirt capabilities.  The junit.xml processing is simply to verify that we have a passing result (ie. the declared driver capabilities match the observed capabilities).  I prefer to include the tool as a separate executable because since KubeVirt is an OpenShift add-on, this extra information will not be interesting to everyone.  We would like to include instructions for running this optional report with the KubeVirt documentation and provide it to storage vendors interested in KubeVirt integration.

smarterclayton:
> I prefer to include the tool as a separate executable because since KubeVirt is an OpenShift add-on, this extra information will not be interesting to everyone. We would like to include instructions for running this optional report with the KubeVirt documentation and provide it to storage vendors interested in KubeVirt integration.

Adding a new orthogonal binary in the tests image is the problem here - this is pretty closely related to CSI execution, which is  coupled with openshift-tests for several reasons.  Someone running the CSI suite is effectively asking for most of the info you've described, so I don't see why a separate binary in this image is suitable.

I would probably recommend that we add the capabilty to display this when the openshift/csi test is executed, perhaps as part of the preamble for the suite.

arnongilboa:
@smarterclayton , following the discussion please review [PR #25483] which is based on your comments.

PR #25440, opened 2020-08-24
Bug 1874281: enable again DR restore cluster after quorum loss

hexfusion:


ironcladlou:
Let er rip!

/lgtm
/approve

PR #25441, opened 2020-08-24
Bug 1863763: Fix statefulset update loop

tnozicka:
If there are 2 revisions, we need to sort them always the same or statefulset may hotloop updating between the 2 revisions that have the same content but different names.

/cc soltysh


soltysh (approved):
/lgtm

tnozicka:
/assign @mfojtik 
for approval

PR #25446, opened 2020-08-25
bug 1875006: update to get line numbers from etcd test

deads2k:
The etcd serial test is still flaking a fair amount, let's get line numbers and try to chase it down

### DIFF
--- a/test/extended/etcd/etcd_storage_path.go
+++ b/test/extended/etcd/etcd_storage_path.go
@@ -8,6 +8,10 @@ import (
 	"strings"
 	"testing"
 
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	o "github.com/onsi/gomega"
+
 	g "github.com/onsi/ginkgo"
 	"golang.org/x/net/context"
 
@@ -221,12 +225,6 @@ func (t *helperT) done() {
 func testEtcd3StoragePath(t g.GinkgoTInterface, kubeConfig *restclient.Config, etcdClient3Fn func() (etcdv3.KV, error)) {
 	defer g.GinkgoRecover()
 
-	// make Errorf fail the test as expected but continue until the end so we can see all failures
-	// we lose line numbers but that does not really matter for this test
-	ht := &helperT{GinkgoTInterface: t}
-	defer ht.done()
-	t = ht
-
 	var tt *testing.T // will cause nil panics that make it easy enough to find where things went wrong
 
 	kubeConfig = restclient.CopyConfig(kubeConfig)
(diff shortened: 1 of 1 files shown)


test/extended/etcd/etcd_storage_path.go
```diff
@@ -246,39 +244,34 @@ func testEtcd3StoragePath(t g.GinkgoTInterface, kubeConfig *restclient.Config, e
...
-		t.Fatalf("error creating test namespace: %#v", err)
-	}
+	_, err := kubeClient.CoreV1().Namespaces().Create(context.Background(), &kapiv1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNamespace}}, metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
 	defer func() {
-		if err := kubeClient.CoreV1().Namespaces().Delete(context.Background(), testNamespace, metav1.DeleteOptions{}); err != nil {
-			t.Fatalf("error deleting test namespace: %#v", err)
-		}
+		kubeClient.CoreV1().Namespaces().Delete(context.Background(), testNamespace, metav1.DeleteOptions{})
```
sttts, line 262: `err := `?

deads2k: > `err := `?

don't need it.  We don't do anything with the information.

marun: I think @sttts might have been trying to point out that without assigning the return value there is zero point in the subsequent error check, which would be performed on the `err` defined on line 258 that has already been verified to be nil.

If your intent is to ignore any given error, I encourage you to make that decision explicit both in code (i.e. `_ = ...`) and as a comment (e.g. `// Any error in deletion should not fail the test`) to simplify the life of future maintainers. Better still might be to assign the error and log it at warning level so that there is traceability in the event that something strange happens. 

Also, I think govet would have caught this. I've certainly seen the value in openshift/kubernetes, where we inherited the upstream check. It caught a lot of little gotchas in our carries when I got it working as part of the 1.18 rebase.

@tnozicka thoughts on adding govet support to the `make verify` command provided by build-machinery-go?

sttts: +1

tnozicka: > @tnozicka thoughts on adding govet support to the make verify command provided by build-machinery-go?
should be there
https://github.com/openshift/build-machinery-go/blob/4066d02dbe3da15b38378c826c1f886c76359e52/make/golang.mk#L8

marun: @tnozicka Hmm, but verify passed. Are the rules we're using too lax then, and it isn't looking for implicitly ignored return values?

tnozicka: I am not sure if ignored return value is a vet error. There is  a lot of these in our code base. Have you tried running `go vet` manually and see if it errors out?
(I also prefer explicitly ignoring the error with `_ = ` so the intention is clear.)

test/extended/etcd/etcd_storage_path.go
```diff
@@ -246,39 +244,34 @@ func testEtcd3StoragePath(t g.GinkgoTInterface, kubeConfig *restclient.Config, e
...
+		o.Expect(err).NotTo(o.HaveOccurred())
```
sttts, line 263: We spend far too much time on this topic. We ignore an err, but leave the err check behind. Why not just fix this and be done with it? Otherwise, the next one will readd the `err := ...` because it really looks like a mistake now.

test/extended/etcd/etcd_storage_path.go
sttts: Can we please address the review issue and be done with it? This code is fishy. Just fix it.

PR #23522, opened 2019-07-31
Bug 1723924: Initializing runningPods on SDN bootup for 3.11

alexanderConstantinescu:
This is a fix for [BUG 1723924](https://bugzilla.redhat.com/show_bug.cgi?id=1723924)

The problem:

As mentioned in the link above, when the openshift networking process dies, the "podManager" object and its attribute "runningPods" is wiped for memory, which can lead to stateful inconsistencies with iptable rules defined for the pods running on that node. 

The fix:

We now initialize the "podManager" with the initial state of all running pods when the process is booted, as such we will have a synchronized state.

PR #25448, opened 2020-08-26
Bug 1857824: Fix incorrect retrieval of sandbox pods and incorrect host port initialization

alexanderConstantinescu:
PR: https://github.com/openshift/origin/pull/23522 fixed syncing initial SDN state for hostport pods. There was however a mistake in that patch, namely: it overlooked that `existingSandboxPods` was a map of sandboxID -> PodSandbox and was retrieving the sandbox using the `podKey` (namespace/name). This lead to other initialization errors when the SDN restarts. This patch reverses the logic, and thus loops over the sandboxes and then retrieves the running pods.

I have not been able to build the images yet, as to test this on a live cluster where I've reproduced the problem. But logically this makes sense. 

/cc @aojea 
/assign @danwinship @squeed

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -347,14 +347,13 @@ func (node *OsdnNode) Start() error {
 		return err
 	}
 
-	if err = node.podManager.InitRunningPods(existingSandboxPods, existingOFPodNetworks, runningPods); err != nil {
+	if err = node.podManager.InitRunningPods(existingSandboxPods, existingOFPodNetworks, runningPods, networkChanged); err != nil {
 		return err
 	}
 
 	glog.V(2).Infof("Starting openshift-sdn pod manager")
 	if err := node.podManager.Start(cniserver.CNIServerRunDir, node.localSubnetCIDR,
-		node.networkInfo.ClusterNetworks, node.networkInfo.ServiceNetwork.String(),
-		networkChanged); err != nil {
+		node.networkInfo.ClusterNetworks, node.networkInfo.ServiceNetwork.String()); err != nil {
 		return err
 	}
 
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -473,7 +473,8 @@ func (node *OsdnNode) UpdatePod(pod kapi.Pod) error {
 	return err
 }
 
-func (node *OsdnNode) GetRunningPods(namespace string) ([]kapi.Pod, error) {
+// GetRunningPods returns a mapping of all running pods, as: namespace/name -> kapi.Pod
+func (node *OsdnNode) GetRunningPods(namespace string) (map[string]kapi.Pod, error) {
 	fieldSelector := fields.Set{"spec.nodeName": node.hostName}.AsSelector()
 	opts := metav1.ListOptions{
 		LabelSelector: labels.Everything().String(),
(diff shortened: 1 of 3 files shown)


alexanderConstantinescu:
/retitle Bug 1857824: Fix incorrect retrieval of sandbox pods

danwinship:
oops
/lgtm

alexanderConstantinescu:
/hold 

there's more to fixing the actual bug itself.

pkg/network/node/node.go
```diff
@@ -471,7 +471,7 @@ func (node *OsdnNode) UpdatePod(pod kapi.Pod) error {
 	return err
 }
 
-func (node *OsdnNode) GetRunningPods(namespace string) ([]kapi.Pod, error) {
+func (node *OsdnNode) GetRunningPods(namespace string) (map[string]kapi.Pod, error) {
```
squeed, line 474: add a quick docblock of what the return parameters are.

alexanderConstantinescu: Done

alexanderConstantinescu:
/hold cancel

Update:

We were also missing something else: previously the SDN only proceeded to setup pods with a defined `.spec.Containers.Ports[*].HostPort` during a CNI_ADD. If the SDN however restarted (and the network had changed) we only flushed the previously created iptables rules, but never re-initialized everything for whatever pods were still running when we restarted. This patch keeps the same flushing mechanism on a detected network change, but proceeds to initializing everything needed for pods who demand host port setup

PTAL, @danwinship / @squeed

alexanderConstantinescu:
/retitle Bug 1857824: Fix incorrect retrieval of sandbox pods and incorrect host port initialization

pkg/network/node/pod.go
```diff
@@ -192,42 +192,34 @@ func (m *podManager) Start(rundir string, localSubnetCIDR string, clusterNetwork
...
+		}
 		var v1Pod v1.Pod
 		if err := kapiv1.Convert_core_Pod_To_v1_Pod(&cPod, &v1Pod, nil); err != nil {
 			glog.Warningf("Could not convert core pod: %s to v1Pod", cKey)
 			continue
 		}
-
 		podPortMapping := constructPodPortMapping(&v1Pod, net.ParseIP(v1Pod.Status.PodIP))
-
 		vnid, err := m.policy.GetVNID(cPod.Namespace)
```
aojea, line 215: we still skip to portmap if we don;t find the GetNVID, is that correct?

alexanderConstantinescu: No, not anymore. We will open host ports and sync them in the step above in `shouldSyncHostports`, we won't add it do the running pods list though.

alexanderConstantinescu: Hmm, it's true that it might be better to have them aligned though and only open host ports and sync the pods that we add to `runningPods`

alexanderConstantinescu: I will correct that

pkg/network/node/pod.go
```diff
@@ -192,42 +184,46 @@ func (m *podManager) Start(rundir string, localSubnetCIDR string, clusterNetwork
...
+func (m *podManager) InitRunningPods(existingSandboxPods map[string]*kruntimeapi.PodSandbox, existingOFPodNetworks map[string]podNetworkInfo, cRunningPods map[string]kapi.Pod, clearHostPorts bool) error {
 	m.runningPodsLock.Lock()
 	defer m.runningPodsLock.Unlock()
-
-	for _, cPod := range cRunningPods {
-		cKey := getPodKey(cPod.Namespace, cPod.Name)
-
+	if m.enableHostports {
+		iptInterface := utiliptables.New(utilexec.New(), utildbus.New(), utiliptables.ProtocolIpv4)
+		m.hostportSyncer = kubehostport.NewHostportSyncer(iptInterface)
```
squeed, line 192: seems strange to me that you have to initialize `m.hostportSyncer` twice; is that correct?

alexanderConstantinescu: Essentially this block of code was executed in `Start`. However we never initialized state for any running pods in that function. Seeing as how  `InitRunningPods`...initializes running pods, and, in particular, now, their host ports: I moved it here. Thus, in case we need to clear the host ports, we do so here and then initialize the host ports right after for any pods needing such.

alexanderConstantinescu:
@danwinship could you please take a look again at this?

PR #24009, opened 2019-10-23
check the olm --version

bparees:

PR #25452, opened 2020-08-27
Revert "check the olm version"

deads2k:
This reverts commit 30a4c70cd4ccc5a63a635a0201040a34556122f1.

reverts https://github.com/openshift/origin/pull/24009

This could be enough to unstick promotion, but we really need to figure out why https://bugzilla.redhat.com/show_bug.cgi?id=1873113 broke.  It seems like ART may be publishing binaries with bad version shas?  Seems like a weird thing to suddenly change.


deads2k:
/hold

we are systemically broken on versions for many binaries.

PR #25454, opened 2020-08-27
WIP: Detect and report unexpected leader elections

ironcladlou:
Under ideal conditions, the following etcd term/election timeline is expected:

```
term  elections  reason
1     -          no bootstrap member
2     1          bootstrap member elected
3     2          bootstrap member removed, master elected
4     3          new etcd pod rev to drop bootstrap member from config
```

Elections 1 and 2 occur during etcd pivot and before Prometheus is scraping any
metrics, and so will be invisible. Election 3 is the first election that should
be collected in the metrics data. Any other elections are suspicious and could
indicate a problem (e.g. IO contention, packet loss) that we want to
investigate.

So, only 1 leader change is expected to be observed unless the test is either
disruptive or an upgrade.

This change adds a new monitor which adds a synthetic flake when unexpected
leader elections are detected so that we can identify and analyze CI runs in an
effort to reduce or eliminate such elections.

If in the future global CI Prometheus metrics are aggregated and made available
for analysis, this monitor can probably be removed.

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	e2epromclient "github.com/openshift/origin/test/extended/prometheus/client"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -21,6 +22,7 @@ import (
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 	clientimagev1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	clientoauthv1 "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	routeclientset "github.com/openshift/client-go/route/clientset/versioned"
 )
 
 // Start begins monitoring the cluster referenced by the default kube configuration until
@@ -40,6 +42,14 @@ func Start(ctx context.Context) (*Monitor, error) {
 	if err != nil {
 		return nil, err
 	}
+	routeClient, err := routeclientset.NewForConfig(clusterConfig)
+	if err != nil {
+		return nil, err
+	}
+	prometheusClient, err := e2epromclient.NewE2EPrometheusRouterClient(client, routeClient)
+	if err != nil {
+		return nil, err
+	}
 
 	if err := StartKubeAPIMonitoring(ctx, m, clusterConfig, 5*time.Second); err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/monitor/etcd.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+)
+
+func startEtcdMonitoring(ctx context.Context, m Recorder, prometheus prometheusv1.API) {
+	expectedCount := 3
```
ironcladlou, line 13: This expectation assumes no upgrades and no disruptive test scenarios

pkg/monitor/etcd.go
```diff
@@ -0,0 +1,40 @@
...
+	ticker := time.NewTicker(15 * time.Second)
```
hexfusion, line 14: prom scrapes etcd at 30s interval is this a bit spammy?

pkg/monitor/etcd.go
ironcladlou: @hexfusion actually, wonder if this could be a function of the pod revision? Is it correct to say after pivot the etcd pods should be at revision 3, and under ideal conditions, a leader election is only expected if we roll out a new revision? If so, would that be a portable value across [non]upgrade jobs?

ironcladlou:
@hexfusion @deads2k here's a run which demonstrates the "flake" injection: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25454/pull-ci-openshift-origin-master-e2e-aws-csi/1299026834821746688

and this shows how it can be discovered:

https://search.ci.openshift.org/?search=.*etcd+observed+unexpected+leader+election.*&maxAge=6h&context=1&type=bug%2Bjunit&name=&maxMatches=5&maxBytes=20971520&groupBy=job

```
Aug 27 17:53:00.492 E etcd observed unexpected leader election, expected total 3, last observed 4
```

ironcladlou:
@hexfusion @retroflexer One thing I don't quite understand yet: when the etcd cluster comes online during pivot, those members should internally begin collecting and serving metrics, including leader changes. When prometheus does finally come online and starts scraping, why is it that the post-bootstrap leader election is not served to prometheus (ref: term 3, election 2 in the table)?

pkg/monitor/api.go
```diff
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	e2epromclient "github.com/openshift/origin/test/extended/prometheus/client"
```
smarterclayton, line 10: This is not allowed - pkg/monitor may not take dependencies on deep test packages in general.

pkg/monitor/etcd.go
```diff
@@ -0,0 +1,73 @@
+package monitor
```
smarterclayton, line 1: Can this entire logic be better rolled into the product by having etcd operator write an event on leader change?

ironcladlou:
Talking more with @smarterclayton and @deads2k, we've decided to take an approach more like I started in https://github.com/openshift/origin/pull/25430:

1. Expand the existing `disruption.Flakef` functionality to be available to the other test suites, which will remove the need for abusing `Monitor` to synthesize flakes
2. Add add a [Late] test (like https://github.com/openshift/origin/pull/25430) to do the metrics detection and synthesize flakes from there

PR #25457, opened 2020-08-31
Bug 1874025: UPSTREAM: <carry>:Add a RBAC checker for external IP ranger

abhat:
Backport of PR#23826


abhat:
Closing for now till we come up with what the right solution should be.

PR #25461, opened 2020-09-02
bug 1857268: UPSTREAM: 94252: If firstTimestamp is not set use eventTime when printing event

ingvagabund:
Backport of https://github.com/kubernetes/kubernetes/pull/94252


soltysh (approved):
/lgtm
/approve

PR #25463, opened 2020-09-02
[release-3.11] Bug 1760807: Update pod affinity check in priority for valid labels

damemi:
This follows up to https://github.com/openshift/origin/pull/25350, which added these checks but only if there were no other pods on the node (making the assumption that these would be checked already if there are other pods on the node). This removes that check


damemi (commented):
/hold cancel
/cc @soltysh @ingvagabund 
could you ptal at this? This PR got kind of forgotten/held up by CI but appears to be green now. Would be great to close this bug that's been long-running

damemi:
the cmd test looks like a lot of failures like:
```
 W1209 14:17:12.593112    1083 dockerimagelookup.go:233] Docker registry lookup failed: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
error: no tags found on matching image stream: "openshift/nodejs" 
```

so there must still be tests that haven't migrated to quay images after the docker rate limit changes went in. @soltysh could we override this test?

I'm also not sure why the extended_conformance_install is failing, it's not giving a reason and the raw output is 34k lines which I haven't sorted through yet

PR #25464, opened 2020-09-02
bump(library-go): to include docker client conversions

damemi:
This bumps library-go to include the changes from https://github.com/openshift/library-go/pull/877

Because oc e2e-cmd are failing in the 1.19 GA bump (https://github.com/openshift/oc/pull/547) due to an out-of-bounds error trying to parse an empty `Items` list: https://github.com/openshift/origin/blob/master/test/extended/cmd/cmd.go#L73-L79

I'm not sure if this should just have a length check, based on the number of failures it seems like every test was failing on this which makes me think it was a bad conversion somewhere (especially since the `tag` field is still set to `latest`)


soltysh:
Replaced with https://github.com/openshift/origin/pull/25467
/close

PR #25471, opened 2020-09-03
test: improve handling of pending pods crashloop detection test

tsmetana:
The crashlooping test is also detecting pods that were pending for the entire run of the test and fails if such a pod is detected. I think the logic behind it is wrong: it creates the list of pending pods every 5 seconds so at the end of the test it is incorrect to claim the pod was pending entire time. I'm adding a time stamp comparison that should make this conclusion valid.

Also -- I don't think it's completely corect to refer to the pods by "namespace/pod_name" in this case: some pods might live for really short time and even get re-created between the polls. Hence the change to use UID instead.

### DIFF
--- a/test/extended/operators/cluster.go
+++ b/test/extended/operators/cluster.go
@@ -55,6 +55,7 @@ func crashloopingContainerCheck(podFilters ...podFilter) {
 	restartingContainers := make(map[containerName]int)
 	podsWithProblems := make(map[string]*corev1.Pod)
 	var lastPending map[string]*corev1.Pod
+	testStartTime := time.Now()
 	wait.PollImmediate(5*time.Second, 4*time.Minute, func() (bool, error) {
 		allPods, err := c.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{})
 		o.Expect(err).NotTo(o.HaveOccurred())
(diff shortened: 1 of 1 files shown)


tsmetana:
@deads2k Could you please take a look at the changes? Do they make sense? I really think the test is not doing exactly what it says...

test/extended/operators/cluster.go
```diff
@@ -136,10 +137,14 @@ func crashloopingContainerCheck(podFilters ...podFilter) {
 			e2e.Logf("Pod status %s/%s ignored for being pending", pod.Namespace, pod.Name)
 			continue
 		}
+		if pod.Status.StartTime.After(testStartTime) {
```
deads2k, line 140: This check should not be present.  A pod that was started before this test began (which is random in time to start) should definitely have time to complete its lifecycle if it is pending.

deads2k, line 140: Is this comment accurate?

```
		// this point lastPending has a list of pods that were pending a few seconds ago, but those pods could have been created
		// a few seconds before that.  What we really want is to know which pods are pending for longer than a predetermined threshold
		// of time.  This test implies that the intent is "find pods which have been pending more than four minutes.
```

tsmetana: Yes, it is.

tsmetana: OK... I'll add the comment.

deads2k:
/lgtm

but I would like to get a better test for pending pods into the monitor I think

/cherry-pick release-4.6

PR #25474, opened 2020-09-03
Bug 1875586: Remove all machine-api tests

michaelgugino:
We have our own e2e tests now.  If something fundamentally
breaks, no worker nodes will get provisioned.  These
tests are also buggy because they do not create new
machinesets for scale operations which unintentionally
disrupts other tests.


elmiko:
we talked this through on slack today, i think it makes sense to remove these tests in favor the ones in openshift/cluster-api-actuator-pkg. the ones here do not look as current as the ones we maintain in the actuator pkg.

deads2k:
per mgugino https://github.com/openshift/cluster-api-actuator-pkg/ runs on MAO and the provider repos

PR #25475, opened 2020-09-03
bug 1875592: move ISI test to disruptive since it leaves an unstable cluster

deads2k:
This test is currently disruptive because it doesn't wait for the cluster state to return to a set of stable operators (everything available, not progressing, not degraded) and the machine config operator to have finished rolling the change to every node. This causes instability in the tests that follow it that expect a stable clusters.  Examples include scheduling tests that require all workers schedulable with functional networking and storage tests which require long running connections on both apiservers and kubelets to exec into them. After meeting the standard for stability, this test is likely to be [Slow]

Found in https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-serial-4.6/1300627052105306112 specifically, but consistent with other unstable tests in storage and scheduling.

### DIFF
--- a/test/extended/images/imagestreamimport.go
+++ b/test/extended/images/imagestreamimport.go
@@ -30,7 +30,14 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
-var _ = g.Describe("[sig-imageregistry][Feature:ImageStreamImport][Serial] ImageStream API", func() {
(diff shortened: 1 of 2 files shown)


bparees:
/lgtm

@dmage fyi

deads2k:
/override ci/prow/e2e-aws-image-registry

This only affects the gcp tests running

dmage:
cc @ricardomaraschini

PR #25476, opened 2020-09-03
bug 1875586: make scaling test wait until the node can take workloads

deads2k:
more constrained version of https://github.com/openshift/origin/pull/25474

We found that we got valuable signal on this test, like https://bugzilla.redhat.com/show_bug.cgi?id=1870343

### DIFF
--- a/test/extended/machines/scale.go
+++ b/test/extended/machines/scale.go
@@ -6,24 +6,20 @@ import (
 	"strconv"
 	"time"
 
-	"k8s.io/client-go/kubernetes"
-
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/restmapper"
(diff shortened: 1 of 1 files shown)


elmiko:
is the thinking here to be a little more conservative on what we remove? (just curious)

deads2k:
> is the thinking here to be a little more conservative on what we remove? (just curious)

Yeah.  I recognized at least one test that I personally added because fips broke it.

elmiko:
makes sense to me, i'll hold off on adding lgtm so that @michaelgugino can take a look.

deads2k:
/hold

This is actually catching bugs for us.  It is failing on a couple platforms.  I'll update tomorrow to make it wait for nodes to be ready with network and schedulable.

deads2k:
/hold cancel

@mfojtik ptal

PR #25479, opened 2020-09-04
WIP: Revert "Bug 1863009: bump(k8s.io/legacy-cloud-providers)"

wking:
This reverts commit 8a9ad2c17359167acd9f585dfd6f097f1ec55b49, #25427.

Might be related to some origin panics.

Manual revert to work around context conflicts with the subsequent 2f12af0fc2025.


wking:
/close

The `Undiagnosed panic detected in pod` JUnit is a non-fatal, informer test case based on some gist scraping logs: openshift/release#11588.  Folks outside the implicated components don't have to worry about it.

PR #25480, opened 2020-09-04
Bug 1868760: Drop quay.io/fedora/fedora:32-x86_64 in favor of docker.io/fedora:32

elmiko:
This change replaces the architecture specific image spec with a spec
that support multi-arch downloads.


elmiko:
@enxebre ptal

PR #25406, opened 2020-08-11
Bug 1828858: use LoadBalancers with ExternalTrafficPolicy set to Local

aojea:
Use  ExternalTrafficPolicy set to Local for the service used in the test
`[sig-network-edge] Application behind service load balancer with PDB is not disrupted ` 
This avoid that we do a double hop inside the cluster to reach the pod, since the traffic goes directly from the load balancer to the node that has the pod running.

PR #25482, opened 2020-09-07
Bug 1828858: use ExternalTrafficPolicy local for zero downtime

aojea:
cherry-pick from #25406

Use ExternalTrafficPolicy set to Local for the service used in the test
[sig-network-edge] Application behind service load balancer with PDB is not disrupted
This avoid that we do a double hop inside the cluster to reach the pod, since the traffic goes directly from the load balancer to the node that has the pod running.


aojea:
@bparees can you please advice on the backport process?
is this the wrong branch or did I something wrong?

PR #25483, opened 2020-09-07
Bug 1879907: Add storage capabilities report for CSI certification tests

arnongilboa:
Following the discussion in [PR #25439](https://github.com/openshift/origin/pull/25439), we add storage capabilities report following the CSI certification tests:
```
$ podman run -v `pwd`:/data:z --rm -it $CONTAINER sh -c "KUBECONFIG=/data/kubeconfig.yaml TEST_CSI_DRIVER_FILES=/data/manifest.yaml /usr/bin/openshift-tests run openshift/csi --junit-dir /data/results"

openshift-tests version: v4.1.0-3107-g627d022
…
Writing JUnit report to /data/results/junit_e2e_20200907-091404.xml

32 pass, 146 skip (4m35s)

Storage Capabilities (guaranteed only on full CSI test suite with 0 fails)
==========================================================================
Driver short name:                  ceph-test
Driver name:                        test.rbd.csi.ceph.com
Storage class:                      ceph-rbd-sc.yaml
Raw block VM disks supported:       true
Live migration supported:           false
VM snapshots supported:             false
Storage-assisted cloning supported: false
```

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -8,6 +8,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -128,7 +129,11 @@ func newRunCommand() *cobra.Command {
 					return err
 				}
 				opt.MatchFn = matchFn
-				return opt.Run(args)
+				err = opt.Run(args)
+				if !opt.DryRun && len(args) > 0 && strings.HasPrefix(args[0], "openshift/csi") {
+					printStorageCapabilities(opt.Out)
+				}
+				return err
 			})
 		},
 	}
--- a/cmd/openshift-tests/storagecaps.go
+++ b/cmd/openshift-tests/storagecaps.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"fmt"
+	"gopkg.in/yaml.v2"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+// Manifest yaml structs
+
+type StorageClass struct {
+	FromFile string `yaml:"FromFile"`
+}
+
+type SnapshotClass struct {
+	FromName bool `yaml:"FromName"`
+}
+
+type Capabilities struct {
+	Block    bool `yaml:"block"`
+	RWX      bool `yaml:"RWX"`
+	Snapshot bool `yaml:"snapshotDataSource"`
+}
+
+type DriverInfo struct {
+	Name         string `yaml:"Name"`
+	Capabilities `yaml:"Capabilities"`
+}
+
+type YamlManifest struct {
+	ShortName     string `yaml:"ShortName"`
+	StorageClass  `yaml:"StorageClass"`
+	SnapshotClass `yaml:"SnapshotClass"`
+	DriverInfo    `yaml:"DriverInfo"`
+}
+
+func printStorageCapabilities(out io.Writer) {
+	manifestFilename := strings.Split(os.Getenv(manifestEnvVar), ",")[0]
+	if manifestFilename == "" {
+		fmt.Fprintln(out, "No manifest filename passed")
+		return
+	}
+
+	yamlFile, err := ioutil.ReadFile(manifestFilename)
+	if err != nil {
+		fmt.Fprintln(out, "Failed to", err)
+		return
+	}
+
+	var yamlManifest YamlManifest
+	err = yaml.Unmarshal(yamlFile, &yamlManifest)
+	if err != nil {
+		fmt.Fprintln(out, "Error parsing", err)
+		return
+	}
+
+	supportsSnapshot := yamlManifest.DriverInfo.Capabilities.Snapshot && yamlManifest.SnapshotClass.FromName
+
+	fmt.Fprintln(out)
+	fmt.Fprintln(out, "Storage Capabilities (guaranteed only on full CSI test suite with 0 fails)")
+	fmt.Fprintln(out, "==========================================================================")
+	fmt.Fprintln(out, "Driver short name:                 ", yamlManifest.ShortName)
+	fmt.Fprintln(out, "Driver name:                       ", yamlManifest.DriverInfo.Name)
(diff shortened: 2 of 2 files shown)


cmd/openshift-tests/storagecaps.go
```diff
@@ -0,0 +1,81 @@
...
+	SnapshotClass `yaml:"SnapshotClass"`
+	DriverInfo    `yaml:"DriverInfo"`
+}
+
+const (
+	testReportEnvVar = "TEST_REPORT"
+)
+
+var (
+	manifestFilename string
```
smarterclayton, line 43: Don't use global variables, pass them explicitly.

cmd/openshift-tests/storagecaps.go
```diff
@@ -0,0 +1,81 @@
...
+
+type YamlManifest struct {
+	ShortName     string `yaml:"ShortName"`
+	StorageClass  `yaml:"StorageClass"`
+	SnapshotClass `yaml:"SnapshotClass"`
+	DriverInfo    `yaml:"DriverInfo"`
+}
+
+const (
+	testReportEnvVar = "TEST_REPORT"
```
smarterclayton, line 39: Go style for single constants omits `(`

cmd/openshift-tests/storagecaps.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var (
+	manifestFilename string
+)
+
+// Generic test reporting with env var selector
+func printReport() {
+	testReport := os.Getenv(testReportEnvVar)
```
smarterclayton, line 48: This should be a flag, not an env var lookup

cmd/openshift-tests/storagecaps.go
```diff
@@ -0,0 +1,81 @@
...
+	var yamlManifest YamlManifest
+	err = yaml.Unmarshal(yamlFile, &yamlManifest)
+	if err != nil {
+		fmt.Println("Error parsing", err)
+		return
+	}
+
+	supportsSnapshot := yamlManifest.DriverInfo.Capabilities.Snapshot && yamlManifest.SnapshotClass.FromName
+
+	fmt.Println()
```
smarterclayton, line 71: Commands are expected to take the stream to print to as an argument, pass an io.Writer in to this method and use `fmt.Fprintf`

arnongilboa:
@smarterclayton fixed according to your comments, please review.

cmd/openshift-tests/openshift-tests.go
```diff
@@ -286,4 +288,5 @@ func bindOptions(opt *testginkgo.Options, flags *pflag.FlagSet) {
 	flags.DurationVar(&opt.Timeout, "timeout", opt.Timeout, "Set the maximum time a test can run before being aborted. This is read from the suite by default, but will be 10 minutes otherwise.")
 	flags.BoolVar(&opt.IncludeSuccessOutput, "include-success", opt.IncludeSuccessOutput, "Print output from successful tests.")
 	flags.IntVar(&opt.Parallelism, "max-parallel-tests", opt.Parallelism, "Maximum number of tests running in parallel. 0 defaults to test suite recommended value, which is different in each suite.")
+	flags.StringVar(&opt.Report, "report", opt.Report, "Print the requested report at the end of the test suite.")
```
smarterclayton, line 291: Reading this further, is there any reason *not* to print the storage report on every storage test execution?  We definitely don't want to print it on dryrun (since dry run is specifically used to output test names to stdout, we don't want to print something else to stdout).  Otherwise we can simply print the storage report if we're not a dryrun and the suite is one of the csi suites.  That might look like:

```
if !opt.DryRun && len(args) > 0 && args[0] == "openshift/csi/..." {
  printStorageReport(...)
}
```

which removes the need for the flag.

smarterclayton:
/lgtm
/approve

But you need to attach a bug

arnongilboa:
@smarterclayton we have a jira issue [CNV-5993](https://issues.redhat.com/browse/CNV-5993) (epic CNV-5018). Can we use this one or we need a bugzilla clone?

> /lgtm
> /approve
> 
> But you need to attach a bug

smarterclayton:
You need a bugzilla

arnongilboa:
@smarterclayton [Bugzilla bug 1879907](https://bugzilla.redhat.com/show_bug.cgi?id=1879907) severity is set. No idea why the `bugzilla/severity-unspecified` flag was not removed.

cmd/openshift-tests/storagecaps.go
```diff
@@ -0,0 +1,73 @@
...
+	fmt.Fprintln(out)
+	fmt.Fprintln(out, "Storage Capabilities (guaranteed only on full CSI test suite with 0 fails)")
+	fmt.Fprintln(out, "==========================================================================")
+	fmt.Fprintln(out, "Driver short name:                 ", yamlManifest.ShortName)
+	fmt.Fprintln(out, "Driver name:                       ", yamlManifest.DriverInfo.Name)
+	fmt.Fprintln(out, "Storage class:                     ", yamlManifest.StorageClass.FromFile)
+	fmt.Fprintln(out, "Raw block VM disks supported:      ", yamlManifest.DriverInfo.Capabilities.Block)
+	fmt.Fprintln(out, "Live migration supported:          ", yamlManifest.DriverInfo.Capabilities.RWX)
+	fmt.Fprintln(out, "VM snapshots supported:            ", supportsSnapshot)
+	fmt.Fprintln(out, "Storage-assisted cloning supported:", supportsSnapshot)
```
jsafrane, line 71: These are CNV features, not CSI features. `openshift/csi` neither tests nor cares about virtualization. The output is very confusing, for example it uses capability `Snapshot` for "Storage-assisted cloning supported" , while the real cloning tests are covered by capability `pvcDataSource`.

I understand that you may need some CNV-tailored output, but please report that as CNV capabilities, not CSI.

Please include @openshift/storage team for any CSI PRs next time.

arnongilboa:
@jsafrane , thanks for your comment. I'll fix the title of the reported capabilities for clarification.

cmd/openshift-tests/storagecaps.go
arnongilboa: @jsafrane , thanks for your comment. I'll fix the title of the reported capabilities for clarification.

PR #25491, opened 2020-09-09
Bug 1847363: test: use ubi instead of centos

r4f4:
There is no centos for s390x architecture, so instead use an image that's guaranteed to be multiarch.


yselkowitz:
ping?

PR #25492, opened 2020-09-10
bug 1877932: don't retry the GRPC connection in our tests

deads2k:
Based on logs from https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-openstack-serial-4.6/1303933478840045568 on the failing test, it looks like the grpc client itself is retrying on failures forever.  This stops that I think


hexfusion:
lets see how it goes

/lgtm

PR #25503, opened 2020-09-15
Bug 1874671: Add NET_RAW capability to internal_ports test

umohnani8:
Fixes #1874671 - helps fix the tests for the PR dropping the capabilities https://github.com/openshift/machine-config-operator/pull/2051

The "TCP and UDP on ports" test does a tcpdump which
requires the NET_RAW capability to be enabled.
We are dropping NET_RAW from the default list of
capabilities in cri-o, so need to explicitly add it
to the test here.

This is blocking https://github.com/openshift/machine-config-operator/pull/2051

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

### DIFF
--- a/test/extended/networking/internal_ports.go
+++ b/test/extended/networking/internal_ports.go
@@ -84,6 +84,11 @@ var _ = ginkgo.Describe("[sig-network] Internal connectivity", func() {
 										},
 									},
 								},
+								SecurityContext: &v1.SecurityContext{
+									Capabilities: &v1.Capabilities{
+										Add: []v1.Capability{"NET_RAW"},
+									},
+								},
 							},
 						},
 					},


umohnani8:
@dcbw PTAL

abhat:
Now that cri-o doesn't set the NET_RAW capability by default, are there other tests where we need to set this as well? 
/lgtm

umohnani8:
@abhat so far, I have only seen this test fail. Will update if I find any more failures.

mrunalp:
This needs to be backported to 4.5 as well.

umohnani8:
@abhat @danwinship can I please get an approve here.

cgwalters (approved):

PR #25517, opened 2020-09-17
Bug 1879385: move olm source SHA test to sig-arch

kevinrizza:
This is a flaky test. The context of the test is valid (ensuring that the downstreaming process doesn't result in potential diffs), but this is not the right place for it nor should it block other release components from merging.

The team is going to work with ART in order to see what can be done to reenable testing for this problem somewhere further down the process.


dinhxuanvu (approved):
/lgtm

deads2k:
/hold

This test has caught two important regressions in the past five weeks.  While I'm sympathetic to the pressure you feel when the test fails, it has proven quite valuable.

dinhxuanvu:
Replaced by #25527

PR #25523, opened 2020-09-18
Bug 1879873: OLM 'can subscribe to the operator' test: skip if redhat-operators is not enabled

vrutkovs:
OKD is defaulting to community-operators only, so the test should 
install operator from community sources. This change would make it pass on OCP and skip this test on OKD. Verified that this test gets skipped with current OKD configuration and runs on OCP.

Later, when community-operators source would contain amq-streams, I'll rework the test to use community source on OKD.

This should resolve [failing OKD promotions](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-okd-installer-e2e-aws-4.6/1306845773220024320)

TODO:
* [ ] [Add OKD tests to the repo](https://github.com/openshift/release/pull/11925)

### DIFF
--- a/test/extended/operators/olm.go
+++ b/test/extended/operators/olm.go
@@ -151,6 +151,38 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
 	})
 })
 
+func hasRedHatOperatorsSource(oc *exutil.CLI) (bool, error) {
+	spec, err := oc.AsAdmin().Run("get").Args("operatorhub/cluster", "-o=jsonpath={.spec}").Output()
+	if err != nil {
+		return true, fmt.Errorf("Error reading operatorhub spec: %s", spec)
+	}
+	type Source struct {
+		Name     string `json:"name"`
+		Disabled bool   `json:"disabled"`
+	}
+	type Spec struct {
+		DisableAllDefaultSources bool     `json:"disableAllDefaultSources"`
+		Sources                  []Source `json:"sources"`
+	}
+	parsed := Spec{}
+	err = json.Unmarshal([]byte(spec), &parsed)
+	if err != nil {
+		return true, fmt.Errorf("Error unmarshalling operatorhub spec: %s", spec)
+	}
+	// Check if default hub sources are used
+	if len(parsed.Sources) == 0 && !parsed.DisableAllDefaultSources {
+		return true, nil
+	}
+
+	// Check if redhat-operators is listed and not disabled
+	for _, source := range parsed.Sources {
+		if source.Name == "redhat-operators" && source.Disabled == false {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
 // This context will cover test case: OCP-23440, author: jiazha@redhat.com
 // Uses nfd operator
 var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
@@ -167,6 +199,13 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
 	files := []string{sub}
 	g.It("can subscribe to the operator", func() {
 		g.By("Cluster-admin user subscribe the operator resource")
+
+		// skip test if redhat-operators is not present or disabled
+		ok, err := hasRedHatOperatorsSource(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if !ok {
+			g.Skip("redhat-operators source not found in enabled sources")
+		}
 
 		// configure OperatorGroup before tests
 		configFile, err := oc.AsAdmin().Run("process").Args("--ignore-unknown-parameters=true", "-f", operatorGroup, "-p", "NAME=test-operator", fmt.Sprintf("NAMESPACE=%s", oc.Namespace())).OutputToFile("config.json")


vrutkovs:
/test e2e-gcp

install failure

test/extended/operators/olm.go
```diff
@@ -168,6 +168,28 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
 	g.It("can subscribe to the operator", func() {
 		g.By("Cluster-admin user subscribe the operator resource")
 
+		// skip test if redhat-operators is not present or disabled
+		sources, err := oc.AsAdmin().Run("get").Args("operatorhub/cluster", "-o", "jsonpath='{.spec.sources}'").Output()
```
kevinrizza, line 172: The logic here is a little bit more complex. In addition to checking if redhat-operators is explicitly disabled in `spec.Sources`, you also need to check if `spec.disableAllDefaultSources` is true, and if it is check `if source.Name == "redhat-operators" && source.Disabled = False`. If that's true, then the redhat-operators source is enabled

test/extended/operators/olm.go
```diff
@@ -168,6 +168,28 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
...
+			Name     string `json:"name"`
+			Disabled bool   `json:"disabled"`
+		}
+		parsed := []Source{}
+		err = json.Unmarshal([]byte(sources), &parsed)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		hasRedHatOperators := false
+		for _, source := range parsed {
+			if source.Name == "redhat-operators" && source.Disabled == false {
```
kevinrizza, line 184: Don't you want to check the inverse of this? if `source.Disabled == true` then the redhat operator catalog is not enabled on the cluster.

test/extended/operators/olm.go
vrutkovs: Added `disableAllDefaultSources` check

test/extended/operators/olm.go
vrutkovs: Updated the following line to read `hasRedHatOperators = true` when "redhat-operators" is present and not disabled

test/extended/operators/olm.go
```diff
@@ -168,6 +168,36 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
 	g.It("can subscribe to the operator", func() {
 		g.By("Cluster-admin user subscribe the operator resource")
 
+		// skip test if redhat-operators is not present or disabled
+		spec, err := oc.AsAdmin().Run("get").Args("operatorhub/cluster", "-o", "jsonpath='{.spec}'").Output()
```
smarterclayton, line 172: This really should be in a function

vrutkovs: Reworked into `hasRedHatOperatorsSource`

vrutkovs:
/test e2e-gcp

Volume test failed

test/extended/operators/olm.go
```diff
@@ -151,6 +151,41 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
...
+		return true, nil
+	}
+
+	// Check if redhat-operators is listed and not disabled
+	for _, source := range parsed.Sources {
+		if source.Name == "redhat-operators" && source.Disabled == false {
+			return true, nil
+		}
+	}
+	e2e.Logf("Skipping test - redhat-operators not found in enabled sources")
```
smarterclayton, line 185: this should not be necessary given the skip rule

test/extended/operators/olm.go
```diff
@@ -151,6 +151,41 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
...
+		Disabled bool   `json:"disabled"`
+	}
+	type Spec struct {
+		DisableAllDefaultSources bool     `json:"disableAllDefaultSources"`
+		Sources                  []Source `json:"sources"`
+	}
+	parsed := Spec{}
+	err = json.Unmarshal([]byte(spec), &parsed)
+	if err != nil {
+		e2e.Logf("Error unmarshalling operatorhub spec: {}", spec)
```
smarterclayton, line 171: wrap the error and remove the log line

test/extended/operators/olm.go
```diff
@@ -151,6 +151,41 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
 	})
 })
 
+func hasRedHatOperatorsSource(oc *exutil.CLI) (bool, error) {
+	spec, err := oc.AsAdmin().Run("get").Args("operatorhub/cluster", "-o=jsonpath={.spec}").Output()
+	if err != nil {
+		e2e.Logf("Error reading operatorhub spec: {}", spec)
```
smarterclayton, line 157: wrap error and remove log line

vrutkovs: Reworked this to return `error` and the test would check if its nil - or skip test with err message

test/extended/operators/olm.go
```diff
@@ -151,6 +151,38 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
...
+		Disabled bool   `json:"disabled"`
+	}
+	type Spec struct {
+		DisableAllDefaultSources bool     `json:"disableAllDefaultSources"`
+		Sources                  []Source `json:"sources"`
+	}
+	parsed := Spec{}
+	err = json.Unmarshal([]byte(spec), &parsed)
+	if err != nil {
+		return fmt.Errorf("Error unmarshalling operatorhub spec: %s", spec)
```
smarterclayton, line 170: This changes the behavior of the test - before you were reporting errors as test failures, now you are reporting them as skips.  It would always be better to report errors as test failures, and skip is a boolean with a hardcoded message.  You don't need to return the message from this function, just inline the message in the test block.

We don't want to suddenly start failing unmarshal (because of a schema change) and accidentally always skip this test

vrutkovs: Oh, I see. Reworked this to return bool and err, which is expected to be nil on OCP and OKD

test/extended/operators/olm.go
```diff
@@ -151,6 +151,38 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
...
+		return true, nil
+	}
+
+	// Check if redhat-operators is listed and not disabled
+	for _, source := range parsed.Sources {
+		if source.Name == "redhat-operators" && source.Disabled == false {
+			return true, nil
+		}
+	}
+	return false, fmt.Errorf("redhat-operators not found in enabled sources")
```
smarterclayton, line 183: not an error, `return false, nil`

vrutkovs:
Install failure

/test e2e-gcp

PR #25524, opened 2020-09-18
Bug 1880683: test/extended/etcd: make excessive leader_change test actionable

hexfusion:

### DIFF
--- a/test/extended/etcd/leader_changes.go
+++ b/test/extended/etcd/leader_changes.go
@@ -2,6 +2,7 @@ package etcd
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	g "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


test/extended/etcd/leader_changes.go
```diff
@@ -16,11 +16,15 @@ var _ = g.Describe("[sig-etcd] etcd", func() {
...
+		t := g.GinkgoT()
 		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
 		o.Expect(err).ToNot(o.HaveOccurred())
 		g.By("Examining the rate of increase in the number of etcd leadership changes for last five minutes")
 		result, _, err := prometheus.Query(context.Background(), "increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[15m:1m])", time.Now())
 		o.Expect(err).ToNot(o.HaveOccurred())
-		o.Expect(result.(model.Vector)[0].Value).To(o.BeNumerically("==", 0))
+		leaderElectionLastFiveMinutes := result.(model.Vector)[0].Value
+		if leaderElectionLastFiveMinutes != 0 {
+			t.Errorf("Leader changes observed last 5m %q: Leader changes are a result of stopping the etcd leader process or from latency (disk or network), review etcd performance metrics", leaderElectionLastFiveMinutes)
```
deads2k, line 27: `o.Expect(fmt.Errorf("***your stuff here***).ToNot(o.HaveOccurred())`  it gives you a line number that way

deads2k (changes requested):

deads2k:
/lgtm
/hold

hold for nit.  Feel free to reapply and remove old after.

test/extended/etcd/leader_changes.go
hexfusion: cool thx

ironcladlou:
Will this trigger a false position when we do the bootstrap removal, or when we churn a rev during upgrade tests?

hexfusion:
> Will this trigger a false positive when we do the bootstrap removal, or when we churn a rev during upgrade tests?

I think we should be clear of these before prom comes up but its a good point.

PR #25545, opened 2020-09-23
WIP: maintain LastTriggeredImageID in BuildConfig status instead of spec

gabemontero:


gabemontero:
/close

between @pweil- @bparees @adambkaplan and myself we've concluded removal of this from spec could break existing users who built on the current behavior

@adambkaplan is driving mitigations with customer ... based on fallout, if some altered form of this is taken on, we'll go from there

PR #25549, opened 2020-09-24
Bug 1880741: Remove broken machine-api taint test

michaelgugino:
This test suite is currently checking for labels
and taints on nodes created by the machine-api.
Other components are adding taints to existing workers
which are sometimes selected to scale down during
testing.  This test does not allow for any modification
of labels and taints.

The machine-api is not the sole or final manager of
labels and taints.  We have coverage in our own
test suite that the machine-api adds labels
and taints appropriately, this test is unnecessary
here and does not add cross-component signaling.

### DIFF
--- a/test/extended/machines/scale.go
+++ b/test/extended/machines/scale.go
@@ -223,15 +223,8 @@ var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Serial] Managed cl
 				return false
 			}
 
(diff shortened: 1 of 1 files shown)


elmiko (approved):
this makes sense to me, if we have out of band actors adding taints this part of the test starts to seem less meaningful to me.
/lgtm

PR #25550, opened 2020-09-24
Bug 1881060: don't wait for an OLM operator to succeed on non-amd64

ecordell:
until we have a good multi-arch operator available for test, skip the check that waits for an operator to succeed.


ecordell:
/hold

not sure that checking runtime arch will indicate the target cluster arch. depends on how non-amd64 arches are tested.

deads2k:
You'll want to restore this on other platforms I think, but thanks for releasing ppc and 390

PR #25557, opened 2020-09-26
Bug 1883235: Adjust wait duration for marketplace test

dinhxuanvu:
Increase from 60s to 2 mins due to recent increase in initial delay
in the catalog pod.

Signed-off-by: Vu Dinh <vdinh@redhat.com>


njhale:
/rename Bug 1883235: Adjust wait duration for marketplace test

ashcrow:
I'm unable to find the original main branch PR in the parent BZs. Can you provide a link or note why it's needed in 4.5 only?

dinhxuanvu:
@ashcrow Hey Stephen. This BZ is to address an issue for a test that is specified to `OperatorSource` which is deprecated in 4.6 so the test is removed in 4.6. The test only exists in 4.5.

PR #25565, opened 2020-09-28
Bug 1873043: Fix annotation generation

marun:
Previously the check for whether a test should be annotated for the origin suite (`[Suite:openshift]`) depended on the repo clone being in a GOPATH (i.e. src/github.com/openshift/origin). https://github.com/openshift/kubernetes/pull/377 avoids this problem by disabling the addition of the annotation - whether a test is defined in origin can be inferred by the absence of the `[Suite:k8s]` annotation. This PR bumps openshift/kubernetes to pick up that change and updates the annotations

/cc @ironcladlou @smarterclayton


sttts:
/lgtm
/approve

Testing only change.

marun:
Updated to fix upgrade test matching (previously relied on `[Suite:openshift]`, now relies on the absence of `[Suite:k8s]`).

PR #25568, opened 2020-09-30
Bug 1882750: extended/apiserver: add test looking out for GracefulTerminationTimeout events

sttts:
Counterpart of https://github.com/openshift/kubernetes/pull/379.


marun:
@sttts You'll need to run `make update` to update the test annotations so that the test will be annotated for inclusion in the test suites.

PR #25572, opened 2020-09-30
bug 1884034: add junit for illegal pod state transitions

deads2k:
pods should never transition back to pending. "invariant violation (bug): pod should not transition Running->Pending even when terminated"

This appears to be frequent https://search.ci.openshift.org/?search=pod+should+not+transition+Running-%3EPending+even+when+terminated&maxAge=48h&context=1&type=build-log&name=4.6&maxMatches=5&maxBytes=20971520&groupBy=job so we'll add a specific failure signal that we can chase.


deads2k:
only its prereq merged. no conflict

PR #25576, opened 2020-10-01
bug 1884252: add test for systemd timeouts

deads2k:
like https://github.com/openshift/origin/pull/25564, but for all test namespaces


deads2k:
no conflicts

PR #25579, opened 2020-10-02
Bug 1881144: Fix Cinder e2e tests not to leak volumes

jsafrane:
To get fixes in Cinder tests. Cinder pre-provisioned tests should not leak volumes.
@openshift/storage 

https://github.com/openshift/kubernetes/pull/380
https://github.com/kubernetes/kubernetes/pull/95003


bertinatto (commented):
/lgtm

PR #25583, opened 2020-10-06
hello-openshift: fix base images

yselkowitz:
/cc @sgreene570


images/hello-openshift/Dockerfile.rhel
```diff
@@ -1,11 +1,9 @@
-#FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-openshift-4.6 AS builder
-FROM registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder:rhel_8_golang_1.14 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-openshift-4.6 AS builder
```
sgreene570, line 1: should this image be `rhel-8-golang-1.15-openshift-4.6` or is that the new default in this case?

yselkowitz: you're right, a bit preoccupied atm.

sgreene570: np! Thanks for making this PR.

PR #25585, opened 2020-10-06
[wip] test/extended/dr: bump timeout for new machines

hexfusion:
Machines take longer to delete/come back in 4.6+ bumping timeout.

```
fail [github.com/openshift/origin/test/extended/dr/quorum_restore.go:195]: Unexpected error:
    <*errors.StatusError | 0xc001cb10e0>: {
        ErrStatus: {
            TypeMeta: {Kind: "Status", APIVersion: "v1"},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "Internal error occurred: resource quota evaluation timed out",
            Reason: "InternalError",
            Details: {
                Name: "",
                Group: "",
                Kind: "",
                UID: "",
                Causes: [
                    {
                        Type: "",
                        Message: "resource quota evaluation timed out",
                        Field: "",
                    },
                ],
                RetryAfterSeconds: 0,
            },
            Code: 500,
        },
    }
    Internal error occurred: resource quota evaluation timed out
occurred
```
ref: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/459/pull-ci-openshift-cluster-etcd-operator-master-e2e-disruptive/1311384501846282240

### DIFF
--- a/test/extended/dr/quorum_restore.go
+++ b/test/extended/dr/quorum_restore.go
@@ -188,7 +188,8 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 								return false, nil
 							}
 							if err != nil {
-								return false, err
+								framework.Logf("Debug: error creating machine %q: %q", master, err.Error())
+								return false, nil
 							}
 							return true, nil
 						})
--- a/test/extended/dr/quorum_restore.go
+++ b/test/extended/dr/quorum_restore.go
@@ -114,7 +114,7 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 						continue
 					}
 
-					framework.Logf("Destroying %s", masterMachine)
+					framework.Logf("Destroying master: %s IPs: %v", masterMachine, node.Status.Addresses)
 					err = ms.Delete(context.Background(), masterMachine, metav1.DeleteOptions{})
 					o.Expect(err).NotTo(o.HaveOccurred())
 				}
@@ -162,7 +162,7 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 				if expectedNumberOfMasters == 1 {
-					framework.Logf("Cannot create new masters, you must manually create masters and update their DNS entries according to the docs")
+					framework.Logf("Cannot create new masters, you must manually create masters")
 				} else {
 					framework.Logf("Create new masters")
 					for _, master := range masterMachines {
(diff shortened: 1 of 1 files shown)


smarterclayton:
Do we know why?

hexfusion:
> Do we know why?

/hold

we should understand why before this merges.

hexfusion:
In speaking with cloud team they felt that the time would be blocking on delete. So I will try to better understand the reasons for that and if we introduced it with the test.

hexfusion:
> real	2m1.618s

Manual delete of the machine takes around 2m in 4.6. Once that machine was removed I wonder if the issue is then around the load balancers still having 3 endpoints. Below is evidence of this machine  `54.177.192.129` was removed.

```
oc get nodes
Unable to connect to the server: dial tcp 54.177.192.129:6443: i/o timeout
```

hexfusion:
It appears we changed the way that LB health checks are performed. But also I think since we delete the first endpoint possibly we could hit this taken from BZ 1869629 from AWS

>Thank you for contacting Elastic Load Balancing about the length of time taken for a target to complete a healthy/Unhealthy state transition on your Network Load Balancer. We can confirm that the expected behavior of the Network Load Balancer health check system includes an additional delay of approximately 10-15 seconds in the time between when a target is detected as changing health state and when the new state is reflected in the load balancer. This is due to the distributed nature of the health check system. This system aggregates data about the health state of the target and distributes it to the Network Load Balancer.

>For example, if you have configured a 30 second health check interval and there is a threshold of 3 health checks to change the state, a healthy target that becomes unhealthy may experience up to 129 seconds of new requests routed to it after failing its first health check. This is due to the possibility that it fails 1 second after passing, so the first period between health checks could be up to 29 seconds. Then there is the 90 seconds of health checks that must fail, in this example as we have configured 30 second health check intervals, and 3 as the health check failure threshold, and then the 10 seconds to aggregate and distribute the health state information.

[1] https://github.com/openshift/installer/pull/3709
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1869629#c5

hexfusion:
/lgtm cancel

@retroflexer it was my fault for not marking this wip untagging lgtm. Do you have any feelings on the above?

hexfusion:
dns issues with bastion last couple runs...

/test e2e-aws-disruptive

hexfusion:
> level=error msg="Cluster operator monitoring Degraded is True with UpdatingopenshiftStateMetricsFailed: Failed to rollout the stack. Error: running task Updating openshift-state-metrics failed: reconciling openshift-state-metrics Deployment failed: updating Deployment object failed: waiting for DeploymentRollout of openshift-monitoring/openshift-state-metrics: got 1 unavailable replicas"

install failure........

/test e2e-aws-disruptive

PR #25586, opened 2020-10-06
Bug 1852289: Add libvirt to the list of unsupported platforms for LB service

r4f4:
The upgrade test suite includes an external load balancer service test
which is not supported in libvirt (used for s390x/ppc64le CI). Maybe
there is a possibility/use-case for metal/libvirt deploys to have
external load balancers, but it'd require a publically routable IP and
managing the access to the IP/port. Not sure if that will be supported
in non-cloud platforms.

So for now let's just add a new suite that skips the LB service,
allowing us to run upgrade tests in CI.

### DIFF
--- a/test/e2e/upgrade/service/service.go
+++ b/test/e2e/upgrade/service/service.go
@@ -49,7 +49,7 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
 	infra, err := configClient.ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
 	framework.ExpectNoError(err)
 	// ovirt does not support service type loadbalancer because it doesn't program a cloud.
-	if infra.Status.PlatformStatus.Type == configv1.OvirtPlatformType || infra.Status.PlatformStatus.Type == configv1.KubevirtPlatformType {
+	if infra.Status.PlatformStatus.Type == configv1.OvirtPlatformType || infra.Status.PlatformStatus.Type == configv1.KubevirtPlatformType || infra.Status.PlatformStatus.Type == configv1.LibvirtPlatformType {
 		t.unsupportedPlatform = true
 	}
 	if t.unsupportedPlatform {


pweil-:
Looks like tests are green now :+1: 

I see no issue with this addition but I'd like @mfojtik or @smarterclayton to provide final approval.  

/assign @smarterclayton 
/assign @mfojtik

r4f4:
@smarterclayton @mfojtik ping

r4f4:
@smarterclayton @mfojtik ping

jaypoulz:
@soltysh @pweil- Would either of you be able to give final approval? We're blocked on upgrade testing by this and I believe the assignees are occupied with other items.

pweil-:
@jaypoulz Sorry for the delay, I've been on PTO.  I've pinged @soltysh on Slack to see if this is something he'd be comfortable approving.

smarterclayton:
/hold

I think we'll want to review what tests don't work and why.  Is that something already recorded somewhere?  In general, openshift is supposed to work the same even on metal, so I need to know what the delta is to advise better.

jaypoulz:
@r4f4 can advise further, but this test suite disables the load balancer tests because libvirt doesn't not have a compatible load-balancing strategy.

Otherwise, the tests do work as advertised. @r4f4 - can you expand further with the details of your investigation?

r4f4:
> @r4f4 can advise further, but this test suite disables the load balancer tests because libvirt doesn't not have a compatible load-balancing strategy.
> 
> Otherwise, the tests do work as advertised. @r4f4 - can you expand further with the details of your investigation?

Exactly. If run as-is the upgrade tests fail with:

```
INFO: Running AfterSuite actions on node 1
fail [k8s.io/kubernetes/test/e2e/framework/service/jig.go:566]: Jun  4 06:04:13.983: Timed out waiting for service "service-test" to have a load balancer

failed: (1h0m55s) 2020-06-04T10:55:05 "[Disruptive] Cluster upgrade should maintain a functioning cluster [Feature:ClusterUpgrade] [Suite:openshift] [Serial]"
```

smarterclayton:
You should add a platform specific skip / alternate logic for baremetal in the service test.  Ultimately this is a gap in baremetal that should be solved by metalLB ootb (as described here https://github.com/openshift/enhancements/blob/508d56c65f832e30a27a3ded4d971d755e7be38f/enhancements/network/on-prem-service-load-balancers.md).  When that happens we would reenable this test.

So platform specific skip instead of a new suite (if that's the only failing test)

r4f4:
@smarterclayton I've changed the code as you suggested. Can you review it again, please?

r4f4:
All tests green, just waiting for approval now.

r4f4:
@smarterclayton ping

PR #25592, opened 2020-10-08
test: Increase allowed telemetry series by 100

smarterclayton:
We recently increased the series published to telemetry by about
100 (to include resource counts for all CRDs). Bump the allowed
test limits similarly to 600.

Consequence of openshift/cluster-monitoring-operator#951


lilic (commented):
/lgtm

PR #25593, opened 2020-10-08
Squash all oc execute methods

soltysh:


soltysh:
Incorporated in https://github.com/openshift/origin/pull/25169

PR #25596, opened 2020-10-08
provide operator upgrade junit results

deads2k:
Provides accurate accounting of individual operators for upgrade and provides durations for how long they take to upgrade.

### DIFF
--- a/test/extended/operators/clusteroperator_upgrade.go
+++ b/test/extended/operators/clusteroperator_upgrade.go
@@ -1,0 +1,217 @@
+package operators
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/openshift/origin/test/extended/util/disruption"
+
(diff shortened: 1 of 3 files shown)


deads2k:
/hold

hold until we see the junit results and like them.

test/extended/operators/clusteroperator_upgrade.go
```diff
@@ -0,0 +1,171 @@
...
+			)
+		}
+	}
+
+	return ret
+}
+
+func minorVersionFor(version string) string {
+	// someday, if you're here because this product is around this long and for loop ran out of matches, send an email
+	// to deads2k from github because I'm horrified that this code lasted that long
```
bparees, line 37: seriously. regex. :)

smarterclayton: Wow

deads2k:
that's weird.  Target version is empty....  It must get filled in after an upgrade request and then read back out...

PR #25602, opened 2020-10-12
Report API server unavailability in e2e tests

jsafrane:
Currently, API server availabilities are tracked under `[sig-arch] Monitor cluster while tests execute`. Create a separate test for them when the API servers are unreachable more than 1% of the test time.

In addition:
* Print status of synthetic tests to stdout for better debug-ability of these test.
* Change monitor interval to 1s to be able to detect outages shorter than 15s.

~WIP: the third commit actually injects some fake errors to see how it works in prow.~

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -24,6 +24,15 @@ import (
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 )
 
+const (
+	LocatorKubeAPIServerNewConnection         = "kube-apiserver-new-connection"
+	LocatorOpenshiftAPIServerNewConnection    = "openshift-apiserver-new-connection"
+	LocatorOAuthAPIServerNewConnection        = "oauth-apiserver-new-connection"
+	LocatorKubeAPIServerReusedConnection      = "kube-apiserver-reused-connection"
+	LocatorOpenshiftAPIServerReusedConnection = "openshift-apiserver-reused-connection"
+	LocatorOAuthAPIServerReusedConnection     = "oauth-apiserver-reused-connection"
+)
+
 // Start begins monitoring the cluster referenced by the default kube configuration until
 // context is finished.
 func Start(ctx context.Context) (*Monitor, error) {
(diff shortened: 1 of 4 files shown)


pkg/test/ginkgo/synthentic_tests.go
```diff
@@ -84,10 +90,44 @@ func createSyntheticTestsFromMonitor(m *monitor.Monitor, eventsForTests []*monit
...
 
+	fmt.Fprintln(buf, "Synthetic test results:")
+	for _, test := range syntheticTestResults {
+		status := "passed"
+		if test.FailureOutput != nil {
+			status = "failed"
+		}
+		fmt.Fprintf(buf, "%s: %s\n", status, test.Name)
+	}
+	fmt.Fprintln(buf, "(duplicities are used to mark some failures as flake and not to fail the whole suite")
```
deads2k, line 105: duplicates

test/extended/util/disruption/disruption.go
```diff
@@ -280,9 +280,21 @@ func createTestFrameworks(tests []upgrades.Test) map[string]*framework.Framework
...
 func ExpectNoDisruption(f *framework.Framework, tolerate float64, total time.Duration, events monitor.EventIntervals, reason string) {
+	duration, describe := GetDisruption(events, "")
+	if percent := float64(duration) / float64(total); percent > tolerate {
+		framework.Failf("%s for at least %s of %s (%0.0f%%):\n\n%s", reason, duration.Truncate(time.Second), total.Truncate(time.Second), percent*100, strings.Join(describe, "\n"))
+	} else if duration > 0 {
+		FrameworkFlakef(f, "%s for at least %s of %s (%0.0f%%), this is currently sufficient to pass the test/job but not considered completely correct:\n\n%s", reason, duration.Truncate(time.Second), total.Truncate(time.Second), percent*100, strings.Join(describe, "\n"))
+	}
+}
+
+func GetDisruption(events monitor.EventIntervals, requiredLocator string) (time.Duration, []string) {
```
deads2k, line 291: I haven't looked at this before.  Can you provide an example eventInterval that shows what you're looking for in this list?

test/extended/util/disruption/disruption.go
```diff
@@ -280,9 +280,21 @@ func createTestFrameworks(tests []upgrades.Test) map[string]*framework.Framework
...
 	var duration time.Duration
 	var describe []string
 	for _, interval := range events {
+		if requiredLocator != "" && requiredLocator != interval.Locator {
+			continue
+		}
 		describe = append(describe, interval.String())
 		i := interval.To.Sub(interval.From)
```
deads2k, line 299: Can I buy some more letters?  I know it's prexisting, but you must know what this means.  What is `i` in this context?  An event summary of something?

This flow suggests some other entity is injecting failed connections into an event stream?

pkg/test/ginkgo/synthentic_tests.go
```diff
@@ -84,10 +90,44 @@ func createSyntheticTestsFromMonitor(m *monitor.Monitor, eventsForTests []*monit
 	syntheticTestResults = append(syntheticTestResults, testPodTransitions(events)...)
 	syntheticTestResults = append(syntheticTestResults, testSystemDTimeout(events)...)
 	syntheticTestResults = append(syntheticTestResults, testPodSandboxCreation(events)...)
+	syntheticTestResults = append(syntheticTestResults, testServerAvailability("kube-apiserver", events, monitorDuration)...)
```
deads2k, line 93: If I'm reading this correctly, the `kube-apiserver` is a well-known name somewhere that is used solely to indicate availability?

pkg/test/ginkgo/synthentic_tests.go
```diff
@@ -8,6 +8,12 @@ import (
 	"time"
 
 	"github.com/openshift/origin/pkg/monitor"
+	"github.com/openshift/origin/test/extended/util/disruption"
+)
+
+const (
+	// Max. duration of API server unreachability, in percents of total test duration.
+	tolerateDisruptionPercent = 1.0
```
deads2k, line 16: 1.0 here mean, "never fail yet", right?  Can I get a comment with a TODO to bring us inline with upgrade tests at 0.08?

jsafrane: percent != percent. In my code, it's really 1% of time it can fail. It upgrade tests, 0.08 means 8%. I can fix to 0.01, but calling something as percentage and using fraction instead is very confusing.

test/extended/util/disruption/disruption.go
jsafrane: Well, I thought you would understand the code more than I am :-).

Apparently, `Monitor` created here https://github.com/openshift/origin/blob/29b338da5bb7fc4309d5b4fe120e3cca61e1daae/pkg/monitor/api.go#L31
watches given Samplers (see `StartKubeAPIMonitoring()`) every second. Once the Sampler becomes ok -> failed -> back ok, Monitor records event "Kube API is not responding to GET requests" with:
* Locator: `kube-apiserver`
* From: time when the Monitor detected the Sampler went ok -> bad
* To: time when the Monitor detected the Sampler went bad -> ok

It took me a looong while to find out this. Still, you can actually see the events in the usual `Monitor cluster while tests execute` output:

```
Oct 12 15:24:26.863 E kube-apiserver Kube API started failing: namespaces "kube-system" not found
Oct 12 15:24:27.759 - 5s    E kube-apiserver Kube API is not responding to GET requests
Oct 12 15:24:33.861 I kube-apiserver Kube API started responding to GET requests
```

So the filter there catches "kube-apiserver Kube API is not responding to GET requests". BTW, you can see that the monitor realizes it almost 1 second after other part of the monitor detected the actual error. The whole outage was 7 seconds, not 5. The whole monitoring code is very confusing. There must have been some reason to make it so...

pkg/test/ginkgo/synthentic_tests.go
jsafrane: fixed

test/extended/util/disruption/disruption.go
jsafrane: fixed

pkg/test/ginkgo/synthentic_tests.go
jsafrane: Added consts to `monitor` pkg.

pkg/test/ginkgo/synthentic_tests.go
jsafrane: changed to 0.01

jsafrane:
Changed test name from "server kube-apiserver is available" to "kube-apiserver should be available"

test/extended/util/disruption/disruption.go
jsafrane: EventInterval is basically what's printed in `Monitor cluster while tests execute output`:

```
Oct 12 15:24:26.863 E kube-apiserver Kube API started failing: namespaces "kube-system" not found
Oct 12 15:24:27.759 - 5s    E kube-apiserver Kube API is not responding to GET requests
Oct 12 15:24:33.861 I kube-apiserver Kube API started responding to GET requests
```

Showing "From" above as the timestamp. Most EventIntervals have `From == To` (== when the event happened), but `kube-apiserver Kube API is not responding to GET requests` has almost the right `From`/`To`  (both +/- 1s) for total duration of the outage.

jsafrane:
Removed the last commit with error injection. You can see its output at https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25602/pull-ci-openshift-origin-master-e2e-gcp/1315968061345370112: 

```

[sig-api-machinery] kube-apiserver should be available
Run #0: Failed 
kube-apiserver was failing for 1m37s seconds (6% of the test duration)
```

jsafrane:
Looking at the test results, [couple of e2e-aws-csi tests failed with `connection refused`, yet I don't see any outage reported by the new test](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25602/pull-ci-openshift-origin-master-e2e-aws-csi/1316293112649224192). The failed tests were the first started, so I added a commit that checks for connectivity immediately, not after 1 second.

deads2k:
/hold

Something has gone wrong in this PR.  The `[sig-arch] Monitor cluster while tests execute` should show dozens of errors as a flake on every run.  In master it does.  I think this PR does something which is making it pass.

pkg/monitor/api.go
```diff
@@ -24,10 +24,19 @@ import (
...
+	LocatorKubeAPIServerReusedConnection      = "kube-apiserver-reused-connection"
+	LocatorOpenshiftAPIServerReusedConnection = "openshift-apiserver-reused-connection"
+	LocatorOAuthAPIServerReusedConnection     = "oauth-apiserver-reused-connection"
+)
+
 // Start begins monitoring the cluster referenced by the default kube configuration until
 // context is finished.
 func Start(ctx context.Context) (*Monitor, error) {
-	m := NewMonitor()
+	m := NewMonitorWithInterval(time.Second)
```
deads2k, line 39: maybe this is somehow not causing the "normal" flakes we see in the e2e monitors in master?

deads2k:
In the upgrade the duration of `[sig-api-machinery] openshift-apiserver-new-connection should be available` is also wrong.  it's showing zero, but it should be the full time of the test run.

jsafrane:
There is something weird with upgrade tests, but I don't think it's this PR. This job run is for a different PR and passed without any flakes: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25646/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1321833443197194240. And I found others too.

jsafrane:
With https://github.com/openshift/origin/pull/25644 merged all tests are flaky again.

deads2k:
/lgtm
/hold

holding in case you want to be sure that it's testing how you like.  Zero second duration is not the same as the upgrade job.  Feel free to release.

PR #25604, opened 2020-10-12
remove retry

deads2k:
preview of https://github.com/openshift/kubernetes/pull/405


deads2k:
/hold

we'll get this after https://github.com/openshift/kubernetes/pull/405 merges and we bump

PR #25605, opened 2020-10-13
tests/extended/OWNERS: add ashcrow to approvers

cgwalters:
As pillar lead for MCO/CoreOS - we will likely do more
tests here and so should have an approver.


ashcrow:
I approve of this :drum:

cgwalters:
Changing OWNERS can't break tests
/override ci/prow/e2e-gcp

PR #25542, opened 2020-09-23
Bug 1879445: Add a route annotation for Samesite

miheer:
1. Add a route annotation for Samesite
https://issues.redhat.com/browse/RFE-600
Bug 1879445 - SameSite hardening breaks sticky load balancing

2. Adding notes that the upstream PR breaks product builds and we now have an "official" way to get the latest haproxy RPM
we reverted their PR because it breaks Red Hat product builds

@jupierce @frobware

PR #25613, opened 2020-10-20
[DO NOT MERGE] Test Flakes

miheer:
test flakes


frobware:
This is to try and gauge whether https://github.com/openshift/origin/pull/25542 has legitimate CI failures.

PR #25614, opened 2020-10-20
Evaluate storage metrics of an e2e test.

jsafrane:
At the end of e2e test, grab storage operation metrics and check that attach and mount times are under expected limits.

To get there, refactor `RunQueries` to return error instead of failing the test, so storage metric checks can declare themselves as test flakes instead of failures.

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
 
 	exutil "github.com/openshift/origin/test/extended/util"
 	helper "github.com/openshift/origin/test/extended/util/prometheus"
@@ -91,7 +92,8 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 		criticalAlertQuery: false,
 	}
 
-	helper.RunQueries(tests, t.oc, ns, execPod.Name, t.url, t.bearerToken)
+	err := helper.RunQueries(tests, t.oc, ns, execPod.Name, t.url, t.bearerToken)
+	o.Expect(err).NotTo(o.HaveOccurred())
 
 	framework.Logf("No critical alerts firing post-upgrade")
 }
--- a/test/extended/prometheus/storage.go
+++ b/test/extended/prometheus/storage.go
@@ -1,0 +1,74 @@
+package prometheus
+
+import (
+	"context"
+	"fmt"
+
+	g "github.com/onsi/ginkgo"
+	"github.com/openshift/origin/pkg/test/ginkgo/result"
+	exutil "github.com/openshift/origin/test/extended/util"
+	helper "github.com/openshift/origin/test/extended/util/prometheus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+const (
+	// storage_operation_duration_seconds_bucket with the mount operations that are considered acceptable.
+	expectedMountTimeSeconds = 120
+	// storage_operation_duration_seconds_bucket with the attach operations that are considered acceptable.
+	expectedAttachTimeSeconds = 120
+)
+
+var _ = g.Describe("[sig-storage][Late] Metrics", func() {
+	defer g.GinkgoRecover()
+	var (
+		oc = exutil.NewCLIWithoutNamespace("prometheus")
+
+		url, bearerToken string
+	)
+
+	g.BeforeEach(func() {
+		var ok bool
+		url, bearerToken, ok = helper.LocatePrometheus(oc)
+		if !ok {
+			e2e.Failf("Prometheus could not be located on this cluster, failing prometheus test")
+		}
+	})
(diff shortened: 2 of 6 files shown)


jsafrane:
cc @openshift/openshift-team-monitoring @openshift/storage PTAL. The overall goal of this PR is to add some context to flaky sig-storage tests when attach/mount takes too long.

test/extended/prometheus/storage.go
```diff
@@ -0,0 +1,67 @@
...
+	}()
+
+	tests := map[string]bool{}
+	// Check "[total nr. of ops] - [nr. of ops < threshold] > 0' and expect failure (all ops should be < threshold).
+	// Using sum(max(...)) to sum all kubelets / controller-managers
+	// Adding a comment to make the failure more readable.
+	e2e.Logf("Checking that Operation %s time of plugin %s should be < % seconds", name, plugins, threshold)
+	queryTemplate := `
+# Operation %[4]s time of plugin %[1]s should be < %[2]d seconds
+  sum(max_over_time(storage_operation_duration_seconds_bucket{job="%[3]s",le="+Inf",operation_name="%[4]s",volume_plugin="%[1]s"}[2h]))
```
gnufied, line 54: Why do we need to sum the `max_over_time` value?

test/extended/prometheus/storage.go
```diff
@@ -0,0 +1,67 @@
...
+		if !ok {
+			e2e.Failf("Prometheus could not be located on this cluster, failing prometheus test")
+		}
+	})
+	g.It("should report short attach times", func() {
+		checkOperation(oc, url, bearerToken, "kube-controller-manager", "volume_attach", 120)
+	})
+	g.It("should report short mount times", func() {
+		checkOperation(oc, url, bearerToken, "kubelet", "volume_mount", 120)
+	})
```
gnufied, line 35: can we move 120 to a constant please?

test/e2e/upgrade/alert/alert.go
```diff
@@ -91,7 +92,8 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 		criticalAlertQuery: false,
 	}
 
-	helper.RunQueries(tests, t.oc, ns, execPod.Name, t.url, t.bearerToken)
+	err := helper.RunQueries(tests, t.oc, ns, execPod.Name, t.url, t.bearerToken)
+	o.Expect(err).NotTo(o.HaveOccurred())
```
gnufied, line 96: Why this btw?

jsafrane: because `RunQueries` had `because `o.Expect(err).NotTo(o.HaveOccurred())` inside and now it returns `error`, so storage metric tests can mark themselves as flaky instead of failed.

test/extended/prometheus/storage.go
jsafrane: `max_over_time` should return the max. nr of operations in given bucket, i.e. the last value in the time series (let me know if there is a better way to get the last value). `sum` then sums nr. of operations from all kubelets / kube-controller-managers.

test/extended/prometheus/storage.go
jsafrane: done

test/extended/util/prometheus/helpers.go
```diff
@@ -171,7 +174,11 @@ func RunQueries(promQueries map[string]bool, oc *exutil.CLI, ns, execPodName, ba
 	if len(queryErrors) != 0 {
 		exutil.DumpPodLogsStartingWith("prometheus-0", oc)
 	}
-	o.Expect(queryErrors).To(o.BeEmpty())
+	var errs []error
+	for _, err := range queryErrors {
+		errs = append(errs, err)
```
bertinatto, line 179: The previous code (`o.Expect(queryErrors).To(o.BeEmpty())`) was outputting the `query => error` map, but now we're discarding the query information (the key). Perhaps we should do something like `errs = append(errs, fmt.Errorf("Query %q failed: %v", query, err))`.

test/extended/prometheus/storage.go
```diff
@@ -0,0 +1,74 @@
...
+	execPod := exutil.CreateUbiExecPodOrFail(oc.AdminKubeClient(), ns, "execpod", nil)
+	defer func() {
+		oc.AdminKubeClient().CoreV1().Pods(ns).Delete(context.Background(), execPod.Name, *metav1.NewDeleteOptions(1))
+	}()
+
+	tests := map[string]bool{}
+	// Check "[total nr. of ops] - [nr. of ops < threshold] > 0' and expect failure (all ops should be < threshold).
+	// Using sum(max(...)) to sum all kubelets / controller-managers
+	// Adding a comment to make the failure more readable.
+	e2e.Logf("Checking that Operation %s time of plugin %s should be < % seconds", name, plugins, threshold)
```
bertinatto, line 58: Isn't this missing a `%d` for `thereshold`?

test/extended/prometheus/storage.go
simonpasquier: It sounds right.

jsafrane:
/assign @tnozicka @soltysh 
for approval

soltysh (approved):
/approve

PR #25619, opened 2020-10-21
WIP: Remove kubectl call from storage tests

jsafrane:
To diagnose https://bugzilla.redhat.com/show_bug.cgi?id=1890131, replace all `kubectl` calls from storage tests with one HTTP POST.

WIP: just test it here, then clean it a bit and post upstream.


jsafrane:
/close
Moved upstream as https://github.com/kubernetes/kubernetes/pull/95799

PR #25620, opened 2020-10-21
check apiserver uptime with both new and re-used connections

deads2k:
Uptime for apiservers can look different between new connections and existing connections.  This adds checks for new connections and runs both sets of tests so we can start to see which fail when.

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -3,24 +3,25 @@ package monitor
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
 	"strings"
 	"sync"
 	"time"
 
+	"k8s.io/client-go/transport"
+
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
-	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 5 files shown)


jsafrane:
Most tests failed with:
```
error: using a custom transport with TLS certificate options or the insecure flag is not allowed error: failed to execute wrapped command: exit status 1
--


```

test/extended/util/disruption/controlplane/controlplane.go
```diff
@@ -14,77 +14,87 @@ import (
...
-
-func (t openShiftAvailableTest) Name() string { return t.availableTest.name }
-func (openShiftAvailableTest) DisplayName() string {
-	return "[sig-api-machinery] OpenShift APIs remain available"
-}
-func (t *openShiftAvailableTest) Test(f *framework.Framework, done <-chan struct{}, upgrade upgrades.UpgradeType) {
-	t.availableTest.test(f, done, upgrade, monitor.StartOpenShiftAPIMonitoring)
+func NewOpenShiftAvailableWithConnectionReuseTest() upgrades.Test {
+	return &availableTest{
+		testName:        "[sig-api-machinery] OpenShift APIs remain available for new connections",
```
sttts, line 56: wrong string

sttts:
Why do we need the connection reuse tests at all?

Also note that Golang 1.15 does not gracefully shutdown http2 connection (@p0lyn0mial has submitted a PR), i.e. you will get flakes when the apiserver stops the server. We should rather check `/readyz` in these tests and shutdown the existing connection ourself.

pkg/monitor/api.go
```diff
@@ -59,117 +68,116 @@ func Start(ctx context.Context) (*Monitor, error) {
...
 
-func StartKubeAPIMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.Config, timeout time.Duration) error {
-	pollingConfig := *clusterConfig
-	pollingConfig.Timeout = timeout
-	pollingClient, err := clientcorev1.NewForConfig(&pollingConfig)
-	if err != nil {
-		return err
-	}
+func StartServerMonitoringWithNewConnections(ctx context.Context, m *Monitor, clusterConfig *rest.Config, timeout time.Duration, resourceLocator, url string) error {
+	return startServerMonitoring(ctx, m, clusterConfig, timeout, resourceLocator, url, true)
```
mfojtik, line 72: nit: can we make `StartServerMonitoring` public?

sttts:
/approve
Lgtm

PR #25625, opened 2020-10-23
Bug 1882556: Access git over https for tests

coreydaley-redhat:
Replace git:// urls with https:// urls for all tests

### DIFF
--- a/test/extended/builds/clone_git_protocol.go
+++ b/test/extended/builds/clone_git_protocol.go
@@ -1,0 +1,48 @@
+package builds
+
+import (
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-builds][Feature:Builds] clone repository using git:// protocol", func() {
+	var (
+		oc = exutil.NewCLI("build-clone-git-protocol")
+	)
+
+	g.Context("", func() {
+
+		g.BeforeEach(func() {
+			exutil.PreTestDump()
+		})
+
+		g.JustBeforeEach(func() {
+			g.By("waiting for openshift namespace imagestreams")
+			err := exutil.WaitForOpenShiftNamespaceImageStreams(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+		})
+
+		g.AfterEach(func() {
+			if g.CurrentGinkgoTestDescription().Failed {
+				exutil.DumpPodStates(oc)
+				exutil.DumpConfigMapStates(oc)
+				exutil.DumpPodLogsStartingWith("", oc)
+			}
+		})
+
+		g.It("should clone using git:// if no proxy is configured", func() {
+			proxyConfigured, err := exutil.IsClusterProxyEnabled(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if proxyConfigured {
+				g.Skip("Skipping test because proxy is configured")
+			}
+
+			g.By("creating a new application using the git:// protocol")
+			err = oc.Run("new-app").Args("git://github.com/sclorg/ruby-ex.git").Execute()
+			o.Expect(err).NotTo(o.HaveOccurred())
+		})
+	})
+})
--- a/test/extended/util/proxy.go
+++ b/test/extended/util/proxy.go
@@ -1,0 +1,20 @@
+package util
+
+import (
+	"context"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// IsClusterProxyEnabled returns true if the cluster has a global proxy enabled
+func IsClusterProxyEnabled(oc *CLI) (bool, error) {
+	proxy, err := oc.AdminConfigClient().ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if kerrors.IsNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return len(proxy.Status.HTTPProxy) > 0 || len(proxy.Status.HTTPSProxy) > 0, nil
+}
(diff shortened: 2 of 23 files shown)


adambkaplan (commented):
I'd like to ensure that we have at least one e2e test that pulls via the `git:` protocol. This test would need to check that we have a cluster-wide proxy configured before running.

coreydaley-redhat:
> I'd like to ensure that we have at least one e2e test that pulls via the `git:` protocol. This test would need to check that we have a cluster-wide proxy configured before running.

So, we would need to fix the proxy (according to the bugzilla) so that git:// urls can be proxied.  Should that be a JIRA then as it is probably a bit more work.  Or are you saying that we need a test that uses a git:// url and make sure that the proxy is NOT configured so that it has a chance to work?

adambkaplan:
> Or are you saying that we need a test that uses a git:// url and make sure that the proxy is NOT configured so that it has a chance to work?

Correct - we should have a test that clones source via `git://`, but when the test starts it checks if a proxy has been configured, skipping if a proxy is present. This should be fairly straightforward (psudo-code here):

```go
func hasProxy() (bool, error) {
    proxy, err := e.ConfigClient.Proxies().Get(ctx, "cluster", metav1.GetOptions{})
    if err != nil && !kerrors.IsNotFound(err) {
        return false, error
    }
    if kerrors.IsNotFound(err) {
        return false, nil
    }
    return len(proxy.Status.HTTPProxy) > 0 || len(proxy.Status.HTTPSProxy) > 0, nil
}
```

coreydaley-redhat:
@adambkaplan ptal

test/extended/builds/clone_git_protocol.go
```diff
@@ -0,0 +1,51 @@
...
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-builds][Feature:Builds] clone repository using git:// protocol", func() {
+	var (
+		oc = exutil.NewCLI("build-pruning")
```
adambkaplan, line 14: nit - this is the prefix for the generated namespace.
```suggestion
		oc = exutil.NewCLI("build-clone-git-protocol")
```

test/extended/builds/clone_git_protocol.go
```diff
@@ -0,0 +1,51 @@
...
+		g.It("should clone using git:// if no proxy is configured", func() {
+			proxyConfigured, err := exutil.HasProxy(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if proxyConfigured {
+				fmt.Fprintf(g.GinkgoWriter, "Skipping test because proxy is configured")
+				return
+			}
+
+			g.By("creating a new application using the git:// protocol")
+			err = oc.Run("new-app").Args("git://github.com/openshift/ruby-hello-world.git").Execute()
```
adambkaplan, line 47: Historically we've had issues with `openshift/ruby-hello-world` - the scl example is a bit more stable:

```suggestion
			err = oc.Run("new-app").Args("git://github.com/sclorg/ruby-ex.git").Execute()
```

test/extended/util/proxy.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// HasProxy returns whether or not the currently installed cluster has a proxy configured
+func HasProxy(oc *CLI) (bool, error) {
+	proxy, err := oc.AdminConfigClient().ConfigV1().Proxies().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil && !kerrors.IsNotFound(err) {
+		return false, err
+	} else if kerrors.IsNotFound(err) {
+		return false, nil
+	}
```
adambkaplan, line 17: Looks like `kerrors.IsNotFound` returns `false` if the input is nil, so we can simplify a bit:

```suggestion
	if kerrors.IsNotFound(err) {
		return false, nil
	}
	if err != nil {
		return false, err
	}
```

adambkaplan (commented):
/approve

Minor suggestions, otherwise looks good.

test/extended/util/proxy.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"context"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// HasProxy returns whether or not the currently installed cluster has a proxy configured
+func HasProxy(oc *CLI) (bool, error) {
```
adambkaplan, line 11: Perhaps need to make it clearer that we're talking about the cluster?

```suggestion
// IsClusterProxyEnabled returns true if the cluster has a global proxy enabled
func IsClusterProxyEnabled(oc *CLI) (bool, error) {
```

test/extended/builds/clone_git_protocol.go
coreydaley-redhat: updated in next push

test/extended/builds/clone_git_protocol.go
coreydaley-redhat: updated in next push

test/extended/util/proxy.go
coreydaley-redhat: updated in next push

test/extended/util/proxy.go
coreydaley-redhat: updated in next push

coreydaley-redhat:
@adambkaplan all suggestions have been applied, ptal

adambkaplan (approved):
/lgtm

wewang58:
Using cluster-bot to launch cluster to test it
/lgtm

test/extended/builds/clone_git_protocol.go
```diff
@@ -0,0 +1,51 @@
...
+				exutil.DumpPodLogsStartingWith("", oc)
+			}
+		})
+
+		g.It("should clone using git:// if no proxy is configured", func() {
+			proxyConfigured, err := exutil.IsClusterProxyEnabled(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if proxyConfigured {
+				fmt.Fprintf(g.GinkgoWriter, "Skipping test because proxy is configured")
+				return
```
soltysh, line 43: `g.Skip` should be used instead.

soltysh: or `e2eskipper.Skipf`.

soltysh (changes requested):
/lgtm cancel

test/extended/builds/clone_git_protocol.go
coreydaley-redhat: updated in current push

coreydaley-redhat:
@soltysh Updated per your comment, ptal.

soltysh (approved):
/lgtm
/approve

PR #25627, opened 2020-10-23
BUG 1891106: UPSTREAM: 95259: allocate service-account flowschema to global-default

tkashem:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

All workloads running using service account (except for the ones distinguished by p&f with a logically higher matching precedence) will match the `service-accounts` flow schema and be assigned to the `workload-low` priority and thus will have only `20` concurrency shares. (~10% of the total)

On the other hand, `global-default` flow schema is assigned to `global-default` priority configuration and thus will have `100` concurrency shares (~50% of the total). If I am not mistaken, `global-default` goes pretty much unused since workloads running with user (not service account) will fall into this category and is not very common. 

Workload with service accounts do not have enough concurrency share and may starve. Increase the concurrency share of `workload-low` from `20` to `100` and reduce that of `global-default` from `100` to `20`. 

Another potential solution: maybe when a concurrency pool is starving, it can borrow from another pool that has concurrency shares to spare? 

```release-note
NONE
```

### DIFF
--- a/vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
+++ b/vendor/k8s.io/kubernetes/staging/src/k8s.io/apiserver/pkg/apis/flowcontrol/bootstrap/default.go
@@ -210,7 +210,7 @@ var (
 		flowcontrol.PriorityLevelConfigurationSpec{
 			Type: flowcontrol.PriorityLevelEnablementLimited,
 			Limited: &flowcontrol.LimitedPriorityLevelConfiguration{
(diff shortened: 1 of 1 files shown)


xingxingxia:
Tried pre-merge verification process in [1891106#c4](https://bugzilla.redhat.com/show_bug.cgi?id=1891106#c4) but failed twice in env installation. Let me continue working on that.

xingxingxia:
@tkashem this PR's pre-merge verification is blocked by the installation failure in [1891106#c5](https://bugzilla.redhat.com/show_bug.cgi?id=1891106#c5)

PR #25628, opened 2020-10-23
Bug 1887850: update SCC test to either pass or fail, but never flake

deads2k:
This will provide a clear signal on regression and prevent unnecessary noise.

Because the CVO starts as a static pod, concurrent with the cluster-policy-controller, there is a time period when namespaces can be created and content placed in them by the CVO before the controller assigning SCC UIDs has had a chance to actually create the range allocations.  Below a certain threshold (I chose five), this is acceptable.  We were dealing with dozens in the past.

### DIFF
--- a/test/extended/authorization/scc.go
+++ b/test/extended/authorization/scc.go
@@ -7,7 +7,6 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


marun:
/retitle Bug 1887850: update SCC test to either pass or fail, but never flake

deads2k:
e2e, no cmd impact

PR #25638, opened 2020-10-27
Move debug.sh & timeout.sh to a proper e2e

soltysh:
In comparison with the original I:
- fixed the messages checked in debug.sh which somehow were completely wrong :scream_cat: 
- fixed the messages which were responsible for many timeout.sh failures
- removed some duplicate tests for `--keep-readiness` and `--keep-liveness` b/c they were checking empty string

/assign @sallyom


hardys:
Looks like backporting this will solve https://bugzilla.redhat.com/show_bug.cgi?id=1998963

soltysh:
@hardys you'll likely need to pick manually what you care about since that's one of many PRs rewriting the tests, so it might not be that simple to backport :disappointed:

hardys:
Thanks @soltysh - I was mistaken anyway - the change I need was in https://github.com/openshift/origin/pull/26415 where the hard-coded busybox image was removed on master, but we're now seeing failures in 4.9 jobs due to the check for only known images

PR #25641, opened 2020-10-27
Bug 1881319: IPtable rules are racy when pod is created before svc

tssurya:
We have a race condition because the `proxier.changesPending` which
was introduced in 0c9e44a08d12c115c72900af6c7513a7d178ea2b is not
locked properly. So most of the times in half of the SDN pods, the
REJECT rule gets created for not finding an endpoint and it does
not get replaced by the correct rules because the `SyncProxyRules`
does not get run because the value of `proxier.changesPending` is racy.

As a consequence of this, until there is a next update that sets
the value of `proxier.changesPending` correctly, the service rejects
requests from users. This patch uses the `proxier.mu.Lock()` on
`proxier.changesPending` when its changed from `Sync()`.
`OnEndpointsUpdate` and `OnServiceUpdate` then in turn call `Sync()`.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go
+++ b/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go
@@ -445,7 +445,9 @@ func (proxier *Proxier) probability(n int) string {
 
 // Sync is called to synchronize the proxier state to iptables as soon as possible.
 func (proxier *Proxier) Sync() {
+	proxier.mu.Lock()
 	proxier.changesPending = true
+	proxier.mu.Unlock()
 	proxier.syncRunner.Run()
 }
 
@@ -476,8 +478,7 @@ func (proxier *Proxier) OnServiceAdd(service *api.Service) {
 
 func (proxier *Proxier) OnServiceUpdate(oldService, service *api.Service) {
 	if proxier.serviceChanges.Update(oldService, service) && proxier.isInitialized() {
-		proxier.changesPending = true
-		proxier.syncRunner.Run()
+		proxier.Sync()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


danwinship:
> Note that the reverse case race was not reproducible (svc created before pod)

hm... I wouldn't rely on that; even if it's true now it could get broken by a later change. Better to use the mutex in all three places. (Well, or change `OnServiceUpdated` and `OnEndpointsUpdated` to just call `proxier.Sync()` rather than duplicating its code and then fix it in `Sync`.)

danwinship:
FTR openshift/sdn `release-4.3` has the same bug but we're no longer pushing updates there. 4.4 and later have the real upstream version of this code which didn't need the `changesPending` hack

tssurya:
> > Note that the reverse case race was not reproducible (svc created before pod)
> 
> hm... I wouldn't rely on that; even if it's true now it could get broken by a later change. Better to use the mutex in all three places. (Well, or change `OnServiceUpdated` and `OnEndpointsUpdated` to just call `proxier.Sync()` rather than duplicating its code and then fix it in `Sync`.)

hmm ack yea. I just wanted a second opinion on whether this should be done or not, I'll update the patch.

tssurya:
/test e2e-gcp
/test extended_conformance_install

failure seems unrelated to the patch.

tssurya:
the failing CI is unrelated: https://bugzilla.redhat.com/show_bug.cgi?id=1902067
We should skip and merge this AFAIK.

knobunc:

This test has been failing consistently across all PRs https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/test_pull_request_origin_extended_conformance_install-release-3.11 and has a bug to track it.

knobunc:

This test has been failing consistently across all PRs https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/test_pull_request_origin_extended_conformance_install-release-3.11 and has a bug to track it (https://bugzilla.redhat.com/show_bug.cgi?id=1902067)

PR #25652, opened 2020-10-31
Add e2e testing of server-side apply for openshift types 

marun:
TODO
 - [x] Fix conversion errors for `oauth.openshift.io/v1` and `user.openshift.io/v1` groups
    - These groups are served by oauth-apiserver

/cc @sttts @stlaz @p0lyn0mial

### DIFF
--- a/test/extended/apiserver/apply.go
+++ b/test/extended/apiserver/apply.go
@@ -24,8 +24,6 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
-// TODO(marun) Evaluate the CRDs that ocp with (~70 at time of
-// writing) which will involve creating the associated stubs.
 var _ = g.Describe("[sig-api-machinery][Feature:ServerSideApply] Server-Side Apply", func() {
 	var (
 		// mapper is used to translate the gvr provided by etcd
--- a/test/extended/apiserver/apply.go
+++ b/test/extended/apiserver/apply.go
@@ -1,0 +1,176 @@
+package apiserver
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	discocache "k8s.io/client-go/discovery/cached"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/restmapper"
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	exetcd "github.com/openshift/origin/test/extended/etcd"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+// TODO(marun) Evaluate the CRDs that ocp with (~70 at time of
+// writing) which will involve creating the associated stubs.
+var _ = g.Describe("[sig-api-machinery][Feature:ServerSideApply] Server-Side Apply", func() {
+	var (
+		// mapper is used to translate the gvr provided by etcd
+		// storage data to the gvk required to create correct resource
+		// yaml.
+		mapper *restmapper.DeferredDiscoveryRESTMapper
+	)
+
+	defer g.GinkgoRecover()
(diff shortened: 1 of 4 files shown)


test/extended/apiserver/apply.go
```diff
@@ -0,0 +1,168 @@
...
+	oc := exutil.NewCLIWithoutNamespace("server-side-apply")
+
+	g.BeforeEach(func() {
+		// Only init once per worker
+		if mapper != nil {
+			return
+		}
+		kubeClient, err := kubernetes.NewForConfig(oc.AdminConfig())
+		o.Expect(err).NotTo(o.HaveOccurred())
+		mapper = restmapper.NewDeferredDiscoveryRESTMapper(discocache.NewMemCacheClient(kubeClient.Discovery()))
```
sttts, line 50: nit: why a new one for each subtest?

test/extended/apiserver/apply.go
```diff
@@ -0,0 +1,168 @@
...
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace && len(testNamespace) == 0 {
+		// Create the namespace on demand
+		namespace = oc.SetupProject()
+		testNamespace = namespace
+	}
+
+	// Ensure that any stub embedding the etcd test namespace
+	// is updated to use local test namespace instead.
+	if len(testNamespace) > 0 {
+		stub = strings.Replace(stub, exetcd.TestNamespace, testNamespace, -1)
```
sttts, line 148: is there a need to use the etcd storage test namespace? Would rather use a unique namespace here to avoid conflicts, and to reduce shared logic.

sttts:
Overall it looks good!

test/extended/apiserver/apply.go
```diff
@@ -0,0 +1,168 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By(fmt.Sprintf("updating the %s via apply", unstructuredObj.GetKind()))
+			obj, err := resourceClient.Patch(
+				context.Background(),
+				unstructuredObj.GetName(),
+				types.ApplyPatchType,
+				serializedObj,
+				metav1.PatchOptions{
+					FieldManager: "apply_test",
```
p0lyn0mial, line 88: is this what it takes to do server-side apply?

sttts: the patch type especially.

p0lyn0mial: oh, okay, it points to the client that made the update.

test/extended/apiserver/apply.go
```diff
@@ -0,0 +1,168 @@
...
+
+			resourceClient, unstructuredObj, namespace := createResource(oc, mapper, gvr, data.Stub, testNamespace)
+			testNamespace = namespace
+			defer deleteResource(resourceClient, unstructuredObj.GetName())
+
+			serializedObj, err := json.Marshal(unstructuredObj.Object)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By(fmt.Sprintf("updating the %s via apply", unstructuredObj.GetKind()))
+			obj, err := resourceClient.Patch(
```
p0lyn0mial, line 82: we don't have to actually change the object to test the server-side apply logic?
is setting a `FieldManager` a good indicator?

sttts: A good first step. It won't tell us that our schema is correct.

test/extended/apiserver/apply.go
marun: I am using a unique namespace - that's `testNamespace`. It only needs to be used when the resource itself is namespaced since many of the types are cluster-scoped. This minimizes the need to provision new namespaces since that is not cheap.

test/extended/apiserver/apply.go
marun: Note the conditional check above. It's initialized once per worker, not once per subtest. My ginkgo-foo is weak, if there's a better way to accomplish this I'm all ears!

test/extended/apiserver/apply.go
marun: The server-side logic has reasonable integration tests in upstream. The goal of this test is validating that openshift types are compatible with server-side apply vs validating server-side apply itself.

marun:
Updated to fix the final test failure: https://github.com/openshift/origin/pull/25652/files#diff-5622a956d54343610dc9861478380f2c89a9050297ea9bc48726e6356c6a9dedR67

test/extended/apiserver/apply.go
marun: This just points to who made an update, and it can be supplied with CREATE, UPDATE or PATCH. The key to performing server-side apply is `types.ApplyPatchType`.

PR #25653, opened 2020-11-02
Remove CSI driver installation from openshift-tests

jsafrane:
AWS EBS CSI driver is now installed by default in an OCP cluster on AWS, therefore we don't need to install it via openshift-install. Remove the installation + update docs of `openshift/csi` test suite.

CC @openshift/storage


smarterclayton:
/lgtm
/approve

Thanks

PR #25655, opened 2020-11-02
UPSTREAM: <carry>: Bug 1893156: include / prefix in instance ID output

Fedosin:
When we want to read an instance ID from the metadata service, cloud provider doesn't include "/" prefix, which is required for successful parsing of provider the ID later.
This commit adds the missing "/" prefix to the output.

Cherry-picked from https://github.com/openshift/kubernetes/pull/343


mandre (commented):
/lgtm

PR #25673, opened 2020-11-10
Bug 1896188: bump wait timeout for build completion in oc debug test

soltysh:
/assign @sallyom 
since you looked at the PR which was changing this
/assign @wking 
since you complained in a good way :wink:

### DIFF
--- a/test/extended/cli/debug.go
+++ b/test/extended/cli/debug.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"time"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/extended/cli/debug.go
```diff
@@ -22,7 +27,7 @@ var _ = g.Describe("[sig-cli] oc debug", func() {
 	g.It("deployment configs from a build", func() {
 		err := oc.Run("create").Args("-f", testCLIDebug).Execute()
 		o.Expect(err).NotTo(o.HaveOccurred())
-		err = wait.Poll(cliInterval, cliTimeout, func() (bool, error) {
+		err = wait.Poll(cliInterval, buildTimeout, func() (bool, error) {
```
wking, line 30: This is a 5x bump from [the old 1m][1], which seems plenty generous.

[1]: https://github.com/openshift/origin/blob/2c51df8035383740abd5784e8c71244094797130/test/extended/cli/admin.go#L23

wking (approved):
/lgtm

PR #25674, opened 2020-11-10
adds e2e for testing admission webhook for OpenShift APIs

p0lyn0mial:
this PR adds an e2e test for checking if calling an admission webhook from `OpenShift` API server works.

it deploys the webhook server from `imageutils.GetE2EImage(imageutils.Agnhost)` image. 
it configures all API servers to call an always-deny validating admission on all resources matching `webhook-marker` label and namespaces matching `webhook-marker` selector.
it then checks if creating an image stream with the label is forbidden by the webhook.

it also provides a function that could be reused to check `oauth-apiserver`

### DIFF
--- a/test/extended/apiserver/webhook.go
+++ b/test/extended/apiserver/webhook.go
@@ -1,0 +1,292 @@
+package apiserver
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	g "github.com/onsi/ginkgo"
+
+	imagev1 "github.com/openshift/api/image/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2edeployment "k8s.io/kubernetes/test/e2e/framework/deployment"
+	imageutils "k8s.io/kubernetes/test/utils/image"
+	"k8s.io/utils/pointer"
+)
+
+var _ = g.Describe("[sig-api-machinery][Feature:AdmissionWebhook]", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLI("apiserver")
+
+	g.It("should call a validating admission webhook when accessing OpenShift API Server", func() {
+		targetNamespace := oc.Namespace()
+		kubeClient := oc.AdminKubeClient()
+		servicePort := int32(8443)
+		containerPort := int32(8444)
+
+		annotateNamespace(kubeClient, targetNamespace, "webhook-marker")
+		createService(kubeClient, targetNamespace, "openshift-test-webhook", "openshift-test-webhook", servicePort, containerPort)
+		deployWebhookServer(kubeClient, imageutils.GetE2EImage(imageutils.Agnhost), targetNamespace, "openshift-test-webhook", "openshift-test-webhook", "openshift-test-webhook", containerPort)
+		removeWebhookConfiguration := createValidatingWebhook(kubeClient, targetNamespace, "webhook-marker", "webhook-marker", "openshift-test-webhook", "openshift-test-webhook", servicePort)
+		defer removeWebhookConfiguration()
+
+		g.By(fmt.Sprintf("attempting to create an image stream in %s namespace (should fail)", targetNamespace))
+		err := hitValidatingWebhook(func() (func(), error) {
+			is := &imagev1.ImageStream{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "openshift-test-webhook",
+					Labels: map[string]string{"webhook-marker": "true"},
+				},
+			}
+			_, err := oc.AdminImageClient().ImageV1().ImageStreams(targetNamespace).Create(context.TODO(), is, metav1.CreateOptions{})
+			removeImageStream := func() {
+				err = oc.AdminImageClient().ImageV1().ImageStreams(targetNamespace).Delete(context.TODO(), is.GetName(), metav1.DeleteOptions{})
+				e2e.ExpectNoError(err, "deleting an image stream %s in namespace %s", is.Name, targetNamespace)
+			}
+			return removeImageStream, err
+
+		})
+		e2e.ExpectNoError(err, "failed to create an image stream, expected to get admission deny error")
+	})
+})
+
+// annotateNamespace adds the label selector so that the webhook is scoped only to this namespace.
+func annotateNamespace(client kubernetes.Interface, namespace string, labelSelectorKey string) {
+	g.By("annotating the namespace so that the webhook is scoped to this namespace")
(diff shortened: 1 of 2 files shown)


test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,292 @@
...
+
+		annotateNamespace(kubeClient, targetNamespace, "webhook-marker")
+		createService(kubeClient, targetNamespace, "openshift-test-webhook", "openshift-test-webhook", servicePort, containerPort)
+		deployWebhookServer(kubeClient, imageutils.GetE2EImage(imageutils.Agnhost), targetNamespace, "openshift-test-webhook", "openshift-test-webhook", "openshift-test-webhook", containerPort)
+		removeWebhookConfiguration := createValidatingWebhook(kubeClient, targetNamespace, "webhook-marker", "webhook-marker", "openshift-test-webhook", "openshift-test-webhook", servicePort)
+		defer removeWebhookConfiguration()
+
+		g.By(fmt.Sprintf("attempting to create an image stream in %s namespace (should fail)", targetNamespace))
+		err := hitValidatingWebhook(func() (func(), error) {
+			is := &imagev1.ImageStream{
```
p0lyn0mial, line 47: I think we would like to check `all` resources managed by this server.

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,292 @@
...
+				Secret: &v1.SecretVolumeSource{SecretName: secretName},
+			},
+		},
+	}
+	containers := []v1.Container{
+		{
+			Name:         "sample-webhook",
+			VolumeMounts: mounts,
+			Args: []string{
+				"webhook",
```
sttts, line 133: what exactly is this doing?

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// hitValidatingWebhook tries to create a resource and expect it to fail (deny)
+// it tries to send a few request before giving up because registering a webhook is not instant and might take up to a few seconds
+func hitValidatingWebhook(callback func() (func(), error)) error {
+	return wait.PollImmediate(100*time.Millisecond, 30*time.Second, func() (bool, error) {
+		cleanup, err := callback()
+		if err != nil {
+			// the always-deny webhook does not provide a reason, so check for the error string we expect
+			if strings.Contains(err.Error(), "denied") {
```
sttts, line 281: what is the full error string?

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,292 @@
+package apiserver
```
sttts, line 1: could we move this to openshift-apiserver repo?

test/extended/apiserver/webhook.go
p0lyn0mial: it runs https://github.com/kubernetes/kubernetes/blob/master/test/images/agnhost/webhook/main.go#L211

the app exposes an API, we use `http.HandleFunc("/always-deny", serveAlwaysDeny)`

test/extended/apiserver/webhook.go
p0lyn0mial: the full error is `admission webhook "deny-all-webhook.openshift.io" denied the request: this webhook denies all requests`

test/extended/apiserver/webhook.go
p0lyn0mial: yes, we could. We would have to move the common code to `library-go` repo.

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,333 @@
...
+						e2e.ExpectNoError(err, "deleting an image stream %s in namespace %s", is.Name, targetNamespace)
+					}
+					return removeImageStream, err
+				},
+			},
+
+			{
+				name: "build config",
+				createFn: func() (func(), error) {
+					bc := &buildv1.BuildConfig{
```
p0lyn0mial, line 71: I need to check if there is a better way of finding and creating resources. I think I could try to reuse data provided by `OpenshiftEtcdStorageData` test.

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,386 @@
...
+		defer removeWebhookConfiguration()
+
+		storageData := exetcd.OpenshiftEtcdStorageData
+		for key := range storageData {
+			gvr := key
+			data := storageData[gvr]
+
+			// apply for core types is already well-tested, so skip
+			// openshift types that are just aliases.
+			aliasToCoreType := data.ExpectedGVK != nil
```
p0lyn0mial, line 61: we could reuse the code with the test that @marun prepared but it seems to be more readable as it is now

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,386 @@
...
+		},
+		Webhooks: []admissionregistrationv1.ValidatingWebhook{
+			{
+				Name: "deny-all-webhook.openshift.io",
+				Rules: []admissionregistrationv1.RuleWithOperations{{
+					Operations: []admissionregistrationv1.OperationType{admissionregistrationv1.Create},
+					Rule: admissionregistrationv1.Rule{
+						APIGroups:   []string{"*"},
+						APIVersions: []string{"*"},
+						Resources:   []string{"*"},
```
tkashem, line 256: okay, this is the same rule they are using (the customer case I am looking into). can we test this scenario?
- create a deploymentconfig
- and then `oc rollout latest dc/{name}`

or any equivalent test path that covers the ^ scenario.

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,386 @@
...
+
+	// supply a value for namespace if the scope requires
+	mapping, err := mapper.RESTMapping(gvk.GroupKind())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// ensure that any stub embedding the etcd test namespace
+	// is updated to use local test namespace instead.
+	stub = strings.Replace(stub, exetcd.TestNamespace, namespace, -1)
```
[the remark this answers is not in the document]
p0lyn0mial: some stubs embed `etcdstoragepathtestnamespace` namespace, for example 
```
Stub:             `{"clientName": "system:serviceaccount:etcdstoragepathtestnamespace:clientg", "metadata": {"name": "user:system:serviceaccount:etcdstoragepathtestnamespace:clientg"}, "scopes": ["user:info"], "userName": "user", "userUID": "cannot be empty"}`
```

this code makes sure that it will be replaced by the target namespace so that it can be found by the test.

test/extended/apiserver/webhook.go
```diff
@@ -0,0 +1,386 @@
...
+
+	_, err = resourceClient.Create(context.Background(), &unstructuredObj, metav1.CreateOptions{})
+	return resourceClient, &unstructuredObj, err
+}
+
+func deleteRes(resourceClient dynamic.ResourceInterface, name string) {
+	err := resourceClient.Delete(context.Background(), name, metav1.DeleteOptions{})
+	if err != nil {
+		e2e.ExpectNoError(err, "Unexpected error deleting resource: %v")
+	}
```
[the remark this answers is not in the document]
p0lyn0mial: it is, thx.

test/extended/apiserver/webhook.go
p0lyn0mial: `deploymentconfig` is already covered by this test - https://github.com/openshift/origin/blob/master/test/extended/etcd/etcd_storage_path.go#L75

we do test all `*.openshift.io` resources

p0lyn0mial:
it would be nice if we could extend the test to check some (all?) subresources as well so that we can find bugs like https://github.com/openshift/openshift-apiserver/pull/165 in the future.

p0lyn0mial:
/hold

the new test fails on `ci/prow/e2e-metal-ipi-ovn-ipv6` (only) with `x509: certificate signed by unknown authority`

PR #24887, opened 2020-04-17
Bug 1816812: Enable mirroring test images offline

smarterclayton:
Some users wish to run the OpenShift conformance suite disconnected.  As a first step, we need to enable a pattern to sync images required by tests from online to offline.  To do that, we maintain a public mirror at `quay.io/openshift/community-e2e-images` which contains unsupported but valid images for each version of Kube we release.  The `openshift-tests` command supports a helper to generate a mirror file to take that repo offline for a given release, which integrates naturally with `oc image mirror`.  After this merges there will be a set of other changes to ensure our tests are using upstream images where possible.

A test user would mirror the images to a local registry:

    openshift-tests images --to-repository myregistry.com/test/location | oc image mirror -f - --filter-by-os=.*
    openshift-tests run openshift/conformance --from-repository myregistry.com/test/location

By default, all tests will run from the default mirror (not from the original location) at `quay.io/openshift/community-e2e-images`

TODO:
* [ ] mark upstream tests that require public access as skipped / or have them skip
* [x] identify openshift images that can be mirrored
* [x] potentially limit the suite if we need more images, or investigate switching to on cluster samples
* [x] fail if an image is used not from the mirror so we catch regressions

---

The `openshift-tests images` command generates a mirror file for use with `oc image mirror` to sync the images used by tests to a private repository (or to disk). The user would then pass `--from-repository` to `openshift-tests run*`.

The `--upstream` flag allows the tool to mirror the upstream images to their canonical location on quay.io for general project use. This is only for admin use within the project and results in the images making it to `quay.io/openshift/community-e2e-images` - full command is:

    openshift-tests images --upstream --to-repository quay.io/openshift/community-e2e-images | oc image mirror -f - --filter-by-os=.*

The change is based around this carry (which I will follow up upstream on). Kube has less practical need for this solution because there is no infrastructure to mirror images in general and this is more specialized to our needs with our offline story (kube has no offline story). Even if this stays a carry, it is fairly compact.

UPSTREAM: <carry>: Allow test images to be in a single mirror

In order to enable offline test execution we need to be able to collect all images used by Kube e2e into a single repo and then mirror them. This structure has been discussed but not fully thought through upstream, but we need to enable customers now.

The presence of the env var switches all pullable images to use a single repository (which has been prepared by openshift-tests) except for those images that are explicitly not intended to be pulled (those tests can't run offline anyway). The original map is returned so the mirror has commands to run.

PR #25675, opened 2020-11-10
skip bld tbr pullscret e2e until https://github.com/openshift/origin/pull/24887 merges and we have a new path forward

gabemontero:
another build e2e change motivated by the diagnosis of https://github.com/openshift/origin/pull/24887

per conversation with @smarterclayton at  https://coreos.slack.com/archives/C014MHHKUSF/p1604952273437500?thread_ts=1604938369.426700&cid=C014MHHKUSF

we'll skip this test for now ... also in that slack thread, @dmage is working on a pattern where tests can spin up additional registries on the test cluster to use in  "validate proper authorization with external registries" tests

the current thought is we can borrow from his pattern

also note, this test was also rendered obsolete when 4.5 came out ... as the TBR secret could be fetched on the node during the image import process, the notion of copying the install pull secret to the local namespace became irrelevant 

/assign @bparees 
/assign @adambkaplan

### DIFF
--- a/test/extended/builds/pullsecret.go
+++ b/test/extended/builds/pullsecret.go
@@ -72,6 +72,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds][Slow] using pull secrets in a b
 				})
 
 				g.It("should be able to use a pull secret in a build", func() {
+					g.Skip("until https://github.com/openshift/origin/pull/24887 merges and we have a pattern from Oleg for spinning up a registry on the test cluster")
 					g.By("creating build config")
 					err := oc.Run("create").Args("-f", pullSecretBuild).Execute()
 					o.Expect(err).NotTo(o.HaveOccurred())


gabemontero:
/hold

late breaking news ... @smarterclayton is going to update 24887 ... see  https://coreos.slack.com/archives/C014MHHKUSF/p1605016753449800?thread_ts=1604938369.426700&cid=C014MHHKUSF

to allow pulling from the nodejs image on registry.redhat.io

smarterclayton:
I have worked around this for now with putting the image on the allow list so you can access it (with ac omment about what it is used for).  The test will still fail in disconnected but we can work on that later.

gabemontero:
this TC now worked around in https://github.com/openshift/origin/pull/24887

eventually when truly disconnected we'll need to refactor this testcase to launch a registry requiring creds, which will host the image we access with the pull secret

PR #25676, opened 2020-11-10
Bug 1896474: Add path based route tests

tnozicka:
In the past there have been several combinations of path based routing broken and when I went to investigate why, I've really felt the need for coverage there. I don't have the haproxy expertise to actually fix those cases, but I do hope that having test coverage will help the team and prevent future regressions.

One instance where path based routing is especially important is exposing token for http-01 ACME validation but our users has likely suffered too at some point. We have done various hacks in github.com/tnozicka/openshift-acme to make it work but it turned out the bug just moved to a different combination of TLS setting.

Currently failing combinations that need fixing:
```
[sig-network][Conformance][Feature:Route] Path based routing should work for "edge+Allow" route and "edge+Redirect" route [Suite:openshift/conformance/parallel/minimal]
[sig-network][Conformance][Feature:Route] Path based routing should work for "edge+Allow" route and "passthrough" route [Suite:openshift/conformance/parallel/minimal]
[sig-network][Conformance][Feature:Route] Path based routing should work for "http-only" route and "edge+Redirect" route [Suite:openshift/conformance/parallel/minimal]
```

/kind bug
/priority important-soon

/assign @openshift/openshift-team-network-edge

### DIFF
--- a/test/extended/include.go
+++ b/test/extended/include.go
@@ -37,6 +37,7 @@ import (
 	_ "github.com/openshift/origin/test/extended/prometheus"
 	_ "github.com/openshift/origin/test/extended/quota"
 	_ "github.com/openshift/origin/test/extended/router"
+	_ "github.com/openshift/origin/test/extended/routes"
 	_ "github.com/openshift/origin/test/extended/security"
 	_ "github.com/openshift/origin/test/extended/tbr_health"
 	_ "github.com/openshift/origin/test/extended/templates"
--- a/test/extended/routes/paths.go
+++ b/test/extended/routes/paths.go
@@ -1,0 +1,644 @@
+package routes
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"math/big"
+	"net/http"
+	"net/url"
+	"time"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	routev1 "github.com/openshift/api/route/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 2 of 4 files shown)


tnozicka:
test-cmd flake
/retest

miheer:
1. 
The insecure policy to allow requests sent on an insecure scheme HTTP.

The insecure policy to redirect requests sent on an insecure scheme HTTP to a secure scheme HTTPS.


So, [sig-network][Conformance][Feature:Route] Path based routing should work for "edge+Allow" route and "edge+Redirect" route [Suite:openshift/conformance/parallel/minimal]  is conflicting and an invalid combination

2.

[sig-network][Conformance][Feature:Route] Path based routing should work for "http-only" route and "edge+Redirect" route [Suite:openshift/conformance/parallel/minimal]

This is also invalid as http-only means terminationPolicy is disabled so http traffic is not allowed

 "edge+Redirect" implies that the edge route sets up a redirect for HTTP requests, but thatconflicts with having an http-only route handle HTTP requests.

The insecure policy redirect requests sent on an insecure scheme HTTP to a secure scheme HTTPS.

So it is conflicting and an invalid combination.



3.[sig-network][Conformance][Feature:Route] Path based routing should work for "edge+Allow" route and "passthrough" route [Suite:openshift/conformance/parallel/minimal]

Path-based routing is not available when using passthrough TLS, as the router does not terminate TLS in that case and cannot read the contents of the request.

So this combination won't work.

jeltz:
@miheer As far as I understand this PR was created to support HTTP01 ACME challenges for passthrough routes (I base this on that @tnozicka is the main developer of https://github.com/tnozicka/openshift-acme). So the problem I have (and I believe this PR tired to demonstrate) is that there is no way to create a route which is only on port 80 and therefore does not conflict with a passthrough route on port 443. Please correct me if I am mistaken about the problem here or if there actually is a solution for this in OpenShift.

jeltz:
Never mind, it seems to work just fine.

PR #25678, opened 2020-11-10
Bug 1895537: Image extract tests should check results of ImageStreamImport

dmage:
Successful creation of ImageStreamImport doesn't mean that images are imported successfully. Status of the returned object should be checked as it may contain errors from the registry.

### DIFF
--- a/test/extended/images/extract.go
+++ b/test/extended/images/extract.go
@@ -2,6 +2,7 @@ package images
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/MakeNowJust/heredoc"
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
Although that doesn't solve the problem from the BZ, it only exposes information why it failed.

dmage:
@soltysh correct, the problem most likely is the rate limit on Docker Hub.

soltysh:
> @soltysh correct, the problem most likely is the rate limit on Docker Hub.

any plans to switch to quay or divide those? I guess you'll know that after this merges.

PR #25679, opened 2020-11-11
IR-152: Monitor availability of image registry during upgrades

dmage:
We expect the registry to be always available during upgrades. This test should help us to ensure this.


dmage:
/assign @soltysh
Maciej, can you take a look while Tomáš is out?

soltysh (approved):
/lgtm
/approve

PR #25683, opened 2020-11-12
Bug 1870050: Mount ImageFS in node container

joelsmith:
Because the atomic-openshift-node container running on RHEL Atomic Host bind mounts `/var/lib/docker/containers` instead of `/var/lib/docker`, the Kubelet gets the wrong disk usage stats when it looks at how much of the `/var/lib/docker` is in use. This results in dangling and unused images never being cleaned up when the ImageFS (i.e. the host's `/var/lib/docker`) fills up, since the Kubelet never triggers its garbage collection.

### DIFF
--- a/images/node/system-container/config.json.template
+++ b/images/node/system-container/config.json.template
@@ -415,8 +415,8 @@
         },
         {
 	    "type": "bind",
-	    "source": "/var/lib/docker/containers",
-	    "destination": "/var/lib/docker/containers",
+	    "source": "/var/lib/docker",
+	    "destination": "/var/lib/docker",
 	    "options": [
 		"bind",
                 "slave",


joelsmith:
I ran a test using the image built by the ci/prow/images job on this PR using a RHEL Atomic Host node in my own test cluster:

```
atomic containers update --rebase registry.build01.ci.openshift.org/ci-op-qs7429lk/stable:node atomic-openshift-node
```

I verified that the `config.json` for the atomic-openshift-node runc container had the updated mount from this PR.  I verified that `/var/lib/docker` is being mounted:

```
# runc exec atomic-openshift-node df -h / /var/lib/docker
Filesystem                   Size  Used Avail Use% Mounted on
/dev/mapper/atomicos-var      10G  4.2G  5.8G  42% /
/dev/mapper/atomicos-docker   20G   16G  4.6G  78% /var/lib/docker
```

(when the same test was run before, it showed the same filesystem for both directories).

Finally, I ran the image pull test described in the bug and verified that the Kubelet notices when `/var/lib/docker` goes over 85% and deletes dangling images.

giuseppe (approved):
LGTM

sdodson:
/hold
@joelsmith Can you please take a look at the failing tests and clear this when you feel that they'll pass.

joelsmith:
This PR https://github.com/openshift/openshift-ansible/pull/12270 should fix e2e-gcp, but I don't know if we need to wait at all for any kind of image build before we run the test. I still haven't figured out what is failing on the other test. It has failed for every run on every PR since 11/11.

/retest

joelsmith:
The 3.11 test extended_conformance_install hasn't passed for any PR for nearly 3 weeks. (The last pass was 2020-11-11 15:01:32 +0000 UTC). https://bugzilla.redhat.com/show_bug.cgi?id=1902067 is tracking the issue of the failing test. Should we ask for a test override exception for this PR since the failure is unrelated to this PR or wait for the failing test to be fixed?

sdodson:
Based on @mrunalp approval above

PR #25688, opened 2020-11-16
Bug 1887488: e2e: node: fix ping tester pod

fromanirh:
Pin the busybox image setting a known-good tag.
It was not done previously because of an oversight, no other reason.

Generalize the code which creates the testing pod to make the
code more flexible and a bit easier to read.

Last, use privileged pods for ping tests. This is needed because of the busybox images used
by the pinger pods.
Another approach could be to change base image (e.g. using centos images) but considering
the nature of the test and the fact that the busybox image is tiny, and much smaller than the closest
good alternative (centos), this approach seems good enough.

Bug-Url: https://bugzilla.redhat.com/1887488

### DIFF
--- a/test/extended/topology_manager/resourcealign.go
+++ b/test/extended/topology_manager/resourcealign.go
@@ -24,6 +24,8 @@ import (
 const (
 	networkAttachmentAnnotation string = "k8s.v1.cni.cncf.io/networks"
 	sriovInterfaceName          string = "sriov1"
+	// same image we use in test/extended/dns/dns.go - anything with busybox is actually enough
+	testerImage string = "gcr.io/google_containers/dnsutils:e2e"
 )
 
 var _ = g.Describe("[Serial][sig-node][Feature:TopologyManager] Configured cluster", func() {
@@ -63,7 +65,7 @@ var _ = g.Describe("[Serial][sig-node][Feature:TopologyManager] Configured clust
 				}
 
 				ns := oc.KubeFramework().Namespace.Name
-				testingPods := pps.MakeBusyboxPods(ns, deviceResourceName)
+				testingPods := pps.MakePods(testerImage, ns, deviceResourceName)
 				// we just want to run pods and check they actually go running
 				updatedPods := createPods(client, ns, testingPods...)
 				defer deletePods(oc, updatedPods)
@@ -128,6 +130,11 @@ var _ = g.Describe("[Serial][sig-node][Feature:TopologyManager] Configured clust
 					g.Skip("no device resource configured")
 				}
 
+				networkCheckImage, ok := os.LookupEnv(networkCheckImageEnvVar)
+				if !ok {
+					g.Skip(fmt.Sprintf("no network check image provided (use %s)", networkCheckImageEnvVar))
+				}
+
 				// any random amount of cores > 2 (to be HT-neutral)
 				pps := PodParamsList{
 					{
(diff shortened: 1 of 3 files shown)


fromanirh:
this
```
fail [k8s.io/kubernetes@v1.19.2/test/e2e/apps/daemon_set.go:102]: Unexpected error:
    <*errors.errorString | 0xc0027641b0>: {
        s: "error while waiting for pods to become inactive daemon-set: there are 3 active pods. E.g. \"daemon-set-qkj95\" on node \"ip-10-0-130-238.us-west-2.compute.internal\"",
    }
    error while waiting for pods to become inactive daemon-set: there are 3 active pods. E.g. "daemon-set-qkj95" on node "ip-10-0-130-238.us-west-2.compute.internal"
occurred
```
can't be caused by this PR changing how we use an image on an unrelated test. Will report as flake.

test/extended/topology_manager/resourcealign.go
```diff
@@ -24,6 +24,7 @@ import (
 const (
 	networkAttachmentAnnotation string = "k8s.v1.cni.cncf.io/networks"
 	sriovInterfaceName          string = "sriov1"
+	busyboxImage                string = "busybox:1.31.0"
```
adambkaplan, line 27: Note that this is pulling from DockerHub without a pull secret, and thus can fail to deploy due to a rate limit.

Can `registry.redhat.io/ubi8/ubi-minimal` be used instead?

fromanirh: Anything which contains ping and it is maintained is fine for this task. I'll research for alternatives.

test/extended/topology_manager/resourcealign.go
```diff
@@ -552,6 +557,13 @@ func (pp PodParams) MakeBusyboxPod(namespace, deviceName string) *corev1.Pod {
 			},
 		}
 
+		if cp.Privileged {
+			isTrue := true
+			cnt.SecurityContext = &corev1.SecurityContext{
+				Privileged: &isTrue,
+			}
+		}
```
adambkaplan, line 565: Are you sure you need to run a pod that's fully privileged? This bypasses every baked in security check.

If the issue is that on busybox you now need root, you can use `runAsUser: 0` in the security context. Not awesome, but better than running a privileged container.

fromanirh: good point. Let me narrow down the privileges.

fromanirh: So `RunAsUser: 0` is not enough, it seems I need at least `CAP_NET_ADMIN` (which is still too much?)
if I removed 'Privileged: true` and I replace with `RunAsUser: 0` I get:
```
Nov 24 17:52:36.585: INFO: Error running /usr/local/bin/oc --namespace=e2e-test-topology-manager-xhjcv --kubeconfig=/home/fromani/clusters/cnflab/kubeconfig rsh -n default -c test-0 test-29pj7 ping -c 3 10.56.217.171:
StdOut>
bind: Permission denied
PING 10.56.217.171 (10.56.217.171): 56 data bytes
command terminated with exit code 2
StdErr>
bind: Permission denied
PING 10.56.217.171 (10.56.217.171): 56 data bytes
command terminated with exit code 2
```

adambkaplan (commented):
Please squash commits, otherwise looks good.

adambkaplan (approved):
/approve

PR #25689, opened 2020-11-16
Bug 1887509: e2e: node: filter out worker nodes without MCD

fromanirh:
The topology manager e2e tests expect MCD running
in order to be able to extract the node Kubelet and
check the test precoditions.

So we add another pre-filter to rule out worker nodes
without MCD running. If no suitable worker nodes are
found, the tests must just skip.

Bug-Url: https://bugzilla.redhat.com/1887509


bparees:
/approve

please get someone from the node team to do the lgtm

PR #25690, opened 2020-11-16
Bug 1898118: Logging more during ephemeral registry deployment

ricardomaraschini:
This patch increases the number of logged information during the
following tests:

ImageStream API TestImportRepositoryFromBlockedRegistry
ImageStream API TestImportImageFromBlockedRegistry
ImageStream API TestImportImageFromInsecureRegistry
ImageStream API TestImportRepositoryFromInsecureRegistry

Tests now log the deployment status. Time between polls was also
increased, no need to rush when checking for the deployment status.

Timeout for endpoints to become available were also increased, from one
to five minutes, logging was also increased during endpoints check.

### DIFF
--- a/test/extended/images/imagestreamimport.go
+++ b/test/extended/images/imagestreamimport.go
@@ -23,10 +23,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
 
 	configv1 "github.com/openshift/api/config/v1"
 	imagev1 "github.com/openshift/api/image/v1"
-	projectv1 "github.com/openshift/api/project/v1"
 
 	exutil "github.com/openshift/origin/test/extended/util"
 )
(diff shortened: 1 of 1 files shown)


test/extended/images/imagestreamimport.go
```diff
@@ -280,49 +318,51 @@ func deployEphemeralImageRegistry(oc *exutil.CLI, proj *projectv1.Project) error
...
 		}
-		return deploy.Status.AvailableReplicas == replicas, nil
+		succeed := deploy.Status.AvailableReplicas == replicas
+		if !succeed {
+			t.Logf("registry deployment not ready yet, status: %+v", deploy.Status)
+		}
+		return succeed, nil
 	}); err != nil {
-		return err
+		e2e.DumpAllNamespaceInfo(oc.AdminKubeClient(), oc.Namespace())
```
dmage, line 338: You don't need to call this function for `oc.Namespace()`. This repository uses [Ginkgo](https://onsi.github.io/ginkgo/#structuring-your-specs) that makes tests more declarative. The namespace is created in `BeforeEach` and is deleted in `AfterEach`, all necessary debug information for the namespace is generated by `AfterEach`. I guess these tests also should use `BeforeEach` and `AfterEach` and shouldn't use `g.GinkgoTInterface` and the latter is a helper for legacy code.

ricardomaraschini: PTAL

PR #25691, opened 2020-11-16
Bug 1886160: Increase resiliency of quorum restore test to potentially recoverable errors

vareti:
After the restore operation in disruptive test, kube-apiserver restarts due to new revision roll outs. As this is expected to happen when new machines are created, it is acceptable to ignore the error response when api requests are made.

### DIFF
--- a/test/extended/dr/quorum_restore.go
+++ b/test/extended/dr/quorum_restore.go
@@ -200,7 +200,7 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 					}
 
 					framework.Logf("Waiting for machines to be created")
-					err = wait.Poll(30*time.Second, 10*time.Minute, func() (done bool, err error) {
+					err = wait.Poll(30*time.Second, 20*time.Minute, func() (done bool, err error) {
 						mastersList, err := ms.List(context.Background(), metav1.ListOptions{
 							LabelSelector: "machine.openshift.io/cluster-api-machine-role=master",
 						})
--- a/test/extended/dr/quorum_restore.go
+++ b/test/extended/dr/quorum_restore.go
@@ -148,9 +148,18 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 				}
 
 				framework.Logf("Perform etcd backup on remaining machine %s (machine %s)", survivingNodeName, survivingMachineName)
-				expectSSH("sudo -i /bin/bash -cx 'rm -rf /home/core/backup; /usr/local/bin/cluster-backup.sh ~core/backup'", survivingNode)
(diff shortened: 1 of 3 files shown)


test/extended/dr/quorum_restore.go
```diff
@@ -188,7 +188,8 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 								return false, nil
 							}
 							if err != nil {
-								return false, err
+								framework.Logf("Error seen when re-creating machines: %v", err)
```
marun, line 191: This is a good start. Would it also make sense to apply this strategy of performing API operations in a poll loop and logging potentially recoverable errors rather than failing? i.e.

- ensure all calls to `scaleEtcdQuorum` occur in a poll loop
- ensure errors returned by `scaleEtcdQuorum` are logged rather than ignored
- ensure failing to list machines (L201) logs rather than fails on error
- ???

Ideally, all our e2e tests would be making the same assumptions as the controllers under test and be resilient to recoverable failure.

marun:
/retitle Bug 1886160: Increase resiliency of quorum restore test to potentially recoverable errors

test/extended/dr/quorum_restore.go
vareti: It actually makes sense given the errors are recoverable and would reduce the number of false negatives. I will make the changes accordingly.

test/extended/dr/quorum_restore.go
```diff
@@ -134,16 +147,14 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 							failures = 0
 						}
 
-						// there is a small chance the cluster restores the default replica size during
```
marun, line 137: I think it's desirable to keep this check here rather than externalizing in a goroutine. It's already in a polling loop, I was just asking to have any error it returned logged for traceability.

marun (commented):
I apologize if my previous comment wasn't clear - I was suggesting that the call to `ensureEtcdQuorum` on L99 be put in a poll loop not creating a new poll loop for the call appearing on L140. Maybe that's superfluous, though, since it's unlikely to be made during disruption. So maybe just ensure that any error from the call on L140 is logged?

Ideally all e2e tests would poll for every single API call since otherwise there is always the possibility of transient failure causing a test failure. Our controllers use reconciling loops to achieve a similar degree of resiliency, and it's absurd that we've written our e2e tests without that same degree of rigor. But for now, ensuring resiliency during a period of disruption should be sufficient.

vareti:
@marun As I understand, test must resilient to transient error only during disruption. Because there is only one instance of kube-apiserver is available after the restore script is run. At L99, cluster is supposed to be healthy and API calls should not be disrupted. So, we cannot add a poll loop here and the call must succeed.

I agree with the other point of just logging the `scaleEtcdQuorum` error rather than moving it to separate loop. Sounds simple.

vareti:
Not a consistent failure. Some of the tests failed even before they started as cluster was not available. Trying a retest.

/test e2e-aws-disruptive

PR #25697, opened 2020-11-18
e2e: Add useroauthaccesstoken tests

stlaz:
/hold
has fake bump from https://github.com/openshift/client-go/pull/167


stlaz:
/close
added these in the component's repo instead

PR #25702, opened 2020-11-20
Bug 1900030: Fix monitoring availability of image registry

dmage:
When the tests are run outside of the cluster (usually they are), the tests cannot access the registry service. They also cannot rely on the registry route, because the registry route is disabled by default.

To check availability of the registry, tests can create additional route for the registry service and monitor it during upgrades.

### DIFF
--- a/test/extended/util/disruption/imageregistry/imageregistry.go
+++ b/test/extended/util/disruption/imageregistry/imageregistry.go
@@ -9,14 +9,19 @@ import (
 
 	"github.com/onsi/ginkgo"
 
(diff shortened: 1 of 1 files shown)


dmage:
/assign @soltysh

Apparently the test currently checks the registry of the CI cluster.

wking (approved):

soltysh (approved):
/lgtm
/approve

wking:
Seems like the two perma-failing jobs need more than retests.

/hold

to save money until someone has time to investigate.

PR #25706, opened 2020-11-23
[WIP] Bug 1882486: Proxy CI: Use router internal service for ingress tests

sgreene570:
This PR is intended to fix the following tests for Proxy CI jobs by forcing them to use the router internal service IP rather than the external load balancer for the default ingress controller:

```
        The HAProxy router should respond with 503 to unrecognized hosts\|
        The HAProxy router should serve routes that were created from an ingress\|
        The HAProxy router should set Forwarded headers appropriately\|
        The HAProxy router should support reencrypt to services backed by a serving certificate automatically\|
```

Tagging WIP since CI will verify if these changes work as intended.

### DIFF
--- a/test/extended/util/router.go
+++ b/test/extended/util/router.go
@@ -21,8 +21,15 @@ func waitForRouterExternalIP(oc *CLI) (string, error) {
 }
 
 func routerShouldHaveExternalService(oc *CLI) (bool, error) {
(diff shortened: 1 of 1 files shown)


sgreene570:
This PR simply won't work on aws.
/close

PR #25707, opened 2020-11-24
Bug 1886160: Improve reliability of e2-aws-disruptive

marun:


marun:
I removed the machine recover shutdown changes and skipped the test pending the outcome of #25750.

marun:
Marking wip - broken pending https://github.com/openshift/cluster-etcd-operator/pull/509

marun:
Updated quorum restore test to force the backup so that the api being down doesn't prevent success.

marun:
The change adding `--force` to the backup script was apparently not available yet? Trying again.

marun:
DR testing on gcp is not yet stable, but aws is looking promising. 

/test e2e-aws-disruptive

test/extended/dr/quorum_restore.go
```diff
@@ -137,17 +138,22 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Disruptive]", func() {
 						// there is a small chance the cluster restores the default replica size during
 						// this loop process, so keep forcing quorum guard to be zero, without failing on
 						// errors
-						scaleEtcdQuorum(pollClient, 0)
+						err = scaleEtcdQuorum(pollClient, 0)
```
smarterclayton, line 141: inline the err into the if unless you're using it outside this loop (which i don't see)

marun: Done.

PR #25718, opened 2020-11-30
Fix minor bit to make oc validation possible & remove unused test files

soltysh:
/assign @sttts


soltysh:
/hold 
failure is real

PR #25720, opened 2020-11-30
apiserver: remove terminates kube-apiserver gracefully test

rphillips:
This proposes removing the `kubelet terminates kube-apiserver gracefully` test. watch-termination does not use a flock to protect concurrent access to the log and lockfile. We changed the static pod UIDs to include an updated UID when they change, so the kubelet starts deleting UID1, while UID2 is starting.

Additionally, there is a case where apiserver is exiting with a 1 exit code. The test case looks at the entire event history looking for a NonGracefulTermination. This is problematic when there is a case where the exit code of 1 is triggered.

watch-termination logs: https://bugzilla.redhat.com/show_bug.cgi?id=1882750#c31

Either the test, exit code strategy, or watch-termination application need to be updated to account for these scenarios.

cc @sttts @deads2k

### DIFF
--- a/test/extended/apiserver/graceful_termination.go
+++ b/test/extended/apiserver/graceful_termination.go
@@ -17,28 +17,6 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Late]", func() {
 
 	oc := exutil.NewCLI("terminating-kube-apiserver")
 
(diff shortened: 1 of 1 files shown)


rphillips:
fixed here https://github.com/kubernetes/kubernetes/pull/98424

PR #25726, opened 2020-12-01
Bug 1896474: Add a subset of path based route tests that currently work

tnozicka:
Picking working tests from https://github.com/openshift/origin/pull/25676 so we don't regress in the meantime.


tnozicka:
cc @openshift/sig-network-edge

tnozicka:
actually this one isn't a bug per say, the tests that are commented out are.
/remove-kind

PR #25730, opened 2020-12-02
Fix cluster quota test to tolerate a new default config map in every namespace

tnozicka:
kube 1.20 brings in default configmap `kube-root-ca.crt` that is created in every namespace so we need to conditionaly adjust the CM counts.

/cc @sttts

### DIFF
--- a/test/extended/quota/clusterquota.go
+++ b/test/extended/quota/clusterquota.go
@@ -12,8 +12,10 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilversion "k8s.io/apimachinery/pkg/util/version"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
 
 	imagev1 "github.com/openshift/api/image/v1"
 	quotav1 "github.com/openshift/api/quota/v1"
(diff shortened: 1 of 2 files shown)


test/extended/quota/clusterquota.go
```diff
@@ -27,7 +29,23 @@ var _ = g.Describe("[sig-api-machinery][Feature:ClusterResourceQuota]", func() {
...
+			versionInfo, err := oc.KubeClient().Discovery().ServerVersion()
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			const kubeRootCAName = "kube-root-ca.crt"
+			version, err := utilversion.ParseSemantic(versionInfo.String())
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			expectKubeRootCACM := version.AtLeast(utilversion.MustParseGeneric("1.20"))
```
sttts, line 43: what does this mean for downgrades or upgrades?

sttts: I would rather unconditionally ignore the kube-root-ca configmap.

tnozicka: it a server version so if you run it before or after upgrade it won't matter. we don't run tests during upgrades afaik.

test/extended/util/wait.go
```diff
@@ -0,0 +1,41 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
+)
+
+func WaitForCMState(ctx context.Context, client corev1client.CoreV1Interface, namespace string, name string, condition func(cm *corev1.ConfigMap) (bool, error)) (*corev1.ConfigMap, error) {
```
sttts, line 17: go doc

sttts: Especialls that it falls over deletion

tnozicka: will add it in a follow up to have this in place for the rebase and avoid going through CI flakes

Deletion isn't a state but an edge, it may be deleted before you start looking or never created at all. Also you get modified currently with deltaFIFO on recreate so it should use watch based Until, to see the edge, not informer based function.

But it fails on unexpected event so you know right away and won't merge a flake.

tnozicka:
gcp infra
/retest

PR #25744, opened 2020-12-07
Bug 1904679: drop docker.io from image tests

soltysh:
/assign @dmage


wking:
Manually picked back in #27321.

PR #25746, opened 2020-12-07
Bug 1904684: get rid of docker.io in tests

soltysh:


test/extended/testdata/cmd/test/cmd/newapp.sh
```diff
@@ -480,9 +458,9 @@ os::cmd::expect_success 'oc delete all -l app=testapp1'
 os::cmd::expect_success 'oc delete all -l app=ruby --ignore-not-found'
 os::cmd::expect_success 'oc delete imagestreams --all --ignore-not-found'
 # newapp does not attempt to create an imagestream that already exists for a container image
-os::cmd::expect_success 'oc new-app docker.io/ruby:2.2'
+os::cmd::expect_success 'oc new-app docker.io/ruby:2.5'
```
adambkaplan, line 461: doesn't this still use docker.io?

adambkaplan (commented):
Generally looks good.

For wildfly replacements, `quay.io/wildfly/wildfly-centos-7` should fill the gap.

test/extended/testdata/cmd/test/cmd/newapp.sh
soltysh: It does :( There are a few sill left, but I can't find anything to replace that temporarily. I'll return to this after this merges.

soltysh: Also newapp.sh is on the blacklist https://github.com/openshift/origin/blob/ee82ae0d236efb69590dd4ed5ad29452eb62a4a0/test/extended/cmd/cmd.go#L28-L34 so it's not executed.

adambkaplan: language nit - we should rename this var to `skipList`. The usage of blacklist/whitelist is discouraged.

soltysh: This will disappear as soon as I'm done with all the migration.

adambkaplan:
LGTM

@soltysh do you plan on squashing commits?

soltysh:
> @soltysh do you plan on squashing commits?

Nope, I'm planning on re-freshing this PR a bit after the dust settles from k8s bump landing.

jaypoulz:
@soltysh is there a way to use images that are UBI based instead of centos?
Centos is not built for s390x, any images based on it will not be compatible for multi-arch tests.

soltysh:
> @soltysh is there a way to use images that are UBI based instead of centos?
> Centos is not built for s390x, any images based on it will not be compatible for multi-arch tests.

I'm fine with doing that as a follow-up, but I'll leave that up to you. Ping me for review, though.

adambkaplan (approved):
/lgtm

yselkowitz:
A number of the files modified in this commit are meant to be synced automatically from openshift/library, meaning that these changes will be lost on the next sync.  The better approach -- presuming both are similarly up to date -- would be to modify the imagestreams at their source to use registry.centos.org instead of docker.io, which would then benefit both OKD users as well as future syncs here.

As for using UBI, that's already being used in imagestreams where available.  However, the databases are not included in UBI and therefore that's not an option for these.

dmage:
It's a 4.7.0 fix, and I want to have it in 4.6.z, but I don't want to do it manually :) I don't think 4.6.z is synced with anything.

yselkowitz:
This was just merged to master though.

dmage:
But it was merged a year ago :)

soltysh:
Yeah, it's not that simple, we'll have to pick it up manually, I'm hoping to get to it tomorrow.

PR #25749, opened 2020-12-07
Bug 1905232: Append image test should be relative to the tools image

smarterclayton:
The tools image may have different layers over different releases
or build types (CI generates more than OCP right now). Make the
calculation relative to the tools image and double check that shell
image matches it.


fabianvf:
`e2e-aws-fips` test failing due to issues that should be fixed by this PR: https://github.com/openshift/cluster-monitoring-operator/pull/1008

fabianvf:
/hold
aforementioned PR is about to merge, want to wait for it to go in before we rerun

PR #25750, opened 2020-12-08
test: Simplify the worker machine health check test for clarity

smarterclayton:
In order to better debug failures with the machine health check,
clarify the test debugging logic and simplify the checks in place.

On GCP this exposes https://bugzilla.redhat.com/show_bug.cgi?id=1905709

### DIFF
--- a/test/extended/dr/machine_recover.go
+++ b/test/extended/dr/machine_recover.go
@@ -34,7 +34,6 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/disruption"
 	"github.com/openshift/origin/test/extended/util/disruption/controlplane"
-	"github.com/openshift/origin/test/extended/util/disruption/frontends"
 )
 
 var _ = g.Describe("[sig-cluster-lifecycle][Feature:DisasterRecovery][Disruptive]", func() {
@@ -44,8 +43,7 @@ var _ = g.Describe("[sig-cluster-lifecycle][Feature:DisasterRecovery][Disruptive
 
 	oc := exutil.NewCLIWithoutNamespace("machine-recovery")
 
-	g.It("[Feature:NodeRecovery] Cluster should survive master and worker failure and recover with machine health checks", func() {
-
+	g.It("[Feature:NodeRecovery] Cluster should survive worker failure and recover with machine health checks", func() {
 		framework.Logf("Verify SSH is available before restart")
 		masters, workers := clusterNodes(oc)
 		o.Expect(len(masters)).To(o.BeNumerically(">=", 3))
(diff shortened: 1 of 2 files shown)


hexfusion:
LGTM thanks @smarterclayton

test/extended/dr/machine_recover.go
```diff
@@ -77,123 +73,69 @@ var _ = g.Describe("[sig-cluster-lifecycle][Feature:DisasterRecovery][Disruptive
...
+				}
 
 				// we use a hard shutdown to simulate a poweroff
 				for _, target := range targets {
 					framework.Logf("Forcing shutdown of node %s", target.Name)
-					_, err = ssh("sudo -i systemctl poweroff --force --force", target)
+					ch := make(chan struct{})
+					go func(node *corev1.Node) {
+						defer close(ch)
+						if _, err := ssh("sudo -i systemctl poweroff --force --force", node); err != nil {
```
marun, line 94: `err` will always be non-nil on successful execution of this command so I'm not sure there's any value in checking it.. In #25707 I've added a new `shutdownNode` function that calls e2essh.SSH directly so that the output can be checked the only accurate indication that shutdown was initiated - the string `Powering off` appearing in stderr.

marun:
@smarterclayton Maybe rebase on #25707 to get related cleanup for free?

marun:
I imagine you're testing with clusterbot, but in case it's handy I've proposed adding `e2e-gcp-disruptive`: https://github.com/openshift/release/pull/14191

test/extended/dr/machine_recover.go
smarterclayton: I don't know that there's even a guarantee that powering off would get returned on all clouds.  it was not on GCP at the current time.  So I think even that isn't sufficient.

marun: How would it not be returned? And why would it be a function of a particular cloud? - afaict it's RHCOS that's sending `Powering off` to the ssh connection before initiating termination.

smarterclayton: It's a function of how the load balancers work.  GCP often leaves dangling connections, and since go SSH is limited (can't do idle timeouts correctly) you can't assume you'll ever get a packet (so the SSH hangs forever).  GCP cuts the proxy connection before you get the "powering off" packet, unlike AWS.

smarterclayton:
This also reproduces bug 1905709 on AWS

marun:
fyi quorum restore test is broken due to the addition of a mandatory dependency on the api when taking a backup intended to ensure that only healthy revisions of static pods are backed up. Fix is pending: https://github.com/openshift/cluster-etcd-operator/pull/509

PR #25755, opened 2020-12-10
allow operators to settle after MCPs are done upgrading

sjenning:
The upgrade test currently just waits for the MCPs to complete their rollout before immediately checking if all ClusterOperators are available and not degraded.  This is causing quite a few false failures in CI, `authentication` and `openshift-apiserver` in particular.

Example job
https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-azure-upgrade-4.7/1336793136516567040

The PR checks the ClusterOperators every 10s for up to 5m to see if they have settled after the MCP rollout.  If they don't settle within 5m, the test fails.

@deads2k @fabianvf

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -439,6 +439,37 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 		return err
 	}
 
+	if err := disruption.RecordJUnit(
+		f,
+		"[sig-cluster-lifecycle] ClusterOperators are available and not degraded after upgrade",
+		func() error {
+			framework.Logf("Waiting for operators to settle after upgrade")
+			if err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+				coList, err := c.ConfigV1().ClusterOperators().List(context.Background(), metav1.ListOptions{})
+				if err != nil {
+					framework.Logf("error getting ClusterOperators %v", err)
+					return false, nil
+				}
+				settled := true
+				for _, co := range coList.Items {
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade/upgrade.go
```diff
@@ -439,6 +439,29 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 		return err
 	}
 
+	framework.Logf("Waiting for operators to settle after upgrade")
```
deads2k, line 442: wrap in a 

```
disruption.RecordJUnit(		if err := disruption.RecordJUnit(
		f,			f,
		"[sig-mco] Machine config pools complete upgrade",			"[sig-mco] Machine config pools complete upgrade",
		func() error {
```

block like above please

test/e2e/upgrade/upgrade.go
```diff
@@ -439,6 +439,29 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
...
+		coList, err := c.ConfigV1().ClusterOperators().List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			return false, err
+		}
+		settled := true
+		for _, co := range coList.Items {
+			// should match failure cases checked in versionMonitor Describe() in monitor.go
+			available := findCondition(co.Status.Conditions, configv1.OperatorAvailable)
+			degraded := findCondition(co.Status.Conditions, configv1.OperatorDegraded)
+			if !conditionHasStatus(available, configv1.ConditionTrue) ||
```
deads2k, line 453: track all the ones not settled for the junit message.

deads2k:
ugly, but workable.  put it in the junit block so you get info on how long it takes and I would take it.

PR #25765, opened 2020-12-14
Bug 1907373: Tests for kube rebase to 1.20.0

soltysh:
This is #25764 with fixed deps.
/assign @tnozicka @marun

PR #25766, opened 2020-12-15
Skip table tests broken by 1.20 rebase

marun:
The following tests fail consistently after the merge of the 1.20 rebase, and this is blocking testing and merge of most repos. Rather than wait for the [1.20 origin bump](https://github.com/openshift/origin/pull/25765) to fix these tests, skip and unblock everyone.

/cc @mfojtik @sttts @tnozicka @soltysh


wking (approved):
/lgtm

smarterclayton:
Why are the tests broken?  This will break skew tests, so we’ll need to fix them or have them be skipped when the skew env var is set.

marun:
> Why are the tests broken? This will break skew tests, so we’ll need to fix them or have them be skipped when the skew env var is set.

As per this PRs description, both tests will be fixed by the origin bump. Skipping just unblocks everyone else while we get the bump into a mergeable condition.

soltysh:
Revert was added to #25765

PR #25769, opened 2020-12-15
test: Update images README and add OWNERS

smarterclayton:
Document the review criteria as part of accepting rebase.

### DIFF
--- a/test/extended/util/image/README.md
+++ b/test/extended/util/image/README.md
@@ -17,7 +17,11 @@ for easy mirroring by the `openshift-tests images` command.
    * Describe your use case in a PR to this file and have it approved
    * Define a standard CI build and publish the image to quay.io in the openshift namespace
    * Add the new reference to the `init()` method in this package
-   * Have the automation promote the image to the quay mirror location.
+   * Reference your image inside your tests tests using `github.com/openshift/origin/test/extended/util/image.LocationFor("my.source/image/location:versioned_tag")`
+   * Regenerate the verify output with `make update` and compare the diff to see your source image located
+   * Ensure your tests fail with "image cannot be pulled" errors.
+   * Have the reviewer promote the image to the quay mirror location.
+   * The reviewer should approve the PR and merge.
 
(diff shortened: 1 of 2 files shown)


bparees:
Can we get some documentation that explains what this process actually does?

Things i'm confused about:

1) if i add a new test that uses a new image (assume i have a good reason), does that image get listed in images.go's init function as an exception?  I wouldn't think so, but the current docs seem to tell you to do that.  I would have assumed that the update tooling would just "find" my image reference in my test and add it to the zz_generated.txt

2) does something being listed in zz_generated.txt result in it being mirrored?  What is doing that mirroring and when?

3) What actually happens at runtime?  My impression is that the idea behind this is that all the image strings in test code get substituted w/ the mirror (quay.io or whatever is specified on the command line) location prior to test execution?  Is that accurate?

I also put some questions in https://github.com/openshift/origin/pull/24887/files#r538841935 that could use clarification.

soltysh (approved):
/lgtm

soltysh (approved):
/lgtm

test/extended/util/image/README.md
```diff
@@ -17,7 +17,11 @@ for easy mirroring by the `openshift-tests images` command.
    * Describe your use case in a PR to this file and have it approved
    * Define a standard CI build and publish the image to quay.io in the openshift namespace
    * Add the new reference to the `init()` method in this package
-   * Have the automation promote the image to the quay mirror location.
+   * Reference your image inside your tests tests using `github.com/openshift/origin/test/extended/util/image.LocationFor("my.source/image/location:versioned_tag")`
```
bparees, line 20:
```suggestion
   * Reference your image inside your tests using `github.com/openshift/origin/test/extended/util/image.LocationFor("my.source/image/location:versioned_tag")`
```

test/extended/util/image/README.md
```diff
@@ -17,7 +17,11 @@ for easy mirroring by the `openshift-tests images` command.
    * Describe your use case in a PR to this file and have it approved
    * Define a standard CI build and publish the image to quay.io in the openshift namespace
```
bparees, line 18: To the best of my understanding CI builds don't push to quay.io nor to the openshift namespace, so i'm confused by this.  

CI builds push to registry.svc.ci.openshift.org and generally in the ocp namespace (though certainly they can push to other namespaces)

PR #25773, opened 2020-12-17
Bug 1907938: Nodes goes into NotReady state (VMware)

p0lyn0mial:
we will carry a patch that allows the clients (Kubelet, KCM, etc.) to detect broken connections to the api server more quickly than the default TCP timeout of 15 minutes.

Note: this patch will be dropped in future releases once http2/ping is introduced.


soltysh (approved):
/lgtm
/approve

p0lyn0mial:
/hold

I need to regenerate the dependencies, `verify` keeps failing.

p0lyn0mial:
> I need to regenerate the dependencies, verify keeps failing.

I had to change the commit msg only.

/hold cancel
/lgtm

markmc:
(patch) affects real-world users, already released on 4.6

markmc:
> > I need to regenerate the dependencies, verify keeps failing.
> 
> I had to change the commit msg only.
> 
> /hold cancel
> /lgtm

Ok, I'll add lgtm back

/lgtm

PR #25778, opened 2021-01-04
Bug 1895053: Verify builds can mount proxy trustedCA

adambkaplan:
Verify builds can mount the cluster's custom PKI trust bundle if one is
set on the cluster proxy configuration. When `mountProxyTrustedCA` is
set to `true`, builds should add the container's trust bundle as a
read-only mount. The custom PKI should be readable during the build
execution, but should not be present in the resulting image.

### DIFF
--- a/test/extended/builds/proxy.go
+++ b/test/extended/builds/proxy.go
@@ -1,8 +1,11 @@
 package builds
 
 import (
+	"context"
+
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	buildv1 "github.com/openshift/api/build/v1"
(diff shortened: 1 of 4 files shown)


adambkaplan:
@bparees something seems to be up with this PR. It is waiting on the status of a job that has never run.

ci/prow/e2e-metal-ipi-ovn-ipv6

bparees:
@openshift/openshift-team-developer-productivity-test-platform can you take a look at why this PR is stuck?  it wants a job result for a context that doesn't seem to exist.

we can manually merge it if need be, but i'd rather find a way to properly push it through.

bparees:
> /test e2e-metal-ipi-ovn-ipv6

wait....that worked?  why didn't the override work then?

alvaroaleman:
> wait....that worked? why didn't the override work then?

Override only works with a status that exists and that status didn't exist

PR #25779, opened 2021-01-06
Enable test skipping via a regex supplied with --skip

marun:
Since golang regex explicitly doesn't support negative lookahead, skipping tests is most effectively done by a regex not
matching. Following ginkgo's example, this change adds a `--skip` argument that can be used to exclude the matching tests from suite execution.

This change is suggested by the need to skip tests matching `.*\[Skipped:Disruptive\].*` when executing the parallel suite after a run of the disruptive suite. Without the ability to skip these checks, the disruptive e2e jobs cannot be guaranteed to consistently succeed.

/cc @smarterclayton


smarterclayton:
We would run the disruptive suite, then run the conformance suite (two successive openshift-tests execution).  I didn’t expect disruptive tests to be in conformance, so what would be skipped?

marun:
> We would run the disruptive suite, then run the conformance suite (two successive openshift-tests execution). I didn’t expect disruptive tests to be in conformance, so what would be skipped?

Synthetic tests that have been annotated to be skipped:

https://github.com/openshift/origin/blob/skip-regex/test/extended/util/annotate/rules.go#L113

I've seen all 3 of the synthetic tests in question fail disruptive jobs.

smarterclayton:
Oh... synthetic tests being run as part of all suites is... not something i would have expected to be implemented like this.  I'll have to look at it, but I don't want behavior of skips on suites, synthetics should be excluded from suites that don't make sense.  The general addition of synthetic "late" tests to non conformance suites is probably not something that should happen in general.  I would probably say synthetics have to be explicitly allowlisted on a suite, not blocked.

smarterclayton:
Or maybe I don't understand - this test should not be running in the disruptive suite to start with (and if it is, that's the bug).

smarterclayton:
Oh.  I see what you are talking about now, sorry.  "How we represent valid conformance tests that won't pass after disruption"

I think valid conformance tests that fail after disruption should be fixed so they pass after disruption in general.  

For the test you mentioned:

```
evs, err := client.CoreV1().Events("openshift-kube-apiserver").List(context.TODO(), metav1.ListOptions{})
```

should be filtered to be from the start of the e2e, or otherwise exclude valid disruption induced prior to test start.

marun:
> For both the tests currently skipped under disruptive:
> 
> ```
> evs, err := client.CoreV1().Events("openshift-kube-apiserver").List(context.TODO(), metav1.ListOptions{})
> ```
> 
> should be filtered to be from the start of the e2e, or otherwise exclude valid disruption induced prior to test start. It is useful to have those run against the time during install, but not (to my knowledge) highly valuable. So we have the option of filtering out events prior to start of test, or explicitly setting a fence condition when we run the conformance suite that gates the time interval (only consider events after time X). I guess it would come down to an argument of how valuable invariant tests are against the install behavior (which maybe stefan wants to weigh on in)

I think we have sufficient coverage of the skipped tests in other jobs that it's not worth ensuring their execution in the post-disruption parallel suite, and that worrying about how to enable them worth neither the effort or the complication.

cc @sttts

smarterclayton:
I disagree, and the  point of getting parallel up to snuff (tests work after disruption) is to do things like run parallel after upgrade and block release promotion on that, as well as disruptive.  You're not hacking around to fix disruptive, you're getting us ready to gate far more (https://github.com/openshift/release/pull/14626 is just the first step)

PR #25780, opened 2021-01-06
Bug 1908217: server-side apply e2e: fix oauthclientauthorization race with KCM

stlaz:
When the oauthclientauthorization is created, it checks the SA from
its clientName for tokens and fails if there are none.

The server-side apply test creates an SA but does not wait for it to
be populated with tokens which may cause the test to fail when
it gets to creating the oauthauthorization object too early.

/assign @marun

### DIFF
--- a/test/extended/apiserver/apply.go
+++ b/test/extended/apiserver/apply.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
@@ -14,9 +15,11 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	discocache "k8s.io/client-go/discovery/cached"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/kubernetes/test/e2e/framework"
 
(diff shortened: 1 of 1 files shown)


test/extended/apiserver/apply.go
```diff
@@ -172,3 +181,16 @@ func createResource(
...
 }
+
+// waitForSATokens waits for the kube-controller-manager to populate the SA from the unstructured object
+// to be populated with a secret defining its tokens
+func waitForSATokens(saClient corev1client.ServiceAccountInterface, unstructuredObj map[string]interface{}) {
+	name, _, err := unstructured.NestedString(unstructuredObj, "metadata", "name")
+	o.Expect(err).NotTo(o.HaveOccurred())
+	err = wait.PollImmediate(time.Second, 15*time.Second, func() (bool, error) {
+		sa, err := saClient.Get(context.Background(), name, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
```
marun, line 192: I'm surprised to see a hard assertion in a polling loop. Is it your intention that a transient failure fail the test vs logging the error and returning false?

marun (commented):
Good catch!

test/extended/apiserver/apply.go
stlaz: You're right, given the wait is only 15 seconds even if the error does not go away there's no need to fail early.

Fixed now.

test/extended/apiserver/apply.go
```diff
@@ -172,3 +181,19 @@ func createResource(
...
 	return resourceClient, &unstructuredObj, testNamespace
 }
+
+// waitForSATokens waits for the kube-controller-manager to populate the SA from the unstructured object
+// to be populated with a secret defining its tokens
+func waitForSATokens(saClient corev1client.ServiceAccountInterface, unstructuredObj map[string]interface{}) {
+	name, _, err := unstructured.NestedString(unstructuredObj, "metadata", "name")
+	o.Expect(err).NotTo(o.HaveOccurred())
+	err = wait.PollImmediate(time.Second, 15*time.Second, func() (bool, error) {
+		sa, err := saClient.Get(context.Background(), name, metav1.GetOptions{})
```
sttts, line 191: isn't there a variant of PollImmediate with context?

test/extended/apiserver/apply.go
```diff
@@ -172,3 +181,19 @@ func createResource(
 
 	return resourceClient, &unstructuredObj, testNamespace
 }
+
+// waitForSATokens waits for the kube-controller-manager to populate the SA from the unstructured object
+// to be populated with a secret defining its tokens
+func waitForSATokens(saClient corev1client.ServiceAccountInterface, unstructuredObj map[string]interface{}) {
```
sttts, line 187: odd interface. Would be simpler with just the name as string, and the string extracted at the call site.

stlaz: The function is very specific to this test case, I merely extracted the necessary steps not to pollute the code above.

test/extended/apiserver/apply.go
stlaz: I couldn't find one in `k8s.io/apimachinery/pkg/util/wait/wait.go`

PR #25786, opened 2021-01-08
Bug 1914343: Replacing centos with ubi in image-annotation test.

jaypoulz:
Centos is not compatible with s390x, resulting in observed flakeyness when this is run on s390x.


ricardomaraschini (approved):
/lgtm

PR #25787, opened 2021-01-11
test: Fail if nodes go unready during a normal parallel run

smarterclayton:
The standard conformance test requires that nodes remain available.
Some suites may deliberately introduce node disruption, so this is
not a general synthetic.


smarterclayton:
Tagging to enforce upgrade is blocked by down nodes afterwards

PR #25789, opened 2021-01-11
test: Only consider rule evaluation failures in current window

smarterclayton:
Some pre-test conditions like upgrade or disruption could result
in rule evaluation failures, so we should check only over the
relevant interval.

Found while testing 4.6->4.7 full e2e upgrades in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/14666/rehearse-14666-release-openshift-origin-installer-e2e-aws-upgrade-4.6-stable-to-4.7-ci/1348694129546104832

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -49,6 +49,8 @@ type Options struct {
 	DryRun        bool
 	PrintCommands bool
 	Out, ErrOut   io.Writer
+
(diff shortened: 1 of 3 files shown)


smarterclayton:
@marun another one that may impact you - it triggers on upgrade, and if we force deleted the worker node that had one of the prometheus it could trigger in the machine health check test

smarterclayton:
Tagging so we can turn on gating

PR #25791, opened 2021-01-13
Bug 1907939: Nodes goes into NotReady state (VMware)

p0lyn0mial:
we will carry a patch that allows the clients (Kubelet, KCM, etc.) to detect broken connections to the api server more quickly than the default TCP timeout of 15 minutes.

Note: this patch will be dropped in future releases once http2/ping is introduced.


vendor/k8s.io/kubernetes/staging/src/k8s.io/client-go/transport/cache.go
```diff
@@ -89,10 +90,7 @@ func (c *tlsTransportCache) get(config *Config) (http.RoundTripper, error) {
 
 	dial := config.Dial
 	if dial == nil {
-		dial = (&net.Dialer{
-			Timeout:   30 * time.Second,
-			KeepAlive: 30 * time.Second,
-		}).DialContext
+		dial = libgonetwork.DefaultClientDialContext()
```
p0lyn0mial, line 93: this is what we want, compare with https://github.com/openshift/origin/pull/25773

glide.lock
```diff
@@ -13,7 +13,7 @@ imports:
 - name: github.com/andreyvit/diff
   version: c7f18ee00883bfd3b00e0a2bf7607827e0148ad4
 - name: github.com/apparentlymart/go-cidr
-  version: c0fb5fbe0acb592411e2db59add389a43260ad44
+  version: d07067359aee8ee9101c1a79a4f286cf0cea059f
```
p0lyn0mial, line 16: is there an easy way to find out why this dep was updated?

sttts: I thought glide needs manual dependency updats. I would remove those manually. Does library-go have the same changes?

p0lyn0mial: I decided to pin them manually. Library-go does have some of them but no all. 
Anyway, now the diff is much smaller.

sttts:
Now verify is unhappy.

glide.yaml
```diff
@@ -114,13 +114,13 @@ import:
...
-  version: release-4.4
+  version: 3e8f22fb0b56808d981b0a98cd2741d31368aec5
 - package: github.com/openshift/client-go
   version: release-4.4
 - package: github.com/openshift/library-go
-  version: release-4.4
+  version: 993204ef63fd55794c4782936c0d369b22b1e6a1
 - package: github.com/openshift/apiserver-library-go
-  version: release-4.4
+  version: ba8b80421765a458f9c50f767bbb5013b5bd5fdc
```
sttts, line 123: We should keep the branch names. Glide has this nice feature of symbolic versions. It makes the next bump easy.

p0lyn0mial: sure, np.

p0lyn0mial: done, slightly bigger diff `15 files` vs `20 files` - not bad.

PR #25795, opened 2021-01-13
[release-3.11] Bug 1760807: Revert #25463 regression in pod affinity check

damemi:
#25463 tried addressing additional bugs that were uncovered when fixing this parent issue, and apparently regressed earlier fixes. This reverts that pr

/priority important-soon
/kind regression


damemi (commented):
/cc @ingvagabund @soltysh

ingvagabund:
@damemi do you plan addressing the original issue in another PR?

damemi:
> @damemi do you plan addressing the original issue in another PR?

@ingvagabund yeah. the main issue was fixed in https://github.com/openshift/origin/pull/25350, meaning that an invalid selector no longer prevented valid pods from being created (this was the important/cluster-breaking part)

The issue now is that an invalid pod is sometimes still scheduled (though it doesn't affect other pods from being created). This is much less severe, but still a problem that requires more digging

damemi:
This is failing on docker rate limiting

damemi:
/hold
this is still failing with no output from CI
@mfojtik could we override these tests?

PR #25800, opened 2021-01-15
Bug 1916660: UPSTREAM: 97916: kube-aggregator: fix apiservice availability gauge

dgrisonnet:
When an apiservice is deleted, its relative
aggregator_unavailable_apiservice metric remains with the value of the
last availability observed. Hence, if an apiservice is deleted while
being unavailable, the metric remains marked as unavailable.
This presents some problems when alerting on unavailable apiservices
as deleted apiservices might trigger the alert indefinitely.

To solve this issue, the aggregator_unavailable_apiservice metric should
only reflect the availability of existing apiservices.

This is achievable by using a custom Collector instead of a GaugeVec and
create throw-away metrics based on an apiservice lister output. With
this approach, on deletion, the apiservice will not be listed anymore,
resulting in its availability metric not being exposed.

### DIFF
--- a/vendor/k8s.io/kubernetes/staging/src/k8s.io/kube-aggregator/pkg/controllers/status/BUILD
+++ b/vendor/k8s.io/kubernetes/staging/src/k8s.io/kube-aggregator/pkg/controllers/status/BUILD
@@ -40,7 +40,10 @@ go_library(
 
 go_test(
     name = "go_default_test",
-    srcs = ["available_controller_test.go"],
+    srcs = [
+        "available_controller_test.go",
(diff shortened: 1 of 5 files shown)


wangke19:
Launched one 4.5 cluster with this PR successfully, verification steps see https://bugzilla.redhat.com/show_bug.cgi?id=1916660#c3

dgrisonnet:
@sttts the `ci/prow/e2e-agnostic-cmd` job seems to be failing constantly for unrelated reasons. Is there anything for me to do in order to have this PR merged?

PR #25803, opened 2021-01-15
Bug 1896188: wait for RC presence before starting debug session

soltysh:
/assign @ingvagabund


soltysh:
/hold
debugging failures

PR #25738, opened 2020-12-03
Bug 1902702: fix ldap group sync tests and move them from bash to golang

stlaz:
/assign @soltysh 
cc @sallyom 

/hold
it's not yet finished, but perhaps already worth a PR

Currently, on my machine, the test takes ~4minutes, which is a huge difference from the original ~20min. It's caused by the original test actually starting a pod and installing a bunch of deps in it...

**edit:** with all tests added, it's now around 8 minutes, but that's still quite an improvement

PR #25808, opened 2021-01-19
Bug 1902008: fix ldap group sync tests and move them from bash to golang

stlaz:
This is a manual backport of https://github.com/openshift/origin/pull/25738, adding it as a fix to the above BZ as it removes one out of the two docker pulls in the test

/assign @soltysh


soltysh (approved):
/lgtm
/approve

PR #25811, opened 2021-01-19
tests: Synthetic tests moved up and out of test infra

smarterclayton:
The disruption suite needs to skip many of the synthetics because
the invariants aren't expected to be maintained. Hoist those use
cases up and out of the code.


smarterclayton:
Tests running multiple times, probably double registered

smarterclayton:
Disruptive test suite must not run tests that don't make sense (synthetics moved up).

/retest

PR #25815, opened 2021-01-21
Bug 1902580: bug fix for e2e test case: [sig-arch] Managed cluster should ensure pods use downstream images [Suite:openshift/conformance/parallel] from our release image with proper ImagePullPolicy

qinpingli:

### DIFF
--- a/test/extended/operators/images.go
+++ b/test/extended/operators/images.go
@@ -117,12 +117,18 @@ var _ = Describe("[sig-arch] Managed cluster", func() {
 					invalidPodContainerImagePullPolicy.Insert(fmt.Sprintf("%s/%s/%s imagePullPolicy=%s", pod.Namespace, pod.Name, container.Name, container.ImagePullPolicy))
 				}
 			}
+
+			if pod.Status.Phase != v1.PodRunning {
+				continue
+			}
 			// check if the container's image from the downstream.
 			for j := range pod.Spec.Containers {
 				containerName := pod.Spec.Containers[j].Name
 				commands := []string{
 					"exec",
 					pod.Name,
+					"-n",
+					pod.Namespace,
 					"-c",
 					containerName,
 					"--",
@@ -163,16 +169,18 @@ func ignoredNamespace(namespace string) bool {
 	// TODO components in openshift-operators may not come from our payload, may want to weaken restriction
 	namespacePrefixes := sets.NewString("kube-", "openshift-")
 	ignoredNamespacePrefixes := sets.NewString("openshift-marketplace", "openshift-must-gather-")
+
+	ret := true
 	for _, prefix := range namespacePrefixes.List() {
-		if !strings.HasPrefix(namespace, prefix) {
-			return true
+		if strings.HasPrefix(namespace, prefix) {
+			ret = false
 		}
 	}
 	for _, prefix := range ignoredNamespacePrefixes.List() {
 		if strings.HasPrefix(namespace, prefix) {
-			return true
+			ret = true
 		}
 	}
-	return false
+	return ret
 
 }
--- a/test/extended/operators/images.go
+++ b/test/extended/operators/images.go
@@ -127,8 +127,6 @@ var _ = Describe("[sig-arch] Managed cluster", func() {
 				commands := []string{
 					"exec",
 					pod.Name,
-					"-n",
-					pod.Namespace,
 					"-c",
 					containerName,
 					"--",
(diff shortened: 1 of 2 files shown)


test/extended/operators/images.go
```diff
@@ -163,16 +169,18 @@ func ignoredNamespace(namespace string) bool {
 	// TODO components in openshift-operators may not come from our payload, may want to weaken restriction
 	namespacePrefixes := sets.NewString("kube-", "openshift-")
 	ignoredNamespacePrefixes := sets.NewString("openshift-marketplace", "openshift-must-gather-")
+
+	ret := true
 	for _, prefix := range namespacePrefixes.List() {
-		if !strings.HasPrefix(namespace, prefix) {
-			return true
+		if strings.HasPrefix(namespace, prefix) {
```
mandre, line 175: This change make the `ignoredNamespace()` function to return false for everything that start with `kube-` or `openshift-`, even for the `openshift-marketplace` and `openshift-must-gather-` prefixes we want to ignore.

The previous implementation seemed correct to me.

test/extended/operators/images.go
```diff
@@ -117,12 +117,18 @@ var _ = Describe("[sig-arch] Managed cluster", func() {
...
 				}
 			}
+
+			if pod.Status.Phase != v1.PodRunning {
+				continue
+			}
 			// check if the container's image from the downstream.
 			for j := range pod.Spec.Containers {
 				containerName := pod.Spec.Containers[j].Name
 				commands := []string{
```
mandre, line 127: I think the problem is that this checks the image is based on RHEL, but doesn't check that it comes from a trusted source. Do we want to hardcode the URL of the image registry in the tests or is there a better way?

test/extended/operators/images.go
```diff
@@ -117,12 +117,18 @@ var _ = Describe("[sig-arch] Managed cluster", func() {
...
 					"exec",
 					pod.Name,
+					"-n",
+					pod.Namespace,
```
mandre, line 131: This certainly doesn't hurt :+1:

test/extended/operators/images.go
qinpingli: @mandre Thanks for the review.  For we have multiple prefixes here, such as `kube-`, `openshift-`, then without the fix, for the example namespace `openshift-cluster-storage-operator`, it will always be ignored with the following loop(it doesn't have prefixed `kube-` and `openshift-`):
```
for _, prefix := range namespacePrefixes.List() {
		if !strings.HasPrefix(namespace, prefix) {
			return true
               }
}
```

qinpingli: With this loop, my fix might have this issue too when the prefix list changed, update my patch too.

test/extended/operators/images.go
qinpingli: I didn't notice `oc.SetNamespace(pod.Namespace)` here, now I fixed the following error from `SetNamespace` function.
```
Feb  7 06:40:32.129: INFO: unable to run command:[exec service-ca-operator-bbb9b5cf7-29z7g -c service-ca-operator -- cat /etc/redhat-release] with error: exit status 1
Feb  7 06:40:32.130: INFO: Running 'oc --kubeconfig=/home/piqin/flexy_output/flexy/workdir/install-dir/auth/kubeconfig exec service-ca-8676df98ff-4hvkv -c service-ca-controller -- cat /etc/redhat-release'
Feb  7 06:40:32.762: INFO: Error running /usr/bin/oc --kubeconfig=/home/piqin/flexy_output/flexy/workdir/install-dir/auth/kubeconfig exec service-ca-8676df98ff-4hvkv -c service-ca-controller -- cat /etc/redhat-release:
StdOut>
Error from server (NotFound): pods "service-ca-8676df98ff-4hvkv" not found
StdErr>
Error from server (NotFound): pods "service-ca-8676df98ff-4hvkv" not found

```

test/extended/operators/images.go
mandre: Hmmm... I understand that the original loop ignores everything that does not start with `kube-` or `openshift-`, but still ignores `openshift-marketplace` and `openshift-must-gather-`.

As such, `openshift-cluster-storage-operator` has the `openshift-` prefix and shouldn't be ignored. Am I missing something?

test/extended/operators/images.go
mandre: Thinking more about this, I think this should be enough to check that every image is pulled from the same registry and namespace as the Release Info image.

test/extended/operators/images.go
qinpingli: for the `openshift-cluster-storage-operator` namespace, with the profixes list ("kube-", "openshift-"), when it enters the following loop:
```
for _, prefix := range namespacePrefixes.List() {
		if !strings.HasPrefix(namespace, prefix) {
			return true
               }
}
```
It doesn't have the prefix `kube-`, then it returns true, then the ns is ignored.

mandre: Oh, I see, you're right. The function will be clearer like the following I think:
```
// ignoredNamespace() returns true if the namespace is to be ignored by the test
func ignoredNamespace(namespace string) bool {
	// a pod in a namespace that begins with kube-* or openshift-* must come from our release payload
	// TODO components in openshift-operators may not come from our payload, may want to weaken restriction
	ignoredNamespacePrefixes := sets.NewString("openshift-marketplace", "openshift-must-gather-")

	ignore := true
	for _, prefix := range namespacePrefixes.List() {
		if strings.HasPrefix(namespace, prefix) {
			ignore = false
			break
		}
	}
	for _, prefix := range ignoredNamespacePrefixes.List() {
		if strings.HasPrefix(namespace, prefix) {
			ignore = true
			break
		}
	}
	return ignore
}
```

qinpingli: @mandre Thank you! Updated.

PR #25819, opened 2021-01-21
 add additional settling debug information

deads2k:
We recently had a BZ about operators not achieving a stable level.  This adds extra debugging to the error message so we know which operator and how long it has been unstable.

TRT identified the problem and https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_openshift-apiserver/167/pull-ci-openshift-openshift-apiserver-master-e2e-aws-upgrade/1351976530204430336 is an example.

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/openshift/origin/pkg/monitor"
+	"github.com/openshift/origin/test/extended/util/operator"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/onsi/ginkgo/config"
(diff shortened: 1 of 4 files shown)


pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -231,6 +232,13 @@ func (opt *Options) Run(args []string) error {
 	includeSuccess := opt.IncludeSuccessOutput
 	if len(tests) == 1 {
 		includeSuccess = true
+	} else {
```
smarterclayton, line 235: This doesn't belong here, it belongs as an `[Early]` test (which is what `[sig-arch][Early] Managed cluster should start all core operators [Suite:openshift/conformance/parallel]` does).

By putting this here you break jobs that are expected to have unsettled operators at the start, of which we don't *want* to wait for them.

Please move into the Early test

smarterclayton: If you had a concern about the early not running in a given suite, that's easy to fix.

smarterclayton: Also... you know this information about operators going unready is already in the event log for all runs, right?  The one you linked has multiple lines about the insights operator:

```
Jan 20 20:59:11.561 E clusteroperator/insights changed Degraded to True: PeriodicGatherFailed: Source clusterconfig could not be retrieved: Get "https://10.0.135.71:10250/containerLogs/openshift-sdn/sdn-64h84/sdn?limitBytes=65536&sinceSeconds=86400": dial tcp 10.0.135.71:10250: connect: connection refused
Jan 20 21:15:41.056 W clusteroperator/insights changed Degraded to False
```

So just add a standard synthetic (operators should not go degraded or unavailable during a run)

PR #25798, opened 2021-01-14
Bug 1915907: replace test/cmd/builds.sh direct/indirect docker.io refs with imagestream image refs

gabemontero:
note at least some of these files were touched by clayton's huge refactor, but a few refs slipped the cracks

/assign @bparees 
(in Adam's absence)

thanks @dmage for the heads up

PR #25821, opened 2021-01-22
Bug 1919453: Update to latest 4.7 state where possible

smarterclayton:
Update all tests to be accurate to 4.7 state, except where obvious 
conflicts exist.


smarterclayton:
Depends on https://github.com/openshift/kubernetes/pull/536

smarterclayton:
@deads2k @gabemontero this is a reset of 4.6 tests code to 4.7 state, removing any backports that were due to changes in kube code packages (maybe 3-4 of those), and then the last commit is places where code compiled but the 4.7 code doesn't pass, so i reverted those as separate commits.  Would squash before merging, but this was all selective reversion to the 4.6 state.

If you check this branch out, it's easiest to review 

```
git diff HEAD..origin/master -- 'test/**/*.go'
```

which shows you all deliberate code differences between this branch and master (deleted lines are 4.6, master is green, so anything in green is something that doesn't pass on 4.6)

gabemontero:
fwiw @smarterclayton after both `git diff HEAD..origin/master -- 'test/**/*.go'`, and `git diff HEAD..origin/master -- 'test/**/*.json'` I saw nothing amiss with what is there, though I admittedly looked closer at the e2e's I'm more familiar with.

I also confirmed my add on after your initial 4.7 drop is there, adding the s2iroot image to the disconnected inventory

Lastly, some test/cmd/build.sh stuff got missed in your initial 4.7 drop as well.  I covered those in 4.7 with https://github.com/openshift/origin/pull/25798 and 4.6 with https://github.com/openshift/origin/pull/25805 ... that said I see several other `test/extended/testdata/cmd/test/cmd` hits in this PR, including moves from docker.io (or implied docker.io) refs to imagestream refs.  So my assumption is your searches etc. caught all those type items... you at least were on the look out.

smarterclayton:
Getting close to updating this again, now that a bunch of 4.7 fixes have gone in.

smarterclayton:
/retest

but I may be missing one more fix from 4.7

PR #25825, opened 2021-01-25
Bug 1902054: Update multicast test image for use in release-4.6

abhat:
In order to get around docker rate limiting, the e2e tests
need to pull images off either quay or the local image
registry. This change updates the multicast image to be
pulled from the local image registry.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>

### DIFF
--- a/test/extended/networking/multicast.go
+++ b/test/extended/networking/multicast.go
@@ -167,7 +167,7 @@ func launchTestMulticastPod(f *e2e.Framework, nodeName string, podName string) e
 			Containers: []kapiv1.Container{
 				{
 					Name:    contName,
-					Image:   "openshift/test-multicast",
+					Image:   "quay.io/openshift/community-e2e-images:e2e-docker-io-openshift-test-multicast-latest-4AxcBBxKg_prX34z",
 					Command: []string{"sleep", "1000"},
 				},
 			},


abhat:
cc @gabemontero.

test/extended/networking/multicast.go
```diff
@@ -10,7 +10,6 @@ import (
 	networkclient "github.com/openshift/client-go/network/clientset/versioned/typed/network/v1"
 	"github.com/openshift/library-go/pkg/network/networkutils"
 	exutil "github.com/openshift/origin/test/extended/util"
-
 	kapiv1 "k8s.io/api/core/v1"
```
danwinship, line 13: keep the blank line there

test/extended/networking/multicast.go
```diff
@@ -167,7 +166,7 @@ func launchTestMulticastPod(f *e2e.Framework, nodeName string, podName string) e
 			Containers: []kapiv1.Container{
 				{
 					Name:    contName,
-					Image:   "openshift/test-multicast",
+					Image:   "quay.io/openshift/community-e2e-images:e2e-docker-io-openshift-test-multicast-latest-4AxcBBxKg_prX34z",
```
danwinship, line 169: That URL seems pretty sketchy. Is that actually stable?

abhat: it's what we have in [4.7](https://github.com/openshift/origin/blob/master/test/extended/util/image/zz_generated.txt#L12). I couldn't find a better way to indirect other than using the whole `image.LocationFor` API which doesn't exist today in 4.6 at all.

PR #25834, opened 2021-01-28
test: etcd leader changes should be measured over the test, not the run

smarterclayton:
Our guarantee is that no leader changes should occur during test runs,
not during installation or scenarios prior to the test. Use the new
test duration to constrain the metric by time. Verified against a cluster
that the resulting query is correct.

### DIFF
--- a/test/extended/etcd/leader_changes.go
+++ b/test/extended/etcd/leader_changes.go
@@ -16,15 +16,19 @@ import (
 var _ = g.Describe("[sig-etcd] etcd", func() {
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLI("etcd-leader-change").AsAdmin()
-	g.It("leader changes are not excessive", func() {
+	g.It("leader changes are not excessive [Late]", func() {
 		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
 		o.Expect(err).ToNot(o.HaveOccurred())
-		g.By("Examining the rate of increase in the number of etcd leadership changes for last five minutes")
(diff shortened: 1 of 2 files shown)


hexfusion:
thanks @smarterclayton 

/lgtm

test/extended/etcd/leader_changes.go
```diff
@@ -16,15 +16,19 @@ import (
...
 		prometheus, err := client.NewE2EPrometheusRouterClient(oc)
 		o.Expect(err).ToNot(o.HaveOccurred())
-		g.By("Examining the rate of increase in the number of etcd leadership changes for last five minutes")
-		result, _, err := prometheus.Query(context.Background(), "increase((max by (job) (etcd_server_leader_changes_seen_total) or 0*absent(etcd_server_leader_changes_seen_total))[15m:1m])", time.Now())
+
+		// we only consider series sent since the beginning of the test
+		testDuration := exutil.DurationSinceStartInSeconds().String()
+
+		g.By("Examining the number of etcd leadership changes over the run")
+		result, _, err := prometheus.Query(context.Background(), fmt.Sprintf("increase((max(max by (job) (etcd_server_leader_changes_seen_total)) or 0*absent(etcd_server_leader_changes_seen_total))[%s:15s])", testDuration), time.Now())
```
wking, line 27: would it be cheaper to do `"max(etcd_server_leader_changes_seen_total) - max(etcd_server_leader_changes_seen_total offset %s)", testDuration`, or would we have to fill in the `or 0*absent...` stuff and make that too stuttery?

smarterclayton: `increase(` is special - it tracks counter resets.  You have to max over them.  It's not expensive to do this query (it's the same data you have to load to do increase) as a subquery.  Absent may not be necessary, but i don't want to refactor that right now.

wking: #25870 says `absent` wasn't necessary after all ;)

PR #25835, opened 2021-01-28
Bug 1921857: Router should require multiple 200 responses (hit multiple instances of the router) before continuing the test

smarterclayton:
Most routers run in HA configuration, so accepting just a single
200 response is not sufficient if the other router is trailing.
Change the test to verify we get 5 successive in a row before
returning.

Identified by a single flake in https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-sdn-network-stress-4.7/1353930135601942528

### DIFF
--- a/test/extended/router/scoped.go
+++ b/test/extended/router/scoped.go
@@ -241,24 +241,31 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 func waitForRouterOKResponseExec(ns, execPodName, url, host string, timeoutSeconds int) error {
 	cmd := fmt.Sprintf(`
 		set -e
-		STOP=$(($(date '+%%s') + %d))
+		pass=%[4]d
+		STOP=$(($(date '+%%s') + %[1]d))
 		while [ $(date '+%%s') -lt $STOP ]; do
 			rc=0
-			code=$( curl -k -s -m 5 -o /dev/null -w '%%{http_code}\n' --header 'Host: %s' %q ) || rc=$?
+			code=$( curl -k -s -m 5 -o /dev/null -w '%%{http_code}\n' --header 'Host: %[2]s' %[3]q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
 				echo $code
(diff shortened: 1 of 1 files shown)


smarterclayton:
@frobware @Miciah this was one of 2 flakes in over 10k runs of the network stress tests, we need to ensure both routers have picked up our change in these e2es before we continue to others, this is similar to code we have in a few other tests (5 retries may not be enough, and I need to verify that repeated curls actually get randomly distributed among all frontends in all clouds, but it should be close)

smarterclayton:
Assuming even load balancing between 2 instances, there is a 31/32 chance you hit both instances in five retries.  I'll check other platforms to see whether sticky load balancing is likely to happen (but if it does, this bug doesn't trigger anyway).

test/extended/router/scoped.go
```diff
@@ -241,24 +241,31 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
...
 		while [ $(date '+%%s') -lt $STOP ]; do
 			rc=0
-			code=$( curl -k -s -m 5 -o /dev/null -w '%%{http_code}\n' --header 'Host: %s' %q ) || rc=$?
+			code=$( curl -k -s -m 5 -o /dev/null -w '%%{http_code}\n' --header 'Host: %[2]s' %[3]q ) || rc=$?
 			if [[ "${rc:-0}" -eq 0 ]]; then
 				echo $code
 				if [[ $code -eq 200 ]]; then
-					exit 0
+					pass=$(( pass - 1 ))
+					if [[ $pass -lt 0 ]]; then
```
Miciah, line 253: Off by one?  If `pass` starts at 5, this logic curls the route 6 times, which is unintuitive.

smarterclayton: Yeah, will change to `-le`.

frobware:
test-images fix has merged https://github.com/openshift/origin/pull/25839

/retest

smarterclayton:
We're green

PR #25837, opened 2021-01-29
Bug 1923737: Add `--fail-fast` to the tests command

smarterclayton:
The `--fail-fast` flag will exit the suite when any test fails. The `--count` flag now accepts -1 to run forever.

Early and late tests are not repeated, and a failure in an earlier phase will not execute tests in later phases.

Also:

32f7be4ef7 (Clayton Coleman, 4 minutes ago)
   test: Don't initialize the cloud provider in run command

   The suite doesn't need to init the cloud provider, each test can accomplish
   that.

8d7bfb87d1 (Clayton Coleman, 34 minutes ago)
   test: Ensure early|late tests are run with other suites

   These tests are not just part of parallel conformance.


smarterclayton:
/hold

Still testing

smarterclayton:
/retest

Both external flakes

PR #25839, opened 2021-01-29
Bug 1904684: fix missing bits in images.sh tests

soltysh:
This is a followup to https://github.com/openshift/origin/pull/25746 I missed some instances which were causing this test to fail consistently.

/assign @sallyom


sallyom:
/lgtm cancel 

waiting for 1 more fix

sallyom:
> /lgtm cancel
> 
> waiting for 1 more fix

/hold  
found 1 more flake, will add the fix here

PR #25842, opened 2021-01-29
Bug 1896558: Skip multiaz e2e

marun:
Pick up skip of flakey upstream multiaz test.


deads2k:
1. this test only fails on GCP: https://sippy.ci.openshift.org/testdetails?release=4.7&test=spread%20the%20pods%20of%20a%20service%20across%20zones
2. it looks like 1/20 services running on GCP will not be balanced across zones.  
3. the reason for skipping is listed as, "the bug is only a medium and we never focused on it before"
4. I see no reason to believe that skipping this test will increase the likelihood that we will fix this scheduling problem on GCP.

It seems to me that we should devote effort to fixing the underlying problem that (apparently) causes this feature to break.

/hold

If we can find proof that the test is broken and the scheduling itself works fine, then it seems more reasonable to skip this.

PR #25844, opened 2021-01-29
Bug 1900989: idle.sh: Drop endpoints mocks from idle tests

sgreene570:
**test/extended/testdata/cmd/test/cmd/idle.sh:**
Remove the need for deleting the test service by removing the custom
runtime-created endpoints. Instead, block test setup on the number of
endpoint addresses actually available in the test endpoint resource.
By being able to keep the service, this test is now compatible with
recent changes to `oc idle`, in which `oc idle` annotates a service
directly (in addition to available endpoints).

**test/extended/testdata/bindata.go:**
Regenerate

---

This PR should unblock CI for https://github.com/openshift/oc/pull/720


soltysh (approved):
Given that our test-cmd are executed on an actual cluster there's no need for fake endpoits.
/lgtm
/approve

PR #25846, opened 2021-02-01
test: Use wildfly 20 since 19 can no longer be accessed at quay

smarterclayton:
The test was looking for an inaccessible image.


smarterclayton:
Merging to clear the queue

Issue #14048, opened 2017-05-04
Extended.idling and unidling idling with a single service and DeploymentConfig [Conformance] should idle the service and DeploymentConfig properly


ncdc:
From https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_extended_conformance_gce/1638/consoleFull

```
• Failure [53.339 seconds]
idling and unidling
/openshifttmp/openshift/build-rpm-release/tito/rpmbuild-origin8YATB3/BUILD/origin-3.6.0/_output/local/go/src/github.com/openshift/origin/test/extended/idling/idling.go:468
  idling
  /openshifttmp/openshift/build-rpm-release/tito/rpmbuild-origin8YATB3/BUILD/origin-3.6.0/_output/local/go/src/github.com/openshift/origin/test/extended/idling/idling.go:276
    with a single service and DeploymentConfig [Conformance]
    /openshifttmp/openshift/build-rpm-release/tito/rpmbuild-origin8YATB3/BUILD/origin-3.6.0/_output/local/go/src/github.com/openshift/origin/test/extended/idling/idling.go:264
      should idle the service and DeploymentConfig properly [It]
      /openshifttmp/openshift/build-rpm-release/tito/rpmbuild-origin8YATB3/BUILD/origin-3.6.0/_output/local/go/src/github.com/openshift/origin/test/extended/idling/idling.go:263

      Expected error:
          <*util.ExitError | 0xc421cfa030>: {
              Cmd: "oc idle --config=/tmp/extended-test-cli-idling-lr7hx-15j1n-user.kubeconfig --namespace=extended-test-cli-idling-lr7hx-15j1n --resource-names-file /tmp/openshift-extended-tests/idling-services-820636018",
              StdErr: "error: unable to mark service \"extended-test-cli-idling-lr7hx-15j1n/idling-echo\" as idled: endpoints \"idling-echo\" is forbidden: endpoint address 172.16.4.79 is not allowedDeploymentConfig \"extended-test-cli-idling-lr7hx-15j1n/idling-echo\" has been idled",
              ExitError: {
                  ProcessState: {
                      pid: 28512,
                      status: 256,
                      rusage: {
                          Utime: {Sec: 0, Usec: 264425},
                          Stime: {Sec: 0, Usec: 28687},
                          Maxrss: 49084,
                          Ixrss: 0,
                          Idrss: 0,
                          Isrss: 0,
                          Minflt: 13901,
                          Majflt: 0,
[...]

PR #25847, opened 2021-02-01
Re-enable test/extended/idling

frobware:
The gist of this was to verify that, if this had been enabled, we would have noticed that idling was broken in the router, see https://bugzilla.redhat.com/show_bug.cgi?id=1900989. However this test doesn't exercise a route and we would have still missed the regression in BZ#1900989.

TODO/TBD:
- land this as is
- extend this to also exercise a route

### DIFF
--- a/test/extended/idling/idling.go
+++ b/test/extended/idling/idling.go
@@ -197,8 +197,9 @@ func checkSingleIdle(oc *exutil.CLI, idlingFile string, resources map[string][]s
 		{
 			Replicas: 2,
 			CrossGroupObjectReference: unidlingapi.CrossGroupObjectReference{
-				Name: resources[resourceName][0],
-				Kind: kind,
+				Name:  resources[resourceName][0],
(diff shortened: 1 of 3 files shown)


frobware:
Some historical context: https://github.com/openshift/origin/issues/14048#issuecomment-328388467

frobware:
Tests marked local are not run: https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/openshift-hack/e2e/annotate/rules.go#L215. Is it reasonable to drop "[Local]" in these tests?

frobware:
I created an explicit e2e test for the router here:https://github.com/openshift/origin/pull/25850.

Will continue to see how what is required to make these tests non-local.

frobware:
Going with https://github.com/openshift/origin/pull/25850 for 4.7. Will continue this outside of https://bugzilla.redhat.com/show_bug.cgi?id=1900989.

PR #25849, opened 2021-02-02
test: Upgrade should prereq early tests

smarterclayton:
Ensure we don't start before install completes or any operators are metastable


smarterclayton:
Closes a flake condition where operators are unstable on startup after install

/retest

PR #25850, opened 2021-02-03
Bug 1900989: test/extended/router: add idle/unidle e2e test

frobware:
An explicit test to assert that a service that is idled will be
 unidled when a GET request occurs on the route.

### DIFF
--- a/test/extended/router/idle.go
+++ b/test/extended/router/idle.go
@@ -1,0 +1,147 @@
+package router
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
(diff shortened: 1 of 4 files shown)


frobware:
This will continue to fail in CI until we pick up the latest router commits:

From slack & @sgreene570:

Looks like the router image isn't using latest HEAD commit from the openshift/router repo. Any ideas why?
ie when I run oc adm release info --commits | grep router for 4.7.0-0.ci-2021-02-03-104157, I get the hash 25b1656, which is from Jan 20, but more commits merged on jan 28, so the current build commit should be 30e2fb5 from my understanding (See https://github.com/openshift/router/commits/master)

PR #25861, opened 2021-02-05
Bug 1920221: Don't initialize zone info repeatedly

smarterclayton:
About 1/6 GCP jobs fails because each test in a suite has to initialize the cloud provider (tests run as sub processes of the suite executor) and the GCP suite lists all zones during initialization. The remainder of the time in initialization is pretty fast, so submit upstream changes that add Zones to the TestContext.CloudConfig and allow the values to be defaulted there. The suite then loads the zones we have machines on once at suite initialization time and ensures each test invocation uses it via the cloud provider config passed down.

Also identify two other significant sources of failures - the pointless per test waits that make node calls (and often flake, causing tests that should be skipped to report a failure), and some of the per suite waits. We make two changes there - a change to ginkgo to let us clear the upstream pre/post suite behavior (which is mostly debug info related, and we ignore), and a change to kube to allow the waits per test to be bypassed when passing -1 to the test.

This should dramatically speed up test execution, deflake errors across all providers that make API calls, and reduce test time.


smarterclayton:
agnostic failing with GCP failure to hit apiserver 10.0.0.2:6443 with i/o timeout

smarterclayton:
/retest

serial failing several times with a hostIP exclusion, it is possible we have a new flake with same node pod cleanup

smarterclayton:
/retest

likely there is a new crio bug getting picked up here

PR #25862, opened 2021-02-05
Bug 1915912: Fix CSI snapshotter image version

jsafrane:
This fixes some niche e2e tests, see the BZ.

In addition, this is pre-requisite for enabling snapshot e2e tests, which we disabled during rebase and forgot to enable in https://bugzilla.redhat.com/show_bug.cgi?id=1906518.

I filed https://bugzilla.redhat.com/show_bug.cgi?id=1925493 for enabling those tests.

PR #25863, opened 2021-02-05
Change registry.svc.ci.openshift.org to registry.ci.openshift.org

soltysh:
While mirroring image for https://github.com/openshift/origin/pull/25862 I've noticed we're still pointing to deprecated images in registry.svc.ci.openshift.org. I've updated them during mirroring but here is the actual PR fixing those.

/assign @adambkaplan 

since you own those bits.


adambkaplan (approved):
/lgtm

PR #25812, opened 2021-01-20
Test that new single replica topology API is taken into account

omertuc:
Tests that deployments have a single replica when https://github.com/openshift/enhancements/pull/555 control plane topology is set to SingleReplica.

This enhancement is yet to be merged as well as the relevant behavior from the various operators, so this PR is in draft mode until then.

PR #25869, opened 2021-02-08
Upgrade openshift/api to 670ac3fc997c

omertuc:
Needed by #25812


deads2k:
/lgtm

for future reference, this could be a part of the PR making use of the API.  No need to do it in advance.

PR #25870, opened 2021-02-09
Bug 1924953: Allow prometheus state clear to pass etcd leader change test

smarterclayton:
If the prometheus process on a conformant cluster has no PVs, the
data in prometheus is lost and the current query will flag that as
a leader election (new prometheus scrapes the old etcd and sees the
increase). The tightened tolerance turned out to be too tight for
serial tests, where we allow the machine set test to replace one
of the workers. If we were using PVs for prometheus in all machine
set clusters this would be ok, but it's valid today to have a
machine set cluster with no PV provisioning.

The test now ignores restarts that happen before the beginning of
prometheus history, and only looks at leader changes over the time
prometheus is up and running with the latest data.


smarterclayton:
/cherry-pick release-4.7

Will be bug 1926556

smarterclayton:
Unrelated, force merging.

PR #25874, opened 2021-02-09
Bug 1926146: test/extended/router/idle: address flakes/failures seen in CI 

frobware:
Fixes for failures/flakes seen in CI runs:

- increase timeout for polling operations to 15 mins

- wait for PODs to be running after the fixture has been created

This also switches from using URLTester to making a direct connection
to the route's hostname. This reduces the need for an execpod and will
make the test run faster.

The changes here will also make the test pass for the proxy CI job.

### DIFF
--- a/test/extended/router/idle.go
+++ b/test/extended/router/idle.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"time"
 
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 
@@ -14,7 +16,6 @@ import (
 
 	unidlingapi "github.com/openshift/api/unidling/v1alpha1"
 	exutil "github.com/openshift/origin/test/extended/util"
-	"github.com/openshift/origin/test/extended/util/url"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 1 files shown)


frobware:
/hold

Reproduced (just the once at the moment) in local testing.

test/extended/router/idle.go
```diff
@@ -25,36 +26,54 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
...
+				// Accumulate current state of all
+				// resources in the test namespace. Do
+				// this here as Namespace.Name is nil
+				// in AfterEach() when the test has
+				// failed.
+				resourceState, _ = oc.Run("get").Args("all", "-o", "yaml").Output()
+			}()
+
+			g.By("Waiting for pods to be running")
+			err = waitForRunningPods(oc, 1, exutil.ParseLabelsOrDie("app=idle-test"), urlTimeout)
```
sgreene570, line 66: Use `timeout` here instead of `urlTimeout`?

frobware: I have one reproducer in local testing today and will fiddle with all the timeouts we're using. Think they're all just too low at the moment.

frobware: Unequivocally, we shouldn't need second test for "pods are running" because a 200 GET is effectively the same test. I'll remove that - just wanted more evidence about where the flakes occur. Seems it can be in various places at the moment, not just one aspect of the test.

frobware:
This test is now passing in the proxy job:

$ ag -n HAProxy e2e.log.1 |grep GET
11357:started: (6/1960/2676) "[sig-network-edge][Conformance][Area:Networking][Feature:Router] The HAProxy router should be able to connect to a service that is idled because a GET on the route will unidle it [Suite:openshift/conformance/parallel/minimal]"
12069:passed: (1m3s) 2021-02-10T16:33:36 "[sig-network-edge][Conformance][Area:Networking][Feature:Router] The HAProxy router should be able to connect to a service that is idled because a GET on the route will unidle it [Suite:openshift/conformance/parallel/minimal]"

knobunc (approved):
/approve

Thanks @frobware

PR #21782, opened 2019-01-14
[WIP] remove oapi

deads2k:
realistically, I'm not going to get to this for a while.

@soltysh @sttts post-rebase I suppose,

PR #25880, opened 2021-02-10
Fix and re-enable test/extended/images/signatures.go

mtrmac:
- Use `docker://` instead of `atomic:`, as recommended back in https://github.com/openshift/origin/pull/21782#issuecomment-459077188 https://github.com/openshift/origin/pull/21782#discussion_r253249564
- Then re-enable the test

### DIFF
--- a/test/extended/images/signatures.go
+++ b/test/extended/images/signatures.go
@@ -83,24 +83,27 @@ var _ = g.Describe("[sig-imageregistry][Serial][Suite:openshift/registry/serial]
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(out).To(o.ContainSubstring("keyring `/var/lib/origin/gnupg/secring.gpg' created"))
 
-		// Create kubeconfig for skopeo
+		// Create kubeconfig for oc
 		g.By("logging as a test user")
 		out, err = pod.Exec("oc login https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT --token=" + token + " --certificate-authority=/run/secrets/kubernetes.io/serviceaccount/ca.crt")
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(out).To(o.ContainSubstring("Logged in"))
 
 		// Sign and copy the memcached image into target image stream tag
-		// TODO: Fix skopeo to pickup the Kubernetes environment variables (remove the $KUBERNETES_MASTER)
 		g.By("signing the memcached:latest image and pushing it into openshift registry")
 		out, err = pod.Exec(strings.Join([]string{
-			"KUBERNETES_MASTER=https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT",
 			"GNUPGHOME=/var/lib/origin/gnupg",
-			"skopeo", "--debug", "copy", "--sign-by", "joe@foo.bar",
+			"skopeo", "--debug",
+			// Disable the default-docker: file sigstore default in /etc/containers/registries.d, so that the X-Registry-Supports-Signatures protocol is used.
+			// Newer versions of Skopeo default to X-R-S-S if present, this test (as of 2021-02) uses skopeo-0.1.40-11.el7_8.x86_64, which defaults to sigstore.
+			"--registries.d", "/this/does/not/exist",
+
+			"copy", "--sign-by", "joe@foo.bar",
 			"--dest-creds=" + user + ":" + token,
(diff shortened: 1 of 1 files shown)


test/extended/images/signatures.go
```diff
@@ -83,24 +82,27 @@ var _ = g.Describe("[sig-imageregistry][Serial][Suite:openshift/registry/serial]
...
+			"skopeo", "--debug",
+			// Disable the default-docker: file sigstore default in /etc/containers/registries.d, so that the X-Registry-Supports-Signatures protocol is used.
+			// Newer versions of Skopeo default to X-R-S-S if present, this test (as of 2020-02) uses skopeo-0.1.40-11.el7_8.x86_64, which defaults to sigstore.
+			"--registries.d", "/this/does/not/exist",
+
+			"copy", "--sign-by", "joe@foo.bar",
 			"--dest-creds=" + user + ":" + token,
 			// TODO: test with this turned to true as well
 			"--dest-tls-verify=false",
 			"docker://docker.io/library/memcached:latest",
```
mtrmac, line 104: Is this external reference registry OK to include, or is there a recommended set of other testing images to use nowadays?

soltysh: Given docker rate-limiting and the requirement to run these tests offline, I'd encourage to pick something from these https://github.com/openshift/origin/blob/master/test/extended/util/image/zz_generated.txt

dmage: It's better to use `test/extended/util/image.ShellImage()` unless you need a specific image.

mtrmac: The only possible concern with an internal image is that the registry has a limited number of signature slots (per image, not per ISTag). Updated, let’s see.

mtrmac: That breaks, apparently because `ShellImage` is `docker://image-registry.openshift-image-registry.svc:5000/openshift/tools:latest` (and that is accessible using the users’ token), but the deduplicated image (same image ID) has `dockerImageReference`:`registry.build01.ci.openshift.org/ci-op-$private/stable@sha256:7932f3a766ddd6a0ee541fef4f40e51e440c9b7bfcf0f138023f6b2efd8cec74` , and that reference to the same underlying image) seems not to be readable the same same token.

We don’t need any _specific_ image but we need an unique image that is not deduplicated to an unaccessible repo — or extra privileges to read the `Image.dockerImageReference` repo.

mtrmac: The test user should have a `system-image:auditor` role, and thus access to all in-cluster images.

If I understand the failure ( https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25880/pull-ci-openshift-origin-master-e2e-aws-image-registry/1359960263960301568 ) correctly (which I’m not at all sure about), `ShellImage()` is an ImageStreamTag referring to `registry.build01.ci.…` (which is _external_ to the cluster, is that right?), with `referencePolicy.type: Source`. So, users of `ShellImage` are expected to contain an external registry directly, using some appropriate credentials.

But https://github.com/openshift/oc/blob/9f54c1d4f68c8530ac9466c655a4e55eb04a1459/pkg/cli/admin/verifyimagesignature/verify-signature.go#L264 blindly passes the current OpenShift token to any registry, even if remote. That looks rather incorrect?!

mtrmac:
/hold

The current version passes, by signing the just-built `signer` image. After signing `test/extended/util/image.ShellImage()` `verify-image-signature` fails, I think in a way that does not prevent re-enabling this test for the basic test case (but suggests that there _is_ a bug to be fixed); anyway, marking on hold to make sure https://github.com/openshift/origin/pull/25880#discussion_r574867309 is discussed.

soltysh:
@mtrmac it looks like the current approach is viable and works, any objections to merging this as is or you want time to resolve the issue you've mentioned in https://github.com/openshift/origin/pull/25880#issuecomment-777907683 ?

mtrmac:
I’m fine with re-enabling the test as is, and dealing with that bug (if it is a bug) separately; I wanted to draw attention to the failure to confirm that I’m interpreting it correctly.

soltysh (approved):
/lgtm
/approve
/hold cancel
/retest

mtrmac:
Updated just strings/commits, no logic:
- so that the new comment says “as of 2021-02”, not “2020-02”
- to update the test operation description in `g.By`
- so that the commit message no longer says WIP
- to re-enable the test in the last commit, not in an intermediate still-broken state.

soltysh (approved):
/lgtm

PR #25882, opened 2021-02-10
test: Calculate leader elections over individual metrics

smarterclayton:
The change to calculation worked when the metric was new (because the serial test cleared the PV data) but not when it was restarted after we added the PV retention.

Simplify the metric with the new data from when we have preserved the change.


smarterclayton:
Force merging to unblock testing.  Has passed 7 times across two releases with no flakes (so confident it fixes the query).

PR #25890, opened 2021-02-11
[release-3.11] Bug 1884421: Try to promote inactive routes following route deletion

Miciah:
Backport https://github.com/openshift/router/pull/126 (modulo potentially incompatible changes related to unit tests).

#### Try to promote inactive routes following route deletion

Before this patch, inactive routes were not being promoted when the conflicting routes were deleted. Fix it so that when deletes happen, all inactive routes are given a chance to be promoted.


Miciah:
Hitting rate limiting on image pulls.  Let's try one test at a time.  
/test cmd

PR #25900, opened 2021-02-16
Bug 1929110: test: add vsphere to unsupported platforms for LB service

vrutkovs:
Required to pass tests in vsphere upgrade workflow.

Ref: https://issues.redhat.com/browse/OTA-308


vrutkovs:
Flakes

/retest

PR #25904, opened 2021-02-17
test: Require no alerts during upgrades

smarterclayton:
It is unacceptable to fire alerts during normal upgrades. Tighten
the constraints on the upgrade tests to ensure that any alerts
fail the upgrade test. If the test is skipped, continue to report
which alerts fired. Test over the entire duration of the test like
we do in the post-run variation.

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -3,44 +3,39 @@ package alert
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	"github.com/prometheus/common/model"
 
 	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/disruption"
 	helper "github.com/openshift/origin/test/extended/util/prometheus"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/test/e2e/framework"
-	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	"k8s.io/kubernetes/test/e2e/upgrades"
 )
 
-const (
-	// Delay after upgrade is complete before checking for critical alerts
-	alertCheckSleepMinutes = 5
-	alertCheckSleep        = alertCheckSleepMinutes * time.Minute
-
-	// Previous period in which to check for critical alerts
-	alertPeriodCheckMinutes = 1
-)
-
-// UpgradeTest runs post-upgrade after alertCheckSleep delay and tests if any critical alerts are firing.
+// UpgradeTest runs verifies invariants regarding what alerts are allowed to fire
+// during the upgrade process.
 type UpgradeTest struct {
 	url         string
 	bearerToken string
 	oc          *exutil.CLI
 }
 
-func (UpgradeTest) Name() string { return "check-for-critical-alerts" }
+func (UpgradeTest) Name() string { return "check-for-alerts" }
 func (UpgradeTest) DisplayName() string {
-	return "[sig-arch] Check if critical alerts are firing after upgrade success"
+	return "[sig-arch] Check if alerts are firing during or after upgrade success"
 }
 
 // Setup creates parameters to query Prometheus
 func (t *UpgradeTest) Setup(f *framework.Framework) {
-	g.By("Setting up post-upgrade alert test")
+	g.By("Setting up upgrade alert test")
 
 	oc := exutil.NewCLIWithFramework(f)
 	url, _, bearerToken, ok := helper.LocatePrometheus(oc)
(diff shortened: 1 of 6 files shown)


test/e2e/upgrade/alert/alert.go
```diff
@@ -54,48 +45,49 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
 
 // Test checks if any critical alerts are firing.
 func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade upgrades.UpgradeType) {
-	g.By("Checking for critical alerts")
+	g.By("Checking for alerts")
 
-	// Recover current test if it fails so test suite can complete
-	defer g.GinkgoRecover()
```
wking, line 60: I'm not clear on what this was about.  Maybe @jottofar remembers from #24786.

test/e2e/upgrade/alert/alert.go
```diff
@@ -54,48 +45,49 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
...
 	<-done
 
-	ctx, cancel := context.WithCancel(context.Background())
-
 	// Additonal delay after upgrade completion
-	g.By("Waiting before checking for critical alerts")
-	time.Sleep(alertCheckSleep)
-	cancel()
+	g.By("Waiting before checking for alerts")
+	time.Sleep(2 * time.Minute)
```
wking, line 58: Not clear to me why we're reducing from `alertCheckSleepMinutes = 5` to 2m here.  I don't have strong feelings either way, but it's nice when folks who do feel like a constant needs tweaking explain their motivation for the tweak, so the rest of us don't forget to consider whatever the motivation was in future changes.

test/e2e/upgrade/alert/alert.go
```diff
@@ -54,48 +45,49 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
...
-	// Query to check if Prometheus has been up and running for entire post-upgrade
-	// period by verifying Watchdog alert has been in firing state
-	watchdogQuery := fmt.Sprintf(`count_over_time(ALERTS{alertstate="firing",alertname="Watchdog", severity="none"}[%dm])`, alertCheckSleepMinutes)
+	testDuration := time.Now().Sub(start).Truncate(time.Second)
+
+	// Query to check if Prometheus has been up and running for entire upgrade
+	// period by verifying Watchdog alert has been in firing state. To account for
+	// range start/end clipping, add one to the numerator and ensure we are greater
+	// than or equal to one.
+	watchdogQuery := fmt.Sprintf(`((count_over_time(ALERTS{alertstate="firing",alertname="Watchdog",severity="none"}[%s:1s])+1) / %d) >= 1`, testDuration.String(), int(testDuration.Seconds()))
```
wking, line 72: I don't think the `.String()` is necessary for `%s` formatting (here and a few lines below).

test/e2e/upgrade/alert/alert.go
```diff
@@ -54,48 +45,49 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
 
 // Test checks if any critical alerts are firing.
```
wking, line 46: nit: drop `critical` here too.

test/e2e/upgrade/alert/alert.go
smarterclayton: Actually, I'm wondering if we should be checking for zero pending alerts here.  Since the wait is basically arbitrary.

test/e2e/upgrade/alert/alert.go
jottofar: As I recall I was having a failure and the Ginko output said " you should add GinkoRecover". This would have been early on when I was still finding my way and I can't say for sure that it is still needed.

test/e2e/upgrade/alert/alert.go
michaelgugino: Pending alerts might flap.  In the case where we don't alert for a condition if the control plane is not ready (or in the future if a particular node is down), as soon as the control plane is ready or the node is up, we will meet the criteria of the alert for pending if a pod can't come up in the monitoring interval.

jottofar: I originally made this 5 minutes based on my test runs and what I thought would always be long enough to ensure the upgrade is complete without being excessively long.
Seems the use of constants is different than what I'm used to although not necessarily wrong.

wking: Things happen during updates that may trip alerts into `pending` without anything actually concerning happening.  I suspect we want to allow that.  But even if we decide we don't want to allow `pending` alerts, this current pivot to blocking almost all `firing` alerts is already big enough, that I'd prefer to punt `pending` discussion to follow-up work.

test/e2e/upgrade/alert/alert.go
smarterclayton: it's more of a warning, it's not necessary

test/e2e/upgrade/alert/alert.go
smarterclayton: I'm changing this so that after upgrade there should be no pending alerts.  I see none that aren't serious bugs in the few samples I have here.

I'm changing the interval to 1m and requiring no pending.

smarterclayton:
I changed the pending check, it didn't make sense (you can't know by looking at a prometheus instance when it was up, because you can't tell the difference between planned outage and crash).  Now the check is "is the watchdog alert firing continuously during upgrade EXCEPT when this other rule is not firing".  The real invariant we're trying to maintain is that "if you query thanos-querier for the range of time over the upgrade, the watchdog query shows up" which is really "thanos querier sees all the data and doesn't violate its SLO to be up continuosly across safe rolling updates" which is a product SLO.

test/e2e/upgrade/alert/alert.go
```diff
@@ -52,50 +43,56 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
...
-	// period by verifying Watchdog alert has been in firing state
-	watchdogQuery := fmt.Sprintf(`count_over_time(ALERTS{alertstate="firing",alertname="Watchdog", severity="none"}[%dm])`, alertCheckSleepMinutes)
+	testDuration := time.Now().Sub(start).Truncate(time.Second)
+
+	// The watchdog alert should be firing continuously during the whole upgrade except when prometheus
+	// is stopped, which the joined query represents by looking at a second metric that is always reported
+	// when the cluster is running. We do not use the prometheus_* metrics because they can't be reviewed
+	// historically.
+	// TODO: switch to using thanos querier and verify that we can query the entire history of the upgrade
+	watchdogQuery := fmt.Sprintf(`sum_over_time((ALERTS{alertstate="firing",alertname="Watchdog",severity="none"} * scalar(absent(max(openshift:prometheus_tsdb_head_samples_appended_total:sum*0+1))))[%s:1s]) > 0`, testDuration)
```
wking, line 73: Poking at [the CI job][1] with PromeCIeus: update [from 2021-02-22T20:35:58Z to 2021-02-22T21:40:29Z][2], and Prom was down from 21:24 to 21:26:

image

Is Prom going down like that mid-update acceptable?  If it is, I think we should drop the `Watchdog` guard.  If not, I think we should drop the `scalar(absent(...` business.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1363938413538971648
[2]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1363938413538971648/artifacts/e2e-gcp-upgrade/gather-extra/artifacts/clusterversion.json

test/e2e/upgrade/alert/alert.go
```diff
@@ -52,50 +43,56 @@ func (t *UpgradeTest) Setup(f *framework.Framework) {
...
+
+	// No non-info level alerts should have fired during the upgrade
+	firingAlertQuery := fmt.Sprintf(`count_over_time(ALERTS{alertname!~"Watchdog|AlertmanagerReceiversNotConfigured",alertstate="firing",severity!="info"}[%s])`, testDuration)
 
-	// Query to check for any critical severity alerts that have occurred within the last alertPeriodCheckMinutes.
-	criticalAlertQuery := fmt.Sprintf(`count_over_time(ALERTS{alertname!~"Watchdog|AlertmanagerReceiversNotConfigured|KubeAPILatencyHigh",alertstate="firing",severity="critical"}[%dm]) >= 1`, alertPeriodCheckMinutes)
+	// There should be no pending alerts 1m after the upgrade completes
+	pendingAlertQuery := `ALERTS{alertname!~"Watchdog|AlertmanagerReceiversNotConfigured",alertstate="pending",severity!="info"}`
```
wking, line 79: nit: I think we can drop `Watchdog` here, because it should never be `pending`.

wking:
[Update job][1] hit [the `hostPort: Invalid value: 10301: Host ports are not allowed to be used` thing][2] and also failed the `Watchdog` guard (empty results `[]`) and the `firing` guard (`ClusterOperatorDegraded` for `authentication` and `openshift-apiserver`, among others), 

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1363972787651219456
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1913069

test/e2e/upgrade/alert/alert.go
smarterclayton: This is the thanos-querier change i mentioned in slack.

smarterclayton:
Last failure was an actual upgraed failure (nodes were still upgrading after cluster allegedly reached level = bad).  

/test e2e-gcp-upgrade

smarterclayton:
Specifically excluded etcdMemberCommunicationSlow from pending - it is reasonable for it to go pending due to the IO latency caused by the node OS upgrade impacting disk latency impacting etcd network latency

smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1365103688250560512 Watchdog was unavailable for 2s, needs investigation

/retest

smarterclayton:
TargetDown is completely flawed for upgrades if you have a per node component (upgrade takes longer than 10m and if machine upgrades are run efficiently it will almost always end up firing because one node is always upgrading) is root cause of https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1366154898957144064.  We need to change TargetDown to not do that (i will look at that this afternoon), and it's also an SRE alert they complain about.

/retest

smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1366424416887508992 caught https://bugzilla.redhat.com/show_bug.cgi?id=1933847 which is that prometheus went down (was unavailable) for 4m during the upgrade because it got scheduled to the same node and had no PDB (needs antiaffinity and PDB)

smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1366489577979645952 Is prometheus down

/retest

smarterclayton:
Another both prometheus instances on the same node, no PDB

/retest

wking:
All green :).  Dunno how disruptive actually landing this will be, but reverts are cheap if it turns out to be pressing us harder than we like :p.

wking:
Waiting on openshift/cluster-monitoring-operator#1072 to avoid excessive pain once this lands.

/hold

smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1369043944154861568 is https://github.com/operator-framework/operator-lifecycle-manager/pull/2024 which is still under review

/retest

smarterclayton:
/test e2e-gcp-upgrade

Was the pods going back to pending issue

smarterclayton:
/retest

I think this is ready for merge (pending is always flake, we have all the exceptions catalogued with bugs), will push it through

smarterclayton:
I am going to merge this to start preventing regressions of new alerts.  We have all known alerts over the last ~20 or so runs in an exception or fixed.

Goodbye alerts during upgrade.

PR #25905, opened 2021-02-18
Bug 1930663: test/e2e/upgrade: skip paused worker pools when checking for update to complete

vrutkovs:
Upgrade test ensures all pool updates are complete during upgrade tests. In order to test  EUS upgrade testing with minimal worker pool disruption steps would set paused in the spec. Upgrade test should consider those updated as it would be verified on later stages.

Ref: https://issues.redhat.com/browse/OTA-359

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -505,6 +505,15 @@ func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (bool
 		framework.Logf("error getting pool %s: %v", name, err)
 		return false, nil
 	}
+	paused, found, err := unstructured.NestedBool(pool.Object, "spec", "paused")
+	if err != nil || !found {
+		return false, nil
+	}
+	if paused && (name == "worker") {
+		framework.Logf("worker pool paused, considering it to be updated")
+		return true, nil
+	}
+
 	conditions, found, err := unstructured.NestedFieldNoCopy(pool.Object, "status", "conditions")
 	if err != nil || !found {
 		return false, nil


vrutkovs:
Verified that the test works on a temp cluster - pause setting can be toggled

/retest

test/e2e/upgrade/upgrade.go
```diff
@@ -422,10 +458,16 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 				}
 				allUpdated := true
 				for _, p := range pools.Items {
-					updated, err := IsPoolUpdated(mcps, p.GetName())
-					if err != nil {
-						framework.Logf("error checking pool %s: %v", p.GetName(), err)
-						return false, nil
+					updated := false
+					// Don't check for worker pool update if its paused
+					if upgradeWorkerPoolPaused && p.GetName() == "worker" {
```
smarterclayton, line 463: Hrm.  I would expect worker pools to be paused as part of the preconfiguration for the job.  This feels more like part of the step registry configuration for the multi-step configuration.  Then this code would only need to skip paused pools.

vrutkovs: Right, lets move this part to step-registry instead

vrutkovs: Reworked, PTAL

test/e2e/upgrade/upgrade.go
```diff
@@ -505,6 +505,15 @@ func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (bool
 		framework.Logf("error getting pool %s: %v", name, err)
 		return false, nil
 	}
+	paused, found, err := unstructured.NestedBool(pool.Object, "spec", "paused")
```
smarterclayton, line 508: I'm restructuring this slightly into #25922

smarterclayton:
/hold

see the linked PR, integrating it

PR #25906, opened 2021-02-18
test: SCC early failures (before suite) flake until fixed

smarterclayton:
Errors during test should hard fail (no tolerance). This change is
due to 44% of jobs failing in master on this, once we fix we can
revert.


wking:
[rhbz#1925698][1] is the bug whose fixing will allow us to revert this.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1925698

PR #25911, opened 2021-02-22
put the SCC test back

deads2k:
Looks like someone nerfed the SCC test while we had the early render in flight.  This puts it back so we can catch when things get worse again.


deads2k:
quota issues

/retest

PR #25912, opened 2021-02-22
test/e2e/upgrade/alert: Extend to testDuration, not just 1m

wking:
We're doing better in updates now, and want to ratchet down to bar critical-alert noise during updates.  The old 1m `alertPeriodCheckMinutes` landed with this test in 3b8cb3ca9b (#24786).  `DurationSinceStartInSeconds`, which I'm using now, landed in ace1345c00 (#25784).

I've also dropped some special-cased `alertname` filtering, because we don't want any critical alerts firing.  [Watchdog is `severity=none
`][1].  [AlertmanagerReceiversNotConfigured is `severity=warning`][2].  KubeAPILatencyHigh was dropped in openshift/cluster-monitoring-operator#898, [4.6][3] and was `severity=warning` anyway.

[1]: https://github.com/openshift/cluster-monitoring-operator/blob/776379a9616be1cbea49dd86086d8be8230370ce/assets/prometheus-k8s/rules.yaml#L2326-L2336
[2]: https://github.com/openshift/cluster-monitoring-operator/blob/776379a9616be1cbea49dd86086d8be8230370ce/assets/prometheus-k8s/rules.yaml#L924-L933
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=1846805#c13

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -21,9 +21,6 @@ const (
 	// Delay after upgrade is complete before checking for critical alerts
 	alertCheckSleepMinutes = 5
 	alertCheckSleep        = alertCheckSleepMinutes * time.Minute
-
-	// Previous period in which to check for critical alerts
-	alertPeriodCheckMinutes = 1
 )
 
 // UpgradeTest runs post-upgrade after alertCheckSleep delay and tests if any critical alerts are firing.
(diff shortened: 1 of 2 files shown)


smarterclayton:
https://github.com/openshift/origin/pull/25904

is intended to remove the ability to throw any alerts during upgrade

wking:
Between #25904 and #25923, everything I was trying to do here is covered.

/close

PR #25916, opened 2021-02-23
Revert "put the SCC test back"

smarterclayton:
Reverts openshift/origin#25911


smarterclayton:
The revert was wrong,

PR #25918, opened 2021-02-23
flake on operator state changes during e2e

deads2k:
If any operator changes available, degraded, or progressing on non-disruptive, non-upgrade jobs, then fail.  It's easy to apply elsewhere and easy to make fail, but let's see where we are.

### DIFF
--- a/cmd/openshift-tests/e2e.go
+++ b/cmd/openshift-tests/e2e.go
@@ -6,16 +6,16 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/origin/pkg/monitor"
+	"github.com/openshift/origin/pkg/synthetictests"
+	"github.com/openshift/origin/pkg/test/ginkgo"
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/disruption"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubectl/pkg/util/templates"
 
-	"github.com/openshift/origin/pkg/monitor"
-	"github.com/openshift/origin/pkg/test/ginkgo"
-
 	_ "github.com/openshift/origin/test/extended"
-	exutil "github.com/openshift/origin/test/extended/util"
 	_ "github.com/openshift/origin/test/extended/util/annotate/generated"
-	"github.com/openshift/origin/test/extended/util/disruption"
 )
 
 func isDisabled(name string) bool {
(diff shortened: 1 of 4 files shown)


cmd/openshift-tests/e2e.go
```diff
@@ -821,3 +823,65 @@ func getEventsByPod(events []*monitor.EventInterval) map[string][]*monitor.Event
...
+	}
+
+	// write a passing test to trigger detection of this issue as a flake. Doing this first to try to see how frequent the issue actually is
+	return []*ginkgo.JUnitTestCase{failure, success}
+}
+
+func testOperatorState(interestingCondition string, events []*monitor.EventInterval) []string {
+	failures := []string{}
+
+	clusterOperatorToEvents := getEventsByOperator(events)
```
wking, line 863: library-go-based operators will event on ClusterOperator changes, but I'm not sure non-library-go operators will event (consistently) on those or not.

deads2k: > library-go-based operators will event on ClusterOperator changes, but I'm not sure non-library-go operators will event (consistently) on those or not.

this wires to e2e monitor events, not real events.  Also, non-library-go operators could learn.... :)

pkg/monitor/operator.go
```diff
@@ -53,6 +78,12 @@ func startClusterOperatorMonitoring(ctx context.Context, m Recorder, client conf
 					if c.Type == configv1.OperatorDegraded && c.Status == configv1.ConditionTrue {
 						level = Error
 					}
+					if c.Type == configv1.OperatorAvailable && c.Status == configv1.ConditionFalse {
+						level = Error
+					}
+					if c.Type == configv1.OperatorProgressing && c.Status == configv1.ConditionTrue {
```
smarterclayton, line 84: I would prefer not to make this an error

deads2k: > I would prefer not to make this an error

Updated to warning for now.

deads2k:
/hold 

until we confirm we get the right test cases.  serial looks like it ought to trip on this

deads2k:
/hold cancel

The tests are now flaking, with good messages.

cmd/openshift-tests/e2e.go
smarterclayton: i do think we could add an OLM monitor addon - olm isn't as sophisticated at CO unfortuntaely.

pkg/monitor/operator.go
```diff
@@ -47,12 +79,18 @@ func startClusterOperatorMonitoring(ctx context.Context, m Recorder, client conf
 					case len(c.Message) > 0:
 						msg = fmt.Sprintf("condition/%s status/%s changed: %s", c.Type, c.Status, c.Message)
 					default:
-						msg = fmt.Sprintf("condition/%s status/%s changed", c.Type, c.Status)
+						msg = fmt.Sprintf("condition/%s status/%s changed: ", c.Type, c.Status)
```
smarterclayton, line 82: why?

deads2k: > why?

laziness in easy machine parsing above

cmd/openshift-tests/e2e.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/origin/pkg/synthetictests"
```
mfojtik, line 9: nit: order

cmd/openshift-tests/e2e.go
```diff
@@ -821,3 +825,113 @@ func getEventsByPod(events []*monitor.EventInterval) map[string][]*monitor.Event
...
+
+	knownOperators := allOperators(events)
+	eventsByOperator := getEventsByOperator(events)
+	for _, condition := range []string{"Available", "Degraded", "Progressing"} {
+		for _, operatorName := range knownOperators.List() {
+			bzComponent := synthetictests.GetBugzillaComponentForOperator(operatorName)
+			if bzComponent == "Unknown" {
+				bzComponent = operatorName
+			}
+			testName := fmt.Sprintf("[bz-%v] clusteroperator/%v should not change condition/%v", bzComponent, operatorName, condition)
```
mfojtik, line 840: would it help to clarify this further like "... should not change condition after installation is complete" ?

PR #25920, opened 2021-02-23
test: Prohibit cluster operators going degraded or unavailable

smarterclayton:
In upgrade or normal e2e testing, operators may not go unavailable
or degraded. Enforce that invariant in CI.


smarterclayton:
/hold

david had cleaner code

smarterclayton:
replaced by https://github.com/openshift/origin/pull/25918

PR #25921, opened 2021-02-23
monitor: Improve cluster operator transition message

smarterclayton:
Make it consistent with node after improvements there.

### DIFF
--- a/pkg/monitor/operator.go
+++ b/pkg/monitor/operator.go
@@ -34,26 +34,26 @@ func startClusterOperatorMonitoring(ctx context.Context, m Recorder, client conf
 		func(co, oldCO *configv1.ClusterOperator) []Condition {
 			var conditions []Condition
 			for i := range co.Status.Conditions {
-				s := &co.Status.Conditions[i]
-				previous := findOperatorStatusCondition(oldCO.Status.Conditions, s.Type)
+				c := &co.Status.Conditions[i]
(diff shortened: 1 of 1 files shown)


pkg/monitor/operator.go
```diff
@@ -34,26 +34,26 @@ func startClusterOperatorMonitoring(ctx context.Context, m Recorder, client conf
...
 					var msg string
 					switch {
-					case len(s.Reason) > 0 && len(s.Message) > 0:
-						msg = fmt.Sprintf("changed %s to %s: %s: %s", s.Type, s.Status, s.Reason, s.Message)
-					case len(s.Message) > 0:
-						msg = fmt.Sprintf("changed %s to %s: %s", s.Type, s.Status, s.Message)
+					case len(c.Reason) > 0 && len(c.Message) > 0:
+						msg = fmt.Sprintf("condition/%s status/%s reason/%s changed: %s", c.Type, c.Status, c.Reason, c.Message)
+					case len(c.Message) > 0:
+						msg = fmt.Sprintf("condition/%s status/%s changed: %s", c.Type, c.Status, c.Message)
```
wking, line 48: Can we make this `reason/unset-cmon-operators-you-need-to-set-a-reason`? ;)

smarterclayton: Do we require reason?

you can find this with regex as written

PR #25922, opened 2021-02-23
test: Detect when the master pool is still updating after upgrade

smarterclayton:
The MCO is required to roll out the master config pool prior to reporting level, and thus CVO should never reach Available at a
new version without the master pool being updated. Add an extra check to the pool rollout - if we see the master pool with an
Updating condition flag it and fail the upgrade job. Also handle paused pools in the wait loop (a paused pool is effectively "we

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/988/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn-upgrade/1364259570968432640 had a possible case where this happened, tighten the test.

```
Feb 23 18:43:19.284: INFO: cluster upgrade is Progressing: Working towards 4.8.0-0.ci.test-2021-02-23-170859-ci-op-m36h5d5b: 666 of 669 done (99% complete)
Feb 23 18:43:29.285: INFO: Completed upgrade to registry.build01.ci.openshift.org/ci-op-m36h5d5b/release@sha256:7884e2856b8e4f2a71dbcaf8333adc36ac7805315726b31d7c075b961a163a94
Feb 23 18:43:29.313: INFO: Waiting on pools to be upgraded
Feb 23 18:43:29.351: INFO: Pool master is still reporting (Updated: false, Updating: true, Degraded: false)
```

^ no, bad

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -414,6 +414,7 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 		return err
 	}
 
+	var errMasterUpdating error
 	if err := disruption.RecordJUnit(
 		f,
 		"[sig-mco] Machine config pools complete upgrade",
@@ -432,12 +433,14 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 				}
 				allUpdated := true
 				for _, p := range pools.Items {
-					updated, err := IsPoolUpdated(mcps, p.GetName())
-					if err != nil {
-						framework.Logf("error checking pool %s: %v", p.GetName(), err)
-						return false, nil
-					}
+					updated, requiresUpdate := IsPoolUpdated(mcps, p.GetName())
 					allUpdated = allUpdated && updated
+
+					// Invariant: when CVO reaches level, MCO is required to have rolled out control plane updates
+					if p.GetName() == "master" && requiresUpdate && errMasterUpdating == nil {
+						errMasterUpdating = fmt.Errorf("the %q pool should be updated before the CVO reports available at the new version", p.GetName())
+						framework.Logf("Invariant violation detected: %s", errMasterUpdating)
+					}
 				}
 				return allUpdated, nil
 			}); err != nil {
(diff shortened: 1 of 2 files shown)


wking (approved):
/lgtm
/hold

Holding so we can look at the presubmit jobs and confirm this is working.

wking:
[images][1]:

```
test/extended/dr/common.go:89:3: cannot use bool value as type error in return argument:
	bool does not implement error (missing Error method)
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk:14: build] Error 2
```

not clear to me how that's related to this PR, but we certainly don't want to land things that fail compilation.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25922/pull-ci-openshift-origin-master-images/1364341889125322752#1:build-log.txt%3A88

smarterclayton:
It’s not, which is concerning 

/retest

test/e2e/upgrade/upgrade.go
```diff
@@ -499,33 +507,39 @@ func recordClusterEvent(client kubernetes.Interface, uid, action, reason, note s
 }
 
 // TODO(runcom): drop this when MCO types are in openshift/api and we can use the typed client directly
-func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (bool, error) {
+func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (poolUpToDate bool, poolIsUpdating bool) {
```
wking, line 510:
```console
$ git --no-pager grep IsPoolUpdated origin/master
origin/master:test/e2e/upgrade/upgrade.go:                                      updated, err := IsPoolUpdated(mcps, p.GetName())
origin/master:test/e2e/upgrade/upgrade.go:func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (bool, error) {
origin/master:test/extended/dr/common.go:               return upgrade.IsPoolUpdated(mcps, "master")
```

So this is causing the compilation error.

smarterclayton:
The flake is not being cooperative, I’m going to merge and then keep watching for it

PR #25923, opened 2021-02-24
Bug 1932618: Don't allow alerts to fire during a test run

smarterclayton:
The alert test has not been working and a PR was able to merge that was firing alerts during the test run. https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-baremetal-operator/110/pull-ci-openshift-cluster-baremetal-operator-master-e2e-agnostic/1364398641615212544 shows the two firing alerts being bypassed by the query in the `shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured` late test

Fix the test logic based on:

1. Remove the filter on KubeAPILatencyHigh - no longer needed in 4.6+
2. Remove the filter on KubePodCrashLooping - no longer needed in 4.6+
3. Sort the metric by number of seconds the alert was firing

The bug that allowed the merge was the filter for KubePodCrashLooping,
to omit a series we must use `unless X`. If we reintroduce a filter we
should use that syntax.

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -65,10 +65,9 @@ var _ = g.Describe("[sig-instrumentation][Late] Alerts", func() {
 		testDuration := exutil.DurationSinceStartInSeconds().String()
 
 		tests := map[string]bool{
-			// Checking Watchdog alert state is done in "should have a Watchdog alert in firing state".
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm

ravisantoshgudimetla (approved):
/lgtm

PR #25927, opened 2021-02-25
Add oom detection test

ravisantoshgudimetla:
Test to identify if there are any OOM events in the cluster.

### DIFF
--- a/test/extended/node/oom_detection.go
+++ b/test/extended/node/oom_detection.go
@@ -1,0 +1,78 @@
+package node
+
+import (
+	"context"
+	"fmt"
+	o "github.com/onsi/gomega"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	exutil "github.com/openshift/origin/test/extended/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO: expand this test to SSH into every node in the cluster and gather kubelet logs in case kubelet doesn't send
+// 		 events to the api-server or `oc adm node-logs`
+var _ = g.Describe("[sig-node][Late]", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLIWithoutNamespace("no-oom-kills")
+
+	g.It("should not have pods that are OOM Killed", func() {
+		kubeClient := oc.AdminKubeClient()
+
+		events, err := kubeClient.CoreV1().Events("").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			g.Fail(fmt.Sprintf("Unexpected error: %v", err))
+		}
+		oomEventNodeList := []string{}
+		suiteStartTime := exutil.SuiteStartTime()
+		errorMsg := "System OOM encountered"
+		for _, event := range events.Items {
+			if event.LastTimestamp.Time.Before(suiteStartTime) {
+				continue
+			}
+			if !strings.Contains(event.Message, errorMsg) {
+				continue
+			}
+			oomEventNodeList = append(oomEventNodeList, event.InvolvedObject.Name)
+		}
+		failMessage := fmt.Sprintf("%v OOM events occured in the following nodes\n", oomEventNodeList)
+		if len(oomEventNodeList) > 0 {
+			g.Fail(failMessage)
+			return
+		}
+		// let us see if need to mark this as a flake
+		// result.Flakef(strings.Join(oomEncounteredEvents, "\n"))
+	})
+})
+
+var _ = g.Describe("[sig-node][Late]", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLIWithoutNamespace("no-oom-kills within worker node journal logs")
+
+	g.It("OOM Killer shouldn't have been invoked in kubelet logs", func() {
+
+		nodeList, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		oomJournalNodeList := []string{}
+		testDuration := exutil.DurationSinceStartInSeconds()
+		for _, node := range nodeList.Items {
+			out, err := oc.Run("adm", "node-logs").Args(node.Name, "--path=journal",
+				"--case-sensitive=false", "--grep=invoked oom-killer", "--since=-"+string(testDuration)).Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if !strings.Contains(strings.ToLower(out), "invoked oom-killer") {
+				continue
+			}
+			oomJournalNodeList = append(oomJournalNodeList, node.Name)
+		}
+		failMessage := fmt.Sprintf("%v OOM events occured in the following nodes from kubelet log\n",
+			oomJournalNodeList)
+		if len(oomJournalNodeList) > 0 {
+			g.Fail(failMessage)
+		}
+		// TOOD: Figure out if we need to do a comparison when the events are not stuck on the node if there is a kernel
+		//		 bug
+	})
+})


ravisantoshgudimetla:
cc @smarterclayton @rphillips @mrunalp @wking

test/extended/node/oom_detection.go
```diff
@@ -0,0 +1,45 @@
...
+			g.Fail(fmt.Sprintf("Unexpected error: %v", err))
+		}
+
+		oomEncounteredEvents := []string{}
+		errorMsg := "System OOM encountered"
+		for _, event := range events.Items {
+			if !strings.Contains(event.Message, errorMsg) {
+				continue
+			}
+			oomEncounteredEvent := fmt.Sprintf("OOM encountered within node %v", event.InvolvedObject.Name)
```
wking, line 34: Do the events say what got OOMed?  That would make it easier to route to the team in charge of the greedy component (for cases where OOM picked the right greedy component).

test/extended/node/oom_detection.go
```diff
@@ -0,0 +1,45 @@
...
+			oomEncounteredEvents = append(oomEncounteredEvents, oomEncounteredEvent)
+		}
+		failMessage := fmt.Sprintf("%d OOM events occured", oomEncounteredEvents)
```
wking, line 37: The count is nice, but having a list of OOMed components here would make routing the resulting bugs easier.

test/extended/node/oom_detection.go
ravisantoshgudimetla: I thought about it but the message only talks about the [victim process](https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/pkg/kubelet/oom/oom_watcher_linux.go#L79). Not sure, if it be beneficial

test/extended/node/oom_detection.go
ravisantoshgudimetla: I think identifying the component which caused OOMKiller(the greedy one) invocation is benefical but AFAIK we can get that info from journal logs.

test/extended/node/oom_detection.go
mrunalp: The process that ends up being the cause of the  OOM Killer invocation isn't always at fault. Any memory allocation by any process could cause the oom killer to be invoked.

nstielau:
I wouldn't block this, but almost tempted to, until we could also measure the same metric across the fleet in aggregate.  If we can solve this problem now, we can use that pattern to get hard-failures in CI and soft-failures in aggregate fleet metrics.  I think that would significantly improve the signal we get.

The two ways ths would be possible would be:
* Write an Alert for this, and check for the firing of the alert rather than the symptom itself
* Collect telemetry for this, and check against that data directly.

I'm sure there are drawbacks to both of these, but I think being able to go from a single test to a single cluster to a subset of clusters to the whole fleet looking at the same signal is powerful compared to one more test.

ravisantoshgudimetla:
> Write an Alert for this, and check for the firing of the alert rather than the symptom itself

I think we already have an alert for these situations. Something like:

`
{alertname="KubeNodeNotReady",severity="warning"} 
`
I think the main problem here is` from the alert how to identify if it's a kernel bug or resource pressure induced by a workload` which is orthogonal to the problem however I agree that having this info captured in a metric and storing it might be useful.

test/extended/node/oom_detection.go
```diff
@@ -0,0 +1,74 @@
...
+	oc := exutil.NewCLIWithoutNamespace("no-oom-kills within worker node journal logs")
+
+	g.It("OOM Killer shouldn't have been invoked in kubelet logs", func() {
+
+		nodeList, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		oomJournalNodeList := []string{}
+		for _, node := range nodeList.Items {
+			// Choosing 150 min as the maximum upgrade duration has been set to 150m in upgrade.go
+			out, err := oc.Run("adm", "node-logs").Args(node.Name, "--unit=kubelet",
```
mrunalp, line 58: These would be in the kernel logs. So either `dmesg` or `journalctl -k` equivalent.

mrunalp: Or get the full journal and grep should be able to find it.

ravisantoshgudimetla: I saw entries like in the kubelet log:

```
Jan 06 23:27:40.051244 ip-10-0-225-225 kernel: prometheus invoked oom-killer: gfp_mask=0x6200ca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=-997
```

Do we need more info that we need to get from kernel logs?

mrunalp: That looks like a kernel log entry to me.

ravisantoshgudimetla: I see, let me remove the `--unit=kubelet` and use `--path=journal`, perhaps we can get entire journal log.


This is what help suggests.
```
--path='journal': Retrieve the specified path within the node's /var/logs/ folder. The 'journal' value will allow
querying the journal on supported operating systems.
```
So, I believe we should be good.

test/extended/node/oom_detection.go
```diff
@@ -0,0 +1,74 @@
...
+
+	g.It("should not have pods that are OOM Killed", func() {
+		kubeClient := oc.AdminKubeClient()
+
+		events, err := kubeClient.CoreV1().Events("").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			g.Fail(fmt.Sprintf("Unexpected error: %v", err))
+		}
+		oomEventNodeList := []string{}
+		errorMsg := "System OOM encountered"
```
smarterclayton, line 29: You need to filter by suite start (see other event types where we filter those out).  We will eventually be running multiple suites in order - we might run disruptive (which causes OOMs) before we run e2e-conformanec (to verify that everything works after ooms).  If you look back before the suite start you may cause failures.

test/extended/node/oom_detection.go
```diff
@@ -0,0 +1,74 @@
...
+	oc := exutil.NewCLIWithoutNamespace("no-oom-kills within worker node journal logs")
+
+	g.It("OOM Killer shouldn't have been invoked in kubelet logs", func() {
+
+		nodeList, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		oomJournalNodeList := []string{}
+		for _, node := range nodeList.Items {
+			// Choosing 150 min as the maximum upgrade duration has been set to 150m in upgrade.go
+			out, err := oc.Run("adm", "node-logs").Args(node.Name, "--path=journal",
```
smarterclayton, line 58: You can get the exact test duration from test util - `DurationSinceStartInSeconds()`

test/extended/node/oom_detection.go
```diff
@@ -0,0 +1,74 @@
...
+				"--case-sensitive=false", "--grep=invoked oom-killer", "--since=-150m").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if !strings.Contains(out, "invoked oom-killer") {
```
smarterclayton, line 61: Were you finding lines returned by --grep that did contain invoked oom killer?  Also, you're doing a case insensitive search above, and then this is case sensitive - why?

test/extended/node/oom_detection.go
ravisantoshgudimetla: Ok makes sense. Let me look at that.

test/extended/node/oom_detection.go
ravisantoshgudimetla: > Were you finding lines returned by --grep that did contain invoked oom killer?

If even it doesn't contain `invoked oom killer`, it was still returning output like:

```
-- Logs begin at Mon 2021-03-01 14:50:39 UTC, end at Mon 2021-03-01 17:10:35 UTC. --
-- No entries --
-- Logs begin at Mon 2021-03-01 14:50:39 UTC, end at Mon 2021-03-01 17:10:35 UTC. --
-- No entries --
```
So, I added the `strings.Contains` check, I think the case sensitivity can be ignored. Will remove it in the next commit.

PR #25928, opened 2021-02-25
test: Fail if a daemonset has a different upgrade strategy than default

smarterclayton:
Now that we have fixed all issues related to network upgrades by moving OVS to the node, we can bulk upgrade daemonsets on the nodes.

Daemonsets shipped with the platform must be able to upgrade without disruption to workloads. Daemonsets that are in the data path must gracefully shutdown and redirect workload traffic, or mitigate outage by holding requests briefly (very briefly!). daemonsets that are part of control plane (anything that changes state) must already be able to be upgraded without disruption by having any components that call them retry when unavailable up to a specific duration. Therefore all workloads must individually be graceful, and so a core daemonset can upgrade up to 10% of pods at a time. In the future, we may allow this percentage to be tuned via a global config, and this test would enforce that.

### DIFF
--- a/test/extended/operators/daemon_set.go
+++ b/test/extended/operators/daemon_set.go
@@ -1,0 +1,107 @@
+package operators
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	"github.com/openshift/origin/pkg/test/ginkgo/result"
+	exutil "github.com/openshift/origin/test/extended/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = g.Describe("[sig-arch] Managed cluster", func() {
+	oc := exutil.NewCLIWithoutNamespace("operator-daemonsets")
+
+	// Daemonsets shipped with the platform must be able to upgrade without disruption to workloads.
+	// Daemonsets that are in the data path must gracefully shutdown and redirect workload traffic, or
+	// mitigate outage by holding requests briefly (very briefly!). Daemonsets that are part of control
+	// plane (anything that changes state) must already be able to be upgraded without disruption by
+	// having any components that call them retry when unavailable up to a specific duration. Therefore
+	// all workloads must individually be graceful, and so a core daemonset can upgrade up to 10% or 33%
+	// of pods at a time. In the future, we may allow this percentage to be tuned via a global config,
+	// and this test would enforce that.
+	//
+	// Use 33% maxUnavailable if you are a workload that has no impact on other workload. This ensures
+	// that if there is a bug in the newly rolled out workload, 2/3 of instances remain working.
+	// Workloads in this category include the spot instance termination signal observer which listens
+	// for when the cloud signals a node that it will be shutdown in 30s. At worst, only 1/3 of machines
+	// would be impacted by a bug and at best the new code would roll out that much faster in very large
+	// spot instance machine sets.
+	//
+	// Use 10% maxUnavailable in all other cases, most especially if you have ANY impact on user
+	// workloads. This limits the additional load placed on the cluster to a more reasonable degree
+	// during an upgrade as new pods start and then establish connections.
+	//
+	// Currently only applies to daemonsets that don't explicitly target the control plane.
+	g.It("should only include cluster daemonsets that have maxUnavailable update of 10 or 33 percent", func() {
+		// iterate over the references to find valid images
+		daemonSets, err := oc.KubeFramework().ClientSet.AppsV1().DaemonSets("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list daemonsets: %v", err)
+		}
+
+		// daemonsets that have an explicit exception, every entry here must have a bug associated
+		// e.g. "openshift-x/foo": "link to bug"
+		bugsAgainstBrokenDaemonSets := map[string]string{
+			"openshift-multus/multus":                 "https://bugzilla.redhat.com/show_bug.cgi?id=1933159",
+			"openshift-multus/network-metrics-daemon": "https://bugzilla.redhat.com/show_bug.cgi?id=1933159",
+		}
+
+		var debug []string
+		var knownBroken []string
+		var invalidDaemonSets []string
+		for _, ds := range daemonSets.Items {
+			if !strings.HasPrefix(ds.Namespace, "openshift-") {
+				continue
+			}
+			if ds.Spec.Selector != nil {
+				selector := labels.SelectorFromSet(ds.Spec.Template.Spec.NodeSelector)
+				if !selector.Empty() && selector.Matches(labels.Set(map[string]string{"node-role.kubernetes.io/master": ""})) {
+					continue
+				}
+			}
+			key := fmt.Sprintf("%s/%s", ds.Namespace, ds.Name)
+			switch {
+			case ds.Spec.UpdateStrategy.RollingUpdate == nil,
+				ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable == nil:
+				violation := fmt.Sprintf("expected daemonset %s to have a maxUnavailable strategy of 10%% or 33%%", key)
+				if bug, ok := bugsAgainstBrokenDaemonSets[key]; ok {
+					knownBroken = append(knownBroken, fmt.Sprintf("%s (bug %s)", violation, bug))
+				} else {
+					invalidDaemonSets = append(invalidDaemonSets, violation)
+				}
+				debug = append(debug, violation)
+			case ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.StrVal != "10%" &&
+				ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.StrVal != "33%":
+				violation := fmt.Sprintf("expected daemonset %s to have maxUnavailable 10%% or 33%% (see comment) instead of %s", key, ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.String())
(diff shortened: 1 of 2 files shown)


test/extended/operators/daemon_set.go
```diff
@@ -0,0 +1,73 @@
...
+					continue
+				}
+			}
+			if bug, ok := skipDaemonSets[fmt.Sprintf("%s/%s", ds.Namespace, ds.Name)]; ok {
+				e2e.Logf("Skipping daemonset %s/%s from consideration: %s", ds.Namespace, ds.Name, bug)
+				continue
+			}
+			switch {
+			case ds.Spec.UpdateStrategy.RollingUpdate == nil,
+				ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable == nil:
```
wking, line 63: None of these today:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-4.8/1364411741282242560/artifacts/e2e-gcp/daemonsets.json | jq -r '.items[] | .metadata.namespace + " " + .metadata.name + " " + (.spec.template.spec.nodeSelector | tostring) + " " + (.spec.updateStrategy | tostring)' | grep -v RollingUpdate
...no hits...
```

Should we make them illegal unless folks talk to you about softening the CI guard?

test/extended/operators/daemon_set.go
```diff
@@ -0,0 +1,73 @@
...
+			case ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.StrVal != "10%":
```
wking, line 64: @michaelgugino doesn't want to be bothered about tuning this for `machine-api-termination-handler`, so either this PR or any subsequent PR that bumps this check needs to allow 100%.

michaelgugino: MAO only has one daemonset, for spot instances watching for termination signals from the cloud provider.  In our case, we could be 100%, it makes no difference in impact to the end user or anything else.  Probably other components could be handled similarly, such as the node tuning bits.  So if we need to change this in the future, I'd prefer to just set 100% so we're never the slow point.

test/extended/operators/daemon_set.go
smarterclayton: oh, good point.  thanks, i screwed that up in my head

"daemonsets must have a rolling update 10% strategy"

test/extended/operators/daemon_set.go
smarterclayton: ok, make it 100% and then i'll add the special case here (should be easy).  thanks for quick response.

smarterclayton: Actually, the one question about 100% is - do you think there is any value in partial degradation (i.e. if you totally botch it in an update, do you want to be all broken or max X percent broken)?

I think there's an analogy here to zone failures - after 33% percent or whatever we just say "we're not doing anything".  Should we make 33% our max so we can say "at least if we miss something 2/3 of the cluster should be working"?

michaelgugino: 33% seems like a fine upper limit to me, and is probably acceptable in the general case for many components as well.

smarterclayton: Ok, I'll list 33% as an acceptable number

wking: 33% vs. 10% doesn't seem like it would be all that significant.  Do we need to give components that flexibility?

test/extended/operators/daemon_set.go
wking: This thread can be marked resolved, now that this is an error case.

test/extended/operators/daemon_set.go
michaelgugino: Any setting other than 'read setting from elsewhere' works for me.  10%, 33%, 100%, doesn't matter, as long as I don't need to have my operator manage that number back and forth.

test/extended/util/annotate/generated/zz_generated.annotations.go
```diff
@@ -847,6 +847,8 @@ var annotations = map[string]string{
 
 	"[Top Level] [sig-arch] Managed cluster should have operators on the cluster version": "have operators on the cluster version [Suite:openshift/conformance/parallel]",
 
+	"[Top Level] [sig-arch] Managed cluster should only include cluster daemonsets that have maxUnavailable update of 10 percent": "should only include cluster daemonsets that have maxUnavailable update of 10 percent [Suite:openshift/conformance/parallel]",
```
wking, line 850: [verify][1]:

```
FAILURE after 0.032s: hack/verify-generated.sh:14: executing 'git diff --quiet /go/src/github.com/openshift/origin/test/extended/util/annotate/generated/' expecting success: the command returned the wrong error code
There was no error output from the command.
```

Not very helpful.  Hopefully a fluke.  But [GCP][2]:

```
panic: unable to find test [Top Level] [sig-arch] Managed cluster should only include cluster daemonsets that have maxUnavailable update of 10 or 33 percent
```

seems more serious.  I guess that's this line needing a regen, although [I'm also fine dropping the 33% back out][3].

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25928/pull-ci-openshift-origin-master-verify/1367128290422165504#1:build-log.txt%3A46
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25928/pull-ci-openshift-origin-master-e2e-gcp/1367128286907338752#1:build-log.txt%3A98
[3]: https://github.com/openshift/origin/pull/25928#discussion_r586725460

test/extended/operators/daemon_set.go
```diff
@@ -0,0 +1,95 @@
...
+			e2e.Failf("unable to list daemonsets: %v", err)
+		}
+
+		// daemonsets that have an explicit exception, every entry here must have a bug associated
+		// e.g. "openshift-sdn/foo": "link to bug"
+		skipDaemonSets := map[string]string{
+			"openshift-multus/multus":                            "https://bugzilla.redhat.com/show_bug.cgi?id=1933159",
+			"openshift-sdn/sdn":                                  "https://bugzilla.redhat.com/show_bug.cgi?id=1933173",
+			"openshift-sdn/ovs":                                  "https://bugzilla.redhat.com/show_bug.cgi?id=1933174",
+			"openshift-cluster-csi-drivers":                      "https://bugzilla.redhat.com/show_bug.cgi?id=1933184",
```
wking, line 53: You can [drop this entry now][1] that the bug is MODIFIED.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1933184

test/extended/operators/daemon_set.go
```diff
@@ -0,0 +1,95 @@
...
+			"openshift-image-registry/node-ca":                   "https://bugzilla.redhat.com/show_bug.cgi?id=1933180",
+			"openshift-network-diagnostics/network-check-target": "https://bugzilla.redhat.com/show_bug.cgi?id=1933179",
```
wking, line 55: You can [drop this entry now][1] that the bug is VERIFIED.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1933179#c2

test/extended/operators/daemon_set.go
```diff
@@ -0,0 +1,95 @@
...
+		daemonSets, err := oc.KubeFramework().ClientSet.AppsV1().DaemonSets("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list daemonsets: %v", err)
+		}
+
+		// daemonsets that have an explicit exception, every entry here must have a bug associated
+		// e.g. "openshift-sdn/foo": "link to bug"
+		skipDaemonSets := map[string]string{
+			"openshift-multus/multus":                            "https://bugzilla.redhat.com/show_bug.cgi?id=1933159",
+			"openshift-sdn/sdn":                                  "https://bugzilla.redhat.com/show_bug.cgi?id=1933173",
```
wking, line 51: You can [drop this entry now][1] that the bug is MODIFIED.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1933173#c3

test/extended/operators/daemon_set.go
```diff
@@ -0,0 +1,92 @@
...
+		daemonSets, err := oc.KubeFramework().ClientSet.AppsV1().DaemonSets("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list daemonsets: %v", err)
+		}
+
+		// daemonsets that have an explicit exception, every entry here must have a bug associated
+		// e.g. "openshift-sdn/foo": "link to bug"
+		skipDaemonSets := map[string]string{
+			"openshift-multus/multus":          "https://bugzilla.redhat.com/show_bug.cgi?id=1933159",
+			"openshift-sdn/ovs":                "https://bugzilla.redhat.com/show_bug.cgi?id=1933174",
```
wking, line 51: We can drop this, because [the bug is MODIFIED][1].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1933174#c3

test/extended/operators/daemon_set.go
```diff
@@ -0,0 +1,92 @@
...
+		// iterate over the references to find valid images
+		daemonSets, err := oc.KubeFramework().ClientSet.AppsV1().DaemonSets("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list daemonsets: %v", err)
+		}
+
+		// daemonsets that have an explicit exception, every entry here must have a bug associated
+		// e.g. "openshift-sdn/foo": "link to bug"
+		skipDaemonSets := map[string]string{
+			"openshift-multus/multus":          "https://bugzilla.redhat.com/show_bug.cgi?id=1933159",
```
wking, line 50: Need to be extended to cover `openshift-multus/network-metrics-daemon` under the same bug.

smarterclayton: i fixed that today

PR #25931, opened 2021-03-02
Bug 1957295: All platform pods should have system-* or openshift-user-critical priority class

ravisantoshgudimetla:
In order to avoid conditions like https://github.com/openshift/cluster-network-operator/pull/994

cc @smarterclayton @soltysh @damemi @ingvagabund

### DIFF
--- a/test/extended/pods/priorityclasses.go
+++ b/test/extended/pods/priorityclasses.go
@@ -1,0 +1,96 @@
+package pods
+
+import (
+	"context"
+	"fmt"
+	"github.com/openshift/origin/pkg/test/ginkgo/result"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = Describe("[sig-arch] Managed cluster should", func() {
+	oc := exutil.NewCLIWithoutNamespace("pod")
+
+	It("ensure platform components have system-* priority class associated", func() {
+		// iterate over the references to find valid images
+		pods, err := oc.KubeFramework().ClientSet.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{})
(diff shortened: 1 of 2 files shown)


ravisantoshgudimetla:
cc @wking

smarterclayton:
/hold

we actually want to move these out of system-* in most cases.  Also, we need to update the guidance in enhancements CONVENTIONS.md doc as well.  Let’s sync tomorrow (but this will absolutely be needed).

test/extended/pods/priorityclasses.go
```diff
@@ -0,0 +1,52 @@
...
+	It("ensure platform components have system-* priority class associated", func() {
+		// iterate over the references to find valid images
+		pods, err := oc.KubeFramework().ClientSet.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list pods: %v", err)
+		}
+
+		// list of pods that use images not in the release payload
+		invalidPodPriority := sets.NewString()
+		// a pod in a namespace that begins with kube-* or openshift-* must come from our release payload
```
aravindhp, line 27: This needn't be true for `openshift-*` namespaces as some OLM managed operators user this also.

ravisantoshgudimetla: If you're just referring to the comment, I don't mind removing it. However please note that this test suite when run in origin wouldn't include those OLM managed components unless explicitly deployed.

aravindhp: Yes I am just referring to the comment. You can reword it to clarify given you are suggesting in your email that OLM managed operators can also pull in this test.

markmc:
> Also, we need to update the guidance in enhancements CONVENTIONS.md doc as well.

Yes. Apparently the guidance now is:

```
* If it is fine for your operator/operand to be preempted by user workload specify and OOMKilled  use openshift-user-critical priority class
* If you want your operator/operand not to be preempted by user-workload but still be OOMKilled use system-cluster-critical priority class
* If you want operator/operand not be preempted by user-workload and OOMKilled last use system-node-critical priority class
```

soltysh (approved):
Consistent with https://github.com/openshift/enhancements/pull/763
/hold cancel
/lgtm
/approve

PR #25932, opened 2021-03-02
test: Graceful termination invariant should separate pre-suite events

smarterclayton:
Termination of the kube-apiserver outside of the suite that is currently running should be a flake, and when the test infra is told to ignore events before a specific time it should do so. Add the test into the upgrade invariants since this would exclude the test from checking upgrades otherwise.

Mitigates impact of pre-install bug in https://bugzilla.redhat.com/show_bug.cgi?id=1928946 but does not fix the issue.


rphillips:
lgtm but will defer to an api approver.

smarterclayton:
I’m suspicious nothing broke...

/test e2e-gcp

smarterclayton:
Split is correct - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25932/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1367163593178484737 caught during test.

PR #25934, opened 2021-03-03
test: Nodes that are deleted should not fire the unready alert

smarterclayton:
The serial test creates a new instance and deletes an old one
(gracefully). This should only cause the "nodes not ready" test
to fail if the node object doesn't get deleted. The query is shifted
1m to the left, so that if at the current time the node doesn't exist
then the ready condition is ignored 1m earlier. This is because
promethues queries can't offset in the future.

This leaves a slight gap at the end of a test (scrape_interval+
1m) where a node could go unready and the test wouldn't catch it.
But in practice it was already (scrape_interval) and the test doesn't
run at the exact end of the test anyway.

Fixes 10% of serial flakes. If 1m isn't long enough we can wait
longer.

### DIFF
--- a/test/extended/machines/cluster.go
+++ b/test/extended/machines/cluster.go
@@ -83,8 +83,12 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
 		testDuration := exutil.DurationSinceStartInSeconds().String()
 
 		tests := map[string]bool{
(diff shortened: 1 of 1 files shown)


wking:
/retitle Bug 2033615: test: Nodes that are deleted should not fire the unready alert

PR #25936, opened 2021-03-04
Allow several alerts in single replica topology

omertuc:
EDIT:
This PR used to be just for [MON-1522](https://issues.redhat.com/browse/MON-1522), but now it's also for  https://bugzilla.redhat.com/show_bug.cgi?id=1945017 and  [API-1136](https://issues.redhat.com/browse/API-1136)

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -16,6 +16,7 @@ import (
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 
+	configv1 "github.com/openshift/api/config/v1"
 	v1 "k8s.io/api/core/v1"
 
 	kapierrs "k8s.io/apimachinery/pkg/api/errors"
@@ -58,6 +59,10 @@ var _ = g.Describe("[sig-instrumentation][Late] Alerts", func() {
 		if len(os.Getenv("TEST_UNSUPPORTED_ALLOW_VERSION_SKEW")) > 0 {
 			e2eskipper.Skipf("Test is disabled to allow cluster components to have different versions, and skewed versions trigger multiple other alerts")
 		}
+
+		infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
 		ns := oc.SetupNamespace()
 		execPod := exutil.CreateExecPodOrFail(oc.AdminKubeClient(), ns, "execpod")
 		defer func() {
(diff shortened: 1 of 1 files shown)


romfreiman:
/lgtm
Can u please connect it to the jira ticket as a workaround?

romfreiman:
@smarterclayton wdyt? Temporary till it will be fixed on the product level. Otherwise we cannot enable sno ci

omertuc:
> Can u please connect it to the jira ticket as a workaround?

That's done automatically when the issue is referenced, but commented anyway

test/extended/prometheus/prometheus.go
```diff
@@ -396,11 +406,20 @@ var _ = g.Describe("[sig-instrumentation] Prometheus", func() {
 				oc.AdminKubeClient().CoreV1().Pods(ns).Delete(context.Background(), execPod.Name, *metav1.NewDeleteOptions(1))
 			}()
 
+			// Checking Watchdog alert state is done in "should have a Watchdog alert in firing state".
+			allowedAlerts := []string{"Watchdog", "AlertmanagerReceiversNotConfigured", "PrometheusRemoteWriteDesiredShards"}
+
+			infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if infra.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
```
smarterclayton, line 414: move these up to the top and give them unique names “allowedAlertsDuringX”, don’t reuse the same variable

omertuc: By top do you mean the top of the `It` block?

Also not sure how I can avoid reusing the same variable without duplicating the Prometheus query's `fmt.Sprintf` - it has to use some variable for `strings.Join(allowedAlerts, "|")`

smarterclayton: Nm, I thought you were fixing the upgrade job which has both of these inline.

smarterclayton: Move the infra config lookup up above allowedAlerts, then init allowedAlerts, then do the if check.  Roughly keep infra lookup high and the init of allowedAlerts and the conditional close to each other.

omertuc: Got it, done

/hold until I verify this works (takes a while)

omertuc: /unhold

omertuc:
/test e2e-aws-single-node

this time with m5d.4xlarge

omertuc:
These alerts have been fixed by now, PR no longer relevant

PR #25939, opened 2021-03-05
test: Node unready should ignore the ns/ in event locator

smarterclayton:
Events from the upgrade job include the namespace (probably because
I'm holding event sending wrong on cluster scoped events).

For now, simply look at the end of the locator rather than the
beginning.


wking (approved):
Can't hurt :)

/lgtm

PR #25941, opened 2021-03-05
Bug 1935707: [release-4.7] test: Detect when the master pool is still updating after upgrade 

vrutkovs:
The MCO is required to roll out the master config pool prior to
reporting level, and thus CVO should never reach Available at a
new version without the master pool being updated. Add an extra
check to the pool rollout - if we see the master pool with an
Updating condition flag it and fail the upgrade job. Also handle
paused pools in the wait loop (a paused pool is effectively "we

Cherr-pick of #25922 on release-4.7

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -404,6 +404,7 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 		return err
 	}
 
+	var errMasterUpdating error
 	if err := disruption.RecordJUnit(
 		f,
 		"[sig-mco] Machine config pools complete upgrade",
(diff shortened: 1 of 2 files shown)


vrutkovs:
/hold

-proxy and -gcp-disruptive permafailing

vrutkovs:
/hold cancel

These tests are no longer required.
/refresh

test/e2e/upgrade/upgrade.go
```diff
@@ -495,9 +509,13 @@ func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (bool
...
+		framework.Logf("Pool %s is paused, treating as up-to-date (Updated: %v, Updating: %v, Degraded: %v)", name, updated, updating, degraded)
+		return true, updating
+	}
 	if updated && !updating && !degraded {
-		return true, nil
+		return true, updating
 	}
 	framework.Logf("Pool %s is still reporting (Updated: %v, Updating: %v, Degraded: %v)", name, updated, updating, degraded)
-	return false, nil
+	return false, updating
```
bparees, line 520:
```suggestion
	return updated, updating
```

test/e2e/upgrade/upgrade.go
```diff
@@ -495,9 +509,13 @@ func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (bool
 			degraded = true
 		}
 	}
+	if paused {
+		framework.Logf("Pool %s is paused, treating as up-to-date (Updated: %v, Updating: %v, Degraded: %v)", name, updated, updating, degraded)
+		return true, updating
+	}
 	if updated && !updating && !degraded {
-		return true, nil
+		return true, updating
```
bparees, line 517:
```suggestion
		return updated, updating
```

test/e2e/upgrade/upgrade.go
```diff
@@ -495,9 +509,13 @@ func IsPoolUpdated(dc dynamic.NamespaceableResourceInterface, name string) (bool
 			degraded = true
 		}
 	}
```
bparees, line 511: do you need to handle the case where you do not find one of the `updated`, `updating`, or `degraded` condition markers?  You're defaulting the values, so if for some reason you don't find one of the conditions you're potentially going to report back a state that doesn't reflect reality.

smarterclayton: This is a backport, so let's take review comments that are generic to 4.8

vrutkovs: Cluster would be severely broken if machine pool is not either of this conditions and we'll have operator failure showing that

PR #25942, opened 2021-03-05
Bug 1935715: [release-4.6] test: Detect when the master pool is still updating after upgrade

vrutkovs:
The MCO is required to roll out the master config pool prior to
reporting level, and thus CVO should never reach Available at a
new version without the master pool being updated. Add an extra
check to the pool rollout - if we see the master pool with an
Updating condition flag it and fail the upgrade job. Also handle
paused pools in the wait loop (a paused pool is effectively "we

Cherry-pick of https://github.com/openshift/origin/pull/25922 on release-4.6

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -398,6 +398,7 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 		return err
 	}
 
+	var errMasterUpdating error
 	if err := disruption.RecordJUnit(
 		f,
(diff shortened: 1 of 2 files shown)


sdodson:
/hold
We should hold this until fixes for https://bugzilla.redhat.com/show_bug.cgi?id=1955929 have made it back to 4.6
We know that this will fail in certain conditions and this is a new constraint not previously imposed, lets at least fix the one condition we're aware of so we don't stall pipeline delivery in 4.6. @smarterclayton @wking Please feel free to say otherwise.
FYI @kikisdeliveryservice

PR #25943, opened 2021-03-08
test/extended/etcd: fix leader change test

simonpasquier:
The test would panic if Prometheus returns no result.

Spotted in https://github.com/openshift/cluster-monitoring-operator/pull/1044

Signed-off-by: Simon Pasquier <spasquie@redhat.com>


simonpasquier:
@deads2k can you review this PR? nothing fancy here, just a bit of test hardening.

PR #25946, opened 2021-03-08
Add list of canary tests to openshift-tests image

bparees:
Adds a list of tests that pass with extremely high reliability.  We will
run these tests in a new e2e job.  The origina list was generated
by taking tests that have passed at a rate of 99% or higher over the last week,
from sippy data.


bparees:
@smarterclayton @stevekuznetsov ptal at this and https://github.com/openshift/release/pull/16595 and tell me if this is an acceptable way to do this.  I'm open to suggestions but fundamentally i want to maintain a text file list of tests somewhere that is release-branch managed, and then have a scheduled CI job (and maybe in the future a PR-based CI job also) that runs those tests.

deads2k:
how will you diff this list over time as tests move in and out?

cmd/openshift-tests/e2e.go
```diff
@@ -340,6 +340,34 @@ var staticSuites = testSuites{
 		},
 		PreSuite: suiteWithProviderPreSuite,
 	},
+	{
+		TestSuite: ginkgo.TestSuite{
+			Name: "experimental/reliability/minimal/parallel",
```
smarterclayton, line 345: Why not combine these into a single run (serial + parallel will automaticalyl get run correctly)?

bparees: i didn't realize it would do that automatically, sure i can combine.

bparees: (but why do we have separate serial jobs today, then?)

bparees: i mean i based this on:

```
	{
		TestSuite: ginkgo.TestSuite{
			Name: "openshift/conformance/parallel",
			Description: templates.LongDesc(`
		Only the portion of the openshift/conformance test suite that run in parallel.
		`),
				if isDisabled(name) {
					return false
				}
				return strings.Contains(name, "[Suite:openshift/conformance/parallel")
			},
			Parallelism:          30,
			MaximumAllowedFlakes: 15,
			SyntheticEventTests:  ginkgo.JUnitForEventsFunc(synthetictests.StableSystemEventInvariants),
		},
		PreSuite: suiteWithProviderPreSuite,
	},
	{
		TestSuite: ginkgo.TestSuite{
			Name: "openshift/conformance/serial",
			Description: templates.LongDesc(`
		Only the portion of the openshift/conformance test suite that run serially.
		`),
				if isDisabled(name) {
					return false
				}
			},
			SyntheticEventTests: ginkgo.JUnitForEventsFunc(synthetictests.StableSystemEventInvariants),
		},
		PreSuite: suiteWithProviderPreSuite,
	},
```

bparees:
updated to a single suite after offline discussion with @smarterclayton, self-tagging.

PR #25953, opened 2021-03-09
indicate which tests correspond to an operator outage

deads2k:
To be expanded next to reflect tests failing during an apiserver availability outage, but that's a little more surgery.

This also adds monitor events for running tests, so we should see those in the monitor now.


deads2k:
Got them in the event stream.

/retest

mfojtik:
i like this!

/lgtm

PR #25954, opened 2021-03-09
NE-544: DNS: Add new dual-stack service DNS test

sgreene570:
test/extended/util/dns/dns.go:

OCP 4.8 now uses CoreDNS v1.8.1, which adds support for dual-stack clusterIP services (https://github.com/coredns/coredns/pull/4339). Add a new test to verify that CoreDNS pods respond to A and AAAA queries for a service URL (ie `<service>.<namespace>.svc`) as well for a service endpointslice URL (ie `<endpointslice_hostname>.<service>.<namespace>.svc`). Note that the other existing test in `dns.go` is disabled since the DNS operator does not support wildcard DNS.

test/extended/util/annotate/generated/zz_generated.annotations.go:
Regenerate

---

This is in support of https://issues.redhat.com/browse/NE-544.

### DIFF
--- a/test/extended/dns/dns.go
+++ b/test/extended/dns/dns.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"strings"
 
 	. "github.com/onsi/ginkgo"
 
 	kapiv1 "k8s.io/api/core/v1"
+	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -20,6 +22,8 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	watchtools "k8s.io/client-go/tools/watch"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
 )
 
 func createDNSPod(namespace, probeCmd string) *kapiv1.Pod {
@@ -109,6 +113,17 @@ func digForARecords(records map[string][]string, expect sets.String) string {
 	for name, ips := range records {
 		fileName := fmt.Sprintf("%s_endpoints@%s", fileNamePrefix, name)
 		probeCmd += fmt.Sprintf(`[ "$$(dig +short +notcp +noall +answer +search %s A | sort | xargs echo)" = "%s" ] && echo %q;`, name, strings.Join(ips, " "), fileName)
+		expect.Insert(fileName)
+	}
+	return probeCmd
+}
+
+func digForAAAARecords(records map[string][]string, expect sets.String) string {
+	var probeCmd string
+	fileNamePrefix := "test"
+	for name, ips := range records {
+		fileName := fmt.Sprintf("%s_endpoints_v6@%s", fileNamePrefix, name)
+		probeCmd += fmt.Sprintf(`[ "$$(dig +short +notcp +noall +answer +search %s AAAA | sort | xargs echo)" = "%s" ] && echo %q;`, name, strings.Join(ips, " "), fileName)
 		expect.Insert(fileName)
 	}
 	return probeCmd
@@ -262,6 +277,19 @@ func createServiceSpec(serviceName string, isHeadless bool, externalName string,
 	return s
 }
 
+func createDualStackServiceSpec(serviceName string, isHeadless bool, externalName string, selector map[string]string) *kapiv1.Service {
+	s := createServiceSpec(serviceName, isHeadless, externalName, selector)
+	s.Spec.IPFamilies = []kapiv1.IPFamily{
+		kapiv1.IPv4Protocol,
+		kapiv1.IPv6Protocol,
+	}
+
+	ipFamilyPolicy := kapiv1.IPFamilyPolicyRequireDualStack
+	s.Spec.IPFamilyPolicy = &ipFamilyPolicy
+
+	return s
+}
+
 func createEndpointSpec(name string) *kapiv1.Endpoints {
 	return &kapiv1.Endpoints{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


sgreene570:
Now that https://github.com/openshift/release/pull/16684 merged
/test e2e-metal-ipi-ovn-dualstack

sgreene570:
Make sure our first green run with this new test wasn't a fluke
/test e2e-metal-ipi-ovn-dualstack

test/extended/dns/dns.go
```diff
@@ -114,6 +118,17 @@ func digForARecords(records map[string][]string, expect sets.String) string {
 	return probeCmd
 }
 
+func digForAAAARecords(records map[string][]string, expect sets.String) string {
+	var probeCmd string
+	fileNamePrefix := "test"
```
candita, line 123: if `fileNamePrefix` never changes, just hard code it in the base string of the `Sprintf`

sgreene570: This is how the other `digForFooRecords` functions are laid out. I think I'll leave this as is for consistency sake.

test/extended/dns/dns.go
```diff
@@ -114,6 +118,17 @@ func digForARecords(records map[string][]string, expect sets.String) string {
...
+	for name, ips := range records {
+		fileName := fmt.Sprintf("%s_endpoints_v6@%s", fileNamePrefix, name)
+		probeCmd += fmt.Sprintf(`[ "$$(dig +short +notcp +noall +answer +search %s AAAA | sort | xargs echo)" = "%s" ] && echo %q;`, name, strings.Join(ips, " "), fileName)
```
candita, line 126: Why do some of the other queries use +tcp in addition to +notcp, and save results to another file, but not A or AAAA queries?

test/extended/dns/dns.go
```diff
@@ -381,4 +460,92 @@ var _ = Describe("[sig-network-edge] DNS", func() {
...
+
+	It("should answer A and AAAA queries for a dual-stack service", func() {
+		// Only run this test on dual-stack enabled clusters.
+		networkConfig, err := oc.AdminConfigClient().ConfigV1().Networks().Get(context.Background(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("unable to get cluster network config: %v", err)
+		}
+		usingIPv4 := false
+		usingIPv6 := false
+		for _, clusterNetworkEntry := range networkConfig.Status.ClusterNetwork {
```
candita, line 472: Is this the only way to determine a dual-stack?   What if it is dual-stack but only has ipv4 addresses for now?

sgreene570: Clusters have to be configured to use both a v4 and v6 network on creation if dual-stack is desired, I believe. I don't think you can change from dualstack to single stack or vice versa after install time.

This is how we check if we are using dual-stack in the ingress operator, by the way:

https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go#L569-L588

Did you have a different way to check this in mind?

test/extended/dns/dns.go
sgreene570: Hm, good question. I'm guessing the other query funcs test both TCP and UDP for extra code coverage. 

I think it's sufficient for this test to perform A and AAAA lookups via UDP only. How does that sound to you?

test/extended/dns/dns.go
candita: Maybe you could check the Load Balancers for `spec.IpFamilyPolicy`, as mentioned in https://kubernetes.io/docs/concepts/services-networking/dual-stack/#service-type-loadbalancer

candita: Oop, never mind - the latter is what you did here.  Checking just the `clusterNetwork.CIDR`.  Looks fine.

sgreene570: I think `networkConfig.Status.ClusterNetwork` could actually have more than just 2 entries. 

See https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/cluster_config.go#L19

test/extended/dns/dns.go
candita: I think it would be better for consistency sake to do it the same as the other queries.

test/extended/dns/dns.go
candita: https://github.com/openshift/cluster-network-operator/blob/3e9bc17218aa6dc6c3516264a301a2e118ceedc5/pkg/network/cluster_config.go#L46 shows it to be an error to have more than two.

sgreene570: That's for the cluster's service network CIDR, not the cluster network CIDR.

test/extended/dns/dns.go
```diff
@@ -381,4 +460,92 @@ var _ = Describe("[sig-network-edge] DNS", func() {
...
+		v6ips := ipsForEndpointSlice(v6)
+
+		var v4ClusterIP string
+		var v6ClusterIP string
+
+		for i, ipFamily := range service.Spec.IPFamilies {
+			if ipFamily == kapiv1.IPv4Protocol {
+				v4ClusterIP = service.Spec.ClusterIPs[i]
+			} else if ipFamily == kapiv1.IPv6Protocol {
+				v6ClusterIP = service.Spec.ClusterIPs[i]
```
Miciah, line 518: The Service API godoc says that the entries in `spec.ipFamilies` and `spec.clusterIPs` "must correspond" and that each field has a maximum of 2 entries; is it safe to assume that means the ordering is the same?

sgreene570: > is it safe to assume that means the ordering is the same?

That's how I interpreted the documentation.

knobunc (approved):
/approve
Thanks Steve!

knobunc:
@candita @Miciah I'm leaving the LGTM to you once you are satisfied with @sgreene570 's answers.

test/extended/dns/dns.go
sgreene570: Maybe we could fix this up in a follow up? Before this PR, `digForARecords` only tested DNS via UDP, so it made sense to me to have `digForAAAARecords` do the same.

test/extended/dns/dns.go
Miciah: Assumptions make me nervous, but worst case, we get spurious CI failures and need to revise the logic, so
/lgtm

PR #25957, opened 2021-03-11
test: When node upgrade sythetic resets, clear intermediate state

smarterclayton:
Each "leg" of an upgrade is independent and the same invariants
apply.


wking (approved):
/lgtm

PR #25958, opened 2021-03-11
test: External image test was silently failing

smarterclayton:
As a result we were not catching images that had not been mirrored.


smarterclayton:
Merging to prevent regressions (upgrade passed with this check)

PR #25963, opened 2021-03-15
Bug 1928946: Move apiserver graceful failure responsibility to invariant

smarterclayton:
The invariant test should fail the e2e run if a disruption occurs during the test. The e2e tests that the invariant was based on are
still useful for picking up install time failures, but they need to not fail (limit start time doesn't exclude install).

For now, failures for these invariants is in the invariant (scoped to the suite), while detection of flakes is the e2e test. We probably
need an openshift-tests run that checks install time invariants but that needs a bit of design work, and right now this failure at
install is not impacting cluster health.

https://github.com/openshift/origin/pull/25932 was the predecessor but confused the cause and solution (invariants are already scoped to the suite execution)


smarterclayton:
Waiting for upgrade status, then I will force merge this

smarterclayton:
Upgrade was unrelated, and the tests did pass.  Given one example of the failure, going to force merge to see if this can get our flake rate down to manageable.

PR #25968, opened 2021-03-16
test/extended/operators/daemon_set: Multus bug has been fixed

wking:
[Fixed][1] in openshift/cluster-network-operator#992.  This is the last bug, so I've torn out the `knownBroken` and `bugsAgainstBrokenDaemonSets` stuff, because we are unlikely to want to regress.  And if we end up wanting to regress, that code is still there in the Git history.

/assign @smarterclayton 

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1933159#c4

### DIFF
--- a/test/extended/operators/daemon_set.go
+++ b/test/extended/operators/daemon_set.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	g "github.com/onsi/ginkgo"
-	"github.com/openshift/origin/pkg/test/ginkgo/result"
 	exutil "github.com/openshift/origin/test/extended/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


test/extended/operators/daemon_set.go
```diff
@@ -46,19 +45,12 @@ var _ = g.Describe("[sig-arch] Managed cluster", func() {
 			e2e.Failf("unable to list daemonsets: %v", err)
 		}
 
-		// daemonsets that have an explicit exception, every entry here must have a bug associated
-		// e.g. "openshift-x/foo": "link to bug"
-		bugsAgainstBrokenDaemonSets := map[string]string{
```
smarterclayton, line 51: Keep all this infrastructure.  Have a comment in the empty map, and describe what the rules are for adding to it.

wking: Under what conditions would we add a new bug carve out, instead of just requiring the broken DaemonSet to be fixed?

PR #25970, opened 2021-03-17
Bug 1933760: Check for recent reboots on pod invariants

ehashman:
ref: https://bugzilla.redhat.com/show_bug.cgi?id=1933760

### DIFF
--- a/pkg/synthetictests/kubelet.go
+++ b/pkg/synthetictests/kubelet.go
@@ -3,6 +3,7 @@ package synthetictests
 import (
 	"bytes"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ehashman:
This seems to have worked. 5 flakes but none of them are the pods go from Running > Pending test, and CI suggests it's present in like 80% of 4.8 failures right now: https://search.ci.openshift.org/?search=pods+should+never+transition+back+to+pending&maxAge=336h&context=1&type=junit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

/cc @rphillips @deads2k @smarterclayton 

/skip
/retest

ehashman:
To gain slightly more confidence I guess I'd like to see this actually fail to make sure it's still catching cases...

This assumes the events are in sorted chronological order to work.

deads2k:
/hold

I think we've talked through this and determined that for non-static pods, the information is present to avoid going from running to pending.  If you have a prior container state that shows a running container and a current state without a running container, at some point in between that container terminated.

https://github.com/openshift/origin/pull/25978 is open to add debugging to show whether that data exists or not.

ehashman:
I'm not sure we want this, closing for now.

See latest update on https://bugzilla.redhat.com/show_bug.cgi?id=1933760#c7

PR #25978, opened 2021-03-17
DO NOT MERGE: Output pod.status when it regresses from running to pending

smarterclayton:


pkg/monitor/pod.go
```diff
@@ -94,11 +95,21 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
...
-						Level:   Warning,
-						Locator: locatePod(pod),
-						Message: fmt.Sprintf("pod moved back to Pending"),
-					})
+					if old == corev1.PodRunning {
+						debugPod, _ := json.MarshalIndent(oldPod.Status, "", "  ")
+						conditions = append(conditions, Condition{
+							Level:   Warning,
+							Locator: locatePod(pod),
+							Message: fmt.Sprintf("pod moved back to Pending: old status: %s", string(debugPod)),
```
ehashman, line 103: This might be a little verbose but honestly I think it might not be the worst to merge this in and see if we find anything weird/unexpected? We should be able to glean any other info we need from the must-gather.

smarterclayton: I was actually opening this to get data from the jobs, not to merge.  Very likely to reproduce within 1-2 retests.

PR #25971, opened 2021-03-17
test: Mark pod transition back to pending as a flake for now

smarterclayton:
It is continuously failing and blocking merges is unnecessary because
it is a symptom and we have an urgent bug.

PR #25979, opened 2021-03-17
[DNM] Mark testPodTransitions as flaky

ehashman:
See also #25970


ehashman:
missed https://github.com/openshift/origin/pull/25971

PR #25980, opened 2021-03-18
pkg/synthetictests/operators: Include the reason string in testOperatorState

wking:
Extending the the logic from 6ec446bad7 (#25918) to include the reason, since that slug both makes it easier to find the relevant operator code and makes it easier to match CI flakes with seen-in-the-wild Telemetry (where we get the status and reason, but not the message).

I'm also moving some string types into structured config/v1 types. Narrowly, this makes it easier for `GetOperatorConditionStatus` to return all its metadata without a big-list-of-strings API.  More broadly, this drops things like "assume status is a boolean, and neglect `Unknown` and unrecognized values".  This also means that we can't use `!status` to guess at what the previous status was, so I've dropped that and now only talk about the data we actually have.

### DIFF
--- a/pkg/monitor/operator.go
+++ b/pkg/monitor/operator.go
@@ -3,7 +3,6 @@ package monitor
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


wking:
[serial][1]:

```
authentication became Progressing=True at 2021-03-18 20:24:30.561470838 +0000 UTC -- reason/OAuthServerDeployment_GenerationNotObserved: OAuthServerDeploymentProgressing: Waiting for OAuth server observed generation 3 to match expected generation 4
authentication became Progressing=False at 2021-03-18 20:25:00.387047765 +0000 UTC -- reason/AsExpected: All is well
authentication became Progressing=True at 2021-03-18 20:25:43.157198745 +0000 UTC -- reason/OAuthServerDeployment_GenerationNotObserved: OAuthServerDeploymentProgressing: Waiting for OAuth server observed generation 6 to match expected generation 7
authentication became Progressing=False at 2021-03-18 20:26:08.163754873 +0000 UTC -- reason/AsExpected: All is well
```

So success :)

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25980/pull-ci-openshift-origin-master-e2e-aws-serial/1372628888714219520

PR #25982, opened 2021-03-18
test: Grant bare metal components an exception for not having requests

smarterclayton:
Bug should be fixed this release.


smarterclayton:
Current the test fails so tagging on the limit

smarterclayton:
WIll force merge if we hit the known flakes

smarterclayton:
All flakes are known flakes

PR #25986, opened 2021-03-18
test: Mark kubelet terminates non-gracefully as a flake

smarterclayton:
Now that we have a root cause, mark this a flake until the crio
fix is in place.


gabemontero:
@smarterclayton perhaps you are already aware, but this *seems* to be happening in 4.7 as well:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_builder/226/pull-ci-openshift-builder-release-4.7-e2e-aws/1374030848268963840

or do you think it is a different flavor?  

but if it is, would picking this back to 4.7 be an option?

rphillips:
/retitle Bug 1955610: test: Mark kubelet terminates non-gracefully as a flake

PR #25988, opened 2021-03-19
test: Synthetics should only return junit tests instead of a bool

smarterclayton:
The bool was complex and easy to get wrong, we now simply expect flakes to
be reported with a duplicate junit.


soltysh (approved):
/lgtm

PR #25990, opened 2021-03-19
test: grant ovirt-csi-driver an exception for not having requests

petr-muller:
Copycat of https://github.com/openshift/origin/pull/25982 for https://bugzilla.redhat.com/show_bug.cgi?id=1940876


smarterclayton:
Will force merge if we're caught on other flakes (to unblock the ovirt team) once i see enough green to know we're good

PR #26001, opened 2021-03-22
serialize the event monitor as json.  I think

deads2k:
step one towards something more interesting.


deads2k:
replaced by https://github.com/openshift/origin/pull/26006

PR #26004, opened 2021-03-22
allow separating monitor event identification from monitor event generation

deads2k:
This avoids package dep cycles we try to build analysis layers. The monitor event generation should not be relied upon directly by analysis built on top.

### DIFF
--- a/cmd/openshift-tests/images.go
+++ b/cmd/openshift-tests/images.go
@@ -9,17 +9,14 @@ import (
 	"strings"
 	"time"
 
+	imagev1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
+	"github.com/openshift/library-go/pkg/image/reference"
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/test/ginkgo"
+	"github.com/openshift/origin/test/extended/util/image"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
-
-	imagev1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
-	"github.com/openshift/library-go/pkg/image/reference"
-
-	"github.com/openshift/origin/pkg/monitor"
-	"github.com/openshift/origin/pkg/test/ginkgo"
-	"github.com/openshift/origin/test/extended/util/image"
-
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 	k8simage "k8s.io/kubernetes/test/utils/image"
 )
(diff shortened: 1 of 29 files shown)


cmd/openshift-tests/images.go
```diff
@@ -9,14 +9,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
```
mfojtik, line 12: nit:order :trollface:

mfojtik: also the package name is weird :-) `pkg/monitor/api` would be nicer ;-)

pkg/monitor/api.go
```diff
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
```
mfojtik, line 13: nit: move to bottom :trollface: 

Also somebody should set `goimport` for David

pkg/monitor/monitorapi/identification.go
```diff
@@ -0,0 +1,63 @@
+package monitorapi
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+func E2ETestLocator(testName string) string {
+	return fmt.Sprintf("e2e-test/%q", testName)
```
mfojtik, line 10: this will produce a weird string, no? `e2e-test/"foo bar`

pkg/monitor/monitorapi/identification.go
```diff
@@ -0,0 +1,63 @@
...
+func NodeFromLocator(locator string) (string, bool) {
+	if !strings.HasPrefix(locator, "node/") {
+		return "", false
+	}
+	parts := strings.SplitN(locator, "/", 2)
+	return parts[1], true
+}
+
+func OperatorLocator(testName string) string {
+	return fmt.Sprintf("node/%q", testName)
```
mfojtik, line 49: this should be `operator/%q`

deads2k: > this should be `clusteroperator/%q`

yay!

cmd/openshift-tests/images.go
deads2k: > also the package name is weird :-) `pkg/monitor/api` would be nicer ;-)

`api` turned into a bad package name in kube and these aren't quite serializeable.  I'll compare against http and httptest: that moment they realized fully qualified names existed for a reason.

pkg/monitor/monitorapi/identification.go
deads2k: > this will produce a weird string, no? `e2e-test/"foo bar"` .. do we want those quotes?

we need them for e2e tests since those can contain slashes.

deads2k: I removed the others.

PR #26005, opened 2021-03-22
create interval for clusteroperators

deads2k:
This shows the from/to behavior that we can build based on our window in time.

Creating intervals like this will make it easier to drive somethign like oschart.


pkg/monitor/monitor.go
```diff
@@ -135,6 +137,11 @@ func (m *Monitor) EventIntervals(from, to time.Time) monitorapi.EventIntervals {
 	intervals := filterSamples(samples, from, to)
 	events = filterEvents(events, from, to)
 
+	// create additional intervals from events
+	intervals = append(intervals, intervalcreation.IntervalsFromEvents_OperatorAvailable(events, from, to)...)
+	intervals = append(intervals, intervalcreation.IntervalsFromEvents_OperatorProgressing(events, from, to)...)
```
smarterclayton, line 142: I assume you know these don't belong hardcoded here and need to be passed in to cmd_runsuite via Options

deads2k: > I assume you know these don't belong hardcoded here and need to be passed in to cmd_runsuite via Options

moved to where we specify the monitors, so they are paired.

PR #26006, opened 2021-03-22
Serialize intervals to json

deads2k:
This will be a serialization that is suited towards building an oschart like view of an e2e test.


pkg/monitor/intervalcreation/operator.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		currentCondition := monitorapi.GetOperatorConditionStatus(event.Message)
+		if currentCondition == nil {
+			continue
+		}
+		if currentCondition.Type != conditionType {
+			continue
+		}
+
+		lastCondition := operatorToInterestingBadCondition[operatorName]
```
mfojtik, line 40: @deads2k probably need a `ok` check?

PR #26009, opened 2021-03-23
fix the rest of parallel tests working with old token format

stlaz:
/assign @sttts

### DIFF
--- a/test/extended/authorization/scopes.go
+++ b/test/extended/authorization/scopes.go
@@ -59,8 +59,9 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] scopes", func() {
 				t.Fatalf("unexpected error: %v", err)
 			}
 
+			whoamiOnlyTokenStr, sha256WhoamiOnlyTokenStr := exutil.GenerateOAuthTokenPair()
 			whoamiOnlyToken := &oauthv1.OAuthAccessToken{
-				ObjectMeta:  metav1.ObjectMeta{Name: "whoami-token-plus-some-padding-here-to-make-the-limit-" + oc.Namespace()},
+				ObjectMeta:  metav1.ObjectMeta{Name: sha256WhoamiOnlyTokenStr},
 				ClientName:  "openshift-challenging-client",
 				ExpiresIn:   200,
 				Scopes:      []string{scope.UserInfo},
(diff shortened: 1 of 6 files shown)


test/extended/authorization/scopes.go
```diff
@@ -59,8 +59,9 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] scopes", func() {
 				t.Fatalf("unexpected error: %v", err)
 			}
 
+			privWhoamiOnlyToken, pubWhoamiOnlyToken := exutil.GenerateOAuthTokenPair()
 			whoamiOnlyToken := &oauthv1.OAuthAccessToken{
-				ObjectMeta:  metav1.ObjectMeta{Name: "whoami-token-plus-some-padding-here-to-make-the-limit-" + oc.Namespace()},
```
sttts, line 63: where is "to make the limit" proporty gone? Do we lose anything? Or this just does not matter anymore because we always hash?

test/extended/authorization/scopes.go
```diff
@@ -59,8 +59,9 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] scopes", func() {
 				t.Fatalf("unexpected error: %v", err)
 			}
 
+			privWhoamiOnlyToken, pubWhoamiOnlyToken := exutil.GenerateOAuthTokenPair()
```
sttts, line 62: whoamiOnlyToken, sha256WhoamiOnlyToken ?

test/extended/authorization/scopes.go
stlaz: the tokens have a requirement on at least 32 characters length, that's something we're keeping even though it may not make too much sense now that we're using hashed tokens

test/extended/oauth/token.go
```diff
@@ -27,42 +27,6 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer] OAuth Authenticator", func()
 	oc := exutil.NewCLI("oauth-access-token-e2e-test")
 	ctx := context.Background()
 
-	g.It(fmt.Sprintf("accepts classic non-prefixed access tokens"), func() {
```
sttts, line 30: does this imply an inverse test when we switch over?

A blocker BZ would be good not to forget.

stlaz: we won't be able to e2e test that since we won't be able to create such a token, but might still be worth a unit test

test/extended/authorization/scopes.go
```diff
@@ -155,8 +156,9 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] scopes", func() {
 				t.Fatalf("unexpected error: %v", err)
 			}
 
+			privNonEscalatingEditToken, pubNonEscalatingEditTOken := exutil.GenerateOAuthTokenPair()
```
sttts, line 159: all the others as well?

test/extended/oauth/client_secret.go
```diff
@@ -20,7 +20,7 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
-const authzTokenName string = "oauth-client-with-plus-with-more-than-thirty-two-characters-in-this-very-long-name"
+var privAuthzToken, pubAuthzToken = exutil.GenerateOAuthTokenPair()
```
sttts, line 23: also here

test/extended/project/project.go
```diff
@@ -572,8 +571,9 @@ func GetScopedClientForUser(oc *exutil.CLI, username string, scopes []string) (*
 		return nil, err
 	}
 
+	privToken, pubToken := exutil.GenerateOAuthTokenPair()
```
sttts, line 574: and here

stlaz:
/test e2e-gcp-builds
didn't think this one needs to pass, it does not have the "Required" label...

PR #26016, opened 2021-03-25
test/extended/prometheus: add test validating alert severity

simonpasquier:
In the context of https://github.com/openshift/enhancements/pull/637, this change adds a test that ensures all alerting rules shipped by core components follow the established standards. In particular, all alerting rules must have a `severity` label and the value should be either `info`, `warning` or `critical` (the `Watchdog` alert is the only exception).

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -40,6 +40,149 @@ import (
 	helper "github.com/openshift/origin/test/extended/util/prometheus"
 )
 
+var _ = g.Describe("[sig-instrumentation][Late] Alerting rules", func() {
+	defer g.GinkgoRecover()
+	var (
+		oc = exutil.NewCLIWithoutNamespace("prometheus")
+
+		url, bearerToken string
+	)
+	g.BeforeEach(func() {
+		var ok bool
+		url, _, bearerToken, ok = helper.LocatePrometheus(oc)
+		if !ok {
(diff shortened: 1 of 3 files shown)


paulfantom (approved):
/lgtm

test/extended/prometheus/prometheus.go
```diff
@@ -36,6 +36,120 @@ import (
...
+					continue
+				}
+
+				if ar.Name == "Watchdog" {
+					continue
+				}
+
+				severity, found := ar.Labels["severity"]
+				if !found {
+					e2e.Failf("alert %q should have a 'severity' label", ar.Name)
```
smarterclayton, line 93: This fails right away, vs showing all problems.  I might suggest adapting the structure in https://github.com/openshift/origin/pull/26020/files#diff-cb149e746da7ca315f8c9b852dd6f407e04e16a7199587be3ae745fa57fe681eR67 which we are using in more tests now to accumulate failures and report them.  that also gives you the ability to define the exceptions for now and merge (without having to chase the miscreants)

simonpasquier: yep I've realized this later on :)

simonpasquier:
@smarterclayton can you have another look please? The test should report all alerting rules that don't have a severity label or define a severity value other than `info`, `warning` or `critical`.

simonpasquier:
@smarterclayton friendly ping

bison:
/close

Closing this (if I can) in favor of #26476.

PR #26017, opened 2021-03-25
test: skip checking nodejs 10 and perl 5.26

vrutkovs:
SCL org has deprecated Perl 5.26 and NodeJS 10, so we should be checking Perl 5.30 and NodeJS 12 instead.

See https://github.com/openshift/cluster-samples-operator/pull/360


adambkaplan (approved):
/lgtm

vrutkovs:
/hold

https://bugzilla.redhat.com/show_bug.cgi?id=1943539 is blocking merges

PR #26021, opened 2021-03-25
test: Do not check github for commits on OLM test

smarterclayton:
We have private repos in the event an embargoed CVE must be verified.
This test would break if we ran one of those builds. Change the logic
to simply ensure the commits are the same.


smarterclayton:
/hold

For someone to confirm my interpretation of the test

smarterclayton:
Failed in https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-aws-serial-4.8/1375162263761063936, and that may have caught an ART bug (art should not be forking the normal repos UNLESS a CVE is present).  We should have another test somewhere that verifies ART is always building from a commit that can be found in the authoritative source repos when it's not a private build.

cgwalters (approved):
Looks sane to me; I'd probably log the two commits that didn't match explicitly in the error, but AFAICS we do log them above too.

PR #26022, opened 2021-03-26
test: App availability test during upgrade must set minReady

smarterclayton:
Apps behind service load balancers must also set a minReadySeconds
threshold longer than the time it takes for the load balancer to
see the new pod, otherwise they can fail to be reachable which can
cause the SLB to route to all pods including the unready ones
(implementation dependent behavior).


cgwalters (approved):
Feels like the 33 could be a constant somewhere?  But looks sane to me.

PR #26027, opened 2021-03-29
Bug 1948634: upgrade: allow upgrades without version change

vrutkovs:
Paused workers test requires watching upgrade to complete on final stage when worker pools is resumed and upgrade proceeds. This commit would record this warning in the log instead of throwing the error

See test results in https://github.com/openshift/release/pull/15939


test/e2e/upgrade/upgrade.go
```diff
@@ -226,9 +226,10 @@ func getUpgradeContext(c configv1client.Interface, upgradeImage string) (*upgrad
 	}
 
 	upgradeImages := strings.Split(upgradeImage, ",")
-	if (len(upgradeImages[0]) > 0 && upgradeImages[0] == current.Image) || (len(upgradeImages[0]) > 0 && upgradeImages[0] == current.Version) {
-		return nil, fmt.Errorf("cluster is already at version %s", versionString(*current))
-	}
+	// Commented version check so that final step in paused workers testing could pass
```
smarterclayton, line 229: Turn it into a log message and i think this is ok - the guard is really just there to prevent oopses but i guess it's reasonable to simply verify the final state and the early state.

PR #25897, opened 2021-02-15
Bug 1947806: test/extended/router: enable http/2, gRPC router tests

frobware:
Needs openshift/cluster-ingress-operator#585.

This PR re-enables the http/2 & gRPC tests. They were previously 
disabled as the default ingresscontroller is not configured with http/2
support by default.

There are a number of changes here:

- The text fixture has been split up so that we can first apply the
pod/service, then apply the configured routes, then roll out a
dedicated ingresscontroller.

- There is a new fixture to rollout a new router shard with http/2
enabled.

- There is a new test fixture for the routes which are now parameterised
with an explicit domain and namespace selector to match on the router
shard type.

- I've deleted all the old test-fixture generation.

- The test now spins up a dedicated ingresscontroller.

PR #26028, opened 2021-03-29
Enable grpc-interop router tests

frobware:
Re-enables the gRPC interop tests.
This tests now stands up its own router shard that has http/2 enabled.

Builds on #25897.


frobware:
/hold

Need to find out what image I can use when building the server code in the test pod.

frobware:
Closing in favour of https://github.com/openshift/cluster-ingress-operator/pull/585

PR #26030, opened 2021-03-29
Bug 1948021: test/extended/operators: Require Upgradeable=True or unset for "start all core operators"

wking:
E.g. [bug#1928141][1] has users who are confused by `Upgradeable=Unknown`, of which we currently have [a few][2]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-aws-4.8/1375324509623554048/artifacts/e2e-aws/clusteroperators.json | jq -r '.items[] | .metadata.name as $n | .status.conditions[] | select(.status == "Unknown") | $n + " " + .type + " " + .reason'
kube-storage-version-migrator Upgradeable NoData
openshift-controller-manager Upgradeable NoData
service-ca Upgradeable NoData
```

This commit adds a post-install guard to reject as surprising any `Upgradeable!=True` conditions.  We should be upgradeable immediately post-install.  Not setting an `Upgradeable` condition is still fine, and is equivalent to explicitly setting `Upgradeable=True`.

This commit also extends `isTransitionToGoodOperatorState` to reject as surprising any condition which is neither `True` nor `False`.  We shouldn't be in any situations during an e2e run that would cause an operator to transition to `Unknown` or other unrecognized status during the run.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1928141
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-aws-4.8/1375324509623554048

### DIFF
--- a/test/extended/operators/operators.go
+++ b/test/extended/operators/operators.go
@@ -200,21 +200,30 @@ func condition(cv objx.Map, condition string) objx.Map {
 // It also returns a slice of types for which a condition entry was
 // expected but not supplied on the ClusterOperator.
 func surprisingConditions(co objx.Map) ([]configv1.ClusterOperatorStatusCondition, []configv1.ClusterStatusConditionType) {
+	name := co.Get("metadata.name").String()
 	var badConditions []configv1.ClusterOperatorStatusCondition
 	var missingTypes []configv1.ClusterStatusConditionType
 	for _, conditionType := range []configv1.ClusterStatusConditionType{
 		configv1.OperatorAvailable,
 		configv1.OperatorDegraded,
+		configv1.OperatorUpgradeable,
 	} {
 		cond := condition(co, string(conditionType))
 		if len(cond) == 0 {
-			missingTypes = append(missingTypes, conditionType)
+			if conditionType != configv1.OperatorUpgradeable {
+				missingTypes = append(missingTypes, conditionType)
+			}
 		} else {
 			expected := configv1.ConditionFalse
-			if conditionType == configv1.OperatorAvailable {
+			if conditionType == configv1.OperatorAvailable || conditionType == configv1.OperatorUpgradeable {
 				expected = configv1.ConditionTrue
 			}
 			if cond.Get("status").String() != string(expected) {
+				if conditionType == configv1.OperatorUpgradeable && (name == "kube-storage-version-migrator" || // https://bugzilla.redhat.com/show_bug.cgi?id=1928141
+					name == "openshift-controller-manager" || // https://bugzilla.redhat.com/show_bug.cgi?id=1948011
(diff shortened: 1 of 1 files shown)


wking:
[e2e-agnostic-cmd][1] finds the expected failures:

```
fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: Mar 29 21:01:54.257: Some cluster operators are not ready: kube-storage-version-migrator (Upgradeable=Unknown NoData: ), openshift-controller-manager (Upgradeable=Unknown NoData: ), service-ca (Upgradeable=Unknown NoData: )
```

@deads2k, @tkashem: idle this PR until those get fixed?  Clone [rhbz#1928141][2] over to the other two operators and add specific, bug-linked exceptions so we can land this guard before the operators get fixed?  Other approaches?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26030/pull-ci-openshift-origin-master-e2e-agnostic-cmd/1376629467992035328
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1928141

wking:
Rebased around #26034 with f22fdd4 -> 60d422292, which also drops my opinions on unrecognized conditions having unrecognized status values.  This is now only about `Upgradeable` during the `start all core operators` test.

deads2k:
I'm fine with the logical change. But you cannot pass CI without building a ratchet and/or fixing the three repos

wking:
I've filed bugs for the remaining two `Upgradeable=Unknown` operators and added exceptions for all three bugs in 60d4222927 -> 82542badb1, which should green up CI.  As the bugs get fixed, we can remove the exceptions.

deads2k:
The ratchet looks reasonable to prevent more.

/lgtm
/hold

holding until you see green CI. once you're green, feel free to remove the hold.

wking:
[e2e-aws-serial][1] is happy, which is enough for me.

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/26030/pull-ci-openshift-origin-master-e2e-aws-serial/1380591687784468480/artifacts/e2e-aws-serial/openshift-e2e-test/artifacts/e2e.log | grep 'passed.*start all core 
operators'
passed: (1s) 2021-04-09T19:39:59 "[sig-arch][Early] Managed cluster should start all core operators [Skipped:Disconnected] [Suite:openshift/conformance/parallel]"
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26030/pull-ci-openshift-origin-master-e2e-aws-serial/1380591687784468480

PR #26037, opened 2021-03-30
require operator availability and non-degraded for upgrades

deads2k:
re-uses the operator monitor to avoid degraded and unavailable operators during an upgrade.


smarterclayton:
lgtm once you see the output and it looks happy to you

deads2k:
yes, this is showing the correct failures for upgrade.  Fun days ahead.

PR #26044, opened 2021-04-05
Bug 1947806: test/extended/router: Re-enable h2spec tests

frobware:
- Requires https://github.com/openshift/cluster-ingress-operator/pull/590
- Builds on https://github.com/openshift/origin/pull/25897 and that should merge first

This PR re-enables the h2spec tests They were previously
disabled as the default ingresscontroller is not configured with http/2
support by default.

There are a number of changes here:

- The text fixture has been split up so that we can first apply the
pod/service, then apply the configured routes, then roll out a
dedicated ingresscontroller.

- There is a new fixture to rollout a new router shard with http/2
enabled.

- There is a new test fixture for the routes which are now parameterised
with an explicit domain and namespace selector to match on the router
shard type.

- The test now spins up a dedicated ingresscontroller.

- Rework and simplify the h2spec tests to only test via a passthrough route.

I ran the tests with `export ROUTER_H2SPEC_SAMPLE=100` and we have 2
failures, both related to GOAWAY; previous discussion https://github.com/haproxy/haproxy/issues/471#issuecomment-591420924

### DIFF
--- a/test/extended/router/h2spec.go
+++ b/test/extended/router/h2spec.go
@@ -2,8 +2,8 @@ package router
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"sort"
 	"strconv"
@@ -13,11 +13,11 @@ import (
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 
-	routev1 "github.com/openshift/api/route/v1"
+	configv1 "github.com/openshift/api/config/v1"
 	routeclientset "github.com/openshift/client-go/route/clientset/versioned"
 	"github.com/openshift/origin/test/extended/router/h2spec"
+	"github.com/openshift/origin/test/extended/router/shard"
 	exutil "github.com/openshift/origin/test/extended/util"
-	"github.com/openshift/origin/test/extended/util/url"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
@@ -31,18 +31,16 @@ type h2specFailingTest struct {
 	TestNumber int
 }
 
-type h2specRouteTypeTest struct {
-	routeType     routev1.TLSTerminationType
-	hostname      string
-	knownFailures map[string]bool
-}
-
 var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Router]", func() {
 	defer g.GinkgoRecover()
 
 	var (
-		configPath = exutil.FixturePath("testdata", "router", "router-h2spec.yaml")
-		oc         = exutil.NewCLI("router-h2spec")
+		h2specServiceConfigPath     = exutil.FixturePath("testdata", "router", "router-h2spec.yaml")
+		h2specRoutesConfigPath      = exutil.FixturePath("testdata", "router", "router-h2spec-routes.yaml")
+		h2specRouterShardConfigPath = exutil.FixturePath("testdata", "router", "router-shard.yaml")
+
+		oc              = exutil.NewCLI("router-h2spec")
+		shardConfigPath string // computed
 	)
 
 	// this hook must be registered before the framework namespace teardown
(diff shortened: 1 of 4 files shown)


test/extended/router/h2spec.go
```diff
@@ -108,30 +148,53 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
...
+			//              Actual: Connection closed
+			//
+			// 147 tests, 145 passed, 0 skipped, 2 failed
+			knownFailures := map[string]bool{
+				"http2/5.4.1.2": true,
+				"http2/6.9.1.2": true,
+			}
+			for _, f := range failures {
+				if _, exists := knownFailures[f.ID()]; exists {
+					failureCount -= 1
```
candita, line 191: Why not do this section with gomega for consistency in the output?

test/extended/router/h2spec.go
```diff
@@ -153,63 +216,43 @@ func failingTests(testSuites []*h2spec.JUnitTestSuite) []h2specFailingTest {
 	return failures
 }
 
-func runConformanceTests(oc *exutil.CLI, t h2specRouteTypeTest) []*h2spec.JUnitTestSuite {
-	podName := "h2spec"
-	e2e.ExpectNoError(e2epod.WaitForPodNameRunningInNamespace(oc.KubeClient(), podName, oc.KubeFramework().Namespace.Name))
-
-	var results []*h2spec.JUnitTestSuite
+func runConformanceTests(oc *exutil.CLI, host, podName string) ([]*h2spec.JUnitTestSuite, error) {
+	g.By("Running the h2spec CLI test")
```
candita, line 220: Shouldn't podName be used in the By clause?  Just in case you want to use this func for some other podName?

test/extended/router/h2spec.go
```diff
@@ -153,63 +216,43 @@ func failingTests(testSuites []*h2spec.JUnitTestSuite) []h2specFailingTest {
...
 
-	o.Expect(wait.PollImmediate(time.Second, 5*time.Minute, func() (bool, error) {
-		g.By("Running the h2spec CLI test")
+	// this is the output file in the pod
+	outputFile := "/tmp/h2spec-results"
```
candita, line 223: Should the `outputFile` be composed from the `podName`, or passed in as a parameter?

candita: Or just hard-coded in the `h2specCommand` func?

test/extended/router/h2spec.go
```diff
@@ -153,63 +216,43 @@ func failingTests(testSuites []*h2spec.JUnitTestSuite) []h2specFailingTest {
...
-		if data == "" {
-			e2e.Logf("[%s] error: zero length results file; retrying...", t.routeType)
-			return false, nil
-		}
+	// h2spec will exit with non-zero if _any_ test in the suite
+	// fails, or if there is a dial timeout, so we ignore the
+	// error. If we can fetch the results and if we can decode the
+	// results and we have > 0 test suites from the decoded
+	// results then assume the test ran.
+	output, _ := e2e.RunHostCmd(oc.Namespace(), podName, h2specCommand(h2specDialTimeoutInSeconds, host, outputFile))
```
candita, line 230: Here is another section which could use gomega too

test/extended/router/h2spec.go
```diff
@@ -219,34 +262,35 @@ func runConformanceTestsAndLogAggregateFailures(oc *exutil.CLI, tests []h2specRo
 		return index
 	}
 
-	printFailures := func(t h2specRouteTypeTest, prefix string, m map[string]int) {
+	printFailures := func(prefix string, m map[string]int) {
 		for _, id := range sortKeys(m) {
-			e2e.Logf("[%s] %sTestCase ID: %q, cumulative failures: %v", t.routeType, prefix, id, m[id])
+			e2e.Logf("%sTestCase ID: %q, cumulative failures: %v", prefix, id, m[id])
```
sgreene570, line 267: Add a space after the first %s here

test/extended/router/h2spec.go
```diff
@@ -55,47 +53,89 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
...
+			g.By("Creating a test-specific router shard")
+			shardConfigPath, err = shard.DeployNewRouterShard(oc, 10*time.Minute, shard.Config{
+				FixturePath: h2specRouterShardConfigPath,
+				Domain:      shardFQDN,
+				Type:        oc.Namespace(),
+			})
+			o.Expect(err).NotTo(o.HaveOccurred(), "new router shard did not rollout")
+
+			g.By("Getting LB service")
+			shardService, err := getRouterService(oc, 10*time.Minute, "router-"+oc.Namespace())
```
sgreene570, line 117: 10 minutes to get the ingress controller's service seems high.

test/extended/router/h2spec.go
```diff
@@ -55,47 +53,89 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(shardService).NotTo(o.BeNil())
+			o.Expect(shardService.Status.LoadBalancer.Ingress).ShouldNot(o.BeEmpty())
+
+			var addrs []string
+
+			if len(shardService.Status.LoadBalancer.Ingress[0].Hostname) > 0 {
```
sgreene570, line 124: Doesn't have to be now but at some point we should consider creating a helper function for this "Ingress Controller DNS resolves as expected" logic.

test/extended/router/h2spec.go
```diff
@@ -55,47 +53,89 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
...
-				urlTester.Within(30*time.Second, url.Expect("GET", "https://"+hostname).Through(hostname).SkipTLSVerification().HasStatusCode(200))
-				routeTypeTests[i].hostname = hostname // now valid for the remaining tests
+				o.Expect(addrs).NotTo(o.BeEmpty())
+			} else {
+				addrs = append(addrs, shardService.Status.LoadBalancer.Ingress[0].IP)
 			}
 
+			g.By("Waiting for route hostname to register in DNS")
+			host := "h2spec-passthrough." + shardFQDN
+			addrs, err = resolveHostAsAddress(oc, time.Minute, 15*time.Minute, host, addrs[0])
```
sgreene570, line 135: DNS might be resolving properly for the h2spec client pod well before it would for the tester binary, but I guess this just puts us on the overly cautious side of things.

sgreene570:
Overall LGTM. Discussed use-case of `ROUTER_H2SPEC_SAMPLE` offline. Will take a final look once https://github.com/openshift/origin/pull/25897 & https://github.com/openshift/cluster-ingress-operator/pull/590 merge.
/hold

test/extended/router/h2spec.go
frobware: I would rather not dictate behaviour. "Please tell me where you'd like the output" is the goal of specifying this here.

test/extended/router/h2spec.go
frobware: Yep, agreed. And I started that and backed out. :)

test/extended/router/h2spec.go
frobware: The behaviour here is now different; we wait for DNS to register for the route under test. Once DNS has registered successfully we now expect the test to run. There should no longer be any need to retry.

test/extended/router/h2spec.go
candita: I.e., something like `o.Expect(knownFailures).ShouldNot(HaveKey(f.ID()))`

test/extended/router/h2spec.go
candita: Otherwise, is there any need to pass in podName if it is always "h2spec"?

sgreene570:
LGTM pending merge of https://github.com/openshift/origin/pull/25897 followed by rebase.

frobware:
fail [k8s.io/kubernetes@v1.20.0/test/e2e/framework/pod/resource.go:483]: failed to create new exec pod in namespace: e2e-test-check-for-alerts-pnjnk

/retest

test/extended/router/h2spec.go
frobware: The purpose is to gather all of them that fail and report them wholesale as opposed to failing on the first one that fails.

PR #26053, opened 2021-04-07
Bug 1945398: e2e requestheader: properly wait for a new oauth-server rollout

stlaz:
Instead of waiting for separate pods, rely on the operator reporting
properly what the status of the rollout is.

Previously, we were waiting for pods for 2 minutes and then the operator
go progressing. But:
- 2 minutes for a pod to rollout does not seem to be enough in 33% of
  the cases with the new rollout strategy
- if the operator is still progressing, some pods might still use the
  old config, hence the original code issue with the old config still
  being used

### DIFF
--- a/test/extended/oauth/requestheaders.go
+++ b/test/extended/oauth/requestheaders.go
@@ -348,57 +348,8 @@ func generateCert(caCert *x509.Certificate, caKey *rsa.PrivateKey, cn string, ek
 }
 
 func waitForNewOAuthConfig(oc *exutil.CLI, caCerts *x509.CertPool, oauthURL string, configChanged time.Time) {
-	// check that the pods running in openshift-authentication NS already reflect our changes
-	err := wait.PollImmediate(time.Second, 2*time.Minute, func() (bool, error) {
-		pods, err := oc.AdminKubeClient().CoreV1().Pods("openshift-authentication").List(context.Background(), metav1.ListOptions{})
-		if err != nil {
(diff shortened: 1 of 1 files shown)


test/extended/oauth/requestheaders.go
```diff
@@ -488,3 +439,32 @@ func getTokenFromResponse(resp *http.Response) string {
...
+		if progressing == nil || progressing.Status != expectedProgressing {
+			e2e.Logf("Waiting for progressing condition to be %q: %s", expectedProgressing, spew.Sdump(authn.Status.Conditions))
+			return false, nil
+		}
+
+		if expectedProgressing == configv1.ConditionFalse {
+			// make additional checkes on availability and degraded status
+			if clusteroperatorhelpers.IsStatusConditionFalse(authn.Status.Conditions, configv1.OperatorAvailable) ||
+				clusteroperatorhelpers.IsStatusConditionTrue(authn.Status.Conditions, configv1.OperatorDegraded) {
+				e2e.Logf("Waiting for available==True, progressing==False, degraded==False: %s", expectedProgressing, spew.Sdump(authn.Status.Conditions))
```
dmage, line 461: one placeholder (%s), two arguments

stlaz: fixed in latest revision

soltysh (approved):
/approve

stlaz:
None of the e2e-aws-serial tests that fail contain the failure of this test, so that's good.
What's bad is that neither of its 6 runs has passed for various other reasons :neutral_face:

PR #26059, opened 2021-04-10
Bug 1948233: Add openshift/network/third-party suite, for CNI plugin conformance

danwinship:
The certification team pointed out that it would make sense to backport this to 4.6, since we will be supporting it as the EUS release for a long time, and it would be simpler to have it use the same process as 4.7+.

/assign @knobunc


knobunc (commented):
/approve
/lgtm

PR #26061, opened 2021-04-12
Bug 1948629: upgrades: add "none" / "minimal" test suites to skip disruption tests

vrutkovs:
For SNO upgrades most disruption tests would fail as there is no HA. This option would allow SNO upgrade test suite to skip those

### DIFF
--- a/cmd/openshift-tests/upgrade.go
+++ b/cmd/openshift-tests/upgrade.go
@@ -77,6 +77,8 @@ func upgradeTestPreTest() error {
 	parseUpgradeOptions(opt.TestOptions)
 	upgrade.SetToImage(opt.ToImage)
 	switch opt.Suite {
+	case "none":
+		return filterUpgrade(upgrade.NoTests(), func(string) bool { return true })
 	case "platform":
 		return filterUpgrade(upgrade.AllTests(), func(name string) bool {
 			return name == controlplane.NewKubeAvailableWithNewConnectionsTest().Name() || name == controlplane.NewKubeAvailableWithNewConnectionsTest().Name()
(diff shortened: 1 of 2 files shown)


cmd/openshift-tests/upgrade.go
```diff
@@ -77,6 +77,8 @@ func upgradeTestPreTest() error {
 	parseUpgradeOptions(opt.TestOptions)
 	upgrade.SetToImage(opt.ToImage)
 	switch opt.Suite {
+	case "none":
+		return filterUpgrade(upgrade.MinimalTests(), func(string) bool { return true })
```
smarterclayton, line 81: Wait... none would be no tests.  None should be "zero things get run".  You still need to skip some tests for single mode (in the test, not here).

vrutkovs: Ah, sure. Lets start with none (mostly for paused workers) / minimal (SNO) suite

smarterclayton: I would prefer you don't add minimal though - that's what I'm specifically asking you not to do :)  To fix the tests.

As far as filtering, it was intended that you could filter test names with --dry-run and grep and then -f -, but it looks like that is broken for the upgrade subset tests.

smarterclayton: I would prefer to not run the invariants for upgrade for now, so none is probably sufficient for SNO

vrutkovs: Remove minimal, keeping none

smarterclayton:
/lgtm

I will try to fix the "filter tests by name"

PR #26063, opened 2021-04-12
Remove non conformance test from the minimal suite

bparees:


cmd/openshift-tests/e2e.go
```diff
@@ -1,18 +1,24 @@
...
 	exutil "github.com/openshift/origin/test/extended/util"
 	"k8s.io/kubectl/pkg/util/templates"
+	"k8s.io/kubernetes/openshift-hack/e2e/annotate"
 
 	_ "github.com/openshift/origin/test/extended"
 	_ "github.com/openshift/origin/test/extended/util/annotate/generated"
 )
 
+var (
+	excludedTestsFilter = regexp.MustCompile(strings.Join(annotate.ExcludedTests, `|`))
```
smarterclayton, line 19: No, you should exclude things that are not in conformance.

cmd/openshift-tests/e2e.go
```diff
@@ -347,13 +353,18 @@ var staticSuites = testSuites{
...
+				if !exists {
+					return false
+				}
+				return !isDisabled(name) && !excludedTestsFilter.MatchString(name)
 			},
 			Parallelism:          20,
 			MaximumAllowedFlakes: 15,
 			SyntheticEventTests:  ginkgo.JUnitForEventsFunc(synthetictests.StableSystemEventInvariants),
 		},
+		PreSuite: suiteWithKubeTestInitializationPreSuite,
```
smarterclayton, line 367: Correct

cmd/openshift-tests/e2e.go
bparees: why?  my goal here is to run any test that has established itself as reliable, regardless of what suite/bucket it is in.

smarterclayton: because tests in conformance mean one thing, and tests not in conformance mean "specialized suites that may not play well with others"

bparees: I thought "disruptive" and "serial" meant "may not play well with others"?

bparees: anyway i've updated to only include conformance tests.

bparees:
/override ci/prow/e2e-aws-csi 

discussion about failing csi is here: https://coreos.slack.com/archives/CBQHQFU0N/p1618578359192800

PR #26066, opened 2021-04-12
test: Expose node changes as intervals

smarterclayton:
Expose both existing node not ready intervals as well as add two new sources of intervals:

* The events generated by MCD creating quasi-accurate intervals (events are lossy and MCD logic omits some events some times)
* The transition between two different sets of config at the node level

image

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -52,60 +52,6 @@
 </script>
 
 <script>
-    var loc = window.location.href;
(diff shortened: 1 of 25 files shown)


smarterclayton:
Viewing and verification only, no bug.  Tagging

smarterclayton:
Bypassing queue - all failures are unrelated and trying to get a handle on better visualization of upgrade disruption

PR #26067, opened 2021-04-12
test: Enforce workload guarantees with respect to spreading and duplication

ravisantoshgudimetla:
Report which workloads do not specify spreading or have undesirable
combinations of rollout strategy via a test.

Made some cosmetic changes but mostly copied from https://github.com/openshift/origin/pull/26038/commits

### DIFF
--- a/test/extended/operators/workload_guarantees.go
+++ b/test/extended/operators/workload_guarantees.go
@@ -148,24 +148,23 @@ var _ = g.Describe("[sig-arch] Managed cluster", func() {
 				nonHAPrefix := "non-HA/"
 				key = nonHAPrefix + key
 
-				if replicas != 1 {
-					rule := fmt.Sprintf("%s/%s", key, "use_single_replica_non_ha")
-					violation := fmt.Sprintf("%s must set replicas to 1 in non-HA mode", rule)
-					if bug, ok := knownBugs[key]; ok {
-						knownBroken = append(knownBroken, fmt.Sprintf("%s (bug %s)", violation, bug))
-					} else {
-						invalidWorkloads = append(invalidWorkloads, violation)
-					}
-				} else if !isSoftAntiAffinity(template.Spec.Affinity, template.Spec.TopologySpreadConstraints) && replicas >= 2 {
-					rule := fmt.Sprintf("%s/%s", key, "use_soft_anti_affinity_non_ha")
-					violation := fmt.Sprintf("%s must set pod anti-affinity preferred during scheduling, with topologyKey 'kubernetes.io/hostname'", rule)
-					if bug, ok := knownBugs[key]; ok {
-						knownBroken = append(knownBroken, fmt.Sprintf("%s (bug %s)", violation, bug))
-					} else {
-						invalidWorkloads = append(invalidWorkloads, violation)
+				// Single replica validation is done via single node topology test
+				if !isSoftAntiAffinity(template.Spec.Affinity, template.Spec.TopologySpreadConstraints) {
+					if replicas > 1 {
+						if meta.Name == "image-registry" && meta.Namespace == "openshift-image-registry" {
+							e2e.Logf("Image registry can have more than one replica even when the cluster " +
+								"is running in non-HA mode")
+							continue
+						}
+						rule := fmt.Sprintf("%s/%s", key, "use_soft_anti_affinity_non_ha")
+						violation := fmt.Sprintf("%s must set pod anti-affinity preferred during scheduling, with topologyKey 'kubernetes.io/hostname'", rule)
+						if bug, ok := knownBugs[key]; ok {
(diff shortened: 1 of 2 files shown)


romfreiman:
@ravisantoshgudimetla can we check how it looks like on SNO? Or, that it's excluded?

ravisantoshgudimetla:
> @ravisantoshgudimetla can we check how it looks like on SNO? Or, that it's excluded?

Yes, Clayton suggested to include scenarios for SNO. I've updated the test to look at them.

test/extended/operators/workload_guarantees.go
```diff
@@ -0,0 +1,269 @@
...
+			}
+			if toleratesMaster {
+				continue
+			}
+
+			key := fmt.Sprintf("%s/%s/%s", workload.GetObjectKind().GroupVersionKind().Kind, meta.Namespace, meta.Name)
+
+			if inHAMode {
+				if replicas == 1 {
+					// TODO: We need to check if this is a misconfiguration.
```
ravisantoshgudimetla, line 136: There are some components that are have just 1 replica today, not sure if upgrades will depend on the availability of them.

test/extended/operators/workload_guarantees.go
```diff
@@ -0,0 +1,269 @@
...
+				}
+				workloadStatus, isInvalidWorkload := isInvalidWorkloadInHAMode(knownBugs, key, workload, template)
+				if isInvalidWorkload && workloadStatus != "" {
+					invalidWorkloads = append(invalidWorkloads, workloadStatus)
+				} else if !isInvalidWorkload && workloadStatus != "" {
+					knownBroken = append(knownBroken, workloadStatus)
+				}
+			} else if !inHAMode {
+				// For SNO cases
+				nonHAPrefix := "non-HA/"
```
ravisantoshgudimetla, line 148: @romfreiman  @dhellmann - Can you please review this block to see if it covers all the SNO cases?

dhellmann: I don't see any components listed in `knownBugs` with a key like the one built here. We know that some components still do not reduce their replica count for single-node deployments. The package server from OLM, for example (https://issues.redhat.com/browse/OLM-2017).

Rather than adding the job and blocking work for a bunch of teams, could you look at a single node deployment to see if there are other components that should be added to the list?

ravisantoshgudimetla: The main goal here is to identify the components that don't confirm to single replica and affinities and create bugs for them. I can access a SNO via CI, if you can assign someone to work with me in merging this test it'll be great. I especially want review from you folks on this block- https://github.com/openshift/origin/pull/26067/files#diff-3e1d6d8499bb3600ca701074807127cdecbd73803989c6178c026e3d2b48ad6cR159. @romfreiman  has mentioned that you are already covering single replica whitelist here - https://github.com/openshift/origin/blob/master/test/extended/single_node/topology.go#L102

test/extended/operators/workload_guarantees.go
```diff
@@ -0,0 +1,269 @@
...
+					continue
+				}
+				meta = &t.ObjectMeta
+				template = &t.Spec.Template
+				replicas = *t.Spec.Replicas
+			default:
+				panic("not an object")
+			}
+
+			if !strings.HasPrefix(meta.Namespace, "openshift-") {
```
omertuc, line 112: This will miss the `openshift` namespace. Not sure if it has any interesting deployments/statefulsets though

test/extended/operators/workload_guarantees.go
```diff
@@ -0,0 +1,269 @@
...
+			if term.TopologyKey == "kubernetes.io/hostname" {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func isSoftAntiAffinity(affinity *corev1.Affinity, spread []corev1.TopologySpreadConstraint) bool {
+	if affinity != nil && affinity.PodAntiAffinity != nil {
```
omertuc, line 199: Do we want to fail in non-HA mode when workloads don't define any affinity? i.e. do we care whether it's `Preferred` or just no anti-affinity defined at all?

ravisantoshgudimetla: Valid point. I'm modified the test suite to include that scenario.

test/extended/operators/workload_guarantees.go
ravisantoshgudimetla: This test suite is focused on platform components and all of the platform components should run in a `openshift-*` namespace

ravisantoshgudimetla (commented):
@omertuc and @dhellmann - Any other concerns, if not can I get an LGTM?

omertuc:
> @omertuc and @dhellmann - Any other concerns, if not can I get an LGTM?

No concerns as far as my understanding of the issue goes (which is not a lot), and that's just on the SNO side too.

PR #26068, opened 2021-04-13
Bug 1949042: test/extended/router: skip h2 related tests on OpenStack

frobware:
Only run http/2, grpc-interop and h2spec tests on platforms where the
default router is exposed by a load balancer service that can also
support http/2.

### DIFF
--- a/test/extended/router/http2.go
+++ b/test/extended/router/http2.go
@@ -90,8 +90,7 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 		g.It("should pass the http2 tests", func() {
 			infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
 			o.Expect(err).NotTo(o.HaveOccurred(), "failed to get cluster-wide infrastructure")
-			switch infra.Status.PlatformStatus.Type {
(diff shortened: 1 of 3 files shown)


test/extended/router/http2.go
```diff
@@ -378,3 +377,15 @@ func resolveHostAsAddress(oc *exutil.CLI, interval, timeout time.Duration, host,
 
 	return result, nil
 }
+
+// platformHasHTTP2LoadBalancerService returns true where the default
+// router is exposed by a load balancer service and can support http/2
+// clients.
+func platformHasHTTP2LoadBalancerService(platformType configv1.PlatformType) bool {
+	switch platformType {
+	case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.GCPPlatformType:
```
candita, line 386: This means we'll never test any new platforms unless we remember to add them here.  Isn't it better to be specific for the platforms we know will not need the test, to prevent silently passing new platforms that we should have tested?

frobware: > This means we'll never test any new platforms unless we remember to add them here. Isn't it better to be specific for the platforms we know will not need the test, to prevent silently passing new platforms that we should have tested?

I've raised https://bugzilla.redhat.com/show_bug.cgi?id=1949154 to cover this. Pragmatically I'm just trying to avoid CI failures at the moment.

sgreene570: The only platform that supports Load Balancer services at the moment is IBM cloud, right?

PR #26071, opened 2021-04-13
Add new E2E test for [SDN-1364] ACL audit logging

astoycos:
This works adds a networking extended test for the new [network
policy audit logging feature ](https://issues.redhat.com/browse/SDN-1364)

The name of the new test is:

`"[sig-network] acl-logging should activate acl-logging [Suite:openshift/conformance/parallel]"`

It first activates the audit logging by applying new annotations to the `acl-logging` ns 

```
ns.Annotations["k8s.ovn.org/acl-logging"] = `{ "deny": "alert", "allow": "alert" }`
```

Next it make's three pods, two(pod[0] and pod[1]) in the ns `acl-logging` and one(pod[3]) in the ns `acl-logging-off` 

Then it makes two networkPolicies in the `acl-logging` ns -> 
1. allow-same-namespace
2. default-deny-all 

Finally it sends packets from pods 1 and 2 to pod 0..... The packet from 
pod[1] will be allowed by the allow-same-namespace policy and the packet 
from pod[2] will be dropped by the default-deny-all policy. This info will be 
captured  by the acl-audit-logs. 

The logs are then collected and verified with an `oc adm node-logs`

### DIFF
--- a/test/extended/networking/acl_audit_log.go
+++ b/test/extended/networking/acl_audit_log.go
@@ -1,0 +1,270 @@
+package networking
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	networkingv1 "k8s.io/api/networking/v1"
+
+	kapiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	maxPokeRetrys = 15
+	retryInterval = 1 * time.Second
+)
+
+var _ = Describe("[sig-network][Feature:Network Policy Audit logging]", func() {
+	var oc *exutil.CLI
+	var ns []string
+	var auditOut string
+	// this hook must be registered before the framework namespace teardown
+	// hook
+	AfterEach(func() {
+		if CurrentGinkgoTestDescription().Failed {
+			// If test fails dump test pods logs
+			exutil.DumpPodLogsStartingWithInNamespace("acl-logging", ns[0], oc.AsAdmin())
+			exutil.DumpPodLogsStartingWithInNamespace("acl-logging", ns[1], oc.AsAdmin())
+			// Dump what audit logs looked like if test failed
+			e2e.Logf("Audit logs are incorrect:\n %v", auditOut)
+		}
+	})
+
+	oc = exutil.NewCLI("acl-logging")
+
+	// The OVNKubernetes subnet plugin should allow acl_logging for network policy.
+	// For Openshift SDN and third party plugins, the behavior is unspecified and we should not run either test.
+	InOVNKubernetesContext(
+		func() {
+			f := oc.KubeFramework()
+			It("should ensure acl logs are created and correct", func() {
+				ns = append(ns, f.Namespace.Name)
+				makeNamespaceScheduleToAllNodes(f)
+				makeNamespaceACLLoggingEnabled(oc, f.Namespace)
+
+				nsNoACLLog := oc.SetupNamespace()
+				By("making namespace " + nsNoACLLog + " with acl-logging disabled")
+				ns = append(ns, nsNoACLLog)
+
+				testACLLogging(f, oc, ns, auditOut)
+			})
+		},
+	)
+})
+
+func makeNamespaceACLLoggingEnabled(oc *exutil.CLI, ns *kapiv1.Namespace) {
+	By("setting the k8s.ovn.org/acl-logging annotation for the namespace: " + oc.KubeFramework().Namespace.Name)
+	var err error
+	if ns.Annotations == nil {
+		ns.Annotations = make(map[string]string, 1)
+	}
+	ns.Annotations["k8s.ovn.org/acl-logging"] = `{ "deny": "alert", "allow": "alert" }`
+	_, err = oc.AdminKubeClient().CoreV1().Namespaces().Update(context.Background(), ns, metav1.UpdateOptions{})
+	expectNoError(err)
+}
+
+// Test the Network policy audit logging feature
+func testACLLogging(f *e2e.Framework, oc *exutil.CLI, ns []string, auditOut string) {
+	// We launch 3 pods total; pod[0] and pod[1] will end up on node[0] in ns "acl-logging-on" , and pod[2]
+	// will end up on node[1] in ns "acl-logging off", to know which acl-logging container to look in
+	var nodes [2]*kapiv1.Node
+	var pods [3]string
+	var ips []string
+	var err error
+	var ipv6 bool
+
+	nodes[0], nodes[1], err = findAppropriateNodes(f, DIFFERENT_NODE)
+	expectNoError(err)
+
+	By("making two pods in the acl-logging-on namespace and one in the acl-logging-off namespace")
+	// make the first two pods in ns acl-logging-on on node[0] and the other in acl-logging-off on node[1]
+	for i := range pods {
+		pods[i] = fmt.Sprintf("acl-logging-%d", i+1)
+		testPod := e2epod.NewAgnhostPod(ns[i/2], pods[i], nil, nil, nil, "netexec", "--http-port=80", "--udp-port=90")
+		testPod.Spec.NodeName = nodes[i/2].Name
+		_, err := f.ClientSet.CoreV1().Pods(ns[i/2]).Create(context.TODO(), testPod, metav1.CreateOptions{})
+		expectNoError(err)
+	}
+
+	// make sure pods come up
+	for i := range pods {
+		tempIp, err := waitForACLLoggingPod(f, ns[i/2], pods[i])
+		expectNoError(err)
+
+		ips = append(ips, tempIp)
+	}
+
+	// Check if pods are have ipv6 addr
+	if isIpv6(ips) {
+		ipv6 = true
+	}
+
+	By("making \"default deny\" and \"allow from same namespace\" network policies")
+	// make two network policies, a default deny all and allow from same namespace in ns "acl-logging-on"
+	_, err = makeDenyAllPolicy(f, ns[0])
+	expectNoError(err)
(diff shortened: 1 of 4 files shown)


astoycos:
/assign @squeed 


Did I do this workflow right? (i.e opening the PR in `openshift/k8s` in order to ensure we don't run the test for the default openshift-sdn plugin)

test/extended/networking/acl_audit_log.go
```diff
@@ -0,0 +1,213 @@
...
+	oc := exutil.NewCLI("acl-logging")
+
+	// The OVNKubernetes subnet plugin should allow acl_logging for network policy.
+	// For third party plugins, the behavior is unspecified and we should not run either test.
+	InPluginContext([]string{},
+		func() {
+			It("should activate acl-logging", func() {
+				f := oc.KubeFramework()
+				makeNamespaceACLLoggingEnabled(oc, f.Namespace)
+				Expect(testACLLogging(f, oc)).To(Succeed())
```
squeed, line 30: Don't Expect() a whole test, it will swallow all context / stack frames and make debugging difficult. Instead, structure your tests with nested Describe() and It().

For a reasonable test structure, check out the router or prometheus tests.

test/extended/networking/acl_audit_log.go
```diff
@@ -0,0 +1,213 @@
...
+
+		if allowLogFound && denyLogFound {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("Acl logs exist but do not contain correct src dst ips and actions")
+}
+
+func launchACLLoggingPod(f *e2e.Framework, namespace string, nodeName string, podName string) error {
```
squeed, line 132: We have access to the whole k8s library here; surely something must provide this function already....

test/extended/networking/acl_audit_log.go
```diff
@@ -0,0 +1,213 @@
...
+			}},
+		},
+	}
+
+	policy, err := f.ClientSet.NetworkingV1().NetworkPolicies(f.Namespace.Name).Create(context.TODO(), policy, metav1.CreateOptions{})
+
+	return policy, err
+}
+
+func cleanupNetworkPolicy(f *e2e.Framework, policy *networkingv1.NetworkPolicy) {
```
squeed, line 208: .do you need to do this? they'll go away when the Namespace is deleted

test/extended/networking/acl_audit_log.go
```diff
@@ -0,0 +1,213 @@
...
+// Test the Network policy audit logging feature
+func testACLLogging(f *e2e.Framework, oc *exutil.CLI) error {
+	// We launch 3 pods total; pod[0] and pod[1] will end up on node[0] in ns "acl-logging-on" , and pod[2]
+	// will end up on node[1] in ns "acl-logging off", to know which acl-logging container to look in
+	var nodes [2]*kapiv1.Node
+	var ns [2]string
+	var err error
+
+	makeNamespaceScheduleToAllNodes(f)
+	nsNoACLLog, err := f.CreateNamespace("acl-logging-off", map[string]string{})
```
squeed, line 55: don't hardcode namespace names; use `c.SetupNamespace()`

squeed:
The test seems reasonable as written; just needs to be structured more like an Openshift / kube test.

squeed:
Easiest way to skip this is in code; just retrieve the network.config.openshift.io object and check the configured plugin.

test/extended/networking/acl_audit_log.go
```diff
@@ -0,0 +1,213 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("[sig-network] acl-logging", func() {
+	oc := exutil.NewCLI("acl-logging")
+
+	// The OVNKubernetes subnet plugin should allow acl_logging for network policy.
+	// For third party plugins, the behavior is unspecified and we should not run either test.
+	InPluginContext([]string{},
```
danwinship, line 25: `InPluginContext` is a hack for running under some openshift-sdn Modes but not others. It's not what you want here, though you might want to add a new `InOVNKubernetesContext` that works similarly but checks the `network.config.openshift.io` object rather than the old sdn `ClusterNetwork` object

astoycos: Ack that makes much more sense, thanks!

test/extended/networking/acl_audit_log.go
astoycos: Unfortunately There isn't a default way to make a pod with the agnhost netexec image for linux, since `framework.pod.createpod` uses the busybox image by default instead.
 This is similar to what is done in the upstream network policy tests where they first 

create the spec manually 
https://github.com/kubernetes/kubernetes/blob/v1.21.1/test/e2e/network/netpol/model.go#L259

And then manually make the Pod 
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/netpol/kubemanager.go#L179

Which is why I do the same thing here

test/extended/networking/acl_audit_log.go
astoycos: Ack sounds good

test/extended/networking/acl_audit_log.go
astoycos: Nope removed :)

test/extended/networking/acl_audit_log.go
astoycos: ack

astoycos:
@danwinship If you could give this another look I'd appreciate it

test/extended/networking/acl_audit_log.go
astoycos: Arg I might have been wrong here, Let me try to use https://pkg.go.dev/k8s.io/kubernetes/test/e2e/framework/pod#NewAgnhostPod

test/extended/networking/acl_audit_log.go
```diff
@@ -0,0 +1,188 @@
...
+	expectNoError(err)
+
+	_, err = makeFromSameNSPolicy(f, ns[0])
+	expectNoError(err)
+	By("sending allowed traffic between pods in same namespace")
+	// Ping pod[0] from pod[1] which should succeed since it's in the same namespace with out == ip[0]
+	// Should hit the `allow-same-namespace` networkpoliocy
+	out, _, err := oc.AsAdmin().Run("exec").Args(pod[1], "-n", ns[0], "--", "curl", "-m", "1", ip[0]+"/clientip").Outputs()
+	if !strings.Contains(out, ip[1]) || err != nil {
+		expectNoError(err)
```
squeed, line 99: if the first conditional is true here, this will proceed since `err` is nil.

test/extended/networking/acl_audit_log.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	By("sending denied traffic between pods in different namespaces")
```
squeed, line 102: Is there any chance that policy implementation will be delayed here? Should we retry the deny-case if it connection inadvertently succeeds?

test/extended/networking/util.go
```diff
@@ -157,14 +183,14 @@ func pluginIsolatesNamespaces() bool {
...
-	return networkPluginName() == networkutils.MultiTenantPluginName
+	return openshiftSDNMode() == networkutils.MultiTenantPluginName
 }
 
 func pluginImplementsNetworkPolicy() bool {
 	switch {
 	case os.Getenv("NETWORKING_E2E_NETWORKPOLICY") == "true":
 		return true
-	case networkPluginName() == networkutils.NetworkPolicyPluginName:
+	case networkPluginName() == openshiftSDNPluginName || networkPluginName() == OVNKubernetesPluginName:
```
squeed, line 193: you also need to check that the openshift-sdn mode is set to policy.

squeed:
The test is good, but needs to be structured a bit differently to ensure it tolerates delays in ovn-kubernetes (and thus doesn't become a flake-fest). ovn sadly sometimes falls behind during load or restarts / upgrades.

I'd recommend doing two things:

1. looping until an expected-to-fail connection does indeed fail
2. looping the ping-and-check-logs step until the expected log line shows up.

Those should work around any asynchronicity.

test/extended/networking/util.go
astoycos: Ack got it

test/extended/networking/acl_audit_log.go
astoycos: Got it

astoycos:
I fixed this concern by ensuring we retest both pings (allow and deny) until the results are as expected, I don't think we need to provide a retry mechanism for the log parsing, since if the traffic is flowing as expected (i.e the OVN ACLs are there) then we should always be seeing the correct logs

test/extended/networking/acl_audit_log.go
astoycos: ack added retry to wait until traffic is flowing as expected

astoycos:
` [sig-network][Feature:Network Policy Audit logging] when using openshift ovn-kubernetes should ensure acl logs are created and correct [Suite:openshift/conformance/parallel]` 

Is now passing on

 `ci/prow/e2e-metal-ipi-ovn-dualstack`
`ci/prow/e2e-metal-ipi-ovn-ipv6`
`ci/prow/e2e-aws-ovn`

astoycos:
retesting those tests for further verification

astoycos:
Otherwise this test should be ready to go

squeed:
@knobunc could you approve this (it touches some helper code)

astoycos:
@squeed I lost the lgtm because I saw a flake on the recent run, and added some better "on failure" logic to see what was going on

astoycos:
Most recent changes should fix the flake (sorry I lost the LGTM again)

astoycos:
Letting this curn in CI for a bit, I will revisit late this afternoon (also looks like I need an accompanying bug for this)

astoycos:
I am seeing the test pass without flakes in 

```
ci/prow/e2e-metal-ipi-ovn-dualstack
ci/prow/e2e-metal-ipi-ovn-ipv6
ci/prow/e2e-aws-ovn
```

astoycos:
@knobunc This is just retesting on non-required tests for some reason.....

knobunc:
ci/prow/e2e-aws-serial

astoycos:
@knobunc Still need an override on `ci/prow/e2e-gcp-builds` :(

knobunc:
/skip
(To reset the required tests)

PR #26078, opened 2021-04-14
Set a minimum width to e2e-intervals chart by default

smarterclayton:
Also fix a makefile issue


smarterclayton:
Passed, no impact to other tests.

PR #26080, opened 2021-04-14
ensure that sampled conditions always have a range of time in their event interval

deads2k:
We don't know exactly how long a single sample condition lasted, but given our sampling interval, we can assume it lasted at least one sample interval and that is better than zero duration.


deads2k:
this updates the event monitor.  adding valid bug

PR #26081, opened 2021-04-15
Bug 1947809: upgrade: register none suite

vrutkovs:
Previously skip rules for None suite (no additional tests to run during upgrade) were added in https://github.com/openshift/origin/pull/26061. This commits ensures "none" suite is properly declared.


smarterclayton:
lgtm as well

PR #26083, opened 2021-04-15
Bug 1949978: test/extended/router: h2spec increase timeout and retry tests

frobware:
Increase the dial timeout and retry the tests if they fail.

### DIFF
--- a/test/extended/router/h2spec.go
+++ b/test/extended/router/h2spec.go
@@ -23,7 +23,7 @@ import (
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 )
 
-const h2specDialTimeoutInSeconds = 15
+const h2specDialTimeoutInSeconds = 30
 
 type h2specFailingTest struct {
 	TestCase   *h2spec.JUnitTestCase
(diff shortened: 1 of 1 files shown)


test/extended/router/h2spec.go
```diff
@@ -214,39 +214,54 @@ func failingTests(testSuites []*h2spec.JUnitTestSuite) []h2specFailingTest {
...
-		return nil, errors.New("zero length results file")
-	}
+		// h2spec will exit with non-zero if _any_ test in the suite
+		// fails, or if there is a dial timeout, so we log the
+		// error. But if we can fetch the results and if we can decode the
+		// results and we have > 0 test suites from the decoded
+		// results then assume the test ran.
+		output, err := e2e.RunHostCmd(oc.Namespace(), podName, h2specCommand(h2specDialTimeoutInSeconds, host, outputFile))
+		if err != nil {
+			e2e.Logf("error running h2spec: %v, but checking on result content", err)
```
candita, line 233: If there is an error, just try again.  Or if there is anything in the `output`, how about including `output` in the error log and start over?

frobware: We don't `return false, nil`, because although the exit code may be non-zero the tests [should] have run. The steps to determine overall failure is an empty results file which is what seems to flake in CI. Does that answer your concern?

PR #26087, opened 2021-04-17
Bug 1938467: test/extended/operators/resources: Autoscaler sets requests

wking:
openshift/cluster-autoscaler-operator#197 and openshift/cluster-autoscaler-operator#201 have landed.


JoelSpeed (commented):
/lgtm

PR #26089, opened 2021-04-19
Bug 1949978: test/extended/router: skip http2, grpc-interop and h2spec on AWS

frobware:
AWS/UPI jobs are failing on http2, grpc-interop and h2spec tests
because deleting the ingresscontroller which was explicitly created
for the test takes more than 10 minutes[1] and that (plus the overall
test time) exceeds the amount of time the test is allowed to run for
before being marked as failing.

This commit skips tests on AWS until we fix BZ #1912413.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1912413.

### DIFF
--- a/test/extended/router/http2.go
+++ b/test/extended/router/http2.go
@@ -389,8 +389,11 @@ func resolveHostAsAddress(oc *exutil.CLI, interval, timeout time.Duration, host,
 // clients.
 func platformHasHTTP2LoadBalancerService(platformType configv1.PlatformType) bool {
 	switch platformType {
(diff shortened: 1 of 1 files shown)


Miciah:
[BZ#1912413](https://bugzilla.redhat.com/show_bug.cgi?id=1912413) (cited in the PR description, commit message, and code comment) is for UPI, so it doesn't seem like the correct BZ.  Presumably [BZ#1949978](https://bugzilla.redhat.com/show_bug.cgi?id=1949978) (cited in the PR title) is the correct reference.

PR #26091, opened 2021-04-20
Bug 1848408: fix validation issues in test files

soltysh:
/assign @adambkaplan

### DIFF
--- a/test/extended/builds/run_fs_verification.go
+++ b/test/extended/builds/run_fs_verification.go
@@ -23,11 +23,11 @@ metadata:
 spec:
   runPolicy: Serial
   source:
+    type: Dockerfile
     dockerfile: |-
       FROM %s
       RUN chmod -R uga+rwx /run
       USER 1001
-  type: Dockerfile
   strategy:
     dockerStrategy:
       env:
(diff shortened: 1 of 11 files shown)


test/extended/builds/run_fs_verification.go
```diff
@@ -23,11 +23,11 @@ metadata:
 spec:
   runPolicy: Serial
   source:
+  	type: Dockerfile
```
adambkaplan, line 26: four spaces instead of two? Should align with `dockerfile` below.

test/extended/builds/run_fs_verification.go
```diff
@@ -44,11 +44,11 @@ metadata:
...
```
adambkaplan, line 47: four spaces instead of two? Should align with `dockerfile` below.

adambkaplan (changes requested):
YAML nits, otherwise looks good.

test/extended/builds/run_fs_verification.go
soltysh: tab vs space :sweat_smile: should be fixed with latest push

adambkaplan (approved):
/lgtm

PR #26034, opened 2021-03-30
visualize monitor event intervals

deads2k:
Create static e2echart HTML.  Something like this for instance.

Screenshot from 2021-03-30 12-56-00

PR #26092, opened 2021-04-20
monitor: mark the starting point for clusteroperators

deads2k:
We recently had a PR that made the operator perma-degraded and CI didn't show it.

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1073/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-serial/1384229570948894720

### DIFF
--- a/pkg/monitor/operator.go
+++ b/pkg/monitor/operator.go
@@ -17,45 +17,6 @@ import (
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 )
 
-// condition/Degraded status/True reason/DNSDegraded changed: DNS default is degraded
-func GetOperatorConditionStatus(message string) *configv1.ClusterOperatorStatusCondition {
-	if !strings.HasPrefix(message, "condition/") {
-		return nil
-	}
-	stanzas := strings.Split(message, " ")
-	condition := &configv1.ClusterOperatorStatusCondition{}
-
-	for _, stanza := range stanzas {
-		keyValue := strings.SplitN(stanza, "/", 2)
-		if len(keyValue) != 2 {
-			continue
-		}
-
-		switch keyValue[0] {
-		case "condition":
-			if condition.Type == "" {
-				condition.Type = configv1.ClusterStatusConditionType(keyValue[1])
-			}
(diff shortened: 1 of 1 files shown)


pkg/monitor/operator.go
```diff
@@ -76,11 +37,10 @@ func startClusterOperatorMonitoring(ctx context.Context, m Recorder, client conf
...
 				c := &co.Status.Conditions[i]
-				previous := findOperatorStatusCondition(oldCO.Status.Conditions, c.Type)
-				if previous == nil {
-					continue
-				}
-				if c.Status != previous.Status {
+				previousCondition := findOperatorStatusCondition(oldCO.Status.Conditions, c.Type)
+				// If we don't have a previous state, then we should always mark the starting state with an event.
+				// We recently had a PR that caused the kube-apiserver operator be permanently degraded and it didn't show up.
+				if previousCondition == nil || c.Status != previousCondition.Status {
```
tkashem, line 43: do we want to log here that previous condition was missing so when we go through the log we have more context?

tkashem: actually, we only log what we changed to, we don't log what we changed from. so this is in keeping with the current behavior.

deads2k: > actually, we only log what we changed to, we don't log what we changed from. so this is in keeping with the current behavior.

correct

pkg/monitor/operator.go
```diff
@@ -17,45 +17,6 @@ import (
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 )
 
-// condition/Degraded status/True reason/DNSDegraded changed: DNS default is degraded
-func GetOperatorConditionStatus(message string) *configv1.ClusterOperatorStatusCondition {
```
wking, line 21: bbd9949c834a14 (#26034) moved this to `pkg/monitor/monitorapi`.  :+1: to dropping this old, unused copy.

deads2k:
only affects the monitor, bypassing bug check

PR #26094, opened 2021-04-20
Bug 1951663: Fix volume loop bugs

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1951663


tsmetana:
The failures were caused by an error in hosted loki installation script which should be fixed now.
/retest

PR #26082, opened 2021-04-15
Bug 1949928: Remove references to EOL ImageStream Tags

adambkaplan:
Removing tests which reference the following EOL imagestream tags:

- nodejs 10, 10-ubi7
- perl 5.26-el7 5.26
- postgresql 10-el7 10
- ruby 2.5-ubi7 2.5

PR #26096, opened 2021-04-21
Bug 1949050: fix images.sh script

soltysh:
This is a followup to https://github.com/openshift/origin/pull/26082 which broke `images.sh` test.

/assign @adambkaplan


adambkaplan (approved):
/lgtm

PR #26103, opened 2021-04-26
test: Mark known alert failure for auth operator to bug

smarterclayton:
The bug is open and covers this for normal e2e and upgrades, so make
sure we exclude it until the team can fix the issue (degraded is
not currently causing other failures).

Ties https://bugzilla.redhat.com/show_bug.cgi?id=1939580 to the tests (I thought https://bugzilla.redhat.com/show_bug.cgi?id=1929922 was the issue but the former is more appropriate).


wking (approved):
/lgtm

PR #26107, opened 2021-04-27
test: Allow KubeAPIErrorBudgetBurn on GCP platforms only

smarterclayton:
The alert seems to fire on GCP in both upgrade and normal cases in 
https://bugzilla.redhat.com/show_bug.cgi?id=1953798, although the symptoms
described there should *also* happen on other clouds, so maybe we have a
general latency issue on GCP.  Disable only on GCP.

Also fix two issues:

test: Alert check should check all conditions

The alert was exiting early after only one matching condition, which means
it was overly aggressively matching the first label. We may have more bugs
after this.

Also add a new custom function to allow conditions to apply only in 
specific cases.

test: Disable upgrade disruption check on GCP

The constant is 'gce', not 'gcp', and the test is failing because of an as
yet undiagnosed issue.

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -8,6 +8,7 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	"github.com/prometheus/common/model"
 
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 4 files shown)


smarterclayton:
/retest

the alert we are suppressing did not fire

smarterclayton:
Merging to clear GCP flakes (10-20% of all runs on KubeAPIErrorBudgetBurn)

PR #26112, opened 2021-04-28
Limit scope of synthetic scc test for disruptive

marun:
Disruptive tests generate events that can trigger synthetic test failures. This change updates the scc synthetic test to ignore events created before the test started for suite types (like disruptive) that set the start time via an env var. Normal execution will be unaffected due to the start time defaulting to 0.

### DIFF
--- a/test/extended/authorization/scc.go
+++ b/test/extended/authorization/scc.go
@@ -30,7 +30,12 @@ var _ = g.Describe("[sig-auth][Feature:SCC][Early]", func() {
 		suiteStartTime := exutil.SuiteStartTime()
 		var preTestDenialStrings, duringTestDenialStrings []string
 
+		eventsAfterTime := exutil.LimitTestsToStartTime()
 		for _, event := range events.Items {
+			if event.LastTimestamp.Time.Before(eventsAfterTime) {
+				continue
+			}
+
 			if !strings.Contains(event.Message, "unable to validate against any security context constraint") {
 				continue
 			}


test/extended/authorization/scc.go
```diff
@@ -30,7 +30,12 @@ var _ = g.Describe("[sig-auth][Feature:SCC][Early]", func() {
 		suiteStartTime := exutil.SuiteStartTime()
 		var preTestDenialStrings, duringTestDenialStrings []string
 
+		eventsAfterTime := exutil.LimitTestsToStartTime()
```
deads2k, line 33: Reading the code around, it has been modified since I first authored it.  What is the difference between this and `preTestDenialStrings` and `duringTestDenialStrings`?

deads2k: Looking at failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25774/pull-ci-openshift-origin-master-e2e-aws-disruptive/1387235806111862784 , any chance they simply forgot to create their rolebinding?  Maybe they aren't reactive?  Perhaps checking creation time of the rolebinding against the events?

At the point when this test is running, is the cluster supposedly stable?

deads2k: > Reading the code around, it has been modified since I first authored it. What is the difference between this and `preTestDenialStrings` and `duringTestDenialStrings`?

Is the time on this bugged perhaps?  Maybe try logging that time first so you can see when the failures happened versus the suiteStartTime?

marun: Arg, this test is already checking the start time. This PR is not needed, the failure is legit indicating an issue with post-disruptive scc propagation.

marun: And now I see that the test is checking `TEST_SUITE_START_TIME`, and this PR is adding a check for `TEST_LIMIT_START_TIME`.

smarterclayton: SUITE_START_TIME is set by openshift-tests.  LIMIT_START_TIME is set in release only.

smarterclayton: synthetics should infer "start time" from the events they are passed.

marun: @smarterclayton I'm not sure how to act on your comments. Is this PR unnecessary, or is more work involved?

marun:
Superceded by https://github.com/openshift/origin/pull/26130

PR #26113, opened 2021-04-28
Bug 1949991: Ignore Catalog update pods in openshift/conformance/parallel

anik120:
The marketplace operator deploys default CatalogSources in the openshift-marketplace namespace.
These CatalogSource have polling feature, [eg](https://github.com/operator-framework/operator-marketplace/blob/master/defaults/01_redhat_operators.cr.yaml#L15-L16)
that poll for updates at regular intervals. In the begining of every interval, a new "update" pod
is [created](https://github.com/operator-framework/operator-lifecycle-manager/blob/d980191ecdb848c6538b721906dfc993e86edd6f/pkg/controller/registry/reconciler/grpc.go#L203) which is [removed immediately](https://github.com/operator-framework/operator-lifecycle-manager/blob/d980191ecdb848c6538b721906dfc993e86edd6f/pkg/controller/registry/reconciler/grpc.go#L297-L299)
if there is no difference between the image digest of the fist pod.

Example of activity in the openshift/marketplace namespace
```
$ k get pods -n openshift-marketplace -w

NAME                                   READY   STATUS    RESTARTS   AGE
community-operators-sbb4p              1/1     Running   0          46m
.
.
.
community-operators-6qtn4              0/1     Pending             0          0s
community-operators-6qtn4              0/1     Pending             0          0s
community-operators-6qtn4              0/1     ContainerCreating   0          0s
community-operators-6qtn4              0/1     ContainerCreating   0          3s
community-operators-6qtn4              0/1     Running             0          6s
community-operators-6qtn4              1/1     Running             0          10s
community-operators-6qtn4              1/1     Terminating         0          10s
community-operators-6qtn4              0/1     Terminating         0          12s
community-operators-6qtn4              0/1     Terminating         0          16s
community-operators-6qtn4              0/1     Terminating         0          16s
```

The `crashloopingContainerCheck` [sees the container terminated forcefully](https://github.com/openshift/origin/blob/e945cb88da780e21c021b6c8b430454bcfb881cf/test/extended/operators/cluster.go#L195),i.e with a non-zero exit code  and adds the update pod to the list of problematic pods.

Example status of a update pod that has been terminated:
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2021-04-28T22:27:55Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2021-04-28T22:28:07Z"
    message: 'containers with unready status: [registry-server]'
    reason: ContainersNotReady
    status: "False"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2021-04-28T22:28:07Z"
    message: 'containers with unready status: [registry-server]'
    reason: ContainersNotReady
    status: "False"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2021-04-28T22:27:55Z"
    status: "True"
    type: PodScheduled
  containerStatuses:
  - containerID: cri-o://db237acad8d46a14d57e6d2e165aa025a149a036c75aa648655c4dabe14bb151
    image: registry.redhat.io/redhat/community-operator-index:v4.8
    imageID: registry.redhat.io/redhat/community-operator-index@sha256:0d84d479552209a84831abc63c48de6f5955c228957a91c09fa26777c0dd9492
    lastState: {}
    name: registry-server
    ready: false
    restartCount: 0
    started: false
    state:
      terminated:
        containerID: cri-o://db237acad8d46a14d57e6d2e165aa025a149a036c75aa648655c4dabe14bb151
        exitCode: 2
        finishedAt: "2021-04-28T22:28:06Z"
        message: |
          time="2021-04-28T22:28:00Z" level=info msg="serving registry" database=/database/index.db port=50051
        reason: Error
        startedAt: "2021-04-28T22:28:00Z"
  hostIP: 10.0.128.4
  phase: Running
  podIP: 10.128.2.40
  podIPs:
  - ip: 10.128.2.40
  qosClass: Burstable
  startTime: "2021-04-28T22:27:55Z"
```
This PR updates the test to ignore all pods created due to polling of default CatalogSources.
CatalogSource update pods are identified with the following label:
```
labels:
    catalogsource.operators.coreos.com/update: community-operators
```

### DIFF
--- a/test/extended/operators/cluster.go
+++ b/test/extended/operators/cluster.go
@@ -191,6 +191,13 @@ func hasImagePullError(pod *corev1.Pod) bool {
 }
 
 func hasFailingContainer(pod *corev1.Pod) bool {
+	// Catalog update pods are deliberately terminated with non-zero exit codes, so this check does not apply to them
+	if _, ok := pod.Labels["catalogsource.operators.coreos.com/update"]; ok {
+		return false
+	}
+	if pod.Status.Phase == corev1.PodFailed {
+		return true
+	}
 	for _, status := range append(append([]corev1.ContainerStatus{}, pod.Status.InitContainerStatuses...), pod.Status.ContainerStatuses...) {
 		if status.State.Terminated != nil && status.State.Terminated.ExitCode != 0 {
 			pod.Status.Message = status.State.Terminated.Message
@@ -200,7 +207,7 @@ func hasFailingContainer(pod *corev1.Pod) bool {
 			return true
 		}
 	}
-	return pod.Status.Phase == corev1.PodFailed
+	return false
 }
 
 func isCrashLooping(pod *corev1.Pod) bool {


test/extended/operators/cluster.go
```diff
@@ -108,8 +108,10 @@ func crashloopingContainerCheck(podFilters ...podFilter) {
 				continue
 			}
 			key := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
-			names = append(names, key)
-			podsWithProblems[key] = pod
+			if _, ok := pod.Labels["catalogsource.operators.coreos.com/update"]; !ok {
```
bparees, line 111: doesn't this mean we'll lose the ability to detect crashlooping catalog pods?  i think we need a more surgical solution.

also i'm not clear why we're getting to this point.  If the pod is running, we should hit the default `continue` above, right?   So is it tripping one of the other cases?  which one? why?

anik120: > doesn't this mean we'll lose the ability to detect crashlooping catalog pods?

We have safety measures in place to protect ourselves from crashlooping update pods, see [this](https://github.com/operator-framework/operator-lifecycle-manager/blob/d980191ecdb848c6538b721906dfc993e86edd6f/pkg/controller/registry/reconciler/grpc.go#L266-L269).
Essentially, if the first catalog pod is running successfully, we don't sacrifice that for a potentially problematic update pod. 

> also i'm not clear why we're getting to this point. If the pod is running, we should hit the default continue above, right?

Apolizies for the confusion, it's hitting the `hasFailingContainer` case, updated my comment in the commit message.

bparees: 1) i think you should add this exemption in `hasFailingContainer` so that we can still catch the other cases if they happen
2) please include a comment in the code explaining why this exemption is needed for these pods so someone else who stumbles upon it doesn't have to dig into git history to understand why it's being done

bparees: > We have safety measures in place to protect ourselves from crashlooping update pods, see this.

that doesn't help us catch the fact that it's happening (raise a signal via a failing test case in our CI)

anik120: @bparees done, PTAL, thanks!

PR #26122, opened 2021-04-30
Bug 1957512: openshift-tests: fix a --provider --dry-run edge case

danwinship:
regression from #25818: Trying to override --provider while not actually connected to a cluster (eg, to look at the output of --dry-run) would fail, rather than just falling back to not trying to autodetect anything.


soltysh (approved):
/lgtm
/approve

PR #26124, opened 2021-04-30
Skip ClientSecretWithPlus test that uses oauth-openshift route for ibm clusters

Joseph-Goergen:


test/extended/oauth/client_secret.go
```diff
@@ -29,6 +33,10 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 
 	g.Describe("ClientSecretWithPlus", func() {
 		g.It(fmt.Sprintf("should create oauthclient"), func() {
+			if e2e.TestContext.Provider == ibmcloud.ProviderName {
+				e2eskipper.Skipf("IBM Cloud clusters do not contain oauth-openshift routes")
+			}
+
```
bparees, line 39: we have a different way of doing this:
https://github.com/openshift/origin/blob/52c153f8e8e11db52f665e7361a3b048dcbe95e0/test/extended/util/annotate/rules.go#L189

Joseph-Goergen:
https://bugzilla.redhat.com/show_bug.cgi?id=1973398

PR #26127, opened 2021-05-03
test: Associate ClusterOperatorDown for machine-config with bug

smarterclayton:
We have opened a high severity bug for this issue and the alert
can be tagged to allow merging.


smarterclayton:
Force merging as this is blocking a significant amount of passing runs

PR #26128, opened 2021-05-03
test/e2e/upgrade: Add manifest delete annotation test

jottofar:


test/e2e/upgrade/manifestdelete/manifest-delete.go
```diff
@@ -0,0 +1,187 @@
...
+	deletes = []object{}
+
+	// Update this array to contain unstructured objects which should have been deleted in this release. For example:
+	//	unstructured{name: "dns",
+	//		namespace: "openshift-dns-operator",
+	//		group:     "monitoring.coreos.com",
+	//		version:   "v1",
+	//		resource:  "prometheusrules"},
+	//
+	unstructuredDeletes = []unstructured{}
```
wking, line 60: can we fill some of these in now?  Doesn't block this PR, but if we have any ready, populating them now saves us a follow-up PR later.

test/e2e/upgrade/manifestdelete/manifest-delete.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (obj object) uniqueName() string {
+	if len(obj.namespace) == 0 {
+		return obj.name
+	}
+	return obj.namespace + "/" + obj.name
+}
+
+func (obj object) String() string {
+	return fmt.Sprintf("%s \"%s\"", obj.kind, obj.uniqueName())
```
wking, line 71: nit: `\"%s\"` -> `%q`.  Or just `%s`, because neither the namespace nor the name are allowed to contain confusing stuff like quotes or whitespace.

test/e2e/upgrade/manifestdelete/manifest-delete.go
jottofar: yes we can

test/e2e/upgrade/manifestdelete/manifest-delete.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (u unstructured) uniqueName() string {
+	if len(u.namespace) == 0 {
+		return u.name
+	}
+	return u.namespace + "/" + u.name
+}
+
+func (u unstructured) String() string {
+	return fmt.Sprintf("unstructured \"%s/%s, %s, %s\"", u.group, u.version, u.resource, u.uniqueName())
```
wking, line 82: I'd rather drop the `\"` here too.  I don't think any of the strings you're formatting are allowed to contain whitespace or quotes or other chars where explicitly quoting the block would improve readability.

test/e2e/upgrade/manifestdelete/manifest-delete.go
```diff
@@ -0,0 +1,187 @@
...
+// An alert firing during an upgrade is a high severity bug - it either points to a real issue in
+// a dependency, or a failure of the component, and therefore must be fixed.
+func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade upgrades.UpgradeType) {
+	if len(deletes) == 0 && len(unstructuredDeletes) == 0 {
+		framework.Logf("No object deletions in this release to verify")
+		return
+	}
+
+	// Block until upgrade is done
+	g.By("Waiting for upgrade to finish before checking for deletes")
```
wking, line 111: I'm a bit fuzzy in the origin/Ginkgo setup, but can we use `[sig-arch][Late] Verify...` above to get slotted in at the end, and then drop this block?  Not a big deal either way.

wking (approved):
/lgtm

test/e2e/upgrade/manifestdelete/manifest-delete.go
```diff
@@ -0,0 +1,192 @@
...
+	g.By("Setting up upgrade delete test")
+	oc := exutil.NewCLIWithFramework(f)
+	t.oc = oc
+	config, err := framework.LoadConfig()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	t.config = config
+	framework.Logf("Post-upgrade delete test setup complete")
+}
+
+// Test checks if alerts are firing at various points during upgrade.
```
deads2k, line 106: old comment?

test/e2e/upgrade/manifestdelete/manifest-delete.go
```diff
@@ -0,0 +1,192 @@
...
+
+var (
+	// Update this array to contain objects which should have been deleted in this release. For example:
+	//	object{kind: "Namespace",
+	//		namespace: "",
+	//		name:      "openshift-foobar"},
+	//
+	deletes = []object{}
+
+	// Update this array to contain unstructured objects which should have been deleted in this release. For example:
```
deads2k, line 52: is there something that prevents someone from using the delete annotation and not adding it to this test?  I don't think it would occur to me to modify this test when adding the annotation to my manifest in my operator.

test/e2e/upgrade/manifestdelete/manifest-delete.go
```diff
@@ -0,0 +1,192 @@
...
+	return u.namespace + "/" + u.name
+}
+
+func (u unstructured) String() string {
+	return fmt.Sprintf("unstructured %s/%s, %s, %s", u.group, u.version, u.resource, u.uniqueName())
+}
+
+func (UpgradeTest) Name() string { return "check-for-deletes" }
+func (UpgradeTest) DisplayName() string {
+	return "[sig-arch] Verify object deletions after upgrade success"
```
deads2k, line 92: This is related to the CVO behavior, not to sig-arch.  `[bz-Cluster Version Operator]` would match what we did for operators.

deads2k:
The test is a little unusual because I think it will drive, but properly attributed, I'm ok with adding it for the CVO.

test/e2e/upgrade/manifestdelete/manifest-delete.go
jottofar: fixed

test/e2e/upgrade/manifestdelete/manifest-delete.go
jottofar: OTA will most likely have to maintain

test/e2e/upgrade/manifestdelete/manifest-delete.go
jottofar: changed

PR #26129, opened 2021-05-04
Extends monitoring of Kube API servers by hitting NodeIPs directly

p0lyn0mial:
Already existing tests monitor the availability of the Kube API servers via the external load balancer.
The issue we see is when the tests fail we cannot tell if it was a misconfigured load balancer or a broken API server.

This PR adds new tests that simply reuse already existing monitors but use the IP addresses assigned to Servers/Nodes instead.

An alternative approach would be to run a separate monitor for each endpoint and using a lister to see if it is still on the list and silence if not.

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -426,7 +426,13 @@ func (rt *kubeAPIEndpointsAwareRT) RoundTrip(r *http.Request) (*http.Response, e
 	r.Host = nextEndpoint
 	r.URL.Host = nextEndpoint
 
-	return rt.delegate.RoundTrip(r)
+	resp, err := rt.delegate.RoundTrip(r)
+	if err == nil {
+		fmt.Println(fmt.Sprintf("status %v", resp.StatusCode))
+	} else {
+		fmt.Println(err)
+	}
+	return resp, err
 }
 
 func (rt *kubeAPIEndpointsAwareRT) nextEndpointLocked() (string, error) {
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"errors"
 
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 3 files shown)


p0lyn0mial:
/hold

for testing

pkg/monitor/api.go
```diff
@@ -203,6 +241,27 @@ func StartKubeAPIMonitoringWithConnectionReuse(ctx context.Context, m *Monitor,
 	return StartServerMonitoringWithConnectionReuse(ctx, m, clusterConfig, timeout, LocatorKubeAPIServerReusedConnection, "/api/v1/namespaces/default")
 }
 
+func StartKubeAPIMonitoringWithConnectionReuseForNodeIPs(ctx context.Context, m *Monitor, clusterConfig *rest.Config, timeout time.Duration) error {
+	client, err := kubernetes.NewForConfig(clusterConfig)
+	if err != nil {
+		return err
+	}
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(client, time.Second*30)
+	endpointsInformer := kubeInformerFactory.Core().V1().Endpoints()
```
smarterclayton, line 250: Watching all endpoints is expensive. Also, endpoints are going away soon.  I'd recommend you only watch in the default namespace

pkg/monitor/api.go
```diff
@@ -54,6 +60,9 @@ func Start(ctx context.Context) (*Monitor, error) {
 	if err := StartKubeAPIMonitoringWithNewConnections(ctx, m, clusterConfig, 5*time.Second); err != nil {
 		return nil, err
 	}
+	if err := StartKubeAPIMonitoringWithNewConnectionsForNodeIPs(ctx, m, clusterConfig, 5*time.Second); err != nil {
```
smarterclayton, line 63: This is the first monitor that needs to be conditional.  If you plan on running as a pod, we don't want to list / watch everything from inside the cluster (we're watching from outside).  You probably need to have a monitor specifically for this, and you should not be running this from outside the cluster.

A proxy connection through the cluster will be disrupted.

PR #26130, opened 2021-05-04
test: Suppress the SCC failure when a limit is present

smarterclayton:
The presence of the LimitSuiteToStartTime flag should both constrain
when the test starts AND flag pre-test failures as flakes. This is
used by disruptive jobs that do their own invariant testing and don't
want our generic tests to look too far back.

Mitigates https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/18229/rehearse-18229-periodic-ci-openshift-release-master-nightly-4.8-e2e-aws-upgrade-rollback-oldest-supported/1389336340990005248

### DIFF
--- a/test/extended/authorization/scc.go
+++ b/test/extended/authorization/scc.go
@@ -8,6 +8,7 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	"github.com/openshift/origin/pkg/test/ginkgo/result"
 	exutil "github.com/openshift/origin/test/extended/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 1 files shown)


marun:
e2e-metal-ipi-ovn-ipv6 is known broken and this PR has no effect on it - please override

smarterclayton:
The failure in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26130/pull-ci-openshift-origin-master-e2e-aws-disruptive/1389644021760004096 was *before* disruptive ran, so it's the existing bug (which actually needs to get prioritized, this is happening a lot more in the last week so it likely regressed)

smarterclayton:
Force merging to allow disruptive to go on other repos (this will fix disruptive -> conformance, which before would always fail on this).

PR #26137, opened 2021-05-07
WIP: Remove jenkins from payload

akram:


yselkowitz:
@akram you need to run `make update` and add the resulting diff to this PR.

yselkowitz:
Also, before the make update, is there any testdata which is used exclusively by these tests that can also be removed?

adambkaplan (commented):
/hold

This should be retitled, as "removing from payload" is misleading. We won't remove Jenkins from the payload until we get Jenkins out of the ART pipeline and into its own CPaaS build pipeline.

This removes Jenkins from origin tests and examples. This PR should be held until equivalent tests are set up in the openshift/jenkins repo.

PR #26138, opened 2021-05-07
Bug 1958094: audit-logs: allow .lock file and flake on lock.log

sttts:
Prereq for https://github.com/openshift/cluster-kube-apiserver-operator/pull/1128.

### DIFF
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -162,7 +162,7 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 
 		tempDir, err := ioutil.TempDir("", "test.oc-adm-must-gather.")
 		o.Expect(err).NotTo(o.HaveOccurred())
-		// defer os.RemoveAll(tempDir)
+		defer os.RemoveAll(tempDir)
 
 		args := []string{
 			"--dest-dir", tempDir,


test/extended/cli/mustgather.go
```diff
@@ -200,7 +201,17 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 						return nil
 					}
 
-					if (strings.Contains(path, "-termination-") && strings.HasSuffix(path, ".log.gz")) || strings.HasSuffix(path, "termination.log.gz") {
+					if path == "lock.log" {
+						apiserver := filepath.Base(auditDirectory)
+						lockLog, err := ioutil.ReadFile(path)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						result.Flakef("kubelet launched %s without waiting for the old process to terminate (lock was still hold): \n\n%s", apiserver, string(lockLog))
```
p0lyn0mial, line 208: you don't want to have a separate test for that?
This test is testing `must-gather` not kas's termination

sttts: done

test/extended/cli/mustgather.go
```diff
@@ -285,6 +290,59 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+		g.Describe("[sig-node] kubelet", func() {
+			g.It("runs apiserver processes strictly sequentially in order to not risk audit log corruption", func() {
+				// On IBM ROKS, events will not be part of the output, since audit logs do not include control plane logs.
+				if e2e.TestContext.Provider == ibmcloud.ProviderName {
+					g.Skip("ROKs doesn't have audit logs")
+				}
+
+				tempDir, err := ioutil.TempDir("", "test.oc-adm-must-gather.")
+				o.Expect(err).NotTo(o.HaveOccurred())
+				// defer os.RemoveAll(tempDir)
```
p0lyn0mial, line 304: should we clean up the tmp dir?

test/extended/cli/mustgather.go
```diff
@@ -285,6 +290,59 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+					err := filepath.Walk(filepath.Join(pluginOutputDir, "audit_logs", apiserver), func(path string, info os.FileInfo, err error) error {
+						g.By(path)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						if info.IsDir() {
+							return nil
+						}
+
+						seen.Insert(apiserver)
+
+						if path != "lock.log" {
```
p0lyn0mial, line 328: Shouldn't we extract the file name from the path?
Otherwise this condition will always be true since the path doesn't contain only the file name.

test/extended/cli/mustgather.go
sttts: Cargo cult. Copied that.

sttts: Fixed

PR #26139, opened 2021-05-07
must-gather: add oauth-apiserver to expected audit logs

sttts:


test/extended/cli/mustgather.go
```diff
@@ -177,11 +177,13 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 		expectedDirectoriesToExpectedCount := map[string]int{
 			path.Join(pluginOutputDir, "audit_logs", "kube-apiserver"):      1000,
 			path.Join(pluginOutputDir, "audit_logs", "openshift-apiserver"): 10, // openshift apiservers don't necessarily get much traffic.  Especially early in a run
+			path.Join(pluginOutputDir, "audit_logs", "oauth-apiserver"):     10, // oauth apiservers don't necessarily get much traffic.  Especially early in a run
```
p0lyn0mial, line 180: Why do we count the number of audit events? This seems to be error-prone. 
Checking if the audit logs contain some events is not enough?

sttts: that's pre-existing. I bet this is @deads2k 's code? @deads2k any idea?

PR #26140, opened 2021-05-07
Bug 1956989: initialize framework.TestContext.IPFamily correctly

fedepaol:
Some tests rely on the TestContext IPFamily field, that is being initialized in the upstream ginkgo suites but not in origin's one.
Here we add the initialization in InitTest so the test find the right value.

### DIFF
--- a/cmd/openshift-tests/provider.go
+++ b/cmd/openshift-tests/provider.go
@@ -61,6 +61,12 @@ func initializeTestFramework(context *e2e.TestContextType, config *exutilcluster
 
 	e2e.AfterReadingAllFlags(context)
 	context.DumpLogsOnFailure = true
+
+	// these constants are taken from kube e2e and used by tests
+	context.IPFamily = "ipv4"
+	if config.HasIPv6 && !config.HasIPv4 {
+		context.IPFamily = "ipv6"
+	}
 	return nil
 }
 
--- a/test/extended/util/annotate/rules.go
+++ b/test/extended/util/annotate/rules.go
@@ -51,12 +51,6 @@ var (
 			// https://bugzilla.redhat.com/show_bug.cgi?id=1953478
 			`\[sig-storage\] Dynamic Provisioning Invalid AWS KMS key should report an error and create no PV`,
 
-			// https://bugzilla.redhat.com/show_bug.cgi?id=1956989
-			`\[sig-network\] Services should be possible to connect to a service via ExternalIP when the external IP is not assigned to a node`,
-			`\[sig-network\] Networking IPerf2 \[Feature:Networking-Performance\] should run iperf2`,
-			`\[sig-network\] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol`,
-			`\[sig-network\] Networking should provide Internet connection for containers`,
-
 			// https://bugzilla.redhat.com/show_bug.cgi?id=1957894
 			`\[sig-node\] Container Runtime blackbox test when running a container with a new image should be able to pull from private registry with secret`,
 		},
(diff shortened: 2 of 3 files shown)


test/extended/util/test.go
```diff
@@ -81,6 +83,11 @@ func InitTest(dryRun bool) error {
 			return err
 		}
 		TestContext.Host = cfg.Host
+		c, err := kclientset.NewForConfig(cfg)
+		if err != nil {
+			klog.Fatal("Error loading client: ", err)
+		}
+		TestContext.IPFamily = getDefaultClusterIPFamily(c)
```
danwinship, line 90: We don't need to call `getDefaultClusterIPFamily()`; we already know whether the cluster is ipv4/ipv6/dual-stack based on reading the OCP network config. You can set `context.IPFamily` in `initializeTestFramework` in `cmd/openshift-tests/provider.go` based on `config.HasIPv4`/`config.HasIPv6`.

test/extended/util/test.go
```diff
@@ -94,6 +101,8 @@ func InitTest(dryRun bool) error {
 	TestContext.CreateTestingNS = createTestingNS
 
 	klog.V(2).Infof("Extended test version %s", version.Get().String())
+
+	e2e.Logf("Cluster IP family: %s", e2e.TestContext.IPFamily)
```
danwinship, line 105: the upstream tests do this but it's basically spam. just kill that

danwinship (changes requested):

test/extended/util/test.go
fedepaol: Cool, this will avoid the need of hitting the apis since we are basing this on config.

fedepaol: Done, I am not totally sure the way I used hasIPv4 / v6 has the same semantic as checking the ipfamily of the primary clusterip (especially in case of a dual stack cluster).

test/extended/util/test.go
```diff
@@ -94,6 +94,7 @@ func InitTest(dryRun bool) error {
 	TestContext.CreateTestingNS = createTestingNS
 
 	klog.V(2).Infof("Extended test version %s", version.Get().String())
+
```
dcbw, line 97: @fedepaol spurious change? might as well just remove this hunk.

cmd/openshift-tests/provider.go
```diff
@@ -61,6 +61,11 @@ func initializeTestFramework(context *e2e.TestContextType, config *exutilcluster
 
 	e2e.AfterReadingAllFlags(context)
 	context.DumpLogsOnFailure = true
+
+	context.IPFamily = "ipv4"
+	if config.HasIPv6 {
+		context.IPFamily = "ipv6"
+	}
```
danwinship, line 68: hm... I think I'd go with `if config.HasIPv6 && !config.HasIPv4`
also maybe add a comment pointing out that these strings come from kube e2e and are not equal to `corev1.IPv4Protocol` / `corev1.IPv6Protocol`

danwinship:
/retitle Bug 1956989: initialize framework.TestContext.IPFamily correctly

soltysh:
@fedepaol fyi, I'm disabling one another test in https://github.com/openshift/origin/pull/26178:

` [sig-network] Networking should provide Internet connection for containers`

I commented in the BZ too.

fedepaol:
> @fedepaol fyi, I'm disabling one another test in #26178:
> 
> ` [sig-network] Networking should provide Internet connection for containers`
> 
> I commented in the BZ too.

Ack, thanks. I'll check if it's (hopefully) fixed by this very same change and in case I'll rebase and un-skip the test here once 26178 merges

test/extended/util/annotate/rules.go
```diff
@@ -182,6 +176,7 @@ var (
 			`\[sig-network\]\[Feature:Router\] The HAProxy router should serve the correct routes when scoped to a single namespace and label set`,
 			`\[sig-network\]\[Feature:Router\] The HAProxy router should set Forwarded headers appropriately`,
 			`\[sig-network\]\[Feature:Router\] The HAProxy router should support reencrypt to services backed by a serving certificate automatically`,
+			`\[sig-network\] Networking should provide Internet connection for containers \[Feature:Networking-IPv6\]`, // ipv6 only clusters do not have external connectivity
```
danwinship, line 179: Drop the comment; that's not an intrinsic fact about IPv6, it's just a fact about the current e2e-metal-ipi-ipv6.

(It's correct to add this test to `[Skipped:Disconnected]`, because the test does require external connectivity and thus needs to be skipped if you don't have external connectivity. It's just wrong to claim that the skipping is somehow related to the IPv6-ness rather than the lack-of-external-connectivity-ness.)

danwinship (changes requested):

test/extended/util/annotate/rules.go
fedepaol: Yeah, that's not what I meant to say.
Will remove the comment

fedepaol: @danwinship done, sorry for the delay

PR #26141, opened 2021-05-07
Bug 1992596: e2e/cli: move half of builds.sh to proper e2e

atiratree:
I found there is a bit of overlap with `[sig-builds]` as they are using oc cli to create builds as well. But it does not test the CLI part thourougly. I am not entirerly sure if it makes sense to combine `sig-builds` with `sig-cli`, but I decided not to at this moment.


test/extended/cli/builds.go
```diff
@@ -0,0 +1,46 @@
...
+
+		g.By("build from a binary with inputs creates a binary build")
+		out, err = oc.Run("new-build").Args("--binary", "--name=binary-test").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.ContainSubstring("buildconfig.build.openshift.io \"binary-test\" created"))
+		out, err = oc.Run("get").Args("bc/binary-test").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.ContainSubstring("Binary"))
+		err = oc.Run("delete").Args("bc/binary-test", "is/binary-test").Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
```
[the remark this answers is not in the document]
atiratree: I went with the function for now - we can change it later

test/extended/cli/builds.go
```diff
@@ -0,0 +1,46 @@
...
+
+	g.It("validates builds and build related commands", func() {
+		g.By("build from Dockerfile with output to ImageStreamTag")
+		out, err := oc.Run("new-build").Args("--to=tests:custom", "--dockerfile", testDockerFile).Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.ContainSubstring("buildconfig.build.openshift.io \"tests\" created"))
+		out, err = oc.Run("get").Args("bc/tests", "--template", bcTemplate).Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.Equal("ImageStreamTag tests:custom"))
+		err = oc.Run("delete").Args("bc/tests", "is/tests").Execute()
```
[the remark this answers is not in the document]
atiratree: resolved

test/extended/cli/builds.go
```diff
@@ -0,0 +1,179 @@
...
+	err = oc.Run("delete").Args(resources...).Execute()
+	o.Expect(err).NotTo(o.HaveOccurred())
+}
+
+var _ = g.Describe("[sig-cli] oc builds", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLI("oc-builds").AsAdmin()
+
+	g.It("new-build", func() {
```
ravisantoshgudimetla, line 77: I'd rather have a `g.It` instead of simple `g.By` to explicitly call out the functionality we are testing.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (r *expectedBuild) setDefaults() {
+	if r.outputType == "ImageStreamTag" && r.isName == "" {
+		r.isName = r.bcName
+	}
+}
+
+func testNewBuild(oc *exutil.CLI, opts expectedBuild) {
+	opts.setDefaults()
```
ravisantoshgudimetla, line 32: We can move this function to `g.JustBeforeEach` as a prerequisite step when each `g.By` becomes a `g.It`

test/extended/cli/builds.go
```diff
@@ -0,0 +1,179 @@
...
+	// check correct output type and image
+	out, err = oc.Run("get").Args("bc/"+opts.bcName, "--template", bcOutputTemplate).Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	if opts.outputType == "" {
+		o.Expect(out).To(o.BeEmpty())
+	} else {
+		o.Expect(out).To(o.Equal(fmt.Sprintf("%v %v", opts.outputType, opts.outputImage)))
+	}
+	// cleanup
+	resources := []string{"bc/" + opts.bcName}
```
ravisantoshgudimetla, line 59: The cleanup can be moved to `g.AfterEach()`

test/extended/cli/builds.go
```diff
@@ -0,0 +1,179 @@
...
+		o.ContainSubstring("Success"),
+	))
+	if opts.sourceType == "Dockerfile" {
+		o.Expect(buildOut).To(o.ContainSubstring(fmt.Sprintf("imagestream.image.openshift.io \"tests\" created"))) // see testDockerFileContent
+	}
+	if opts.isName != "" {
+		o.Expect(buildOut).To(o.ContainSubstring(fmt.Sprintf("imagestream.image.openshift.io \"%v\" created", opts.isName)))
+	}
+	// check source type
+	out, err := oc.Run("get").Args("bc/" + opts.bcName).Output()
```
ravisantoshgudimetla, line 47: What would happen if opts.bcName is empty? Would we error out or ignore?

I'd rather be explicit in the individual test rather than trying to get arg and then do validation here.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,179 @@
...
+		r.isName = r.bcName
+	}
+}
+
+func testNewBuild(oc *exutil.CLI, opts expectedBuild) {
+	opts.setDefaults()
+	// new build
+	buildOut, err := oc.Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(buildOut).To(o.And(
```
ravisantoshgudimetla, line 36: Shouldn't we wait for the build? Isn't the whole purpose of moving to go code is to be reasonable with the time frame and have knobs to control it?

ravisantoshgudimetla (changes requested):
Thank you for working on this @atiratree. The code looks correct to me. However I have the following concerns 
- The tests are not explicit. It'd also be hard to parse the [output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/26141/pull-ci-openshift-origin-master-e2e-gcp/1391681995452452864/artifacts/e2e-gcp/openshift-e2e-test/artifacts/e2e.log) when everything passes 
- The generic method `testNewBuild` is trying to be too generic. I believe you've gone with this approach since you did not want to repeat the code block. However in case of tests, I'd rather be explicit and be repetitive if it helps in debugging.
- I think we should wait for the build to finish and throw an error

atiratree:
> Thank you for working on this @atiratree. The code looks correct to me. However I have the following concerns
> 
> * The tests are not explicit. It'd also be hard to parse the [output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/26141/pull-ci-openshift-origin-master-e2e-gcp/1391681995452452864/artifacts/e2e-gcp/openshift-e2e-test/artifacts/e2e.log) when everything passes
> * The generic method `testNewBuild` is trying to be too generic. I believe you've gone with this approach since you did not want to repeat the code block. However in case of tests, I'd rather be explicit and be repetitive if it helps in debugging.

I wasn't sure with too many `It`s, since it generate a lot of new tests. 

Originally, I thought it would not be so bad for debugging - since one can still infer the place where bug happened from the logs. Although I agree it is a bit harder.

So, it is generally not advisible to do logic extraction? 
Would `for` which generates multiple `It`s help? Or can I extract some of the checks into functions? Or should all the logic be copy pasted? 

> * I think we should wait for the build to finish and throw an error

That sounds like a good idea. I wasn't sure how deep I should go. If I should just rewrite the tests. Or if I should analyze the CLI functionality regarding builds completely and enhance the tests? It would take considerably longer though.

test/extended/cli/builds.go
atiratree: I wasn't sure with too many `It`s, since it generate a lot of new tests and `By` doesn't. I agree, it makes sense to use `It`s

test/extended/cli/builds.go
atiratree: I am not sure if I should have AfterEach since I still haven't ported all the other tests and they will have different resources.

Or should  I just call `oc delete all --all` for simplicity there?

test/extended/cli/builds.go
atiratree: we could add a check there

test/extended/cli/builds.go
atiratree: Yes, I could add it. Please consider also https://github.com/openshift/origin/pull/26141#issuecomment-836731670

test/extended/cli/builds.go
atiratree: removed the config and simplified the flow so `g.JustBeforeEach` is not needed anymore

test/extended/cli/builds.go
atiratree: I have not researched the framework properly before. It seems it handles the deletion of the namespace in `AfterEach`, so additional logic should not be needed.

test/extended/cli/builds.go
atiratree: I am using different functions now and I added checks for bcName empty, but it seems superfluous now in these...

test/extended/cli/builds.go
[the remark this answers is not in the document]
atiratree: After a discussion changed to multiple standalone `It`s to support redability and better debugging.

atiratree:
@ravisantoshgudimetla thanks for the review. I tried to fix all the problems, but still haven't ported all the logic yet.

atiratree:
@ravisantoshgudimetla 

I have ported the first half of the `builds.sh`, but I am still struggling with waiting for the builds to complete. I cannot make any of the test builds successfully complete on my cluster. All end up in a Failed state. I will debug that more tomorrow.

Currently I am just waiting for the Build object to appear without checking the status/progress.

ravisantoshgudimetla:
>  but I am still struggling with waiting for the builds to complete. I cannot make any of the test builds successfully complete on my cluster. All end up in a Failed state. I will debug that more tomorrow.

Not an expert myself but can you use [`StartBuildAndWait`](https://github.com/openshift/origin/blob/master/test/extended/builds/build_timing.go#L70) function and see if it works?

soltysh:
> I found there is a bit of overlap with `[sig-builds]` as they are using oc cli to create builds as well. But it does not test the CLI part thourougly. I am not entirerly sure if it makes sense to combine `sig-builds` with `sig-cli`, but I decided not to at this moment.

@atiratree I forgot to mention, if you see duplicates feel free to remove as long as oc bits are properly tested in the existing tests. It's possible that for some of them it'll be true and it's ok to drop these tests, but some will be 1:1 duplicates, but the pre-existing will be using client-go, whereas yours will be strictly oc-focused - those should stay.

test/extended/cli/builds.go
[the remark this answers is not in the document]
soltysh: There's no strict definition how to group them reasonable, I'll leave it up to you to decide so that they make a logical sense.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+	"fmt"
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	t "github.com/onsi/gomega/types"
+	exutil "github.com/openshift/origin/test/extended/util"
+	"k8s.io/apimachinery/pkg/util/wait"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"regexp"
+	"strings"
+	"time"
```
soltysh, line 14: Nit, but it's nice to split imports into the following logical groups:
1. built-in imports
2. 3rd party imports
3. openshift imports

You'll notice this being mostly applied to all of our source code and in k8s as well.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+func waitForBuild(oc *exutil.CLI, bcName, buildNumber string) error {
+	args := []string{"builds"}
+
+	if buildNumber != "" {
+		args = append(args, fmt.Sprintf("%v-%v", bcName, buildNumber))
+	} else {
+		args = append(args, "-l", fmt.Sprintf("openshift.io/build-config.name=%v", bcName))
+	}
+
+	return wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
```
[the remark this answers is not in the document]
soltysh: First, you don't need to use `PollImmediate` since you're starting a build so I doubt it'll finish that soon, `wait.Poll` first waits and then performs poll which makes more sense in this case.

> I am not sure if I should wait here with `oc` or directly with go API

I'd probably go with an oc call, since we're focusing on that. The other tests are testing client-go calls.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+		lines := strings.Split(out, "\n")
+
+		return len(lines) >= 2, nil
+	})
+}
+
+func getExpectedBCOutputMatcher(bcName string, isNames ...string) t.GomegaMatcher {
+	matcher := o.ContainSubstring("Success")
+	if bcName != "" {
+		matcher = o.And(matcher, o.ContainSubstring(fmt.Sprintf("buildconfig.build.openshift.io \"%v\" created", bcName)))
```
soltysh, line 48: `%q` will be more appropriate, it'll handle quoting the string for you.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+	} else {
+		e2e.Logf("no bcName specified")
+	}
+
+	for _, isName := range isNames {
+		if isName != "" {
+			matcher = o.And(matcher, o.ContainSubstring(fmt.Sprintf("imagestream.image.openshift.io \"%v\" created", isName)))
```
soltysh, line 55: Ditto.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+
+	if buildNumber != "" {
+		args = append(args, fmt.Sprintf("%v-%v", bcName, buildNumber))
+	} else {
+		args = append(args, "-l", fmt.Sprintf("openshift.io/build-config.name=%v", bcName))
+	}
+
+	return wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+		out, err := oc.Run("get").Args(args...).Output()
+
```
soltysh, line 35: Unnecessary empty line, you're checking the error of the `Run` invocation, so it does make sense to group invocation and error check together. I see you're a fan of too many empty lines :wink:

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+	return matcher
+}
+
+func getBCInfoByTemplate(oc *exutil.CLI, bcName, template string) string {
+	if bcName == "" {
+		e2e.Logf("no bcName specified")
+		return ""
+	}
+	out, err := oc.Run("get").Args("bc/"+bcName, "--template", template).Output()
+
```
soltysh, line 67: Drop emtpy line.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+	return getBCInfoByTemplate(oc, bcName, bcOutputTemplate)
+}
+
+var _ = g.Describe("[sig-cli] oc builds", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLI("oc-builds").AsAdmin()
+
+	g.Describe("new-build", func() {
+		g.It("build from a binary with no inputs requires name", func() {
```
soltysh, line 89: Yeah, I agree with @ravisantoshgudimetla, I'd change those `It`'s to `By`'s.

soltysh: Also, it's nice and simplifies review if the `By` description matches that from https://github.com/openshift/origin/blob/d704a4d2ab5e55731d11770c11eacd666940b944/test/extended/testdata/cmd/test/cmd/builds.sh#L31 which it does here, but not necessarily everywhere.

test/extended/cli/builds.go
```diff
@@ -0,0 +1,339 @@
...
+			o.Expect(getBCOutputType(oc, "binary-test")).To(o.Equal("ImageStreamTag binary-test:latest"))
+		})
+
+		g.It("build from git with output to ImageStreamTag", func() {
+			out, err := oc.Run("new-build").Args("centos/ruby-26-centos7", "https://github.com/openshift/ruby-hello-world.git").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(out).To(getExpectedBCOutputMatcher("ruby-hello-world", "ruby-hello-world"))
+			o.Expect(getBCSourceType(oc, "ruby-hello-world")).To(o.Equal("Git"))
+			o.Expect(getBCOutputType(oc, "ruby-hello-world")).To(o.Equal("ImageStreamTag ruby-hello-world:latest"))
+			o.Expect(waitForBuild(oc, "ruby-hello-world", "")).NotTo(o.HaveOccurred())
```
soltysh, line 109: I'm looking at the original test in https://github.com/openshift/origin/blob/d704a4d2ab5e55731d11770c11eacd666940b944/test/extended/testdata/cmd/test/cmd/builds.sh#L27-L29 and it doesn't wait for the build to succeed, actually, it only checks the definition of the newly created BuildConfig matches the expectation. Waiting for every build to complete unnecessarily lengthens the test time, which is not the goal. The build correctness is verified elsewhere, we care about oc here, mostly.

test/extended/cli/builds.go
soltysh: My suggestion for organization is:
1. `Describe` as the top level
2. `It` as grouping, for example you have `new-build`, `complex-build`, etc.
3. `By` to mark every "group" from the original file

test/extended/testdata/application-template-dockerbuild.json
```diff
@@ -0,0 +1,443 @@
+{
```
soltysh, line 1: Don't copy this file, it already lives here: https://github.com/openshift/origin/blob/master/examples/sample-app/application-template-dockerbuild.json re-use that instead. We currently have 2 copies already, my goal is to go to only a single one, b/c it's easier to maintain.

test/extended/testdata/cmd/test/cmd/builds.sh
```diff
@@ -9,114 +9,15 @@ trap os::test::junit::reconcile_output EXIT
   exit 0
 ) &>/dev/null
 
-
-url=":${API_PORT:-6443}"
```
soltysh, line 13: When done, feel free to entirely drop this file.

soltysh (changes requested):

test/extended/cli/builds.go
atiratree: +1, looks much nicer

test/extended/cli/builds.go
atiratree: will try to keep the empty lines to a reasonable amount :)  , hope it is better now

test/extended/cli/builds.go
atiratree: I added one wait to the `start-build` test and removed the rest

test/extended/testdata/application-template-dockerbuild.json
atiratree: resolved

test/extended/cli/builds.go
```diff
@@ -0,0 +1,316 @@
...
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const (
+	bcOutputTemplate       = "{{with .spec.output.to}}{{.kind}} {{.name}}{{end}}"
+	bcOutputToNameTemplate = "{{ .spec.output.to.name }}"
+	bcSourceTypeTemplate   = "{{ .spec.source.type }}"
+	testDockerFileContent  = "FROM image-registry.openshift-image-registry.svc:5000/openshift/tests:latest"
```
soltysh, line 24: We should be using images from here https://github.com/openshift/origin/blob/e87c32d71e0462e3c32ee96e4a198d087bc6f001/test/extended/util/image/image.go#L128 I think that `ShellImage()` should work here. The reason for that is that those images are then available for mirroring and thus for offline testing.

soltysh (approved):
/lgtm
/approve
That nit can be fixed in a followup.

atiratree:
skpping the new tests on Disconnected clusters - the CI should pass now

/hold cancel

soltysh:
/retitle Bug 1992596: e2e/cli: move half of builds.sh to proper e2e

soltysh (approved):
/lgtm

PR #26147, opened 2021-05-10
Bug 1947800: Router: Switch from networking v1beta1 to v1

sgreene570:
test/extended/testdata/router/ingress.yaml:

Change `apiVersion` from `networking.k8s.io/v1beta1` to
`networking.k8s.io/v1`.

To ensure compatibility with the new v1 API, use a `pathType`
of `Prefix` for each of the test Ingress resources (which should be
compatible with existing test logic).

Also, use the correct service backend syntax, which requires a trivial
yaml layout change.

test/extended/testdata/bindata.go:

Regenerate.

This commit is in support of Bug 1947800.

---

Edit 5/11/21:
This PR will affect any backports that touch `test/extended/testdata/router/ingress.yaml` going forward.

### DIFF
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -48868,7 +48868,7 @@ var _testExtendedTestdataRouterIngressYaml = []byte(`kind: List
 apiVersion: v1
 items:
 # an ingress that should be captured as individual routes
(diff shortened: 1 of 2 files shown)


sgreene570:
Thanks Ben!


/test e2e-aws-serial

PR #26155, opened 2021-05-13
Bug 1989158: rewrite idling tests to not be [Local]

danwinship:
While investigating bz 1953705, I discovered that most of the unidling e2e tests have been disabled for years because they were written under the assumption that the test runner was running inside the cluster. (There is one unidling test in `test/extended/router/` that is not disabled, so we weren't _completely_ without test coverage...)

### DIFF
--- a/test/extended/idling/idling.go
+++ b/test/extended/idling/idling.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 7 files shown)


frobware:
@danwinship FYI - the test in `test/extended/router/` is new this year (https://github.com/openshift/origin/pull/25850).

squeed:
good catch
/approve
/lgtm

squeed:
@knobunc can you approve this, please?

danwinship:
/test e2e-gcp
(i think the sdn 1.22 rebase may have broken this :-)

PR #26156, opened 2021-05-13
Bug 1959564: builds: skip /run verification tests

adambkaplan:
Skipping verification tests for the /run directory in builds. A change was
recently made at the node/cri-o layer which added additional content to
the /run directory.

This change should be reverted once the contents added to /run have been
vetted.

### DIFF
--- a/test/extended/builds/run_fs_verification.go
+++ b/test/extended/builds/run_fs_verification.go
@@ -144,6 +144,7 @@ valid_fields.json
 
 		g.Describe("do not have unexpected content", func() {
 			g.It("using a simple Docker Strategy Build", func() {
+				g.Skip("Bug 1959564: skipping while additional content in /run is investigated")
 				g.By("calling oc create with yaml")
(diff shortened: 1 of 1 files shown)


test/extended/builds/run_fs_verification.go
```diff
@@ -144,6 +144,7 @@ valid_fields.json
 
 		g.Describe("do not have unexpected content", func() {
 			g.It("using a simple Docker Strategy Build", func() {
+				g.Skip("Bug 1959564: skipping while additional content in /run is investigated")
```
ravisantoshgudimetla, line 147: How about using `result.Flakef` so that we get a clear signal?

adambkaplan: This isn't a flaky test - cri-o broke a conformance test for builds. This unblocks teams from merging fixes.

It will still appear in the list of skipped tests - I do concede that the skipped list can be quite large.

PR #26158, opened 2021-05-13
test: SetupNamespace in cli helper should wait for default service account

smarterclayton:
Otherwise pods running in that namespace won't be able to access all
images on the cluster.

At least one cause of failure to pull openshift/tools from registry - registry requires auth, if the pod is created before SA provisioning then you can't pull that image (because the pull secret isn't added).

Likely introduced when we switched to docker.


smarterclayton:
This is fully replaced with #26159

PR #26159, opened 2021-05-13
Wait longer in tests where namespace cleanup is slow

smarterclayton:
Also fix a case where we waited for dockercfg but not the token (which we need).

This fixes the failure to pull openshift/tools in a number of tests (we raced with the secret controller)

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -984,13 +984,23 @@ func WaitForServiceAccount(c corev1client.ServiceAccountInterface, name string) 
 			return false, fmt.Errorf("Failed to get service account %q: %v", name, err)
 		}
 		secretNames := []string{}
+		var hasDockercfg, hasToken bool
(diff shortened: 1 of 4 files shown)


test/extended/util/framework.go
```diff
@@ -984,13 +984,20 @@ func WaitForServiceAccount(c corev1client.ServiceAccountInterface, name string)
 			return false, fmt.Errorf("Failed to get service account %q: %v", name, err)
 		}
 		secretNames := []string{}
+		var hasDockercfg, hasToken bool
 		for _, s := range sc.Secrets {
-			if strings.Contains(s.Name, "dockercfg") {
-				return true, nil
+			if strings.Contains(s.Name, "-dockercfg-") {
```
deads2k, line 989: This one needs to be in `sc.ImagePullSecrets`.

smarterclayton: Good catch

PR #26160, opened 2021-05-13
test: Every control plane component should have PDB

ravisantoshgudimetla:
This ensures every component is having PDB associated. Need to figure out how to ensure only operands have the PDB not the operators. One way to do it is to exclude `-operator` namespace but not all components are following the convention.

it is currently missing static pods, will add them soon.

cc @smarterclayton @wking @deads2k @soltysh

### DIFF
--- a/test/extended/pods/pdb.go
+++ b/test/extended/pods/pdb.go
@@ -1,0 +1,182 @@
+package pods
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/api/policy/v1beta1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes"
+	"reflect"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	oapi "github.com/openshift/api/config/v1"
+	"github.com/openshift/origin/pkg/test/ginkgo/result"
+	exutil "github.com/openshift/origin/test/extended/util"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+// getAllControlPlaneStaticPods gets the static pod
+func getAllControlPlaneStaticPods(kubeClient kubernetes.Interface) (sets.String, error) {
+	pod, err := kubeClient.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	var controlPlanePods = sets.NewString()
+	for _, pod := range pod.Items {
+		_, isStaticPod := pod.Annotations["kubernetes.io/config.mirror"]
+		_, isControlPlanePod := pod.Annotations["target.workload.openshift.io/management"]
+		if isStaticPod && isControlPlanePod && strings.Contains(pod.Namespace, "openshift-") {
+			staticPodName := strings.Split(pod.Name, "-ip")[0]
+			if staticPodName == "etcd" {
+				// etcd pod name needs to be appended with -quorum as it has quorum guard associated
+				staticPodName = staticPodName+"-quorum"
+			}
+			controlPlanePods.Insert(staticPodName)
+		}
+	}
+	return controlPlanePods, nil
+}
+
+var _ = g.Describe("[sig-arch] Managed cluster", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLIWithoutNamespace("ensure-pdb")
+
+	g.It("all platform components should have PDBs configured", func() {
+		kubeClient := oc.AdminKubeClient()
+
+		deployments, err := kubeClient.AppsV1().Deployments("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list deployments: %v", err)
+		}
+		statefulsets, err := kubeClient.AppsV1().StatefulSets("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list statefulsets: %v", err)
+		}
+		// ovn-kube, openshift-controller-manager are running as DS.
+		daemonsets, err := kubeClient.AppsV1().DaemonSets("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list daemonsets: %v", err)
+		}
+
+		var items []runtime.Object
+		for i := range deployments.Items {
+			items = append(items, &deployments.Items[i])
+		}
+		for i := range statefulsets.Items {
+			items = append(items, &statefulsets.Items[i])
+		}
+		for i := range daemonsets.Items {
+			items = append(items, &daemonsets.Items[i])
+		}
+		// iterate over the references to find valid images
+		infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("unable to get infrastructure: %v", err)
+		}
+		var inHAMode bool
+		if infra.Status.InfrastructureTopology == oapi.HighlyAvailableTopologyMode {
+			inHAMode = true
+		}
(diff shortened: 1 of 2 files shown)


test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+			items = append(items, &daemonsets.Items[i])
+		}
+		// iterate over the references to find valid images
+		infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("unable to get infrastructure: %v", err)
+		}
+		var inHAMode bool
+		if infra.Status.InfrastructureTopology == oapi.HighlyAvailableTopologyMode {
+			inHAMode = true
```
wking, line 83: This seems like it should be orthogonal.  Can't we get at this just by ignoring `replicas < 2` workloads?  Because even if they allow surging, we don't surge on drains.

test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		//controlPlaneStaticPodNames := sets.NewString("etcd", "kube-apiserver", "kube-controller-manager",
+		//	"openshift-kube-scheduler")
```
wking, line 86: nit: did you mean to drop this comment now that you have `getAllControlPlaneStaticPods`?

test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+				meta = &t.ObjectMeta
+				selectorLabels = t.Spec.Selector
+				replicas = t.Status.Replicas
+			case *appsv1.DaemonSet:
+				t.Kind = "DaemonSet"
+				meta = &t.ObjectMeta
+				selectorLabels = t.Spec.Selector
+				replicas = t.Status.DesiredNumberScheduled
+			default:
+				panic("not an object")
```
wking, line 116: nit: `panic(fmt.Sprintf("unrecognized workload type %T", workload))` or some such to give more context?

test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+			if !strings.Contains(meta.Namespace, "openshift-") {
+				// Not every component is following the convention of -operator at the end, there are some namespaces
+				// like openshift-monitoring and openshift-network-diagnostics and some of them run on worker nodes
+				// as well. So, need to figure out how to exclude operator components and components running on worker node
```
wking, line 121: I don't understand the need to exclude operators from this.  If we have an operator with `replicas > 1`, can't we apply the same PDB considerations to it as well?

test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+		deployments, err := kubeClient.AppsV1().Deployments("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list deployments: %v", err)
+		}
+		statefulsets, err := kubeClient.AppsV1().StatefulSets("").List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			e2e.Failf("unable to list statefulsets: %v", err)
+		}
+		// ovn-kube, openshift-controller-manager are running as DS.
+		daemonsets, err := kubeClient.AppsV1().DaemonSets("").List(context.Background(), metav1.ListOptions{})
```
michaelgugino, line 61: daemonsets can't be evicted today via drain, so enforcing PDBs without upstream changes on these would not be useful.

wking, line 61: DaemonSets aren't drained.  There may still be caps around how many we want to disrupt by shutting down nodes, but I dunno if PDBs actually protect them from this today.

michaelgugino (changes requested):
Many things don't require PDBs, like the machine-api.  That runs with 1 replica, short periods of unavailability while it is rescheduled to another host are fine.

test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+					} else {
+						e2e.Logf("Workload %s has PDB %s configured", key, pdb.Name)
+					}
+					break
+				}
+			}
+			if inHAMode {
+				key = "HA/" + key
+				if !isPDBConfigured {
+					pdbMissingWorkloads = append(pdbMissingWorkloads, key)
```
wking, line 155: I don't think we need to complain about missing PDBs on `replicas=1` workloads; they're clearly interruptible.  Can we restructure this like:

```go
if replicas > 1 {  // no reason to set this unless we are trying to be HA, right?
  if currentPDB == nil { // you'd need to make currentPDB a pointer for this to work
    pdbMissingWorkloads = append(pdbMissingWorkloads, key)
  } else {
    // checks to see if we were comfortable with the PDB configuration
  }
}
```

test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+
+				} else if isPDBConfigured && replicas > 2 {
+					// We expect atleast 1 pod to be available while evicting/draining.
+					if currentPDB.Spec.MaxUnavailable != nil && currentPDB.Spec.MaxUnavailable.IntValue() < 1 {
+						pdbMisconfiguredWorkloads = append(pdbMisconfiguredWorkloads, key)
+					} else if currentPDB.Spec.MinAvailable != nil && currentPDB.Spec.MinAvailable.IntValue() < 1 {
+						pdbMisconfiguredWorkloads = append(pdbMisconfiguredWorkloads, key)
```
wking, line 162: We should probably add a message about why we are mad at this key, by either making `pdbMisconfiguredWorkloads` a map from workload key to string:

```go
pdbMisconfiguredWorkloads[key] = fmt.Sprintf("sets minAvailable %s, but $WHY_WE_WANT_IT_UNSET_OR_GREATER_THAN_ZERO", currentPDB.Spec.MinAvailable)
```

or a map from a misconfigured-reason to a slice of keys:

```go
pdbMisconfiguredWorkloads[pdbMaxUnavailableTooSmall] = append(pdbMisconfiguredWorkloads[pdbMaxUnavailableTooSmall], key)
```

test/extended/pods/pdb.go
```diff
@@ -0,0 +1,182 @@
...
+				e2e.Failf("Unexpected error getting PDBs: %v", err)
+			}
+			for _, pdb := range pdbs.Items {
+				if reflect.DeepEqual(pdb.Spec.Selector, selectorLabels) {
+					isPDBConfigured = true
+					currentPDB = pdb
+					operatorsHavingPDBS = append(operatorsHavingPDBS, key)
+					// The convention we came up with for static pod control plane components is to have a deployment
+					// doing health check for the corresponding component. This can be changed to labels later.
+					// As of now, I am assuming if the deployment name has `-guard` in the name and it runs in
```
deads2k, line 140: this seems reasonable

michaelgugino:
FYI, if we're interested in draining daemonsets, I have a patch upstream ready to go: https://github.com/kubernetes/kubernetes/pull/88345

This would allow us to put PDBs that cover daemonsets to add back pressure to rolling nodes, however, we will be exposed to endless looping of drain if any daemonset takes longer to stop than the timeout, as a new one would be present when we requeue the drain operation.  We would need to implement some kind of filtering mechanism.  We added custom pod filters for draining, so it's possible to do that today, we just need to work out what that might look like if we're interested in pursuing this.

test/extended/pods/pdb.go
ravisantoshgudimetla: Fair enough, I think we have a flag which evicts DS pods too but I can keep DS out of the PR now.

test/extended/pods/pdb.go
ravisantoshgudimetla: Of course :(

test/extended/pods/pdb.go
ravisantoshgudimetla: The goal here is to get control plane components most of which run on master nodes.

test/extended/pods/pdb.go
ravisantoshgudimetla: I can change it. In fact, I can ignore it before the loop

test/extended/pods/pdb.go
wking: OpenShift-core operators will also run on the control-plane machines.  If any of them have `replicas > 1` (maybe none of them), why wouldn't we want them covered by PDBs?

PR #26164, opened 2021-05-17
Bug 1959564: move verify run test from tools to cli image to avoid additional /run content

gabemontero:
@adambkaplan @nalind FYI

Assuming we can get this to work like we want it I'll associate the bz, fully remove the skip, etc.

### DIFF
--- a/test/extended/builds/run_fs_verification.go
+++ b/test/extended/builds/run_fs_verification.go
@@ -35,7 +35,7 @@ spec:
           value: "10"
       imageOptimizationPolicy: SkipLayers
     type: Docker
-`, image.ShellImage())
+`, image.LimitedShellImage())
 		testVerifyRunFSContentsBuildConfigYaml = fmt.Sprintf(`
 apiVersion: build.openshift.io/v1
 kind: BuildConfig
@@ -56,7 +56,7 @@ spec:
           value: "10"
       imageOptimizationPolicy: SkipLayers
     type: Docker
-`, image.ShellImage())
+`, image.LimitedShellImage())
 		lsRSlashRun = `
 /run:
 lock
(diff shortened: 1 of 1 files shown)


test/extended/builds/run_fs_verification.go
```diff
@@ -35,7 +35,7 @@ spec:
           value: "10"
       imageOptimizationPolicy: SkipLayers
     type: Docker
-`, image.ShellImage())
+`, image.LimitedShellImage())
```
gabemontero, line 38: Or as @nalind mentioned during scrum, the cli image is built off of 

`registry.ci.openshift.org/ocp/4.8:base`

There is not helper method yet in the image component that exposes that image, but we could either try to add one there, or just reference it directly here.

gabemontero: But it gets complicated cause we really should use an image stream image reference, and there is not imagestream for the base image right now.

This gets into the barriers we may encounter that I alluded to wrt adding imagestreams to the openshift namespace.
Similarly getting `registry.ci.openshift.org/ocp/4.8:base` into https://github.com/openshift/origin/blob/master/test/extended/util/image/image.go#L15 could prove difficult.

IMO the cli image via `LimitedShellImage` is the path of least resistance

gabemontero:
missed a spot where I had to switch

gabemontero:
ok my test cluster is now up, and this is passing for me locally

let's see what happens in CI

gabemontero:
e2e-gcp-builds passes @adambkaplan @nalind

gabemontero:
GCP IAM pain on e2e-gcp failure

/test e2e-gcp

gabemontero:
e2e-gcp still has IAM pain, cluster can't get started

gabemontero:
unrelated failures in e2e-aws-csi and e2e-aws-disruptive

gabemontero:
all green tests for this bz PR @adambkaplan

xiuwang:
The two tests are passed , the result is in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26164/pull-ci-openshift-origin-master-e2e-aws-fips/1394396549101916160

/label qe-approved

adambkaplan (approved):
/lgtm

gabemontero:
/cherrypick release-4.7

@dmage per exchange in slack

PR #26165, opened 2021-05-17
Bug 1961201: DNS: Use endpointslice v1 for DNS dual-stack test

sgreene570:
test/extended/dns/dns.go:

Use the `discoveryv1` instead of `discoveryv1beta1`
API since `discoveryv1beta1` is officially deprecated in
Kubernetes 1.21.

Note that the test modified in this commit
only runs on dual-stack enabled clusters.

This commit is in support of Bug 1961201.

### DIFF
--- a/test/extended/dns/dns.go
+++ b/test/extended/dns/dns.go
@@ -13,7 +13,7 @@ import (
 	. "github.com/onsi/ginkgo"
 
 	kapiv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


sgreene570:
https://github.com/openshift/coredns/pull/57 should unblock 
` [sig-network-edge] DNS should answer A and AAAA queries for a dual-stack service is constantly failing`, however,
this PR is also important since discoveryv1beta1 is deprecated in kube 1.21.

PR #26167, opened 2021-05-18
Bug 1961582: Do not install python3-cinderclient on s390x

joepvd:
The openstack repository does not yet exist for s390x, where
python3-cinderclient should be coming from. Safelisting the
existing arches.


mandre:
Shouldn't this target `master` instead of `release-4.8` ?

joepvd:
/close

in favor of https://github.com/openshift/origin/pull/26170

PR #26131, opened 2021-05-04
Bug 1957131: test: Add the Cinder client to the test image

pierreprinetti:
Multiple tests in the e2e Conformance suite require the Cinder client to be installed when run
against OpenStack.

In the step-registry CI structure, the "test" ref is shared across
platforms; it is therefore not desirable to run the OpenStack tests
using a separate image.

This change adds the Cinder client to the container image used to run
the tests in the CI.

Implements [OSASINFRA-2308](https://issues.redhat.com/browse/OSASINFRA-2308)

cf. "[sig-storage] In-tree Volumes [Driver: cinder] [Testpattern:
Inline-volume (default fs)] volumes should store data
[Suite:openshift/conformance/parallel] [Suite:k8s]"

PR #26170, opened 2021-05-19
Bug 1961582: Do not install python3-cinderclient on s390x

joepvd:
The openstack repository does not yet exist for s390x, where
python3-cinderclient should be coming from. Safelisting the
existing arches.

### DIFF
--- a/images/tests/Dockerfile.rhel
+++ b/images/tests/Dockerfile.rhel
@@ -7,7 +7,9 @@ RUN make; \
 
 FROM registry.ci.openshift.org/ocp/4.8:tools
 COPY --from=builder /tmp/build/openshift-tests /usr/bin/
-RUN yum install --setopt=tsflags=nodocs -y git gzip util-linux python3-cinderclient && yum clean all && rm -rf /var/cache/yum/* && \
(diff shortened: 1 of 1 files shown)


pierreprinetti:
/assign bparees

https://github.com/openshift/origin/pull/26131 turned out to be a blocker for architectures where the openstack repository is not available.

images/tests/Dockerfile.rhel
```diff
@@ -7,7 +7,9 @@ RUN make; \
 
 FROM registry.ci.openshift.org/ocp/4.8:tools
 COPY --from=builder /tmp/build/openshift-tests /usr/bin/
-RUN yum install --setopt=tsflags=nodocs -y git gzip util-linux python3-cinderclient && yum clean all && rm -rf /var/cache/yum/* && \
+RUN PACKAGES="git gzip util-linux" && \
+    if [ $HOSTTYPE = x86_64] || [ $HOSTTYPE = ppc64le ]; then PACKAGES="$PACKAGES python3-cinderclient"; fi && \
```
EmilienM, line 11: breaking bash, see https://github.com/openshift/origin/pull/26192

PR #26173, opened 2021-05-19
do not allow an excessive count of an individual event: they indicate a failure

deads2k:
If the same event is emitted more than X times (20?), something has gone wrong and should be fixed.

Let's see what this exposes.


deads2k:
merged in a different PR

PR #26188, opened 2021-05-27
Bug 1957894: Enable pull from private registry with secret test

harche:
Signed-off-by: Harshal Patil <harpatil@redhat.com>


harche:
I have been running `/test e2e-metal-ipi-ovn-ipv6` on this PR. Although the test job shows flakiness it doesn't failed due to `the test Container Runtime blackbox test when running a container with a new image should not be able to pull from private registry without secret` reported as a part of https://bugzilla.redhat.com/show_bug.cgi?id=1957894

PR #26192, opened 2021-05-31
Bug 1961582: Dockerfile.rhel: fix a bash typo

EmilienM:
A regression was introduced in the Dockerfile, with a typo in
the condition:
https://github.com/openshift/origin/pull/26170

This patch fixes it.


mandre (commented):
/lgtm

pierreprinetti:
/assign bparees
🙏
This one is going to drive us crazy 😝

PR #26197, opened 2021-06-03
Bug 1953127: rebase k8s, undisable NetworkPolicy tests

danwinship:
undisables the NetworkPolicy tests which got disabled in the 1.21 rebase

rebase pulls in
 - https://github.com/openshift/kubernetes/pull/737 add more context to log after a request times out
 - https://github.com/openshift/kubernetes/pull/756 Does not prevent pod creation because of no nodes reason when it runs under the regular cluster
 - https://github.com/openshift/kubernetes/pull/770 work around broken NetPol DNS rules
 - https://github.com/openshift/kubernetes/pull/783 client-go: add retry logic for Watch and Stream

(where 770 is the part needed to fix the NetworkPolicy tests)


danwinship:
included in #26200

PR #26203, opened 2021-06-07
[sig-arch][Late] operators should not create watch channels very often [Suite:openshift/conformance/parallel]

ingvagabund:
Measure the amount of watch requests per operator per bucket and upper bound it in case
an operator introduces an error leading to opening too many watch requests in short time.

https://issues.redhat.com/browse/WRKLDS-291

### DIFF
--- a/test/extended/apiserver/api_requests.go
+++ b/test/extended/apiserver/api_requests.go
@@ -9,11 +9,14 @@ import (
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 	apiserverv1 "github.com/openshift/api/apiserver/v1"
+	configv1 "github.com/openshift/api/config/v1"
 	apiserverclientv1 "github.com/openshift/client-go/apiserver/clientset/versioned/typed/apiserver/v1"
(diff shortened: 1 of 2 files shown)


ingvagabund:
From `pull-ci-openshift-origin-master-e2e-gcp`:
```
[36mINFO[0m[2021-06-07T13:22:33Z] passed: (1.9s) 2021-06-07T13:20:00 "[sig-arch][Late] operators should not create watch channels very often [Suite:openshift/conformance/parallel]" 
```

soltysh (approved):
My only worry is keeping those upper bounds up-to-date, especially that they are very granular (per platform). I'd like to have 2nd pair of eye on it.
/lgtm
/approve
/hold
/assign @deads2k

soltysh:
Spoke with David on slack
/hold cancel

deads2k:
Having a cap on the number of requests made by parts of the payload is important. We cut the number by 20% in 4.8.  A test that keeps us from getting worse is a good idea.

I haven't reviewed this implementation, but current state plus a percentage is a good starting point.  It may eventually be necessary to phrase it as qps in some way.

PR #26204, opened 2021-06-07
Bug 1963730: updates local kubeconfig conformance test to cover multiple paths

p0lyn0mial:
https://github.com/openshift/library-go/pull/1098 moved the cert directory under a new path hence we need to cover both the old and the new paths.

This PR also unblocks https://github.com/openshift/cluster-kube-apiserver-operator/pull/1138

### DIFF
--- a/test/extended/apiserver/kubeconfigs.go
+++ b/test/extended/apiserver/kubeconfigs.go
@@ -3,10 +3,13 @@ package apiserver
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+
 	exutil "github.com/openshift/origin/test/extended/util"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/kubernetes/test/e2e/framework"
 )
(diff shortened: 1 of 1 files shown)


test/extended/apiserver/kubeconfigs.go
```diff
@@ -21,21 +24,59 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
...
+						// it looks like relying on the exit code from the debug command is not enough
+						// make sure the output contains what we have asked for
+						commandSucceeded := err == nil && strings.Contains(out, `"name": "kube-system"`)
+
+						if commandSucceeded && !foundGoodKubeConfig {
+							foundGoodKubeConfig = true
+							continue // make sure the other paths that follow will fail
+						}
+						if commandSucceeded && foundGoodKubeConfig {
+							g.Fail("only one path should have a valid kubeconfig")
```
p0lyn0mial, line 70: unless we run the test against previous versions as well.

p0lyn0mial:
/hold

as it is not clear if the DR scripts will work

p0lyn0mial:
/hold

for 4.8 we decided to create certificates in an atomic way. In 4.9 we are going to change the paths.

PR #26205, opened 2021-06-07
test: Tolerate control plane unavailability before 4.8

smarterclayton:
If we are upgrading from versions before 4.8, some necessary fixes
are not present and we need to avoid failing the test (because we
deliberately decided not to tighten these restrictions). Only
require 0 disruption going forward. This applies to all platforms.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/RangelReale/osincli v0.0.0-20160924135400-fababb0555f2
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/auth0/go-jwt-middleware v0.0.0-20201030150249-d783b5c46b39 // indirect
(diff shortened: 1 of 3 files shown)


smarterclayton:
4.7 to this PR test upgrade in https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1402012014028525568, previous version was wrong.

Will tag when that passes.

ravisantoshgudimetla (approved):
/lgtm

smarterclayton:
4.7 to this PR test upgrade in https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1402043649365118976 worked, previous version was wrong.

Force merging

PR #26208, opened 2021-06-08
Bug 1970315: testPodSandboxCreation: skip sandbox errors for pods which were not deleted during network update

vrutkovs:
"pods should successfully create sandboxes" test should mark pod events as flakes if network is being updated. 
During network update CNI binaries may be in the middle of update. This may cause sandbox errors like:
* `error adding container to network "ovn-kubernetes": failed to send CNI request: Post "http://dummy/": EOF`
* `Multus: [openshift-dns/dns-default-nbkz2]: have you checked that your default network is ready? still waiting for readinessindicatorfile @ /var/run/multus/cni/net.d/10-ovn-kubernetes.conf. pollimmediate error: timed out waiting for the condition`
* `error adding container to network "openshift-sdn": failed to find plugin "openshift-sdn" in path [/opt/multus/bin /var/lib/cni/bin /usr/libexec/cni]`

["never deleted" search in 4.7 -> 4.8 upgrades for last 7 days](https://search.ci.openshift.org/?search=never+deleted&maxAge=168h&context=1&type=junit&name=periodic-ci-openshift-release-master-ci-4.8-upgrade-from-stable-4.7.*&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job)

If these events are occuring during network/machine-config update and sandboxes eventually get created (i.e. the pod never gets deleted) these events are marked as flakes.

Test runs:
* [test upgrade 4.7 openshift/origin#26208 gcp](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1408101759766761472)
* [test upgrade 4.7 openshift/origin#26208 gcp,ovn](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1408101788195753984)
* [test upgrade 4.8.0-rc.0 openshift/origin#26208 gcp](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1408101871096172544)
* [test upgrade 4.8.0-rc.0 openshift/origin#26208 gcp,ovn](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1408101880483024896)

### DIFF
--- a/pkg/synthetictests/networking.go
+++ b/pkg/synthetictests/networking.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/origin/pkg/monitor/intervalcreation"
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 
 	"github.com/openshift/origin/pkg/test/ginkgo"
@@ -33,6 +34,10 @@ func testPodSandboxCreation(events monitorapi.Intervals) []*ginkgo.JUnitTestCase
 
 	failures := []string{}
 	flakes := []string{}
+	operatorsProgressing := intervalcreation.IntervalsFromEvents_OperatorProgressing(events, time.Time{}, time.Time{})
+	networkOperatorProgressing := operatorsProgressing.Filter(func(ev monitorapi.EventInterval) bool {
+		return ev.Locator == "clusteroperator/network" || ev.Locator == "clusteroperator/machine-config"
+	})
 	eventsForPods := getEventsByPod(events)
 	for _, event := range events {
 		if !strings.Contains(event.Message, "reason/FailedCreatePodSandBox Failed to create pod sandbox") {
@@ -44,8 +49,21 @@ func testPodSandboxCreation(events monitorapi.Intervals) []*ginkgo.JUnitTestCase
 		}
 		deletionTime := getPodDeletionTime(eventsForPods[event.Locator], event.Locator)
 		if deletionTime == nil {
-			// this indicates a failure to create the sandbox that should not happen
-			failures = append(failures, fmt.Sprintf("%v - never deleted - %v", event.Locator, event.Message))
+			// mark sandboxes errors as flakes if networking is being updated
+			match := -1
+			for i := range networkOperatorProgressing {
+				matchesFrom := event.From.After(networkOperatorProgressing[i].From)
+				matchesTo := event.To.Before(networkOperatorProgressing[i].To)
+				if matchesFrom && matchesTo {
+					match = i
+					break
+				}
+			}
+			if match != -1 {
+				flakes = append(flakes, fmt.Sprintf("%v - never deleted - network rollout - %v", event.Locator, event.Message))
+			} else {
+				failures = append(failures, fmt.Sprintf("%v - never deleted - %v", event.Locator, event.Message))
+			}
 		} else {
 			timeBetweenDeleteAndFailure := event.From.Sub(*deletionTime)
 			switch {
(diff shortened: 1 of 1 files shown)


pkg/synthetictests/networking.go
```diff
@@ -44,16 +45,30 @@ func testPodSandboxCreation(events monitorapi.Intervals) []*ginkgo.JUnitTestCase
 		}
 		deletionTime := getPodDeletionTime(eventsForPods[event.Locator], event.Locator)
 		if deletionTime == nil {
+			// mark sandboxes errors as flakes if networking is being updated
+			// these pods eventually get created
+			operatorsProgressing := intervalcreation.IntervalsFromEvents_OperatorProgressing(events, event.From, event.To)
```
smarterclayton, line 50: Hrm, this could be O(N^2) on a pretty big N, have you verified that IntervalsFromEvents uses binary search?

vrutkovs: Seems `IntervalsFromEvents_OperatorProgressing` is `O(N)`

smarterclayton: The list should be sorted, so if you know from to you can do a binary search o(logn) to find the start and then same for the end.  Or maybe just at the beginning do a single pass and calculate all the intervals that the operator is progressing (which should be very small O) and then just do the smaller loop here?

smarterclayton: (see intervals.go / monitor.go for a method that already uses sort.Search() to do this)

vrutkovs: Reworked this to use monitorapi functions:
* `CopyAndSort` to create a copy of events and sort them by type
* `IntervalsFromEvents_OperatorProgressing` to build a list of operator progressing events
* `sort.Search` to find events for network/machine-config

pkg/synthetictests/networking.go
```diff
@@ -15,7 +17,7 @@ type testCategorizer struct {
 	substring string
 }
 
-func testPodSandboxCreation(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
+func testPodSandboxCreation(unsortedEvents monitorapi.Intervals) []*ginkgo.JUnitTestCase {
```
smarterclayton, line 20: Wait, these should be sorted by default.  Are they not?

smarterclayton: (I thought they would be sorted)

vrutkovs: I don't think these are sorted.
We're reading those in [`EventsFromFile`](https://github.com/openshift/origin/blob/2b30f8e6a2a55ea3f7f1415ba8fdcf5135888c4b/pkg/monitor/serialization/serialize.go#L38) and written in [`EventsToJSON`](https://github.com/openshift/origin/blob/2b30f8e6a2a55ea3f7f1415ba8fdcf5135888c4b/pkg/monitor/serialization/serialize.go#L67-L76), neither is explicitely sorting events.

In any case, if these are sorted, it would take a constant time to sort them, as we do it once when running the test

vrutkovs: Ah, we do [sort](https://github.com/openshift/origin/blob/2b30f8e6a2a55ea3f7f1415ba8fdcf5135888c4b/pkg/monitor/serialization/serialize.go#L73) them on write, my bad.

Removing sorting in the test

pkg/synthetictests/networking.go
```diff
@@ -44,16 +47,27 @@ func testPodSandboxCreation(events monitorapi.Intervals) []*ginkgo.JUnitTestCase
 		}
 		deletionTime := getPodDeletionTime(eventsForPods[event.Locator], event.Locator)
 		if deletionTime == nil {
-			// this indicates a failure to create the sandbox that should not happen
-			failures = append(failures, fmt.Sprintf("%v - never deleted - %v", event.Locator, event.Message))
+			// look for network/machine-config progressing in operator events, stop when either found
+			operatorsProgressing := intervalcreation.IntervalsFromEvents_OperatorProgressing(events, event.From, event.To)
```
smarterclayton, line 51: So... is there a reason why you don't grab these out up top and prefilter by networking?  Then while it's O(N^2) it's at least O(MN) where M is small.

vrutkovs: Right, its easier to fetch all progressing operator intervals, filter by network - and then find if event is in the progressing networking via `.Slice()`. Implementing that in 41a2e0c6cb

vrutkovs:
/hold

Doesn't seem to be helping, still debugging this

pkg/synthetictests/networking.go
vrutkovs: `sort.Search`/`Slice` was an incorrect approach, reworked this in 54660eec. See PR description for test result links, all have sandboxes errors marked as flakes as expected

ravisantoshgudimetla (approved):
/lgtm

Thank you for working on this @vrutkovs

vrutkovs:
/skip

e2e-aws-upgrade failing due to https://github.com/openshift/release/pull/19836

PR #26211, opened 2021-06-08
Bug 1969561: Use cluster-logging for operator installation smoke test.

benluddy:
A new catalog-operator alert is being introduced that fires when the
API server produces warnings during operator installation.

The latest amq-streams catalog entry hasn't completed its migration
away from v1beta1 CRDs, which results in deprecation warnings that in
turn cause the catalog-operator alert to fire.


anik120 (approved):
/lgtm

exdx:
/cherry-pick release-4.7

context: https://bugzilla.redhat.com/show_bug.cgi?id=1980187

PR #26212, opened 2021-06-08
Add test/extended/operators OWNERS file

kevinrizza:


test/extended/operators/OWNERS
```diff
@@ -0,0 +1,20 @@
...
+  - ecordell
+approvers:
+  - smarterclayton
+  - bparees
+  - mfojtik
+  - knobunc
+  - ironcladlou
+  - adambkaplan
+  - kevinrizza
+  - ecordell
```
bparees, line 20: you should only list OLM people here.

it's inherited, so i'll still have permissions from the top level files.

kevinrizza:
@bparees updated. please take another look

PR #26213, opened 2021-06-09
test: Don't reuse route.spec.host in tests run in parallel

smarterclayton:
Tests that hardcode spec.host can fail to be admitted while other instances of the same test are running (in parallel / stress jobs).
We don't need a hardcoded host anyway.

Also, max_sessions tracking is unreliable in the presence of too many restarts, so simply verify after a restart it's zero for our route (as it should be).

### DIFF
--- a/test/extended/router/metrics.go
+++ b/test/extended/router/metrics.go
@@ -31,6 +31,8 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 
 	configv1 "github.com/openshift/api/config/v1"
+	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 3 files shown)


smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1403052129899253760 passed, which is good.  Will wait for one more.

smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1403096383774265344 also passed, force merging to clear the stress test

PR #26214, opened 2021-06-09
upgrade test: Fix firing PDBAtLimit alert

ravisantoshgudimetla:
Exclude PodDisruptionBudgetAtLimit which is a warning
level that is bound to trigger during upgrades. As we
upgrade one node at a time and desired number of healthy
pods match current number of healthy pods as
maxUnavailable is set to 1

cc @sttts @smarterclayton @soltysh

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -86,7 +86,13 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 			},
 		},
 	}
-
+	allowedFiringAlerts := helper.MetricConditions{
+		{
+			// TODO: Add all the control-plane components once all of them have PDBs configured
+			Selector: map[string]string{"alertname": "PodDisruptionBudgetAtLimit", "namespace": "openshift-etcd"},
+			Text:     "Excluded because PodDisruptionBudgetAtLimit is a warning level that is bound to trigger during upgrades as we upgrade one node at a time and desired number of healthy pods match current number of healthy pods as maxUnavailable is set to 1",
+		},
+	}
 	pendingAlertsWithBugs := helper.MetricConditions{
 		{
 			Selector: map[string]string{"alertname": "ClusterMonitoringOperatorReconciliationErrors"},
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade/alert/alert.go
```diff
@@ -86,7 +86,13 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 			},
 		},
 	}
-
+	allowedFiringAlerts := helper.MetricConditions{
+		{
+			// TODO: Add all the control-plane components once all of them have PDBs configured
+			Selector: map[string]string{"alertname": "PodDisruptionBudgetAtLimit", "namespace": "openshift-etcd"},
+			Text:     "Excluded because PodDisruptionBudgetAtLimit is a warning level that is bound to trigger during upgrades as we upgrade one node at a time and desired number of healthy pods match current number of healthy pods as maxUnavailable is set to 1",
```
bparees, line 93: i'm ok with this as a fast fix to get CI running again, but i don't consider this a fix for the fundamental bug(that the alert fires during upgrade) because we are trying to reduce the number of alerts that fire during upgrades because it causes customer pain/confusion.

we need to find a way to avoid this alert entirely during upgrades, not just mask it in our e2e.

test/e2e/upgrade/alert/alert.go
```diff
@@ -153,6 +159,9 @@ count_over_time(ALERTS{alertstate="firing",severity!="info",alertname!~"Watchdog
 		violation := fmt.Sprintf("alert %s fired for %s seconds with labels: %s", series.Metric["alertname"], series.Value, helper.LabelsAsSelector(labels))
 		if cause := firingAlertsWithBugs.Matches(series); cause != nil {
 			knownViolations.Insert(fmt.Sprintf("%s (open bug: %s)", violation, cause.Text))
+		} else if cause = allowedFiringAlerts.Matches(series); cause != nil {
```
smarterclayton, line 162: No, please open a bug.  The alert is no different than any other alert.  You should be adding the bug that fixes this to the section above.

test/e2e/upgrade/alert/alert.go
smarterclayton: I think we should revert the alert for now, instead of working around it here.  Alerts shouldn't fire during normal upgrades

/hold

soltysh: Revert in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/533

ravisantoshgudimetla:
/close

As we decided to tune the metric instead of modifying the test.

PR #26215, opened 2021-06-09
add check of lease length to allow 60s of kube-apiserver  communication disruption

deads2k:
We cannot find them perfectly, but we can find those that are obviously broken.

There will need to be a library-go PR.

### DIFF
--- a/test/extended/operators/leases.go
+++ b/test/extended/operators/leases.go
@@ -1,0 +1,81 @@
+package operators
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = g.Describe("[sig-arch] Leases", func() {
+	defer g.GinkgoRecover()
+
+	g.It("should be able to span 60s kube-apiserver disruption", func() {
+		ctx := context.Background()
+
+		kubeClient, err := e2e.LoadClientset()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		shortLeases := []string{}
+
+		configMaps, err := kubeClient.CoreV1().ConfigMaps("").List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		for _, configmap := range configMaps.Items {
+			leaderElection, ok := configmap.Annotations[resourcelock.LeaderElectionRecordAnnotationKey]
+			if !ok {
+				continue
+			}
+			leaderElectionRecord := &resourcelock.LeaderElectionRecord{}
+			if err := json.Unmarshal([]byte(leaderElection), leaderElectionRecord); err != nil {
+				o.Expect(err).NotTo(o.HaveOccurred())
+			}
+			if leaderElectionRecord.LeaseDurationSeconds < 107 {
+				shortLeases = append(shortLeases, fmt.Sprintf("configmap/%s used by %q, has too short a lease (%d) to span 60s kube-apiserver disruption.  Try 107s leaseDuration with 13s retryPeriod and a 85s renewDeadline.  Be sure you have the graceful release properly wired.", configmap.Name, leaderElectionRecord.HolderIdentity, leaderElectionRecord.LeaseDurationSeconds))
+			}
+		}
+
+		endpoints, err := kubeClient.CoreV1().Endpoints("").List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		for _, endpoint := range endpoints.Items {
+			leaderElection, ok := endpoint.Annotations[resourcelock.LeaderElectionRecordAnnotationKey]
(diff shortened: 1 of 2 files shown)


test/extended/operators/leases.go
```diff
@@ -0,0 +1,77 @@
...
+		for _, configmap := range configMaps.Items {
+			leaderElection, ok := configmap.Annotations[resourcelock.LeaderElectionRecordAnnotationKey]
+			if !ok {
+				continue
+			}
+			leaderElectionRecord := &resourcelock.LeaderElectionRecord{}
+			if err := json.Unmarshal([]byte(leaderElection), leaderElectionRecord); err != nil {
+				o.Expect(err).NotTo(o.HaveOccurred())
+			}
+			if leaderElectionRecord.LeaseDurationSeconds < 90 {
```
romfreiman, line 40: Use 90 as consts.
Didnt we say 60?

test/extended/operators/leases.go
```diff
@@ -0,0 +1,77 @@
+package operators
+
+import (
```
romfreiman, line 3: do we want to start from exclude list so the test will be green?

test/extended/operators/leases.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			if leaderElectionRecord.LeaseDurationSeconds < 90 {
+				shortLeases = append(shortLeases, fmt.Sprintf("configmap/%s used by %q, has too short a lease to span 60s kube-apiserver disruption.  Try 99s leaseDuration with 13s retryPeriod and a 17s renewDeadline.  Be sure you have the graceful release properly wired.", configmap.Name, leaderElectionRecord.HolderIdentity))
+			}
+		}
+
+		endpoints, err := kubeClient.CoreV1().Endpoints("").List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		for _, endpoint := range endpoints.Items {
+			leaderElection, ok := endpoint.Annotations[resourcelock.LeaderElectionRecordAnnotationKey]
```
romfreiman, line 48: maybe to have this as an function and call it both for cm and andpoints?

test/extended/operators/leases.go
```diff
@@ -0,0 +1,77 @@
...
+			if leaderElectionRecord.LeaseDurationSeconds < 90 {
+				shortLeases = append(shortLeases, fmt.Sprintf("configmap/%s used by %q, has too short a lease to span 60s kube-apiserver disruption.  Try 99s leaseDuration with 13s retryPeriod and a 17s renewDeadline.  Be sure you have the graceful release properly wired.", endpoint.Name, leaderElectionRecord.HolderIdentity))
+			}
+		}
+
+		leases, err := kubeClient.CoordinationV1().Leases("").List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		for _, lease := range leases.Items {
+			if lease.Spec.LeaseDurationSeconds != nil && *lease.Spec.LeaseDurationSeconds < 90 {
+				shortLeases = append(shortLeases, fmt.Sprintf("configmap/%s used by %q, has too short a lease to span 60s kube-apiserver disruption.  Try 99s leaseDuration with 13s retryPeriod and a 17s renewDeadline.  Be sure you have the graceful release properly wired.", lease.Name, lease.Spec.HolderIdentity))
```
romfreiman, line 65: leases, not configmaps

test/extended/operators/leases.go
```diff
@@ -0,0 +1,81 @@
...
+
+	g.It("should be able to span 60s kube-apiserver disruption", func() {
+		ctx := context.Background()
+
+		kubeClient, err := e2e.LoadClientset()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		shortLeases := []string{}
+
+		configMaps, err := kubeClient.CoreV1().ConfigMaps("").List(ctx, metav1.ListOptions{})
```
p0lyn0mial, line 29: how about using a paginated list instead?

test/extended/operators/leases.go
```diff
@@ -0,0 +1,81 @@
...
+			leaderElectionRecord := &resourcelock.LeaderElectionRecord{}
+			if err := json.Unmarshal([]byte(leaderElection), leaderElectionRecord); err != nil {
+				o.Expect(err).NotTo(o.HaveOccurred())
+			}
+			if leaderElectionRecord.LeaseDurationSeconds < 107 {
+				shortLeases = append(shortLeases, fmt.Sprintf("endpoint/%s used by %q, has too short a lease (%d) to span 60s kube-apiserver disruption.  Try 107s leaseDuration with 13s retryPeriod and a 85s renewDeadline.  Be sure you have the graceful release properly wired.", endpoint.Name, leaderElectionRecord.HolderIdentity, leaderElectionRecord.LeaseDurationSeconds))
+			}
+		}
+
+		leases, err := kubeClient.CoordinationV1().Leases("").List(ctx, metav1.ListOptions{})
```
p0lyn0mial, line 61: it looks like `kube-apiservers` use the lease, we could filter them out

romfreiman:
@deads2k is this test will run as part of the fips suite?

romfreiman:
@deads2k I assume you have to update the values according to the enhancement?
https://github.com/openshift/enhancements/pull/832/files#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fR183

romfreiman:
@deads2k dont we want to merge this test?

PR #26218, opened 2021-06-10
bump(k8s)

tkashem:


tkashem:
we don't need it, current master already has client-go `watch` and `Stream` retry

/close

PR #26220, opened 2021-06-10
test/e2e/upgrade/alert: Allow AggregatedAPIDown to unblock 4.7->4.8 CI

wking:
We're getting:

    alert AggregatedAPIDown fired for 210 seconds with labels: {name="v1beta1.metrics.k8s.io", namespace="default", severity="warning"}

and such pretty consistently in those jobs.  Tracked in [rhbz#1970624][1].  Until that gets fixed, ignore the alert, so we are more likely to notice other breakage.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1970624

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -59,6 +59,10 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=1939580",
 		},
 		{
+			Selector: map[string]string{"alertname": "KubePodNotReady", "namespace": "openshift-kube-apiserver-operator"},
+			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=1939580",
+		},
+		{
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade/alert/alert.go
```diff
@@ -143,7 +143,7 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 	// Invariant: No non-info level alerts should have fired during the upgrade
 	firingAlertQuery := fmt.Sprintf(`
 sort_desc(
-count_over_time(ALERTS{alertstate="firing",severity!="info",alertname!~"Watchdog|AlertmanagerReceiversNotConfigured"}[%[1]s:1s])
+count_over_time(ALERTS{alertstate="firing",severity!="info",alertname!~"Watchdog|AlertmanagerReceiversNotConfigured|AggregatedAPIDown"}[%[1]s:1s])
```
smarterclayton, line 146: Please make this a known violation rather than an exclusion, and add more label selectors

wking: Done in a965a20 -> ea88b65.  Would be good to tighten this down with minor-update `Matches`, which I'm still working on, but I dunno if we need to block on that.

PR #26222, opened 2021-06-14
[3.11] Bug 1970977: UPSTREAM: multiple: Fix corruption of FibreChannel volumes

jsafrane:
There are two upstream PRs, a bot will link them. The code is in different files in 3.11 than in 4.x, however, the backport went quite smooth.

Fix corruption of FS devices when a pod is deleted when kubelet is not running. The volume reconstruction in this case must not fail, otherwise the volume is marked as unused and can be mounted on another node, corrupting the filesystem on it.

Therefore try hard not to fail 1) volume reconstruction and 2) unmounting the volume on error.

Upstream cherry pick to 1.21: kubernetes/kubernetes#102656


jsafrane:
I tested this on a 3.11 cluster and I was not able to reproduce `could not get consistent content of /proc/mounts after X attempts` (either with or without this PR), even with hundreds of processes constantly mounting + unmounting tmpfs in parallel. At least it did not make things worse, an iSCSI volume disguised as FC was always unmounted + multipath device flushed + removed. No corrupted filesystem found.

bparees:
/override ci/openshift-jenkins-cmd
/override ci/openshift-jenkins/extended_clusterup

failing for unrelated reasons

jsafrane:
Override does not help, re-testing
/test ci/openshift-jenkins/extended_clusterup
/test ci/openshift-jenkins/cmd

jsafrane:
/test extended_clusterup
/test cmd

does it work on overriden test?

jsafrane:
Rebased to today's release-3.11, hoping to get fresh tests...

PR #26223, opened 2021-06-14
Bug 1971590: Enable TestEndpointAdmission test only for OpenShift SDN

ardaguclu:
This PR applies only enabling TestEndpointAdmission test for OpenShift SDN

### DIFF
--- a/test/extended/networking/endpoint_admission.go
+++ b/test/extended/networking/endpoint_admission.go
@@ -24,16 +24,12 @@ var _ = g.Describe("[sig-network][endpoints] admission", func() {
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLI("endpoint-admission")
 
-	g.It("TestEndpointAdmission", func() {
-		TestEndpointAdmission(g.GinkgoT(), oc)
+	InOpenShiftSDNContext(func() {
+		g.It("blocks manual creation of Endpoints pointing to the cluster or service network", func() {
+			TestEndpointAdmission(g.GinkgoT(), oc)
+		})
 	})
 })
-
-var exampleAddresses = map[string]string{
-	"cluster":  "10.128.0.2",
-	"service":  "172.30.0.2",
-	"external": "1.2.3.4",
-}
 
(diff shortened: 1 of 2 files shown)


test/extended/networking/endpoint_admission.go
```diff
@@ -29,10 +30,9 @@ var _ = g.Describe("[sig-network][endpoints] admission", func() {
 	})
 })
 
-var exampleAddresses = map[string]string{
-	"cluster":  "10.128.0.2",
-	"service":  "172.30.0.2",
-	"external": "1.2.3.4",
+var exampleServiceAddresses = map[string]string{
```
dcbw, line 33: @ardaguclu I would move this into the TestEndpointAdmission() function since it's only used there and it's confusing to have it as a global when testOne() also creates a map of the same name, but in local scope.

dcbw: In fact, why even bother with the map? Why not just assign the addresses directly since exampleServiceAddresses is only used in one place?

ardaguclu: Thanks for review @dcbw, I changed according to your suggestions.

danwinship (changes requested):
There was discussion about this test somewhere else but I don't remember where.

This is only relevant for openshift-sdn in Multitenant mode, and thus there's no need to test it under IPv6. The test should just be wrapped in `InOpenShiftSDNContext()` so it gets skipped for other plugins.

ardaguclu:
Thanks @danwinship, I did changes and only enable this test for OpenShift sdn is enabled cases.

test/extended/networking/endpoint_admission.go
```diff
@@ -24,17 +24,13 @@ var _ = g.Describe("[sig-network][endpoints] admission", func() {
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLI("endpoint-admission")
 
-	g.It("TestEndpointAdmission", func() {
-		TestEndpointAdmission(g.GinkgoT(), oc)
+	InOpenShiftSDNContext(func() {
+		g.It("TestEndpointAdmission", func() {
```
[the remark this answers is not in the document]
ardaguclu: Done. Thanks.

PR #26224, opened 2021-06-14
Mention BZs of affected APIs in 'clients should not use APIs that are removed in upcoming releases' e2e

ingvagabund:
To avoid repeated assesments of why the APIs are still in use.


ingvagabund:
Once https://github.com/openshift/kubernetes/pull/862 gets merged, this PR can be closed.

PR #26226, opened 2021-06-14
Disable quorum restore dr test

marun:
The machine approver does not reliably approve new nodes after restore at present, and the quorum restore test failing in response prevents getting signal of the regular restore case.


marun:
/hold 

Waiting for signal.

marun:
aws disruptive is passing, gcp not so much.

/test e2e-gcp-disruptive

marun:
Not sure if the other failures are infra or master being unstable.

PR #26228, opened 2021-06-15
add test checking ns/pods for annotations allowing/preferring management cores

jerpeter1:
…ent cores


jerpeter1:
Deprecating this in favor of https://github.com/openshift/origin/pull/27770

PR #26231, opened 2021-06-15
Bug 1953127: undisable NetworkPolicy tests

danwinship:
Previously #26197; I had closed that because someone else did the rebase from o/k, but I forgot that we still needed the commit fixing rules.go


knobunc:
It's perma-failing, and not required...

danwinship:
/skip
(to make prow notice that e2e-gcp-builds is no longer needed)

PR #26236, opened 2021-06-16
Bug 1963160: don't iterate over pod objects

aojea:
Bug 1963160 shows that this function is the more expensive:
```
Showing top 5 nodes out of 97

flat  flat%   sum%        cum   cum%
18620ms 55.62% 55.62%    18620ms 55.62%  runtime.duffcopy
4700ms 14.04% 69.65%     7420ms 22.16%  runtime.mapiternext
1870ms  5.59% 75.24%     1870ms  5.59%  runtime.aeshashbody
850ms  2.54% 77.78%    28180ms 84.17%  github.com/openshift/origin/pkg/network/node.(*networkPolicyPlugin).selectPods
780ms  2.33% 80.11%      780ms  2.33%  runtime.memeqbody
```

In order to avoid excessive copies of data better to iterate over the key of the map so we don't have to copy the value in each iteration.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -406,9 +406,21 @@ func (np *networkPolicyPlugin) selectPods(npns *npNamespace, lsel *metav1.LabelS
 		utilruntime.HandleError(fmt.Errorf("ValidateNetworkPolicy() failure! Invalid PodSelector: %v", err))
 		return ips
 	}
-	for _, pod := range np.pods {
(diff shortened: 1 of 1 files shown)


aojea:
/hold
This is in incorrect

aojea:
/hold cancel
pods are a map not an slice
```
	pods map[ktypes.UID]kapi.Pod
```

aojea:
I should be doing something wrong here because the verify jobs is failing here and it doesn't fail in other places

aojea:
/close
in favor of #26247

PR #26247, opened 2021-06-18
Bug 1963160: don't iterate over pod objects

aojea:
Bug 1963160 shows that this function is the more expensive:
Showing top 5 nodes out of 97

flat  flat%   sum%        cum   cum%
18620ms 55.62% 55.62%    18620ms 55.62%  runtime.duffcopy
4700ms 14.04% 69.65%     7420ms 22.16%  runtime.mapiternext
1870ms  5.59% 75.24%     1870ms  5.59%  runtime.aeshashbody
850ms  2.54% 77.78%    28180ms 84.17%  github.com/openshift/origin/pkg/network/node.(*networkPolicyPlugin).selectPods
780ms  2.33% 80.11%      780ms  2.33%  runtime.memeqbody

In order to avoid excessive copies of data better to iterate over the key instead of the value.

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -406,9 +406,11 @@ func (np *networkPolicyPlugin) selectPods(npns *npNamespace, lsel *metav1.LabelS
 		utilruntime.HandleError(fmt.Errorf("ValidateNetworkPolicy() failure! Invalid PodSelector: %v", err))
 		return ips
 	}
-	for _, pod := range np.pods {
-		if (npns.name == pod.Namespace) && sel.Matches(labels.Set(pod.Labels)) {
-			ips = append(ips, pod.Status.PodIP)
+
+	// Iterate over the key instead of the value for performance reasons
+	for k := range np.pods {
+		if (npns.name == np.pods[k].Namespace) && sel.Matches(labels.Set(np.pods[k].Labels)) {
+			ips = append(ips, np.pods[k].Status.PodIP)
 		}
 	}
 	return ips


pkg/network/node/networkpolicy.go
```diff
@@ -406,9 +406,11 @@ func (np *networkPolicyPlugin) selectPods(npns *npNamespace, lsel *metav1.LabelS
 		utilruntime.HandleError(fmt.Errorf("ValidateNetworkPolicy() failure! Invalid PodSelector: %v", err))
 		return ips
 	}
-	for _, pod := range np.pods {
-		if (npns.name == pod.Namespace) && sel.Matches(labels.Set(pod.Labels)) {
-			ips = append(ips, pod.Status.PodIP)
+
+	// Iterate over the key instead of the value for performance reasons
+	for k := range np.pods {
```
danwinship, line 411: hm... would it be better to make `np.pods` be a map of `*kapi.Pod` rather than `kapi.Pod` instead? As it is now you're dereferencing `np.pods[k]` three times which can't be great?

In 4.x we dropped `np.pods` since the informer already has its own cache (https://github.com/openshift/sdn/commit/4bac7c7a568937dfe4680bb8d8357b8093ec616d + https://github.com/openshift/sdn/commit/eaf68af70cade6ea53663a0e24843149e9c63907) but that also required rewriting the unit tests to use a fake kubeclient (https://github.com/openshift/sdn/commit/2e60296cc8a78033234129b8a6e84e56b7dfed3a) and I'm not sure how much divergence there's been since 3.11 so that might be too much to backport...

aojea: I took the easy path because I really don't know this code

Let me try to check how difficult are the backports

aojea: the patches doesn't apply directly ... it is not straight forward to backport all that stuff

PR #26250, opened 2021-06-18
Update ruby-hello-world images

soltysh:
React to changes from openshift/ruby-hello-world#125

/assign @mfojtik


aojea:
/lgtm
not familiar with the code but the tests now pass, and these only change the failing tests

PR #26256, opened 2021-06-21
Bug 1919453: Resolve e2e test image pull rate limit

rvanderp3:
The intent of this PR is to complete the work begun in https://github.com/openshift/origin/pull/25821 .


rvanderp3:
There is one remaining test that is pulling busybox from docker.io.  I'm investigating.  It is possible that we could merge this and the 4.6 CI signals would improve just due to the decreased likelihood of exhausting the rate limit.

PR #26258, opened 2021-06-22
Bug 1963160: Backport Bug 1896958: NetworkPolicy performance (pod caching) #226

aojea:
Backport network policy performance changes from 4.x, adapting the code to 3.11

https://github.com/openshift/origin/pull/26247#discussion_r654645326

> In 4.x we dropped np.pods since the informer already has its own cache (openshift/sdn@4bac7c7 + openshift/sdn@eaf68af) but that also required rewriting the unit tests to use a fake kubeclient (openshift/sdn@2e60296) and I'm not sure how much divergence there's been since 3.11 so that might be too much to backport...

### DIFF
--- a/pkg/network/node/networkpolicy.go
+++ b/pkg/network/node/networkpolicy.go
@@ -43,8 +43,6 @@ type networkPolicyPlugin struct {
 	namespaces map[uint32]*npNamespace
 	// nsMatchCache caches matches for namespaceSelectors; see selectNamespaceInternal
 	nsMatchCache map[string]*npCacheEntry
-
-	pods map[ktypes.UID]kapi.Pod
 }
 
 // npNamespace tracks NetworkPolicy-related data for a Namespace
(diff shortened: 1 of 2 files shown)


pkg/network/node/networkpolicy.go
```diff
@@ -64,7 +68,10 @@ type npNamespace struct {
 type npPolicy struct {
 	policy            networking.NetworkPolicy
 	watchesNamespaces bool
-	watchesPods       bool
+	// not used in 3.11
+	// added in https://github.com/openshift/sdn/commit/3c35081bda7f4e0be1ae0f5e85f3ad3b3e6d1224
```
danwinship, line 72: yeah, this depends on other changes that weren't backported to 3.11. I'd drop `networkpolicy: consolidate OVS transactions` and `networkpolicy: do policy recalculation in the bfr thread to avoid blocking`. Basically, take the patches involved in getting rid of `np.pods`, but not the rest, at least for now.

danwinship (changes requested):

pkg/network/node/networkpolicy.go
aojea: will do

aojea:
@danwinship it seems jobs that are ok are passing

PR #26259, opened 2021-06-22
Bug 1928862: [4.7] Bump openshift/kubernetes 

jsafrane:
Updated to today's head of release-4.7 of openshift/kubernetes. To get updated test annotation with enabled VolumeSnapshot tests.
 
Regenerated test tags in a separate commit.


jsafrane:
`[Feature:APIServer][Late] kubelet terminates kube-apiserver gracefully` is flaky, but it was flaky even before this PR. I don't think it's related to this PR.
/retest

And `ci/prow/artifacts` is something new that never succeeded, at least in 4.7.

PR #26261, opened 2021-06-22
Bug 1983614: Add renamed Prometheus test into skipped disconnected list

ardaguclu:
This PR updates `Skipped:Disconnected` list with the renamed test.

Previously;
`[sig-instrumentation][Late] Alerts shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured`

Currently;
`[sig-instrumentation][Late] Alerts shouldn't report any alerts in firing or pending state apart from Watchdog and AlertmanagerReceiversNotConfigured and have no gaps in Watchdog firing`

### DIFF
--- a/test/extended/util/annotate/rules.go
+++ b/test/extended/util/annotate/rules.go
@@ -166,7 +166,7 @@ var (
 			`\[sig-instrumentation\] Prometheus when installed on the cluster when using openshift-sdn should be able to get the sdn ovs flows`,
 			`\[sig-instrumentation\]\[Late\] Alerts should have a Watchdog alert in firing state the entire cluster run`,
 			`\[sig-instrumentation\]\[Late\] Alerts shouldn't exceed the 500 series limit of total series sent via telemetry from each cluster`,
-			`\[sig-instrumentation\]\[Late\] Alerts shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured`,
+			`\[sig-instrumentation\]\[Late\] Alerts shouldn't report any alerts in firing or pending state apart from Watchdog and AlertmanagerReceiversNotConfigured and have no gaps in Watchdog firing`,
 			`\[sig-instrumentation\]\[sig-builds\]\[Feature:Builds\] Prometheus when installed on the cluster should start and expose a secured proxy and verify build metrics`,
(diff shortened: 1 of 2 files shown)


deads2k:
I lack some background here.  Why should alerts be allowed to fire in the disconnected case?

If we can detect this, why not simply determine which alerts can fire as opposed to those that shouldn't?

ardaguclu:
@deads2k, tests under the ICSP section try to pull `image-registry.openshift-image-registry.svc:5000/openshift/tools:latest` image but they can not accomplish that in disconnected as filed in https://bugzilla.redhat.com/show_bug.cgi?id=1918376 (which seems verified but not backported to 4.8 yet.). Thereby, this Prometheus test in this PR could not go beyond the first step of the test which create a test pod.

As you said, if it did not have any image pull problem, it would be expected to not fire alerts even in disconnected.

test/extended/util/annotate/rules.go
```diff
@@ -166,7 +166,7 @@ var (
 			`\[sig-instrumentation\] Prometheus when installed on the cluster when using openshift-sdn should be able to get the sdn ovs flows`,
 			`\[sig-instrumentation\]\[Late\] Alerts should have a Watchdog alert in firing state the entire cluster run`,
 			`\[sig-instrumentation\]\[Late\] Alerts shouldn't exceed the 500 series limit of total series sent via telemetry from each cluster`,
-			`\[sig-instrumentation\]\[Late\] Alerts shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured`,
+			`\[sig-instrumentation\]\[Late\] Alerts shouldn't report any alerts in firing or pending state apart from Watchdog and AlertmanagerReceiversNotConfigured and have no gaps in Watchdog firing`,
```
smarterclayton, line 169: Are we still firing alerts in disconnected mode?  Or is this because we can't access the test?  This is such a fundamental safety net that the fact that we shipped 4.8 with this failing means we have a huge knowledge gap as to whether disconnected is safe.

Is there a story covering reenabling this test because "we've fixed this"?  I see there's a comment up above for "ICSP" - isn't that bug fixed?  In the future, please add bugzilla links to these sections that have a bug we can track for "this problem is now fixed".

PR #26263, opened 2021-06-23
Bug 1973667: undisable NetworkPolicy tests [4.8]

danwinship:
backport of #26231 
/assign @knobunc

PR #26266, opened 2021-06-24
Bug 1980141: Skip the new "NetPol" tests for now

danwinship:
The test

`[sig-network] Netpol [LinuxOnly] NetworkPolicy between server and client should stop enforcing policies after they are deleted [Feature:NetworkPolicy] [Skipped:Network/OpenShiftSDN/Multitenant] [Skipped:Network/OpenShiftSDN] [Suite:openshift/conformance/parallel] [Suite:k8s]`

assumes that after deleting a NetworkPolicy, the system will have been updated to reflect the new state within 3 seconds. This fails sometimes in CI, especially with ovn-kubernetes. The test needs to be fixed to do a retry loop (like every other NP test!)

(I've already added this commit to #26263 for 4.8 as well.)

**EDIT**: Actually, the entire "Netpol" suite is broken, that test is just more broken than everything else.

/assign @knobunc


yselkowitz:
Any reason not to prefix the subject with `Bug 1975865:` so as to attach this to the BZ and allow better tracking in sippy etc.?

danwinship:
It's not _fixing_ 1975865 though.

Well... I guess there are two problems here:
1. there is a broken e2e test case that needs to be fixed upstream and then backported
2. the broken e2e test case is causing CI failures

Linking this PR to 1975865 would be saying that 1975865 represents the second part, but then I'd need to create a separate bz for the first part so that it could be linked from `rules.go`. But since we don't actually _need_ a bz for the "CI failures" part, I just made the existing bz be for the "needs an upstream fix" part instead.

danwinship:
/cc @astoycos 
ok, I already had a PR to disable the NetPol tests, I had just been hoping not to merge it...

astoycos:
/lgtm 

Leaving the bug open is a must, we do not want to leave these tests skipped.  Hopefully https://github.com/kubernetes/kubernetes/issues/103802 will help the resource consumption issues once it's complete

astoycos:
Leaving the bug open is a must, we do not want to leave these tests skipped. Hopefully kubernetes/kubernetes#103802 will help the resource consumption issues once it's complete

soltysh (approved):
/approve

PR #26273, opened 2021-06-25
clusterup.sh lacks creds in CI

brenton:
Normally 'oc cluster up' looks in $HOME/.docker/config and wires that up
for imagestream imports.  That isn't happening in the clusterup.sh test
since it does clever things with $HOME.


brenton:
I debugged at least one of the problems with clusterup.sh.  The job passed. I'm going to green button this to hopefully unblock more important PRs.

PR #26072, opened 2021-04-13
Use UBI8 instead of SCL images

yselkowitz:
This allows tests to pass on other architectures, which do not have RHEL
7 and/or CentOS container content.

Goes together with https://github.com/openshift/ruby-hello-world/pull/125

/cc @gabemontero @adambkaplan @soltysh @bparees

PR #26278, opened 2021-06-28
builds: comment out RHEL-only sample-pipeline template test

yselkowitz:
This test relies on pulling yaml from github rather than from bindata,
which makes it difficult to pre-commit testing of changes which affect
it.  @akram and @gabemontero will move this to the jenkins client plugin
E2E, but in the meantime, we comment it out.

https://github.com/openshift/origin/pull/26072#issuecomment-862307401

/cc @gabemontero


gabemontero:
/lgtm

some further clarification - this RHEL only file is the intended regression testing subset to run on multi arch to sniff test jenkins on those arch's, where we only use images based on RHEL (which is a restriction we do not have on x86)

the remaining set of tests here are still a sufficient sniff test

PR #26285, opened 2021-06-29
Bug 1977454: Use nodejs to test service connection

adambkaplan:
Redis has tighter constraints on client HTTP calls to prevent malicious
attacks. Only official redis clients are guaranteed to receive reliable
HTTP responses. The build test for in-cluster service connections does
not use a redis client, resulting in a test HTTP curl request receiving an
empty repsonse.

This updates the in-cluster service test to use a "hello world" nodejs
application and service. The use of the upstream Redis image is also
removed.

### DIFF
--- a/test/extended/builds/service.go
+++ b/test/extended/builds/service.go
@@ -10,7 +10,6 @@ import (
 	kapierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kdeployutil "k8s.io/kubernetes/test/e2e/framework/deployment"
-	k8simage "k8s.io/kubernetes/test/utils/image"
 
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/image"
@@ -23,7 +22,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds] build can reference a cluster s
 		testDockerfile = fmt.Sprintf(`
 FROM %s
 RUN cat /etc/resolv.conf
-RUN curl -vvv hello-openshift:6379
+RUN curl -vvv hello-nodejs:8080
 `, image.ShellImage())
 	)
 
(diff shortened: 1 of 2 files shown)


test/extended/builds/service.go
```diff
@@ -43,14 +42,20 @@ RUN curl -vvv hello-openshift:6379
 
 		g.Describe("with a build being created from new-build", func() {
 			g.It("should be able to run a build that references a cluster service", func() {
-				g.By("standing up a new hello world service")
-				err := oc.Run("new-app").Args("--name", "hello-openshift", k8simage.GetE2EImage(k8simage.Redis)).Execute()
+				g.By("standing up a new hello world redis service")
```
wking, line 45: I don't understand why we are adding the `redis` string here, when the rest of this PR seems to be removing Redis from the test-case.

adambkaplan: an oversight on my part...

test/extended/builds/service.go
```diff
@@ -43,14 +42,20 @@ RUN curl -vvv hello-openshift:6379
 
 		g.Describe("with a build being created from new-build", func() {
 			g.It("should be able to run a build that references a cluster service", func() {
-				g.By("standing up a new hello world service")
-				err := oc.Run("new-app").Args("--name", "hello-openshift", k8simage.GetE2EImage(k8simage.Redis)).Execute()
+				g.By("standing up a new hello world nodejs service via oc new-app")
+				err := oc.Run("new-app").Args("nodejs~https://github.com/sclorg/nodejs-ex.git", "--name", "hello-nodejs").Execute()
```
gabemontero, line 46: I've occasionally been curious on the side how the upcoming disconnected tests are going to deal with git cloning from the outside :-)

wking: Can this be "we build the image on the outside with the Git bits already installed, and then mirror the built image into the restricted-network environment before running the tests"?  That would make cloning-from-Git or other on-the-fly internet access technical debt that we'd have to clean out before we had the test-case green in restricted-networks where that internet access is not available.

gabemontero: can - yes
do it in relatively minimal cycles like this change - no .... a much much much more involved change based on other recent build test image work we have had to wrt the disconnected e2e stuff

short term, e2e-aws-proxy needs to be unblocked

but as I mentioned to @adambkaplan in slack, our team needs to huddle up at some point and see how we will deal with the git cloning elements of our e2e's (which aside from testing git clone, test several of the upstream s2i base image implementations as part of vetting those)

adambkaplan: That's going to be a much bigger endeavor. "Clone source from git over the internet" is the most basic thing OCP builds do, and we do it everywhere in the build suite.

Is having a fully air-gapped CI test a goal? Or is the goal having CI tests run on a cluster that is restricted, but with some caveats. Example - proxy in a DMZ that has an allowlist of domains to access.

PR #26288, opened 2021-06-30
[release-4.6] Bug 1977093: Add consolequickstarts and consoleplugins extensions to the authenticated rules

jhadvig:
/assign @bparees


jhadvig:
@bparees @spadgett is there a way how we could override the labels here, since for all the later versions we dont need this change, just for 4.6.

bparees:
yeah i'll override

/lgtm

bparees:
will still need to wait for cherry pick approval but hopefully that happens today (i'm not patch manager this week)

PR #26293, opened 2021-06-30
Bug 1977097: Improve GC Check for Builds

adambkaplan:
When checking that Build objects are garbage collected after their
parent BuildConfig was deleted, we used a crude means of verifying that
associated ConfigMaps were deleted. This improves the test by checking
that any remaining ConfigMaps in the namespace do not have an owner
reference.

### DIFF
--- a/test/extended/builds/remove_buildconfig.go
+++ b/test/extended/builds/remove_buildconfig.go
@@ -2,6 +2,7 @@ package builds
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	g "github.com/onsi/ginkgo"
@@ -31,6 +32,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds] remove all builds when build co
 
 		g.AfterEach(func() {
 			if g.CurrentGinkgoTestDescription().Failed {
+				exutil.DumpBuilds(oc)
 				exutil.DumpPodStates(oc)
 				exutil.DumpConfigMapStates(oc)
 				exutil.DumpPodLogsStartingWith("", oc)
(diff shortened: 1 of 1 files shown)


adambkaplan:
/assign @alicerum 

This is an (old) test that checks that everything related to a Build gets cleaned up in a reasonable time period when the parent BuildConfig is deleted. In past OCP releases, we've observed k8s GC get trapped which led to build objects, pods, and generated ConfigMaps being orphaned.

I suspect what is happening in some of our narrower test cases is as follows:

1. New project is created
2. The initial count of ConfigMaps is checked, which is zero
3. Kubernetes (and other controllers) add ConfigMaps. One of them is `kube-root.ca`
4. The builds start, run, and are then deleted.
5. We check to see if there are any ConfigMaps left. `kube-root.ca` is left behind, because it is system generated.

I use owner references as a proxy to check if a ConfigMap is owned by a Build pod. IMO we would be better off if we added labels to the ConfigMaps generated by the build, and simply looked for ConfigMaps that had our label key. But that is a bit of tech debt we can take on separately.

xiuwang:
The test suite passed https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26293/pull-ci-openshift-origin-master-e2e-aws-fips/1410315650714308608
/label qe-approved

coreydaley:
I would like to see the test differentiate between whether it was the builds that failed to be removed, or the associated configmaps. Maybe have an `err` variable outside of the `wait.Poll` that gets updated inside of that block so we know what failed?

coreydaley:
Or a couple of variables that let you expose the `Items` count for the builds and the configmaps that can be appended to the end of the `timed out` error.

adambkaplan:
@coreydaley updated so on timeout we log the number of builds and ConfigMaps left.

sdodson:
Frequently failing test and the failures look unrelated to this change

PR #26302, opened 2021-07-01
Bug 1980338: test/cmd/newapp: update app name to match existing tag

vrutkovs:
Fixes several `test/cmd` entries to match the current state of imagestreams.
Includes #26292 to have Dockerhub pullsecret placed, so that import would not hit rate limits

### DIFF
--- a/hack/test-cmd.sh
+++ b/hack/test-cmd.sh
@@ -34,6 +34,30 @@ function find_tests() {
 }
 tests=( $(find_tests ${1:-.*}) )
 
+# install_image_pull_secret will look for ~/docker/config.json and create an image pull secret.  If no suitable credentials are found a warning will be displayed.  It assumes a running server.
+#
+# Globals:
+#  - ADMIN_KUBECONFIG
+#  - REAL_HOME
+# Arguments:
+#  -  namespace (optional)
+# Returns:
+#  None
+function install_image_pull_secret() {
+	local ns="${1:-}"
+	local dockerconfig="${REAL_HOME}/.docker/config.json"
+
+	if [[ -f $dockerconfig ]]; then
+		os::log::debug "Installing imagestreamsecret..."
+		oc create secret generic imagestreamsecret --dry-run -o yaml --type=kubernetes.io/dockerconfigjson --from-file=.dockerconfigjson=${dockerconfig} -n ${ns} --config=$ADMIN_KUBECONFIG >& /tmp/secret.yaml
+		oc apply -f /tmp/secret.yaml -n ${ns} --config=$ADMIN_KUBECONFIG
+	else
+		os::log::debug "No docker credentials found.  Image pull secret will not be installed."
+	fi
+}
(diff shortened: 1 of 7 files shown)


test/cmd/newapp.sh
```diff
@@ -27,9 +27,9 @@ os::cmd::try_until_failure 'oc get istag php:latest -n openshift'
 os::cmd::expect_failure 'oc new-app --image-stream=openshift/php https://github.com/sclorg/cakephp-ex'
 
 # should succeed and create the php:latest tag in the current namespace
-os::cmd::expect_success 'oc new-app --docker-image=library/php https://github.com/sclorg/cakephp-ex --strategy=source'
-os::cmd::try_until_success 'oc get istag php:latest -n test-imagestreams'
-os::cmd::expect_success 'oc create istag php:latest --from=openshift/php:7.1 -n openshift'
+os::cmd::expect_success 'oc new-app --docker-image=quay.io/centos7/php-72-centos7 https://github.com/sclorg/cakephp-ex --strategy=source --name=php'
```
cuppett, line 30: If it helps, here's where the an official UBI imagestream comes from:

registry.access.redhat.com/ubi7/php-73

https://catalog.redhat.com/software/containers/ubi7/php-73/5da0228fdd19c70159f4601b

7.2 is already EOL, 7.3 is next and 7.4 is only published for ubi8 currently.

cuppett: I see in 9dc7e4bb750ea7f2e9339a39649e428cf6a68eda you are using ubi8. In that case, registry.access.redhat.com/ubi8/php-74 will give you the most legs. :)

bparees:
for the image pullspec changes themselves, as long as the test is still passing you are probably fine to change them, the tests are exercising the commands, not the images, so as long as the new image your ref'ing is a functional/behavioral equivalent to the old one, it should be fine.

hack/test-cmd.sh
```diff
@@ -34,6 +34,30 @@ function find_tests() {
 }
 tests=( $(find_tests ${1:-.*}) )
 
+# install_image_pull_secret will look for ~/docker/config.json and create an image pull secret.  If no suitable credentials are found a warning will be displayed.  It assumes a running server.
+#
+# Globals:
+#  - ADMIN_KUBECONFIG
+#  - REAL_HOME
+# Arguments:
+#  None
```
dcbw, line 43: @vrutkovs doesn't it take an optional namespace as an argument?

test/cmd/newapp.sh
vrutkovs: Updated quay.io images to registry.access.redhat.com. Its a mix of UBI7/UBI8 now, we can improve this later on

hack/test-cmd.sh
vrutkovs: Good catch, thanks. Fixed this in 6757e9f

soltysh (approved):
/lgtm
/approve

PR #26304, opened 2021-07-02
test/extended: Remove the failure allowed logic in the cluster topology SNO test

timflannagan:
Update the test/extended/single_node/topology.go single-node validation
logic that allows the PackageServer Deployment resource to be more than
one replica when running tests.

OLM will be moving that component from the CVO payload to a second-level
Operator that reads the Infrastructure cluster singleton resource and
propagates the HA/non-HA expectations to the PackageServer CSV resource.

### DIFF
--- a/test/extended/single_node/topology.go
+++ b/test/extended/single_node/topology.go
@@ -93,20 +93,13 @@ func isInfrastructureDeployment(deployment appsv1.Deployment) bool {
 	return false
 }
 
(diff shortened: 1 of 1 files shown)


timflannagan:
Blocked until https://github.com/openshift/operator-framework-olm#99 merges.

/hold

timflannagan:
The linked PR has merged - removing the hold.

/test all
/hold cancel

bparees:

one of these times it'll make it....

PR #26305, opened 2021-07-05
Allow HighlyAvailableWorkloadIncorrectlySpread alert to fire

dgrisonnet:
Allow the HighlyAvailableWorkloadIncorrectlySpread alert to fire until
we can safely enable hard anti-affinity on hostname to workload with
persistent storage, without breaking customers. The timeline for that
should be 4.9. Note that for upgrade tests, we will need to wait one
more release to remove the alert from the allow list since we will
need both the source version and the target to have hard anti-affinity
enabled to prevent the alert from firing.

I linked the alert scoped to prometheus-k8s to [bug 1929262](https://bugzilla.redhat.com/show_bug.cgi?id=1949262) and the 
one scoped to alertmanager-main to [bug 1955489](https://bugzilla.redhat.com/show_bug.cgi?id=1955489). These bugzillas 
track the effort of adding hard anti-affinity constraints to these
components.

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -96,6 +96,16 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 				return framework.ProviderIs("gce")
 			},
 		},
+		{
+			// Should be removed one release after the attached bugzilla is fixed.
+			Selector: map[string]string{"alertname": "HighlyAvailableWorkloadIncorrectlySpread", "namespace": "openshift-monitoring", "workload": "prometheus-k8s"},
+			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=1949262",
+		},
+		{
+			// Should be removed one release after the attached bugzilla is fixed.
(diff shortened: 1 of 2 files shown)


wking (approved):
/lgtm

wking:
Doesn't seem like it would be unique to updates, so probably add it [here][1] too?

[1]: https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L68

dgrisonnet:
Yes definitely :+1:.

test/e2e/upgrade/alert/alert.go
```diff
@@ -96,6 +96,11 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 				return framework.ProviderIs("gce")
 			},
 		},
+		{
+			// Should be removed one release after the attached bugzilla is fixed.
+			Selector: map[string]string{"alertname": "HighlyAvailableWorkloadIncorrectlySpread"},
```
smarterclayton, line 101: Can you scope it to the component that is impacted (add an additional label)?  Or does the alert not have an extra label?  A namespace?

dgrisonnet: Good point, I scoped them to the namespace/workload.

Since we have now have two different alerts allowed, I attached the relevant bugzilla to each of them.

PR #26307, opened 2021-07-05
WIP: Recover from panics in ClusterUpgrade test

fpetkovski:
Whenever there is a failure in the cluster upgrade test suite, the framework panics with the message:
```
Ginkgo panics to prevent subsequent assertions from running.
Normally Ginkgo rescues this panic so you shouldn't see it.
But, if you make an assertion in a goroutine, Ginkgo can't capture the panic.
To circumvent this, you should call
	defer GinkgoRecover()
```

This PR adds `GinkgoRecover()` to the cluster upgrade test to help Ginkgo recover from the panic and output a clean failure log message.

### DIFF
--- a/test/extended/util/disruption/disruption.go
+++ b/test/extended/util/disruption/disruption.go
@@ -316,6 +316,7 @@ func ExpectNoDisruption(f *framework.Framework, tolerate float64, total time.Dur
 	duration := events.Duration(0)
 	describe := events.Strings()
 	if percent := float64(duration) / float64(total); percent > tolerate {
+		defer g.GinkgoRecover()
 		framework.Failf("%s for at least %s of %s (%0.0f%%):\n\n%s", reason, duration.Truncate(time.Second), total.Truncate(time.Second), percent*100, strings.Join(describe, "\n"))
 	} else if duration > 0 {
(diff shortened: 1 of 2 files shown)


test/extended/util/disruption/disruption.go
```diff
@@ -187,6 +187,7 @@ func (cma *chaosMonkeyAdapter) Test(sem *chaosmonkey.Semaphore) {
 	cma.test.Setup(cma.framework)
 	defer cma.test.Teardown(cma.framework)
 	ready()
+	defer g.GinkgoRecover()
```
soltysh, line 190: I wouldn't expect `GinkgoRecover` in places other than just the body of the Ginkgo test, ie. inside `It`, `Describe`, `Before*` etc.

fpetkovski: That was also my assumption, but I was still seeing panics and wanted to see if adding a recover as close to the test would help.

soltysh: You should add those in the tests where you see those failures, if they don't have it, yet.

fpetkovski: I think all of the panics are all coming from the cluster upgrade test. I added a recover statement at the top of that test[1], but I don't think this solved the problem completely. Could the upgrade test spawn additional goroutines which would also need the recover call?

[1] https://github.com/openshift/origin/pull/26307/files#diff-a3e17408eaaa387d9a91030c6d3cd0fe5ad10f976a605dc1d081b56a57f79162R134

PR #26308, opened 2021-07-06
Bug 1977095: correcting test case

prashbnair:


bparees:
overriding newer release bug dep req since this is a test issue and it's easily confirmed that the code is correct in the newer branches

PR #26311, opened 2021-07-07
Create a new serial test suite with tests assuming high-availability removed

omertuc:
While running the serial test suite on single-node openshift clusters, some tests
add taints to the node, causing disruptions to tests that come after them. This commit
adds a new suite, duplicating the existing serial suite, but excluding the tests mentioned
above because they're not applicable in non-highly-available clusters

### DIFF
--- a/cmd/openshift-tests/e2e.go
+++ b/cmd/openshift-tests/e2e.go
@@ -91,6 +91,24 @@ var staticSuites = testSuites{
 	},
 	{
 		TestSuite: ginkgo.TestSuite{
+			Name: "openshift/conformance/serial-non-highly-available",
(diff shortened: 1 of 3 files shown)


omertuc:
Following tests removed:

```
meld <(./openshift-tests run openshift/conformance/serial-non-highly-available --dry-run) <(./openshift-tests run openshift/conformance/serial --dry-run) 
```
image

omertuc:
Does this make https://github.com/openshift/origin/pull/26311 obsolete? Should I close it?

omertuc:
Skip tests dynamically rather than creating a new suite https://github.com/openshift/origin/pull/26314

PR #26312, opened 2021-07-07
always discard v1beta1 in etcd test of apiextensions and  admissionregistration

atiratree:
a workaround had to be added to merge https://github.com/openshift/cluster-kube-apiserver-operator/pull/1162 - this is removing the workaround so that v1beta1 of apiextensions.k8s.io and admissionregistration.k8s.io are always discarded in etcd_storage_path test


marun:
@atiratree Maybe include context in the description so a reviewer and future maintainers can discern your intent? The title of the PR and your commit description describe the 'what' rather than the 'why', with the latter being essential to understanding why something  was or should be merged.

atiratree:
@marun sorry, you are right it isn't clear from that - updated

marun:
/lgtm 

@hexfusion Is the test in question really something that the etcd team should be maintaining? Seems more like an apiserver concern.

PR #26314, opened 2021-07-08
on SNO, skip disruptive or multinodes tests

osherdp:
Skipping disruptive tests on single-node clusters which are not considered as highly available.
This will mark those kind of tests with the ``[Skip:NonHighlyAvailable]`` tag, and make any run with a single-node cluster to skip tests with this tag.

### DIFF
--- a/test/extended/util/annotate/rules.go
+++ b/test/extended/util/annotate/rules.go
@@ -74,6 +74,17 @@ var (
 		"[Skipped:ovirt]": {},
 		"[Skipped:gce]":   {},
 
+		"[Skipped:SingleReplicaTopology]": {
+			`\[sig-apps\] Daemon set \[Serial\] should rollback without unnecessary restarts \[Conformance\]`,
+			`\[sig-node\] NoExecuteTaintManager Single Pod \[Serial\] doesn't evict pod with tolerations from tainted nodes`,
+			`\[sig-node\] NoExecuteTaintManager Single Pod \[Serial\] eventually evict pod with finite tolerations from tainted nodes`,
+			`\[sig-node\] NoExecuteTaintManager Single Pod \[Serial\] evicts pods from tainted nodes`,
+			`\[sig-node\] NoExecuteTaintManager Single Pod \[Serial\] removing taint cancels eviction \[Disruptive\] \[Conformance\]`,
+			`\[sig-node\] NoExecuteTaintManager Multiple Pods \[Serial\] evicts pods with minTolerationSeconds \[Disruptive\] \[Conformance\]`,
+			`\[sig-node\] NoExecuteTaintManager Multiple Pods \[Serial\] only evicts pods without tolerations from tainted nodes`,
+			`\[sig-cli\] Kubectl client Kubectl taint \[Serial\] should remove all the taints with the same key off a node`,
+		},
+
 		// Tests that don't pass on disconnected, either due to requiring
 		// internet access for GitHub (e.g. many of the s2i builds), or
 		// because of pullthrough not supporting ICSP (https://bugzilla.redhat.com/show_bug.cgi?id=1918376)
--- a/test/extended/util/cluster/cluster.go
+++ b/test/extended/util/cluster/cluster.go
@@ -38,6 +38,10 @@ type ClusterConfiguration struct {
 	// Disconnected is set for test jobs without external internet connectivity
 	Disconnected bool
 
+	// SingleReplicaTopology is set for disabling disruptive tests or tests
+	// that require high availability
+	SingleReplicaTopology bool
+
 	// NetworkPlugin is the "official" plugin name
 	NetworkPlugin string
 	// NetworkPluginMode is an optional sub-identifier for the NetworkPlugin.
@@ -64,10 +68,11 @@ func (c *ClusterConfiguration) ToJSONString() string {
 // ClusterState provides information about the cluster that is used to generate
 // ClusterConfiguration
 type ClusterState struct {
-	PlatformStatus *configv1.PlatformStatus
-	Masters        *corev1.NodeList
-	NonMasters     *corev1.NodeList
-	NetworkSpec    *operatorv1.NetworkSpec
+	PlatformStatus       *configv1.PlatformStatus
+	Masters              *corev1.NodeList
+	NonMasters           *corev1.NodeList
+	NetworkSpec          *operatorv1.NetworkSpec
+	ControlPlaneTopology *configv1.TopologyMode
 }
 
 // DiscoverClusterState creates a ClusterState based on a live cluster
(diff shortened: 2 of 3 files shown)


osherdp:
depends on openshift/release#20053

eranco74 (commented):
/lgtm

romfreiman:
@osherdp also this one: https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/test/e2e/node/taints.go#L209

romfreiman:
Jul 05 23:04:20.948 - 127s  I e2e-test/"[sig-node] NoExecuteTaintManager Single Pod [Serial] doesn't evict pod with tolerations from tainted nodes [Suite:openshift/conformance/serial] [Suite:k8s]" e2e test finished As "Passed"

osherdp:
@romfreiman done
also other tests from here https://github.com/openshift/origin/pull/26311 ?

omertuc:
Does this make https://github.com/openshift/origin/pull/26311 obsolete?

osherdp:
@omertuc yes, some people are against creating a new test suite for single-node, and prefer to skip tests dynamically by the cluster mode
I implemented it now using control-plane topology, can you take another look?

test/extended/util/cluster/cluster.go
```diff
@@ -38,6 +38,10 @@ type ClusterConfiguration struct {
 	// Disconnected is set for test jobs without external internet connectivity
 	Disconnected bool
 
+	// SingleReplicaTopology is set for disabling disruptive tests or tests
+	// that require high availability
+	SingleReplicaTopology bool
```
omertuc, line 43: Maybe reuse the MultiMaster field in this struct, but change it so it gets set as `*state.ControlPlaneTopology == configv1.SingleReplicaTopologyMode,` rather than `len(state.Masters.Items) > 1,`

osherdp: there is one usage for the multimaster field, and I do not intend to break it:
https://github.com/openshift/origin/blob/6adc3079a51f5ca677cdf445879a9feb407a22ac/vendor/k8s.io/kubernetes/test/e2e/framework/nodes_util.go#L119

omertuc: I didn't say remove the Multimaster field, I said reuse the MultiMaster field

omertuc: You can assert that indeed `!(len(state.Masters.Items) > 1)` whenever  `*state.ControlPlaneTopology == configv1.SingleReplicaTopologyMode` to verify the definitions have no diverged in any way

osherdp: also it's not conforming to the field name
you were right the first time, I need to use the topology not the master-counter field :)

omertuc: I didn't say you shouldn't use the topology, I said we don't need a new field for it, i.e.:

```go
	config := &ClusterConfiguration{
		MultiMaster:           *state.ControlPlaneTopology != configv1.SingleReplicaTopologyMode,
		MultiZone:             zones.Len() > 1,
		Zones:                 zones.List(),
	}
```

osherdp: it's not the same thing
it seems like they use multimaster to distinguish between multi-zone and multi-regional cases on cloud, and it's not the same as cluster topology 
I prefer creating a new field with the right meaning to its value, and see afterwards if it makes sense for the old field to be merged to the new one, because right now I doubt it

omertuc: Okay, not insisting, just making sure my comments are understood

test/extended/util/cluster/cluster.go
```diff
@@ -128,9 +137,10 @@ func LoadConfig(state *ClusterState) (*ClusterConfiguration, error) {
 	zones.Delete("")
 
 	config := &ClusterConfiguration{
-		MultiMaster: len(state.Masters.Items) > 1,
-		MultiZone:   zones.Len() > 1,
-		Zones:       zones.List(),
+		MultiMaster:           len(state.Masters.Items) > 1,
```
romfreiman, line 140: hm... can MultiMaster be True and SingleReplica topology can be 1?

omertuc: "We're not really sure let's just create a new field" is the conclusion from [this](https://github.com/openshift/origin/pull/26314#discussion_r667475569) resolved conversation

sttts:
Sgtm.

test/extended/util/cluster/cluster.go
osherdp: I think ``MultiMaster`` is not needed
Because the only usage in this repo refers to something which can be handled by ``MultiZone`` field imho

eranco74: ControlPlaneTopology == == configv1.SingleReplicaTopologyMode is true only for SNO.
So as long we don't add master nodes to SNO, when SingleReplicaTopology is true,  multiMaster is false and vice versa.

romfreiman, line 140: @smarterclayton @wking

test/extended/util/annotate/rules.go
```diff
@@ -74,6 +74,17 @@ var (
 		"[Skipped:ovirt]": {},
 		"[Skipped:gce]":   {},
 
+		"[Skipped:SingleReplicaTopology]": {
+			`\[sig-apps\] Daemon set \[Serial\] should rollback without unnecessary restarts`,
```
romfreiman, line 78: @rphillips @soltysh

test/extended/util/cluster/cluster.go
omertuc: > I think `MultiMaster` is not needed
> Because the only usage in this repo refers to something which can be handled by `MultiZone` field imho

This repo is also running vendored tests from the k8s repo, which originally defined the MultiMaster variable in the first place, so it makes sense to populate it just in case upstream makes use of it

test/extended/util/annotate/rules.go
rphillips: Looks like a good list.

test/extended/util/cluster/cluster.go
ravisantoshgudimetla: Usually the convention is to disable upstream tests in [o/k](https://github.com/openshift/kubernetes/blob/master/openshift-hack/test-kubernetes-e2e.sh) and then vendor o/o. This ensures we have single source of truth but in this case it could be ok considering `SingleReplicaTopology` var is defined here.

test/extended/util/annotate/rules.go
ravisantoshgudimetla: This a conformance test. Can you add the tag [Conformance] so that we can easily identify that we are skipping conformance tests as well, just like how you did for other conformance tests.

osherdp: you're right
done

soltysh (approved):
this looks reasonable, can you squash and ping me, I'll tag

osherdp:
@soltysh sure thing :)
done

PR #26319, opened 2021-07-12
Bug 1981246: [4.8]: test/e2e: allow workload incorrectly spread alert

dgrisonnet:
Backport of #26305 to release-4.8.

This is not a blocker so it can wait for 4.8.z before being merged.
Also, the best would be to merge this PR before https://github.com/openshift/cluster-monitoring-operator/pull/1276 since once added to 4.8, the alert will fire from time to time and thus make the e2e tests fail.


simonpasquier (commented):
/lgtm

aravindhp:
@dgrisonnet in the future, if there are dependencies between PRs please add a hold

PR #26323, opened 2021-07-13
Do not allow an excessive count of an individual event: they indicate a failure

dmage:
Based on @deads2k work #26173.

### DIFF
--- a/pkg/synthetictests/events.go
+++ b/pkg/synthetictests/events.go
@@ -1,0 +1,149 @@
+package synthetictests
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/test/ginkgo"
+)
+
+func combinedRegexp(arr ...*regexp.Regexp) *regexp.Regexp {
+	s := ""
+	for _, r := range arr {
+		if s != "" {
+			s += "|"
+		}
+		s += r.String()
+	}
+	return regexp.MustCompile(s)
+}
+
+var knownEvents = combinedRegexp(
+	// [sig-apps] StatefulSet Basic StatefulSet functionality [StatefulSetBasic] should not deadlock when a pod's predecessor fails [Suite:openshift/conformance/parallel] [Suite:k8s]
+	// PauseNewPods intentionally causes readiness probe to fail.
+	regexp.MustCompile(`ns/e2e-statefulset-[0-9]+ pod/ss-[0-9] node/[a-z0-9.-]+ - reason/Unhealthy Readiness probe failed: `),
+
+	// [sig-apps] StatefulSet Basic StatefulSet functionality [StatefulSetBasic] should perform rolling updates and roll backs of template modifications [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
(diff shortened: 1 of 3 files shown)


pkg/synthetictests/events.go
```diff
@@ -10,6 +10,42 @@ import (
...
+			s += "|"
+		}
+		s += r.String()
+	}
+	return regexp.MustCompile(s)
+}
+
+var knownEvents = combinedRegexp(
+	// [sig-apps] StatefulSet Basic StatefulSet functionality [StatefulSetBasic] should not deadlock when a pod's predecessor fails [Suite:openshift/conformance/parallel] [Suite:k8s]
+	// PauseNewPods causes readiness probe to fail.
```
deads2k, line 26: nit: PauseNewPods intentionally causes

pkg/synthetictests/events.go
```diff
@@ -10,6 +10,42 @@ import (
...
+	regexp.MustCompile(`ns/e2e-statefulset-[0-9]+ pod/ss-[0-9] node/[a-z0-9.-]+ - reason/Unhealthy Readiness probe failed: `),
+
+	// [sig-apps] StatefulSet Basic StatefulSet functionality [StatefulSetBasic] should perform rolling updates and roll backs of template modifications [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
+	regexp.MustCompile(`ns/e2e-statefulset-[0-9]+ pod/ss2-[0-9] node/[a-z0-9.-]+ - reason/Unhealthy Readiness probe failed: HTTP probe failed with statuscode: 404`),
```
deads2k, line 30: also intentional?

deads2k:
These look plausible.  if they work, we'll merge.  The approach is very good and should be continued until we're able to merge.  This will be a big help latching our progress.

pkg/synthetictests/events.go
dmage: yes, it calls `breakPodHTTPProbe`. Perhaps I need to add a commend about it

deads2k:
/approve

This is good in concept and the ratchet looks solid.  You may end up needing a set of excludes specific to upgrades.

dmage:
e2e-agnostic-cmd:

```
event happened 23 times, something is wrong: ns/cmd-basicresources deploymentconfig/test-deployment-config - reason/DeploymentAwaitingCancellation Deployment of version 2 awaiting cancellation of older running deployments
event happened 21 times, something is wrong: ns/openshift-machine-api machine/ci-op-r7y800m4-098b0-vgskh-worker-westus-sdqlb - reason/Updated Updated machine "ci-op-r7y800m4-098b0-vgskh-worker-westus-sdqlb"
```

Perhaps these can be ignored?

/retest

deads2k:
This looks like a good starting point to me.  If some platforms break, then we can punch out for them using what is developed here or fix them depending on whether the fix is fast.

/lgtm

dmage:
/hold

```
event happened 23 times, something is wrong: ns/openshift-network-diagnostics pod/network-check-target-qxkld node/ip-10-0-138-212.us-west-1.compute.internal - reason/NetworkNotReady network is not ready: container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: No CNI configuration file in /etc/kubernetes/cni/net.d/. Has your network provider started?
```

PR #26324, opened 2021-07-13
Bug 1942164: Fix time calc ordering for upgrades

jluhrsen:
The original PR https://github.com/openshift/origin/pull/26202 use
parens in the wrong place and the actual time calcs were happening
wrong and garbage values were being used, like this job [0] where
the test output looked like:
  : [sig-cluster-lifecycle] cluster upgrade should complete in -1386946h6m26.707003392s minutes

This should fix that.

[0] https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.9-upgrade-from-stable-4.8-e2e-gcp-ovn-upgrade/1414923427529101312


wking (approved):
/lgtm

PR #26325, opened 2021-07-14
Suppress scary looking panic message in disruption tests and that isn't actually a panic

deads2k:
The "Your test failed." message is a structured ginkgo panic used to
stop execution.  This commit moves the reporting to exceptional cases
only and provides a more unique strick to search the codebase for to
find the reporting line.

Suppresses
> Jul 14 12:05:36.523: INFO: recover: 
Your test failed.
Ginkgo panics to prevent subsequent assertions from running.
Normally Ginkgo rescues this panic so you shouldn't see it.
> 
> But, if you make an assertion in a goroutine, Ginkgo can't capture the panic.
To circumvent this, you should call
> 
> 	defer GinkgoRecover()
> 
> at the top of the goroutine that caused this panic.

from disruptive tests.  https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_kubernetes/856/pull-ci-openshift-kubernetes-release-4.8-e2e-aws-upgrade/1415239271618449408 as an example

### DIFF
--- a/test/extended/util/disruption/disruption.go
+++ b/test/extended/util/disruption/disruption.go
@@ -230,7 +230,6 @@ func finalizeTest(start time.Time, tc *junit.TestCase, ts *junit.TestSuite, f *f
 		}
 		return
 	}
-	framework.Logf("recover: %v", r)
 
 	switch r := r.(type) {
 	case ginkgowrapper.FailurePanic:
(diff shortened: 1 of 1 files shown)


deads2k:
/hold

wonderful.  Now hitting the next block

```go
	// if we have a panic but it hasn't been recorded by ginkgo, panic now
	if !g.CurrentGinkgoTestDescription().Failed {

```

deads2k:
just hit this again today.  awesome.

PR #26326, opened 2021-07-14
[release-4.6] Bug 1981582: Use cluster-logging for operator installation smoke test

kevinrizza:
A new catalog-operator alert is being introduced that fires when the
API server produces warnings during operator installation.

The latest amq-streams catalog entry hasn't completed its migration
away from v1beta1 CRDs, which results in deprecation warnings that in
turn cause the catalog-operator alert to fire.


kevinrizza:
Closing in favor of https://github.com/openshift/origin/pull/26355

PR #26329, opened 2021-07-15
WIP: Split [sig-api-machinery] and [sig-network] monitors

mfojtik:


pkg/monitor/api.go
```diff
@@ -182,6 +193,117 @@ func startServerMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.
...
+	if err != nil {
+		return err
+	}
+	var httpTransport *http.Transport
+
+	if disableConnectionReuse {
+		httpTransport = &http.Transport{
+			Dial: (&net.Dialer{
+				Timeout:   timeout,
+				KeepAlive: -1, // this looks unnecessary to me, but it was set in other code.
```
aojea, line 211: I will keep this, it will detect a connection hanging , it is a TCP option and vI think desirable
https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/net/dial.go;l=73-80

aojea:
I'm chasing these days a problem upstream due to the "abuse" of TCP
https://github.com/kubernetes/kubernetes/pull/103641#issuecomment-878384805

We should be careful with the tests, TCP ephemeral ports are finite, by default in linux there are:
```
$ cat /proc/sys/net/ipv4/ip_local_port_range
32768   60999
```

also, each socket in TCP is not inmediatly available, TCP connections in TIME_WAIT state are canceled after 60 seconds by defaut
```
$ cat /proc/sys/net/ipv4/tcp_fin_timeout
60
```

If the e2e binary runs "outside" of the cluster under test, it may happen that the connections are nated too, so you have to take into account the NAT range from the box that is natting, there was a bug in the CI related to this, don't know if this applies here too, but is important to consider it.

Creating new connections is expensive and Werner Heisenberg 'Uncertainty Principle' applies here, I would limit the test of new connections to something reasonable so we don't influence the measurement (1 connection every 10 seconds ?) , is not realistic a customer sends http connections every 1 second, is it?

pkg/monitor/api.go
```diff
@@ -182,6 +193,117 @@ func startServerMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.
...
+		}
+
+	}
+
+	httpClient := http.Client{
+		Transport: roundTripper,
+	}
+
+	go NewSampler(m, time.Second, func(previous bool) (condition *monitorapi.Condition, next bool) {
+		resp, err := httpClient.Get(clusterConfig.Host + url)
```
aojea, line 243: I think we should use a context to avoid the server keeps us hanging reading the body or some weird thing, just to be paranoid
```
ctx, cancel := context.WithTimeout(context.Background(), timeout)
defer cancel()
req, err := http.NewRequestWithContext(ctx, http.MethodGet, clusterConfig.Host + url, nil)
resp, err := httpClient.Do(req)
```

pkg/monitor/api.go
```diff
@@ -182,6 +193,117 @@ func startServerMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.
...
+		Level:   monitorapi.Error,
+		Locator: resourceLocator,
+		Message: fmt.Sprintf("%s is not responding to network requests", resourceLocator),
+	})
+
+	return nil
+}
+
+var apiMachineryErrors = []string{
+	"connection refused", // TODO: Likely a networking issue, but require more analysis to tell
```
aojea, line 274: we can document it, usually it happens something like this:
- we send a SYN
- we receive a RST

This means that there is actually some sort of communication because we receive a RST packet, who sends the RST is the key, it can be a firewall, kube-proxy does this for services without endpoints too, it can be the there is nothing listening on the host and the OS RSTs the connection to inform the client

pkg/monitor/api.go
```diff
@@ -182,6 +193,117 @@ func startServerMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.
...
+	"TLS handshake timeout",
+	"EOF", // could be ungraceful termination
+}
+
+var networkErrors = []string{
+	"i/o timeout",
+	"connection reset by peer",
```
aojea, line 281: meanwhile `connection refused` happens when we are starting a connection, this means that the connection was already established and it was working, however, something resets it.
I saw this a lot on loadbalancers and proxies, i.e, the upstream server hangs or ir not available.
It also can be sent by the server if it is overloaded or not able to process a request ... 
This is tricky to categorize ... but the fact that a connection has already been established makes me think that is more an apiMachinery error than a network error

pkg/monitor/api.go
```diff
@@ -182,6 +193,117 @@ func startServerMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.
...
+func IsAPIMachineryError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if IsNetworkError(err) {
+		return false
+	}
+	// If it is not networking error, it is always API server problem
+	return true
+}
```
aojea, line 293: ni
```suggestion
	return !IsNetworkError(err) 
}
```

pkg/monitor/api.go
```diff
@@ -182,6 +193,117 @@ func startServerMonitoring(ctx context.Context, m *Monitor, clusterConfig *rest.
...
+			return err
+		}
+
+	}
+
+	httpClient := http.Client{
+		Transport: roundTripper,
+	}
+
+	go NewSampler(m, time.Second, func(previous bool) (condition *monitorapi.Condition, next bool) {
```
aojea, line 242: ref https://github.com/openshift/origin/pull/26329#issuecomment-884786869,

PR #26336, opened 2021-07-16
Bug 1925524: tweak jenkins imagestreamtag podtemplate test to label imagestream; jenkins multi namespace test has been moved to client plugin

gabemontero:
So the commented out test case is now handled by https://github.com/openshift/jenkins-client-plugin/pull/364

And the add of the label to the imagestream in the imagestreamtag pod template is happening as part of the scalability / existing behavior trade off we will be making once we release whatever version of the sync plugin after v1.0.46 we go with per the saga recorded in https://github.com/openshift/jenkins/pull/1297

I'll then be submitting a PR to sync plugin with my fixes from today that should pass tests once this merges.

/assign @adambkaplan 
/assign @coreydaley 

@openshift/openshift-team-build-api @openshift/team-devtools-jenkins FYI

### DIFF
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -43319,6 +43319,8 @@ func testExtendedTestdataImagestreamJenkinsSlavePodsYaml() (*asset, error) {
 var _testExtendedTestdataImagestreamtagJenkinsSlavePodsYaml = []byte(`apiVersion: v1
 kind: ImageStream
 metadata:
+  labels:
+    role: jenkins-slave
   name: slave-jenkins
 spec:
   tags:
--- a/test/extended/testdata/imagestreamtag-jenkins-slave-pods.yaml
+++ b/test/extended/testdata/imagestreamtag-jenkins-slave-pods.yaml
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: ImageStream
 metadata:
+  labels:
+    role: jenkins-slave
   name: slave-jenkins
 spec:
   tags:
--- a/test/extended/builds/pipeline_origin_bld.go
+++ b/test/extended/builds/pipeline_origin_bld.go
@@ -265,66 +265,6 @@ var _ = g.Describe("[sig-builds][Feature:Builds][sig-devex][Feature:Jenkins][Slo
 
 				})
 
-				/* to be moved to jenkins client plugin e2e
-				g.By("should handle multi-namespace templates", func() {
-					g.By("create additional projects")
-					namespace := oc.Namespace()
-					namespace2 := oc.Namespace() + "-2"
-					namespace3 := oc.Namespace() + "-3"
-
-					err := oc.Run("new-project").Args(namespace2).Execute()
-					o.Expect(err).NotTo(o.HaveOccurred())
-					err = oc.Run("new-project").Args(namespace3).Execute()
-					o.Expect(err).NotTo(o.HaveOccurred())
-					// no calls to delete these two projects here; leads to timing
-					// issues with the framework deleting all namespaces
-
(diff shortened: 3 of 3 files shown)


test/extended/testdata/bindata.go
```diff
@@ -43319,6 +43319,8 @@ func testExtendedTestdataImagestreamJenkinsSlavePodsYaml() (*asset, error) {
 var _testExtendedTestdataImagestreamtagJenkinsSlavePodsYaml = []byte(`apiVersion: v1
 kind: ImageStream
 metadata:
+  labels:
+    role: jenkins-slave
```
adambkaplan, line 43323: Does the role have to be `jenkins-slave` - can it be `jenkins-agent`?

We should discourage the use of "slave" wherever possible, as it does not conform to our inclusive language standards (or those of the Jenkins community).

test/extended/testdata/imagestreamtag-jenkins-slave-pods.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: ImageStream
 metadata:
+  labels:
+    role: jenkins-slave
```
adambkaplan, line 5: ditto

adambkaplan:
/approve

If language concerns are addressed, another team member can lgtm

test/extended/testdata/bindata.go
gabemontero: yeah thses tests came into existence, are legacy conventions, before those standards were put into place.

a little more wide spread and non-trivial of a change than you think, but I'll start on it right now

gabemontero: actually we are going to have to stage this @adambkaplan 

1) the sync plugin actually looks for both the label key `role` and the specific value `jenkins-slave`, and only that right now
2) so this change without a sync plugin change fails the test
3) and moving forward, while we can start looking for `role: jenkins-agent` we cannot break existing users still looking at the old value

So the path forward will be 
1) merge this as is
2) I'll update my sync plugin PR to look for either value
3) when that is in the image (i.e. 1.0.48 is what we pull in), we can go back and changes these tests

gabemontero:
> /approve
> 
> If language concerns are addressed, another team member can lgtm

see https://github.com/openshift/origin/pull/26336#discussion_r671531586 for details on the multi PR roadmap required to address things

gabemontero:
e2e-aws-jenkins is green ... that is the verification we need

rest is just non-related flakes getting in the way

/retest

gabemontero:
still with the various ipi-ovn-ipv6 can't get off the ground issues ... totally unrelated to this change

seeing if things got addressed over the weekend

/retest

gabemontero:
talked with @adambkaplan during scrum and he is OK with the multi PR plan I noted above

test/extended/testdata/bindata.go
adambkaplan: > yeah thses tests came into existence, are legacy conventions, before those standards were put into place.

My main concern was that we were introducing a new labeling requirement that was not inclusive. If this was already in place, then we will need to watch for the old value and the new one.

gabemontero:
OK the latest ipi-ovn-ipv6 failure was https://bugzilla.redhat.com/show_bug.cgi?id=1975476

I've linked the failed run here in that bz

@bparees can we get an override for e2e-metal-ipi-ovn-ipv6 so we can make progress with this escalated customer BZ?  the e2e-aws-jenkins run passed, which is what we tweaked here.

gabemontero:
as sig-storage flake in e2e-gcp 

/test e2e-gcp

akram:
All tests passed successfully, but tide seems to re-run them. Probably because of rebase requirement.

PR #26337, opened 2021-07-19
Bug 1999561: Add API rollout resiliency test

osherdp:
Adding test that validates cluster is resilient to API restarts.
Ultimately, it will make validations on other cluster components, but for now we'll validate API is getting up fast enough (40 seconds of outage).
This test will only run on single-node clusters, and should be part of suite of disruptive, serial tests.

### DIFF
--- a/test/extended/apiserver/resiliency.go
+++ b/test/extended/apiserver/resiliency.go
@@ -1,0 +1,117 @@
+package apiserver
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	types "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+
+	"github.com/openshift/origin/test/extended/scheme"
+	"github.com/openshift/origin/test/extended/single_node"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
+	f := framework.NewDefaultFramework("cluster-resiliency")
+	f.SkipNamespaceCreation = true
+	f.SkipPrivilegedPSPBinding = true
+
+	oc := exutil.NewCLIWithoutNamespace("cluster-resiliency")
+
+	ginkgo.It("should allow a fast rollout of kube-apiserver", func() {
+		controlPlaneTopology, _ := single_node.GetTopologies(f)
+
+		if controlPlaneTopology != configv1.SingleReplicaTopologyMode {
+			e2eskipper.Skipf("Test is only relevant for single replica topologies")
+		}
+
+		config, err := framework.LoadConfig()
+		framework.ExpectNoError(err)
+
+		setRESTConfigDefaults(config)
+		restClient, err := rest.RESTClientFor(config)
+		framework.ExpectNoError(err)
+
+		httpClient := restClient.Client
+
+		ginkgo.By("Making sure no previous rollout is in progress")
+		clusterApiServer, err := oc.AdminOperatorClient().OperatorV1().KubeAPIServers().Get(context.Background(), "cluster", metav1.GetOptions{})
+		framework.ExpectNoError(err)
+		gomega.Expect(clusterApiServer.Status.NodeStatuses[0].TargetRevision).To(gomega.Equal(int32(0)))
+
+		ginkgo.By("Forcing API rollout")
+		forceApiRollout(oc)
+
+		ginkgo.By("Expecting API to become unavailable")
+		wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+			ready, _, err := isApiReady(config, httpClient)
+			return !ready, err
+		})
+
+		start := time.Now()
+
+		ginkgo.By("Expecting API to become ready")
+		wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+			ready, _, _ := isApiReady(config, httpClient)
+			return ready, nil
+		})
+
+		end := time.Now()
+
+		ginkgo.By("Measuring disruption duration time")
+		gomega.Expect(end.Sub(start)).To(gomega.BeNumerically("<", 40*time.Second))
+	})
+})
+
+func setRESTConfigDefaults(config *rest.Config) {
+	if config.GroupVersion == nil {
+		config.GroupVersion = &schema.GroupVersion{Group: "", Version: "v1"}
+	}
+
+	if config.NegotiatedSerializer == nil {
+		config.NegotiatedSerializer = scheme.Codecs
+	}
+}
+
+func forceApiRollout(oc *exutil.CLI) {
+	redeploymentReason := fmt.Sprintf(`{"spec":{"forceRedeploymentReason":"resiliency-test-%v"}}`, uuid.NewUUID())
+
+	_, err := oc.AdminOperatorClient().OperatorV1().KubeAPIServers().Patch(context.Background(), "cluster", types.MergePatchType,
+		[]byte(redeploymentReason), metav1.PatchOptions{})
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+}
+
+func isApiReady(clusterConfig *rest.Config, httpClient *http.Client) (ready bool, reason string, err error) {
+	resp, err := httpClient.Get(clusterConfig.Host + "/readyz")
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
+	if err != nil {
+		return false, "client failed to make the request", err
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err == nil {
+			return false, fmt.Sprintf("got status code %v from the server: %v", resp.Status, body), nil
+		}
+
+		return false, fmt.Sprintf("got status code %v from the server", resp.Status), err
+	}
+
+	return true, "kube-apiserver is ready", nil
+}
--- a/test/extended/apiserver/resiliency.go
+++ b/test/extended/apiserver/resiliency.go
@@ -72,7 +72,11 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
 		end := time.Now()
 
 		ginkgo.By("Measuring disruption duration time")
-		gomega.Expect(end.Sub(start)).To(gomega.BeNumerically("<", 40*time.Second))
+		disruptionDuration := end.Sub(start)
+		// For more information: https://github.com/openshift/origin/pull/26337/files#r698435488
+		gomega.Expect(disruptionDuration).To(gomega.BeNumerically("<", 40*time.Second),
+			fmt.Sprintf("Total time of disruption is %v which is more than 40 seconds. ", disruptionDuration)+
+				"Actual SLO for this is 60 seconds, yet we want to be notified about major regressions")
 	})
 })
 
--- a/test/extended/single_node/topology.go
+++ b/test/extended/single_node/topology.go
@@ -43,7 +43,7 @@ func getNamespaceStatefulSets(f *framework.Framework, namespace corev1.Namespace
 	return list.Items
 }
 
-func getTopologies(f *framework.Framework) (controlPlaneTopology, infraTopology v1.TopologyMode) {
+func GetTopologies(f *framework.Framework) (controlPlaneTopology, infraTopology v1.TopologyMode) {
 	oc := exutil.NewCLIWithFramework(f)
 	infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(),
 		"cluster", metav1.GetOptions{})
@@ -132,7 +132,7 @@ var _ = ginkgo.Describe("[sig-arch] Cluster topology single node tests", func() 
 	f := framework.NewDefaultFramework("single-node")
 
 	ginkgo.It("Verify that OpenShift components deploy one replica in SingleReplica topology mode", func() {
-		controlPlaneTopology, infraTopology := getTopologies(f)
+		controlPlaneTopology, infraTopology := GetTopologies(f)
 
 		if controlPlaneTopology != v1.SingleReplicaTopologyMode && infraTopology != v1.SingleReplicaTopologyMode {
 			e2eskipper.Skipf("Test is only relevant for single replica topologies")
--- a/test/extended/util/annotate/generated/zz_generated.annotations.go
+++ b/test/extended/util/annotate/generated/zz_generated.annotations.go
@@ -15,6 +15,8 @@ var annotations = map[string]string{
 
 	"[Top Level] [Conformance][sig-api-machinery][Feature:APIServer] local kubeconfig \"localhost.kubeconfig\" should be present on all masters and work": "\"localhost.kubeconfig\" should be present on all masters and work [Suite:openshift/conformance/parallel/minimal]",
 
+	"[Top Level] [Conformance][sig-sno][Serial] Cluster should allow a fast rollout of kube-apiserver": "should allow a fast rollout of kube-apiserver [Suite:openshift/conformance/serial/minimal]",
+
 	"[Top Level] [Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [IdP] test RequestHeaders IdP": "test RequestHeaders IdP [Suite:openshift/conformance/serial]",
 
 	"[Top Level] [Serial][sig-node][Feature:TopologyManager] Configured cluster [Disabled:Broken] with non-gu workload should run with no regressions with single pod, single container requesting 1 core": "with single pod, single container requesting 1 core",


test/extended/single_node/resiliency.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+var _ = ginkgo.Describe("[sig-api-machinery] Cluster resiliency", func() {
+	f := framework.NewDefaultFramework("single-node")
+	f.SkipNamespaceCreation = true
+	f.SkipPrivilegedPSPBinding = true
+
+	oc := exutil.NewCLIWithoutNamespace("cluster-resiliency")
+
+	ginkgo.It("Verify that OpenShift components are resilient to kubeapi server rollback", func() {
```
omertuc, line 74: s/rollback/rollout

test/extended/single_node/resiliency.go
```diff
@@ -0,0 +1,91 @@
...
+		controlPlaneTopology, infraTopology := getTopologies(f)
+
+		if controlPlaneTopology != v1.SingleReplicaTopologyMode && infraTopology != v1.SingleReplicaTopologyMode {
+			e2eskipper.Skipf("Test is only relevant for single replica topologies")
+		}
+
+		disruption.Run(f, "Cluster Resiliency to API shutdown", "cluster-resiliency", disruption.TestData{},
+			[]upgrades.Test{
+				&controlplane.ResiliencyTest{},
```
omertuc, line 83: I guess you'll want to have one test for checking that API was down for less than 60 seconds, and a separate test to make sure components didn't restart during that time

test/extended/single_node/resiliency.go
```diff
@@ -0,0 +1,91 @@
...
+
+func forceApiRollback(oc *exutil.CLI) {
+	redeploymentReason := fmt.Sprintf(`{"spec":{"forceRedeploymentReason":"resiliency-test-%v"}}`, time.Now().Format(time.RFC3339))
+
+	_, err := oc.AdminOperatorClient().OperatorV1().KubeAPIServers().Patch(context.Background(), "cluster", types.MergePatchType,
+		[]byte(redeploymentReason), metav1.PatchOptions{})
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+}
+
+func waitForApiToShutdown(f *framework.Framework) {
```
omertuc, line 31: I'm guessing you're not using this anymore?

test/extended/util/disruption/controlplane/controlplane.go
```diff
@@ -162,3 +163,43 @@ func allClusterVersionsAreGTE(version semver.Version, config *rest.Config) (bool
...
+	// wait to ensure API is still up after the test ends
+	<-done
+	time.Sleep(5 * 60 * time.Second)
+	cancel()
+	end := time.Now()
+
+	// TODO: validate API is down for a gracefull amount of time, and validate other operators are not affected
+	toleratedDisruption := 0.08
+	events := m.Intervals(time.Time{}, time.Time{})
+	disruption.ExpectNoDisruption(f, toleratedDisruption, end.Sub(start), events, "API was unreachable during disruption (AWS has a known issue: https://bugzilla.redhat.com/show_bug.cgi?id=1943804)")
```
omertuc, line 200: You planned to call `ExpectDisruption` instead, right?

test/extended/single_node/resiliency.go
osherdp: ignore

test/extended/single_node/resiliency.go
osherdp: ack

test/extended/single_node/resiliency.go
osherdp: yes, we'll have several tests

test/extended/util/disruption/controlplane/controlplane.go
osherdp: ignore for now

osherdp: I'll have to use other validation from those two options I think

test/extended/single_node/resiliency.go
```diff
@@ -0,0 +1,73 @@
...
+	f.SkipNamespaceCreation = true
+	f.SkipPrivilegedPSPBinding = true
+
+	oc := exutil.NewCLIWithoutNamespace("cluster-resiliency")
+
+	ginkgo.It("[Conformance] Cluster should recover fast on kube-apiserver rollout [Disruptive]", func() {
+		controlPlaneTopology, infraTopology := getTopologies(f)
+
+		if controlPlaneTopology != v1.SingleReplicaTopologyMode && infraTopology != v1.SingleReplicaTopologyMode {
+			e2eskipper.Skipf("Test is only relevant for single replica topologies")
```
omertuc, line 48: Is it? Why not run it on multi-node as well? It's interesting there as well

osherdp: ack, fixing it

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,65 @@
...
+		disruption.Run(f, "Cluster Resiliency to API shutdown", "cluster-resiliency", disruption.TestData{},
+			[]upgrades.Test{
+				&controlplane.ApiFastRestorationTest{},
+			},
+			func() {
+				ginkgo.By("Forcing API rollout")
+				forceApiRollout(oc)
+
+				ginkgo.By("Expecting API downtime")
+				wait.PollImmediate(3*time.Second, time.Minute, func() (bool, error) {
```
omertuc, line 52: What's the difference between this polling and the polling in ApiFastRestorationTest?

osherdp: so this part should just do he rollout and verify rollout is occurring
the ``ApiFastRestorationTest`` (and pod-watches which I'll add in the future) should include everything we want to actually measure

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,65 @@
...
+func forceApiRollout(oc *exutil.CLI) {
+	redeploymentReason := fmt.Sprintf(`{"spec":{"forceRedeploymentReason":"resiliency-test-%v"}}`, uuid.NewUUID())
+
+	_, err := oc.AdminOperatorClient().OperatorV1().KubeAPIServers().Patch(context.Background(), "cluster", types.MergePatchType,
+		[]byte(redeploymentReason), metav1.PatchOptions{})
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+}
+
+func isApiUp(oc *exutil.CLI) (bool, error) {
+	_, err := oc.AdminKubeClient().CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
```
sttts, line 31: this is not a good check. Check readiness. This answers success far too early.

osherdp: fixed

osherdp:
(build was broken because of upstream change)

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/disruption"
+	"github.com/openshift/origin/test/extended/util/disruption/controlplane"
+	types "k8s.io/apimachinery/pkg/types"
+	"k8s.io/kubernetes/test/e2e/upgrades"
```
sttts, line 20: order

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,66 @@
...
+		[]byte(redeploymentReason), metav1.PatchOptions{})
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+}
+
+func isApiUp(client clientset.Interface) (bool, error) {
+	_, err := client.CoreV1().RESTClient().Get().RequestURI("/readyz").DoRaw(context.TODO())
+	return err == nil, err
+}
+
+var _ = ginkgo.Describe("[Conformance][sig-api-machinery][Serial]", func() {
```
sttts, line 36: This is wrong. `[sig-sno]`. We don't own this.

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,66 @@
...
+	f := framework.NewDefaultFramework("cluster-resiliency")
+	f.SkipNamespaceCreation = true
+	f.SkipPrivilegedPSPBinding = true
+
+	oc := exutil.NewCLIWithoutNamespace("cluster-resiliency")
+
+	ginkgo.It("should allow a fast kube-apiserver rollout", func() {
```
sttts, line 43: if you want to add operators here as well, rename it (= remove mentioning kube-apiserver and mention the tested operators). In the way it is named now, API team will get bugs due to operators. That's wrong.

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,66 @@
...
+		disruption.Run(f, "Cluster Resiliency", "cluster-resiliency", disruption.TestData{},
+			[]upgrades.Test{
+				&controlplane.ApiFastRestorationTest{},
```
sttts, line 46: if this is a subtest and more will come, separate them as ginkgo subtest.

test/extended/util/disruption/controlplane/controlplane.go
```diff
@@ -124,7 +124,8 @@ func (t *availableTest) Test(f *framework.Framework, done <-chan struct{}, upgra
 			toleratedDisruption = 0
 		}
 	}
-	disruption.ExpectNoDisruption(f, toleratedDisruption, end.Sub(start), m.Intervals(time.Time{}, time.Time{}), fmt.Sprintf("API %q was unreachable during disruption (AWS has a known issue: https://bugzilla.redhat.com/show_bug.cgi?id=1943804)", t.name))
+	events := m.Intervals(time.Time{}, time.Time{})
```
sttts, line 127: what is this? Add a comment.

test/extended/util/disruption/controlplane/controlplane.go
```diff
@@ -162,3 +163,38 @@ func allClusterVersionsAreGTE(version semver.Version, config *rest.Config) (bool
 	}
 	return true, nil
 }
+
+type ApiFastRestorationTest struct{}
+
+func (t ApiFastRestorationTest) Name() string { return "api-fast-restoration-test" }
+func (t ApiFastRestorationTest) DisplayName() string {
+	return "API is being restored within restricted time"
```
sttts, line 171: which time? be concrete.

test/extended/util/disruption/controlplane/controlplane.go
```diff
@@ -162,3 +163,38 @@ func allClusterVersionsAreGTE(version semver.Version, config *rest.Config) (bool
 	}
 	return true, nil
 }
+
+type ApiFastRestorationTest struct{}
```
sttts, line 167: what is this supposed to do in addition to a `/readyz` check

test/extended/util/disruption/disruption.go
```diff
@@ -322,6 +322,15 @@ func ExpectNoDisruption(f *framework.Framework, tolerate float64, total time.Dur
 	}
 }
 
+func ExpectScopedDisruption(f *framework.Framework, maxDuration time.Duration, events monitorapi.Intervals, reason string) {
```
sttts, line 325: document

sttts:
/hold

This is not owned by sig-api-machinery, but by SNO team.

test/extended/apiserver/resiliency.go
romfreiman: hm... isnt it api behavior?

sttts: yet, who is getting BZs about SNO problem? My last information was that this is the telco team.

sttts: Note that I read "Ultimately, it will make validations on other cluster components, but for now we'll validate API is getting up fast enough (40 seconds of outage)." as it will get operator recovery tests as well here.

sttts: see comment below https://github.com/openshift/origin/pull/26337/files#r686730057

romfreiman: sure, but how test coverage related to customer BZ? I mean failure in this test will probably cause API bz, no? At least untill the operators part is introduced

sttts:
General question: what is the target platform for this? SNO? Why doesn't it say anything about the topology? @romfreiman

test/extended/util/disruption/controlplane/controlplane.go
```diff
@@ -162,3 +163,38 @@ func allClusterVersionsAreGTE(version semver.Version, config *rest.Config) (bool
...
+func (t ApiFastRestorationTest) Name() string { return "api-fast-restoration-test" }
+func (t ApiFastRestorationTest) DisplayName() string {
+	return "API is being restored within restricted time"
+}
+
+// Setup does nothing
+func (t *ApiFastRestorationTest) Setup(f *framework.Framework) {
+}
+
+// Test runs a connectivity check to the core APIs.
```
sttts, line 178: why should we check a connectivity invariant while rolling an kube-apiserver revision out that causes 60s of outage?

test/extended/apiserver/resiliency.go
romfreiman: @osherdp it can be sig-sno, I dont really mind.

romfreiman:
> General question: what is the target platform for this? SNO? Why doesn't it say anything about the topology? @romfreiman

Because @omertuc and I suggested that it will run on every variant. But thinking about it again, you implemented the fast rollout only in the sno context, right? so the api will behave differently in multi node scenario - then, I assume, we should take topology into the account. Although, maybe we should do it only for the api timing - the operators part should behave the same in multi and sno, right?

test/extended/util/disruption/disruption.go
osherdp: done

test/extended/util/disruption/controlplane/controlplane.go
osherdp: refactored the code to make it a lot simpler, please have another look

test/extended/util/disruption/controlplane/controlplane.go
osherdp: no, it does not
refactored the code to make it a lot simpler (and only use ``/readyz``), please have another look

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,66 @@
...
+				wait.PollImmediate(3*time.Second, time.Minute, func() (bool, error) {
+					up, _ := isApiUp(f.ClientSet)
+					return !up, nil
+				})
+
+				ginkgo.By("Expecting API to become reachable")
+				wait.PollImmediate(3*time.Second, 2*time.Minute, func() (bool, error) {
+					up, _ := isApiUp(f.ClientSet)
+					return up, nil
+				})
```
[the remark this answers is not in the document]
osherdp: we first make sure we see api going down and then up before making the actual validation on rollout time
anyway have a look after the refactor, it should be more understandable

test/extended/apiserver/resiliency.go
osherdp: ok, I removed it as an invariant
we'll first have the basic test and the pods-watching logic will join afterwards as an invariant

test/extended/apiserver/resiliency.go
osherdp: more invariants will come in the future, for now I kept it simple

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, event := range events {
+		if strings.Contains(event.Message, message) {
+			return true
+		}
+	}
+
+	return false
+}
+
+var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
```
osherdp, line 44: @sttts any idea how should I disable the cluster monitor synthetic events test?
it seems like the infra runs it for every test

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,132 @@
...
+			return nil, err
+		}
+	}
+
+	return &http.Client{
+		Transport: roundTripper,
+	}, nil
+}
+
+func isApiReady(clusterConfig *rest.Config, httpClient *http.Client) (bool, error) {
```
sttts, line 66: nit: helpers are usually at the bottom in go code

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+	httpTransport := &http.Transport{
+		Dial: (&net.Dialer{
+			Timeout: timeout,
+		}).Dial,
+		TLSClientConfig:     tlsConfig,
+		TLSHandshakeTimeout: timeout,
+		IdleConnTimeout:     timeout,
+	}
```
sttts, line 51: why not use the existing helpers? `rest.RESTClientFor(config).Client`

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err == nil {
+			err = fmt.Errorf("error running request: %v: %v", resp.Status, string(body))
+		} else {
+			err = fmt.Errorf("error running request: %v", resp.Status)
+		}
+
+		return false, err
```
sttts, line 84: this is an odd signature for a func, returning false and err. The case here is no error case. The typical signature in go would be `ready bool, reason string, err error`.

test/extended/apiserver/resiliency.go
sttts: @mfojtik

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,132 @@
...
+	f.SkipNamespaceCreation = true
+	f.SkipPrivilegedPSPBinding = true
+
+	oc := exutil.NewCLIWithoutNamespace("cluster-resiliency")
+
+	ginkgo.It("should allow a fast rollout", func() {
+		controlPlaneTopology, infraTopology := single_node.GetTopologies(f)
+
+		if controlPlaneTopology != configv1.SingleReplicaTopologyMode && infraTopology != configv1.SingleReplicaTopologyMode {
+			e2eskipper.Skipf("Test is only relevant for single replica topologies")
```
sttts, line 101: isn't a check for controlPlaneTopology enough? We are verifying the control plane here, don't we?

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
+	f := framework.NewDefaultFramework("cluster-resiliency")
+	f.SkipNamespaceCreation = true
+	f.SkipPrivilegedPSPBinding = true
+
+	oc := exutil.NewCLIWithoutNamespace("cluster-resiliency")
+
+	ginkgo.It("should allow a fast rollout", func() {
```
sttts, line 97: fast rollout of kube-apiserver

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		config, err := framework.LoadConfig()
+		framework.ExpectNoError(err)
+
+		httpClient, err := createHttpClient(config, time.Second)
+		framework.ExpectNoError(err)
+
+		ginkgo.By("Forcing API rollout")
+		forceApiRollout(oc)
```
sttts, line 111: I would add a step before checking that the kube-apiserver is not progressing, i.e. nodeStatus.targetRevision is zero.

osherdp: done

test/extended/apiserver/resiliency.go
osherdp: done

test/extended/apiserver/resiliency.go
osherdp: I understand it to not be critical, as it will just mark flakiness and not fail the build

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,117 @@
...
+		ginkgo.By("Expecting API to become ready")
+		wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+			ready, _, _ := isApiReady(config, httpClient)
+			return ready, nil
+		})
+
+		end := time.Now()
+
+		ginkgo.By("Measuring disruption duration time")
+		gomega.Expect(end.Sub(start)).To(gomega.BeNumerically("<", 40*time.Second))
```
sttts, line 75: why 40? We said 60, didn't we?

osherdp: @romfreiman any preference on that one?
We can choose 60s as the original SLA or 40s to track regressions

sttts: where are we today on average and on 99% quantile? I can see to put the 99% value here if it is very clear that the SLO is 60s, but we put a lower number to not regress, both in go doc and in test output.

osherdp: We're way better from 40 seconds. Most of the time it's between 7-10 seconds, the maximal time I've got is 11 seconds in a couple of runs that I've tried

osherdp: Let me update the message real quick

test/extended/apiserver/resiliency.go
```diff
@@ -0,0 +1,121 @@
...
+		ginkgo.By("Making sure no previous rollout is in progress")
+		clusterApiServer, err := oc.AdminOperatorClient().OperatorV1().KubeAPIServers().Get(context.Background(), "cluster", metav1.GetOptions{})
+		framework.ExpectNoError(err)
+		gomega.Expect(clusterApiServer.Status.NodeStatuses[0].TargetRevision).To(gomega.Equal(int32(0)))
+
+		ginkgo.By("Forcing API rollout")
+		forceApiRollout(oc)
+
+		ginkgo.By("Expecting API to become unavailable")
+		wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
```
mfojtik, line 59: what if it is not unavaiable within 1 minute? woudln't that skew the next test?

osherdp: That's possible
Do we have some way to stop the whole suite from running?

osherdp:
/bugzilla refresh
(outage has been resolved probably)

PR #26340, opened 2021-07-20
[sig-arch] platform pods should not fail to start - add QPS BZ flake

kikisdeliveryservice:
We are seeing  error "pull QPS exceeded" in some runs from:
https://bugzilla.redhat.com/show_bug.cgi?id=1983833

### DIFF
--- a/pkg/synthetictests/kubelet.go
+++ b/pkg/synthetictests/kubelet.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
-
 	"github.com/openshift/origin/pkg/test/ginkgo"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
cc: @wking

pkg/synthetictests/kubelet.go
```diff
@@ -108,11 +107,17 @@ func testContainerFailures(events monitorapi.Intervals) []*ginkgo.JUnitTestCase
 
 	const failToStartTestName = "[sig-architecture] platform pods should not fail to start"
 	if len(failures) > 0 {
+		qpsMsg := ""
+		for  _, failure := range failures:
```
wking, line 111: I think you were in Python-mode or something.  [Go needs braces][1]:

```go
for _, failure := range failures {
  if strings.Contains(failure, "pull QPS exceeded") {
    qpsMsg = "..."
    break
  }
}
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26340/pull-ci-openshift-origin-master-images/1417295007534026752#1:build-log.txt%3A69

pkg/synthetictests/kubelet.go
```diff
@@ -108,11 +107,17 @@ func testContainerFailures(events monitorapi.Intervals) []*ginkgo.JUnitTestCase
...
+			if strings.Contains(failure, "pull QPS exceeded"):
```
smarterclayton, line 112: That's just one type of failure.  We don't want to pass if other types are added.

Roughly, up above when see something htat could be a failure, if it has this string it should go into a new `flakes` slice instead.  Then if failures greater than zero, you should also print flakes in this block, but if failures are zero and flakes are present, you should print a specific junit test case for fail to start in a net new block below.

pkg/synthetictests/kubelet.go
kikisdeliveryservice: im gonna pretend this didnt happen

PR #26346, opened 2021-07-23
WIP: [3.11] Bug 1970977: UPSTREAM: multiple: Fix corruption of FibreChannel volumes

jsafrane:
This is the same PR as #26222, re-created to run CI jobs for jobs that were overriden in the previous PR.

---

There are two upstream PRs, a bot will link them. The code is in different files in 3.11 than in 4.x, however, the backport went quite smooth.

Fix corruption of FS devices when a pod is deleted when kubelet is not running. The volume reconstruction in this case must not fail, otherwise the volume is marked as unused and can be mounted on another node, corrupting the filesystem on it.

Therefore try hard not to fail 1) volume reconstruction and 2) unmounting the volume on error.

Upstream cherry pick to 1.21: kubernetes/kubernetes#102656


jsafrane:
/test extended_clusterup
/test cmd
/test end_to_end
/test extended_conformance_install
These were not started automatically

jsafrane:
/close
CI is now working in the other PR

PR #26347, opened 2021-07-23
Added E2E test for multus-cni

nicklesimba:
This test performs a similar test to the upstream multus-cni E2E test...
- Creates a net-attach-def
- Creates a pod with an annotation that refers to the net-attach-def
- Checks for newly created net1 interface and passes test if found

Feedback on the test contents/structure is welcome =)

### DIFF
--- a/test/extended/networking/multus.go
+++ b/test/extended/networking/multus.go
@@ -1,0 +1,51 @@
+package networking
+
+import (
+	"fmt"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/api/core/v1"
+	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
+)
+
+var _ = g.Describe("[sig-network][Feature:Multus]", func() {
+	var oc *exutil.CLI
+	var ns string // namespace
+
+	oc = exutil.NewCLI("multus-e2e")
+
+	f := oc.KubeFramework()
+	podName := "multus-test-pod-"
+
+	// Multus is already installed on origin. These tests aims to verify the integrity of the installation.
+
+	g.It("should use multus to create net1 device from network-attachment-definition", func() {
+		var err error
+		ns = f.Namespace.Name
+
+		g.By("creating a net-attach-def using bridgeCNI")
+		nad_yaml := exutil.FixturePath("testdata", "net-attach-defs", "bridge-nad.yml")
+		g.By(fmt.Sprintf("calling oc create -f %s", nad_yaml))
+		err = oc.AsAdmin().Run("create").Args("-f", nad_yaml).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred(), "created net-attach-def")
+
+		g.By("launching pod with an annotation to use the net-attach-def")
+		annotation := map[string]string{
+			"k8s.v1.cni.cncf.io/networks": "bridge-nad",
+		}
+		testPod := frameworkpod.CreateExecPodOrFail(f.ClientSet, ns, podName, func(pod *v1.Pod) {
+			pod.ObjectMeta.Annotations = annotation
+		})
+
+		g.By("checking for net1 interface on pod")
+		output, err := oc.AsAdmin().Run("exec").Args(testPod.Name, "--", "ip", "a", "show", "dev", "net1").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		fmt.Println(output)
+		o.Expect(output).Should(o.ContainSubstring("net1"))
+		o.Expect(output).Should(o.ContainSubstring("inet 10.10.0.1/24"))
+	})
+
+	// TODO: additional multus tests (see https://issues.redhat.com/browse/SDN-1180)
+})
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -435,6 +435,7 @@
 // test/extended/testdata/marketplace/opsrc/02-opsrc.yaml
 // test/extended/testdata/multi-namespace-pipeline.yaml
 // test/extended/testdata/multi-namespace-template.yaml
+// test/extended/testdata/net-attach-defs/bridge-nad.yml
 // test/extended/testdata/oauthserver/cabundle-cm.yaml
 // test/extended/testdata/oauthserver/oauth-network.yaml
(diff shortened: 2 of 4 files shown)


test/extended/networking/multus.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: podName,
+		},
+		Spec: kapiv1.PodSpec{
+			Containers: []kapiv1.Container{
+				{
+					Name:    contName,
+					Image:   image.LocationFor("docker.io/centos/httpd-24-centos8:latest"),
+					Command: []string{"/bin/bash", "-c", "trap : TERM INT; sleep infinity & wait"},
```
danwinship, line 89: We want to avoid adding unnecessary new images to the e2e suite. You can use `e2epod.CreateExecPodOrFail` (from `k8s.io/kubernetes/test/e2e/framework/pod`) to create a pod using the standard "agnhost" testing image (pass a callback as the `tweak` arg to that function so you can adjust the pod definition before it's created, eg to add the NAD annotation) and then use `f.ExecShellInPod` or the like to run `ip` in it

test/extended/networking/multus.go
```diff
@@ -0,0 +1,118 @@
...
+					SecurityContext: &kapiv1.SecurityContext{
+						Capabilities: &kapiv1.Capabilities{
+							Add: []kapiv1.Capability{"NET_ADMIN"},
```
danwinship, line 92: You don't actually need this capability do you?

test/extended/networking/multus.go
```diff
@@ -0,0 +1,118 @@
...
+		"k8s.v1.cni.cncf.io/networks": "bridge-nad",
+	}
+
+	_, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.Background(), pod, metav1.CreateOptions{})
+	expectNoError(err)
+	return pod
+}
+
+// waits till pod is ready
+func waitForTestMultusPod(f *e2e.Framework, namespace string, podName string) (string, error) {
```
danwinship, line 111: (this won't be needed if you use `CreateExecPodOrFail` because it does this for you)

danwinship (commented):
Unfortunately most of the existing networking e2e tests in origin are really old and not very well written, so you copied some bad code...

nicklesimba:
Thanks for the extensive feedback! Will be working on addressing these, really appreciate it.

test/extended/networking/multus.go
```diff
@@ -0,0 +1,118 @@
...
+	var ns string // namespace
+
+	oc = exutil.NewCLI("multus-e2e")
+
+	f := oc.KubeFramework()
+	g.It("should ensure that multus-cni functions correctly", func() {
+		makeNamespaceScheduleToAllNodes(f)
+		ns = f.Namespace.Name
+		fmt.Println()
+		o.Expect(testMultus(f, oc, ns)).To(o.Succeed())
```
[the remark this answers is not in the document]
nicklesimba: I'd like to keep this as is, if that's okay. There won't be multiple contexts running the testMultus function, but it will eventually look something like this:

```
g.It("should ensure that multus-cni functions correctly", func() {
	ns = f.Namespace.Name
	o.Expect(testMultus(f, oc, ns, valid_annotation)).To(o.Succeed())
	o.Expect(testMultus(f, oc, ns, invalid_annotation)).NotTo(o.Succeed())
	// Another two or three multus tests will live here in the future.
})
```

Or...I'm also open to a "proper" way of setting such a structure up.

test/extended/networking/multus.go
```diff
@@ -0,0 +1,118 @@
...
+	nad_yaml := exutil.FixturePath("testdata", "net-attach-defs", "bridge-nad.yml")
+	g.By(fmt.Sprintf("calling oc create -f %s", nad_yaml))
+
+	err = oc.AsAdmin().Run("create").Args("-f", nad_yaml).Execute()
+	expectNoError(err)
+
+	oc.AsAdmin().Run("get").Args("network-attachment-definitions").Execute() // Print verification that it's come up
+
+	// 2. Then make a pod with the right annotation to use the net-attach-def
+	_ = launchTestMultusPod(f, podName)
```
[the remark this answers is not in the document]
nicklesimba: This function will be used for other multus e2e tests in the future.

test/extended/networking/multus.go
```diff
@@ -0,0 +1,52 @@
...
+		testPod := frameworkpod.CreateExecPodOrFail(f.ClientSet, ns, podName, func(pod *v1.Pod) {
+			pod.ObjectMeta.Annotations = annotation
+		})
+
+		g.By("checking for net1 interface on pod")
+		output, err := oc.AsAdmin().Run("exec").Args(testPod.Name, "--", "ip", "a", "show", "dev", "net1").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		fmt.Println(output)
+		o.Expect(strings.Contains(output, "net1")).Should(o.Equal(true))
+		o.Expect(strings.Contains(output, "inet 10.10.0.1/24")).Should(o.Equal(true))
```
danwinship, line 48:
```suggestion
		o.Expect(output).Should(o.ContainSubstring("net1"))
		o.Expect(output).Should(o.ContainSubstring("inet 10.10.0.1/24"))
```

test/extended/util/image/zz_generated.txt
```diff
@@ -1,4 +1,5 @@
 # This file is generated by hack/update-generated-bindata.sh
+docker.io/centos/httpd-24-centos8:latest quay.io/openshift/community-e2e-images:e2e-docker-io-centos-httpd-24-centos8-latest-00vUTwFFxuIiiWjJ
```
danwinship, line 2: you need to rerun "make update" or whatever to regenerate this file. (well, or just revert the change to it)

test/extended/networking/multus.go
```diff
@@ -0,0 +1,52 @@
...
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/api/core/v1"
+	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
+)
+
+var _ = g.Describe("[sig-network][Feature:Multus-CNI functionality]", func() {
```
danwinship, line 14: just `[Feature:Multus]` is fine

PR #26353, opened 2021-07-27
bug 1987230: Operators should not create watch channels very often: bump apirequests upperbounds in 4.9

ingvagabund:


test/extended/apiserver/api_requests.go
```diff
@@ -126,9 +126,9 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 				"ingress-operator":                       371,
 				"kube-apiserver-operator":                260,
 				"kube-controller-manager-operator":       145,
-				"kube-storage-version-migrator-operator": 36,
+				"kube-storage-version-migrator-operator": 58,
```
deads2k, line 129: wow, I wonder what we did.

test/extended/apiserver/api_requests.go
```diff
@@ -205,10 +205,10 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 				"console-operator":                       154,
 				"csi-snapshot-controller-operator":       59,
 				"dns-operator":                           59,
-				"etcd-operator":                          162,
+				"etcd-operator":                          191,
```
deads2k, line 208: wow, this is now more than 50% higher than other platforms..

test/extended/apiserver/api_requests.go
```diff
@@ -137,29 +137,29 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+				"kube-storage-version-migrator-operator": 33,
+				"machine-api-operator":                   31,
+				"marketplace-operator":                   11,
 				"openshift-apiserver-operator":           215,
-				"openshift-config-operator":              37,
-				"openshift-controller-manager-operator":  110,
-				"openshift-kube-scheduler-operator":      102,
+				"openshift-config-operator":              39,
+				"openshift-controller-manager-operator":  124,
+				"openshift-kube-scheduler-operator":      116,
```
soltysh, line 162: 10% for both kcm-o and ks-o would be nice to check why is that

soltysh (approved):
/approve

PR #26354, opened 2021-07-28
bug 1987230: Single node apirequestscount upper bounds

ingvagabund:
Increasing the bounds based on
```
aws-ebs-csi-driver-operator, watchrequestcount=284, upperbound=216, ratio=1.3148148148148149
cloud-credential-operator, watchrequestcount=127, upperbound=124, ratio=1.0241935483870968
cluster-monitoring-operator, watchrequestcount=70, upperbound=64, ratio=1.09375
cluster-storage-operator, watchrequestcount=403, upperbound=310, ratio=1.3
csi-snapshot-controller-operator, watchrequestcount=197, upperbound=104, ratio=1.8942307692307692
etcd-operator, watchrequestcount=328, upperbound=250, ratio=1.312
kube-storage-version-migrator-operator, watchrequestcount=136, upperbound=72, ratio=1.8888888888888888
openshift-apiserver-operator, watchrequestcount=514, upperbound=452, ratio=1.1371681415929205
openshift-config-operator, watchrequestcount=99, upperbound=94, ratio=1.053191489361702
openshift-controller-manager-operator, watchrequestcount=360, upperbound=298, ratio=1.2080536912751678
service-ca-operator, watchrequestcount=261, upperbound=214, ratio=1.219626168224299
```
to make the test less blocking. We may decrease the bounds later once the number of api requests is reduced.

### DIFF
--- a/test/extended/apiserver/api_requests.go
+++ b/test/extended/apiserver/api_requests.go
@@ -10,7 +10,6 @@ import (
 	o "github.com/onsi/gomega"
 	apiserverv1 "github.com/openshift/api/apiserver/v1"
 	configv1 "github.com/openshift/api/config/v1"
-	v1 "github.com/openshift/api/config/v1"
 	apiserverclientv1 "github.com/openshift/client-go/apiserver/clientset/versioned/typed/apiserver/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/origin/pkg/test/ginkgo/result"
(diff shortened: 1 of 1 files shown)


test/extended/apiserver/api_requests.go
```diff
@@ -278,8 +279,49 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+				"openshift-controller-manager-operator":  180,
+				"openshift-kube-scheduler-operator":      179,
+				"prometheus-operator":                    90,
+				"service-ca-operator":                    131,
+			},
+		}
+
+		var upperBound platformUpperBound
+
+		if infra.Status.ControlPlaneTopology == v1.SingleReplicaTopologyMode {
```
omertuc, line 315: Maybe a comment / commit message explaining why SNO needs higher bounds? I guess even "we're not sure" will be better than nothing. Since it's not really obvious.

eranco74 (commented):
/lgtm

test/extended/apiserver/api_requests.go
ingvagabund: All the upper bounds are measured experimentally. So "we are not sure" in any cases. It's all a guess based on the CI measurements. The entire test should be treated as such. The main idea is to have a test which will tell us if the upper bound was exceeded ad more importantly how much. To see if the increase is linear or exponential.

omertuc: Fair enough

test/extended/apiserver/api_requests.go
```diff
@@ -10,6 +10,7 @@ import (
 	o "github.com/onsi/gomega"
 	apiserverv1 "github.com/openshift/api/apiserver/v1"
 	configv1 "github.com/openshift/api/config/v1"
+	v1 "github.com/openshift/api/config/v1"
```
soltysh, line 13: Nit: you have these duplicated

soltysh (approved):
/lgtm
/approve

PR #26358, opened 2021-07-29
Count only events that are observed during tests run

dmage:


pkg/synthetictests/events.go
```diff
@@ -72,19 +79,45 @@ func testDuplicatedEvents(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
+			continue
+		}
+
+		prev, ok := displayToCount[eventDisplayMessage]
+		// If we see the event for the first time, but it has already happened
+		// multiple, usually it means that previous events happened before the
+		// tests run. There might be exceptions (for example, combined events)
+		// if the event message changes.
+		if !ok && times > 1 && !strings.Contains(eventDisplayMessage, "(combined from similar events): ") {
+			displayToCount["before tests: "+eventDisplayMessage] = times - 1
```
deads2k, line 95: I don't agree with this change.  The fact that there's an ongoing issue is a problem.

dmage: It's not an ongoing issue. Let's say you have a pod with 15 readiness probe failures during installation, and it has 7 failures due to reconfiguration that is made by a test. In total it has 22 failures, but they happened at complete different times and have different reasons. An event counter has a flaw that these events are counted together.

pkg/synthetictests/events.go
```diff
@@ -72,19 +79,45 @@ func testDuplicatedEvents(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
 
 	displayToCount := map[string]int{}
+	ignoredCounts := map[string]int{}
 	for _, event := range events {
 		eventDisplayMessage, times := getTimesAnEventHappened(fmt.Sprintf("%s - %s", event.Locator, event.Message))
-		if times > 20 {
-			if knownEvents.MatchString(eventDisplayMessage) {
-				continue
-			}
+		if times == 0 {
```
bparees, line 85: isn't times always at least 1 if we saw the event?  or is this a count of how many times it repeated after the first instance?

pkg/synthetictests/events.go
```diff
@@ -72,19 +79,45 @@ func testDuplicatedEvents(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
+			continue
+		}
+
+		prev, ok := displayToCount[eventDisplayMessage]
+		// If we see the event for the first time, but it has already happened
+		// multiple, usually it means that previous events happened before the
```
bparees, line 91: i'm confused how this test works....I thought that it just ran after the e2es and looked at all the events in the cluster, but it sounds like it's continuously getting run each time an event is observed in the cluster?

bparees: can't we use/fix https://github.com/openshift/origin/blob/master/pkg/test/ginkgo/cmd_runsuite.go#L330 to ensure we only see events that have a timestamp that occurred during test execution?  We know when we started running tests(and when we finished)

dmage: The monitor collects all events from the watcher channel, during capturing we convert their property `count` into the message: https://github.com/openshift/origin/blob/9bc7ae93d1596a9bcb3266a0d370ced270c02c54/pkg/monitor/event.go#L75-L78
This `count` property has interesting properties, it knows about the object UID, so it doesn't deduplicate events for different objects with the same name. If I create a pod, then delete it, and then create it again, I won't have an event like _pod foo created (2 times)_.

Then indeed we select only events that happened during the test run
https://github.com/openshift/origin/blob/9bc7ae93d1596a9bcb3266a0d370ced270c02c54/pkg/test/ginkgo/cmd_runsuite.go#L353
but `count` for these events (or it's better to say `message`, as we haven't saved `count` as a separate field) stay the same.

Once the tests are over, the collected events will be handled into this function and it should detect pathologies. There are several options to define "repeated events":

1. Use `count`, but it remembers what happened during the installation.
2. Count events by ourselves, but we don't have UIDs at this point, so we'll smash together events like `pod foo created`.

The current implementation uses the first approach. But I also added a heuristic that tries to adjust `count` as if we didn't have any events before the tests.

pkg/synthetictests/events.go
dmage: That's an implementation oddity. This function returns `"", 0` if the message doesn't have the suffix `(X times)`. This suffix is added only if the k8s event met the condition `event.Count > 1`.

dmage:
This was replaced by ignoreFns approach.

PR #26359, opened 2021-07-29
Bug 1979962: Fix aws network stress tests

npinaeva:
Failing job - https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.9-e2e-aws-network-stress/1420053950479994880
The most often failing tests are:
```
[sig-network] Internal connectivity for TCP and UDP on ports 9000-9999 is allowed [Suite:openshift/conformance/parallel]
[sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
```
Add [Serial:Self] label for tests that can't be run in parallel with a copy of itself, exclude these tests from "openshift/network/stress" suite.
"Internal connectivity for TCP and UDP on ports 9000-9999 is allowed" uses exclusive resources (node-role.kubernetes.io/master + ports 9000,9999) and can't be run in parallel with multiple copies
"HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol" also uses host ip + port

### DIFF
--- a/test/extended/networking/internal_ports.go
+++ b/test/extended/networking/internal_ports.go
@@ -35,7 +35,7 @@ const (
 var _ = ginkgo.Describe("[sig-network] Internal connectivity", func() {
 	f := framework.NewDefaultFramework("k8s-nettest")
 
-	ginkgo.It("for TCP and UDP on ports 9000-9999 is allowed", func() {
+	ginkgo.It("for TCP and UDP on ports 9000-9999 is allowed [Serial:Self]", func() {
 		e2eskipper.SkipUnlessNodeCountIsAtLeast(2)
 		clientConfig := f.ClientConfig()
 
--- a/cmd/openshift-tests/e2e.go
+++ b/cmd/openshift-tests/e2e.go
@@ -321,6 +321,10 @@ var staticSuites = testSuites{
 				if strings.Contains(name, "[Feature:NetworkPolicy]") {
 					return false
 				}
+				// Serial:Self are tests that can't be run in parallel with a copy of itself
+				if strings.Contains(name, "[Serial:Self]") {
+					return false
+				}
 				return (strings.Contains(name, "[Suite:openshift/conformance/") && strings.Contains(name, "[sig-network]")) || isStandardEarlyOrLateTest(name)
 			},
 			Parallelism:         60,
--- a/test/extended/util/annotate/generated/zz_generated.annotations.go
+++ b/test/extended/util/annotate/generated/zz_generated.annotations.go
@@ -1783,7 +1783,7 @@ var annotations = map[string]string{
 
 	"[Top Level] [sig-network] Firewall rule should have correct firewall rules for e2e cluster": "should have correct firewall rules for e2e cluster [Disabled:SpecialConfig] [Suite:k8s]",
 
-	"[Top Level] [sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]": "validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]",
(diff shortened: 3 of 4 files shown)


npinaeva:
original job that was failing - https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.9-e2e-aws-network-stress/1420053950479994880
test job for this pr - https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1420790499387641856
I'm planning to fix this test too  
`[sig-network][Feature:Router] The HAProxy router should expose prometheus metrics for a route [Skipped:Disconnected] [Suite:openshift/conformance/parallel]` as it is similar to `The HAProxy router should enable openshift-monitoring to pull metrics` test fixed in this pr

npinaeva:
Test job - https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1424627533705383936
At least no failures for mentioned tests.

npinaeva:
@smarterclayton can you please confirm that `[Serial:Self]` label is implemented the way you suggested?

jluhrsen:
this /lgtm. I was hoping to run the e2e-network-test presubmit here, but looks like it's not available here. I think the
[job](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1424627533705383936) you linked in comment above is showing what we expect.

test/extended/networking/internal_ports.go
```diff
@@ -35,7 +35,7 @@ const (
 var _ = ginkgo.Describe("[sig-network] Internal connectivity", func() {
 	f := framework.NewDefaultFramework("k8s-nettest")
 
-	ginkgo.It("for TCP and UDP on ports 9000-9999 is allowed", func() {
+	ginkgo.It("for TCP and UDP on ports 9000-9999 is allowed [Serial:Self]", func() {
```
danwinship, line 38: There's no reason the test has to always use ports 9000 and 9999 exactly... it should just pick two random ports in that range.

Although I guess then you have to worry about conflicting with other components that use ports in that range, so I guess it will need some retrying if the first attempt at pod creation fails. (But then, that makes it more resilient against other copies of itself too.)

Anyway, if not that, I don't think we want the `[Serial:Self]` tag; the `[Serial]` and `[Disruptive]` tags are understood to cover a pretty broad range of incompatible test behaviors, so you can just use `[Serial]` for this case.

npinaeva: I agree that if we make it choose random ports, there is no need for this tag, but then we need to add some try-cycle for different port numbers so that the test won't fail if it chooses some service's port. Also finding out if deployment fails because of someone legally using chosen port or not may be difficult and time consuming. So maybe using just edge port values is not a bad idea, what do you think?

But in general, `[Serial:Self]` tests are only excluded from network-stress tests, and basically any test that uses some exclusive resources should be marked like that, but using `[Serial]` label will exclude the test from all parallel test suits (where it passes). And  I think we don't want to exclude tests for no reason.

Also, the problem with running copies of the same test in parallel may be fixed by changing test scheduler behaviour by making it schedule the next copy of the same test only when the previous one is finished, and when this is done, we can remove  `[Serial:Self]` label.

test/extended/router/metrics.go
```diff
@@ -257,7 +257,7 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 			}()
 
 			o.Expect(wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
-				contents, err := getBearerTokenURLViaPod(ns, execPod.Name, fmt.Sprintf("%s/api/v1/targets", prometheusURL), token)
+				contents, err := getBearerTokenURLViaPod(ns, execPod.Name, fmt.Sprintf("%s/api/v1/targets?state=active", prometheusURL), token)
```
smarterclayton, line 260: This belongs in its own commit with a comment describing why it was necessary.

npinaeva: It's already merged https://github.com/openshift/origin/commit/0fbe9178794c72636b372d10d6132b6fac82b372

PR #26361, opened 2021-07-30
Bug 1988264: initializeTestFramework: initialize TextContext ip family

fedepaol:
Some tests rely on the TestContext IPFamily field, that is being initialized in the upstream ginkgo suites but not in origin's one.
Here we add the initialization in initializeTestFramework based on config.HasIPV6 / config.HasIPV4.

This is the manual cherry-pick of https://github.com/openshift/origin/pull/26140

### DIFF
--- a/test/extended/util/annotate/rules.go
+++ b/test/extended/util/annotate/rules.go
@@ -51,11 +51,8 @@ var (
 			// https://bugzilla.redhat.com/show_bug.cgi?id=1953478
 			`\[sig-storage\] Dynamic Provisioning Invalid AWS KMS key should report an error and create no PV`,
 
-			// https://bugzilla.redhat.com/show_bug.cgi?id=1956989
-			`\[sig-network\] Services should be possible to connect to a service via ExternalIP when the external IP is not assigned to a node`,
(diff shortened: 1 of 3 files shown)


test/extended/util/annotate/rules.go
```diff
@@ -184,6 +178,7 @@ var (
 			`\[sig-network\]\[Feature:Router\] The HAProxy router should serve the correct routes when scoped to a single namespace and label set`,
 			`\[sig-network\]\[Feature:Router\] The HAProxy router should set Forwarded headers appropriately`,
 			`\[sig-network\]\[Feature:Router\] The HAProxy router should support reencrypt to services backed by a serving certificate automatically`,
+			`\[sig-network\] Networking should provide Internet connection for containers \[Feature:Networking-IPv6\]`,
```
danwinship, line 181: this needs to remain in `[Disabled:Broken]` as in #26362

danwinship (changes requested):

danwinship:
/lgtm cancel
(as above, one of the skips needs to be changed to basically merge in a later fix from master)

fedepaol:
> /lgtm cancel
> (as above, one of the skips needs to be changed to basically merge in a later fix from master)

Done

fedepaol:
@danwinship ptal

PR #26363, opened 2021-07-31
temporary search for memory.max in build container during quota test

gabemontero:


gabemontero:
/assign @adambkaplan 

The quota test needs a bit more tweaking, as the cgroup v2 `memory.max` file is not exactly where the script is anticipating it to be.  So I have added some find's of the file, which we can come back and remove once it is all sorted out.  Once this merges, the next cgroupv2 run in https://github.com/openshift/builder/pull/252 should confirm exactly if/where the `memory.max` file is located in the build container, and we can go back and tweak the assemble script as needed.

The last failed run there shows the current changes in openshift/origin now firing for both cgroup v1 and v2.  It is just not finding either.  The failing test debug also shows the util_linux.go files changes in this PR are running as expected, with the build pod's memory.max file being located and the correct memory limit getting pulled.  It is just a question of where in the build container to find it.  Perhaps with the build container being in a privileged pod, it is in the pod specific subdir ??? ... the find should confirm.

gabemontero:
> /assign @adambkaplan
> 
> The quota test needs a bit more tweaking, as the cgroup v2 `memory.max` file is not exactly where the script is anticipating it to be. So I have added some find's of the file, which we can come back and remove once it is all sorted out. Once this merges, the next cgroupv2 run in [openshift/builder#252](https://github.com/openshift/builder/pull/252) should confirm exactly if/where the `memory.max` file is located in the build container, and we can go back and tweak the assemble script as needed.
> 
> The last failed run there shows the current changes in openshift/origin now firing for both cgroup v1 and v2. It is just not finding either. The failing test debug also shows the util_linux.go files changes in this PR are running as expected, with the build pod's memory.max file being located and the correct memory limit getting pulled. It is just a question of where in the build container to find it. Perhaps with the build container being in a privileged pod, it is in the pod specific subdir ??? ... the find should confirm.

with @adambkaplan on PTO, can you approve @bparees ... per the details above, I need these finds short term to nail down in https://github.com/openshift/builder/pull/252 where to look for the `memory.max` file in cgroups v2.

Unfortunately, although maybe @vrutkovs can correct me if I am wrong here, there is no way yet to launch a cgroupv2 cluster via cluster-bot.

Once this merges and we sort out the location, I'll then change this file again to go to the file in question.  And if there are more changes needed in https://github.com/openshift/builder/pull/252 to get the file in the expected location, we'll drive that as well.  But ultimately, I believe we can remove these find's eventually.

/assign @coreydaley in case @bparees just wants to approve and wants an lgtm from someone on the team.

thanks

PR #26371, opened 2021-08-03
Bug 1988828: fix test - reduce minimum size of expected files in must-gather tests

eranco74:
The "oc adm must-gather" test in the e2e-aws-single-node suite is failing because
the openshift-apiserver.audit_logs_listing and oauth-apiserver.audit_logs_listing files content is only 38 bytes
while the test is expecting the size to be at least 50 bytes.
Since single node has just one host the file content is shorter.
This commit change the minimum required size to 20 bytes.

https://bugzilla.redhat.com/show_bug.cgi?id=1988828


soltysh (approved):
/lgtm
/approve

PR #26373, opened 2021-08-03
bug 1989961: refine apiserver downtime sampling and calculations

deads2k:
Downtime calculations didn't consider the average downtime *before* detection, so a minimum value is required.  This value is different than cases where the value is actually zero.

The intervals also calculated the startTime of a failure from when the failed request returned as opposed to when the failed request was originally made.  This can result in short downtime calculations, when the actual downtime was significant.


deads2k:
added revert

PR #26375, opened 2021-08-03
allow console readiness probe failures only during the initial installation

deads2k:
This PR also splits the allow list into upgrade and normal operation.

/assign @dmage


deads2k:
proof worked

/hold cancel

pkg/synthetictests/duplicated_events.go
```diff
@@ -21,7 +29,7 @@ func combinedRegexp(arr ...*regexp.Regexp) *regexp.Regexp {
 	return regexp.MustCompile(s)
 }
 
-var knownEvents = combinedRegexp(
+var allowedRepeatedEventPatterns = []*regexp.Regexp{
```
aojea, line 32: allowedRepeatedEventPatterns take precedence over allowedRepeatedEventFns,  if an event match the former it will no be evaluated by the later, right?

pkg/synthetictests/duplicated_events.go
```diff
@@ -64,20 +76,75 @@ var knownEventProblems = []struct {
 	},
 }
 
+type duplicateEventsEvaluator struct {
```
aojea, line 79: if you foresee more uses of the client config here, maybe you can make it part of the struct and not having to plumb it down as an argument

deads2k: > if you foresee more uses of the client config here, maybe you can make it part of the struct and not having to plumb it down as an argument

it's part of the signature on synthetics now, so passing it through is consistent with others.

pkg/synthetictests/duplicated_events.go
deads2k: > allowedRepeatedEventPatterns take precedence over allowedRepeatedEventFns, if an event match the former it will no be evaluated by the later, right?

True.   Effectively that makes it a logical OR which I think people expect.

pkg/monitor/api.go
```diff
@@ -33,40 +33,45 @@ const (
 	LocatorOAuthAPIServerReusedConnection     = "oauth-apiserver-reused-connection"
 )
 
-// Start begins monitoring the cluster referenced by the default kube configuration until
-// context is finished.
-func Start(ctx context.Context) (*Monitor, error) {
-	m := NewMonitorWithInterval(time.Second)
+func GetMonitorRESTConfig() (*rest.Config, error) {
```
aojea, line 36: I fail to understand why you modify Start() and extract this function, is something related to the openshift tests structure?

pkg/synthetictests/duplicated_events.go
```diff
@@ -147,3 +214,85 @@ func getTimesAnEventHappened(message string) (string, int) {
...
+	}
+
+	// this block gets the actual event from the API.  This is ugly, but necessary because we don't have real events.
+	// It may be interesting to track real events, but it would be expensive.
+	kubeClient, err := kubernetes.NewForConfig(kubeClientConfig)
+	if err != nil {
+		// default to OK
+		return true
+	}
+	consoleEvents, err := kubeClient.CoreV1().Events("openshift-console").List(context.TODO(), metav1.ListOptions{})
```
aojea, line 273: we are going to query this list for each event?
is this ok or can be overwhelming?

deads2k: > we are going to query this list for each event?
> is this ok or can be overwhelming?

Good question.  Since we decimate the  list before, we only do this for readiness probe failures for console pods in their namespace.  About 8 or so of these usually.

pkg/monitor/api.go
deads2k: > I fail to understand why you modify Start() and extract this function, is something related to the openshift tests structure?

Yeah.  trying to keep the clients consistent and "obvious" in the flow.

aojea:
LGTM

deads2k:
test only change to make it easier to selectively exclude events.  overriding bug requirements

PR #26377, opened 2021-08-04
Bug 1986003: Retry kubeconfig checks, when kube-apiserver is temporarily unavailable

soltysh:
/assign @p0lyn0mial 

since you complained about this particular to me some time ago, maybe not necessarily about this particular problem but it's a good starting point :wink: which I noticed fails pretty frequently

### DIFF
--- a/test/extended/apiserver/kubeconfigs.go
+++ b/test/extended/apiserver/kubeconfigs.go
@@ -3,12 +3,15 @@ package apiserver
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
-	exutil "github.com/openshift/origin/test/extended/util"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
 )
 
 var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local kubeconfig", func() {
@@ -29,13 +32,29 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
 			framework.Logf("Discovered %d master nodes.", len(masterNodes.Items))
 			o.Expect(masterNodes.Items).NotTo(o.HaveLen(0))
 			for _, master := range masterNodes.Items {
-				g.By("Testing master node " + master.Name)
-				kubeconfigPath := "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/" + kubeconfig
-				framework.Logf("Verifying kubeconfig %q on master %s", master.Name)
-				out, err := oc.AsAdmin().Run("debug").Args("node/"+master.Name, "--", "chroot", "/host", "/bin/bash", "-euxo", "pipefail", "-c", fmt.Sprintf(`oc --kubeconfig "%s" get namespace kube-system`, kubeconfigPath)).Output()
+				retry, err := testNode(oc, kubeconfig, master.Name)
+				for retries := 2; retries > 0; retries-- {
+					if !retry {
+						break
+					}
+					g.By("There was a retryable error for " + fmt.Sprintf("%s/%s", master.Name, kubeconfig))
+					retry, err = testNode(oc, kubeconfig, master.Name)
+				}
 				o.Expect(err).NotTo(o.HaveOccurred())
-				framework.Logf(out)
 			}
 		})
 	}
 })
+
+func testNode(oc *exutil.CLI, kubeconfig, masterName string) (bool, error) {
+	g.By("Testing master node " + masterName)
+	kubeconfigPath := "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/" + kubeconfig
+	framework.Logf("Verifying kubeconfig %q on master %q", kubeconfig, masterName)
+	out, err := oc.AsAdmin().Run("debug").Args("node/"+masterName, "--", "chroot", "/host", "/bin/bash", "-euxo", "pipefail", "-c",
+		fmt.Sprintf(`oc --kubeconfig "%s" get namespace kube-system`, kubeconfigPath)).Output()
+	framework.Logf(out)
+	// retry error when kube-apiserver was temporarily unavailable, this matches oc error coming from:
+	// https://github.com/kubernetes/kubernetes/blob/cbb5ea8210596ada1efce7e7a271ca4217ae598e/staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go#L237-L243
+	matched, _ := regexp.MatchString("The connection to the server .+ was refused - did you specify the right host or port", out)
+	return !matched, err
+}


test/extended/apiserver/kubeconfigs.go
```diff
@@ -29,13 +32,25 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
...
+
+func testNode(oc *exutil.CLI, kubeconfig, masterName string) (bool, error) {
+	g.By("Testing master node " + masterName)
+	kubeconfigPath := "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/" + kubeconfig
+	framework.Logf("Verifying kubeconfig %q on master %q", kubeconfig, masterName)
+	out, err := oc.AsAdmin().Run("debug").Args("node/"+masterName, "--", "chroot", "/host", "/bin/bash", "-euxo", "pipefail", "-c",
+		fmt.Sprintf(`oc --kubeconfig "%s" get namespace kube-system`, kubeconfigPath)).Output()
+	framework.Logf(out)
+	// retry error when kube-apiserver was temporarily unavailable
+	retry := strings.Contains(out, "The connection to the server localhost:6443 was refused")
```
p0lyn0mial, line 54: would you be open to adding a precondition before running the test instead? 
not sure this is the only error we might get, for example on an IPv6 cluster we might get `[::1]:6443: connect: connection refused`

we could check if the cluster is in stable condition (not progressing, the pods on the same revision) for X min.

p0lyn0mial:
> since you complained about this particular to me some time ago, maybe not necessarily about this particular problem but it's a good starting point 😉 which I noticed fails pretty frequently

wasn't me :)

test/extended/apiserver/kubeconfigs.go
```diff
@@ -29,13 +32,25 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
 			framework.Logf("Discovered %d master nodes.", len(masterNodes.Items))
 			o.Expect(masterNodes.Items).NotTo(o.HaveLen(0))
 			for _, master := range masterNodes.Items {
-				g.By("Testing master node " + master.Name)
-				kubeconfigPath := "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/" + kubeconfig
-				framework.Logf("Verifying kubeconfig %q on master %s", master.Name)
-				out, err := oc.AsAdmin().Run("debug").Args("node/"+master.Name, "--", "chroot", "/host", "/bin/bash", "-euxo", "pipefail", "-c", fmt.Sprintf(`oc --kubeconfig "%s" get namespace kube-system`, kubeconfigPath)).Output()
+				retry, err := testNode(oc, kubeconfig, master.Name)
+				if retry {
```
aojea, line 36: is one retry enough?

soltysh: yeah, after consideration I did 2 :wink:

soltysh:
@aojea updated, ptal

test/extended/apiserver/kubeconfigs.go
soltysh: Checking progressing might slow this down, but we can add that as last resort, re-tries seems simpler, b/c theoretically we can even pass the test when rollout is in progress.

soltysh: Added check for ipv6, though

soltysh:
or @p0lyn0mial

test/extended/apiserver/kubeconfigs.go
```diff
@@ -29,13 +32,29 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
...
+func testNode(oc *exutil.CLI, kubeconfig, masterName string) (bool, error) {
+	g.By("Testing master node " + masterName)
+	kubeconfigPath := "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/" + kubeconfig
+	framework.Logf("Verifying kubeconfig %q on master %q", kubeconfig, masterName)
+	out, err := oc.AsAdmin().Run("debug").Args("node/"+masterName, "--", "chroot", "/host", "/bin/bash", "-euxo", "pipefail", "-c",
+		fmt.Sprintf(`oc --kubeconfig "%s" get namespace kube-system`, kubeconfigPath)).Output()
+	framework.Logf(out)
+	// retry error when kube-apiserver was temporarily unavailable
+	retry := strings.Contains(out, "The connection to the server localhost:6443 was refused") ||
+		strings.Contains(out, "[::1]:6443: connect: connection refused")
```
aojea, line 58: what are the messages so different between protocols? is just curiosity , using curl they are more consisten
```
$ curl localhost:6443
curl: (7) Failed to connect to localhost port 6443: Connection refused
$ curl [::1]:6443
curl: (7) Failed to connect to ::1 port 6443: Connection refused
$ curl 127.0.0.1:6443
curl: (7) Failed to connect to 127.0.0.1 port 6443: Connection refused
```

soltysh:
> lgtm, just a question to double check the "retry error messages"

I just checked that, you're right the message should be identical, only the address will change, it's coming from here: https://github.com/kubernetes/kubernetes/blob/cbb5ea8210596ada1efce7e7a271ca4217ae598e/staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go#L237-L243, so I've updated accordingly the PR.  

With:
```golang

File: test/extended/apiserver/kubeconfigs.go [hunk 4]
```diff
@@ -29,13 +32,29 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
...
+
+func testNode(oc *exutil.CLI, kubeconfig, masterName string) (bool, error) {
+	g.By("Testing master node " + masterName)
+	kubeconfigPath := "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/" + kubeconfig
+	framework.Logf("Verifying kubeconfig %q on master %q", kubeconfig, masterName)
+	out, err := oc.AsAdmin().Run("debug").Args("node/"+masterName, "--", "chroot", "/host", "/bin/bash", "-euxo", "pipefail", "-c",
+		fmt.Sprintf(`oc --kubeconfig "%s" get namespace kube-system`, kubeconfigPath)).Output()
+	framework.Logf(out)
+	// retry error when kube-apiserver was temporarily unavailable, this matches oc error coming from:
+	// https://github.com/kubernetes/kubernetes/blob/cbb5ea8210596ada1efce7e7a271ca4217ae598e/staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go#L237-L243
```
aojea, line 57: the message you've linked ends with a quesiton tag, but the regex work fine
https://play.golang.org/p/aBGVR3_3YsQ

soltysh: Yeah, I didn't want to be that precise :wink:

aojea: those symbols on regexps :grimacing: 
:rofl:

soltysh:
/override ci/prow/e2e-aws-single-node
These are not required and I'll continue debugging test-cmd separately.

PR #26379, opened 2021-08-05
Add apiserver pod readiness check for kubeconfig tests

ardaguclu:
`local kubeconfig * should be present on all masters and work` tests are flaky. The reason is installer sometimes upgrades new revisions of apiserver pods(probably another test triggers this) and within this time window, above kubeconfig tests get `The connection to the server localhost:6443 was refused` errors and fail.

These kubeconfig tests assume that apiserver pods are running and healthy state. That's why, this PR adds readiness check for them prior to execute real test commands.

### DIFF
--- a/test/extended/apiserver/kubeconfigs.go
+++ b/test/extended/apiserver/kubeconfigs.go
@@ -3,11 +3,13 @@ package apiserver
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


soltysh:
/close 
in favor of https://github.com/openshift/origin/pull/26377

PR #26380, opened 2021-08-05
add several skip cases for pathological events

deads2k:
builds on https://github.com/openshift/origin/pull/26375, only the last commit is pertinent.

Started burning down the list of pathological events.


pkg/synthetictests/duplicated_events.go
```diff
@@ -64,18 +64,62 @@ var knownEventProblems = []struct {
...
+		knownRepeatedEventsBugs:      knownEventsBugs,
+	}
+
+	return evaluator.testDuplicatedEvents(events)
+}
+
+// isRepeatedEventOKFunc takes a monitorEvent as input and returns true if the repeated event is OK.
+// this commonly happens for known bugs and for cases where events are repeated intentionally by tests.
+// the string is the message to display for the failure.
+type isRepeatedEventOKFunc func(monitorEvent monitorapi.EventInterval) (bool, string)
```
wking: nit: thoughts about:

```go
// checkRepeatedEventFunc takes a monitorEvent as input and returns an error if the repeated event is not ok.  Returns nil for known bugs, events repeated intentionally by tests, and other acceptable events.
type checkRepeatedEventFunc func(monitorEvent monitorapi.EventInterval) error
```

So you don't have to explain a `bool, string` return combo?

pkg/synthetictests/duplicated_events.go
```diff
@@ -77,40 +89,40 @@ type knownProblem struct {
...
 
-	return evaluator.testDuplicatedEvents(events)
+	return evaluator.testDuplicatedEvents(events, kubeClientConfig)
 }
 
 // isRepeatedEventOKFunc takes a monitorEvent as input and returns true if the repeated event is OK.
 // this commonly happens for known bugs and for cases where events are repeated intentionally by tests.
 // the string is the message to display for the failure.
-type isRepeatedEventOKFunc func(monitorEvent monitorapi.EventInterval) (bool, string)
+type isRepeatedEventOKFunc func(monitorEvent monitorapi.EventInterval, kubeClientConfig *rest.Config) bool
```
wking, line 154: here you're dropping the string response, but still talk about it in the godocs two lines up.  I [still][1] think it would be more convenient as an `error` return.  And the `context.TODO()` in `isConsoleReadinessDuringInstallation` suggests it should take a `ctx context.Context` argument too.

[1]: https://github.com/openshift/origin/pull/26380#discussion_r683889039

pkg/synthetictests/duplicated_events.go
```diff
@@ -191,3 +214,85 @@ func getTimesAnEventHappened(message string) (string, int) {
...
+// isConsoleReadinessDuringInstallation returns true if the event is for console readiness and it happens during the
+// initial installation of the cluster.
+// we're looking for something like
+// > ns/openshift-console pod/console-7c6f797fd9-5m94j node/ip-10-0-158-106.us-west-2.compute.internal - reason/ProbeError Readiness probe error: Get "https://10.129.0.49:8443/health": dial tcp 10.129.0.49:8443: connect: connection refused
+// with a firstTimestamp before the cluster completed the initial installation
+func isConsoleReadinessDuringInstallation(monitorEvent monitorapi.EventInterval, kubeClientConfig *rest.Config) bool {
+	if !strings.Contains(monitorEvent.Locator, "ns/openshift-console") {
+		return false
+	}
+	if !strings.HasPrefix(monitorEvent.Locator, "pod/console-") {
```
wking, line 262: your example a few lines up in the godocs opens with `ns/openshift-console `, so would expect `Contains` instead of `HasPrefix` here.

pkg/synthetictests/duplicated_events.go
```diff
@@ -191,3 +214,85 @@ func getTimesAnEventHappened(message string) (string, int) {
...
+	}
+	if !strings.Contains(monitorEvent.Locator, "Readiness probe") {
+		return false
+	}
+	if !strings.Contains(monitorEvent.Locator, "connect: connection refused") {
+		return false
+	}
+	tokens := strings.Split(monitorEvent.Locator, " ")
+	tokens = strings.Split(tokens[1], "/")
+	podName := tokens[1]
```
wking, line 273: if we're going to split into tokens, can we just make a `map[string][string]` instead of using the `Locator` `Contains` business above?  It is unlikely that `ns/openshift-console` is a substring of something in the `Locator` that is not the namespace entry, but it feels weird to do some unstructured checking before unmarshalling when we could unmarshal and then do structured checking.  And then here, when you just assume that the second token is the pod name, you could use `tokens["pod"]` to conveniently get that pod name without being as brittle.

pkg/synthetictests/duplicated_events.go
```diff
@@ -191,3 +214,85 @@ func getTimesAnEventHappened(message string) (string, int) {
...
+		return true
+	}
+	initialInstallHistory := clusterVersion.Status.History[len(clusterVersion.Status.History)-1]
+	if initialInstallHistory.CompletionTime == nil {
+		// default to OK
+		return true
+	}
+
+	// this block gets the actual event from the API.  This is ugly, but necessary because we don't have real events.
+	// It may be interesting to track real events, but it would be expensive.
```
wking, line 302: would it be expensive to track the real event IDs, so you could ask for them by name here instead of pulling them out of a list response?

pkg/synthetictests/duplicated_events.go
```diff
@@ -38,11 +38,31 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
...
-	// [sig-node] Probing container should *not* be restarted with a non-local redirect http liveness probe [Suite:openshift/conformance/parallel] [Suite:k8s]
-	regexp.MustCompile(`ns/e2e-container-probe-[0-9]+ pod/liveness-[0-9a-f-]+ node/[a-z0-9.-]+ - reason/ProbeWarning Liveness probe warning: <a href="http://0\.0\.0\.0/">Found</a>\.\n\n`),
+	// [sig-node] Probing container ***
+	// these tests intentionally cause repeated probe failures to ensure good handling
+	regexp.MustCompile(`ns/e2e-container-probe-[0-9]+ .* probe failed: `),
+	regexp.MustCompile(`ns/e2e-container-probe-[0-9]+ .* probe warning: `),
+
+	// Kubectl Port forwarding ***
+	// The same pod name is used many times for all these tests with a tight readiness check to make the tests fast.
+	// This results in hundreds of events while the  pod isn't ready.
```
wking, line 48: nit: `the  pod` -> `the pod`, removing a doubled space

deads2k:
this only touches a recently added test, bypassing BZ requirements.

Slack thread says this looks ok and can be refined later.

Need to stop the bleeding on master, so forcing this in.

pkg/synthetictests/duplicated_events.go
deads2k: > if we're going to split into tokens, can we just make a `map[string][string]` instead of using the `Locator` `Contains` business above? It is unlikely that `ns/openshift-console` is a substring of something in the `Locator` that is not the namespace entry, but it feels weird to do some unstructured checking before unmarshalling when we could unmarshal and then do structured checking. And then here, when you just assume that the second token is the pod name, you could use `tokens["pod"]` to conveniently get that pod name without being as brittle.

This is a good general idea given the event compression format.  I think I'm going to try to move the other way and keep all the event content as a real event.

PR #26387, opened 2021-08-09
Bug 1981736: Fix a lock contention problem in the proxy

danwinship:
(oops, filed against ose before, refiling against origin now)

Using the mutex in `proxier.Sync()` to protect `proxier.changesPending` means that a call to `proxier.Sync()` can end up blocking waiting for a call to `proxier.syncProxyRules()` to finish, which can take a long time.

In particular, in a large cluster `Sync()` will get called thousands of times during startup, and some of those calls will block waiting on `syncProxyRules()` calls, which may take a few seconds, and "thousands of times" times "a few seconds" can add up to 10 or 20 minutes of delays in synchronizing.

Fix this by using atomics to set/get `changesPending` rather than fighting with `syncProxyRules()` for the mutex.

----

fixup to https://github.com/openshift/origin/pull/25641; this is fallout from our backporting the iptables performance fixes to 3.11. (The bug is specific to how we hacked the backport into 3.11; no 4.x releases have the same problem.)

I don't have a cluster where I can reproduce this, so I'm not totally sure it's right, so sanity-checking my logic is appreciated.

(The code for `setChangesPending`/`hasChangesPending` is just copied from `setInitialized`/`isInitialized`)

/cc @tssurya @dcbw @aojea @squeed

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go
+++ b/vendor/k8s.io/kubernetes/pkg/proxy/iptables/proxier.go
@@ -182,7 +182,7 @@ type Proxier struct {
 	endpointsSynced bool
 	servicesSynced  bool
 	initialized     int32
-	changesPending  bool
+	changesPending  int32
 	syncRunner      *async.BoundedFrequencyRunner // governs calls to syncProxyRules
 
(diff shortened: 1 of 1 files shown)


squeed:
/approve
(if that even works)

PR #26388, opened 2021-08-09
Bug 1991980: Skipped non-standard openshift installed crd testing for ibmcloud

Joseph-Goergen:
https://bugzilla.redhat.com/show_bug.cgi?id=1991980

### DIFF
--- a/test/extended/cli/explain.go
+++ b/test/extended/cli/explain.go
@@ -82,9 +82,6 @@ var (
 
 		// openshift.io groups:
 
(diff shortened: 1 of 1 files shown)


Joseph-Goergen:
https://bugzilla.redhat.com/show_bug.cgi?id=1991980

csrwng:
@ironcladlou ptal
This is to skip crd checks for resources that don't exist in a hosted cluster install

PR #26389, opened 2021-08-10
Bug 1986413: Do not use egressIP on reply packets

danwinship:
EgressIP namespaces should only force the egressIP when the pod is the client. If the pod is the server we want to reply normally.

Backport of openshift/sdn#236


danwinship:
/assign @abhat 
sorry, refiled from ose
zzhao says QE can't pre-confirm 3.11 bugs so this has to merge first and then get tested

PR #26392, opened 2021-08-11
kubelet: detect process overlap during apiserver rollout

mfojtik:
https://github.com/openshift/kubernetes/pull/876/files introduced a detection mechanism that uses an event when there are two overlapping processes running of apiserver. This should never happen and if it happens it should become a blocker bug for the release to fix it.


pkg/synthetictests/event_junits.go
```diff
@@ -41,6 +42,7 @@ func SystemUpgradeEventInvariants(events monitorapi.Intervals, duration time.Dur
 	tests = SystemEventInvariants(events, duration, kubeClientConfig)
 	tests = append(tests, testContainerFailures(events)...)
 	tests = append(tests, testDeleteGracePeriodZero(events)...)
+	tests = append(tests, testKubeApiserverProcessOverlap(events)...)
```
deads2k, line 45: this looks like a systemeventinvariant to me.  no matter what, we shouldn't have two processes running.

mfojtik: @deads2k it exists in both

pkg/synthetictests/kubelet.go
```diff
@@ -135,6 +135,26 @@ func testContainerFailures(events monitorapi.Intervals) []*ginkgo.JUnitTestCase
...
+		if strings.Contains(event.Message, "reason/TerminationProcessOverlapDetected") {
+			failures = append(failures, event.Locator)
+		}
+	}
+
+	failure := &ginkgo.JUnitTestCase{
+		Name:      testName,
+		SystemOut: strings.Join(failures, "\n"),
+		FailureOutput: &ginkgo.FailureOutput{
+			Output: fmt.Sprintf("The following events detected:\n\n%s", strings.Join(failures, "\n")),
```
deads2k, line 152: you're going to want to display the time as well.

mfojtik: fixed

mfojtik:
adding valid-bug manually as this is a test and does not require QE.

PR #26395, opened 2021-08-12
BUILD-278: search for correct memory.max file with cgroupv2 and build quota test

gabemontero:
See https://github.com/openshift/builder/pull/252#issuecomment-897226011 and https://github.com/openshift/builder/pull/252#issuecomment-896891553 for rationale as to why we have to do this with cgroupv2 and privileged pods

per recent review conventions, wrt to bash, my invocation of spellcheck returned clean:

```shell
gmontero ~ $ shellcheck ~/go/src/github.com/openshift/origin/test/extended/testdata/builds/build-quota/.s2i/bin/assemble 
gmontero ~ $ 
```


gabemontero:
ok e2e-gcp-builds and e2e-gcp passed ... those are the e2e's affected by this change 

/assign @coreydaley 
/assign @bparees 

/retest

PR #26401, opened 2021-08-17
Bug 1994491: add HPA event exception

soltysh:
/assign @ingvagabund 
/cc @atiratree 
since I'm sending that BZ for initial investigation.


soltysh:
it's optional and is being currently looked at

soltysh:
Replaced with https://github.com/openshift/origin/pull/26409

PR #26402, opened 2021-08-17
(experimental) Bug 1980141: Test upstream netPol Test Changes

astoycos:
/cc @astoycos 

This is a test to see how the recent upstream netPol test work effects our downstream CI 
https://github.com/kubernetes/kubernetes/pull/104349
https://github.com/kubernetes/kubernetes/pull/104086


astoycos:
And the build is broken :( Let's see what happened

jayunit100:
yay thanks for this @astoycos !

lsmcfadden:
@astoycos is this something in plan to complete for OCP 4.9 or considered low priority?

astoycos:
I'd say lower priority and no for 4.9 because we still have networkPolicy test coverage with the legacy tests.

 However, I think I got enough data from this to push for puling down the upstream K8's changes and reactivating the `netPol` tests in 4.10, I can also look at backporting after that

astoycos:
Will re-open this when It's ready

PR #26406, opened 2021-08-17
Add 'DeploymentAwaitingCancellation' as an allowed repeated event

bparees:
we're hitting this one a lot and i don't think it is indicative of an actual underlying problem (it results from queuing up multiple rollouts and cancelations, which the deploymentconfig tests do)

https://search.ci.openshift.org/?search=DeploymentAwaitingCancellation+Deployment+of+version+2+awaiting+cancellation+of+older+running+deployments&maxAge=12h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job


deads2k:
The PR bumped into

> event happened 21 times, something is wrong: ns/e2e-horizontal-pod-autoscaling-8238 horizontalpodautoscaler/rc-light - reason/FailedGetResourceMetric failed to get cpu utilization: unable to get metrics for resource cpu: no metrics returned from resource metrics API

by design?

PR #26410, opened 2021-08-19
Bug 1992596: e2e/cli: Move annotations.sh to proper e2e

deejross:
This moves the testing from a bash script to a Ginkgo test files.


soltysh (approved):
/approve
/assign @atiratree

test/extended/testdata/cmd/test/cmd/annotations.sh
```diff
@@ -1,28 +0,0 @@
...
-# This test validates empty values in key-value pairs set by the annotate command
-os::cmd::expect_success_and_text 'oc create -f ${TEST_DATA}/hello-openshift/hello-pod.json' 'pod/hello-openshift created'
-os::cmd::expect_success_and_text 'oc annotate pod hello-openshift node-selector=""' 'pod/hello-openshift annotated'
-os::cmd::expect_success_and_not_text 'oc get pod hello-openshift --template="{{index .metadata.annotations \"node-selector\"}}"' '.'
-
-echo "annotate: ok"
-os::test::junit::declare_suite_end
-
-os::test::junit::declare_suite_start "cmd/label"
-# This test validates empty values in key-value pairs set by the label command
```
atiratree, line 23: can we also include the tests for empty values?

deejross: Latest commit includes removing annotations and labels and validating they are empty

atiratree: I was asking specifically for the test we have here (

test/extended/cli/annotation.go
```diff
@@ -0,0 +1,48 @@
...
+		out, err = oc.Run("annotate").Args("pod", "hello-openshift", "new-anno=hello").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.ContainSubstring("pod/hello-openshift annotated"))
+
+		g.By("validating new annotation")
+		out, err = oc.Run("get").Args("pod", "hello-openshift", "--template", podAnnotationTemplate).Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.Equal("hello"))
+
+		g.By("removing the annotation")
```
atiratree, line 38: +1, having additional test doesn't hurt anything

atiratree, line 38: +1, having additional test doesn't hurt anything

yselkowitz:
The test as currently written points to the "openshift/hello-openshift" image, which is not multi-arch enabled.   This will cause these tests to fail in both disconnect installs, and on any install on non-x86 architectures.  Please rework this test according to the [procedures for e2e test images](https://github.com/openshift/origin/tree/master/test/extended/util/image).
/hold

deejross:
@yselkowitz I ran into something similar with the next PR where the old bash scripts were using images that weren't allowed. That PR #26415 includes several fixes for these types of issues, and I'm considering closing this one and moving it over there to take advantage of the newer resource builder which uses approved e2e images.

deejross:
This has been moved into #26415 to better handle multi-arch and disconnected installs.

PR #26098, opened 2021-04-21
refactor TestMultipleImageChangeBuildTriggers, add verification of new status fields

gabemontero:
So reworking this test was motivated by 2 things:
- validating the changes in [BUILD-186](https://issues.redhat.com/browse/BUILD-186) where last triggered image ID is getting set in status (and will eventually stop getting set in spec) ... though this changes is not required for that Jira to drop

- the error output from runs like https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_openshift-controller-manager/171/pull-ci-openshift-openshift-controller-manager-master-e2e-aws/1384874641511682048 were totally unhelpful:
```
I0421 15:46:24.629732   76714 test_context.go:459] Tolerating taints "node-role.kubernetes.io/master" when considering if nodes are ready
[BeforeEach] [Top Level]
  github.com/openshift/origin/test/extended/util/framework.go:1440
[BeforeEach] [Top Level]
  github.com/openshift/origin/test/extended/util/framework.go:1440
[BeforeEach] [Top Level]
  github.com/openshift/origin/test/extended/util/test.go:59
[BeforeEach] [sig-imageregistry][Feature:ImageTriggers] Image change build triggers
  github.com/openshift/origin/test/extended/util/client.go:141
STEP: Creating a kubernetes client
[BeforeEach] [sig-imageregistry][Feature:ImageTriggers] Image change build triggers
  github.com/openshift/origin/test/extended/util/client.go:116
Apr 21 15:46:24.969: INFO: configPath is now "/tmp/configfile722106540"
Apr 21 15:46:24.970: INFO: The user is now "e2e-test-image-change-build-trigger-kdcch-user"
Apr 21 15:46:24.970: INFO: Creating project "e2e-test-image-change-build-trigger-kdcch"
Apr 21 15:46:25.181: INFO: Waiting on permissions in project "e2e-test-image-change-build-trigger-kdcch" ...
Apr 21 15:46:25.198: INFO: Waiting for ServiceAccount "default" to be provisioned...
Apr 21 15:46:25.326: INFO: Waiting for ServiceAccount "deployer" to be provisioned...
Apr 21 15:46:25.448: INFO: Waiting for ServiceAccount "builder" to be provisioned...
Apr 21 15:46:25.570: INFO: Waiting for RoleBinding "system:image-pullers" to be provisioned...
Apr 21 15:46:25.602: INFO: Waiting for RoleBinding "system:image-builders" to be provisioned...
[...]

PR #26411, opened 2021-08-19
WIP - BUILD-188 image trigger tests update for LastTriggeredImageID

alicerum:
Since LastTriggeredImageID field of BuildConfig Spec is now never set,
and same field of BC Status is now used instead, we should check for
Status field, and confirm that Spec field is empty.

This is a WIP pull request, that fixes `Image Trigger` tests, there are a few other tests to fix, specifically Build Controller tests, so this still should not be merged.

Possible that this pull request should also incorporate changes by @gabemontero from #26098


test/extended/images/imagechange_buildtrigger.go
```diff
@@ -311,8 +311,10 @@ func runTest(t g.GinkgoTInterface, oc *exutil.CLI, testname string, projectAdmin
 
 		// the first tag did not have an image id, so the last trigger field is the pull spec
 		expectedLastTriggerTag := registryHostname + "/openshift/test-image-trigger:" + tag
-		lastTriggeredImageId := updatedConfig.Spec.Triggers[0].ImageChange.LastTriggeredImageID
-		o.Expect(lastTriggeredImageId).To(o.Equal(expectedLastTriggerTag))
+		lastTriggeredImageIdSpec := updatedConfig.Spec.Triggers[0].ImageChange.LastTriggeredImageID
+		lastTriggeredImageIdStatus := updatedConfig.Status.ImageChangeTriggers[0].LastTriggeredImageID
+		o.Expect(lastTriggeredImageIdStatus).To(o.Equal(expectedLastTriggerTag))
+		o.Expect(lastTriggeredImageIdSpec).To(o.BeEmpty())
```
gabemontero, line 317: I think you have a "chicken or egg" situation here @alicerum 

1) you are trying to merge your openshift-apiserver PR where you only set status
2) but that is no merging because the current version of this test attempt to verify spec
3) but this test is currently failing in this PR because you openshift-apiserver PR which ensures spec is not set has not merged

I think you will need to update this test in 2 PRs.
a) change this test first so you only verify what status is set to
b) get your openshift-apisever PR to merge with the test update from 1)
c) come back here in a new PR and add the verification that spec is empty

alicerum: Yeah, i think it is the right way to go :) Will probably go for that for now, thank you

adambkaplan:
@alicerum is this still WIP? The tests look good to me and are passing

adambkaplan (commented):
/approve

Marking this PR as approved to merge - generally looks good to me.

gabemontero:
/lgtm

yeah I think the WIP should be removed as well, but will let @alicerum do that

adambkaplan:
/retitle Bug 1986562: Only check status for image trigger tests

alicerum:
Unmarking `WIP` because it is a first part of BUILD-188 changes, required to make https://github.com/openshift/openshift-apiserver/pull/239 pass

PR #26241, opened 2021-06-17
Bug 1972684: Add failing tests in OVN-K into skipped list

ardaguclu:
e2e-metal-ipi-ovn-dualstack jobs are permanently failing.

This PR adds 3 tests that are failing in dualstack into the `[Skipped:Network/OVNKubernetes]` list. As stated in https://bugzilla.redhat.com/show_bug.cgi?id=1972684#c5, these tests are not supported by OVN-Kubernetes.

PR #26414, opened 2021-08-20
Bug 1996097: Skip renamed dualstack tests in OVNKubernetes

ardaguclu:
Previously skipped dualstack tests(https://github.com/openshift/origin/pull/26241) have been renamed after the latest bump.
This PR adds renamed tests into OVN-K skipped list for dualstack.

### DIFF
--- a/test/extended/util/annotate/rules.go
+++ b/test/extended/util/annotate/rules.go
@@ -295,11 +295,11 @@ var (
 			`\[sig-network\] Services should have session affinity timeout work for NodePort service`,
 			`\[sig-network\] Services should have session affinity timeout work for service with type clusterIP`,
 
-			// https://bugzilla.redhat.com/show_bug.cgi?id=1972684 - [Feature:IPv6DualStack] tests are failing in dualstack
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1996097 - [Feature:IPv6DualStack] tests are failing in dualstack
 			// https://jira.coreos.com/browse/SDN-510: OVN-K doesn't support session affinity
-			`\[sig-network\] \[Feature:IPv6DualStack\] \[LinuxOnly\] Granular Checks: Services Secondary IP Family should function for client IP based session affinity: http`,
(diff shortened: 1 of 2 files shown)


test/extended/util/annotate/rules.go
```diff
@@ -300,6 +300,9 @@ var (
 			`\[sig-network\] \[Feature:IPv6DualStack\] \[LinuxOnly\] Granular Checks: Services Secondary IP Family should function for client IP based session affinity: http`,
 			`\[sig-network\] \[Feature:IPv6DualStack\] \[LinuxOnly\] Granular Checks: Services Secondary IP Family should function for client IP based session affinity: udp`,
 			`\[sig-network\] \[Feature:IPv6DualStack\] \[LinuxOnly\] should have ipv4 and ipv6 node podCIDRs`,
+			`\[sig-network\] \[Feature:IPv6DualStack\] Granular Checks: Services Secondary IP Family \[LinuxOnly\] should function for client IP based session affinity: http`,
```
deads2k, line 303: the bug referenced for this skip set is verified and the jira referenced for this block has its PR merged.  Why are we still skipping these?

ardaguclu: Thanks @deads2k. I think, referenced PR in jira has been merged in upstream https://github.com/ovn-org/ovn-kubernetes/pull/2294 but not in downstream https://github.com/openshift/ovn-kubernetes/pull/666.

Since service affinity is not supported, bug was verified after skipping these three tests for OVN-K.

ardaguclu:
@danwinship could you please take a look at?. Thanks.

test/extended/util/annotate/rules.go
trozet: service affinity should work once 666 merges. The only thing that isnt supported is:
https://github.com/ovn-org/ovn-kubernetes/blob/master/test/scripts/e2e-kind.sh#L30

dcbw:
/lgtm
(test has nothing to do with dual-stack since it's an SDN test, and dual-stack is OVN-only)

PR #26417, opened 2021-08-20
Whereabouts e2e test

nicklesimba:
Created an e2e test to make 4 pods with whereabouts IPAM that only allows 3. The expected result is that the first 3 come up successfully and the 4th pod doesn't.


test/extended/networking/whereabouts.go
```diff
@@ -0,0 +1,95 @@
...
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
+)
+
+var _ = g.Describe("[sig-network][Feature:Whereabouts]", func() {
+	var oc *exutil.CLI
+	var ns string // namespace
```
dougbtv, line 20: Can this be scoped more locally to the `g.It()` block?

test/extended/networking/whereabouts.go
```diff
@@ -0,0 +1,95 @@
...
+			pod.ObjectMeta.Annotations = annotation
+		})
+		testPod2 := frameworkpod.CreateExecPodOrFail(f.ClientSet, ns, podName, func(pod *v1.Pod) {
+			pod.ObjectMeta.Annotations = annotation
+		})
+		testPod3 := frameworkpod.CreateExecPodOrFail(f.ClientSet, ns, podName, func(pod *v1.Pod) {
+			pod.ObjectMeta.Annotations = annotation
+		})
+		// Fourth pod should not come up.
+		fmt.Println("...creating 4th exec pod, expecting failure due to IPAM")
```
dougbtv, line 54: Really fond of this, to test the case where addresses are exhausted :+1:

test/extended/networking/whereabouts.go
```diff
@@ -0,0 +1,95 @@
...
+		output, err = oc.AsAdmin().Run("exec").Args(testPod2.Name, "--", "ip", "a").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		fmt.Println(output)
+		o.Expect(output).Should(o.ContainSubstring("192.168.2.229"))
+		// pod 3
+		output, err = oc.AsAdmin().Run("exec").Args(testPod3.Name, "--", "ip", "a").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		fmt.Println(output)
+		o.Expect(output).Should(o.ContainSubstring("192.168.2.230"))
+
```
dougbtv, line 93: Something to consider would be adding a check of the "results object" -- that is, the resulting `network-status` annotation on the pod, to see that the proper IP addresses were returned to Multus, and that they show up the status annotation properly.

This would be kind of buttering both sides of the bread, but, might be a nice addition.

dougbtv (changes requested):

PR #26420, opened 2021-08-21
Add imagepull backoff to frequent event allowlist

bparees:
Ran into this here:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.9-e2e-aws-canary/1428864201471823872

did not appear to cause any test failures (not sure if that job runs any actual olm tests.  If it doesn't, that could explain it, or the pulls eventually succeeded before the test ran).

### DIFF
--- a/pkg/synthetictests/duplicated_events.go
+++ b/pkg/synthetictests/duplicated_events.go
@@ -56,9 +56,6 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
 	// The pod is shaped to intentionally not be scheduled.  Looks like an artifact of the old integration testing.
 	regexp.MustCompile(`ns/e2e-test-scc-[a-z0-9]+ pod/.* - reason/FailedScheduling.*`),
 
-	// this image is used specifically to be one that cannot be pulled in our tests
-	regexp.MustCompile(`.*reason/BackOff Back-off pulling image "webserver:404"`),
-
 	// Security Context ** should not run with an explicit root user ID
 	// Security Context ** should not run without a specified user ID
 	// This container should never run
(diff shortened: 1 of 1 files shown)


bparees:
/hold

i'm slightly concerned so many instances of this are imagepull backoffs for the OLM catalog images.  I'm not sure what to make of that (and depending what it reveals, it might be indicative this entry is worth keeping around, though it also seems to generate at least a handful of false positives.

@ecordell @kevinrizza curious what you make of this:

https://search.ci.openshift.org/?search=reason%2FBackOff+Back-off+pulling+image&maxAge=48h&context=1&type=junit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

pkg/synthetictests/duplicated_events.go
```diff
@@ -69,7 +66,11 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
 	regexp.MustCompile(`ns/e2e-persistent-local-volumes-test-[0-9]+ pod/pod-[a-z0-9.-]+ reason/FailedScheduling`),
 
 	// various DeploymentConfig tests trigger this by canceling multiple rollouts
-	regexp.MustCompile(`.*reason/DeploymentAwaitingCancellation Deployment of version [0-9].+ awaiting cancellation of older running deployments`),
+	regexp.MustCompile(`reason/DeploymentAwaitingCancellation Deployment of version [0-9].+ awaiting cancellation of older running deployments`),
+
+	// Can be caused by registry hiccups and not explicitly tied to any particular component failure
+	// (if image pulls fail catastrophically we'd expect them to lead to test failures)
```
deads2k, line 72: I think we want to catch the non-catastrophic ones too.  Consistent lack of availability seems like a problem we want to find.

I'm more flexible for e2e namespaces if you restrict it to ns/e2e-something *** Back-off pulling image

bparees: So the consistent OLM pull failures seem to have gone away (must have been a registry/image issue).

but we're still at 0.13% of all failures being caused by imagepull backoff event repetition:
https://search.ci.openshift.org/?search=reason%2FBackOff+Back-off+pulling+image&maxAge=48h&context=1&type=junit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

I guess i'm trying to figure out how we draw the line between "this might be useful/good to know about" and "this is causing CI jobs to fail that otherwise would have succeeded and no one is going to investigate this"

whitelisting "e2e-*" as a namespace would probably cover most of them, not clear if you're amenable to that or not.

PR #26421, opened 2021-08-23
Fix regex pattern for `DeploymentAwaitingCancellation` event

andfasano:
Fix regex expression for the allowed repeated pattern for `DeploymentAwaitingCancellation` event


deads2k:
/lgtm

I hate regexes

PR #26422, opened 2021-08-23
Bug 1995198: Skip admin can subscribe to operator using olm test on arm64

anik120:


test/extended/operators/olm.go
```diff
@@ -208,6 +209,11 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
 			g.Skip("redhat-operators source not found in enabled sources")
 		}
 
+		// TODO: Come up with an arch agnostic operator to deploy during this test.
+		// For now this does not work for arm64
+		if e2e.NodeOSDistroIs("arm64") {
```
Prashanth684, line 214: is `NodeOSDistroIs` used for identifying the OS type rather than arch? is there an equivalent for architecture?

timflannagan: It looks like `NodeOSArchIs` may be more applicable here:

```go
// NodeOSArchIs returns true if the node OS arch is included in the supportedNodeOsArchs. Otherwise false.
func NodeOSArchIs(supportedNodeOsArchs ...string) bool {
	for _, arch := range supportedNodeOsArchs {
		if strings.EqualFold(arch, TestContext.NodeOSArch) {
			return true
		}
	}
	return false
}
```

There also looks like there's a `SkipUnlessNodeOSArchIs(...)` that will skip tests unless the list of arch(es) provided as arguments match the test context.

anik120: Oops. I meant to use `NodeOSArchIs`. Thanks for catching it @Prashanth684!

anik120: @timflannagan yea unfortunately doesn't look like there's a `SkipIfNodeOSArchIs` :( and hard coding the list of archs we want the tests to run on doesn't seem like a good idea. So just used `NodeOSArchIs`.

timflannagan:
This can be a stopgap until https://github.com/operator-framework/operator-marketplace/pull/420 merges sometimes around 4.9 GA.

/lgtm

timflannagan:
These changes won't affect core APIs becoming unreachable during the upgrade tests.

PR #26430, opened 2021-08-26
Bug 1998029: CustomNoUpgrade should not fire no-upgrade failures

pierreprinetti:
CustomNoUpgrade should be treated just as TechPreviewNoUpgrade, and
should not be a condition that triggers no-upgrade failures in tests.
This is useful in cases where we want to test a specific set of feature
gates in isolation from other tech preview feature gates.

Specifically we are using CustomNoUpgrade to test
external-cloud-provider CCM promotions in isolation in
https://github.com/openshift/release/blob/992f297ef8a1ba749f3a0b9b40438696f18a1bc4/ci-operator/step-registry/ccm/conf/pre-create-feature-gate/ccm-conf-pre-create-feature-gate-commands.sh#L22

Co-Authored-By: Matthew Booth <mbooth@redhat.com>

### DIFF
--- a/test/extended/operators/operators.go
+++ b/test/extended/operators/operators.go
@@ -63,7 +63,7 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should", func() {
 
 		fg := objx.Map(fgObj.UnstructuredContent())
 		featureSet := fg.Get("spec.featureSet").String()
-		isTechPreview := featureSet == "TechPreviewNoUpgrade"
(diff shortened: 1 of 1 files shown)


pierreprinetti:
/assign @kevinrizza 
as @benluddy suggested (thanks)

pierreprinetti:
Apparently I was chasing the wrong team.

Let's try again!

/assign bparees

pierreprinetti:
So apparently, the right person here might be @deads2k instead.

/assign @deads2k

PR #26432, opened 2021-08-30
Bug 1986562: check buildConfig spec LastTriggeredImageID is empty

alicerum:
After the #26411 regarding this bug, we are already checking that correct value is in `LastTriggeredImageID` field of `BuildConfig`'s status.
Now, since `openshift-apiserver` changes have been merged, we can also check that same spec field is actually empty, and not filled.


adambkaplan (approved):
/lgtm

PR #26439, opened 2021-09-01
ETCD-234: pkg/synthetictests: add etcd quorum-gaurd duplicate events to known problems

hexfusion:
Bug and Jira were created to track future optimizations.

cc @lilic @bparees 


bparees:
thanks @hexfusion!

bparees:
slapping valid-bz on this so we can potentially benefit from it in 4.9 CI too rather than waiting for 4.10 to open.

PR #26441, opened 2021-09-01
Openstack tests

pierreprinetti:
This extension to openshift/conformance/parallel will test
OpenStack-specific invariants by interrogating the OpenStack API.

This patch adds one test, asserting that the control plane servers all
sit in the same server group.

In a separate commit:

* Add the required dependencies to interrogate the OpenStack API.
* Bump gophercloud/gophercloud to the most recent version (v0.9.0) supported by the in-tree cloud provider.

Implements [OSASINFRA-2618](https://issues.redhat.com/browse/OSASINFRA-2618)


pierreprinetti:
I have rewritten the two commits, as part of the vendoring was happening in the wrong one. The patch in its entirety is unchanged.

mandre (commented):
/lgtm

pierreprinetti:
/retest
/test e2e-openstack

Note that this does not require a BZ and can wait until the master branch is open again.

pierreprinetti:
/hold
As suggested by @itzikb-redhat, the test should be skipped unless running against OCP v4.8+.

In general, this test framework is designed for feature testing. Since each feature comes in a specific version, we can assume that all tests will require to set a minimum version.

This PR should provide a generic mechanism for defining the minimum version against which the test should not be skipped.

pierreprinetti:
/hold cancel
The test is now skipped when the target cluster does not support the feature.

pierreprinetti:
Rebased so that hopefully this PR does not require the vsphere job any more.

test/extended/openstack/OWNERS
```diff
@@ -0,0 +1,16 @@
+approvers:
+- EmilienM
+- adduarte
+- iamemilio
+- mandre
+- mdbooth
+- pierreprinetti
+- stephenfin
```
sttts, line 8: that are a lot of approvers.

pierreprinetti: This is the team in charge of the component (OpenShift on OpenStack). What is the concern?

sttts: We loosely follow the definition from upstream: https://github.com/kubernetes/community/blob/master/community-membership.md. Seeing as many reviewers as approvers is unusual.

mandre: @sttts I don't disagree with you it's important to pay attention keeping the list of approvers up-to-date with active developers in upstream kubernetes, however I think we need to consider what this OWNERS file gives access to. We're talking about unit tests, and an internal team that maintains them.

sttts: Breaking these tests will break CI for the org and make hundred people unable to merge anything, at least as soon as you move to having an always running e2e job (which I expect you want if your want into origin). So there is quite some responsibility.

sttts: There is no clear rule about this. I think there is risk and we should take e2e tests serious, especially when they run from origin and break the whole org if something is wrong. My opinion is that 7 (whole team) is wrong, versus a healthy  2 or 3. IMO: start small and add people when they gain experience and have proven to deliver reliable and non-flaky code.

I took the chance and checked the other packages:
```
cmd: reviewers 3, approvers 3
cluster: reviewers 2, approvers 1
router: reviewers 6, approvers 6
cluster: reviewers 2, approvers 1
dr: reviewers 4, approvers 4
operators: reviewers 11, approvers 4
annotate: reviewers 1, approvers 2
image: reviewers 3, approvers 3
image_ecosystem: reviewers 3, approvers 3
images: reviewers 2, approvers 2
networking: reviewers 4, approvers 9
baremetal: reviewers 0, approvers 3
builds: reviewers 3, approvers 3
imageapis: reviewers 2, approvers 2
tbr_health: reviewers 3, approvers 3
etcd: reviewers 4, approvers 4
extended: reviewers 8, approvers 8
prometheus: reviewers 5, approvers 5
rbac: reviewers 4, approvers 0
router: reviewers 6, approvers 6
```

I don't want to block this PR. Delegating to @mfojtik.

mandre: Fair enough, I get your point and I see that indeed, OWNERS files in origin/tests have a low number of approvers.
It should be fine to keep the whole team as reviewers but we should make the list of approvers shorter. I think we need at least @pierreprinetti and myself in it.

pierreprinetti: With the possible exception of myself, each of the 7 approvers I propose in this PR are talented engineers who bring a demonstrable track record of valid contributions in open source.

I take testing very seriously and I believe that each of these individuals is worth in this repository the same level of trust they have earned over the years in many others across the OpenShift and OpenStack organisations.

sttts: @pierreprinetti being approver or not has nothing to do with talent or valid contributions in open source. Also nobody want to keep away approver rights from teams. Just looking at the contribution count of the mentioned github handles I see only some with contributions to this repo.

In general, I am not against full ownerships (everybody approver) in team owned repos. But this is a shared code base, with shared maintenance, shared CI pain and potentially impact on a whole org of a hundred developers. Hence, I feel like it is healthy to follow "earned authority" rather than "authority by role".

mdbooth (changes requested):

pierreprinetti:
/close
As agreed with @sttts we are going forward in a separate repo.

Thank you @mdbooth for the review, that will be helpful because the code there will likely be very similar to this.

test/extended/openstack/servergroup.go
```diff
@@ -0,0 +1,76 @@
...
+	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = g.Describe("[sig-installer][Feature:openstack] OpenStack platform should", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLI("openstack")
```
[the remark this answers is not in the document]
pierreprinetti: As fancy as it may sound, `exutil.NewCLI` generates BeforeAll and AfterAll functions, and must therefore be called in the Describe context.

test/extended/openstack/servergroup.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+		}
+
+		g.By("getting information on the OpenStack Server groups")
+		var groups []servergroups.ServerGroup
+		{
+			computeClient, err := client(serviceCompute)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			allPages, err := servergroups.List(computeClient).AllPages()
+			o.Expect(err).NotTo(o.HaveOccurred())
```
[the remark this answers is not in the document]
pierreprinetti: the server_group property [has been added to the "get server" API call in microversion 1.71](https://docs.openstack.org/api-ref/compute/?expanded=show-server-details-detail#show-server-details); this is unfortunately beyond what we have in some of our CI clouds.

pierreprinetti:
For the record: github.com/openshift/openstack-test

PR #26442, opened 2021-09-02
Bug 1997509: delete only necessary resources instead of all

atiratree:


test/extended/builds/pipeline_origin_bld.go
```diff
@@ -255,14 +255,6 @@ var _ = g.Describe("[sig-builds][Feature:Builds][sig-devex][Feature:Jenkins][Slo
 					g.By("get build console logs and see if succeeded")
 					_, err = j.GetJobConsoleLogsAndMatchViaBuildResult(brservices, "Finished: SUCCESS")
 					o.Expect(err).NotTo(o.HaveOccurred())
-					g.By("clean up openshift resources for next potential run")
```
soltysh, line 258: Why removing these?

soltysh (commented):
One question, otherwise this looks good
/approve

test/extended/builds/pipeline_origin_bld.go
atiratree: @soltysh The removal of e2e namespace should cleanup these by default. I removed them for better readability/less clutter but I can revert the change if you prefer.

soltysh: You're right, I just noticed `defer` at the beginning of the test.

soltysh (approved):
/lgtm

PR #26444, opened 2021-09-02
Bug 2000589: Revert "Re-enable crictl node test"

saschagrunert:
Reverts openshift/origin#26320 since the test does not get excluded in the periodics.


mrunalp (approved):
/lgtm

PR #26445, opened 2021-09-02
Split the "Cluster frontend ingress" test into 4

Miciah:
Before this change, the "Cluster frontend ingress" test used four separate, concurrently running clients to monitor availability of the console and OAuth routes during upgrade.  The monitoring recorded any periods of observed disruption and then reported the sum total amount of time during which it had observed disruption.  Because the four clients ran concurrently, it was possible for the period of disruption observed by one client to overlap with periods of disruption observed by other clients, with the result that summing up these periods could lead to quadruple accounting of the same period during which disruption was observed by multiple clients.

This change replaces the single "Cluster frontend ingress remain available" test with four tests:

* OAuth remains available via cluster frontend ingress using new connections

* OAuth remains available via cluster frontend ingress using reused connections

* Console remains available via cluster frontend ingress using new connections

* Console remains available via cluster frontend ingress using reused connections

These tests independently observe disruption, and each reports the sum total disruption that the respective test observes so that there is no double-, triple-, or quadruple accounting.

* `test/e2e/upgrade/upgrade.go` (`AllTests`):
* `test/extended/dr/machine_recover.go`: Replace the old single cluster frontend ingress availability test with the four new tests.
* `test/extended/util/disruption/frontends/frontends.go` (`AvailableTest`): Replace this...
(`availableTest`): ...with this.  Replace the slice of frontends with a single frontend as now each `availableTest` instance tests exactly one frontend.  Add `test` and `testName` fields, as well as a `startMonitoring` field to specify the function that starts the monitoring.  Instead of using `AvailableTest`, callers now must use one of the new functions to get a specific test.
(`NewOAuthAvailableWithConnectionReuseTest`, `NewOAuthRouteAvailableWithNewConnectionsTest`, `NewConsoleRouteAvailableWithNewConnectionsTest`, `NewConsoleRouteAvailableWithConnectionReuseTest`): New functions to get an
`availableTest` instance for a specific cluster frontend ingress availability test, using the new `oauthRouteFrontend` and `consoleRouteFrontend` variables and `startEndpointMonitoringWithNewConnections` and `startEndpointMonitoringWithConnectionReuse` helpers.
(`oauthRouteFrontend`, `consoleRouteFrontend`): New variables to describe the frontend routes that the cluster frontend ingress tests use.
(`Name`, `DisplayName`): Return the name from the `availableTest` receiver.
(`Setup`): Set up only the single frontend of the `availableTest`.
(`Test`): Use the `startMonitoring` function from `availableTest`.
(`startEndpointMonitoring`): Add a `disableConnectionReuse` parameter.  Instead of setting up two clients (one that uses a new connection for every request and one that reuses connections), set up only one client configured per the new parameter's value.
(`startEndpointMonitoringWithNewConnections`, `startEndpointMonitoringWithConnectionReuse`): New functions.


deads2k:
leaving the detailed review to Stephen, but this looks like it's moving in the right direction.

/assign @stbenjam 

/approve

PR #26446, opened 2021-09-02
Bug 2009879: Add admin ack Upgradeable condition gate test

jottofar:
Since only later 4.8.z releases will contain the admin ack functionality, the test first checks for presence of the admin-gates configmap. If not found, indicating that  the admin ack functionality is not part the baseline being tested, the test simply returns successfully.
Otherwise, test first verifies that Upgradeable condition is false for correct reason and with correct message. It then modifies the admin-acks configmap to ack the necessary admin ack gate and then waits for the Upgradeable condition to change to true.

### DIFF
--- a/test/e2e/upgrade/adminack/adminack.go
+++ b/test/e2e/upgrade/adminack/adminack.go
@@ -1,0 +1,52 @@
+package adminack
+
+import (
+	"context"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/upgrades"
+)
+
+// UpgradeTest contains artifacts used during test
+type UpgradeTest struct {
+	oc     *exutil.CLI
+	config *restclient.Config
+}
+
+func (UpgradeTest) Name() string { return "check-for-admin-acks" }
+func (UpgradeTest) DisplayName() string {
+	return "[bz-Cluster Version Operator] Verify presence of admin ack gate blocks upgrade until acknowledged"
+}
+
+// Setup creates artifacts to be used by Test
+func (t *UpgradeTest) Setup(f *framework.Framework) {
+	g.By("Setting up admin ack test")
+	oc := exutil.NewCLIWithFramework(f)
+	t.oc = oc
+	config, err := framework.LoadConfig()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	t.config = config
+	framework.Logf("Admin ack test setup complete")
+}
+
+// Test simply returns successfully if admin ack functionality is not part the baseline being tested. Otherwise,
+// test first verifies that Upgradeable condition is false for correct reason and with correct message. It then
+// modifies the admin-acks configmap to ack the necessary admin-ack gate and then waits for the Upgradeable
+// condition to change to true.
+func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade upgrades.UpgradeType) {
+	ctx := context.Background()
+
+	adminAckTest := &exutil.AdminAckTest{Oc: t.oc, Config: t.config}
+	adminAckTest.Test(ctx)
+}
+
+// Teardown cleans up any remaining objects.
+func (t *UpgradeTest) Teardown(f *framework.Framework) {
+	// rely on the namespace deletion to clean up everything
+}
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -30,6 +30,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+	"github.com/openshift/origin/test/e2e/upgrade/adminack"
 	"github.com/openshift/origin/test/e2e/upgrade/alert"
 	"github.com/openshift/origin/test/e2e/upgrade/service"
 	"github.com/openshift/origin/test/extended/util/disruption"
@@ -47,6 +48,7 @@ func NoTests() []upgrades.Test {
 // AllTests includes all tests (minimal + disruption)
 func AllTests() []upgrades.Test {
 	return []upgrades.Test{
+		&adminack.UpgradeTest{},
 		controlplane.NewKubeAvailableWithNewConnectionsTest(),
 		controlplane.NewOpenShiftAvailableNewConnectionsTest(),
 		controlplane.NewOAuthAvailableNewConnectionsTest(),
--- a/test/extended/adminack/adminack.go
+++ b/test/extended/adminack/adminack.go
@@ -1,0 +1,28 @@
+package adminack
+
+import (
+	"context"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = g.Describe("[sig-cluster-lifecycle]", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("cli-deployment")
+
+	g.Describe("TestAdminAck", func() {
+		g.It("should succeed", func() {
+			config, err := framework.LoadConfig()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			ctx := context.Background()
+
+			adminAckTest := &exutil.AdminAckTest{Oc: oc, Config: config}
+			adminAckTest.Test(ctx)
+		})
+	})
+})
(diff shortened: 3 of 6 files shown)


test/e2e/upgrade/adminack/adminack.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			framework.Failf(fmt.Sprintf("Error accessing configmap openshift-config-managed/admin-gates, err=%v", err))
+		} else {
+			framework.Logf("Skipping admin ack test. Admin ack is not in this baseline.")
+			return
+		}
+	}
+	var msg string
+	if msg = gateCm.Data["ack-4.8-kube-122-api-removals-in-4.9"]; msg == "" {
```
jianlinliu, line 61: Sound like the cm name is changed to "ack-4.8-kube-1.22-api-removals-in-4.9" mentioned in https://github.com/openshift/cluster-version-operator/pull/647/files

jottofar: Yes, that was a late breaking change. Thanks for catching.

jottofar: fixed

test/e2e/upgrade/adminack/adminack.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	var msg string
+	if msg = gateCm.Data["ack-4.8-kube-1.22-api-removals-in-4.9"]; msg == "" {
+		framework.Failf("Configmap openshift-config-managed/admin-gates does not contain ack-4.8-kube-1.22-api-removals-in-4.9 gate")
+	}
+	c, err := configv1client.NewForConfig(t.config)
+	if err != nil {
+		framework.Failf(fmt.Sprintf("NewForConfig client request failed, err=%v", err))
+	}
+	cv, err := c.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
```
wking, line 68: nit: to reduce the number of `context.TODO()` and prepare for some hypothetical future where `Test` takes a `context.Context` argument, I prefer opening the function like:

```go
func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade upgrades.UpgradeType) {
  ctx := context.Background()
```

And then using `ctx` throughout.  That way porting to passed-in contexts will be:

```go
func (t *UpgradeTest) Test(ctx context.Context, f *framework.Framework, upgrade upgrades.UpgradeType) {
```

without needing as many invasive touches throughout the function.

And you can use `Background` instead of `TODO`.  From [the `Background` docs][1]:

> It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

[1]: https://pkg.go.dev/context#Background

test/e2e/upgrade/adminack/adminack.go
```diff
@@ -0,0 +1,134 @@
...
+		if !apierrors.IsNotFound(err) {
+			framework.Failf(fmt.Sprintf("Error accessing configmap openshift-config-managed/admin-gates, err=%v", err))
+		} else {
+			framework.Logf("Skipping admin ack test. Admin ack is not in this baseline.")
+			return
+		}
+	}
+	var msg string
+	if msg = gateCm.Data["ack-4.8-kube-1.22-api-removals-in-4.9"]; msg == "" {
+		framework.Failf("Configmap openshift-config-managed/admin-gates does not contain ack-4.8-kube-1.22-api-removals-in-4.9 gate")
```
wking, line 62: we want this to be a log-and-skip, right?  Because otherwise it seems like it would fail when run on release outside of 4.8.

jottofar:
/retitle Bug 1999092: Add admin ack Upgradeable condition gate test

test/e2e/upgrade/adminack/adminack.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		framework.Failf(fmt.Sprintf("NewForConfig client request failed, err=%v", err))
+	}
+	cv, err := c.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
+	if err != nil {
+		framework.Failf(fmt.Sprintf("Error accessing clusterversion, err=%v", err))
+	}
+	// Verify Upgradeable is false and set correctly
+	cond := getUpgradeableStatusCondition(cv.Status.Conditions)
+	checkUpgradeableStatusCondition(cond, msg)
```
wking, line 74: openshift/release#22368 will break your assumption that we are going into the update test with the ack unset.  Perhaps the test can be reshuffled like:

1. Check the current ack value, and confirm that the cluster is behaving as expected for the current ack setting.
2. If the ack is unset, set it, and confirm that the cluster is goes `Upgradeable=True`.

We'd exercise the vanilla 4.8 case during our 4.7 -> 4.8 CI, because for 4.7 and earlier installs, openshift/release#22368 would be a no-op.

test/e2e/upgrade/adminack/adminack.go
jottofar: I was thinking that this test would only be ran on 4.8 releases. If so and if we've gotten this far then we had found the gate configmap and it for 4.8 it the key should always have a value.

test/e2e/upgrade/upgrade.go
```diff
@@ -47,6 +48,7 @@ func NoTests() []upgrades.Test {
 // AllTests includes all tests (minimal + disruption)
 func AllTests() []upgrades.Test {
 	return []upgrades.Test{
+		&adminack.UpgradeTest{},
```
wking, line 51: I don't know if the order here is important, but I'd like this test to run at the end of the suite (I'd have expected to have that happen via a `[Late]` annotation).  But the `alert.UpgradeTest{}` precedent makes it seem like order does matter here.  I'm not entirely sure how these are positioned relative to the update itself, but the logging makes them seem like they're currently running early in the job:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/26446/pull-ci-openshift-origin-release-4.8-e2e-aws-upgrade/1438885249361645568/artifacts/e2e-aws-upgrade/openshift-e2e-test/artifacts/e2e.log | grep -n 'Cluster should remain functional during upgrade\|admin ack' | grep -B100 passed
84:started: (0/5/5) "[sig-arch][Feature:ClusterUpgrade] Cluster should remain functional during upgrade [Disruptive] [Serial]"
95:[It] Cluster should remain functional during upgrade [Disruptive] [Serial]
167:STEP: Setting up admin ack test
170:Sep 17 16:10:42.260: INFO: Skipping admin ack test. Admin ack is not in this baseline.
1751:passed: (2h47m1s) 2021-09-17T18:57:41 "[sig-arch][Feature:ClusterUpgrade] Cluster should remain functional during upgrade [Disruptive] [Serial]"
```

(with line 170 presumably being written quite a while before line 1751).  Running later (post update) also protects 4.7 -> 4.8 jobs from having `Upgradeable=False` alerting go off, because the openshift/release#22368 approach will not protect those jobs (and some jobs run slow parallel-conformance suites after completing the update).

test/e2e/upgrade/adminack/adminack.go
wking: I think we'll need this in 4.8 and later too, at least as long as we want thing like 4.7 -> 4.8 -> 4.9... chained updates and such to not alert on `Upgradeable=False`.  I'm not entirely sure which `origin` branch feeds the test suite for those chained updates, but I'd expect the update that takes you into 4.8 for the first time (which may be running a 4.9 or later `origin` test suite) to be setting the ack once it arrives on 4.8, and before the cluster has gotten too mad about the lack-of-ack.

jottofar: But independent of a specific release, I wouldn't think we ever want to allow a gate to be defined without a description.

test/e2e/upgrade/adminack/adminack.go
wking: This still needs to get addressed, because now that openshift/release#22368 has landed, we're failing with][1]:

```
disruption_tests: [bz-Cluster Version Operator] Verify presence of admin ack gate blocks upgrade until acknowledged | 1s
-- | --
Sep 30 18:51:09.381: Upgradeable status condition not found  github.com/openshift/origin/test/e2e/upgrade/adminack.(*UpgradeTest).Test(0xc000a102f0, 0xc00115b8c0, 0xc000e5a1e0, 0x2) 	github.com/openshift/origin/test/e2e/upgrade/adminack/adminack.go:74 +0x37d 
...
```

Instead, we should be noticing that the ack is already set, and then celebrating the lack of an `Upgradeable` condition (semantically equivalent to an explicit `Upgradeable=True`).

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26446/pull-ci-openshift-origin-release-4.8-e2e-aws-upgrade/1443635824443789312

wking:
/retitle Bug 2009879: Add admin ack Upgradeable condition gate test

No target on bug 2009879 yet, while we work out [whether this PR will be 4.8-only or not][1], but pointing away from the old bug series so the Bugzilla bot doesn't attempt to re-attach this PR.

[1]: https://github.com/openshift/origin/pull/26446#discussion_r718907254

wking:
Replaced by #26635, backporting the generic logic that landed in `master` and 4.9.

/close

PR #26451, opened 2021-09-05
Bug 1999561: Extend API rollout resiliency test to validate pods restart during the rollout

eranco74:
Extend the test to validate that pods aren't getting restarted during the API outage.

### DIFF
--- a/test/extended/apiserver/resiliency.go
+++ b/test/extended/apiserver/resiliency.go
@@ -3,8 +3,12 @@ package apiserver
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/origin/test/extended/operators"
 	"io/ioutil"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/onsi/ginkgo"
@@ -22,6 +26,7 @@ import (
 	"github.com/openshift/origin/test/extended/scheme"
 	"github.com/openshift/origin/test/extended/single_node"
 	exutil "github.com/openshift/origin/test/extended/util"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
 )
 
 var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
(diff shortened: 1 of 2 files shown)


test/extended/apiserver/resiliency.go
```diff
@@ -52,6 +57,13 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
 		framework.ExpectNoError(err)
 		gomega.Expect(clusterApiServer.Status.NodeStatuses[0].TargetRevision).To(gomega.Equal(int32(0)))
 
+		ginkgo.By("Checking pods restart count")
+		restartingContainers := make(map[operators.ContainerName]int)
+		c, err := e2e.LoadClientset()
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		// This will just load the restartingContainers map with the current restart count
+		_ = GetRestartedPods(c, restartingContainers)
```
sttts, line 65: Why this code block? It needs a comment. It's not clear to me why we should care about pod restart counts.

eranco74: Updated the comment

test/extended/apiserver/resiliency.go
```diff
@@ -52,6 +57,14 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
 		framework.ExpectNoError(err)
 		gomega.Expect(clusterApiServer.Status.NodeStatuses[0].TargetRevision).To(gomega.Equal(int32(0)))
 
+		ginkgo.By("Checking pods restart count")
```
sttts, line 60: this does not fit to the test string "allow a fast rollout". People will assume the API is at fault while the workloads or kubelet are.

eranco74: The test first validates that the disruption caused by kube-apiserver rollout is lower than allowed downtime.
It's true that I'm using this test to validate that the pods (in the openshift-* and kube-* namespaces) don't get restarted during the kube-apiserver rollout.
  
Rephrased this line to say: `ginkgo.By("Initializing pods restart count")` 
Added: `ginkgo.It("with no pods restarts during API disruption", func() {` before validating the pods restarts, so that in case it fails due to the pods restarts it would show up as:
`[Conformance][sig-sno][Serial] Cluster should allow a fast rollout of kube-apiserver with no pods restarts during API disruption`

eranco74:
Making sure the test still pass
/test e2e-aws-single-node

PR #26454, opened 2021-09-06
Bug 1961204: upgrade: skip early tests which verify events during install

vrutkovs:
Upgrade tests should ignore possible install errors during install. An example would be "should not have pod creation failures during install".

See https://search.ci.openshift.org/?search=should+not+have+pod+creation+failures+during+install&maxAge=48h&context=1&type=junit&name=.*upgrade.*&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

Upgrade tests:
* [test upgrade 4.9 openshift/origin#26454 aws](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1434921954061586432)
* [test upgrade 4.9 openshift/origin#26454 gcp](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1434921967084900352)

### DIFF
--- a/cmd/openshift-tests/upgrade.go
+++ b/cmd/openshift-tests/upgrade.go
@@ -16,6 +16,13 @@ import (
 	"k8s.io/kubernetes/test/e2e/upgrades"
 )
 
+func isStandardEarlyUpgradeTest(name string) bool {
+	if isStandardEarlyTest(name) && !strings.Contains(name, "during install") {
+		return true
+	}
+	return false
+}
+
 // upgradeSuites are all known upgade test suites this binary should run
 var upgradeSuites = testSuites{
 	{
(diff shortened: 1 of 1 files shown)


deads2k:
/hold

Given the effort to run upgrade jobs repeatedly and given that we run regular e2e-parallel tests after upgrades and given that such install failures are likely bugs that would be valid for backport to prior, I don't think that trying to hide these will be helpful.

Can you provide an example of an early test failure which would not get caught in the e2e-parallel that runs right after the upgrade and which is not a bug that is valid for backport to the prior release?

vrutkovs:
>Can you provide an example of an early test failure which would not get caught in the e2e-parallel that runs right after the upgrade and which is not a bug that is valid for backport to the prior release?

Several useful early tests are being run on upgrade:
* `[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should have same number of Machines and Nodes`
* `[sig-arch][Early] Managed cluster should start all core operators`
* `Prometheus when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured`

All of these verify that installed version is ready to be updated. But `should not have pod creation failures during install` is different - it would fail on fully-functional cluster if it previously had some pod creation events. 

Moreover, during minor version upgrade, the job success rate depend on backport status to previous release. This muddies the waters even more - it looks as if upgrade jobs are still failing and more fixes are required until we proceed with backporting fixes to previous minor.

PR #26458, opened 2021-09-09
Bug 2002552: Retry kubeconfig checks, when kube-apiserver is temporarily unavailable

soltysh:
Backporting #26377 & #26407
/assign @mfojtik


mfojtik:
approving this because it has potential to improve the metal-ipi nightlies in 4.8

PR #26459, opened 2021-09-10
Bug 1960674: images: port image signature workflow test to OCP4/UBI8

yselkowitz:
This fixes portability issues with this test, given that origin-control-plane:3.11 (besides being obsolete) is only available for x86_64.  Also, copying the oc binary from the machine running the test into a container on the cluster would cause the test to fail when run on a different architecture, or when oc was built for a newer glibc.  There was also no guarantee that the oc found on the test machine matched that of the payload being tested.

Instead, this uses the payload cli container, which inherits UBI 8.  The rest of the changes are a consequence of adapting to the newer versions of gpg2 and skopeo.

### DIFF
--- a/test/extended/images/signatures.go
+++ b/test/extended/images/signatures.go
@@ -2,7 +2,6 @@ package images
 
 import (
 	"fmt"
-	"os/exec"
 	"strings"
 
 	g "github.com/onsi/ginkgo"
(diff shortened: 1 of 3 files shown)


yselkowitz:
Test in question passed in e2e-aws-serial and e2e-aws-image-registry.  Failures appear to be flakes.
/retest
/assign @danwinship

PR #26463, opened 2021-09-10
Bug 2003252: builds: fix BUILD_LOGLEVEL tests for nightlies

yselkowitz:
These tests assume that builder's OS_GIT_MAJOR starts with 'v' as it
does in CI.  However, productized builds (nightlies and stable) are
built with OS_GIT_MAJOR=4.  When these tests from the openshift/build
testsuite are run on a nightly, the first fails, and the second could
return a false negative.

/cc @adambkaplan


adambkaplan (approved):
/lgtm

PR #26472, opened 2021-09-17
e2e: Wait for an individual imagestream import

adambkaplan:
Refactor e2e framework code so that tests can wait for an individual
imagestream to be imported. This can help address CI flakes where a test
needs to wait for an imagestream to be imported.


adambkaplan:
fyi @dmage @ricardomaraschini

test/extended/util/framework.go
```diff
@@ -236,127 +236,152 @@ func processScanError(log string) error {
 // WaitForOpenShiftNamespaceImageStreams waits for the standard set of imagestreams to be imported
 func WaitForOpenShiftNamespaceImageStreams(oc *CLI) error {
 	ctx := context.Background()
+	langs := []string{"ruby", "nodejs", "perl", "php", "python", "mysql", "postgresql", "jenkins"}
```
gabemontero, line 239: I mentioned this in the slack threads around this, but allow me to echo here.  And perhaps item asked in the various slack threads was not the root cause of the CI failure(s).

But with that preamable :-) 

The only thing the install payload imagestreams defined at https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml have to do with samples operator is that Clayton decided to put the imagestream defs in the manifest dir of the samples operator repo.

No creation/update, no monitoring, no tracking, no management, no retries from samples operator for those.

The CVO in fact creates/updates the items in https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml

One thing I'll highlight:  they all have the `scheduled` bit turned on.  i.e. https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml#L89

That means if the initial import fails, the imagestream related controllers in the OCM will try again in 10 to 15 minutes :-)

test/extended/util/framework.go
```diff
@@ -236,127 +236,152 @@ func processScanError(log string) error {
...
+		e2e.Logf(strbuf.String())
+	}
+	return pollErr
+}
+
+// retrySamplesOperatorImagestreamImport immediately retries an import for the provided imagestream if an imagestream managed by the samples operator fails to import.
+// This allows imagestreams to be reimported at a faster cadence than what the samples operator currently provides.
+// Imagestreams which use images in the OCP payload are not managed by the samples operator and therefore will not be retried.
+// Returns true if the imagestream import was retried.
+func retrySamplesOperatorImagestreamImport(ctx context.Context, oc *CLI, imagestream string) (bool, error) {
```
gabemontero, line 296: I understand the name of the bool param could indicate that retries may not occur, but I think that is still somewhat subtle.

I also think we can lose `Operator` from the name.

I'd perfer `retryAsNeededSampleImagestreamImport`

test/extended/util/framework.go
```diff
@@ -236,127 +236,152 @@ func processScanError(log string) error {
...
+		switch {
+		case condition.Type == configv1.OperatorDegraded && condition.Status == configv1.ConditionTrue:
+			// if degraded, bail ... unexpected results can ensue
+			return false, processScanError(fmt.Sprintf("samples-operator is degraded with reason: %s", condition.Reason))
+		case condition.Type == configv1.OperatorProgressing:
+			// if the imagestreams for one of our langs above failed, we abort,
+			// but if it is for say only EAP streams, we allow
+			if condition.Reason == "FailedImageImports" {
+				msg := condition.Message
+				if strings.Contains(msg, " "+imagestream+" ") || strings.HasSuffix(msg, " "+imagestream) {
```
gabemontero, line 312: @dperaza4dustbit - I actually made a mistake when talking with @adambkaplan about this section of the code, thinking it had to be redone.  It does not.

Quiz for you:  can you post some links from the code in https://github.com/openshift/cluster-samples-operator as to why this is correct, even after the rewrite I did in 4.7, where we started employing configmaps to track imagestream status?

gabemontero:
> Refactor e2e framework code so that tests can wait for an individual
> imagestream to be imported. This can help address CI flakes where a test
> needs to wait for an imagestream to be imported.

put another way, this effectively extends the amount of time we wait ... we were previously waiting the fixed set of time for the entire list of "language + jenkins" imagestreams to import

now we are waiting the same fixed set of time for each of those imagestreams

gabemontero (approved):
I'll

/lgtm

as I mentioned in my first comment here, we are in the ends extending the time we will wait as part of this code reorg, but I will also 

/hold

as @dperaza4dustbit needs to understand what is going on here.  I even posted a quiz of sorts for him in my review.

He can cancel the hold when he's done reviewing, adding comments/questions, answering my question to him.

I also hopped up to the podium for a sec to reiterate some clarification around the "install payload" imagestreams that were discussed in some of the

adambkaplan:
> put another way, this effectively extends the amount of time we wait ... we were previously waiting the fixed set of time for the entire list of "language + jenkins" imagestreams to import
> 
> now we are waiting the same fixed set of time for each of those imagestreams

true, however this also enables other teams to wait for a particular imagestream import to finish in a test. Since imagestream imports are run in parallel, I would expect we would only have a marginal/extreme cases where the wait time increases for our "image eco" imagestreams but doesn't fail a test.

gabemontero:
> > put another way, this effectively extends the amount of time we wait ... we were previously waiting the fixed set of time for the entire list of "language + jenkins" imagestreams to import
> > now we are waiting the same fixed set of time for each of those imagestreams
> 
> true, however this also enables other teams to wait for a particular imagestream import to finish in a test. Since imagestream imports are run in parallel, I would expect we would only have a marginal/extreme cases where the wait time increases for our "image eco" imagestreams but doesn't fail a test.

ok yep I see that now that you mention it and I go back and see https://github.com/openshift/origin/pull/26472/files#diff-35a89a7a7362642eebb559fb8564e857b00d6f7dd6322c3adabaf1adbd609d35R252

thanks for the clarification - good info

gabemontero:
just realized I lgtm'ed, but I had also asked for a name change on one of the methods introduced

so how about 

/lgtm cancel

until we sort that out

I'll leave the hold on for now in case I repost the lgtm before @dperaza4dustbit gets to this

test/extended/util/framework.go
```diff
@@ -236,127 +236,152 @@ func processScanError(log string) error {
 // WaitForOpenShiftNamespaceImageStreams waits for the standard set of imagestreams to be imported
 func WaitForOpenShiftNamespaceImageStreams(oc *CLI) error {
 	ctx := context.Background()
+	langs := []string{"ruby", "nodejs", "perl", "php", "python", "mysql", "postgresql", "jenkins"}
+	e2e.Logf("waiting for image ecoystem imagestreams to be imported")
```
sthaha, line 240: nit: ecosystem

test/extended/util/framework.go
sthaha: Can this list also include the imagestream - `tools` that provides `openshift/tools` image  which is used extensively throughout the tests referred to as `ShellImage()`. 
Also, All calls to `CreateExecPodOrFail` use this image.

test/extended/util/framework.go
```diff
@@ -236,127 +236,152 @@ func processScanError(log string) error {
...
+	// pollErr will be not nil if there was an immediate error, or we timed out.
+	if pollErr == nil {
+		return nil
+	}
+	DumpImageStream(oc, "openshift", imagestream)
+	// If retried=true at this point, it means that we have repeatedly tried to reimport the imagestream and failed to do so.
+	// This could be an indicator that the Red Hat Container Registry (registry.redhat.io) is experiencing an outage, since most samples operator imagestream images are hosted there.
+	if retried {
+		strbuf := bytes.Buffer{}
+		strbuf.WriteString("Failed immagestream imports may indicate an issue with the Red Hat Container Registry (registry.redhat.io).\n")
```
sthaha, line 283: nit: immage -> image

sthaha: nice!

sthaha (commented):
/lgtm 

Thanks for the patch 

* Could you please include the `tools` imagestream as well
* Minor nits that you may want to correct.

test/extended/util/framework.go
gabemontero: no it cannot @sthaha , at least not in this function, the tools imagestream will not show up in the samples operator object that is inspected below and the initiation of retries that is done if the import fails, since it is not being tracked at 
per my comment at https://github.com/openshift/origin/pull/26472/files#r711270195

but I will note that all the in payload imagesterams have scheduled import turned on, so the imagestream controller will retry every 15 minutes if there is a failure

more restructure would be needed if you wanted to expedite retry on things like tools for imagestreams not managed by samples operator

test/extended/util/framework.go
gabemontero: @dperaza4dustbit and I confirmed his understanding in slack ... and fwiw it basically correlates with my answer to @sthaha with https://github.com/openshift/origin/pull/26472/files#r715648199

adambkaplan:
@gabemontero updated the function name and cleaned up the godoc.

gabemontero:
> @gabemontero updated the function name and cleaned up the godoc.

thanks for that @adambkaplan .... you executed what  I previously asked for wrt `retryAsNeededSampleImagestreamImport`

..that said ...

per a conversation I had with various folks on #forum-release-oversight yesterday, there is still confustion with the distinction between imagestreams in the openshift namespace that are samples and are managed by the samples operator, versus those in payload imagestreams that are created by the CVO, but are "associated" with samples because of their declaration in one of the sample operator manifest files, but were those in payload imagestreams are in no way managed by samples operator.

So it occurs to me that we need to try and make that clear in our methods here as well

So an additional ask I did not thing of to make previously (apologies).

Let's change `WaitForOpenShiftNamespaceImagestream` to `WaitForOpenShiftNamespaceImagestreamsManagedBySamplesOperator`

I'll make the comment in line as well

in the meantime, I'll release the hold

/hold cancel

test/extended/util/framework.go
```diff
@@ -236,127 +236,157 @@ func processScanError(log string) error {
...
+		if err != nil {
+			DumpSampleOperator(oc)
+			return err
+		}
+	}
+	return nil
+}
 
+// WaitForOpenShiftNamespaceImagestream waits for an individual imagestream in the openshift namespace to be imported.
+func WaitForOpenShiftNamespaceImagestream(ctx context.Context, oc *CLI, imagestream string) error {
```
gabemontero, line 252: aside from the name change here @adambkaplan to `WaitForOpenShiftNamespaceImagestreamsManagedBySamplesOperator` let's make sure godoc is clear that sampels operator does not managae the imagestreams defined in https://github.com/openshift/cluster-samples-operator/blob/master/manifests/08-openshift-imagestreams.yaml and that this method and the ones it calls by no means confirms whether the imports on those in payload imagestreams has completed

i.e. the progressing / available / degraded conditions on samples operator are in no way affected by the import status of those in payload imagestreams

gabemontero:
/lgtm

/retest

the remaining e2e failures are flakes unrelated to these changes

PR #25959, opened 2021-03-13
test: Prevent limits from being set on pods and enforce requests

smarterclayton:
Enforce the rules we define in CONVENTIONS.md

PR #26483, opened 2021-09-27
Skip pod limit check on single-node topology

jerpeter1:
When workload partitioning is enabled on SNO clusters, limits are being set on management.workload.openshift.io/cores, but setting limits on management.workload.openshift.io/cores does not have the same effect as setting cpu.limits (there is no throttling). This PR disables the test which checks for cpu.limits being set on single-node topologys.

```
[root@sno-test4 core]# kubectl get pods -n openshift-kube-apiserver kube-apiserver-sno-test4 -o json | jq '.spec.containers[] | select(.name=="kube-apiserver") | .resources'
{
  "limits": {
    "management.workload.openshift.io/cores": "265"
  },
  "requests": {
    "management.workload.openshift.io/cores": "265",
    "memory": "1Gi"
  }
}

[root@sno-test4 core]# kubectl get pods -n openshift-kube-apiserver kube-apiserver-sno-test4 -o json | jq -r '.status.containerStatuses[] | select (.name=="kube-apiserver") | .containerID' | sed 's/cri-o:\/\///g'
30923592458b8e5d58ba2d3e9a6cbaa1ef27700768a497663db8b063712801b9

[root@sno-test4 core]# crictl inspect 30923592458b8e5d58ba2d3e9a6cbaa1ef27700768a497663db8b063712801b9 | jq '.info.pid'
108174

[root@sno-test4 core]# crictl inspect 30923592458b8e5d58ba2d3e9a6cbaa1ef27700768a497663db8b063712801b9 | jq '.info.runtimeSpec.linux.resources'
{
  "devices": [
    {
      "allow": true,
      "access": "rwm"
    }
  ],
  "cpu": {
    "shares": 271,
    "quota": 0,
    "period": 100000,
    "cpus": "0-1"
  },
  "pids": {
    "limit": 1024
  },
  "hugepageLimits": [
    {
      "pageSize": "2MB",
      "limit": 0
    },
    {
      "pageSize": "1GB",
      "limit": 0
    }
  ]
}

[root@sno-test4 crio-30923592458b8e5d58ba2d3e9a6cbaa1ef27700768a497663db8b063712801b9.scope]# cat /proc/108174/cgroup | grep cpu
9:cpuset:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod65718bbe_7c00_4a7a_a881_4920da8cd3e6.slice/crio-30923592458b8e5d58ba2d3e9a6cbaa1ef27700768a497663db8b063712801b9.scope
2:cpu,cpuacct:/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod65718bbe_7c00_4a7a_a881_4920da8cd3e6.slice/crio-30923592458b8e5d58ba2d3e9a6cbaa1ef27700768a497663db8b063712801b9.scope

[root@sno-test4 ~]# cat /sys/fs/cgroup/cpu,cpuacct/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod65718bbe_7c00_4a7a_a881_4920da8cd3e6.slice/crio-30923592458b8e5d58ba2d3e9a6cbaa1ef27700768a497663db8b063712801b9.scope/cpu.stat
nr_periods 0
nr_throttled 0
throttled_time 0
```


jerpeter1:
> For now the test should be skipped when workload partitioning is enabled in general

Done

```
$ ./openshift-tests run-test "[sig-arch] Managed cluster should set requests but not limits [Suite:openshift/conformance/parallel]"
[BeforeEach] [Top Level]
  github.com/openshift/origin/test/extended/util/framework.go:1453
[BeforeEach] [Top Level]
  github.com/openshift/origin/test/extended/util/framework.go:1453
[BeforeEach] [Top Level]
  github.com/openshift/origin/test/extended/util/test.go:61
[BeforeEach] [sig-arch] Managed cluster
  github.com/openshift/origin/test/extended/util/client.go:142
STEP: Creating a kubernetes client
[It] should set requests but not limits [Suite:openshift/conformance/parallel]
  github.com/openshift/origin/test/extended/operators/resources.go:55
STEP: checking namespaces for WP annotation
Sep 28 10:59:59.170: INFO: test is not relevant on clusters with workload partitioning enabled
[AfterEach] [sig-arch] Managed cluster
  github.com/openshift/origin/test/extended/util/client.go:140
[AfterEach] [sig-arch] Managed cluster
  github.com/openshift/origin/test/extended/util/client.go:141
skip [github.com/openshift/origin/test/extended/operators/resources.go:39]: test is not relevant on clusters with workload partitioning enabled
```

test/extended/operators/resources.go
```diff
@@ -13,8 +13,33 @@ import (
...
+func skipIfWPEnabled(oc *exutil.CLI) {
+	g.By("checking namespaces for WP annotation")
+
+	namespaces, err := oc.KubeFramework().ClientSet.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		e2e.Failf("unable to list namespaces: %v", err)
+	}
+
+	annotationForAllowingManagementCores := "workload.openshift.io/allowed"
+	annotationValueForAllowingManagementCores := "management"
```
omertuc, line 28: I think any workload partitioning would cause issues, not just management

test/extended/operators/resources.go
```diff
@@ -13,8 +13,33 @@ import (
...
+
+	for _, ns := range namespaces.Items {
+		if val := ns.Annotations[annotationForAllowingManagementCores]; strings.Contains(val, annotationValueForAllowingManagementCores) {
+			wpEnabled = true
+			break
+		}
+	}
+
+	if wpEnabled {
+		e2eskipper.Skipf("test is not relevant on clusters with workload partitioning enabled")
```
omertuc, line 39: I think it's still relevant, it's just that the way it's currently written it doesn't work. Maybe the message should be changed to indicate that the test needs to be improved to be made aware of WP

test/extended/operators/resources.go
```diff
@@ -13,8 +13,33 @@ import (
...
+	if err != nil {
+		e2e.Failf("unable to list namespaces: %v", err)
+	}
+
+	annotationForAllowingManagementCores := "workload.openshift.io/allowed"
+	annotationValueForAllowingManagementCores := "management"
+	wpEnabled := false
+
+	for _, ns := range namespaces.Items {
+		if val := ns.Annotations[annotationForAllowingManagementCores]; strings.Contains(val, annotationValueForAllowingManagementCores) {
```
omertuc, line 32: so maybe just check if the annotation is there, rather than checking if it contains/equals `management` specifically

test/extended/operators/resources.go
jerpeter1: I put up another change when fixes up the test to support WP.

omertuc:
Consider squashing the commits to a single one with a commit message explaining what this is all about (linking to the enhancement would give a lot of the needed context)

test/extended/operators/resources.go
```diff
@@ -144,6 +144,9 @@ var _ = g.Describe("[sig-arch] Managed cluster", func() {
 					// Pods may not set limits
```
njhale, line 144: I'm not sure why these tests are under the `operators` test package. From what I can see, they don't look related to the operator-framework. Were these tests placed here by mistake?

omertuc: Looks like it was originally added in #25959

njhale (approved):
After grokking the threads here, I think this makes sense, so:

 /approve

njhale:
> Standard error from the command:\nerror: the path \"/var/tests/test/cmd/testdata/multiport-service.yaml\" does not exist

This doesn't look like something that can be fixed by retesting. Seems unrelated, but couldn't find any issues on it either!

(see https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26483/pull-ci-openshift-origin-master-e2e-agnostic-cmd/1443639193447174144#1:build-log.txt%3A108)

njhale:
/hold

Not sure if retest is enabled, but considering https://github.com/openshift/origin/pull/26483#issuecomment-933552158 it doesn't look like this will pass until those unrelated tests are fixed.

PR #26484, opened 2021-09-28
Add baremetal operator validating webhook healthy test

ardaguclu:
This PR adds test to verify baremetal operator validating
webhook is healthy. It does this by trying to update bootMACAddress
which is not allowed and updating itself which should not get error.

Enabling BMO validating webhook PR https://github.com/openshift/cluster-baremetal-operator/pull/203 is prerequisite of this PR.

### DIFF
--- a/test/extended/baremetal/hosts.go
+++ b/test/extended/baremetal/hosts.go
@@ -93,4 +93,34 @@ var _ = g.Describe("[sig-installer][Feature:baremetal] Baremetal platform should
 			expectBoolField(h, "spec.online").To(o.BeTrue())
 		}
 	})
+
+	g.It("not allow updating BootMacAddress", func() {
+		skipIfNotBaremetal(oc)
+
+		dc := oc.AdminDynamicClient()
+		bmc := baremetalClient(dc)
+
+		hosts, err := bmc.List(context.Background(), v1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(hosts.Items).ToNot(o.BeEmpty())
+
+		host := hosts.Items[0]
+		expectStringField(host, "spec.bootMACAddress").ShouldNot(o.BeNil())
+		// Already verified that bootMACAddress exists
+		bootMACAddress, _, _ := unstructured.NestedString(host.Object, "spec", "bootMACAddress")
+		testMACAddress := "11:11:11:11:11:11"
+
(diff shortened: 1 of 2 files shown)


ardaguclu:
@andfasano you are right, I renamed test name and add more granularity as you suggested. Thanks for review.

test/extended/baremetal/hosts.go
```diff
@@ -93,4 +93,40 @@ var _ = g.Describe("[sig-installer][Feature:baremetal] Baremetal platform should
...
+		o.Expect(err).To(o.HaveOccurred())
+		o.Expect(err.Error()).To(o.ContainSubstring("bootMACAddress can not be changed once it is set"))
+
+		g.By("verify bootMACAddress is not updated")
+		h, err := bmc.Get(context.Background(), host.GetName(), v1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		check, _, _ := unstructured.NestedString(h.Object, "spec", "bootMACAddress")
+		o.Expect(check).To(o.Equal(bootMACAddress))
+
+		g.By("updating it's own bootMACAddress which is allowed")
```
andfasano, line 126: Is it really required this check? I'm not clear about the added value

ardaguclu: Since initial implementation was testing webhook functionality, it had a meaning; this operation is not allowed and this operation is allowed. However, now from customer's point of view, it is redundant. Thanks @andfasano, I'll remove that part.

andfasano:
@ardaguclu is it possible to test the PR against a version where the webhook _wasn't_ available (ie 4.8 I guess)?

ardaguclu:
> @ardaguclu is it possible to test the PR against a version where the webhook _wasn't_ available (ie 4.8 I guess)?

@andfasano this test will only for work 4.10+, what we get doing that?

andfasano:
> > @ardaguclu is it possible to test the PR against a version where the webhook _wasn't_ available (ie 4.8 I guess)?
> 
> @andfasano this test will only for work 4.10+, what we get doing that?

Oh, right, there wouldn't be any way that this test will get executed on versions < 4.10 (unless explicitly backported ofc)

ardaguclu:
@bparees could you please have a look at this newly added test?, or who should we ask about this?

bparees:
/approve

@andfasano (or someone else from the team)( should still provide the lgtm.

PR #26487, opened 2021-09-28
stop adding individual serviceaccount namespace groups to SCC for e2e tests

deads2k:
after https://github.com/openshift/origin/pull/26486 and https://github.com/openshift/kubernetes/pull/984, this will eliminate the race during e2e tests to get the "right" SCC permissions.

### DIFF
--- a/test/extended/util/test.go
+++ b/test/extended/util/test.go
@@ -18,7 +18,6 @@ import (
 
 	kapiv1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kclientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


test/extended/util/test.go
```diff
@@ -210,32 +208,23 @@ func createTestingNS(baseName string, c kclientset.Interface, labels map[string]
 		baseName = "e2e-" + baseName
 	}
 
+	// we skip SCC if this is a kube e2e test
+	isKubeE2ENamespace := strings.HasPrefix(baseName, "e2e-k8s-") || (isKubernetesE2ETest() && !skipTestNamespaceCustomization())
+	if isKubeE2ENamespace{
+		labels["security.openshift.io/disable-securitycontextconstraints"] = "true"
```
aojea, line 214: is this safe? or should it check that `labels` is not nil to avoid panicing?

aojea: I see some places that are setting labels to nil IIUIC
https://github.com/openshift/origin/blob/aa7e2387a3c8ac48293307340fbe5c40fc186577/vendor/k8s.io/kubernetes/test/e2e/apimachinery/namespace.go#L89

s-urbaniak: agreed :+1:

PR #26490, opened 2021-09-29
Bug 2008926: etcd-storage-test: mark node ready to save it for 5min from cloud provider

sttts:
Upstream runs this as integration test, but we do as e2e. Hence, nodes can be deleted.


elmiko (commented):
+1

dgoodwin:
+1 from me thank you @sttts

aojea:
the test passed 
```
: [sig-api-machinery] API data in etcd should be stored at the correct location and version for all resources [Serial] [Suite:openshift/conformance/serial]
```
/retest

dgoodwin:
/lgtm

Worth a shot.

sttts:
Different flakes.

/retest

PR #26492, opened 2021-10-01
Bug 1980127: include the pod name in the cni binary

JacobTanenbaum:
this will make sure that if the sdn pod is restarting there is no change
that the exiting pod will delete the openshift-sdn binary that the
currently starting pod is assuming is there


JacobTanenbaum:
depends on https://github.com/openshift/openshift-ansible/pull/12354

PR #26494, opened 2021-10-01
record only known images test pass so we can get a percentage of passes

deads2k:
lack of passes in addition to failures on this test impacted https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/aggregated-gcp-sdn-upgrade-4.10-micro-release-openshift-release-analysis-aggregator/1443917511878774784

though the run was likely doomed overall.


deads2k:
getting the split for CI as we're adding new history for additional jobs.

PR #26495, opened 2021-10-01
split the event should not repeat pathologically into e2e and core

deads2k:
The core are more severe.  The e2e namespaces caught an imagestream problem and a single-node upgrade problem this week, so they have value.    This gives us two different names to react to depending on urgency.

/assign @stbenjam


deads2k:
getting the split for CI as we're adding new history for additional jobs.

PR #26498, opened 2021-10-02
Tolerate API downtime in upgrade disruption tests on single-node topologies

omertuc:
Currently the tests tolerate 8% API disruption on <4.8 clusters and 0% API disruption on
\>=4.8 azure,aws,gce clusters

This test makes it so no matter what version / cloud provider is used, if the cluster control
plane topology is single node, then 15% API disruption is allowed during an upgrade. This is
because there is no way to avoid disruption on single-node topologies, but we would still
like to make sure that we don't regress to too much downtime.

### DIFF
--- a/test/extended/util/disruption/controlplane/controlplane.go
+++ b/test/extended/util/disruption/controlplane/controlplane.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/kubernetes/test/e2e/upgrades"
 
 	"github.com/blang/semver"
+	apiconfigv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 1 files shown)


test/extended/util/disruption/controlplane/controlplane.go
```diff
@@ -124,6 +135,12 @@ func (t *availableTest) Test(f *framework.Framework, done <-chan struct{}, upgra
 			toleratedDisruption = 0
 		}
 	}
+
+	if controlPlaneTopology, _ := getTopologies(f); controlPlaneTopology == apiconfigv1.SingleReplicaTopologyMode {
```
deads2k, line 139: add to switch above please

deads2k:
the idea is sound

/approve

PR #26502, opened 2021-10-04
Bug 1992596: e2e/cli: More basicresources.sh

deejross:
Moved more of the basicresources.sh over to Go tests, restored a missing fixture, and provide a way to replace images used in fixtures to ensure we always use known images.

### DIFF
--- a/cmd/openshift-tests/images.go
+++ b/cmd/openshift-tests/images.go
@@ -244,6 +244,7 @@ func pulledInvalidImages(fromRepository string) ginkgo.JUnitForEventsFunc {
 			if !ok {
 				byImage = sets.NewString()
 				pulls[image] = byImage
+				fmt.Printf("[sig-arch] unknown image: %s\n", image)
 			}
 			byImage.Insert(event.Locator)
 		}
--- a/test/extended/cli/basics.go
+++ b/test/extended/cli/basics.go
@@ -1,0 +1,121 @@
+package cli
+
+import (
+	"os"
+	"path/filepath"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	k8simage "k8s.io/kubernetes/test/utils/image"
+)
+
+var _ = g.Describe("[sig-cli] oc basics", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc                   = exutil.NewCLI("oc-basics")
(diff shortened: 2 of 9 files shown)


cmd/openshift-tests/images.go
```diff
@@ -244,6 +244,7 @@ func pulledInvalidImages(fromRepository string) ginkgo.JUnitForEventsFunc {
 			if !ok {
 				byImage = sets.NewString()
 				pulls[image] = byImage
+				fmt.Printf("[sig-arch] unknown image: %s\n", image)
```
soltysh, line 247: This is not needed, we print them in the next if clause.

test/extended/cli/basics.go
```diff
@@ -0,0 +1,121 @@
...
+	})
+
+	g.It("can describe an OAuth access token", func() {
+		// need admin here
+		ocAdmin := oc.AsAdmin()
+
+		err := ocAdmin.Run("create").Args("-f", oauthAccessTokenFile).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		out, err := ocAdmin.Run("describe").Args("oauthaccesstoken", "sha256~efaca6fab897953ffcb4f857eb5cbf2cf3a4c33f1314b4922656303426b1cfc9").Output()
```
soltysh, line 114: Nit: you forgot to drop that from basicresource.sh here: https://github.com/openshift/origin/blob/e000efe2de764b8f0789bbea7bc2ebe248eb5275/test/extended/testdata/cmd/test/cmd/basicresources.sh#L198-L201

test/extended/cli/resources.go
```diff
@@ -176,3 +179,46 @@ func writeObjectToFile(obj runtime.Object) (string, error) {
 
 	return name, nil
 }
+
+// replaceImageInFile reads in a file, replaces the given image name, then writes to a new temporary file.
+func replaceImageInFile(source, oldImage, newImage string) (string, error) {
```
soltysh, line 184: I'm not a fan of manually replacing like so, I'd suggest using templates similarly how we do that upstream, see: https://github.com/openshift/origin/blob/e000efe2de764b8f0789bbea7bc2ebe248eb5275/vendor/k8s.io/kubernetes/test/e2e/kubectl/kubectl.go#L296-L303

soltysh (approved):
I'm merging this as is to unblock oc tests, but I expect the bits mentioned here to be fixed in a followup.

/lgtm
/approve

PR #26504, opened 2021-10-05
test/extended/prometheus: Validate alerting rules

bison:
The [OpenShift Alerting Consistency][1] enhancement defines a style
guide for the alerts shipped as part of OpenShift.  This adds a test
validating some of the guidelines considered required.

This was originally added in #26476, but was reverted in #26499 due to
failures with OVN and vSphere clusters.  This adds the tests back, but
adds exceptions for the non-compliant alerts as well as marking the
failing tests as flakes for now.  We'll gather data and make the tests
required once we're reasonably sure things are passing with all the
existing alerts.

[1]: https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md


bison:
I've run this on an OVN cluster and it passed. I haven't been able to get a vSphere cluster to spin up successfully, but I added exceptions for the problematic rules I saw in CI and that are listed in the bug that was created.

bison:
@simonpasquier: I found a new alert that needed an exception, so lost the LGTM. Can you have another look? That's the only change.

bison:
@deads2k, this adds the tests for alerting rules again after they were reverted. The problematic tests are marked as flakes now, in addition to having added temporary exceptions for the OVN and vSphere alerts. Can you have a look?

The `e2e-gcp-upgrade` tests are failing, but it looks unrelated.

PR #26510, opened 2021-10-07
Bug 1960674: images: port image signature workflow test to OCP4/UBI8

yselkowitz:
This fixes portability issues with this test, given that origin-control-plane:v3.11 (besides being obsolete) is only available for x86_64.  Also, copying the oc binary from the machine running the test into a container on the cluster would cause the test to fail when run on a different architecture, or when oc was built for a newer glibc. There was also no guarantee that the oc found on the test machine matched that of the payload being tested.

Instead, this uses the payload cli container, which is based on RHEL 8.  Because 4.y CI builds do not have any publicly accessible yum repos, UBI 8 is added manually to get skopeo and its dependencies. The rest of the changes are a consequence of adapting to the newer versions of gpg2 and skopeo.

v2: fix key type for RHEL 8 FIPS compatibility (see #26479 and #26485)

/assign @stbenjam

### DIFF
--- a/test/extended/images/signatures.go
+++ b/test/extended/images/signatures.go
@@ -2,7 +2,6 @@ package images
 
 import (
 	"fmt"
-	"os/exec"
 	"strings"
 
 	g "github.com/onsi/ginkgo"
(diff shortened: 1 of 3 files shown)


yselkowitz:
Both e2e-aws-serial and e2e-gcp-fips-serial passed, so this should be in good shape; the remaining failures are unrelated.
/retest

yselkowitz:
it already passed e2e-gcp-fips-serial, so yes it's ok.

mandre:
We see the test consistently fail the [periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.9-e2e-openstack-proxy job](https://prow.ci.openshift.org/job-history/origin-ci-test/logs/periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.9-e2e-openstack-proxy).
/cherry-pick release-4.9

PR #26514, opened 2021-10-07
Added 2 Whereabouts e2e tests

nicklesimba:
Created 2 e2e tests for whereabouts. 

The first makes 4 pods with a whereabouts IPAM that only allows 3 IP addresses. The expected result is that the first 3 come up successfully and the 4th pod doesn't.

The second makes 2 pods, one with a parameter to sleep for ~20 seconds~ 5 seconds (configurable via `whereabouts-race-sleepy.yml`, located in `test/extended/testdata/net-attach-defs`) after getting its IP address, and the second with no such parameter. The expected result is that a race condition should not occur on working code, and both pods should be assigned unique IPs.

### DIFF
--- a/test/extended/networking/whereabouts.go
+++ b/test/extended/networking/whereabouts.go
@@ -1,0 +1,171 @@
+package networking
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	g "github.com/onsi/ginkgo"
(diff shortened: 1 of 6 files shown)


nicklesimba:
Altered yaml files to work with configurable sleep_for_race parameter.

PR #26515, opened 2021-10-08
ignore ClusterNotUpgradeable alert on a TechPreviewNoUpgrade cluster

jkhelil:
the e2e-aws-builds-techpreview test failed because the ClusterNotUpgradeable alert was fired.
On a TechPreviewNoUpgrade cluster we must ignore the  ClusterNotUpgradeable alert
fired by the Kube API Operator. the same as the TechPreviewNoUpgrade  alert


gabemontero:
@jkhelil - you should put some context in thsi PR's description and summarize why we are adding this alert; otherwise, @jan--f is naturally going to have some questions about why we are doing this.

@adambkaplan 's comment in your openshift/release PR is a good starting point

jkhelil:
> @jkhelil - you should put some context in thsi PR's description and summarize why we are adding this alert; otherwise, @jan--f is naturally going to have some questions about why we are doing this.
> 
> @adambkaplan 's comment in your openshift/release PR is a good starting point

@gabemontero I updated the PR description
/assign @jan--f

PR #26519, opened 2021-10-08
WIP test/e2e/upgrade/alert: attach found alerts to junit as EventIntervals

vrutkovs:
e2e test is printing which alerts have fired during the test. It would be useful to render these on e2e timeline as well


vrutkovs:
Probably a dupe of #26508

PR #26530, opened 2021-10-19
Bug 2011293: add default search image registry test to build e2es

gabemontero:


gabemontero:
Examples of the new test failing without the fix
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26530/pull-ci-openshift-origin-master-e2e-gcp/1450536862744580096
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26530/pull-ci-openshift-origin-master-e2e-aws-single-node/1450536861901524992
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26530/pull-ci-openshift-origin-master-e2e-gcp/1450536862744580096

gabemontero:
bump @adambkaplan all green for the new e2e for build default search registries

test/extended/builds/docker_pullsearch.go
```diff
@@ -0,0 +1,48 @@
...
+				exutil.DumpPodLogsStartingWith("", oc)
+			}
+		})
+
+		g.Describe("Building from a Dockerfile whose FROM image ref does not specify the image registry", func() {
+			g.It("should create a docker build that has buildah search from our predefined list of image registries and succeed", func() {
+
+				g.By(fmt.Sprintf("calling oc create -f %q", buildFixture))
+				err := oc.Run("create").Args("-f", buildFixture).Execute()
+				o.Expect(err).NotTo(o.HaveOccurred())
```
adambkaplan, line 38: I believe the full `oc` invocation is logged if the test fails - `g.By` should use a more general statement:

```suggestion
				g.By("creating a BuildConfig whose base image does not have a fully qualified registry name")
				err := oc.Run("create").Args("-f", buildFixture).Execute()
				o.Expect(err).NotTo(o.HaveOccurred())
```

test/extended/testdata/builds/test-build-search-registries.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata:
+  name: ubi
+spec:
+  source:
+    # ubi8/ubi:latest only resolves to an image on registry.redhat.io or registry.access.redhat.com.
+    dockerfile: FROM ubi8/ubi:latest
```
adambkaplan, line 8: I think we need to add this to the list of allowed images. This may require some level of scrutiny because this test covers functionality that impacts disconnected use cases:

https://github.com/openshift/origin/blob/master/test/extended/util/image/image.go

adambkaplan (changes requested):

test/extended/testdata/builds/test-build-search-registries.yaml
gabemontero: I picked ubi8/ubi because it was already being used elsewhere:  https://github.com/openshift/origin/blob/master/test/extended/testdata/builds/test-build-proxy.yaml#L101

you in fact introduced that with https://github.com/openshift/origin/commit/bb063138df44488b9fb0f02040fee87a1a63cd95

And that PR https://github.com/openshift/origin/pull/25778 merged in April, at which point https://github.com/openshift/origin/blob/master/test/extended/util/image/image.go was in full throttle, reviewing its git history

So either we overlooked the mirroring hoopla back then, or ubi8/ubi is so core/essential it gets pulled in elsewhere.

But if it was an oversight, I'm more inclined to use the busybox entry already in there, k8s.gcr.io/e2e-test-images/busybox:1.29-1

test/extended/builds/docker_pullsearch.go
gabemontero: suggested commit pulled in, fmt import removed, commits squashed ... thanks @adambkaplan

test/extended/testdata/builds/test-build-search-registries.yaml
adambkaplan: the busybox image breaks the premise of the test. `k8s.gcr.io` is not a container registry on our search list, and moreover `busybox` is a default image on Docker Hub.

adambkaplan: > So either we overlooked the mirroring hoopla back then, or ubi8/ubi is so core/essential it gets pulled in elsewhere.

It's possible the ubi layers are present in other images we use. That would depend on a lot of things.
Feels like there was an oversight here, because we indirectly reference the image in our test. This is also explicitly a part of the build suite, which means it isn't run in the normal conformance test.

I think we are OK merging here, then having a follow-up PR that adds the image to the mirror process.

gabemontero: > the busybox image breaks the premise of the test. `k8s.gcr.io` is not a container registry on our search list, and moreover `busybox` is a default image on Docker Hub.

duh - wasn't thinking :-/ thanks @adambkaplan

gabemontero: > > So either we overlooked the mirroring hoopla back then, or ubi8/ubi is so core/essential it gets pulled in elsewhere.
> 
> It's possible the ubi layers are present in other images we use. That would depend on a lot of things. Feels like there was an oversight here, because we indirectly reference the image in our test. This is also explicitly a part of the build suite, which means it isn't run in the normal conformance test.
> 
> I think we are OK merging here, then having a follow-up PR that adds the image to the mirror process.

sounds like a plan thanks

gabemontero:
unrelated flakes in e2e-gcp-builds

our new test, which is the only thing this PR touches, still passes:

`

:  [sig-builds][Feature:Builds][pullsearch] docker build where the  registry is not specified  Building from a Dockerfile whose FROM image  ref does not specify the image registry should create a docker build  that has buildah search from our predefined list of image registries and  succeed [Suite:openshift/conformance/parallel] | 34s
-- | --


`

adambkaplan (approved):
/lgtm

gabemontero:
non sig-build flakes have been hitting e2e-gcp-builds

our new test here is passing

PR #26537, opened 2021-10-21
Bug 2015624: Refactor IBM Cloud skips to External

cjschaef:
Refactor code related to skipping tests on IBM Cloud to skip based
on External ControlPlaneTopology, or Managed offering. IBM Cloud
will be offering an OCP, unmanaged, cluster that should not skip
such tests.


test/extended/authorization/authorization.go
```diff
@@ -250,16 +252,19 @@ func (test resourceAccessReviewTest) run() {
 				}
 			}
 
+			infra, err := test.ocClient.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
```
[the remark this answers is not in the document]
cjschaef: I hopefully have something sufficient to do that now

test/extended/util/framework.go
```diff
@@ -2002,3 +2006,18 @@ func AllClusterVersionsAreGTE(version semver.Version, config *rest.Config) (bool
 	}
 	return true, nil
 }
+
+// GetControlPlaneTopology retrieves the cluster infrastructure TopologyMode
+func GetControlPlaneTopology(ocClient *CLI) (*configv1.TopologyMode, error) {
+	if ControlPlaneTopology == nil {
```
alvaroaleman, line 2012: I think this needs to be threadsafe/protected by a sync.Mutex as tests could end up calling this in parallel

cjschaef: Added a mutex lock to the function

alvaroaleman: the mutex already needs to be used for reading ( `if ControlPlaneTopology == nil`) as reading/writing to a variable in parallel is a datarace.

cjschaef: Will the additional locking be acceptable, if we always perform it when this function is called, compared with the additional check after receiving the lock, prior to attempting to set the value?

I considered using the builtin `once`, but was not confident that that would work properly if an error was raised while attempting to get the `Infrastructure.Status.ControlPlaneTopology`.

alvaroaleman: If we just lock in the very beginning of the function prior to reading we are good. An additional check prior to setting the value would only be needed if we used a RWLock, but I don't think we need that.

cjschaef: okay, I'll make that change, thanks!

cjschaef: updated

alvaroaleman (approved):
thanks for your work!

PR #26548, opened 2021-10-27
write disruption data

deads2k:
write a data file that can be consumed by aggregation instead of parsing out of junit.

/hold

holding so that we can see the content is correct.  There is a followup needed to fix the four different ways we have to watch for disruption today. I didn't make it worse, but I think it's time to make it better.

/assign @stbenjam


deads2k:
/hold cancel

content is present and it doesn't appear to cause an issue.

PR #26555, opened 2021-11-01
Oauth server tests: Get Oauth issuer url from discovery instead of route

alvaroaleman:
The tests currently fetch a route and use its status.Ingress[0].Host to
find the oauth issuer. This route doesn't exist in the Hypershift
architecture, as the oauth server doesn't run in-cluster. Use the oauth
discovery information from the kube-apiserver instead, just like `oc`
does.

/assign @stlaz 
/cc @csrwng

### DIFF
--- a/test/extended/oauth/client_secret.go
+++ b/test/extended/oauth/client_secret.go
@@ -62,8 +62,7 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 			o.Expect(err).NotTo(o.HaveOccurred())
 			oc.AddResourceToDelete(userv1.GroupVersion.WithResource("users"), user)
 
-			oauthRoute, err := oc.AdminRouteClient().RouteV1().Routes("openshift-authentication").Get(ctx, "oauth-openshift", metav1.GetOptions{})
-			o.Expect(err).NotTo(o.HaveOccurred())
+			oauthDiscoveryData := getOAuthWellKnownData(oc)
 
 			for _, createTokenReq := range []func(string) *http.Request{
 				queryClientAuthRequest,
@@ -89,7 +88,7 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 
 				g.By("querying for a token")
 
-				tokenReq := createTokenReq(oauthRoute.Status.Ingress[0].Host)
+				tokenReq := createTokenReq(oauthDiscoveryData.TokenEndpoint)
 				tokenReq.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 				requestDump, err := httputil.DumpRequest(tokenReq, true)
 				o.Expect(err).NotTo(o.HaveOccurred())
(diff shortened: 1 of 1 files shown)


test/extended/oauth/client_secret.go
```diff
@@ -62,7 +65,7 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 			o.Expect(err).NotTo(o.HaveOccurred())
 			oc.AddResourceToDelete(userv1.GroupVersion.WithResource("users"), user)
 
-			oauthRoute, err := oc.AdminRouteClient().RouteV1().Routes("openshift-authentication").Get(ctx, "oauth-openshift", metav1.GetOptions{})
+			oauthIssuerURL, err := getOAuthEndpoint(oc.AdminConfig().Host)
```
csrwng, line 68: please feel free to ignore, but it seems that what you really want is the token_endpoint, since that's what ends up being used in the request functions ... like https://github.com/openshift/origin/pull/26555/files#diff-6548703e85e00efd742cc59e67cea251a6501651f05151ec0892e02b915dfdc6L117

alvaroaleman: True, but I wanted to keep the change minimal and that would require changing all of the request functions. I don't really have much feelings one way or another though.

alvaroaleman:

test/extended/oauth/client_secret.go
csrwng: It does reduce the assumptions about the request path being what it is, but yeah, not a big deal

test/extended/oauth/client_secret.go
```diff
@@ -113,6 +116,31 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 	})
 })
 
+func getOAuthEndpoint(apiserverURL string) (string, error) {
+	requestURL := strings.Join([]string{apiserverURL, ".well-known/oauth-authorization-server"}, "/")
+	request, err := http.NewRequest(http.MethodGet, requestURL, nil)
+	if err != nil {
+		return "", err
+	}
+	client := &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}}
```
stlaz, line 125: Needs Proxy as well

test/extended/oauth/client_secret.go
```diff
@@ -113,6 +116,31 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 	})
 })
 
+func getOAuthEndpoint(apiserverURL string) (string, error) {
```
stlaz, line 119: this could be replaced by https://github.com/openshift/origin/blob/5baea95eac9b776d483ca2ba89f6eca9e264e80d/test/extended/oauth/helpers.go#L44-L54

alvaroaleman: perfect, updated it to use that

test/extended/oauth/client_secret.go
alvaroaleman: I changed it to use the existing method and am assuming that will use a proxy if needed

alvaroaleman:
/hold
`clone repository using git:// protocol should clone using git://` fails because GitHub disabled that scheme and will not recover through retesting

alvaroaleman:
https://github.com/openshift/origin/pull/26560 merged
/hold cancel
/retest-required

PR #26559, opened 2021-11-02
Bug 2011895: Add synthetic test for cloud API throttling

jsafrane:
"[sig-arch] cloud API quota should not be exceeded" tells if OCP cluster under test encountered throttling from the underlyig cloud.

cc @openshift/storage @deads2k

### DIFF
--- a/pkg/synthetictests/storage.go
+++ b/pkg/synthetictests/storage.go
@@ -1,0 +1,53 @@
+package synthetictests
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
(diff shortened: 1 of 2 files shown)


pkg/synthetictests/storage.go
```diff
@@ -0,0 +1,53 @@
...
+			{
+				Name: testName,
+				FailureOutput: &ginkgo.FailureOutput{
+					Output: output,
+				},
+			},
+			// Mark the test as a flake
+			{
+				Name: testName,
+			},
```
dobsonj, line 49: We probably discussed this at one point, but why mark the test as a flake?
Is this just so we can easily find out if we hit quota errors during other tests?

PR #26565, opened 2021-11-03
Disable topology tests for vSphere in-tree storage plugin

bertinatto:
We don't support this setup in our CI.

CC @openshift/storage


bertinatto:
/test e2e-vsphere

```
: [sig-network] services when using OpenshiftSDN in a mode that isolates namespaces by default should allow connections to services in the default namespace from a pod in another namespace on the same node [Suite:openshift/conformance/parallel] 
```

soltysh (approved):
/approve

rvanderp3:
We are still seeing some tests failing.   #26575  has been started to disable those tests until we can address them.

PR #26568, opened 2021-11-04
Fix sno resiliency test

eggfoobar:
Running the `sig-sno serial` test locally was failing due to ginkgo error `You may only call It from within a Describe, Context or When`. Updated test to remove nested `It` call.


ashcrow:
Reassigning as I've changed roles.

eggfoobar:
Closing, tracked in PR #26576

PR #26569, opened 2021-11-04
Add baremetal high availability tests

ardaguclu:
This PR adds new baremetal test, namely
`[sig-installer][Feature:baremetal][Serial] Baremetal platform should ensure cluster baremetal operator and metal3 deployment return back healthy after they are deleted`.

This test will delete cluster baremetal operator and metal3 deployment
sequentially and ensures that baremetalhosts are in correct and healthy
state.

### DIFF
--- a/test/extended/baremetal/high_availability.go
+++ b/test/extended/baremetal/high_availability.go
@@ -1,0 +1,151 @@
+package baremetal
+
+import (
+	"context"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	clusteroperatorhelpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	deleteInterval           = 1 * time.Second
+	deleteWaitTimeout        = 3 * time.Minute
+	restartInterval          = 10 * time.Second
+	restartWaitTimeout       = 10 * time.Minute
+	baremetalNamespace       = "openshift-machine-api"
+	clusterBaremetalOperator = "cluster-baremetal-operator"
+	metal3Deployment         = "metal3"
+)
+
+var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platform should ensure", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLI("baremetal")
+
+	g.It("cluster baremetal operator and metal3 deployment return back healthy after they are deleted", func() {
+		skipIfNotBaremetal(oc)
+
+		c, err := e2e.LoadClientset()
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		g.By("delete cluster baremetal operator")
+		err = c.AppsV1().Deployments(baremetalNamespace).Delete(context.Background(), clusterBaremetalOperator, v1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("wait until cluster baremetal operator is deleted")
+		err = wait.PollImmediate(deleteInterval, deleteWaitTimeout, func() (bool, error) {
+			_, err := c.AppsV1().Deployments(baremetalNamespace).Get(context.Background(), clusterBaremetalOperator, v1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return true, nil
+				}
+
+				e2e.Logf("Error getting cluster baremetal operator deployment: %v", err)
+				return false, nil
+			}
+
+			return false, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("wait until cluster baremetal operator returns back healthy")
+		err = wait.Poll(restartInterval, restartWaitTimeout, func() (bool, error) {
+			dc, err := c.AppsV1().Deployments(baremetalNamespace).Get(context.Background(), clusterBaremetalOperator, v1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return false, nil
+				}
+
+				e2e.Logf("Error getting cluster baremetal operator deployment: %v", err)
+				return false, nil
+			}
+
+			if dc.Status.AvailableReplicas != 1 {
+				return false, nil
+			}
+
+			baremetalCO, err := oc.AdminConfigClient().ConfigV1().ClusterOperators().Get(context.Background(), "baremetal", v1.GetOptions{})
+			if err != nil {
+				e2e.Logf("Error getting baremetal operator: %v", err)
+				return false, nil
+			}
+
+			if clusteroperatorhelpers.IsStatusConditionFalse(baremetalCO.Status.Conditions, configv1.OperatorAvailable) {
+				return false, nil
+			}
+
+			return true, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("delete metal3 deployment")
+		err = c.AppsV1().Deployments(baremetalNamespace).Delete(context.Background(), metal3Deployment, v1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("wait until metal3 deployment is deleted")
+		err = wait.PollImmediate(deleteInterval, deleteWaitTimeout, func() (bool, error) {
+			_, err := c.AppsV1().Deployments(baremetalNamespace).Get(context.Background(), metal3Deployment, v1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return true, nil
+				}
+
+				e2e.Logf("Error getting metal3 deployment: %v", err)
+				return false, nil
+			}
+
+			return false, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("wait until metal3 deployment returns back healthy")
+		err = wait.Poll(restartInterval, restartWaitTimeout, func() (bool, error) {
+			dc, err := c.AppsV1().Deployments(baremetalNamespace).Get(context.Background(), metal3Deployment, v1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return false, nil
+				}
+
+				e2e.Logf("Error getting metal3 deployment: %v", err)
+				return false, nil
+			}
+
+			if dc.Status.AvailableReplicas != 1 {
+				return false, nil
+			}
+
+			return true, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("verify that baremetal hosts are healthy and correct state")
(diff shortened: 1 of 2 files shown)


ardaguclu:
This https://github.com/openshift/release/pull/23296 is prerequisite for this PR to test.

ardaguclu:
@andfasano @elfosardo @bfournie could you please have a look to new test?

andfasano:
/test e2e-metal-ipi

Just to double-check that the tests will not appear in the parallel suite

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,117 @@
...
+
+		c, err := e2e.LoadClientset()
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		g.By("delete cluster baremetal operator")
+		err = c.AppsV1().Deployments(baremetalNamespace).Delete(context.Background(), "cluster-baremetal-operator", v1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("wait until cluster baremetal operator returns back healthy")
+		err = wait.Poll(interval, waitTimeout, func() (bool, error) {
```
andfasano, line 40: I think that In case of a slow environment this single wait loop could miss the delete.
I'd suggest to add first a `wait.PollImmediate` (and checking until `apierrors.IsNotFound`),
and then another `wait.PollImmediate` until the new deployment is found

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,117 @@
...
+		c, err := e2e.LoadClientset()
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		g.By("delete metal3 deployment")
+		err = c.AppsV1().Deployments(baremetalNamespace).Delete(context.Background(), "metal3", v1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("wait until metal3 deployment returns back healthy")
+		err = wait.Poll(interval, waitTimeout, func() (bool, error) {
+			dc, err := c.AppsV1().Deployments(baremetalNamespace).Get(context.Background(), "metal3", v1.GetOptions{})
```
andfasano, line 74: Same here

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,117 @@
...
+			if dc.Status.AvailableReplicas != 1 {
+				return false, nil
+			}
+
+			return true, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("verify that baremetal hosts are healthy and correct state")
+		checkMetal3DeploymentHealthy(oc)
```
andfasano, line 59: I'm not sure about the usefulness of this check in case of a CBO delete

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,117 @@
...
```
andfasano, line 92: Also in this case I don't think the resources are affected after the deletion? In such cases a waiting condition could be required instead of an immediate check

test/extended/baremetal/high_availability.go
ardaguclu: Good point @andfasano

test/extended/baremetal/high_availability.go
ardaguclu: I don't agree. A couple of months ago, there was a bug that caused nodes are stuck in registration error because of a bug in BMO or CBO. I think, it is useful to check that this deletion(I mean both cases) must not have any effect on the bmh statuses. I think, this is even more important than CBO permanently stays in degraded status.

test/extended/baremetal/high_availability.go
ardaguclu: I deliberately did not wrap this with wait condition, because metal3 deployments should not be affected by this deletion.

test/extended/baremetal/high_availability.go
andfasano: It seems different things get mixed, and maybe a different approach could be required. The goal of this test is to verify that CBO returns operative after a deletion. Since the main task of CBO is to deploy a metal3 deployment when missing, one possible way to check that CBO is continuing to operate as normal would be to delete the metal3 deployment _after_ CBO came back, and then wait for it to become ready (and this probably would mean merging the case below here)

ardaguclu: Why do you think different things are mixed because by deleting CBO and only waiting it returns back is testing the CVO nothing more than that. For all baremetal tests, I think we always ensure that BMHs are in correct state(all operators we have exist for baremetalhosts).

But I agree with you, we can boil these two tests down to one test(deleting sequentially) seems like a more correct approach. Delete CBO, wait ready, delete metal3, wait ready, and check bmh statuses?

andfasano: Mainly due the check on the pre-existing resources. Thus the `Delete CBO, wait ready, delete metal3, wait ready, and check bmh statuses` sequence seems fine to me for such test :+1:

ardaguclu: Thanks @andfasano, this test will be a good candidate catching possible errors.

ardaguclu:
/test e2e-metal-ipi-serial

@andfasano what do you think about the latest version?

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,135 @@
...
+
+		g.By("delete cluster baremetal operator")
+		err = c.AppsV1().Deployments(baremetalNamespace).Delete(context.Background(), "cluster-baremetal-operator", v1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("wait until cluster baremetal operator is deleted")
+		err = wait.PollImmediate(deleteInterval, deleteWaitTimeout, func() (bool, error) {
+			_, err := c.AppsV1().Deployments(baremetalNamespace).Get(context.Background(), "cluster-baremetal-operator", v1.GetOptions{})
+			if err != nil {
+				if strings.Contains(err.Error(), `deployments.apps "cluster-baremetal-operator" not found`) {
```
andfasano, line 45: I think `if errors.IsNotFound(err) {` could be used for that

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,135 @@
...
+					return true, nil
+				}
+				return false, err
```
andfasano, line 48: Other types of error shouldn't be considered as a test failure?

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,135 @@
...
+			dc, err := c.AppsV1().Deployments(baremetalNamespace).Get(context.Background(), "cluster-baremetal-operator", v1.GetOptions{})
+			if err != nil {
+				// This error is expected at first run
+				if strings.Contains(err.Error(), `deployments.apps "cluster-baremetal-operator" not found`) {
+					return false, nil
+				}
+				return false, err
+			}
+
+			if dc.Status.AvailableReplicas != 1 {
```
andfasano, line 65: Not sure if it could be worth checking also the co here

test/extended/baremetal/high_availability.go
ardaguclu: Thanks. Done.

test/extended/baremetal/high_availability.go
ardaguclu: Instead failing the test, maybe we should log for all and wait until timeout.

test/extended/baremetal/high_availability.go
ardaguclu: I've added a CO availability check in here, let's see the results.

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,158 @@
...
+					return true, nil
+				}
+
+				e2e.Logf("Error getting cluster baremetal operator deployment: %v", err)
+				return false, nil
+			}
+
+			return false, nil
+		})
+
```
andfasano, line 62: Isn't missing a `o.Expect(err).NotTo(o.HaveOccurred())` here?

ardaguclu: Exactly :), thanks.

ardaguclu:
@andfasano are you ok with the latest version of PR?, thanks.

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,160 @@
...
+		expectStringField(h, "baremetalhost", "status.operationalStatus").To(o.BeEquivalentTo("OK"))
+		if strings.Contains(h.GetName(), "master") {
+			expectStringField(h, "baremetalhost", "status.provisioning.state").To(o.BeEquivalentTo("externally provisioned"))
+			expectBoolField(h, "baremetalhost", "spec.online").To(o.BeTrue())
+		} else {
+			// extra workers are in available status and online field is set to false
+			expectStringField(h, "baremetalhost", "status.provisioning.state").To(
+				o.Or(
+					o.BeEquivalentTo("provisioned"),
+					o.BeEquivalentTo("available")))
```
andfasano, line 157: What's preventing using the same code block used in this [test](https://github.com/openshift/origin/blob/f1cfa98d3dd57179844c0acad5f9fb673c965673/test/extended/baremetal/hosts.go#L136-L146) to verify that the metal3 resources are in a good shape?

ardaguclu: The test you referenced is parallel suite test and there is no extra worker. We can safely ensure that every host is provisioned and online. However, this test is serial, there is an extra worker(status: available, online:false).

ardaguclu: But after our slack discussions, I started thinking that maybe we may not check worker's statuses and even we can remove all status check. What do you think?

andfasano: Here the provisioning state check is fundamental, since it ensures that the restart of the metal3 deployment didn't affect the (already existing) baremetal resources. The main problem is related due the extraworker management, which anyhow we're going to change (thus the logic will become the same of the other tests, ie checking that everything is provisioned or externally provisioned))

ardaguclu: Yes, agree. This PR can wait until extra worker management is finalized. After that, this check will be unified with the parallel test check.

ardaguclu:
This PR is waiting https://github.com/openshift/origin/pull/26681 and will be changed according to it.

ardaguclu:
@andfasano I think, there is no blocker for this PR and we can move forward. What do you think?

ardaguclu:
@bparees could you please take a look at this new metal high availability test?, thanks.

test/extended/baremetal/high_availability.go
```diff
@@ -0,0 +1,151 @@
...
+				return false, nil
+			}
+
+			if dc.Status.AvailableReplicas != 1 {
+				return false, nil
+			}
+
+			baremetalCO, err := oc.AdminConfigClient().ConfigV1().ClusterOperators().Get(context.Background(), "baremetal", v1.GetOptions{})
+			if err != nil {
+				e2e.Logf("Error getting authentication operator: %v", err)
```
bparees, line 81: authentication operator?  or baremetal operator?

ardaguclu: It should definitely be baremetal operator.

bparees:
/approve

fyi @sadasu should be able to approve these:
https://github.com/openshift/origin/blob/master/test/extended/util/annotate/generated/OWNERS
https://github.com/openshift/origin/blob/master/OWNERS_ALIASES#L44

ardaguclu:
Thanks @bparees.

Since the general Centos issue, metal-ipi jobs are failing. Thus, waiting to retest until the problem is fixed.

ardaguclu:
@andfasano all tests passed and I think we just need lgtm label. Could you please label it?, thanks.

PR #26574, opened 2021-11-05
CLI tests: Softcode the kubernetes.default cluster IP

alvaroaleman:
Currently, the tests always fail if a different service ip range is used
which is always the case in Hypershfit clusters.

/assign @soltysh


soltysh (approved):
/lgtm
/approve

PR #26579, opened 2021-11-08
Managed cluster should ensure control plane operator: Allow hypershift kapi proxy

alvaroaleman:
This pod is used in Hypershift to proxy the connection to the kube
apiserver from nodes. As even the kubelet uses it, it has to be a static
pod and that results in a keyless noExecute toleration being added by
the node.
    
### DIFF
--- a/test/extended/operators/tolerations.go
+++ b/test/extended/operators/tolerations.go
@@ -32,6 +32,9 @@ var _ = Describe("[sig-arch] Managed cluster should", func() {
 		excludedNamespaces := sets.NewString("openshift-kube-apiserver", "openshift-kube-controller-manager", "openshift-kube-scheduler", "openshift-etcd", "openshift-openstack-infra", "openshift-ovirt-infra")
 		// exclude these pods from checks
 		whitelistPods := sets.NewString("network-operator", "dns-operator", "olm-operators", "gcp-routes-controller", "ovnkube-master", "must-gather")
+		// The kube-apiserver proxy exists only in Hypershift. It is used to proxy the kubelet->kube-apiserver connection, hence it must be a static pod
+		// which makes the kubelet add a keyless noExecution toleration, so we have to exclude it here.
+		whitelistPods.Insert("kube-apiserver-proxy")
 		for _, pod := range pods.Items {
 			// exclude non-control plane namespaces
 			if !hasPrefixSet(pod.Namespace, namespacePrefixes) {


bparees:
why is this not already a problem?  (presumably because the pods fetched by the test in question don't actually include static pods).

and if it's a new problem specific to hypershift topologies then i'm not understanding why, when does hypershift create static pods that would fail this test?

alvaroaleman:
@bparees thanks for your feedback. The reason this must be a static pod and can not be a workload that lives in the API is that IBM uses it to loadblance the kubelet -> apiserver connection, so it has to be up before the kubelet contacts the apiserver: https://github.com/openshift/hypershift/pull/685#pullrequestreview-806712341

bparees:
i'm still not understanding what is new/novel about the static pods that hypershift is creating, as compared to the static pods that already exist and don't cause this test to fail?

test/extended/operators/tolerations.go
```diff
@@ -31,7 +31,7 @@ var _ = Describe("[sig-arch] Managed cluster should", func() {
 		namespacePrefixes := sets.NewString("kube-", "openshift-")
 		excludedNamespaces := sets.NewString("openshift-kube-apiserver", "openshift-kube-controller-manager", "openshift-kube-scheduler", "openshift-etcd", "openshift-openstack-infra", "openshift-ovirt-infra")
 		// exclude these pods from checks
-		whitelistPods := sets.NewString("network-operator", "dns-operator", "olm-operators", "gcp-routes-controller", "ovnkube-master", "must-gather")
+		whitelistPods := sets.NewString("network-operator", "dns-operator", "olm-operators", "gcp-routes-controller", "ovnkube-master", "must-gather", "kube-apiserver-proxy")
```
bparees, line 34: why add the pod name to the whitelist *and* add logic to generally exclude staticpods?

alvaroaleman: That is wrong, I am happy to do either

alvaroaleman:
> i'm still not understanding what is new/novel about the static pods that hypershift is creating, as compared to the static pods that already exist and don't cause this test to fail?

@bparees nothing, the existing static pods are explicitly excluded instead. I don't really have an opinion on either approach, any favorite?

bparees:
> @bparees nothing, the existing static pods are explicitly excluded instead. I don't really have an opinion on either approach, any favorite?

i think i prefer the explicit whitelist approach, it avoids the risk that we are ignoring pods we shouldn't be.

alvaroaleman:
> i think i prefer the explicit whitelist approach, it avoids the risk that we are ignoring pods we shouldn't be.

Thanks, updated the change to do that.

test/extended/operators/tolerations.go
```diff
@@ -32,6 +32,8 @@ var _ = Describe("[sig-arch] Managed cluster should", func() {
 		excludedNamespaces := sets.NewString("openshift-kube-apiserver", "openshift-kube-controller-manager", "openshift-kube-scheduler", "openshift-etcd", "openshift-openstack-infra", "openshift-ovirt-infra")
 		// exclude these pods from checks
 		whitelistPods := sets.NewString("network-operator", "dns-operator", "olm-operators", "gcp-routes-controller", "ovnkube-master", "must-gather")
+		// The kube-apiserver proxy exists only in Hypershift
```
bparees, line 35: please give this a full comment on why it is being whitelisted (e.g. the summary from your PR)

alvaroaleman: Done

PR #26583, opened 2021-11-09
Set upper bounds to 50 percentile

ingvagabund:
Instead of taking the maximum through all apirequestscount CRs through operators, take median of maximum of all 60 minutes long interval buckets from audit logs. The new approach gives better approximation of the actual maximum.

Currently taking median (50-th percentile) as most of the maximums of apirequestscount falls under it.

I am still mining the percentiles from all `redhat-openshift-ocp-release-4.10-informing` jobs to improve the upper bound/percentiles.

/hold


test/extended/apiserver/api_requests.go
```diff
@@ -109,172 +109,182 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
-				"cluster-monitoring-operator":            51,
-				"cluster-node-tuning-operator":           48,
-				"cluster-samples-operator":               28,
-				"cluster-storage-operator":               187,
-				"console-operator":                       149,
-				"csi-snapshot-controller-operator":       48,
-				"dns-operator":                           41,
-				"etcd-operator":                          142,
-				"gcp-pd-csi-driver-operator":             750,
-				"ingress-operator":                       488,
```
deads2k, line 181: doh. @ingvagabund this one got smaller and we're failing now :(

PR #26586, opened 2021-11-09
WIP- Add puma to s2i-environment-build-app

adambkaplan:
With ruby 2.7, puma is required to run the web server used in the test.


adambkaplan:
/close

For Jenkins tests on OCP 3.11, we won't need this update.

PR #26590, opened 2021-11-10
Revert "add search default image registries test to build e2es"

deads2k:
This reverts commit f66fc8cb2651a6a673cffcc597260a8efda7b297.

This new test is causing a perma-fail in nightly promotion.  We need to get promotion running again to unblock feature and bug delivery.  The unrevert of change can merge *after* it shows a pass on metal-ovn-ipv6.


gabemontero:
according to @dgoodwin https://github.com/openshift/origin/pull/26581/files should address the situation with the nightlies

it makes not sense / we don't want this test to run in disconnected envs

deads2k:
> according to @dgoodwin https://github.com/openshift/origin/pull/26581/files should address the situation with the nightlies
> 
> it makes not sense / we don't want this test to run in disconnected envs

The linked PR ran metal-ovn-ipv6 successfully.

/close

PR #26594, opened 2021-11-11
oauth e2e: fix the expected http2 error to match golang 1.17+

stlaz:
Golang 1.17 made it so that a TLS connection is terminated when
a client with supported ALPN protocols attempts to reach a server
that does not support any ALPN protocols or the clients' ALPN
protocols don't match those of the server.


deads2k:
/lgtm

I want my payload! :)

deads2k:
GCP tests are blocked on another problem, I'm going to take a chance on this.

PR #26602, opened 2021-11-12
pkg/synthetictests: add toleration for missing versioned etcd-endpoints cm during upgrade

hexfusion:
In 4.10 the cluster-etcd operator will begin to version the etcd-endpoints cm resource as part of its static pod resources. During upgrade the installer controller can complain that the resource doesn't exist for the existing version. This can result in a few events as the controller requeues. This issue will resolve with the creation of the next revision which is the result of the upgrade itself.  Thus expected, transient, and benign.

cc @hasbro17  

ref: https://github.com/openshift/cluster-etcd-operator/pull/701#issuecomment-966455818


hasbro17 (approved):
/lgtm

hexfusion:
cc @bparees

PR #26604, opened 2021-11-12
wait for auth cache on secondary apiserver for the  templateinstance …

deads2k:
…impersonate tests

We currently wait for one server to be ok and we have no way to wait for all servers to be ok.  We can build retries in everywhere or we can try to see it settles with an additional 5s wait after the first apiserver reports ok.


adambkaplan (approved):
/lgtm

petr-muller:
Heh, now the jobs got spawned correctly, but the UI is lagging because the status of the PRPQR is badly updated (I think that the job status controller races with itself while attempting to control WHOLE status)

Jobs are there though (and I think the PRPQR will update its status as the jobs are finishing and we get new events to react on): https://prow.ci.openshift.org/?job=openshift-origin-26604-*

PR #26605, opened 2021-11-15
prevent rare alert gathering panic

deads2k:
found in https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.10-e2e-aws-serial/1459871441305473024

This just does its best to fill in something for missing data.  We don't know the actual end, but we know we don't want to fail and that some alert has started.


deads2k:
loki error. This is a serial fix

/override ci/prow/e2e-gcp

PR #26616, opened 2021-11-18
Specify default namespace when getting API endpoint

derekhiggins:
Instead of switching the current context to default
just Specify it in the status command preventing the
need to write out to KUBECONFIG, which may effect
other test reading from the file at the same time.

### DIFF
--- a/test/extended/cli/builds.go
+++ b/test/extended/cli/builds.go
@@ -204,9 +204,7 @@ var _ = g.Describe("[sig-cli] oc builds", func() {
 
 		g.JustBeforeEach(func() {
 			g.By("getting the api server host")
-			err := oc.Run("project").Args("default").Execute()
-			o.Expect(err).NotTo(o.HaveOccurred())
-			out, err := oc.WithoutNamespace().Run("status").Args().Output()
+			out, err := oc.WithoutNamespace().Run("--namespace=default", "status").Args().Output()
 			o.Expect(err).NotTo(o.HaveOccurred())
 			e2e.Logf("got status value of: %s", out)
 			matcher := regexp.MustCompile("https?://.*?443")


derekhiggins:
/hold
e2e-metal-ipi triggered the error case being tested, need to look into it

test/extended/cli/builds.go
```diff
@@ -204,9 +204,7 @@ var _ = g.Describe("[sig-cli] oc builds", func() {
 
 		g.JustBeforeEach(func() {
 			g.By("getting the api server host")
-			err := oc.Run("project").Args("default").Execute()
-			o.Expect(err).NotTo(o.HaveOccurred())
-			out, err := oc.WithoutNamespace().Run("status").Args().Output()
+			out, err := oc.WithoutNamespace().Run("--namespace=default", "status").Args().Output()
```
deads2k, line 207: the `oc project default` switches the project which impacts the process and create later on.  Which namespace are those supposed to be in?

derekhiggins:
```
oc project default
oc status
```
This sequence of commands was being used to get the url of the api server, in the
majority of cases "oc project default" is most likely a NOOP with the negative 
side-effect of leaving KUBECONFIG as a empty file for a brief
period of time (causing other tests to fail).  

```
oc --namespace=default status
```
does the same thing without writing to KUBECONFIG so doesn't effect other tests

derekhiggins: > the oc project default switches the project which impacts the process and create later on. Which namespace are those supposed to be in?

The process and create commands are being run with a namespace e.g.

soltysh (approved):
/lgtm
/approve
/label backport-risk-assessed
/retest-required

test/extended/cli/builds.go
soltysh: What this does can actually be done simpler with either:
- `oc whoami --show-server`
- `oc config view -o jsonpath='{.clusters[*].cluster}'`

PR #26617, opened 2021-11-18
Bug 2023238: Skipping Django Test until bug is fixed

dperaza4dustbit:
This is not fixing the bug it self but unblocking the CI pipeline
as a workaround. The fix for this bug will come from the python
image stream team, most likely, will send another PR to enable.


gabemontero:
image eco failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26617/pull-ci-openshift-origin-master-e2e-gcp-image-ecosystem/1461357681024438272
was unrelated to this change, and the test we are skipping showed up in the skip list

just need to retest until we get past flakes and this merges

/retest

PR #26623, opened 2021-11-19
Local kubeconfig tests: Allow no master nodes when external controlpane topology

alvaroaleman:
External controlplane topology implies there are no master nodes, so
allow that.

### DIFF
--- a/test/extended/apiserver/kubeconfigs.go
+++ b/test/extended/apiserver/kubeconfigs.go
@@ -9,6 +9,7 @@ import (
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 
+	configv1 "github.com/openshift/api/config/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
@@ -29,6 +30,13 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
 	} {
 		kubeconfig := kc
 		g.It(fmt.Sprintf("%q should be present on all masters and work", kubeconfig), func() {
+			// external controlplane topology doesn't have master nodes
+			controlPlaneTopology, err := exutil.GetControlPlaneTopology(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if *controlPlaneTopology == configv1.ExternalTopologyMode {
+				g.Skip("ExternalControlPlaneTopology doesn't have master node kubeconfigs")
+			}
+
 			masterNodes, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{
 				LabelSelector: `node-role.kubernetes.io/master`,
 			})


alvaroaleman:
@bparees could you have a look, please?

test/extended/apiserver/kubeconfigs.go
```diff
@@ -34,7 +35,14 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
 			})
 			o.Expect(err).NotTo(o.HaveOccurred())
 			framework.Logf("Discovered %d master nodes.", len(masterNodes.Items))
-			o.Expect(masterNodes.Items).NotTo(o.HaveLen(0))
+
+			controlPlaneTopology, err := exutil.GetControlPlaneTopology(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			// external controlplane topology doesn't have master nodes
+			if *controlPlaneTopology != configv1.ExternalTopologyMode {
```
bparees, line 43: why not move this check to the top of the test (before trying to get the master nodes) and then Skip out of it?

1) faster skip
2) no risk of it failing due to an issue trying to get the master nodes list when you don't need it anyway
3) if someone adds more logic to this test later that also doesn't work w/ external control planes, it's a non-issue.

bparees: (unless there is also a desire to fundamentally confirm that the 4 kubeconfigs the test client has available work correctly?)

alvaroaleman: The reason is that we share the ExternalTopologyMode with IBM roks where it seems this test passes somehow as I can not find a skip anywhere.Hence the idea was to continue testing this even if we have an external controlplane topology if there are master nodes.

bparees: we sure ROKS is actually running this test and not skipping it in some other way?  it is hard to understand how it would pass.

alvaroaleman: No, not sure I just wanted to err on the side of not breaking anything. But I suppose we can change it again if there really is an issue, so moved it to the top.

test/extended/apiserver/kubeconfigs.go
```diff
@@ -29,6 +30,13 @@ var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] local ku
 	} {
 		kubeconfig := kc
 		g.It(fmt.Sprintf("%q should be present on all masters and work", kubeconfig), func() {
+			// external controlplane topology doesn't have master nodes
+			controlPlaneTopology, err := exutil.GetControlPlaneTopology(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if *controlPlaneTopology != configv1.ExternalTopologyMode {
+				g.Skip("ExtternalControlPlaneToplogy doesn't have master node kubeconfigs")
```
bparees, line 37:
```suggestion
				g.Skip("ExternalControlPlaneTopology doesn't have master node kubeconfigs")
```

bparees:
/approve

fix the nit and i or someone else can lgtm

test/extended/apiserver/kubeconfigs.go
alvaroaleman: Fixed

bparees: (i'm not trying to be a pain, getting this stuff consistent is useful/important when people start grepping the codebase trying to find references to certain strings)

alvaroaleman: Updated

alvaroaleman:
@bparees ptal, I kept the condition the wrong way round from the previous iteration (i.E. go into the conditional block when ControlplaneTopology is not External)

PR #26624, opened 2021-11-19
Allow even greater downtime on Azure single-node upgrade

omertuc:
On AWS single-node upgrade we observe around 10-14% downtime so the threshold is 15%.
On Azure we need a different threshold because we observe around 19-20% downtime.


deads2k:
/lgtm

This isn't a statement that this is good. This is a recognition of the state we're currently in.

PR #26627, opened 2021-11-19
Allow operator-lifecycle-manager-packageserver to not define related objects when external controlplane

alvaroaleman:

/cc @bparees


alvaroaleman:
>  error: unable to extract layer sha256:1a4d86f39503912ca

/retest

PR #26631, opened 2021-11-22
prometheus: adjust telemetry test for remote_write switch

jan--f:
The switch from `telemeter-client` to `remote_write` is up at https://github.com/openshift/cluster-monitoring-operator/pull/1416.

One of these two PRs will have to merge with failing tests in order to get past the chicken & egg situation.


jan--f:
/hold
Hold this for now until some final details are decided.

PR #26637, opened 2021-11-22
Fix 'oc expose can ensure the expose command' for custom service cidrs

alvaroaleman:
Hypershift uses a different service network, which results in the test
failing as it can not create the services because they have clusterIPs
in the wrong cidr.

From all that I can tell the value of those clusterIPs doesn't need to
be passed and is not asserted anywhere, hence just remove it. I've
tested that the test now passes on both standard self-hosted and
hypershift clusters.


soltysh (approved):
Yeah, those are from old test-cmd times.
/lgtm
/approve

PR #26638, opened 2021-11-22
OLM should have imagePullPolicy: Fix when ExternalControlplane

alvaroaleman:
These run in the hypershift management cluster so:
* Add code to get a client for that cluster and find the correct namespace
* Update the tests to check for the controlplanetoplogy and if its external, use the code from the first commit to talk to the correct cluster

### DIFF
--- a/test/extended/util/client.go
+++ b/test/extended/util/client.go
@@ -143,6 +143,32 @@ func NewCLIWithoutNamespace(project string) *CLI {
 	return cli
 }
 
+// NewHypershiftManagementCLI returns a CLI that interacts with the Hypershift management cluster.
+// Contrary to a normal CLI it does not perform any cleanup and it must not be used for any mutating
+// operations. Also contrary to a normal CLI it must be constructed inside an `It` block. This is
+// because retrieval of hypershift management cluster config can fail, but assertions are only
+// allowed inside an It block. AfterEach and BeforeEach are not allowed there though.
+func NewHypershiftManagementCLI(project string) *CLI {
+	kubeconfig, _, err := GetHypershiftManagementClusterConfigAndNamespace()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return &CLI{
+		kubeFramework: &framework.Framework{
+			SkipNamespaceCreation:    true,
+			BaseName:                 project,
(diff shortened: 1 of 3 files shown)


alvaroaleman:
/assign @awgreene
could you have a look, please?

awgreene:
Accidentally added both the approve and lgtm labels
/approve cancel

joelanford:
@alvaroaleman Are these tests applicable to hypershift clusters, and we're only skipping them because the tests don't have the necessary clients?

If so, rather than skipping the tests, can we update the tests such that the management cluster clients are available and used? Or at least make a BZ to remind us to go back and re-enable these tests for hypershift?

test/extended/util/client.go
```diff
@@ -143,6 +143,32 @@ func NewCLIWithoutNamespace(project string) *CLI {
 	return cli
 }
 
+// NewHypershiftManagementCLI returns a CLI that interacts with the Hypershift management cluster.
+// Contrary to a normal CLI it does not perform any cleanup and it must not be used for any mutating
+// operations. Also contrary to a normal CLI it must be constructed inside an `It` block. This is
+// because retrieval of hypershift management cluster config can fail, but assertions are only
+// allowed inside an It block. AfterEach and BeforeEach are not allowed there though.
+func NewHypershiftManagementCLI(project string) *CLI {
```
alvaroaleman, line 151: /assign @bparees 

Sorry I have to bother you again for the changes under test/extended/util (first commit).

I am not particularly happy with this behaving differently than a normal cli with the lack of cleanup, but its a chicken-egg problem, we only know if we need it after checking for the controlplane toplogy, in order to check we have to be in an It block, once we are in an it block we can not register the cleanup anymore. Ginkgo can be pretty frustrating to work with.

As we don't assume tests manipulate anything in the management cluster this seems okayish to me and I didn't manage to come up with a better alternative. Very open to suggestions though if you have any.

bparees: thanks for explaining this and including it in the comment.

presumably there is nothing to actually clean up anyway because this doesn't create a new project?

alvaroaleman: Yeah there is nothing to cleanup, hence it's okayish

PR #26639, opened 2021-11-23
Fix 'should only include cluster daemonsets that have maxUnavailable' for Hypershift

alvaroaleman:
The test allows DaemonSets to ignore this rule if they have a
labelSelector for master nodes. In hypershift there are never master
nodes, hence these DSes don't have a selector for that. Fix the test by
also allowing to ignore the rule if there is just a toleration for
master nodes.

/cc @csrwng

### DIFF
--- a/test/extended/operators/daemon_set.go
+++ b/test/extended/operators/daemon_set.go
@@ -7,10 +7,13 @@ import (
 	"strings"
 
 	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
 	exutil "github.com/openshift/origin/test/extended/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/kube-openapi/pkg/util/sets"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 )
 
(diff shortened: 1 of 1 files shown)


test/extended/operators/daemon_set.go
```diff
@@ -88,3 +86,20 @@ var _ = g.Describe("[sig-arch] Managed cluster", func() {
...
 })
+
+func toleratesOrSelectsMasterNodes(spec corev1.PodSpec, masterReq *labels.Requirement) bool {
+	labelSet := labels.Set(spec.NodeSelector)
+	if !labelSet.AsSelector().Empty() && masterReq.Matches(labelSet) {
+		return true
+	}
+
+	// Hypershift doesn't have master nodes so there is never a selector for master nodes. The toleration still exists
+	// though so grant an exception for that as well.
```
bparees, line 97: tolerating master nodes seems irrelevant if there are no master nodes.

i'm not sure i follow the logic of the original test case, but it seems to be:

1) if you're a daemonset that cannot/does not run on master nodes, you must have a certain maxUnavailable percent.
2) If you're a daemonset that runs on master nodes(via nodeselector), then you don't need to have that maxUnavailable set (presumably because running on masters guarantees you some availability anyway?  it's not really clear from the test code why the exception is granted)

it may also be that the exemption for daemonsets that target the control plane is granted because "10% maxunavailable" is impossible to achieve if you only have 3 control plane nodes and your daemonset only runs on control plane nodes.  (33% should be achievable though).

but just having a toleration of master nodes (especially if there aren't actually going to be any in the cluster) doesn't seem like the same thing.

So before we add this exception, can i get a better explanation of why it's ok to exempt these daemonsets from this test?

alvaroaleman: That we have any system daemonsets that normally have a master labelselector is wrong in the first place, as it results in us running nodecount replicas of it rather than three like in selfhosted.

That being said, today it is unclear how the we will be doing networking in the future for Hypershift so investing time in converting it to a different workload type for Hypershift doesn't seem worthwhile. Additionally, it is only wrong in the sense that it is wasteful but I don't think it results in correctness issues.

So basically want I want here is to disable this test for DS that were originally intended to run on master nodes until we figure out what to do with them which is blocked by the networking question.

alvaroaleman: I created https://issues.redhat.com/browse/HOSTEDCP-279 to track future work on this

bparees: Can you just whitelist the daemonsets in question? It seems like there could be DSes that tolerate masters, while not nodeselecting for them, and those DSes should still be subject to this maxUnavailable rule.  The only DSes that should be exempted from it (based on my admittedly poor understanding of the intent of the original test) are ones that *only* run on masters, because the unavailability expectations are justifiably different if you can only run on a master node, than if you can run on any node, or on any non-master node.

alvaroaleman: Updated the code to the static exception and also added a better description and a link to the jira issue

PR #26642, opened 2021-11-23
double space preventing proper reporting for disruption

deads2k:
This caused the differences found in https://github.com/openshift/ci-tools/pull/2518

### DIFF
--- a/pkg/monitor/api.go
+++ b/pkg/monitor/api.go
@@ -251,9 +251,9 @@ func DisruptionContinuingMessage(locator string, connectionType BackendConnectio
 	case NewConnectionType:
 		return fmt.Sprintf("%s is not responding to GET requests over new connections: %v", locator, err)
 	case ReusedConnectionType:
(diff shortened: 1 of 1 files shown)


deads2k:
this produces an event like https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.10-e2e-gcp-upgrade/1460417551287717888/artifacts/e2e-gcp-upgrade/openshift-e2e-test/artifacts/junit/AdditionalEvents__sig_network_edge_Console_remains_available_via_cluster_frontend_ingress_using_reused_connections_2021-11-16T02:46:07Z.json which does not match https://github.com/openshift/origin/blob/176aeb92845af9eb50b1d0fe8e98a78dee29215e/pkg/monitor/monitorapi/disruption.go#L18

deads2k:
whitespace only change.  We need this to fix the disruption recording, so force merging

PR #26598, opened 2021-11-12
Adjust build test images for multi-arch support

yselkowitz:
This includes a [test image rebase](https://github.com/openshift/origin/blob/master/test/extended/util/image/README.md).

/cc @sttts 
/cc @dmage

PR #26646, opened 2021-11-29
Bug 2037075: Verify Builds with CSI Volume Sources

jkhelil:
https://issues.redhat.com/browse/BUILD-275

### DIFF
--- a/test/extended/builds/volumes.go
+++ b/test/extended/builds/volumes.go
@@ -1,11 +1,18 @@
 package builds
 
 import (
+	"context"
 	"path/filepath"
 	"time"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	kapierrs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	buildv1 "github.com/openshift/api/build/v1"
+	configv1 "github.com/openshift/api/config/v1"
 
 	deploymentutil "github.com/openshift/origin/test/extended/deployments"
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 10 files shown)


test/extended/testdata/builds/volumes/csi-docker-buildconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+    to:
+      kind: ImageStreamTag
+      name: mydockerstream:latest
+  postCommit:
+    script: cat /var/run/secrets/some-secret/key
+  triggers: []
+  runPolicy: Serial
+  source:
+    dockerfile:
+      'FROM registry.ci.openshift.org/ocp/4.8:test-build-simples2i'
```
yselkowitz, line 18: Please note that this image is being relocated for multi-arch support in #26598

jkhelil: ack @yselkowitz

gabemontero:
you have to run `make update` to update the bindata @jkhelil 

you also minimally have to assign @adambkaplan to this PR to get teh approved label per https://github.com/openshift/origin/pull/26646#issuecomment-981671576

test/extended/testdata/builds/volumes/csi-docker-buildconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+          value: "5"
+      volumes:
+      - mounts:
+        - destinationPath: "/var/run/secrets/some-secret"
+        name: my-csi-shared-secret
+        source:
+          csi:
+            driver: csi.sharedresource.openshift.io
+            volumeAttributes:
+              sharedSecret: my-secret
```
gabemontero, line 33: Per my comment to you in https://docs.google.com/document/d/1F3VFkiLUJbsUkFJcjnA2oyROYTI5RRlTJYl4ZWQoNls/edit?disco=AAAASUDEeYE add the `‘refreshResource’ attribute, and set it to a string value of false

gabemontero: Another option as well, which I might like better, is that in your OCM controller PR, if you see CSI driver volume mounts with our shared resource driver here, you adjust their volume attributes to force `refreshResources` to 'false'

WDYT @adambkaplan

test/extended/testdata/builds/volumes/csi-s2i-buildconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+        name: quay.io/redhat-developer/test-build-simples2i:1.2
+      volumes:
+      - mounts:
+        - destinationPath: "/var/run/secrets/some-secret"
+        name: my-csi-shared-secret
+        source:
+          csi:
+            driver: csi.sharedresource.openshift.io
+            volumeAttributes:
+              sharedSecret: my-secret
```
gabemontero, line 33: same refreshResource to false comment

test/extended/testdata/builds/volumes/csi-sharedresourcerolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: RoleBinding
+metadata:
+  name: shared-resource-secret
+  namespace: build-volumes-1
+roleRef:
+  name: shared-resource-secret
+  namespace: build-volume
+subjects:
+- kind: ServiceAccount
+  name: default
```
gabemontero, line 11: builds use the 'builder' service account  ... change default to builder

gabemontero (changes requested):
I'm guessing the ginkgo code of the e2e's is just following the pattern @coreydaley established with his original build volume work 

assign this PR to him as well so he can review these

otherwise I added some shared resource specific comments, along with a question to @adambkaplan that might mean a tweak to what you are doing in your OCM PR

test/extended/testdata/builds/volumes/csi-docker-buildconfig.yaml
jkhelil: @gabemontero I added the refreshResource attribute

test/extended/testdata/builds/volumes/csi-s2i-buildconfig.yaml
jkhelil: I added the refreshResource attribute

test/extended/testdata/builds/volumes/csi-sharedresourcerolebinding.yaml
jkhelil: Updated

test/extended/testdata/builds/volumes/csi-docker-buildconfig.yaml
gabemontero: OK putting a comment to reflect our team discussion yesterday on this after I posted my last set of comments, namely
we are *NOT* going to have the OCM force shared resource CSI volumes to have refreshResource set to false.

That said, it is a supported option we do want to let user opt into.

So I'll need to adjust my request to you here @jkhelil ..... for each of your shared resource test items here, have one where you do not set refreshResource, and then one where you add the refreshsResource: 'false' setting

gabemontero:
/assign @gabemontero 

/assign @coreydaley 
(to minimally review your volume test cases)

/assign @adambkaplan 
(to minimally get the approve label)

see if you can remember @jkhelil to start doing these assigns yourself.

jkhelil:
ack @gabemontero , I do remember, Just not clear for me if I should wait until all tests are green to assign the PR for approval or not

gabemontero:
> ack @gabemontero , I do remember, Just not clear for me if I should wait until all tests are green to assign the PR for approval or not

I would say if there is more than likely that the failed test are non-related flakes, go ahead and assign / start the review process.

test/extended/testdata/builds/volumes/csi-docker-buildconfig.yaml
jkhelil: @gabemontero what is the purpose for adding tests without refreshsResource: 'false', because we know that it shouldnt happen for builds and it will fail(correct me if i am wring about it)

gabemontero: no @jkhelil both scenarios should work
- if the user does not add the `refreshResource` volume attribute, it should work
- if the user does add the `refreshResourece: 'false'` volume attribute, it should work

Hence, I want you to make sure both paths work but having 2 flavors of your various tests here.

jkhelil: @gabemontero can you have a look please  to the changes regarding this last comment

gabemontero: ok I'm good with the mix you have for this @jkhelil

gabemontero: git the approve label from @adambkaplan and see if @coreydaley is OK with the way you added to the volumes tests he did, and we'll go from there

also, if the agnostic-cmd test is the only failure in the tests, and if you examine the tests and provide a comment to use explaining that all the sig-builds test have passed, and none of any remaining flakes are build related, we can use the skip prow command to bypass that one

adambkaplan (changes requested):
@jkhelil we should also have a test on a standard cluster which tries to mount a CSI volume. In this test case, we would expect the build to be created, and then later fail.

adambkaplan:
/retitle Bug 2037075: Verify Builds with CSI Volume Sources

jkhelil:
retrying test flaking, failure is not related to the PR
/test e2e-gcp-upgrade

gabemontero:
sig-builds in the last e2e-aws-single-node jobs all passed, and the flake was unrelated

since that test appears optional (does not have the required tag in tide) going to see if we can skip it

/skip

gabemontero:
yep appears so  ... down to the upgrade job

/retest

gabemontero:
> @jkhelil we should also have a test on a standard cluster which tries to mount a CSI volume. In this test case, we would expect the build to be created, and then later fail.

@jkhelil I still don't see where you added a test of the type @adambkaplan called for above ^^

with such a test, I would expect a tweak in your `JustBeforeEach` logic where you do not skip the test when it is on a non-tech preview cluster and you want to verify that the build fails .... or you may have to move that skip test outside of the `JustBeforeEach` depending on how you can or cannot parameterize that block.

jkhelil:
> > @jkhelil we should also have a test on a standard cluster which tries to mount a CSI volume. In this test case, we would expect the build to be created, and then later fail.
> 
> @jkhelil I still don't see where you added a test of the type @adambkaplan called for above ^^
> 
> with such a test, I would expect a tweak in your `JustBeforeEach` logic where you do not skip the test when it is on a non-tech preview cluster and you want to verify that the build fails .... or you may have to move that skip test outside of the `JustBeforeEach` depending on how you can or cannot parameterize that block.

gabemontero (approved):
works for me

/lgtm

still minimally need an approve from @adambkaplan I believe

test/extended/builds/volumes.go
```diff
@@ -126,3 +133,233 @@ var _ = g.Describe("[sig-builds][Feature:Builds][volumes] build volumes", func()
...
+
+		g.AfterEach(func() {
+			if g.CurrentGinkgoTestDescription().Failed {
+				exutil.DumpPodStates(oc)
+				exutil.DumpConfigMapStates(oc)
+				exutil.DumpPodLogsStartingWith("", oc)
+			}
+		})
+
+		if !techPreviewNoUpgrade {
```
adambkaplan, line 179: I _don't_ think this will work the way you expect it to. I'm pretty positive this breaks how openshift-tests works with ginkgo.

I would use skips in each test instead, and/or separate context/Describe sections for tech preview vs. standard cluster.

See https://github.com/openshift/origin/blob/master/test/extended/README.md

test/extended/builds/volumes.go
```diff
@@ -126,3 +133,233 @@ var _ = g.Describe("[sig-builds][Feature:Builds][volumes] build volumes", func()
...
+				out, err := oc1.Run("rsh").Args("dc/mydockertest", "cat", "/var/run/secrets/some-secret/key").Output()
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(out).To(o.ContainSubstring("cat: /var/run/secrets/some-secret/key: No such file or directory"))
+			})
+		}
+	})
+})
+
+// isNoUpgrade checks if a cluster is a TechPreviewNoUpgrade cluster
+func isNoUpgrade(config *rest.Config) (bool, error) {
```
adambkaplan, line 350: Be explicit that you're checking for `TechPreviewNoUpgrade`. There are other feature sets that end with `NoUpgrade`.

```suggestion
// isTechPreviewNoUpgrade checks if a cluster is a TechPreviewNoUpgrade cluster
func isTechPreviewNoUpgrade(config *rest.Config) (bool, error) {
```

test/extended/builds/volumes.go
```diff
@@ -126,3 +133,233 @@ var _ = g.Describe("[sig-builds][Feature:Builds][volumes] build volumes", func()
...
+	dc, err := dynamic.NewForConfig(config)
```
adambkaplan, line 351: I don't think you need to use the dynamic client. The test framework includes typed clients for all OpenShift APIs.

test/extended/builds/volumes.go
jkhelil: fixed

test/extended/builds/volumes.go
jkhelil: fixed

test/extended/builds/volumes.go
jkhelil: Thank you @adambkaplan , Good catch
Fixed now

gabemontero:
bump @adambkaplan 

are you at least willing to apply the approve label to @jkhelil PR and have others complete the review, or do you think you'll be able to get lgtm level cycles to look at this soon.

other than non related flakes in the agnostic cmd suite,  this appears green

adambkaplan (approved):
/lgtm

PR #26650, opened 2021-11-29
test/extended/util/openshift/clusterversionoperator: Delegate admin-ack ownership

wking:
So folks don't think Oleg is responsible for these, despite his test/extended/util/OWNERS entry.  Per Ben, he was only added as a starting point to deal w/ the fact that every team gets bottlenecked on approval whenever they touch something that impacts generated code.  This commit restructures so that [existing cluster-version operator approvers][1] (or other folks the origin maintainers want to use instead) can update these test-cases without needing to pull in more generic test-framework review and approval.

[1]: https://github.com/openshift/cluster-version-operator/blob/bc972aafe17f22b0f7f9f6552a327196c77dddf8/OWNERS#L3-L7

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,13 @@
+# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#owners_aliases
+
+aliases:
+  cluster-version-operator-test-case-approvers:
+  - LalatenduMohanty
+  - jottofar
+  - vrutkovs
+  - wking
+  cluster-version-operator-test-case-reviewers:
+  - LalatenduMohanty
+  - jottofar
+  - vrutkovs
+  - wking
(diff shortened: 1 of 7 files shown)


OWNERS_ALIASES
```diff
@@ -0,0 +1,13 @@
+# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#owners_aliases
+
+aliases:
+  update-approvers:
+  - LalatenduMohanty
+  - jottofar
+  - vrutkovs
+  - wking
+  update-reviewers:
```
bparees, line 9: shouldn't these names be more specific/clearly point to the team/component?  "update" is pretty vague/likely to be interpreted in other ways

wking: Moved to a `cluster-version-operator-test-case-` prefix in 27d2c285ed -> 802d3898b8.

bparees:
/approve
will let team lgtm

LalatenduMohanty (approved):
/lgtm

PR #26652, opened 2021-11-29
Move shared OWNERS file into generated dir

bparees:


bparees:
@dmage fyi, i think this is what we should have done originally

PR #26655, opened 2021-11-30
Skip oc version test in disconnected

ardaguclu:
`[sig-cli] oc version check kubernetes version matches the io.openshift.build.versions label`
test tries to exec `oc image info` command in master nodes to compare
versions.

However, in disconnected configuration, this command can not access registry by
throwing `network is unreachable`.

This PR skips this test in disconnected.


ardaguclu:
e2e-aws-proxy is also perm failing but it is not a release blocker.

soltysh (approved):
/approve

PR #26656, opened 2021-12-01
Bug 2027929: test/e2e/upgrade/adminack: Poll gates for duration of update

wking:
Picking #26649 back to 4.9, but manually because I don't have #26650 in 4.9.

### DIFF
--- a/test/extended/util/adminack.go
+++ b/test/extended/util/adminack.go
@@ -215,6 +215,9 @@ func setAdminGate(ctx context.Context, gateName string, gateValue string, oc *CL
 	if err != nil {
 		return err
 	}
+	if ackCm.Data == nil {
+		ackCm.Data = make(map[string]string)
+	}
 	ackCm.Data[gateName] = gateValue
 	if _, err := oc.AdminKubeClient().CoreV1().ConfigMaps("openshift-config").Update(ctx, ackCm, metav1.UpdateOptions{}); err != nil {
 		return fmt.Errorf("Unable to update configmap openshift-config/admin-acks: %w", err)
--- a/test/extended/util/adminack.go
+++ b/test/extended/util/adminack.go
@@ -22,6 +22,7 @@ import (
 type AdminAckTest struct {
 	Oc     *CLI
 	Config *restclient.Config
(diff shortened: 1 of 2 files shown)


wking:
/hold

`master`/4.10 will also need to drop the current skip.

wking:
I've pushed 83d48587ba -> ac8b7905c6, rebasing on 4.9's current tip and picking up #26661 and #26668.  The current mapping from 4.10 -> 4.9 commits is:

* 05374fd501 -> ac8b7905c6 Bug 2026806: clusterversionoperator/adminack.go: Check for nil cm map
* 25d7ccd5c8 -> 415403d92f test/extended/util/openshift/clusterversionoperator: Drop waitForUpgradeable
* e5d0eaeda7 -> b186659968 test/extended/util/openshift/clusterversionoperator: Return errors
* 32c8f30e9a -> 8246a6abd2 test/extended/util/openshift/clusterversionoperator: Simplify Fail(f) calls
* 4d1a25aee9 -> 5095fb3ec8 test/e2e/upgrade/adminack: Poll gates for duration of update

No significant manual work, although there were some minor conflicts to resolve, because 4.9 lacks #26650.

test/extended/util/adminack.go
```diff
@@ -36,6 +37,24 @@ var adminAckGateRegexp = regexp.MustCompile(adminAckGateFmt)
 // admin-acks configmap to ack the given admin-ack gate. Once all gates have been ack'ed, the test waits for the
 // Upgradeable condition to change to true.
```
jottofar: Replace with:
`// If the gate has been ack'ed the test verifies that the Upgradeable condition does not complain about the ack. Test`
`// then clears the ack and verifies that the Upgradeable condition complains about the ack. Test then sets the ack`
`// and verifies that the Upgradeable condition no longer complains about the ack.`

wking:
there's still [one nit][1] with the 4.10 test-case, but I think we can punt that one to follow-up work and take this PR back as it stands.

/hold cancel

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2026806#c8

PR #26658, opened 2021-12-01
Fix etcd leader changes are not excessive for Hypershift

alvaroaleman:
This fixes the 'etcd leader changes are not excessive' for Hypershift.
As the etcd runs in the management cluster there, we need to talk to
that cluster and limit our query to not get the management clusters own
etcd metric.

/assign @csrwng

### DIFF
--- a/test/extended/etcd/leader_changes.go
+++ b/test/extended/etcd/leader_changes.go
@@ -9,6 +9,7 @@ import (
 	o "github.com/onsi/gomega"
 	"github.com/prometheus/common/model"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/origin/test/extended/prometheus/client"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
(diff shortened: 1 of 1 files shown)


test/extended/etcd/leader_changes.go
```diff
@@ -17,14 +18,24 @@ var _ = g.Describe("[sig-etcd] etcd", func() {
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLIWithoutNamespace("etcd-leader-change").AsAdmin()
 	g.It("leader changes are not excessive [Late]", func() {
+		controlPlaneTopology, err := exutil.GetControlPlaneTopology(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if *controlPlaneTopology == configv1.ExternalTopologyMode {
```
csrwng, line 23: should you exclude the case that platform is IBM Cloud ?

alvaroaleman: No idea, they shouldn't have that metric inside the guest cluster either so it will just fail differently there now or not?

alvaroaleman: @csrwng does that address the concern?

csrwng: Yes, let's just leave it as is and if needed, it can be updated later to accomodate ROKS

hexfusion:
hypershift controls their own destiny 

/lgtm

PR #26643, opened 2021-11-28
Add serial tests for baremetal resources

bfournie:
Add new serial tests for hostfirmwaresettings:
- A test to confirm invalid settings are handled properly
- A test to change a setting and bring a worker through cleaning to confirm that new setting is reflected in the BMC, using BMC emulator (sushy)

PR #26660, opened 2021-12-01
Bug 2040741: Allow baremetal tests to run on multiple platforms

sadasu:
Metal3 pods can now be deployed on platforms: Baremetal, None,
OpenStack and VSphere.
But, in platforms other than BareMetal, baremetal hosts can be booted only via virtual media.

### DIFF
--- a/test/extended/baremetal/hosts.go
+++ b/test/extended/baremetal/hosts.go
@@ -30,6 +30,55 @@ func skipIfNotBaremetal(oc *exutil.CLI) {
 	}
 }
 
+// Starting from 4.10, metal3 resources could be created in the vSphere, OpenStack and None
+// Platforms in addition to the Baremetal Platform. This method can be used to check for
+// the presence of supported Platforms and also the specific ProvisioningNetwork config
+// supported in non-Baremetal platforms.
+func skipIfUnsupportedPlatformOrConfig(oc *exutil.CLI, dc dynamic.Interface) {
+	g.By("checking supported platforms")
+
+	infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	switch infra.Status.PlatformStatus.Type {
+	case configv1.BareMetalPlatformType:
+		return
+	case configv1.OpenStackPlatformType:
+		fallthrough
+	case configv1.VSpherePlatformType:
+		fallthrough
+	case configv1.NonePlatformType:
+		provisioningNetwork := getProvisioningNetwork(dc)
+		if provisioningNetwork != "Disabled" {
+			e2eskipper.Skipf("Unsupported config in supported platform detected")
+		} else if provisioningNetwork == "" {
+			e2eskipper.Skipf("Unable to read ProvisioningNetwork from Provisioning CR")
+		} else {
+			return
+		}
+	default:
+		e2eskipper.Skipf("No supported platform detected")
+	}
+}
+
+func getProvisioningNetwork(dc dynamic.Interface) string {
+	provisioningGVR := schema.GroupVersionResource{Group: "metal3.io", Resource: "provisionings", Version: "v1alpha1"}
+	provisioningClient := dc.Resource(provisioningGVR)
+	provisioningConfig, err := provisioningClient.Get(context.Background(), "provisioning-configuration", metav1.GetOptions{})
+	if err != nil {
+		return ""
+	}
+	provisioningSpec, found, err := unstructured.NestedMap(provisioningConfig.UnstructuredContent(), "spec")
+	if !found || err != nil {
+		return ""
+	}
+	provisioningNetwork, found, err := unstructured.NestedString(provisioningSpec, "provisioningNetwork")
+	if !found || err != nil {
+		return ""
+	}
+	return provisioningNetwork
+}
+
 func baremetalClient(dc dynamic.Interface) dynamic.ResourceInterface {
 	baremetalClient := dc.Resource(schema.GroupVersionResource{Group: "metal3.io", Resource: "baremetalhosts", Version: "v1alpha1"})
 	return baremetalClient.Namespace("openshift-machine-api")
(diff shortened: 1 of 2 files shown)


ardaguclu:
Maybe we should run openstack and vsphere tests for actual impact;
/e2e-openstack
/e2e-openstack-serial
/e2e-vsphere

ardaguclu:
@sadasu Are we going to be able to run new tests which require extra workers in these platforms?. Because I don't know we'll have extra available workers or should we run those ones only in baremetal?

cc: @andfasano

test/extended/baremetal/hosts.go
```diff
@@ -19,15 +19,48 @@ import (
...
-func skipIfNotBaremetal(oc *exutil.CLI) {
+func skipIfNotSupported(oc *exutil.CLI, dc dynamic.Interface) {
 	g.By("checking platform type")
 
 	infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
-	if infra.Status.PlatformStatus.Type != configv1.BareMetalPlatformType {
-		e2eskipper.Skipf("No baremetal platform detected")
+	provisioningNetwork := getProvisioningNetwork(dc)
```
andfasano, line 28: It seems this could be moved down just for the `NonePlatformType` case, no need to run it everytime

andfasano:
Agree with @ardaguclu . Previous run shown that the changes are ok for the baremetal platform type, but for completeness we should also run them on the newly added platforms to check for regressions (ie to verify also if they require an extra worker as already mentioned)

test/extended/baremetal/hosts.go
sadasu: Yes, I can move it within the switch statement. Please note that the OpenStack and vSpehere cases have a fallthrough which means this will be called for those platforms too.

andfasano:
@sadasu I don't think this approach could work smoothly, considering also all the new pending tests/infrastructure we have currently in progress (ie https://github.com/openshift/origin/pull/26681, https://github.com/openshift/origin/pull/26569 or https://github.com/openshift/origin/pull/26643).
For example, some of tests may require additional extra workers - provided by a dev-scripts deployment, which is the standard for the baremetal platform type but not for the others.
I think that in this case could work better the reversed approach: define a new separated `Describe` block (even in a separate file) to contain all those baremetal tests that makes sense to have in non-baremetal platform (and skip if baremetal in such case). I guess that this also will help in making clear which baremetal features are expected to work in non-baremetal contexts.

ardaguclu:
> @sadasu I don't think this approach could work smoothly, considering also all the new pending tests/infrastructure we have currently in progress (ie #26681, #26569 or #26643). For example, some of tests may require additional extra workers - provided by a dev-scripts deployment, which is the standard for the baremetal platform type but not for the others. 

I agree with @andfasano 
> I think that in this case could work better the reversed approach: define a new separated `Describe` block (even in a separate file) to contain all those baremetal tests that makes sense to have in non-baremetal platform (and skip if baremetal in such case). I guess that this also will help in making clear which baremetal features are expected to work in non-baremetal contexts.

Instead duplicating tests and maintain them in two place, I think we can skip tests we don't want to run on non-baremetal platforms just creating a `SkipForNonBaremetal`, etc.

sadasu:
> @sadasu I don't think this approach could work smoothly, considering also all the new pending tests/infrastructure we have currently in progress (ie #26681, #26569 or #26643). For example, some of tests may require additional extra workers - provided by a dev-scripts deployment, which is the standard for the baremetal platform type but not for the others. 

@andfasano Currently, only one test is being enabled for multiple platforms in addition to running it on the Baremetal platform. And that is to check if the metal3 pod is running. 

I think that in this case could work better the reversed approach: define a new separated `Describe` block (even in a separate file) to contain all those baremetal tests that makes sense to have in non-baremetal platform (and skip if baremetal in such case). I guess that this also will help in making clear which baremetal features are expected to work in non-baremetal contexts.
At this time (in 4.10) we have almost no information about the functionality that running metal3 would enable in the other platforms. And I suspect it would be different for each additional platform we are supporting. Validating that specific platform specific use case is not the goal here. Those platform specific tests should be added in the CI tests of those platforms. 
IMHO, the goal of the CI tests in the metal-ipi CI jobs should be to make sure that changes done for the metal platform should not prevent metal3 from running on these additional platforms.

andfasano:
> At this time (in 4.10) we have almost no information about the functionality that running metal3 would enable in the other platforms. And I suspect it would be different for each additional platform we are supporting. Validating that specific platform specific use case is not the goal here. Those platform specific tests should be added in the CI tests of those platforms. IMHO, the goal of the CI tests in the metal-ipi CI jobs should be to make sure that changes done for the metal platform should not prevent metal3 from running on these additional platforms.

Ok, I understand that this lack of information currently does not allow a super-detailed planning, and the current minimal goal it's just to ensure that the metal3 deployment it's up and running, thus I'd reasonably expect in the longer term to have a (probably small) subset of tests applied also to non-baremetal platforms (in addition to the one currently enabled).
Still, it should be clear for the moment - at least for the baremetal test author - how to correctly apply a skipping condition, so I'd suggest to add some comments on the new skipping function (to make it clear when it should be used) and to rename it to something more speaking (ie `SkipIfNotBaremetalExtended`)

andfasano:
@sadasu I saw that for the openstack and vsphere jobs the test was skipped because the provisioning network is not disabled:

```
<testcase name="[sig-installer][Feature:baremetal] Baremetal/OpenStack/vSphere/None platforms have a metal3 deployment [Suite:openshift/conformance/parallel]" time="2.3">
<skipped message="skip [github.com/openshift/origin/test/extended/baremetal/hosts.go:53]: Unsupported config in supported platform detected"/>
<system-out>
</system-out>
```

Is this something expected given the current job configurations?

sadasu:
> @sadasu I saw that for the openstack and vsphere jobs the test was skipped because the provisioning network is not disabled:
> 
> ```
> <testcase name="[sig-installer][Feature:baremetal] Baremetal/OpenStack/vSphere/None platforms have a metal3 deployment [Suite:openshift/conformance/parallel]" time="2.3">
> <skipped message="skip [github.com/openshift/origin/test/extended/baremetal/hosts.go:53]: Unsupported config in supported platform detected"/>
> <system-out>
> </system-out>
> ```
> 
> Is this something expected given the current job configurations?

Yes, this is expected. The Provisioning resource is not created in the vSphere and OpenStack platforms yet and so, this test would be skipped. When the above platforms are ready to bring up the metal3 components, they can do that by creating the resource and this test will make sure that any changes we make for the metal platform doesn't fundamentally break them.

PR #26661, opened 2021-12-02
Bug 2026806: Admin ack ignore unrelated upgradeable false

wking:
Two fairly unrelated cleanup commits, while I was getting my bearings in this code, and then the real fix which drops `waitForUpgradeable`.

Since we began polling in 4d1a25aee9 (#26649), there's an increased chance that this test is poking at acks while something else is going on in `Upgradeable` in parallel.  For [example][1]:

    Dec  1 21:01:44.722: FAIL: Error while waiting for Upgradeable to go true, err=timed out waiting for the condition Upgradeable: Status=False, Reason=KubeletMinorVersion_KubeletMinorVersionUnsupportedNextUpgrade, Message="Cluster operator kube-apiserver should not be upgraded between minor versions: KubeletMinorVersionUpgradeable: Kubelet minor versions on 6 nodes will not be supported in the next OpenShift minor version upgrade.".
    
With the incoming logic, this test-case would say "great, nothing in `Upgradeable` complaining about any acks", without getting hung up on the fact that `Upgradeable` is complaining about something else (in this case, Kubelet skew, which will shortly resolve once the machine-config gets around to updating nodes and catches them up with the API-server).

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.7-to-4.8-to-4.9-to-4.10-ci/1466135844220833792#1:build-log.txt%3A866

### DIFF
--- a/test/extended/util/openshift/clusterversionoperator/adminack.go
+++ b/test/extended/util/openshift/clusterversionoperator/adminack.go
@@ -59,18 +59,18 @@ func (t *AdminAckTest) Test(ctx context.Context) {
 func (t *AdminAckTest) test(ctx context.Context, exercisedGates map[string]struct{}) {
 	exists := struct{}{}
 
-	gateCm, errMsg := getAdminGatesConfigMap(ctx, t.Oc)
-	if len(errMsg) != 0 {
-		framework.Fail(errMsg)
+	gateCm, err := getAdminGatesConfigMap(ctx, t.Oc)
+	if err != nil {
+		framework.Fail(err.Error())
 	}
 	// Check if this release has admin ack functionality.
 	if gateCm == nil || (gateCm != nil && len(gateCm.Data) == 0) {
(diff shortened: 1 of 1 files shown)


jottofar:
These changes LGTM but I have question about this line:

https://github.com/openshift/origin/blob/25d7ccd5c82edc9759a3a30b4e214c8e31c9aa81/test/extended/util/openshift/clusterversionoperator/adminack.go#L78

This indicates the gate is a duplicate which I think is grounds to fail the test.

jottofar:
> These changes LGTM but I have question about this line:
> 
> https://github.com/openshift/origin/blob/25d7ccd5c82edc9759a3a30b4e214c8e31c9aa81/test/extended/util/openshift/clusterversionoperator/adminack.go#L78
> 
> This indicates the gate is a duplicate which I think is grounds to fail the test.

As discussed with @wking, this is keeping track across test calls. If dup's are possible unit test would be a better place I believe to check so this is resolved.

PR #26674, opened 2021-12-08
ensure that we always produce output for successful pod sandbox tests

deads2k:


deads2k:
will force merge on a green e2e that shows these tests present.

PR #26675, opened 2021-12-08
collapse service LB with PDB into standard disruption check

deads2k:
This is a precursor to fixing the disruption sampling.


deads2k:
replaced by https://github.com/openshift/origin/pull/26703

PR #26676, opened 2021-12-08
Test baremetal deployment without provisioning network

elfosardo:


elfosardo:
waiting on #26681

test/extended/baremetal/hosts.go
```diff
@@ -187,3 +188,37 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
 		o.Expect(found).To(o.BeFalse())
 	})
 })
+
+var _ = g.Describe("[sig-installer][Feature:baremetal] A baremetal deployment without a provisioning network should", func() {
```
ardaguclu, line 192: Maybe we can move this under `baremetal platform should` tests as `g.It`?

test/extended/baremetal/hosts.go
```diff
@@ -187,3 +188,37 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
...
+
+	oc := exutil.NewCLI("baremetal")
+	skipIfNotBaremetal(oc)
+	dc := oc.AdminDynamicClient()
+	skipIfProvisioningNetworkSet(dc)
+
+	g.By("show the Provisioning Network as 'Disabled'")
+	o.Expect(getProvisioningNetwork(dc)).To((o.BeEquivalentTo("Disabled")))
+
+	g.It("not allow setting the ProvisioningNetwork to \"Managed\" with invalid values", func() {
```
ardaguclu, line 203: nit: \`not allow setting the ProvisioningNetwork to "Managed" with invalid values\`?

ardaguclu (commented):
This PR looks good to me(and it also covers most of part of what I'm currently implementing).

Maybe you forgot to run `make update`?

test/extended/baremetal/common.go
```diff
@@ -58,9 +58,26 @@ func skipIfUnsupportedPlatformOrConfig(oc *exutil.CLI, dc dynamic.Interface) {
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	provisioningSpec, _, err := unstructured.NestedMap(provisioningConfig.UnstructuredContent(), "spec")
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	provisioningNetworkCIDR, _, err := unstructured.NestedString(provisioningSpec, "provisioningNetworkCIDR")
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	if provisioningNetworkCIDR != "" {
+		e2eskipper.Skipf("Provisioning Network set, skipping")
```
andfasano, line 75: nit: "skipping" could be avoided here

test/extended/baremetal/hosts.go
andfasano: Wouldn't be safer to move this test under the `serial` suite? If the update action will have success - by mistake - it could create some disruption for the subsequent baremetal tests

test/extended/baremetal/hosts.go
elfosardo: mmm I need to escape the double quotes here, or I can change that to single quotes

ardaguclu: I mean, if you use "`" instead double quotes, you don't need to escape double quotes.

elfosardo: oh, right! ok

test/extended/baremetal/hosts.go
elfosardo: this is probably a good idea

test/extended/baremetal/hosts.go
```diff
@@ -187,3 +187,80 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
...
 })
+
+var _ = g.Describe("[sig-installer][Feature:baremetal] A baremetal deployment without a provisioning network should", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc = exutil.NewCLI("baremetal")
+	)
+
+	g.It("show the Provisioning Network as 'Disabled'", func() {
```
ardaguclu, line 198: I think this might another `g.By` in below test instead a new test.

test/extended/baremetal/hosts.go
elfosardo: should I move the entire test block inside the block at L154 or just add the Serial notation here ?

andfasano: That block is mostly required in case your test would need an extra worker to work with, otherwise it seems fine to just add the serial tag (it will save some time by avoiding to deploy the extra worker)

test/extended/baremetal/common.go
```diff
@@ -58,9 +58,26 @@ func skipIfUnsupportedPlatformOrConfig(oc *exutil.CLI, dc dynamic.Interface) {
...
+	g.By("checking Provisioning Network CIDR")
+
+	provisioningClient := provisioningClient(dc)
+	provisioningConfig, err := provisioningClient.Get(context.Background(), "provisioning-configuration", metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	provisioningSpec, _, err := unstructured.NestedMap(provisioningConfig.UnstructuredContent(), "spec")
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	provisioningNetworkCIDR, _, err := unstructured.NestedString(provisioningSpec, "provisioningNetworkCIDR")
```
ardaguclu, line 71: Does `provisioningNetworkCIDR` always exist even if it is null?. If no, maybe also worth to check bool value?

test/extended/baremetal/hosts.go
```diff
@@ -187,3 +187,72 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
...
+		err = unstructured.SetNestedField(provisioning.Object, validProvisioningNetworkCIDR, "spec", "provisioningNetworkCIDR")
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		err = unstructured.SetNestedField(provisioning.Object, validProvisioningIP, "spec", "provisioningIP")
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		err = unstructured.SetNestedField(provisioning.Object, "Managed", "spec", "provisioningNetwork")
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		_, err = provisioningClient.Update(context.Background(), &provisioning, metav1.UpdateOptions{})
```
ardaguclu, line 254: We are changing `provisioningIP` and `provisioningNetwork` but we are not reverting back after test is completed. Is it safe for the other tests?

ardaguclu (commented):
/test e2e-metal-ipi-serial

test/extended/baremetal/common.go
elfosardo: after checking again, it looks like the provisioningNetworkCIDR entry should be absent in case of a Disabled network, so we can just set the bool and not even the value

test/extended/baremetal/hosts.go
elfosardo: I guess we could move the provisioningNetwork back to Disabled ?

ardaguclu: Yes, I think so. Otherwise, other tests might be affected.

test/extended/baremetal/common.go
ardaguclu: I don't know is it meaningful `provisioningNetworkCIDR` can exist but it is empty

test/extended/baremetal/common.go
```diff
@@ -91,6 +107,11 @@ func preprovisioningImagesClient(dc dynamic.Interface) dynamic.ResourceInterface
 	return ppiClient.Namespace("openshift-machine-api")
 }
 
+func provisioningClient(dc dynamic.Interface) dynamic.ResourceInterface {
+	pClient := dc.Resource(schema.GroupVersionResource{Group: "metal3.io", Resource: "provisionings", Version: "v1alpha1"})
+	return pClient.Namespace("openshift-machine-api")
```
ardaguclu, line 112: I used without namespace in here and it worked https://github.com/openshift/origin/pull/26832/files#diff-aa7a742599cf4f1743a327027b6221c693e5dfedd562ad3d9c7dd50659ca331aR96.

I think `provisionings`'s scope is cluster wide and no need to use namespace in here.

test/extended/baremetal/common.go
```diff
@@ -58,9 +58,25 @@ func skipIfUnsupportedPlatformOrConfig(oc *exutil.CLI, dc dynamic.Interface) {
...
+	g.By("checking Provisioning Network CIDR")
+
+	provisioningClient := provisioningClient(dc)
+	provisioningConfig, err := provisioningClient.Get(context.Background(), "provisioning-configuration", metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	provisioningSpec, _, err := unstructured.NestedMap(provisioningConfig.UnstructuredContent(), "spec")
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	_, provisioningNetworkCIDRfound, err := unstructured.NestedString(provisioningSpec, "provisioningNetworkCIDR")
```
ardaguclu, line 71: Instead getting `provisioningSpec` first and after `provisioningNetworkCIDR`, maybe we can do this in one step like in here https://github.com/openshift/origin/blob/d335c644c13b920b3bc9ab404133788ebd76856a/test/extended/baremetal/helper.go#L184

ardaguclu (commented):
/test e2e-metal-ipi-serial

test/extended/baremetal/hosts.go
```diff
@@ -187,3 +187,81 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
...
+	g.It("show the Provisioning Network as 'Disabled'", func() {
+		skipIfNotBaremetal(oc)
+
+		dc := oc.AdminDynamicClient()
+
+		skipIfProvisioningNetworkSet(dc)
+
+		o.Expect(getProvisioningNetwork(dc)).To((o.BeEquivalentTo("Disabled")))
+
+		g.By("Not allow setting the ProvisioningNetwork to 'Managed' with invalid values")
```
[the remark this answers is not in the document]
elfosardo: I set this as a continuation of the same case as it uses the same parameters
I could separate it, but it would just duplicate the authentication part and leave only L205 to the precedent test

PR #26677, opened 2021-12-08
Fix TestFIPS on Hypershift

alvaroaleman:
There are two changes needed:
* Don't test master nodes, as there are none
* Don't execute the command in the machine-config-daemon, but in an oc
  debug pod. The only reason the machine-config-daemon was originaly
  used is convenience, the test command can be executed in any pod as
  long as its on the right node

Ref regarding "can be executed in any pod": https://coreos.slack.com/archives/C999USB0D/p1638981587077200?thread_ts=1638981423.077100&cid=C999USB0D

/assign @bparees

### DIFF
--- a/test/extended/security/fips.go
+++ b/test/extended/security/fips.go
@@ -8,6 +8,7 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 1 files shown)


alvaroaleman:
Jiraref https://issues.redhat.com/browse/HOSTEDCP-257

alvaroaleman:
/hold
failure looks genuine, someting with the new podsecurity feature?

alvaroaleman:
Okay, the FIPS test passed now, could you have another look @bparees ?

PR #26678, opened 2021-12-08
Hypershift: Skip Prometheus should have important platform topology metrics

alvaroaleman:
None of these metrics exist there, as they come out of the apiserver
operator, require someone to manually install the cluster or are based
on etcd metrics.

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/client-go/rest"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	"k8s.io/kubernetes/test/e2e/framework/skipper"
 	"k8s.io/kubernetes/test/e2e/framework/statefulset"
 	"k8s.io/kubernetes/test/utils/image"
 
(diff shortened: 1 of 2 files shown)


test/extended/util/framework.go
```diff
@@ -2055,3 +2056,11 @@ func GetHypershiftManagementClusterConfigAndNamespace() (string, string, error)
 
 	return hypershiftManagementClusterKubeconfig, hypershiftManagementClusterNamespace, nil
 }
+
+func SkipIfExternalControlplaneTopolocy(oc *CLI, reason string) {
```
bparees, line 2060:
```suggestion
func SkipIfExternalControlplaneTopology(oc *CLI, reason string) {
```

alvaroaleman: Fixed. I will look for a vim spelling plugin that can deal with camelcase :upside_down_face:

test/extended/util/framework.go
```diff
@@ -2055,3 +2056,11 @@ func GetHypershiftManagementClusterConfigAndNamespace() (string, string, error)
 
 	return hypershiftManagementClusterKubeconfig, hypershiftManagementClusterNamespace, nil
 }
+
+func SkipIfExternalControlplaneTopology(oc *CLI, reason string) {
+	controlPlaneTopology, err := GetControlPlaneTopology(oc)
+	o.Expect(err).NotTo(o.HaveOccurred())
```
bparees, line 2062:
```suggestion
	o.ExpectWithOffset(1,err).NotTo(o.HaveOccurred())
```

that way the failure gets reported as coming from the test in question and not the helper function

https://pkg.go.dev/github.com/onsi/gomega#ExpectWithOffset

alvaroaleman: Updated

PR #26679, opened 2021-12-08
separate disruption sampling from the monitor sampler

deads2k:
the existing monitor sampler is odd.  On the one hand, it appears to be trying to provide even samples with direct recording to the monitor.  But it also attempts to have the monitor itself poll.  It seems that we actually used *both* of these techniques at the same time.

In addition, it doesn't provide even sampling over time (the sample func can skew the disruption calculation). 

The simplest thing to do was simply divorce the disruption from the sampler.  I think there are now zero uses of sampler now and ultimately it's likely easier to have the in-memory sampler forever work as a push model.  Which also obviates the need for an actual sampler.  So perhaps we simply delete it in a followup.

/assign @dgoodwin

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -95,6 +95,9 @@
     }
 
     function isEndpointConnectivity(eventInterval) {
+        if (!eventInterval.message.includes("stopped responding to GET requests")){
+            return false
+        }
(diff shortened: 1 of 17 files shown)


deads2k:
replaced by https://github.com/openshift/origin/pull/26682

PR #26680, opened 2021-12-09
baremetal: add test helper

andfasano:
This patch adds the BaremetalTestHelper type (and also refactors a little bit the code) to start using the feature introduced in https://github.com/openshift-metal3/dev-scripts/pull/1315


andfasano:
Addressed by #26681

PR #26681, opened 2021-12-09
Bug 2035534: Baremetal skip inspection

andfasano:
This patch adds a new test for the baremetal ipi platform, in order to verify that the inspection is skipped. Also, it adds the BaremetalTestHelper type (and also refactors a little bit the code) to start using the feature introduced in openshift-metal3/dev-scripts#1315.
The existing scale test has been also retouched to adopt the new add worker mechanism in case of baremetal ipi platform


andfasano:
/hold

Requires https://github.com/openshift/origin/pull/26680

andfasano:
cc @ardaguclu @bfournie @elfosardo

andfasano:
/hold cancel

Dependency on #26680 has been removed

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+	o.Expect(err).ToNot(o.HaveOccurred())
+
+	yamlData, ok := ew.Data[b.extraWorkerKey(index)]
+	o.Expect(ok).To(o.BeTrue(), fmt.Sprintf("unable to find data for extra worker %d", index))
+
+	jsonData, err := yaml.YAMLToJSON(yamlData)
+	o.Expect(err).ToNot(o.HaveOccurred())
+
+	var host unstructured.Unstructured
+	err = host.UnmarshalJSON(jsonData)
```
ardaguclu, line 93: Isn't there UnmarshallYaml function and that's why we are using `YAMLToJSON`?

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+
+// WaitForProvisioningState waits for the given baremetal host to reach the specified provisioning state.
+// If successfull, returns the updated host resource, otherwise the test will fail
+func (b *BaremetalTestHelper) WaitForProvisioningState(host *unstructured.Unstructured, expectedProvisioningState string) *unstructured.Unstructured {
+
+	hostName := host.GetName()
+	var newHost *unstructured.Unstructured
+
+	g.By(fmt.Sprintf("wait until %s becomes %s", hostName, expectedProvisioningState))
+	err := wait.PollImmediate(5*time.Second, 2*time.Minute, func() (done bool, err error) {
```
ardaguclu, line 155: Isn't 2 minutes low for that?

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (b *BaremetalTestHelper) Setup() {
+
+	//This code will ensure that any previous extra worker will be cleaned up
+	//This should make easier the migration of the serial metal ipi jobs
+	//for the adoption of the new way of deployng extra worker
+	hosts, err := b.bmcClient.List(context.Background(), metav1.ListOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	for _, host := range hosts.Items {
```
ardaguclu, line 187: Clean up is already done in here https://github.com/openshift/origin/pull/26681/files#diff-7e730e2c7566572f6d8e0ef47a5b3ca546de7f0d1bd9f0468ff6bed4064a037cR140. Moreover, it waits deletion is completed. 

Is there any possibility that is executed?

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+			return false, err
+		}
+		return false, nil
+	})
+	o.Expect(err).NotTo(o.HaveOccurred())
+}
+
+// WaitForProvisioningState waits for the given baremetal host to reach the specified provisioning state.
+// If successfull, returns the updated host resource, otherwise the test will fail
+func (b *BaremetalTestHelper) WaitForProvisioningState(host *unstructured.Unstructured, expectedProvisioningState string) *unstructured.Unstructured {
```
ardaguclu, line 149: It looks like there is no point that calling `CreateExtraWorker` and not calling this function. Maybe, we can use it in `CreateExtraWorker` instead giving call responsibility to test owner?

ardaguclu (commented):
Thanks @andfasano. These changes look great. I left some comments and I have one question;

We have to wait changes in `Machine Scale` test to merge this PR?, because it still assumes that machine is available.

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+
+// BaremetalTestHelper is an helper class for the baremetal tests,
+// providing support for the most common operations perfomed by
+// the tests. It will also help in reducing the noise in the test
+// definition by hiding technical details
+type BaremetalTestHelper struct {
+	clientSet *kubernetes.Clientset
+	bmcClient dynamic.ResourceInterface
+
+	hostsCreated []*unstructured.Unstructured
```
bfournie, line 31: Seems that this should be called extraWorkers

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+
+		err := b.bmcClient.Delete(context.Background(), hostName, metav1.DeleteOptions{})
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		b.waitForDeletion(host)
+	}
+}
+
+func (b *BaremetalTestHelper) waitForDeletion(host *unstructured.Unstructured) {
+	err := wait.PollImmediate(1*time.Second, 1*time.Minute, func() (done bool, err error) {
```
bfournie, line 133: Maybe could use o.Eventually() here to be more consistent with gomega

test/extended/baremetal/helper.go
bfournie: Same comment with using o.Eventually()

test/extended/baremetal/helper.go
andfasano: Yes

test/extended/baremetal/helper.go
andfasano: I didn't want to overengineer the API of the helper too much for the moment, and considered that the WaitForProvisioningState could be a more generic function useful for other tests (and other scenarios). I expect we're going to refine the interface as long as we're using it in more tests

test/extended/baremetal/helper.go
andfasano: This block is in reality independent, and just required to support the migration, given that there could be other tests using the APPLY_EXTRA_WORKERS feature. Once we'll get rid of it, we'll remove it

test/extended/baremetal/helper.go
andfasano: @bfournie I tried to use `Eventually()` but for this case it doesn't seem a better choice than `wait.PollIImmediate`: the comparison (on the provisioning state) must be moved outside the block, and also the timeouts must be specified at the end. In addition, the error check is present but must be positional (without the possibility to specify a message if required).
So the entire logical flow appeared a little bit strange, while in the current shape is anyhow pretty linear (also, `wait.poll...` seems to be a used more frequently in the repo rather than `Eventually`)

bfournie: OK, I actually like the timeouts being specified at the end, but not being able to specify an error message seems like a limitation. Anyway, thanks for checking it out.

andfasano:
/hold

Found a problem during the `Setup` cleanup logic, fixing it

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+	yamlData, ok := ew.Data[b.extraWorkersSecretKey(index)]
+	o.Expect(ok).To(o.BeTrue(), fmt.Sprintf("unable to find data for extra worker %d", index))
+
+	var secret v1.Secret
+	err = yaml.Unmarshal(yamlData, &secret)
+	o.Expect(err).ToNot(o.HaveOccurred())
+	return &secret
+}
+
+// getExtraWorkerSecretData gets and decodes the baremetal host associated for the
```
bfournie, line 81: Nit - comment should be `getExtraWorkerData`

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+	host, err = b.bmcClient.Create(context.Background(), host, metav1.CreateOptions{})
+	o.Expect(err).ToNot(o.HaveOccurred())
+
+	b.extraWorkers = append(b.extraWorkers, host)
+
+	return host, secret
+}
+
+// DeleteAllExtraWorkers deletes all the extra workers created in the current session
+func (b *BaremetalTestHelper) DeleteAllExtraWorkers() {
```
bfournie, line 122: Does anything need to be done to delete the secret that was created for the worker?

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (b *BaremetalTestHelper) waitForDeletion(host *unstructured.Unstructured) {
+	err := wait.PollImmediate(1*time.Second, 1*time.Minute, func() (done bool, err error) {
+		_, err = b.bmcClient.Get(context.TODO(), host.GetName(), metav1.GetOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return true, nil
+			}
+
+			return false, err
```
bfournie, line 139: Just curious, this will retry every second, so if there are any problems accessing the ApiServer on any of the retries the test will fail. It sounds like this will be fragile, but I guess this is how other tests work?

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,198 @@
...
+	for _, worker := range b.extraWorkers {
+		err := b.bmcClient.Delete(context.Background(), worker.GetName(), metav1.DeleteOptions{})
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		b.waitForDeletion(worker)
+	}
+}
+
+func (b *BaremetalTestHelper) waitForDeletion(host *unstructured.Unstructured) {
+	err := wait.PollImmediate(1*time.Second, 1*time.Minute, func() (done bool, err error) {
```
bfournie, line 132: Would be good to have a log message here, similar to waitForProvisioningState()

test/extended/baremetal/hosts.go
```diff
@@ -204,14 +109,53 @@ var _ = g.Describe("[sig-installer][Feature:baremetal] Baremetal platform should
...
+		hostData, secretData := helper.GetExtraWorkerData(0)
+
+		// Set inspection annotation as disabled
+		unstructured.SetNestedField(hostData.Object, "disabled", "metadata", "annotations", "inspect.metal3.io")
+
+		// Deploy extra worker and wait
+		host, _ := helper.CreateExtraWorker(hostData, secretData)
+		host = helper.WaitForProvisioningState(host, "available")
+
+		g.By("Check hardware to be empty")
```
bfournie, line 156: Maybe change log msg - "Check that hardware field in status is emptry"

test/extended/baremetal/hosts.go
```diff
@@ -204,14 +109,53 @@ var _ = g.Describe("[sig-installer][Feature:baremetal] Baremetal platform should
...
+		_, found, err := unstructured.NestedString(host.Object, "status", "hardware")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(found).To(o.BeFalse())
```
bfournie, line 159: So hardware is a pointer to a HardwareDetails struct, shouldn't there be a check for the value being nil instead of it being found?

test/extended/baremetal/commons.go
```diff
@@ -0,0 +1,109 @@
...
+		o.Expect(ok).To(o.BeTrue())
+		if t == condType {
+			s, ok := condition["status"]
+			o.Expect(ok).To(o.BeTrue())
+			o.Expect(s).To(o.Equal(condStatus))
+		}
+	}
+}
+
+func getField(host unstructured.Unstructured, resource string, nestedField string, fieldGetter FieldGetterFunc) string {
```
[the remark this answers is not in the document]
andfasano: there's a name clash with the inner func but agree

test/extended/baremetal/helper.go
andfasano: Secrets will be automatically garbage collected when the related host is deleted, so no need to explicitly manage them. Anyhow I found a case where the secret was not yet deleted, and the subsequent test started, and since this could cause a problem (the resource already exist) better to wait for their deletion as well

test/extended/baremetal/helper.go
andfasano: In theory it's right, but  if we have issues in the apiserver in a serial test (so nothing else is running) then the situation is probably even worse :). Another strategy could be to filter other kinds of errors, it's probably something we could tweak in future in case of need (other tests adopted also the approach to watch for a particular event, but this one seems better to me)

test/extended/baremetal/hosts.go
andfasano: We don't have a structured type, so if not present it does not get unmarshalled at all (and thus the need to check the bool flag)

test/extended/baremetal/helper.go
andfasano: Yep, good idea

elfosardo (approved):

ardaguclu:
`e2e-metal-ipi-serial` failed due to the test in this PR, it looks like some fix needed.

andfasano:
> `e2e-metal-ipi-serial` failed due to the test in this PR, it looks like some fix needed.

Yep, still working on it ([hold](https://github.com/openshift/origin/pull/26681#issuecomment-990988074) not yet removed)

andfasano:
Requires https://github.com/metal3-io/baremetal-operator/pull/1054 (otherwise the deletion of extraworkers will intermittently fail)

bfournie:
Seeing failure here which shouldn't be affected by changes:
```
Failing tests:

[sig-cluster-lifecycle][Feature:Machines][Serial] Managed cluster should grow and decrease when scaling different machineSets simultaneously [Suite:openshift/conformance/serial]
```

Also looks like a bugzilla needs to be added.

test/extended/baremetal/helper.go
andfasano: Added `DeployExtraWorker`

test/extended/machines/scale.go
```diff
@@ -186,6 +187,35 @@ func scaleMachineSet(name string, replicas int) error {
...
+		cfg, err := e2e.LoadConfig()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		c, err = e2e.LoadClientset()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		dc, err = dynamic.NewForConfig(cfg)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		// For baremetal platforms, an extra worker must be previously
+		// deployed to allow subsequent scaling operations
+		helper = bmhelper.NewBaremetalTestHelper(dc)
```
ardaguclu, line 208: I think it would be better to do these operations only if platformType is equal to Baremetal. I know `NewBaremetalTestHelper` is intelligent enough to skip for other platform types but from this test's perspective, test just can not rely on it.

test/extended/machines/scale.go
```diff
@@ -186,6 +187,35 @@ func scaleMachineSet(name string, replicas int) error {
...
+		if helper.CanDeployExtraWorkers() {
+			helper.Setup()
+			helper.DeployExtraWorker(0)
+		}
+	})
+
+	g.AfterEach(func() {
+		helper.DeleteAllExtraWorkers()
```
ardaguclu, line 216: I know helper in here is actually `NewBaremetalTestHelper` but reader can not easily understand this operation is also for other platforms?

test/extended/machines/scale.go
```diff
@@ -186,6 +187,35 @@ func scaleMachineSet(name string, replicas int) error {
...
 
 var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Serial] Managed cluster should", func() {
+
+	var (
+		c      *kubernetes.Clientset
+		dc     dynamic.Interface
+		helper *bmhelper.BaremetalTestHelper
+	)
+
+	g.BeforeEach(func() {
```
ardaguclu, line 197: IMHO, instead wrapping up with BeforeEach and AfterEach just for Baremetal, I think it would be more convenient to use if baremetal and defer statements. Because in reality, this change has no effect on other platforms.

test/extended/machines/scale.go
```diff
@@ -26,7 +27,7 @@ const (
 	machineAPIGroup                = "machine.openshift.io"
 	machineSetOwningLabel          = "machine.openshift.io/cluster-api-machineset"
 	machineSetDeleteNodeAnnotaiton = "machine.openshift.io/cluster-api-delete-machine"
-	scalingTime                    = 12 * time.Minute
+	scalingTime                    = 15 * time.Minute
```
ardaguclu, line 30: As clearly seen you intentionally increased it. I'm asking just because of curiosity; what is the reason that?

ardaguclu (commented):
Thanks @andfasano I left some comments

test/extended/machines/scale.go
andfasano: In my local test I started to have timeout failure, as it seems that the provisioning time somehow slightly increased, so I also retouched it to see it if it works in the CI

test/extended/machines/scale.go
andfasano: BeforeEach/AfterEach are the ways supported by [ginkgo](https://onsi.github.io/ginkgo/) for structuring the setup/teardown in a clear/readable mannder, and also ensuring the proper order of execution. I'd find complicated to follow other custom approaches for that (the helper teardown will not have any effect in case of non-baremetal platform)

test/extended/machines/scale.go
andfasano: Considering that we are slightly enlarging the scopes of baremetal related operators (ie CBO) to other platforms, my intention was to have a more granular helper class. If we could have the same support on other platforms in future, this test could continue to work. Also, the setup/teardown code would be less cumbersome (across different tests), given that the helper will not required it (personally, I'd limit the platform type check only to decide if a whole test must be skipped or not).
I'm not very happy of the teardown method name, probably naming it `Teardown` would make the helper class even more generic.

test/extended/machines/scale.go
andfasano: See previous comment on the usability of the helper class

elfosardo (approved):

test/extended/baremetal/helper.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// CreateExtraWorker creates a new BaremetalHost (and associated Secret).
+// If successfull, returns the newly created host and secret resources
+func (b *BaremetalTestHelper) CreateExtraWorker(host *unstructured.Unstructured, secret *v1.Secret) (*unstructured.Unstructured, *v1.Secret) {
+
+	secret, err := b.clientSet.CoreV1().Secrets("openshift-machine-api").Create(context.Background(), secret, metav1.CreateOptions{})
+	o.Expect(err).ToNot(o.HaveOccurred())
+
+	host, err = b.bmcClient.Create(context.Background(), host, metav1.CreateOptions{})
```
bfournie, line 128: Wondering if its intentional to use context.Background() here and context.TODO() in other places (e.g. line 149)?

test/extended/machines/scale.go
```diff
@@ -186,6 +187,35 @@ func scaleMachineSet(name string, replicas int) error {
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		dc, err = dynamic.NewForConfig(cfg)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		// For baremetal platforms, an extra worker must be previously
+		// deployed to allow subsequent scaling operations
+		helper = bmhelper.NewBaremetalTestHelper(dc)
+		if helper.CanDeployExtraWorkers() {
+			helper.Setup()
+			helper.DeployExtraWorker(0)
```
bfournie, line 211: This will only deploy one extra worker, is a mechanism needed to deploy more than one?

test/extended/machines/scale.go
```diff
@@ -186,6 +187,35 @@ func scaleMachineSet(name string, replicas int) error {
...
+		}
+	})
+
+	g.AfterEach(func() {
+		helper.DeleteAllExtraWorkers()
+	})
+
```
bfournie, line 218: When scaling down the number of replicas, would it be possible to define that a particular extra worker should be scaled down using the equivalent of "oc annotate machine <extra-worker> machine.openshift.io/cluster-api-delete-machine=yes"? Perhaps the test is already doing that and I missed it.

andfasano: Yes @bfournie the test is exactly using that approaching, by marking the newly created machine with the `cluster-api-delete-machine` annotation (see [here](https://github.com/openshift/origin/blob/838021cc5dfa1a75b6efacfc77ecae65850f564e/test/extended/machines/scale.go#L260-L262))

ardaguclu:
e2e-metal-ipi-serial is perm failing. Fix should be https://github.com/openshift/release/pull/25304

ardaguclu:
@andfasano after this https://github.com/openshift/origin/pull/26660 is merged, there is merge conflict.

andfasano:
> @andfasano after this #26660 is merged, there is merge conflict.

Thanks, done. @ardaguclu @bfournie could you please have another look?

bparees:
> The existing scale test has been also retouched to adopt the new add worker mechanism in case of baremetal ipi platform

how was this test passing on baremetal ipi before, if the additional worker is needed?

andfasano:
> > The existing scale test has been also retouched to adopt the new add worker mechanism in case of baremetal ipi platform
> 
> how was this test passing on baremetal ipi before, if the additional worker is needed?

Previously the extra worker was directly created by the dev-scripts setup via the [APPLY_EXTRA_WORKERS](https://github.com/openshift/release/blob/7a8dd52be2f1d9418102be47762d19572ffe9cb7/ci-operator/step-registry/baremetalds/e2e/serial/ipv4/baremetalds-e2e-serial-ipv4-workflow.yaml#L15) config variable (so it was available just at the end of the setup phase, before the e2e test step).
We plan to remove the `APPLY_EXTRA_WORKERS` as soon as this PR lands (current code also takes care of the migration)

PR #26682, opened 2021-12-09
fix disruption sampler

deads2k:
This unreverts, extends the timeout, and then fixes the sampler.


deads2k:
> /hold bug in here.

you found it?

test/extended/util/disruption/backend_sampler_tester.go
```diff
@@ -0,0 +1,130 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/upgrades"
+)
+
+type BackendSampler interface {
+	GetDisruptionBackendName() string
+	GetLocator() string
+	GetURL() (string, error)
+	StartEndpointMonitoring(ctx context.Context, m monitor.Recorder, eventRecorder events.EventRecorder) error
+}
```
dgoodwin, line 21: Could this interface move over to the package that implements it, it's a little off in the weeds here.

test/extended/util/disruption/disruption.go
```diff
@@ -325,6 +325,22 @@ func ExpectNoDisruption(f *framework.Framework, tolerate float64, total time.Dur
 	}
 }
 
+// ExpectNoDisruptionForDuration fails if the sum of the duration of all events exceeds allowedDisruption, reports a
+// disruption flake if any disruption occurs, and uses reason to prefix the message.
+func ExpectNoDisruptionForDuration(f *framework.Framework, allowedDisruption time.Duration, total time.Duration, events monitorapi.Intervals, reason string) {
```
dgoodwin, line 330: "events" should probably be renamed to "intervals", a bit confusing given we're using kube events very close to this.

dgoodwin:
> > /hold bug in here.
> 
> you found it?

No that was based on your comment on Friday. What is the symptom you saw that led you to believe there is a bug? The CI jobs look pretty good.

dgoodwin (commented):
I don't fully understand all of this yet but got a ways into it. Primary request is to godoc what is relatively complex code such that someone else has a decent shot of maintaining it next time a change is needed. Might take another pass at this later today.

dgoodwin:
I'd also really like to see some instructions for how someone could develop this, how could you compile and run one of these tests against an arbitrary backend you could play with to cause disruption and see how the sampler behaves. This would go a LONG way for the next person to try to work in this code.

test/extended/util/disruption/backend_sampler_tester.go
deads2k: > Could this interface move over to the package that implements it, it's a little off in the weeds here.

it's a cyclical import thing and golang encourages local interfaces with duck typing.

pkg/monitor/disruption_backend_sampler.go
```diff
@@ -0,0 +1,509 @@
...
+
+		switch {
+		case currentlyAvailable && previouslyAvailable:
+			// we are continuing to function.  no condition change.
+
+		case !currentlyAvailable && !previouslyAvailable:
+			// we are continuing to fail, check to see if the error is new
+			if previousError.Error() == currentError.Error() && !firstSample {
+				// if the error is the same and this isn't the first sample we have, skip
+				break
```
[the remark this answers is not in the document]
deads2k: > Should this be a continue?

breaking from the switch I think, but we still want the closing lines of hte loop

pkg/monitor/disruption_backend_sampler.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	uncast := b.activeSamplers.Front()
+	return uncast.Value.(*disruptionSample)
+}
+
+func (b *disruptionSampler) newSample(ctx context.Context) *disruptionSample {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+	currentDisruptionSample := newDisruptionSample(time.Now())
+	b.activeSamplers.PushBack(currentDisruptionSample)
```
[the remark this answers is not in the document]
deads2k: > Why do I not see anything getting popped from this list? I'd have expected something to be removed from active samplers at some point

yep, that was one of the bugs.

deads2k:
think I got it.

/hold cancel

test/extended/util/disruption/backend_sampler_tester.go
dgoodwin: Ah local to where it's used, makes sense.

dgoodwin:
Couple style nits missed but it looks good, glad the review actually helped find an issue. :tada: 

The payload results look troubling though so will readd the hold.

/lgtm
/hold

deads2k:

impacts upgrades, those are looking good enough to gather data.

PR #26684, opened 2021-12-09
force a new payload

deads2k:


deads2k:
https://github.com/openshift/origin/pull/26683 appears to have cancelled its image build (merged too soon?), so if we don't get a new payload flowing, I'll try this one on a successful image build.

PR #26686, opened 2021-12-10
Revert "Alter junit test suite names for jobs with multiple junit xmls."

dgoodwin:
Reverts openshift/origin#26667

This is blocking payloads because on Azure, we now correctly report the sandbox test as a failure, whereas previously the pass+fail tests were incorrectly put together and considered a flake. Aggregation code and bigquery thinks the historical pass rate is 97%, and that is not the case at all on azure. With deads out and required to push any changes to fix the image manually, I don't think I can code a way to skip the test. 

For now I will revert this fix, until we can figure out how to keep payloads flowing and get the test data updated.

### DIFF
--- a/cmd/openshift-tests/upgrade.go
+++ b/cmd/openshift-tests/upgrade.go
@@ -16,7 +16,7 @@ import (
 	"k8s.io/kubernetes/test/e2e/upgrades"
 )
 
-// upgradeSuites are all known upgrade test suites this binary should run
+// upgradeSuites are all known upgade test suites this binary should run
 var upgradeSuites = testSuites{
 	{
 		TestSuite: ginkgo.TestSuite{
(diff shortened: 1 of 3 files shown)


cmd/openshift-tests/upgrade.go
```diff
@@ -16,7 +16,7 @@ import (
 	"k8s.io/kubernetes/test/e2e/upgrades"
 )
 
-// upgradeSuites are all known upgrade test suites this binary should run
+// upgradeSuites are all known upgade test suites this binary should run
```
DennisPeriquet, line 19: mis-spelling

dgoodwin: Whoops, thanks!

dgoodwin: Oh I guess I fixed it in my commit, and reverting is removing it. :) This is probably ok, we'll revert the revert once I can get guidance from deads, and that will bring it back in.

PR #26691, opened 2021-12-14
test: avoid double slash in path

giuseppe:
every user of configObjectsDir already appends a slash after
configObjectsDir.  Avoid generating a path that has a double slash
as `var/oauth/configobjects//htpasswd`.

The reason for such fix is that the crun version currently used in the
CI has a bug when dealing with paths with multiple slashes.  The issue
has already been addressed upstream.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


harche:
Addresses failures like https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/23609/rehearse-23609-pull-ci-openshift-kubernetes-release-4.10-e2e-aws-crun/1470678801741516800

PR #26695, opened 2021-12-15
Bug 2034688: allow Prometheus/Thanos to return 401 or 403 when the request isn't authenticated

raptorsun:
Both tests are querying prometheus/thanos endpoints without any authentication information, the response HTTP status code should be [401 unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) (the user is not yet authenticated) instead of [403 forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) (an authenticated user has no access to the requested resource.

Do not merge yet. This can block tests of Cluster Monitoring Operator which is returning the code 403 from Oauth-proxy at the moment.

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -580,8 +580,8 @@ var _ = g.Describe("[sig-instrumentation] Prometheus", func() {
 				return false, nil
 			})).NotTo(o.HaveOccurred(), fmt.Sprintf("Did not find tsdb_samples_appended_total, tsdb_head_samples_appended_total, or prometheus_tsdb_head_samples_appended_total"))
 
-			g.By("verifying the oauth-proxy reports a 403 on the root URL")
-			err := helper.ExpectURLStatusCodeExec(ns, execPod.Name, url, 403)
+			g.By("verifying the Thanos querier service requires authentication")
+			err := helper.ExpectURLStatusCodeExec(ns, execPod.Name, url, 401, 403)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			g.By("verifying a service account token is able to authenticate")
--- a/test/extended/util/prometheus/helpers.go
+++ b/test/extended/util/prometheus/helpers.go
@@ -235,17 +235,20 @@ func RunQueries(promQueries map[string]bool, oc *exutil.CLI, ns, execPodName, ba
 }
 
 // ExpectURLStatusCodeExec attempts connection to url returning an error
-// upon failure or if status return code is not equal to statusCode.
-func ExpectURLStatusCodeExec(ns, execPodName, url string, statusCode int) error {
+// upon failure or if status return code is not equal to any of the statusCodes.
+func ExpectURLStatusCodeExec(ns, execPodName, url string, statusCodes ...int) error {
 	cmd := fmt.Sprintf("curl -k -s -o /dev/null -w '%%{http_code}' %q", url)
 	output, err := framework.RunHostCmd(ns, execPodName, cmd)
 	if err != nil {
 		return fmt.Errorf("host command failed: %v\n%s", err, output)
 	}
-	if output != strconv.Itoa(statusCode) {
-		return fmt.Errorf("last response from server was not %d: %s", statusCode, output)
+	for _, statusCode := range statusCodes {
+		if output == strconv.Itoa(statusCode) {
+			return nil
+		}
 	}
-	return nil
+
+	return fmt.Errorf("last response from server was not in %v: %s", statusCodes, output)
 }
 
 // ExpectPrometheusEndpoint attempts to connect to the metrics endpoint with
@@ -253,7 +256,7 @@ func ExpectURLStatusCodeExec(ns, execPodName, url string, statusCode int) error 
 func ExpectPrometheusEndpoint(namespace, podName, url string) {
 	var err error
 	for i := 0; i < maxPrometheusQueryAttempts; i++ {
-		err = ExpectURLStatusCodeExec(namespace, podName, url, 403)
+		err = ExpectURLStatusCodeExec(namespace, podName, url, 401, 403)
 		if err == nil {
 			break
 		}


test/extended/prometheus/prometheus.go
```diff
@@ -580,8 +580,11 @@ var _ = g.Describe("[sig-instrumentation] Prometheus", func() {
 				return false, nil
 			})).NotTo(o.HaveOccurred(), fmt.Sprintf("Did not find tsdb_samples_appended_total, tsdb_head_samples_appended_total, or prometheus_tsdb_head_samples_appended_total"))
 
-			g.By("verifying the oauth-proxy reports a 403 on the root URL")
-			err := helper.ExpectURLStatusCodeExec(ns, execPod.Name, url, 403)
+			g.By("verifying the oauth-proxy reports a 401 or 403 on the root URL")
```
simonpasquier, line 583:
```suggestion
			g.By("verifying the root URL requires authentication")
```

test/extended/prometheus/prometheus.go
```diff
@@ -580,8 +580,11 @@ var _ = g.Describe("[sig-instrumentation] Prometheus", func() {
...
+			err := helper.ExpectURLStatusCodeExec(ns, execPod.Name, url, 401)
```
simonpasquier, line 584: I would rather modify `ExpectURLStatusCodeExec()` to check a list of possible status codes, e.g.

```
func ExpectURLStatusCodeExec(ns, execPodName, url string, statusCode ...int) error {
}
```

simonpasquier:
/retitle allow Prometheus/Thanos to return 401 or 403 when the request isn't authenticated

test/extended/prometheus/prometheus.go
raptorsun: Good idea, I will modify it in this way. :)

simonpasquier:
/retitle WIP: Bug xxx: allow Prometheus/Thanos to return 401 or 403 when the request isn't authenticated

simonpasquier:
/skip
/lgtm
/retitle Bug 2034688: allow Prometheus/Thanos to return 401 or 403 when the request isn't authenticated

simonpasquier:
/hold

this isn't needed right now since https://github.com/openshift/cluster-monitoring-operator/pull/1512 doesn't replace the existing oauth-proxy sidecar by kube-rbac-proxy.

raptorsun:
/unhold
Now the work on replacing oauth-proxy by kube-rbac-proxy for Thanos deployments resumes :) 
This fix will be needed.

test/extended/prometheus/prometheus.go
```diff
@@ -580,8 +580,8 @@ var _ = g.Describe("[sig-instrumentation] Prometheus", func() {
 				return false, nil
 			})).NotTo(o.HaveOccurred(), fmt.Sprintf("Did not find tsdb_samples_appended_total, tsdb_head_samples_appended_total, or prometheus_tsdb_head_samples_appended_total"))
 
-			g.By("verifying the oauth-proxy reports a 403 on the root URL")
-			err := helper.ExpectURLStatusCodeExec(ns, execPod.Name, url, 403)
+			g.By("verifying the oauth-proxy requires authentication")
```
simonpasquier, line 583:
```suggestion
			g.By("verifying the Thanos querier service requires authentication")
```

test/extended/util/prometheus/helpers.go
```diff
@@ -235,25 +235,32 @@ func RunQueries(promQueries map[string]bool, oc *exutil.CLI, ns, execPodName, ba
...
 	if err != nil {
 		return fmt.Errorf("host command failed: %v\n%s", err, output)
 	}
-	if output != strconv.Itoa(statusCode) {
-		return fmt.Errorf("last response from server was not %d: %s", statusCode, output)
+	for _, statusCode := range statusCodes {
+		if output == strconv.Itoa(statusCode) {
+			return nil
+		}
 	}
```
simonpasquier, line 249:
```suggestion
        strStatusCodes := make([]string, len(statusCodes))
	for i, statusCode := range statusCodes {
		if output == strconv.Itoa(statusCode) {
			return nil
		}
		strStatusCodes[i] = strconv.Itoa(statusCode)
	}
```

test/extended/util/prometheus/helpers.go
```diff
@@ -235,25 +235,32 @@ func RunQueries(promQueries map[string]bool, oc *exutil.CLI, ns, execPodName, ba
...
-	return nil
+	strStatusCodes := make([]string, len(statusCodes))
+	for i, statusCode := range statusCodes {
+		strStatusCodes[i] = strconv.Itoa(statusCode)
+	}
+
+	return fmt.Errorf("last response from server was not %s: %s", strings.Join(strStatusCodes, ","), output)
```
simonpasquier, line 255:
```suggestion
	return fmt.Errorf("last response from server was not in [%s]: %s", strings.Join(strStatusCodes, ","), output)
```

raptorsun:
This PR is now required by https://github.com/openshift/cluster-monitoring-operator/pull/1580
In CMO we are replacing OAuth-proxy with Kube-RBAC-proxy, so we have to allow both 401 and 403 status codes as a response to unauthenticated users.

raptorsun:
Hello @bparees , Could you please have a look at this change? It can deblock a current development in cluster monitoring operator :)

test/extended/util/prometheus/helpers.go
```diff
@@ -235,25 +235,30 @@ func RunQueries(promQueries map[string]bool, oc *exutil.CLI, ns, execPodName, ba
...
 		return fmt.Errorf("host command failed: %v\n%s", err, output)
 	}
-	if output != strconv.Itoa(statusCode) {
-		return fmt.Errorf("last response from server was not %d: %s", statusCode, output)
+	strStatusCodes := make([]string, len(statusCodes))
+	for i, statusCode := range statusCodes {
+		if output == strconv.Itoa(statusCode) {
+			return nil
+		}
+		strStatusCodes[i] = strconv.Itoa(statusCode)
```
bparees, line 250: this seems unnecessary.  just use %v, e.g:

https://go.dev/play/p/dFdJ7HBi639

raptorsun:
ready to be reviewed again :)

raptorsun:
The job ci/prow/e2e-gcp-builds fails when git clone this repository using the HTTPS protocol:
`fatal: unable to access 'https://github.com/openshift/origin.git/': Could not resolve host: github.com`
It is strange it fails to resolve the domain name of github.com 🤔

PR #26702, opened 2021-12-16
Add tests for ErrImagePull monitor intervals.

dgoodwin:


pkg/test/ginkgo/synthentic_tests.go
```diff
@@ -73,11 +83,52 @@ func createSyntheticTestsFromMonitor(events monitorapi.Intervals, monitorDuratio
...
+	if errImagePullTimeouts > 0 {
+		syntheticTestResults = append(
+			syntheticTestResults,
+			&JUnitTestCase{
+				Name:     errImagePullTestName,
+				SystemOut: buf.String(),
+				Duration:  monitorDuration.Seconds(),
+				FailureOutput: &FailureOutput{
+					// Not including the entire buffer again as it's very large and is added to junit xml above.
+					Output: fmt.Sprintf("%d ErrImagePull read: connection timed out events were detected during this test run", errImagePullTimeouts),
```
deads2k, line 112: very reasonable test.  Try to factor it like https://github.com/openshift/origin/blob/master/pkg/synthetictests/event_junits.go#L18

and capture:
1. errimagepull in openshift-* namespaces for any reason
2. errimagepull in openshift-* namespaces for read connection timed out
3. errimagepull in non-openshift-* namespaces for any reason
4. errimagepull in non-openshift-* namespaces for read connection timed out

as four different tests.  you may want to start all of them as flaking.

pkg/synthetictests/kubelet.go
```diff
@@ -363,3 +364,59 @@ func testSystemDTimeout(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
 	// write a passing test to trigger detection of this issue as a flake. Doing this first to try to see how frequent the issue actually is
 	return []*ginkgo.JUnitTestCase{failure, success}
 }
+
+var errImagePullTimeoutRE = regexp.MustCompile("ErrImagePull.*read: connection timed out")
```
deads2k, line 368: my kryptonite!

pkg/synthetictests/kubelet.go
```diff
@@ -363,3 +364,59 @@ func testSystemDTimeout(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
+	const testName = "[sig-node] should not encounter ErrImagePull in non-openshift namespace pods"
+	return buildTestsFailIfRegexMatch(testName, errImagePullGenericRE, errImagePullTimeoutRE, false, events)
+}
+
+func buildTestsFailIfRegexMatch(testName string, matchRE, dontMatchRE *regexp.Regexp, inOCPNS bool, events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
+
+	matchedErrMsgs := []string{}
+	for _, event := range events {
+		estr := event.String()
+		if matchRE.MatchString(estr) && (dontMatchRE == nil || !dontMatchRE.MatchString(estr)) {
```
deads2k, line 396: unnest

```go
if !namespace-match{
  continue
}

if !matchRE{
  continue
}

if match(don't match){
  continue
}

matchedErrorMsg

```

pkg/test/ginkgo/synthentic_tests.go
```diff
@@ -73,10 +74,20 @@ func createSyntheticTestsFromMonitor(events monitorapi.Intervals, monitorDuratio
 			// write a passing test to trigger detection of this issue as a flake, indicating we have no idea whether
 			// these are actual failures or not
 			&JUnitTestCase{
-				Name:     "[sig-arch] Monitor cluster while tests execute",
+				Name:     monitorTestName,
 				Duration: monitorDuration.Seconds(),
 			},
 		)
+	} else {
+		// even if no error events, add a passed test including the output so we can scan with search.ci:
```
deads2k, line 82: this is fine, but how about always printing `buf` to stdout.

dgoodwin: Literally printing to stdout in this code, or always assigning to SystemOut field on the test case?

I didn't want to include it on the above "flake" success because its' a huge wall of text that's already included on the failure on line 70, and didn't seem worthwhile to duplicate it on the success test.

pkg/synthetictests/kubelet.go
dgoodwin: done!

pkg/synthetictests/kubelet.go
```diff
@@ -363,3 +364,70 @@ func testSystemDTimeout(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
+	success := &ginkgo.JUnitTestCase{Name: testName}
+	if len(matchedErrMsgs) > 0 {
+		return []*ginkgo.JUnitTestCase{success}
+	}
+
+	failure := &ginkgo.JUnitTestCase{
+		Name:      testName,
+		SystemOut: strings.Join(matchedErrMsgs, "\n"),
+		FailureOutput: &ginkgo.FailureOutput{
+			Output: fmt.Sprintf("%d pod event intervals matched regex %s\n\n%s",
```
deads2k, line 425: I dislike error messages like this.  Identifying pods and sorting by them or identifying images would be helpful.  the point is not the regex, the point is the impact of the failure.

pkg/synthetictests/kubelet.go
```diff
@@ -363,3 +364,70 @@ func testSystemDTimeout(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
+	const testName = "[sig-node] should not encounter ErrImagePull in openshift namespace pods"
+	return buildTestsFailIfRegexMatch(testName, errImagePullGenericRE, errImagePullTimeoutRE, true, events)
+}
+
+func testErrImagePullGeneric(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
+	const testName = "[sig-node] should not encounter ErrImagePull in non-openshift namespace pods"
+	return buildTestsFailIfRegexMatch(testName, errImagePullGenericRE, errImagePullTimeoutRE, false, events)
+}
+
+func buildTestsFailIfRegexMatch(testName string, matchRE, dontMatchRE *regexp.Regexp, inOCPNS bool, events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
```
deads2k, line 391: bools in APIs like this age poorly.  in the future, choose something like a namespace selector function or a struct that contains these options.

pkg/synthetictests/kubelet.go
deads2k, line 425: you didn't limit to pods, did you mean to?

deads2k:
I dislike the factorization, but not enough to skip gaining signal from the output.  I think the API will age poorly if we have to tweak it overtime and someone will try to use this "generic" matching somewhere else to ill effect.


/lgtm
/approve

deads2k:
/hold

it's actually bugged somewhere in teh regex stuff

pkg/synthetictests/kubelet.go
```diff
@@ -363,3 +364,70 @@ func testSystemDTimeout(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
+var errImagePullGenericRE = regexp.MustCompile("ErrImagePull")
+
+func testErrImagePullConnTimeoutOpenShiftNamespaces(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
+	const testName = "[sig-node] should not encounter ErrImagePull read connection timeout in openshift namespace pods"
+	return buildTestsFailIfRegexMatch(testName, errImagePullTimeoutRE, nil, true, events)
+}
+
+func testErrImagePullConnTimeout(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
+	const testName = "[sig-node] should not encounter ErrImagePull read connection timeout in non-openshift namespace pods"
+	return buildTestsFailIfRegexMatch(testName, errImagePullTimeoutRE, nil, false, events)
```
deads2k, line 378: `nil, false` reads really badly. if you make a struct to hold your three configuration settings you could consistently name them.  If you made a const in here, you coudl refer to it by name in-line

pkg/synthetictests/kubelet.go
```diff
@@ -363,3 +364,70 @@ func testSystemDTimeout(events monitorapi.Intervals) []*ginkgo.JUnitTestCase {
...
+		// Skip if we *do* match the don't match regex:
+		if dontMatchRE != nil && dontMatchRE.MatchString(estr) {
+			continue
+		}
+
+		matchedErrMsgs = append(matchedErrMsgs, event.String())
+	}
+
+	success := &ginkgo.JUnitTestCase{Name: testName}
+	if len(matchedErrMsgs) > 0 {
```
deads2k, line 417: this is backwards, right?

pkg/synthetictests/kubelet.go
dgoodwin: No just assumed it would limit itself to pods, I'll remove the word, I want to see any event with that string regardless. They would show up as non-openshift- namespaces.

pkg/synthetictests/kubelet.go
dgoodwin: Yikes yes.

pkg/synthetictests/kubelet.go
dgoodwin: I tried the struct but not a fan of how that's coming out, everything then starts looking optional when in reality, only the dontMatchRegex is optional, and then the function starts having to do a lot of nil/empty checking and error handling because it's not as clear how to invoke it. More than one optional param and I can see the value in the struct.

I added an enum for the in openshift- or not bool instead though. Hopefully this is clear enough what is being done.

pkg/synthetictests/kubelet.go
dgoodwin: Replacing with an enum

pkg/synthetictests/kubelet.go
dgoodwin: I will add a pod clause to the regex, I think I see how to handle that now.

dgoodwin: I don't want to get into parsing locators, but I have changed it so the FailureOutput has sorted list of locators (ns/blah pod/foo container/bar) only, and the SystemOut has the full event String. (which  can be very large)

dgoodwin: Scratch that, I do want to catch them even if somehow magically not including a pod in the locator.

dgoodwin:
I do not know what you dislike about the factoring, as I thought this is what you requested earlier, one for ErrImagePull with read: connection timeout, and one for ErrImagePull without, which I called a generic ErrImagePull. If there's a better name or more clarification on what your issue is I'd be happy to address.

dgoodwin:
Waiting to see test results, two possible fixes in here, the bad logic after refactoring on the return success, and the assumption that ns/openshift- was at the start of the .String() output, I don't think it actually was, changed to contains.

dgoodwin:
Looks to be working, I think ErrImagePull is going to be very common in non-openshift namespaces from what I'm seeing.

dgoodwin:
Rebased and needs another lgtm. :(

deads2k:
adding a test to measure impact. overriding bug requirement

PR #26703, opened 2021-12-16
use standard disruption checker for load balancer

deads2k:
replaces https://github.com/openshift/origin/pull/26675, built on https://github.com/openshift/origin/pull/26682

### DIFF
--- a/pkg/monitor/backenddisruption/disruption_backend_sampler.go
+++ b/pkg/monitor/backenddisruption/disruption_backend_sampler.go
@@ -198,6 +198,10 @@ func (b *BackendSampler) bodyMatches(body []byte) error {
 	return nil
 }
 
+func (b *BackendSampler) SetHost(host string) {
+	b.host = host
+}
+
 func (b *BackendSampler) GetDisruptionBackendName() string {
 	return b.disruptionBackendName
 }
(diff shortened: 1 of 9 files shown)


deads2k:
/payload 4.10 ci blocking
/payload 4.10 nightly blocking

/hold

holding so we can see the payload results.  I think this will actually reduce the downtime observed for service LB because it's more likely to end up as one second chunks.

pkg/monitor/backenddisruption/disruption_backend_sampler.go
```diff
@@ -198,6 +198,10 @@ func (b *BackendSampler) bodyMatches(body []byte) error {
 	return nil
 }
 
+func (b *BackendSampler) SetHost(host string) {
+	b.host = host
+}
```
dgoodwin, line 203: Is breaking the "With" pattern above intentional due to the late binding nature of it?

dgoodwin (commented):
Looks fine to me otherwise.

/lgtm

pkg/monitor/backenddisruption/disruption_backend_sampler.go
deads2k: > Is breaking the "With" pattern above intentional due to the late binding nature of it?

yes

deads2k:
4.9 doesn't honor the scc label.  I'll have to tweak the namespace too.  Oh teh ugly.

deads2k:
no conflict with previous.  merging.

PR #26705, opened 2021-12-16
Bug 2033341: tests: skip some tests for proxy

EmilienM:
Some tests are known to be unstable since they need Internet access and
therefor cause issue when running behind a secured proxy.

Let's skip them for now.

(cherry picked from commit b2d569128f2abb5434ea5d36d9bf829efab6562b)

### DIFF
--- a/test/extended/util/annotate/rules.go
+++ b/test/extended/util/annotate/rules.go
@@ -94,6 +94,14 @@ var (
 			`\[sig-cli\] Kubectl Port forwarding With a server listening on localhost should support forwarding over websockets`,
 			`\[sig-cli\] Kubectl Port forwarding With a server listening on 0.0.0.0 should support forwarding over websockets`,
 			`\[sig-node\] Pods should support remote command execution over websockets`,
+
+			// These tests are flacky and require internet access
+			// See https://bugzilla.redhat.com/show_bug.cgi?id=2019375
+			`\[sig-builds\]\[Feature:Builds\] build can reference a cluster service with a build being created from new-build should be able to run a build that references a cluster service`,
+			`\[sig-builds\]\[Feature:Builds\] oc new-app should succeed with a --name of 58 characters`,
+			`\[sig-network\] DNS should resolve DNS of partial qualified names for services`,
+			`\[sig-arch\] Only known images used by tests`,
+			`\[sig-network\] DNS should provide DNS for the cluster`,
 		},
 		"[Skipped:SingleReplicaTopology]": {
 			`\[sig-apps\] Daemon set \[Serial\] should rollback without unnecessary restarts \[Conformance\]`,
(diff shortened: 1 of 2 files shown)


test/extended/util/annotate/rules.go
```diff
@@ -94,6 +94,14 @@ var (
...
 			`\[sig-cli\] Kubectl Port forwarding With a server listening on 0.0.0.0 should support forwarding over websockets`,
 			`\[sig-node\] Pods should support remote command execution over websockets`,
+
+			// These tests are flacky and require internet access
+			// See https://bugzilla.redhat.com/show_bug.cgi?id=2019375
+			`\[sig-builds\]\[Feature:Builds\] build can reference a cluster service with a build being created from new-build should be able to run a build that references a cluster service`,
+			`\[sig-builds\]\[Feature:Builds\] oc new-app should succeed with a --name of 58 characters`,
+			`\[sig-network\] DNS should resolve DNS of partial qualified names for services`,
+			`\[sig-arch\] Only known images used by tests`,
+			`\[sig-network\] DNS should provide DNS for the cluster`,
```
bparees, line 104: at least some of these seem like tests that should work in a proxy environment, has the investigation been done to understand why they are not?

i'm ok w/ disabling them while the investigation is done (via new bugs) but i'm not so sure we should just be disabling them w/o tracking a plan to fix them.

EmilienM: @bparees I've created a BZ and assigned it to the right team: https://bugzilla.redhat.com/show_bug.cgi?id=2033628
I'll keep an eye and hopefully at some point remove these tests fro that list, so we can exercise them if it makes sense.

Thanks

test/extended/util/annotate/rules.go
```diff
@@ -94,6 +94,14 @@ var (
 			`\[sig-cli\] Kubectl Port forwarding With a server listening on localhost should support forwarding over websockets`,
 			`\[sig-cli\] Kubectl Port forwarding With a server listening on 0.0.0.0 should support forwarding over websockets`,
 			`\[sig-node\] Pods should support remote command execution over websockets`,
+
+			// These tests are flacky and require internet access
+			// See https://bugzilla.redhat.com/show_bug.cgi?id=2019375
```
bparees, line 99: can you make a note here listing the bugs that are being used to investigate the failures (i.e. the bugs that, when resolved, will allow us to either re-enable these tests, or decide to skip them permanently)?

would be good to have that note in master as well.  This is how someone 6 months from now who is digging through this, can follow up on what ended up happening and ensure the tests are properly re-enabled(if they should be).

test/extended/util/annotate/rules.go
bparees: fyi it's probably going to end up needing more than one bug since these tests are owned by different teams

test/extended/util/annotate/rules.go
EmilienM: @bparees done via https://github.com/openshift/origin/pull/26710 - do you want me to add it here as well?

bparees: being done in master here:
https://github.com/openshift/origin/pull/26710

less critical to have it in the z-branch, since it'll eventually carry backwards from master over time and we're unlikely to backport the fixes to re-enable these tests in 4.9.

PR #26706, opened 2021-12-16
find disruption backends from the event intervals instead of a hardcoded list

deads2k:
builds on https://github.com/openshift/origin/pull/26703 and should give us a complete backend .json

### DIFF
--- a/pkg/monitor/disruption_known_backends.go
+++ b/pkg/monitor/disruption_known_backends.go
@@ -13,34 +13,6 @@ import (
 // we also got stuck on writing the disruption backends.  We need a way to track which disruption checks we have started,
 // so we can properly write out "zero"
 
-var (
-	LocatorKubeAPIServerNewConnection         = backenddisruption.LocateDisruptionCheck("kube-api", backenddisruption.NewConnectionType)
-	LocatorKubeAPIServerReusedConnection      = backenddisruption.LocateDisruptionCheck("kube-api", backenddisruption.ReusedConnectionType)
-	LocatorOpenshiftAPIServerNewConnection    = backenddisruption.LocateDisruptionCheck("openshift-api", backenddisruption.NewConnectionType)
-	LocatorOpenshiftAPIServerReusedConnection = backenddisruption.LocateDisruptionCheck("openshift-api", backenddisruption.ReusedConnectionType)
-	LocatorOAuthAPIServerNewConnection        = backenddisruption.LocateDisruptionCheck("oauth-api", backenddisruption.NewConnectionType)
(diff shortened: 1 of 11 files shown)


pkg/monitor/disruption_known_backends.go
```diff
@@ -13,34 +13,6 @@ import (
...
-// the job aggregator.
-var BackendDisruptionLocatorsToName = map[string]string{
-	LocatorKubeAPIServerNewConnection:         "kube-api-new-connections",
-	LocatorOpenshiftAPIServerNewConnection:    "openshift-api-new-connections",
-	LocatorOAuthAPIServerNewConnection:        "oauth-api-new-connections",
-	LocatorKubeAPIServerReusedConnection:      "kube-api-reused-connections",
-	LocatorOpenshiftAPIServerReusedConnection: "openshift-api-reused-connections",
-	LocatorOAuthAPIServerReusedConnection:     "oauth-api-reused-connections",
-	backenddisruption.LocateRouteForDisruptionCheck("openshift-authentication", "oauth-openshift", "ingress-to-oauth-server", backenddisruption.NewConnectionType):    "ingress-to-oauth-server-new-connections",
-	backenddisruption.LocateRouteForDisruptionCheck("openshift-authentication", "oauth-openshift", "ingress-to-oauth-server", backenddisruption.ReusedConnectionType): "ingress-to-oauth-server-used-connections",
```
deads2k, line 35: we will lose history on this name :(

it changes to `ingress-to-oauth-server-reused-connections` .  We may be able (if we wish) to do a bulk update in the table.  I'm not completely sure it will work.

deads2k:
also holding to confirm compatibility with the uploader.

/hold

deads2k:
comparing against existing, we now get consistency, but lose history on
1. ingress-to-console-used-connections (now ingress-to-console-reused-connections)
2. ingress-to-oauth-server-used-connections (now ingress-to-oauth-server-reused-connections)
3. service-load-balancer-with-pdb-new-connections (now service-loadbalancer-with-pdb-new-connections)
4. service-load-balancer-with-pdb-reused-connections (now service-loadbalancer-with-pdb-reused-connections)

I'm going to tweak the service LB test name to match so we don't lose that, but I think we eat the ingress to console changes.

I'm also going to re-capitalize the connection types so the table stays clean

deads2k:
nwo pushed with update

deads2k:
looks good now

/hold cancel
failed on existing flakes
/override ci/prow/e2e-gcp 
promtail

deads2k:
no conflict with existing, addresses CI need.

PR #26707, opened 2021-12-16
add downtime checks for console and oauth-server

deads2k:
/hold

builds on other PRs which need proof first.


deads2k:
gathering data on ingress outages during all e2e runs

PR #26709, opened 2021-12-17
extend tcp dial timeout and http response header timeout

deads2k:
azure is weirdly slow.  We may even have to bump again, but let's see what this does

ref https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade/1471779397802397696


deads2k:
relaxing timeout to clear more payloads.

PR #26713, opened 2021-12-17
extend timeouts with 5s for each part

deads2k:
tolerate even more disruption with the more precise measurement.


deads2k:
trying to find the limit.

PR #26714, opened 2021-12-17
test/extended/util/annotate/generated: Delegate to team leads

wking:
Expand on the scaffolding set up with e6f298436d (#26652) to empower all OpenShift team leads to manage skips and such for test-cases covering their own components.

Alias content based on `misc/tools/bin/leads.py` from the internal `li` repo (sorry external folks) as of 4c5bc167f9 (2021-12-17), with subsequent manual pruning for teams that had missing data or didn't seem like they'd care about origin content.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -11,3 +11,43 @@ aliases:
   - jottofar
   - vrutkovs
   - wking
+  team-leads:
+  - JAORMX # Infrastructure Security and Compliance
+  - JoelSpeed # Cluster Infrastructure
+  - LalatenduMohanty # Over the Air Updates
+  - Miciah # Network Edge
+  - adambkaplan # Build + Jenkins
+  - ashishkamra # Performance and Latency Sensitive Application Platform
+  - beekhof # Machine Lifecycle
+  - brenton # Technical Release Team
+  - bthurber # Partner Accelerators
+  - c3d # OpenShift Sandboxed Containers
+  - celebdor # Ecosystem Edge
+  - dmage # Image Registry
+  - dulek # Kuryr Networking
+  - hexfusion # etcd
+  - jaypoulz # Multi Arch
+  - jcantrill # Cluster Logging
+  - jmrodri # Operator SDK
+  - joelddiaz # Hive
+  - jparrill # Portfolio Integration
+  - jsafrane # Storage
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -11,3 +11,43 @@ aliases:
...
+  - rphillips # Node
+  - s-urbaniak # Authentication
+  - sadasu # Metal Platform
+  - selansen # Windows Containers
+  - simonpasquier # Monitoring - In Cluster
+  - sinnykumari # MCO
+  - soltysh # Workloads
+  - sosiouxme # Automated Release Tooling
+  - staebler # OpenShift Installer
+  - stevekuznetsov # Test Platform
```
bparees, line 49: this should be petr muller

wking: Yeah, I told Petr that the li config is stale.

OWNERS_ALIASES
```diff
@@ -11,3 +11,43 @@ aliases:
...
+  - sttts # API Server
+  - thomasmckay # Quay-Quay.io
```
dmage, line 51: He is no longer in the Quay team. But I'm almost sure that the Quay team doesn't use this repo.

wking: If these team leads are not accurate, folks should update the Shiftzilla config in the li repository.  But I've pushed 81c439a4bc -> fc34a9be48 to drop Quay entirely, so there's no need to hold this PR on an li fix for this entry.

OWNERS_ALIASES
wking: Backing data was updated, I've refreshed here with fc34a9be48 -> 36b4e72d39, and this entry now correctly references Petr.

bparees:
moves the ball in the right direction, thanks @wking 

/lgtm
/approve

wking:
This won't merge until one of:

a. Someone decides it's worth a bug; unlikely, or
b. Someone manually labels it `bugzilla/valid-bug` because it's not touching anything that affects the shipped payload (`OWNERS` isn't compiled into shipped binaries), or
c. `master` re-opens for bugless PRs.

Until then, no sense in banging away with retest requests, so:

/hold

bparees:
i'll mark it valid-bug

/hold cancel

PR #26719, opened 2021-12-22
Bug 2026488: Skip Duplicate openshift-controller-manager Events

adambkaplan:
The full build suite includes several tests that force rollouts of
openshift-controller-manager. This causes duplicate
"reason/SuccessfulDelete" events to be detected, even though these
events are being triggered by separate rollouts. To fully address this
issue, openshift-controller-manager should be installed as a Deployment
rather than a DameonSet. This has better performance semantics with
respect to rollouts, which may benefit the install and upgrade
experience and reduce these "SuccessfulDelete" events.

This change provides a mechanism for known duplicate events to be
linked to a specific test suite, in addition to a specific platform or
topology. The known problem here is limited to the openshift/build test
suite, which can flake on this check and generally fails when the suite
is run on a cluster that enables tech preview features.

This skip should be removed when Bug 2034984 is fixed.

See also:
- https://bugzilla.redhat.com/show_bug.cgi?id=2026488
- https://bugzilla.redhat.com/show_bug.cgi?id=2034984

### DIFF
--- a/cmd/openshift-tests/images.go
+++ b/cmd/openshift-tests/images.go
@@ -211,7 +211,7 @@ func pulledInvalidImages(fromRepository string) ginkgo.JUnitForEventsFunc {
 
 	// any image not in the allowed prefixes is considered a failure, as the user
 	// may have added a new test image without calling the appropriate helpers
-	return func(events monitorapi.Intervals, _ time.Duration, cfg *rest.Config) []*ginkgo.JUnitTestCase {
+	return func(events monitorapi.Intervals, _ time.Duration, cfg *rest.Config, testSuite string) []*ginkgo.JUnitTestCase {
 		imageStreamPrefixes, err := imagePrefixesFromNamespaceImageStreams("openshift")
 		if err != nil {
 			klog.Errorf("Unable to identify image prefixes from the openshift namespace: %v", err)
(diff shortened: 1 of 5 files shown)


deads2k:
skipping the entire early/late packages is not appropriate.  If there is an issue with one particular test when running this suite, i'm open to targeting a single case of a single invariant/test check.  I think this is an attempt to skip a single repeated event during a single suite, right?

adambkaplan:
> skipping the entire early/late packages is not appropriate. If there is an issue with one particular test when running this suite, i'm open to targeting a single case of a single invariant/test check. I think this is an attempt to skip a single repeated event during a single suite, right?

Correct - updated the PR to skip the particular event, however the code as is will skip the check for all tests, which isn't desired. Is there a simple way to wire the invoked test suite here?

adambkaplan:
(also I did not realize we already had a mechanism for skipping certain events)

gabemontero:
doesn't look like we  can run any of the tech preview e2e's out of openshift/origin PRs:  https://github.com/openshift/release/blob/14ba81a04ef61b960df40dcb798ee38d617bd16b/ci-operator/config/openshift/origin/openshift-origin-master.yaml

suppose we'll just have to see how the tech preview periodic does after this merges ... I've been tracking that any this release given shared resource's including into tech preview

will look for 

"[sig-arch] events should not repeat pathologically" test occasionally fails with the following flake:

```
event happened 22 times, something is wrong: ns/openshift-controller-manager daemonset/controller-manager - reason/SuccessfulDelete (combined from similar events): Deleted pod: controller-manager-74rw5
```

/lgtm

PR #26724, opened 2022-01-03
Skip scs update when loadbalancer is unsupported

ardaguclu:
If platform(ie. ovirt, metal) does not support load balancer, load balancer disruption tests are skipped. But in any case, SCS is updated.

This PR skips SCS update when platform is unsupported.

### DIFF
--- a/test/e2e/upgrade/service/service.go
+++ b/test/e2e/upgrade/service/service.go
@@ -85,27 +85,6 @@ func (t *serviceLoadBalancerUpgradeTest) RequiresKubeNamespace() bool {
 func shouldTestPDBs() bool { return true }
 
 func (t *serviceLoadBalancerUpgradeTest) loadBalancerSetup(f *framework.Framework, backendSampler disruption.BackendSampler) error {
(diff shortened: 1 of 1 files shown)


ardaguclu:
`disruption_tests: [sig-network-edge] Application behind service load balancer with PDB remains available using new connections` test are failing intermittently in metal-ipi upgrade jobs. However, load balancer test are skipped in platforms such as ovirt, metal.

Example;
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.10-e2e-metal-ipi-upgrade-ovn-ipv6/1477067324841267200

Although root cause still not clear because it is pretty difficult to reproduce it on local environment, but this PR looks safe since just skipping SCS update for unsupported platforms.

@deads2k what do you think about this patch?, or if there is any other way to find the root cause maybe we can apply this patch in actual fix PR?, thanks.

ardaguclu:
I'm closing this PR, because it becomes redundant after https://github.com/openshift/origin/pull/26723

PR #26729, opened 2022-01-05
Bug 1919453: [release-4.6] configure tests to pull images from mirror

rvanderp3:
The intention of this PR is to enable some tests in the `openshift/conformance/parallel` suite which are not subject to the mirror configuration in `/etc/containers/registries.conf` to use a mirror as the source of their required images.  

Note: The pattern in this PR could be applied to other suites as well.

This is needed due to docker.io rate limiting causing CI on some platforms on 4.6 to be in permafail.  To use the feature in this PR:
- Prior to running `openshift-tests`, `export TEST_IMAGE_MIRROR_REGISTRY=<pull-through-cache-host>:<port>`
- Configure a pull-through cache for docker.io with a certificate signed by a well-known CA on a bastion that is only reachable by the nodes executing e2e tests. For example:
~~~
podman run -d -p 5000:5000 -v /home/admin/registry_storage:/var/lib/registry:Z -v /home/admin/registry_certs:/certs:Z --restart always --name registry-docker.io -e REGISTRY_HTTP_ADDR=0.0.0.0:5000 -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key -e REGISTRY_PROXY_REMOTEURL="https://registry-1.docker.io" -e REGISTRY_PROXY_USERNAME=<optional user name> -e REGISTRY_PROXY_PASSWORD=<optional password> registry:2
~~~

To enable resources which can take advantage of the mirror registry, apply a [machine-config](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/vsphere/registry/ipi-install-vsphere-registry-commands.sh#L34-L52) which configures the pull-through cache as a docker.io registry mirror.  This will allow, over the course of a day or so(based on pull-rate exhaustion), for all required images to be pulled and cached.

This PR is an alternate to solution to that proposed in #26256

### DIFF
--- a/test/extended/images/append.go
+++ b/test/extended/images/append.go
@@ -2,6 +2,7 @@ package images
 
 import (
 	"context"
+	"os"
 	"strings"
 
 	"github.com/MakeNowJust/heredoc"
@@ -15,6 +16,7 @@ import (
 	"github.com/openshift/library-go/pkg/image/imageutil"
 
 	exutil "github.com/openshift/origin/test/extended/util"
+	images "github.com/openshift/origin/test/extended/util/image"
 )
 
 func cliPodWithPullSecret(cli *exutil.CLI, shell string) *kapiv1.Pod {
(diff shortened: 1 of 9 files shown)


rvanderp3:
/hold

A possible regression in e2e-aws/e2e-gcp needs to be addressed.  Will remove hold once done.

PR #26731, opened 2022-01-06
Round before comparing disruption

deads2k:
builds on https://github.com/openshift/origin/pull/26730, only the last commit is unique

round before we compare the disruption values.  We aren't really accurate below two seconds (if I remember nyquist correctly), so the rounding is fine and makes the comparison make more sense.  I think the p95 is still accurate enough because it has so many combined results.

This is to avoid messages like 
> Jan  6 14:35:12.184: disruption/service-load-balancer-with-pdb connection/reused was unreachable during disruption for at least 2s of 30m57s (maxAllowed=2s):

which are nonsensical.  That's close enough for this test I think.


deads2k:
no conflict, merging.

PR #26736, opened 2022-01-11
Bug 2026802: copy api-request upperbound for cluster-monitoring-operator

sthaha:
This patch addresses the bug by increasing the api-request upperbound of
cluster-monitoring-operator by copying the values from `master` for only
those platforms where the upperbound is lower than in `master`.
I.E. This only updates the upperbound of AWS, Azure, and GCP and not
BareMetal, VSphere and OpenStack since the ones in release-4.9 is higher
than the ones in master branch.

Signed-off-by: Sunil Thaha <sthaha@redhat.com>


sthaha:
@ingvagabund I see that you have worked extensively on setting the api-request limits. Could you please see if this looks good to you?

ingvagabund:
@sthaha thank you for updating the upper bounds!!! The upper bounds set on the master branch are more realistic estimate. Unfortunately, the bounds are still changing so we might need another update in 4.9 eventually.

/approve
/lgtm

sthaha:
@ingvagabund Thank you for the review, I think `/backport-risk-assessed` is a no-op 

isn't it `/label backport-risk-assessed` ?

soltysh (approved):
/approve
/label backport-risk-assessed

ingvagabund:
@juzhao would you mind labeling the PR with `cherry-pick-approved` to complete the qa part?

test/extended/apiserver/api_requests.go
```diff
@@ -115,7 +115,7 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 				"cluster-autoscaler-operator":            44,
 				"cluster-baremetal-operator":             31,
 				"cluster-image-registry-operator":        119,
-				"cluster-monitoring-operator":            33,
+				"cluster-monitoring-operator":            124,
```
bparees, line 118: this is a big increase.  is it well understood why the number of watches increased so much?

ingvagabund: 4.9 threshold was computed by taking maximum from the last N jobs. 4.10 uses percentiles computed from jobs ran in the last 2 months. Still, both approaches are experimental. Unfortunately, there's currently no automation that would adjust the bounds periodically. So it is possible the 4.9 upper bound for CMO is more than enough.

PR #26737, opened 2022-01-11
Bug 2038389: Fix vsphere testing workarounds

gnufied:
We changed the operator that is responsible for marking cluster as un-upgradeable. so update the e2e workarounds accordingly for UPI installs.

xref https://bugzilla.redhat.com/show_bug.cgi?id=2038389

### DIFF
--- a/pkg/synthetictests/duplicated_events.go
+++ b/pkg/synthetictests/duplicated_events.go
@@ -78,7 +78,7 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
 	// currently UPI install defaults to HW 13 which causes events in 4.10 CI
 	// since 4.10 still supports vSphere 6.5, we can't default to HW 15 in RHCOS image but such clusters are unupgradable to 4.11 and hence
 	// events are still valid.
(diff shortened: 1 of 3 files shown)


pkg/synthetictests/duplicated_events.go
```diff
@@ -78,7 +78,7 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
 	// currently UPI install defaults to HW 13 which causes events in 4.10 CI
 	// since 4.10 still supports vSphere 6.5, we can't default to HW 15 in RHCOS image but such clusters are unupgradable to 4.11 and hence
 	// events are still valid.
-	regexp.MustCompile(`ns/openshift-cluster-storage-operator\s+deployment/vsphere-problem-detector-operator - reason/VSphereOlderVersionDetected.+vmx-13`),
+	regexp.MustCompile(`ns/openshift-cluster-csi-drivers\s+deployment/vmware-vsphere-csi-driver-operator - reason/check_deprecated_hw_version.+vmx-13`),
```
gnufied, line 81: Once https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/65 merges, we will not need this workaround.

jsafrane: It has merged, should we remove the regexp?

gnufied: done.

PR #26739, opened 2022-01-12
test/extended: check apiserver request latency

dgrisonnet:
Add test verifying that the apiserver request latency isn't heavily impacted by changes.

### DIFF
--- a/test/extended/apiserver/latency.go
+++ b/test/extended/apiserver/latency.go
@@ -14,38 +14,41 @@ import (
 	promhelper "github.com/openshift/origin/test/extended/util/prometheus"
 )
 
+const (
+	apiserverReadVerbs  = "GET|LIST"
+	apiserverWriteVerbs = "POST|PUT|PATCH|DELETE"
+)
+
 var _ = g.Describe("[sig-api-machinery][Late] API Server latency", func() {
 	defer g.GinkgoRecover()
 
 	oc := exutil.NewCLIWithoutNamespace("apiserver-latency")
 
-	// TODO: refine latency thresholds based on actual CI run results
-	//
+	// Latency thresholds defined below are based on the following CI jobs:
+	// - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26739/pull-ci-openshift-origin-master-e2e-gcp/1481905976666755072
+	// - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26739/pull-ci-openshift-origin-master-e2e-gcp/1481879797410828288
+	// - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26739/pull-ci-openshift-origin-master-e2e-aws-fips/1481879797377273856
+	// - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26739/pull-ci-openshift-origin-master-e2e-aws-fips/1481905976616423424
 	// Latency thresholds should be calculed based on the following queries:
-	// - histogram_quantile(0.99,sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",scope=~"resource|"}[$testDuration])) by (le,verb))
-	// - histogram_quantile(0.99,sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",scope="namespace"}[$testDuration])) by (le,verb))
-	// - histogram_quantile(0.99,sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",scope="cluster"}[$testDuration])) by (le,verb))
+	// - histogram_quantile(0.99,sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"GET|LIST"}[$testDuration])) by (le,scope))
+	// - histogram_quantile(0.99,sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"POST|PUT|PATCH|DELETE"}[$testDuration])) by (le,scope))
 	g.It("should verify that the apiserver requests latency are within expected thresholds", func() {
 		promClient := oc.NewPrometheusClient(context.TODO())
 
 		// 99th percentile of resource-scoped requests latency
 		scope := "resource|"
-		expectRequestLatency(promClient, "GET", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "POST", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "PUT", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "PATCH", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "DELETE", scope, parseDuration("100ms"))
+		expectRequestLatency(promClient, apiserverReadVerbs, scope, parseDuration("200ms"))    // got 130ms
+		expectRequestLatency(promClient, apiserverWriteVerbs, scope, parseDuration("1s250ms")) // got 850ms
 
 		// 99th percentile of namespace-scoped requests latency
 		scope = "namespace"
-		expectRequestLatency(promClient, "LIST", scope, parseDuration("500ms"))
-		expectRequestLatency(promClient, "POST", scope, parseDuration("500ms"))
-		expectRequestLatency(promClient, "DELETE", scope, parseDuration("500ms"))
+		expectRequestLatency(promClient, apiserverReadVerbs, scope, parseDuration("150ms"))    // got 100ms
+		expectRequestLatency(promClient, apiserverWriteVerbs, scope, parseDuration("1s350ms")) // got 900ms
 
 		// 99th percentile of cluster-scoped requests latency
 		scope = "cluster"
-		expectRequestLatency(promClient, "LIST", scope, parseDuration("5s"))
-		expectRequestLatency(promClient, "POST", scope, parseDuration("5s"))
+		expectRequestLatency(promClient, apiserverReadVerbs, scope, parseDuration("1s350ms")) // got 900ms
+		expectRequestLatency(promClient, apiserverWriteVerbs, scope, parseDuration("300ms"))  // got 200ms
 	})
 })
 
@@ -60,7 +63,7 @@ func expectRequestLatency(prometheusClient prometheusv1.API, verb string, scope 
 		context.TODO(),
 		prometheusClient,
 		fmt.Sprintf(
-			`histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb="%s",scope=~"%s"}[%s])) by (le))`,
+			`histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"%s",scope=~"%s"}[%s])) by (le))`,
 			verb, scope, exutil.DurationSinceStartInSeconds().String(),
 		),
 	)
--- a/test/extended/apiserver/latency.go
+++ b/test/extended/apiserver/latency.go
@@ -1,0 +1,75 @@
+package apiserver
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+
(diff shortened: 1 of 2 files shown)


test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,79 @@
...
+		defer func() {
+			oc.AdminKubeClient().CoreV1().Pods(ns).Delete(context.Background(), execPod.Name, *metav1.NewDeleteOptions(1))
+		}()
+
+		// consider series since the beginning of the test
+		testDuration := exutil.DurationSinceStartInSeconds().String()
+
+		resp, err := promhelper.RunQuery(
+			fmt.Sprintf(`
+				rate(apiserver_request_duration_seconds_sum{job="apiserver",verb=~"LIST|GET"}[%s])
```
deads2k, line 52: how about we separate the verbs: get, list, update, patch, delete should all be relatively constant time.

deads2k, line 52: can I get a comment telling me what a rate divided by a rate gives me?

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,85 @@
...
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	promhelper "github.com/openshift/origin/test/extended/util/prometheus"
+)
+
+var (
+	oc                   *exutil.CLI
+	url, bearerToken, ns string
+	execPod              *v1.Pod
+)
```
dgrisonnet, line 22: These will either be removed or moved to a function once https://github.com/openshift/origin/pull/26740 is merged. I just exposed them in the meantime to avoid changing the signature of `expectRequestLatency` when the PR will be merged.

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,84 @@
...
+		expectRequestLatency("DELETE", scope, 0.75)
+
+		// 99th percentile of cluster-scoped requests latency
+		scope = "cluster"
+		expectRequestLatency("LIST", scope, 1)
+		expectRequestLatency("POST", scope, 0.5)
+	})
+})
+
+func expectRequestLatency(verb string, scope string, threshold float64) {
```
deads2k, line 72: is threshold logically a time.Duration?  I think it would help me to see it as time.Duration and this function manipulates it however is required.

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,84 @@
...
+		fmt.Sprintf(
+			`histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb="GET",scope=~"%s"}[%s])) by (le))`,
+			scope, exutil.DurationSinceStartInSeconds().String()
+		),
+		ns, execPod.Name, url, bearerToken,
+	)
+
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(resp.Data.Result).NotTo(o.BeEmpty())
+	o.Expect(resp.Data.Result[0].Value).Should(o.BeNumerically("<"), threshold)
```
deads2k, line 83: we want to control this message very exactly so that it indicates somethign like

"p99 of requests like this (for example: list of configmaps across all namespaces) have a historical p99 of duration.  We set the threshold to other-duration.  This job run has a p99 of third-duration"

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,84 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		var ok bool
+		url, _, bearerToken, ok = promhelper.LocatePrometheus(oc)
+		if !ok {
+			e2e.Failf("Prometheus could not be located on this cluster, failing prometheus test")
+		}
+	})
+
+	g.It("should verify that the apiserver requests latency are within expected thresholds", func() {
```
deads2k, line 50: provide a comment that indicates where you got these actual p99s. Not so I can criticize it, but so that you can collect it again.

test/extended/apiserver/latency.go
dgrisonnet: I changed it to:
```go
	o.Expect(latency).Should(o.BeNumerically("<"), threshold,
		fmt.Sprintf("expected 99th percentile of apiserver request latency with verb=%q and scope=%q to be inferior to %s, got %s", verb, scope, threshold, latency),
	)
```
Doing it per resource would be very tedious here. To me, this should be a very high-level warning that something is degraded and further investigation is needed to figure exactly what.
Since the investigation is far from being trivial as we've seen during the past couple of days, it might be worth attaching a link explaining the steps that need to be taken in order to figure out what is wrong, similar to the document I created. wdyt?

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,101 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		var ok bool
+		url, _, bearerToken, ok = promhelper.LocatePrometheus(oc)
+		if !ok {
+			e2e.Failf("Prometheus could not be located on this cluster, failing prometheus test")
+		}
+	})
+
+	// Lantecy thresholds collected from:
```
deads2k, line 45: Can you specify the query you used?  I think that will give me enough, when combined with these links, that I could rebuild what you got.

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,101 @@
...
+		),
+		ns, execPod.Name, url, bearerToken,
+	)
+
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(resp.Data.Result).NotTo(o.BeEmpty())
+
+	latency := time.Duration(float64(resp.Data.Result[0].Value) * float64(time.Second))
+	o.Expect(latency).Should(o.BeNumerically("<"), threshold,
+		fmt.Sprintf("expected 99th percentile of apiserver request latency with verb=%q and scope=%q to be inferior to %s, got %s", verb, scope, threshold, latency),
```
deads2k, line 99: to be less than

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,101 @@
...
+
+		// TODO: access Prometheus via route instead of exec pod
+		execPod = exutil.CreateExecPodOrFail(oc.AdminKubeClient(), ns, "execpod")
+		defer func() {
+			oc.AdminKubeClient().CoreV1().Pods(ns).Delete(context.Background(), execPod.Name, *metav1.NewDeleteOptions(1))
+		}()
+
+		// 99th percentile of resource-scoped requests latency
+		scope := "resource|"
+		expectRequestLatency("GET", scope, parseDuration("75ms"))
```
deads2k, line 60: can you indicate the actuals you got so that when this later fails and Abu or Stefan say, "this isn't significantly worse", I can say, "yes, it is significantly worse, see this historical value from 1/15/2022".

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,101 @@
...
+func parseDuration(s string) time.Duration {
+	d, err := time.ParseDuration(s)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return d
+}
+
+func expectRequestLatency(verb string, scope string, threshold time.Duration) {
+	resp, err := promhelper.RunQuery(
+		fmt.Sprintf(
+			`histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb="GET",scope=~"%s"}[%s])) by (le))`,
```
deads2k, line 88: why is this hardcoded to GET instead of substituting "verb"

deads2k:
I think this is close to being ready.

test/extended/apiserver/latency.go
dgrisonnet: done

test/extended/apiserver/latency.go
dgrisonnet: make sense

test/extended/apiserver/latency.go
dgrisonnet: :facepalm: thank you for catching that :laughing:

dgrisonnet:
Since the values that I observed in CI were varying a lot depending on the infrastructure and the kind of the job, I would take it slowly and first try to test these thresholds across all platforms before breaking anyone. To hasten the process, It might be worth adding this test as flaky for now until we have figured out thresholds that won't break anything and still be efficient, wdyt?

deads2k:
you need to hack/update-generate

deads2k:
/lgtm

thresholds can be reassessed based on actual data, but if they get looser that data should be based on runs from *before* the kube update.

dgrisonnet:
I rebased the PR on top of https://github.com/openshift/origin/pull/26740, PTAL @deads2k

dgrisonnet:
Holding to find more appropriate thresholds for the time being.

/hold

test/extended/apiserver/latency.go
```diff
@@ -14,38 +14,41 @@ import (
...
 
 		// 99th percentile of resource-scoped requests latency
 		scope := "resource|"
-		expectRequestLatency(promClient, "GET", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "POST", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "PUT", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "PATCH", scope, parseDuration("100ms"))
-		expectRequestLatency(promClient, "DELETE", scope, parseDuration("100ms"))
+		expectRequestLatency(promClient, apiserverReadVerbs, scope, parseDuration("200ms")) // got 130ms in CI
+		expectRequestLatency(promClient, apiserverWriteVerbs, scope, parseDuration("1s"))   // threshold defined by sig-scalability
```
deads2k, line 41: the point of this is to latch *our* thresholds based on our observed CI behavior.  Not to mirror something generic.

I'm willing to start nearly anywhere though.

dgrisonnet: Updated

deads2k:
```


2/549 Tests Failed. | expand_less
-- | --
: [sig-instrumentation][Late] API Server latency should verify that the apiserver requests latency are within expected thresholds [Suite:openshift/conformance/parallel] expand_less0sfail [github.com/openshift/origin/test/extended/apiserver/latency.go:72]: Expected     <model.Vector \| len:0, cap:0>: [] not to be empty | : [sig-instrumentation][Late] API Server latency should verify that the apiserver requests latency are within expected thresholds [Suite:openshift/conformance/parallel] expand_less | 0s | fail [github.com/openshift/origin/test/extended/apiserver/latency.go:72]: Expected     <model.Vector \| len:0, cap:0>: [] not to be empty
: [sig-instrumentation][Late] API Server latency should verify that the apiserver requests latency are within expected thresholds [Suite:openshift/conformance/parallel] expand_less | 0s
fail [github.com/openshift/origin/test/extended/apiserver/latency.go:72]: Expected     <model.Vector \| len:0, cap:0>: [] not to be empty


```

test/extended/apiserver/latency.go
```diff
@@ -0,0 +1,78 @@
...
+	exutil "github.com/openshift/origin/test/extended/util"
+	promhelper "github.com/openshift/origin/test/extended/util/prometheus"
+)
+
+const (
+	apiserverReadVerbs  = "GET|LIST"
+	apiserverWriteVerbs = "POST|PUT|PATCH|DELETE"
+)
+
+var _ = g.Describe("[sig-instrumentation][Late] API Server latency", func() {
```
deads2k, line 22: sig-api-machinery

dgrisonnet:
There was a problem with the query, it should be good now

deads2k:
/lgtm

in slack we agreed to next steps: https://coreos.slack.com/archives/C02NZBANL3G/p1642169820339000?thread_ts=1642110108.326700&cid=C02NZBANL3G


David Eads  3 hours ago
so, breaking it into steps
1. I'll merge pretty much whatever you want for 26739
2. we need a perma-flaking test that always outputs the number we're looking for in some way that we can automatically parse
3. we backport the PR from #2 to 4.9
4. in 4.10, we produce a cleaner, "write this data out" mechanism using the method I showed you for alerts.
5. we write the uploader in ci-tools and write the data into bigquery
6. we query it back out
does that sound good?


David Eads  3 hours ago
your choice on 26739.  At this point, something is better than nothing there

Damien Grisonnet  3 hours ago
yeah that's exactly what I was thinking

dgrisonnet:
PTAL @deads2k

PR #26711, opened 2021-12-17
Bug 2033751: Rebase 1.23.0

damemi:
Depends on https://github.com/openshift/kubernetes/pull/1087

PR #26740, opened 2022-01-12
use route to access prometheus instead of exec pods.

deads2k:
This reduces the dependency chain and eliminates a source of error.  Just use the route directly.

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+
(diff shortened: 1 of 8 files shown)


tkashem:
why did we use exec pods originally, any historical context?

/lgtm

tkashem:
waiting for https://github.com/openshift/origin/pull/26711 to merge first

/hold

deads2k:
alert query still work!

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26740/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1481296112697479168

pushed additional commit to handle initialization without clientconfig. no longer requires updated library-go. removing hold and retagging.

deads2k:
bypassing bug requirement to address CI problem from kube update

PR #26741, opened 2022-01-12
watch specific alerts for pending and firing

deads2k:
this lets us track burn-budget directly.

### DIFF
--- a/pkg/synthetictests/allowedalerts/basic_alert.go
+++ b/pkg/synthetictests/allowedalerts/basic_alert.go
@@ -1,0 +1,285 @@
+package allowedalerts
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/openshift/origin/pkg/monitor"
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/synthetictests/platformidentification"
+	"github.com/openshift/origin/pkg/test/ginkgo"
+	testresult "github.com/openshift/origin/pkg/test/ginkgo/result"
+	exutil "github.com/openshift/origin/test/extended/util"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+type AlertTest interface {
+	// TestNamePrefix is the prefix for this as a late test: [bz-component][late] something
+	TestNamePrefix() string
+	// LateTestNameSuffix is name for this as a late test: alert/foo should not be pending
+	LateTestNameSuffix() string
+	// InvariantTestName is name for this as an invariant test
+	InvariantTestName() string
+
+	// AlertName is the name of the alert
+	AlertName() string
+	// AlertState is the threshold this test applies to.
+	AlertState() AlertState
+
+	TestAlert(ctx context.Context, prometheusClient prometheusv1.API, restConfig *rest.Config) error
+	InvariantCheck(ctx context.Context, restConfig *rest.Config, intervals monitorapi.Intervals) ([]*ginkgo.JUnitTestCase, error)
+
+	// FailAfter is the amount of time that an alert can be at or above the current state before failing a test
+	FailAfter(jobType platformidentification.JobType) time.Duration
+	// FlakeAfter is the amount of time that an alert can be at or above the current state before flaking a test
+	FlakeAfter(jobType platformidentification.JobType) time.Duration
+}
+
+// AlertState is the state of the alert. They are logically ordered, so if a test says it limits on "pending", then
+// any state above pending (like info or warning) will cause the test to fail.
+type AlertState string
+
+const (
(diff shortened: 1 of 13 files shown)


deads2k:
bypassing bug requirement to address CI problem from kube update

pkg/synthetictests/allowedbackenddisruption/matches.go
```diff
@@ -7,73 +7,19 @@ import (
...
-	topology := ""
-	switch infrastructure.Status.ControlPlaneTopology {
-	case configv1.HighlyAvailableTopologyMode:
-		topology = "ha"
-	case configv1.SingleReplicaTopologyMode:
-		topology = "single"
-	}
 
-	return GetClosestP95Value(backendName, release, fromRelease, platform, networkType, topology), nil
+	return GetClosestP95Value(backendName, jobType.Release, jobType.FromRelease, jobType.Platform, jobType.Network, jobType.Topology), nil
```
dgoodwin, line 22: Was this mistakenly lumped into this PR? Looks quite separate.

pkg/synthetictests/platformidentification/types.go
```diff
@@ -0,0 +1,101 @@
...
+	if len(versionParts) < 2 {
+		return ""
+	}
+
+	version := versionParts[0] + "." + versionParts[1]
+	if strings.HasPrefix(version, "v") {
+		version = version[1:]
+	}
+	return version
+}
```
dgoodwin, line 101: Separate PR?

pkg/synthetictests/allowedalerts/basic_alert.go
```diff
@@ -0,0 +1,303 @@
...
+
+	// FailAfter is the amount of time that an alert can be at or above the current state before failing a test
+	FailAfter(jobType platformidentification.JobType) time.Duration
+	// FlakeAfter is the amount of time that an alert can be at or above the current state before flaking a test
+	FlakeAfter(jobType platformidentification.JobType) time.Duration
+}
+
+// AlertState is the state of the alert. They are logically ordered, so if a test says it limits on "pending", then
+// any state above pending (like info or warning) will cause the test to fail.
+type AlertState string
```
dgoodwin, line 50: Would it be helpful to have these map to integers for a programatic way to compare if a state is above your threshold, instead of having to hard code those assumptions when used? Seems like someone may miss that distinction when using at some point especially with info after pending, and this may be error prone to maintain. Those assumptions appear in couple places in this PR alone. 

Alternatively an AtOrAboveAlertStateThreshold func on the interface might do the trick.

pkg/synthetictests/allowedalerts/basic_alert.go
```diff
@@ -0,0 +1,303 @@
...
+	pendingDuration := pendingIntervals.Duration(1 * time.Second)
+
+	jobType, err := platformidentification.GetJobType(ctx, restConfig)
+	if err != nil {
+		return fail, err.Error()
+	}
+
+	switch {
+	case durationAtOrAboveLevel > a.FailAfter(*jobType):
+		return fail, fmt.Sprintf("%s was at or above %s at least %s (maxAllowed=%s): pending for %s, firing for %s:\n\n%s",
```
dgoodwin, line 227: "for at least"

pkg/synthetictests/allowedalerts/basic_alert.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+type AlertTest interface {
+	// TestNamePrefix is name for this as a late test
+	TestNamePrefix() string
+	// LateTestNameSuffix is name for this as a late test
+	LateTestNameSuffix() string
```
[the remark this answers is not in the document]
dgoodwin: Presumably run on test data after the suite is completed.

pkg/synthetictests/allowedbackenddisruption/matches.go
deads2k: > Was this mistakenly lumped into this PR? Looks quite separate.

it unifies the detection of job categorization.  not a mistake.

pkg/synthetictests/platformidentification/types.go
deads2k: > Separate PR?

required to pass the job categorization criteria to the "how long a duration do you tolerate".

pkg/synthetictests/allowedalerts/basic_alert.go
deads2k: Where would such an integer be used in this PR?

dgoodwin: I was thinking in the switches/filtering where we take level X and above, but if this doesn't trigger any ideas that would be an improvement then disregard.

deads2k:
comments addressed.

deads2k:
doesn't conflict with the wardle fix.  This will help unlock payload promotion. forciing.

PR #26744, opened 2022-01-13
Record zero for known alerts

deads2k:
We have a list of known alerts now that is fairly easy to add to.  This is useful to allow us to indicate those runs where this alert did not fire.  This will allow us to find proper percentile values.

/assign @dgoodwin


dgoodwin (approved):
/lgtm

deads2k:
no conflicts, green tests, forcing in so we get full alert data over the weekend.

PR #26745, opened 2022-01-14
Display ErrImagePull messages.

dgoodwin:
Previously when listing the image pull problems we only showed the
location of the problem pod, but we do not see the actual message. As
such it's not possible to tell where we're pulling from.

Modified to include the interval message on each line which should
indicate where the pull problem was from.


deads2k:
/lgtm

This is gathering more debug information for existing problems.  merging

PR #26746, opened 2022-01-14
for upgrades, skip alerts with individual invariant tests

deads2k:
found in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26739/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1481993338713280512

/cc @dgrisonnet 
/assign @dgoodwin

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -6,16 +6,14 @@ import (
 	"strings"
 	"time"
 
-	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
-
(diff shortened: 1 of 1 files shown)


deads2k:
got comments in slack with a few ious.  Merging to hit next payload test.

dgrisonnet:
As a follow-up to this PR, it would be great to either mention which test is replacing an alert instead of saying `has a separate e2e test` or at least have comments in https://github.com/openshift/origin/blob/master/pkg/synthetictests/allowedalerts/all.go to make sure we keep track of the link between tests and allowed alerts.

PR #26747, opened 2022-01-14
allow burn budget to flake until kube-apiserver team finds and fixes it

deads2k:
/hold

for confirmation of understand from @tkashem and @wallylewis that this 
1. does not mean that https://bugzilla.redhat.com/show_bug.cgi?id=2039539 is not a blocker bug
2. does not mean that an appropriate fix is change the alert.  https://github.com/openshift/cluster-kube-apiserver-operator/pull/1284 fixes a symptom, but does not address a cause
3. The bug does mean that the source of latency needs to be found and either fixed or tightly categorized for impact.  p&f width/seats on lists is something @deads2k suspects.

The purpose of this PR is to unblock our merge queue, while leaving an "easy" signal to find for prospective fixes.  If the issue is fixed, then job runs will stop showing this as a flake.

### DIFF
--- a/pkg/synthetictests/allowedalerts/all.go
+++ b/pkg/synthetictests/allowedalerts/all.go
@@ -22,7 +22,7 @@ func AllAlertTests() []AlertTest {
 		newAlert("etcd", "etcdHighNumberOfLeaderChanges").firing(),
 
 		newAlert("kube-apiserver", "KubeAPIErrorBudgetBurn").pending().neverFail(),
-		newAlert("kube-apiserver", "KubeAPIErrorBudgetBurn").firing(),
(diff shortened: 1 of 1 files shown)


deads2k:
with this PR, the process for proving a PR fixes the latency problem is, "get aggregated runs and check each run for the burnbudget alert flake.  If you don't get burn budget alert flakes, you're fixed"

deads2k:
got an ack in slack from wally too https://coreos.slack.com/archives/C02NZBANL3G/p1642184474366000

/hold cancel

deads2k:
merging to relieve  the merge queue post kube update.

PR #26749, opened 2022-01-14
handle refactoring concerns on the alert separation

deads2k:
handling debt from other PRs.

### DIFF
--- a/pkg/synthetictests/allowedalerts/all.go
+++ b/pkg/synthetictests/allowedalerts/all.go
@@ -5,36 +5,36 @@ func AllAlertTests() []AlertTest {
 		newWatchdogAlert(),
 
 		newAlert("etcd", "etcdMembersDown").pending().neverFail(),
-		newAlert("etcd", "etcdMembersDown").firing(),
+		newAlert("etcd", "etcdMembersDown").info(),
 		newAlert("etcd", "etcdGRPCRequestsSlow").pending().neverFail(),
-		newAlert("etcd", "etcdGRPCRequestsSlow").firing(),
+		newAlert("etcd", "etcdGRPCRequestsSlow").info(),
 		newAlert("etcd", "etcdHighNumberOfFailedGRPCRequests").pending().neverFail(),
-		newAlert("etcd", "etcdHighNumberOfFailedGRPCRequests").firing(),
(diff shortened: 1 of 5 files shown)


pkg/synthetictests/allowedalerts/all.go
```diff
@@ -3,33 +3,33 @@ package allowedalerts
...
 		newAlert("etcd", "etcdInsufficientMembers").pending().neverFail(),
-		newAlert("etcd", "etcdInsufficientMembers").firing(),
+		newAlert("etcd", "etcdInsufficientMembers").info(),
 		newAlert("etcd", "etcdHighNumberOfLeaderChanges").pending().neverFail(),
-		newAlert("etcd", "etcdHighNumberOfLeaderChanges").firing(),
+		newAlert("etcd", "etcdHighNumberOfLeaderChanges").info(),
 
 		newAlert("kube-apiserver", "KubeAPIErrorBudgetBurn").pending().neverFail(),
-		newAlert("kube-apiserver", "KubeAPIErrorBudgetBurn").firing().neverFail(), // https://bugzilla.redhat.com/show_bug.cgi?id=2039539
+		newAlert("kube-apiserver", "KubeAPIErrorBudgetBurn").info().neverFail(), // https://bugzilla.redhat.com/show_bug.cgi?id=2039539
```
dgrisonnet, line 25: The refactoring looks good to me, but I am curious about this line. The `KubeAPIErrorBudgetBurn` alert has two severities either `critical` or `warning` so if I am not wrong, this line doesn't do anything since it allows the info level alert 
that doesn't exist to fire?

deads2k: > The refactoring looks good to me, but I am curious about this line. The `KubeAPIErrorBudgetBurn` alert has two severities either `critical` or `warning` so if I am not wrong, this line doesn't do anything since it allows the info level alert
> that doesn't exist to fire?

`.info` fails when it is at or above info level.  otherwise you can get weird results like passing when the alert is critical.  this line is what is adding the flake to so many tests today.

dgrisonnet: Ah make sense, I didn't understood it correctly in the first place, thank you for clarifying.

PR #26755, opened 2022-01-17
Alter junit test suite names for jobs with multiple junit xmls.

DennisPeriquet:
Reverts [26686](https://github.com/openshift/origin/pull/26686)

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -306,7 +306,7 @@ func newRunCommand() *cobra.Command {
 				if !opt.DryRun {
 					fmt.Fprintf(os.Stderr, "%s version: %s\n", filepath.Base(os.Args[0]), version.Get().String())
 				}
-				err = opt.Run(&suite.TestSuite)
+				err = opt.Run(&suite.TestSuite, "openshift-tests")
 				if suite.PostSuite != nil {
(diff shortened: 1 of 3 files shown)


deads2k:
/approve

@dgoodwin if you could summarize your plan for handling this as we land, I'd feel better.  I'm fine with a few bumps, but I don't want a week of them.

deads2k:
This allows us to distinguish different failures for the same testcase name from two different runs.  upgrade-conformance runs an upgrade suite and an e2e parallel suite and we need to distinguish the different testcases so we can properly build the historical data.

deads2k:
I have changed  my mind. this is breaking our test grid data on https://testgrid.k8s.io/redhat-openshift-ocp-release-4.10-informing#periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade&show-stale-tests=

And that causes it to hide perma-fails.  I can see my way to a path where this works.  I'll give @DennisPeriquet first crack.

/lgtm

dgoodwin:
My plan was we bring this in, and then update aggregation to skip the now failing test as we have recently for several others.

dgoodwin:
We will need a manual label to merge to origin.

deads2k:
required for accurate aggregation.  serial is failing on the lokie event exporter.  forcing in.

PR #26757, opened 2022-01-17
make it easy to create multiple flavors of interval charts

deads2k:
refactors interval html creation to optimize different views for different purposes


deads2k:
making unique debug data easier to gather.

PR #26759, opened 2022-01-17
Add test for OS updates being staged within 5-10m.

dgoodwin:
Recent test failure investigation led to a possibility of slow disk
speeds doing OS content extraction. Add a test to ensure that we move
from OSUpdateStarted to OSUpdateStaged within 5-10 minutes, which
primarily just encompasses extraction of the content.

If any node is over 10 minutes we will fail the test, if some are over 5
we'll flake instead.

### DIFF
--- a/pkg/synthetictests/operators.go
+++ b/pkg/synthetictests/operators.go
@@ -73,6 +73,89 @@ func testOperatorStateTransitions(events monitorapi.Intervals, conditionTypes []
 	return ret
 }
 
+func testOperatorOSUpdateStaged(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	testName := "[bz-Machine Config Operator] Nodes should reach OSUpdateStaged in a timely fashion"
+	success := &junitapi.JUnitTestCase{Name: testName}
+	flakeThreshold := 5 * time.Minute
+	failThreshold := 10 * time.Minute
+
+	type startedStaged struct {
+		// OSUpdateStarted is the event Reason emitted by the machine config operator when a node begins extracting
+		// it's OS content.
+		OSUpdateStarted time.Time
+		// OSUpdateStaged is the event Reason emitted by the machine config operator when a node has extracted it's
+		// OS content and is ready to begin the update. For the purposes of this test, we're looking for how long it
+		// took from Started -> Staged to try to identify disk i/o problems that may be occurring.
+		OSUpdateStaged time.Time
+	}
+
+	// Scan all OSUpdateStarted and OSUpdateStaged events, sort by node.
+	nodeOSUpdateTimes := map[string]*startedStaged{}
+	for _, e := range events {
+		if strings.Contains(e.Message, "reason/OSUpdateStarted") {
+			// locator will be of the form: node/ci-op-j34hmfqt-253f3-cq852-master-1
+			_, ok := nodeOSUpdateTimes[e.Locator]
+			if !ok {
+				nodeOSUpdateTimes[e.Locator] = &startedStaged{}
+			}
+			// for this type of event, the from/to time are identical as this is a point in time event.
+			ss := nodeOSUpdateTimes[e.Locator]
+			ss.OSUpdateStarted = e.To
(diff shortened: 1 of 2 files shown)


pkg/synthetictests/operators.go
```diff
@@ -73,6 +73,77 @@ func testOperatorStateTransitions(events monitorapi.Intervals, conditionTypes []
 	return ret
 }
 
+func testOperatorOSUpdateStaged(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	testName := "[bz-Machine Config Operator] Nodes should reach OSUpdateStaged in a timely fashion"
+	success := &junitapi.JUnitTestCase{Name: testName}
+
+	type startedStaged struct {
+		OSUpdateStarted time.Time
```
deads2k, line 81: give me some comments about what this state means

pkg/synthetictests/operators.go
```diff
@@ -73,6 +73,77 @@ func testOperatorStateTransitions(events monitorapi.Intervals, conditionTypes []
...
+		OSUpdateStaged  time.Time
```
deads2k, line 82: give me some comments about what this state means

deads2k:
/approve

a request for some comments and lgtm

pkg/synthetictests/operators.go
```diff
@@ -73,6 +73,82 @@ func testOperatorStateTransitions(events monitorapi.Intervals, conditionTypes []
...
+
+	// Iterate the data we assembled looking for any nodes with an excessive time between Started/Staged, or those
+	// missing a Staged
+	slowStageMessages := []string{}
+	var failTest bool // set true if we see anything over 10 minutes, our failure threshold
+	for node, ss := range nodeOSUpdateTimes {
+		if ss.OSUpdateStaged.IsZero() {
+			slowStageMessages = append(slowStageMessages, fmt.Sprintf("%s OSUpdateStarted at %s, did not make it to OSUpdateStaged", node, ss.OSUpdateStarted))
+			failTest = true
+		} else if ss.OSUpdateStaged.Sub(ss.OSUpdateStarted) > 5*time.Minute {
```
wking, line 122: possible worth an additional case here for `ss.OSUpdateStarted.IsZero()` for "best-effort event submission must have dropped the started event, so I dunno how long this node too to get from there to staged".  I guess I'd expect that to be a flake, but should be rare enough that if we decide to fail on it the number of runs that fail as a result should be small.

dgoodwin:
@wking has pointed out that we may have suites that do multi-stage updates, meaning we move through these conditions several times. I'm also not sure this currently properly handles install time events. Will revisit asap tomorrow.

pkg/synthetictests/operators.go
```diff
@@ -73,6 +73,82 @@ func testOperatorStateTransitions(events monitorapi.Intervals, conditionTypes []
...
+
+			// Separate check, did we see anything over 10 minutes, in which case we want to fail the test
+			if ss.OSUpdateStaged.Sub(ss.OSUpdateStarted) > 10*time.Minute {
+				failTest = true
+			}
+		}
+
+	}
+	if len(slowStageMessages) > 0 {
+		output := fmt.Sprintf("%d nodes took over 5 minutes to stage OSUpdate:\n\n%s",
```
wking, line 134: might be a local variable for `5*time.Minute` that we can format in here, so folks don't bump the threshold but forget to bump this output string.

deads2k:
> @wking has pointed out that we may have suites that do multi-stage updates, meaning we move through these conditions several times. I'm also not sure this currently properly handles install time events. Will revisit asap tomorrow.

followup is acceptable to me.

deads2k:
cleaning up CI signal on tests

dgoodwin:
/hold 

Something's wrong

dgoodwin:
Fixed a pointer bug causing zero start times and empty staged times, added vars for the flake/fail thresholds, added the check for edge case on zero start time (which will fail the test for now), formatted timestamps better for printing, and *may* have handled test suites with multiple update cycles.

PR #26765, opened 2022-01-18
localResourceAccessReviewTest: allow event explorer

vrutkovs:
Skip event-explorer SA from loki namespace - it records events from all namespaces. This was introduced in https://github.com/openshift/release/pull/25344.

The alternative to this PR is https://github.com/openshift/release/pull/25376


vrutkovs:
https://github.com/openshift/release/pull/25376 is preferred

PR #26770, opened 2022-01-20
Bug 1980127: delete sdn binary in goroutine when SIGTERM is provided

JacobTanenbaum:
arguments cannot be passed to preStop hooks, delete the binary when sdn
process exits

### DIFF
--- a/pkg/network/node/node.go
+++ b/pkg/network/node/node.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package node
@@ -7,9 +8,11 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


pkg/network/node/node.go
```diff
@@ -388,6 +390,13 @@ func (node *OsdnNode) Start() error {
 	if len(podName) > 0 {
 		cniType = cniType + "-" + podName
 	}
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		os.Remove("/host/opt/cni/bin/openshift-sdn-" + podName)
+		<-sig
```
danwinship, line 397: This is backwards isn't it? You want to wait for the signal and then remove the file, not remove the file and then wait for the signal.

The sdn binary doesn't currently know anything about the path `"/host/opt/cni/bin/"` so it's awkward to suddenly add this here... but I guess 3.11 is supposed to be dead soon so, meh.

danwinship: also, this won't remove the file if the sdn binary crashes...

maybe you could just have the DaemonSet clean old binaries at startup, rather than trying to clean itself up at shutdown? ie,

```
          rm -f /host/opt/cni/bin/openshift-sdn-*
          cp -f /opt/cni/bin/openshift-sdn /host/opt/cni/bin/openshift-sdn-${OPENSHIFT_SDN_POD}

```

danwinship: (I haven't thought about this bug in a while, so double-check my logic that this won't just reintroduce the original race condition...)

JacobTanenbaum: @danwinship did you get the chance to think about this? deleting on startup MIGHT work but I think there was a reason why I just can't remember it at the moment either.

danwinship: Oh, no, I meant _you_ should double-check my logic, since you've been more closely involved with this.

PR #26771, opened 2022-01-20
allow loki target to be down

deads2k:
prevent failures like https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade/1484205905842016256

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -108,7 +108,12 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 			},
 		},
 	}
-	allowedFiringAlerts := helper.MetricConditions{}
+	allowedFiringAlerts := helper.MetricConditions{
+		{
+			Selector: map[string]string{"alertname": "TargetDown", "namespace": "openshift-e2e-loki"},
+			Text:     "Loki is nice to have, but we can allow it to be down",
+		},
+	}
 
 	pendingAlertsWithBugs := helper.MetricConditions{
 		{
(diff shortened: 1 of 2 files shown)


test/e2e/upgrade/alert/alert.go
```diff
@@ -108,7 +108,12 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 			},
 		},
 	}
-	allowedFiringAlerts := helper.MetricConditions{}
+	allowedFiringAlerts := helper.MetricConditions{
+		{
+			Selector: map[string]string{"alertname": "TargetDown", "namespace": "openshift-e2e-loki"},
+			Text:     "Loki is nice to have, but we can allow it to be down briefly",
```
wking, line 114: `TargetDown` is already [`for: 15m`][1], so not clear to me that we can claim "briefly" when allowing it to fire.  Dropping "briefly" and just saying `...but we can allow it to be down.` would be internally consistent.

[1]: https://github.com/openshift/cluster-monitoring-operator/blob/4c19e806220b7b2db11449c52fc024811a25f489/assets/cluster-monitoring-operator/prometheus-rule.yaml#L28

wking (approved):
Sounds like pending alerts are non-fatal today, so I don't think we need the `allowedPendingAlerts` exception, but doesn't matter too much to me, so:

/lgtm
/hold

Pull the hold if you want to keep the `allowedPendingAlerts` around in the window before we get a real fix and revert this, but looking at [the run][1], there are a whole bunch of `TargetDown` pending alerts today during various phases of the updates, and we don't have explicit exceptions for those (because we're only concerned about alerts that are still pending when the update completes).

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade/1484205905842016256

deads2k:
this built.  we want to make the next payload. merging.

PR #26772, opened 2022-01-20
adds a synthetic test for checking if the pod.spec.nodeName field is immutable

p0lyn0mial:
a simple monitor test for checking immutability of `pod.spec.nodeName` field. 
The field is not allowed to change once set i.e. a pod assigned to a node must stay on it.

It will help gather more data (if any) for https://bugzilla.redhat.com/show_bug.cgi?id=2042657

### DIFF
--- a/pkg/monitor/pod.go
+++ b/pkg/monitor/pod.go
@@ -313,6 +313,18 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
 			}
 			return conditions
 		},
+		// inform when a pod gets reassigned to a new node
+		func(pod, oldPod *corev1.Pod) []monitorapi.Condition {
+			var conditions []monitorapi.Condition
+			if len(oldPod.Spec.NodeName) > 0 && pod.Spec.NodeName != oldPod.Spec.NodeName {
+				conditions = append(conditions, monitorapi.Condition{
+					Level:   monitorapi.Error,
+					Locator: locatePod(pod),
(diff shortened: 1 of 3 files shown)


pkg/monitor/pod.go
```diff
@@ -313,6 +313,18 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
...
 			return conditions
 		},
+		// inform when a pod gets reassigned to a new node
+		func(pod, oldPod *corev1.Pod) []monitorapi.Condition {
+			var conditions []monitorapi.Condition
+			if len(oldPod.Spec.NodeName) > 0 && pod.Spec.NodeName != oldPod.Spec.NodeName {
+				conditions = append(conditions, monitorapi.Condition{
+					Level:   monitorapi.Error,
+					Locator: locatePod(pod),
+					Message: fmt.Sprintf("invariant violation, pod once assigned to a node must stay on it. The pod previously scheduled to %s, has just been assigned to a new node %s", oldPod.Spec.NodeName, pod.Spec.NodeName),
```
deads2k, line 323: nit: if this gets cleared, you will constantly wonder what that nodename is.  I suggest 
1. include the resourceversion for old and new.
2. use %q for your node names
3. inspect the field owners in metadata for old and new to determine who overwrite the value.

deads2k:
this is significantly better than what we have.  I think my suggestions are good/valuable, but if you aren't going to get to them, you can release the hold

/lgtm
/hold

deads2k:
trying to work out the scope of our problem, bypassing bug requirement

deads2k:
looks like he's out.  This is strictly better than nothing.  merging.

PR #26773, opened 2022-01-21
storage operator can go upgradeable=unknown if vsphere is unreachable

deads2k:
cleans up failures like https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.10-e2e-vsphere-upi-serial/1484206475835346944

/assign @wking @gnufied

### DIFF
--- a/test/extended/operators/operators.go
+++ b/test/extended/operators/operators.go
@@ -242,6 +242,7 @@ func surprisingConditions(co objx.Map) ([]configv1.ClusterOperatorStatusConditio
 			if cond.Get("status").String() != string(expected) {
 				reason := cond.Get("reason").String()
 				message := cond.Get("message").String()
+				status := cond.Get("status").String()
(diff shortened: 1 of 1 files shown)


test/extended/operators/operators.go
```diff
@@ -242,6 +242,7 @@ func surprisingConditions(co objx.Map) ([]configv1.ClusterOperatorStatusConditio
 			if cond.Get("status").String() != string(expected) {
 				reason := cond.Get("reason").String()
 				message := cond.Get("message").String()
+				status := cond.Get("status").String()
```
wking, line 245: nit: worth moving this up in front of the `if` so we don't have to get it twice?  And then we could also use that local `status` variable when appending to `badConditions` below, instead of getting it out a third time.

deads2k:
this is borking payload informing.  it builds, it's not worse, merging

PR #26763, opened 2022-01-17
Bug 2038481: Flake failed sandboxes from bug in new guard pods

jluhrsen:
more info in https://bugzilla.redhat.com/show_bug.cgi?id=2038481

essentially, two new guard pods are being started in 4.10+ and
are incorrectly being restarted on a cordoned node and when the
node is rebooted those pods fail to set up sandboxes right away
before the underlying network config file is present. This can
be reverted when the PR to fix this is merged:
  https://github.com/openshift/library-go/pull/1287

Signed-off-by: Jamo Luhrsen <jluhrsen@gmail.com>

PR #26776, opened 2022-01-21
Bug 2038481: Flake failed sandboxes from bug in new guard pods

deads2k:
more info in https://bugzilla.redhat.com/show_bug.cgi?id=2038481

essentially, two new guard pods are being started in 4.10+ and
are incorrectly being restarted on a cordoned node and when the
node is rebooted those pods fail to set up sandboxes right away
before the underlying network config file is present. This can
be reverted when the PR to fix this is merged:
  https://github.com/openshift/library-go/pull/1287

https://github.com/openshift/origin/pull/26763 with gofmt goodness


deads2k:
built, tested green before.

PR #26777, opened 2022-01-24
Bug 2043042: Additional checks for requestheaders.go

slaskawi:
https://bugzilla.redhat.com/show_bug.cgi?id=2043042

This Pull Request adds a generation check and a missing `X-CSRF-Token` header to the `[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [IdP]`. This should help stabilizing the test or at least, make failure investigation easier in the future.

It also adds additional check to make explicitly sure the CAO is stable before executing the test.


slaskawi:
First one went fine, let's see the second pass.

/retest

slaskawi:
Second run is fine, let's try more

/test all

slaskawi:
Fourth run is fine, last time lucky...

/test all

deads2k:
heavily tested, no conflict with other merges

PR #26781, opened 2022-01-24
use real alert data to determine flake and failure timeframes for alerts

deads2k:
This will prevent us from getting worse without noticing it, but it will not force an improvement.

### DIFF
--- a/pkg/synthetictests/allowedalerts/types.go
+++ b/pkg/synthetictests/allowedalerts/types.go
@@ -1,0 +1,111 @@
+package allowedalerts
+
+import (
+	"bytes"
+	_ "embed"
+	"encoding/json"
+	"strconv"
+	"sync"
+)
+
+const (
+	// p95Query produces the query_results.json.  Take this query and run it against bigquery, then export the results
+	// as json and place them query_results.json.
+	// This query produces the p95 and p99 alert firing seconds for the named alerts on a per platform, release, topology,
+	// network type basis.
+	p95Query = `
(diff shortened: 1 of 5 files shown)


pkg/synthetictests/allowedalerts/types.go
```diff
@@ -0,0 +1,101 @@
...
+  alertName = "etcdHighNumberOfLeaderChanges" or 
+  alertName = "KubeAPIErrorBudgetBurn" or 
+  alertName = "KubeClientErrors" or 
+  alertName = "KubePersistentVolumeErrors" or 
+  alertName = "MCDDrainError" or 
+  alertName = "PrometheusOperatorWatchErrors"
+order by 
+ AlertName, Release, FromRelease, Topology, Platform, Network
+`
+)
```
dgoodwin, line 32: This appears unused. How does one update the json file? We'd need that documented somewhere, and I do not know where that somewhere should be. It's a thing that someone should do periodically right? 

I'm actually a little worried about this going in hard-coded at all here, will we struggle with this getting out of date and not matching what's going on in aggregation? What are your thoughts around maintenance of this.

Pie in the sky, someday, what if this data was exposed in sippy via an unathenticated json API, and the test could hit that endpoint to fetch latest data.

dgoodwin (commented):
Looks like a big improvement just need to hear your thoughts on maintenance and documentation.

deads2k:
I've updated comments to explain how to build this.  I am in favor of creating a command to generate these json files that lives in openshift/origin as a jira card.  

I'm going to get this merged to clean up where we are, but I will address additional comments as followups if you leave them.

PR #26787, opened 2022-01-26
Add tests verifying that alerts expressions are valid

dgrisonnet:
Add a test verifying that the underlying expressions of an alert are still relevant. The goal behind such tests is to detect alerts that are out-of-sync with the platform, whether it is because a metric was removed or one of its labels was changed. To do so, we want to make sure that all the metrics and their selectors aren't absent. However, this comes with the trade-off that we won't be able to detect if an error-related alert is still relevant since in most cases, error metrics are absent if no error occurred.

One use case that we recently ran into is when the buckets of a metric changed and an alert using them hasn't been updated. In that case, we want to detect that the alert isn't doing anything anymore in order to fix it.


dgrisonnet:
@openshift/openshift-team-monitoring could you please have a look at this PR?

test/extended/prometheus/prometheus.go
```diff
@@ -278,6 +279,150 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules", func(
...
+				}
+				return nil
+			})
+			q := strings.Join(exprs, " or ")
+
+			resp, err := helper.RunQuery(context.TODO(), promClient, q)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			if len(resp.Data.Result) > 0 {
+				return sets.NewString("one of the metrics used in the query is absent")
```
sthaha, line 415: would we be able to point out the metric? and print the query?

test/extended/prometheus/prometheus.go
```diff
@@ -278,6 +279,150 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules", func(
...
+		promClient := oc.NewPrometheusClient(context.TODO())
+
+		err := helper.ForEachAlertingRule(alertingRules, func(alert promv1.AlertingRule) sets.String {
+			if exceptions.Has(alert.Name) {
+				return nil
+			}
+
+			// Remove backslashs added by the HTTP API since they can't be parsed by
+			// the promQL parser.
+			query := strings.Replace(alert.Query, "\\", "", -1)
```
sthaha, line 395: should this sanitisation be part of the `helper.ForEachAlertingRule` itself?

test/extended/prometheus/prometheus.go
```diff
@@ -278,6 +279,150 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules", func(
...
+			expr, err := parser.ParseExpr(query)
+			if err != nil {
+				return sets.NewString("couldn't parse query")
```
sthaha, line 398: should we also add embed the query that failed to parse?

test/extended/prometheus/prometheus.go
```diff
@@ -278,6 +279,150 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules", func(
...
+					exprs = append(exprs, fmt.Sprintf("absent(%s)", n))
+				}
+				return nil
+			})
+			q := strings.Join(exprs, " or ")
+
+			resp, err := helper.RunQuery(context.TODO(), promClient, q)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			if len(resp.Data.Result) > 0 {
```
sthaha, line 414: could you please help clarify why `len(resp.Data.Result) > 0` results in a metric being absent?

sthaha (commented):
@dgrisonnet very useful patch! I have few questions.

test/extended/prometheus/prometheus.go
dgrisonnet: yes that would make sense

test/extended/prometheus/prometheus.go
dgrisonnet: sure

test/extended/prometheus/prometheus.go
dgrisonnet: the absent function will return a timeserie if a given metric is absent, so by checking that the result isn't empty means that the metrics is absent

test/extended/prometheus/prometheus.go
dgrisonnet: doing that might be a bit too expensive. Currently, in order to avoid sending too many requests to Prometheus, I am transforming the query into:
```
absent(metrics1) or absent(metrics2) or absent(metrics3) or ...
```
IIRC we are currently checking 300+ alerts and if we were to check each metric used by an alert one by one that would amount to a lot of requests

PR #26789, opened 2022-01-26
Flake instead of fail the timely OSUpdateStaged test on metal and ovirt.

dgoodwin:
This new test seems to fail a fair bit on those platforms causing
unecessary failed jobs on platforms that can't yet meet this guideline.
Make sure we only flake on these platforms for now.


deads2k:
it's worth trying to gather timing on this eventually, but this will stem unnecessary failures for now, while latching other platforms

/lgtm

deads2k:
cleaning up CI

PR #26793, opened 2022-01-27
Bug 2047362: Prometheus check targets endpoint

jan--f:
This change is needed for openshift/cluster-monitoring-operator#1532


simonpasquier (commented):
/lgtm

jan--f:
/retitle Bug 2047362: Prometheus check targets endpoint

PR #26795, opened 2022-01-27
Bug 1961233: Add DNS availability upgrade test

suleymanakbas91:
Add a new test to the upgrade test suite to validate DNS availability during and after an upgrade. The new test implements the simple [UpgradeTest interface](https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/test/e2e/upgrades/upgrade.go#L46). 
- `test/e2e/upgrade/dns/dns.go`: 
(`Name()`) and (`DisplayName()`) provides a name for the test to be used by the framework (e.g. for namespace creation, printing in the logs). 
(`Setup()`) performs the intial setup before the upgrade starts. It retrieves DNS Service Cluster IP and creates a DaemonSet (`createDNSTestDaemonSet()`) which performs a simple dig command (`while true; do dig +short @CLUSTER_IP google.com || echo $(date) fail && sleep 1; done`) to check DNS availability using the obtained Cluster IP.
(`Test()`) is called during upgrade and the `done` channel is closed when the upgrade is finished. `validateDNSResults()` is the function called within `Test()` to test the availability during and after upgrade (i.e. after `done` is closed). It retrieves logs from DaemonSet Pods, counts number of sucessful and failed DNS queries, and calculates the success rate. It fails the test if the success rate is less than 99%. 
(`Teardown()`) is an empty function created only to implement the interface. Framework deletes the namespace created for this test after the test suit finishes. That's why there is no any other clean up need for this test.
- `test/e2e/upgrade/upgrade.go`: Add the new DNS upgrade test to the list of `AllTests` so that it gets executed as a part of ClusterUpgrade tests.

### DIFF
--- a/test/e2e/upgrade/dns/dns.go
+++ b/test/e2e/upgrade/dns/dns.go
@@ -1,0 +1,133 @@
+package dns
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/onsi/ginkgo"
+
+	kappsv1 "k8s.io/api/apps/v1"
+	kapiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/upgrades"
+	imageutils "k8s.io/kubernetes/test/utils/image"
+)
+
+type UpgradeTest struct{}
+
+var appName string
+
+func (t *UpgradeTest) Name() string { return "check-for-dns-availability" }
+func (UpgradeTest) DisplayName() string {
+	return "[sig-network-edge] Verify DNS availability during and after upgrade success"
+}
+
+// Setup creates a DaemonSet to verify DNS availability during and after upgrade
+func (t *UpgradeTest) Setup(f *framework.Framework) {
+	ginkgo.By("Setting up upgrade DNS availability test")
+
+	ginkgo.By("Getting DNS Service Cluster IP")
+	dnsServiceIP := t.getServiceIP(f)
+
+	ginkgo.By("Creating a DaemonSet to verify DNS availability")
+	appName = fmt.Sprintf("dns-test-%s", string(uuid.NewUUID()))
+	t.createDNSTestDaemonSet(f, dnsServiceIP)
+}
+
+// Test checks for logs from DNS availability test a minute after upgrade finishes
+// to cover during and after upgrade phase, and verifies the results.
+func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, _ upgrades.UpgradeType) {
+	ginkgo.By("Validating DNS results during upgrade")
+	t.validateDNSResults(f)
+
+	// Block until upgrade is done
+	<-done
+
+	ginkgo.By("Sleeping for a minute to give it time for verifying DNS after upgrade")
+	time.Sleep(1 * time.Minute)
+
+	ginkgo.By("Validating DNS results after upgrade")
+	t.validateDNSResults(f)
+}
+
+// getServiceIP gets Cluster IP from DNS Service
+func (t *UpgradeTest) getServiceIP(f *framework.Framework) string {
+	dnsService, err := f.ClientSet.CoreV1().Services("openshift-dns").Get(context.Background(), "dns-default", metav1.GetOptions{})
+	framework.ExpectNoError(err)
+	return dnsService.Spec.ClusterIP
+}
+
+// createDNSTestDaemonSet creates a DaemonSet to test DNS availability
+func (t *UpgradeTest) createDNSTestDaemonSet(f *framework.Framework, dnsServiceIP string) {
+	cmd := fmt.Sprintf("while true; do dig +short @%s google.com || echo $(date) fail && sleep 1; done", dnsServiceIP)
+	_, err := f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Create(context.Background(), &kappsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{Name: appName},
+		Spec: kappsv1.DaemonSetSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{"app": appName},
+			},
+			Template: kapiv1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"app": appName},
+				},
+				Spec: kapiv1.PodSpec{
+					Containers: []kapiv1.Container{
+						{
+							Name:    "querier",
+							Image:   imageutils.GetE2EImage(imageutils.JessieDnsutils),
+							Command: []string{"sh", "-c", cmd},
+						},
+					},
+				},
+			},
+		},
+	}, metav1.CreateOptions{})
+	framework.ExpectNoError(err)
+}
+
+// validateDNSResults retrieves the Pod logs and validates the results
+func (t *UpgradeTest) validateDNSResults(f *framework.Framework) {
+	ginkgo.By(fmt.Sprintf("Listing Pods with label app=%s", appName))
+	podClient := f.ClientSet.CoreV1().Pods(f.Namespace.Name)
+	selector, _ := labels.Parse(fmt.Sprintf("app=%s", appName))
+	pods, err := podClient.List(context.Background(), metav1.ListOptions{LabelSelector: selector.String()})
+	framework.ExpectNoError(err)
+
+	ginkgo.By("Retrieving logs from all the Pods belonging to the DaemonSet and asserting no failure")
+	for _, pod := range pods.Items {
+		r, err := podClient.GetLogs(pod.Name, &kapiv1.PodLogOptions{Container: "querier"}).Stream(context.Background())
+		framework.ExpectNoError(err)
+
+		failureCount := 0.0
+		successCount := 0.0
+		scan := bufio.NewScanner(r)
+		for scan.Scan() {
+			line := scan.Text()
+			if strings.Contains(line, "fail") {
+				failureCount++
+			} else if ip := net.ParseIP(line); ip != nil {
+				successCount++
+			}
+		}
+
+		if successRate := (successCount / (successCount + failureCount)) * 100; successRate < 99 {
+			err = fmt.Errorf("success rate is less than 99%% on the node %s: [%0.2f]", pod.Spec.NodeName, successRate)
+		} else {
+			err = nil
+		}
+		framework.ExpectNoError(err)
+	}
+}
+
+// Teardown cleans up any objects that are created that
+// aren't already cleaned up by the framework.
+func (t *UpgradeTest) Teardown(_ *framework.Framework) {
+	// rely on the namespace deletion to clean up everything
+}
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -9,8 +9,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/openshift/origin/test/e2e/upgrade/service"
-
 	v1 "k8s.io/api/core/v1"
 	eventsv1 "k8s.io/api/events/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,146 @@
...
+func (t *UpgradeTest) getServiceIP(f *framework.Framework) string {
+	dnsService, err := f.ClientSet.CoreV1().Services("openshift-dns").Get(context.Background(), "dns-default", metav1.GetOptions{})
+	framework.ExpectNoError(err)
+	return dnsService.Spec.ClusterIP
+}
+
+// createDNSTestDaemonSet creates a DaemonSet to test DNS availability
+func (t *UpgradeTest) createDNSTestDaemonSet(f *framework.Framework, dnsServiceIP string) {
+	// daemon sets should auto replace local references
+	cmd := fmt.Sprintf("while true; do dig +noall @%s google.com && echo success || echo fail; done", dnsServiceIP)
```
brandisher, line 71: It might be a good idea to add a `sleep` in here. Maybe `sleep 1` just to space out the lookups a bit. We should also consider whether caching is an important aspect here...for example, if the default caching is 3 seconds we could get a cached record for _n_ lookups before receiving a failure.

I also noticed that `dig` doesn't differentiate between a successful lookup and an `NXDOMAIN`.
```
$ dig +noall asdlkfhsdlfhsdaklhfklsjflklskhflsdflkadflhsdafkljhsdakjhfsdf.com
$ echo $?
0
$ dig +noall google.com
$ echo $?
0
```
The only problem with that is we have no way of programmitcally differentiating between a failed lookup for a domain that we know should be successful (e.g. google.com) and a domain that we know should not be successful. To solve this, it might make sense to just do:
```
while true; do dig +short @%s google.com || echo fail; done
```
Then when looping through the pod logs you can validate that a line is an IP address. If a line is "fail" or not an IP, you know something else went wrong and can fail accordingly. Hopefully that all makes sense!

brandisher (changes requested):

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+	}, metav1.CreateOptions{})
+	framework.ExpectNoError(err)
+}
+
+// validateDNSResults retrieves the Pod logs and validates the results
+func (t *UpgradeTest) validateDNSResults(f *framework.Framework) {
+	ginkgo.By("Listing Pods with label app=dns-test")
+	podClient := f.ClientSet.CoreV1().Pods(f.Namespace.Name)
+	selector, _ := labels.Parse("app=dns-test")
```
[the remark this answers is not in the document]
suleymanakbas91: I appended a random UUID to the end so that it will be unique.

test/e2e/upgrade/upgrade.go
```diff
@@ -3,6 +3,7 @@ package upgrade
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/origin/test/e2e/upgrade/dns"
```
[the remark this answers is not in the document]
suleymanakbas91: Done.

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,146 @@
...
+	pods, err := podClient.List(context.Background(), metav1.ListOptions{LabelSelector: selector.String()})
+	framework.ExpectNoError(err)
+
+	ginkgo.By("Retrieving logs from all the Pods belonging to the DaemonSet and asserting no failure")
+	for _, pod := range pods.Items {
```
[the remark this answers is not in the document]
suleymanakbas91: Nice tooling, but I'm not sure if we can use it here. We might use it to look for the `fail` keyword, but in our case, we should also check if there is anything else in the logs rather than IP addresses or `fail` keyword. Not sure how to achieve this with this tooling.

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,146 @@
...
+
+		ginkgo.By("Retrieving the Pod logs")
+		r, err := podClient.GetLogs(pod.Name, &kapiv1.PodLogOptions{Container: "querier"}).Stream(context.Background())
+		framework.ExpectNoError(err)
+		out, err := ioutil.ReadAll(r)
+		framework.ExpectNoError(err)
+
+		ginkgo.By("Looking for the results for any failures")
+		err = assertNoFailure(bytes.NewBuffer(out))
+		framework.ExpectNoError(err)
```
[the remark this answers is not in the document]
suleymanakbas91: I've applied the suggestion. It's much simpler and more efficient. Thanks!

test/e2e/upgrade/dns/dns.go
suleymanakbas91: I've applied the suggestion. Thanks!

brandisher (approved):
This looks good to me! Ready for review by the broader team.

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,129 @@
+package dns
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"github.com/onsi/ginkgo"
```
candita, line 7: We follow a conventional pkg ordering pattern of:
* single-name pkg alphabetically ordered
* blank line
* github pkgs alphabetically ordered
* blank line
* k8s pkgs alphabetically ordered
* blank line
* other pkgs in alphabetical order

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,129 @@
...
+
+	ginkgo.By("Creating a DaemonSet to verify DNS availability")
+	appName = fmt.Sprintf("dns-test-%s", string(uuid.NewUUID()))
+	t.createDNSTestDaemonSet(f, dnsServiceIP)
+}
+
+// Test checks for logs from DNS availability test a minute after upgrade finishes
+// to cover during and after upgrade phase, and verifies the results.
+func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, _ upgrades.UpgradeType) {
+	// Block until upgrade is done
```
candita, line 45: We want to test during an upgrade as well, you need to add a test for during the upgrade.

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,129 @@
...
+
+// getServiceIP gets Cluster IP from DNS Service
+func (t *UpgradeTest) getServiceIP(f *framework.Framework) string {
+	dnsService, err := f.ClientSet.CoreV1().Services("openshift-dns").Get(context.Background(), "dns-default", metav1.GetOptions{})
+	framework.ExpectNoError(err)
+	return dnsService.Spec.ClusterIP
+}
+
+// createDNSTestDaemonSet creates a DaemonSet to test DNS availability
+func (t *UpgradeTest) createDNSTestDaemonSet(f *framework.Framework, dnsServiceIP string) {
```
candita, line 64: Any reason you can't do this with just a pod?  https://github.com/openshift/origin/blob/master/test/extended/dns/dns.go#L30

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,129 @@
...
+	cmd := fmt.Sprintf("while true; do dig +short @%s google.com || echo fail && sleep 1; done", dnsServiceIP)
+	_, err := f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Create(context.Background(), &kappsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{Name: appName},
```
candita, line 67: Anything you create using f.ClientSet, you will need to clean up later by deleting it, usually done in a `defer` block.

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,129 @@
...
+			err = nil
+		}
+		framework.ExpectNoError(err)
+	}
+}
+
+// Teardown cleans up any objects that are created that
+// aren't already cleaned up by the framework.
+func (t *UpgradeTest) Teardown(_ *framework.Framework) {
+	// rely on the namespace deletion to clean up everything
```
candita, line 128: I don't see where you created or deleted a namespace for the new objects?

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,129 @@
...
+					},
+				},
+			},
+		},
+	}, metav1.CreateOptions{})
+	framework.ExpectNoError(err)
+}
+
+// validateDNSResults retrieves the Pod logs and validates the results
+func (t *UpgradeTest) validateDNSResults(f *framework.Framework) {
```
candita, line 92: Overall, I think you should model this test re-using more from https://github.com/openshift/origin/blob/master/test/extended/dns/dns.go.

test/e2e/upgrade/dns/dns.go
suleymanakbas91: Thanks! I re-ordered them based on this. Is there a place where I can see all the team conventions?

test/e2e/upgrade/dns/dns.go
suleymanakbas91: Fixed. Now I run the test during and after upgrade.

test/e2e/upgrade/dns/dns.go
suleymanakbas91: If the `SkipNamespaceCreation` is not set explicitly, a namespace is being created and deleted by the framework: https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/test/e2e/framework/framework.go#L238

test/e2e/upgrade/dns/dns.go
suleymanakbas91: The namespace being used here will already get deleted by the framework at the end of the test. That's why, AFAIU, deleting these test resources explicitly should not be required.

test/e2e/upgrade/dns/dns.go
suleymanakbas91: No, but this was Miciah's idea to use a DaemonSet to test the DNS availability from all the Pods.

test/e2e/upgrade/dns/dns.go
candita: Mostly just what `go fmt` and `goimports` do.  There are also https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#api and https://go.dev/doc/effective_go#formatting

test/e2e/upgrade/dns/dns.go
candita: After discussion, I will leave this up to the approver/s.

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,129 @@
...
+// getServiceIP gets Cluster IP from DNS Service
+func (t *UpgradeTest) getServiceIP(f *framework.Framework) string {
+	dnsService, err := f.ClientSet.CoreV1().Services("openshift-dns").Get(context.Background(), "dns-default", metav1.GetOptions{})
+	framework.ExpectNoError(err)
+	return dnsService.Spec.ClusterIP
+}
+
+// createDNSTestDaemonSet creates a DaemonSet to test DNS availability
+func (t *UpgradeTest) createDNSTestDaemonSet(f *framework.Framework, dnsServiceIP string) {
+	cmd := fmt.Sprintf("while true; do dig +short @%s google.com || echo fail && sleep 1; done", dnsServiceIP)
```
candita, line 69: After discussion, it would be better if you could also echo a timestamp here when it fails.

candita:
After discussion, @suleymanakbas91  if you could please check: https://github.com/openshift/origin/pull/26795/commits/c10e64e1f5f591c45ad15993dca4147b20b21a7d#r817073378

Otherwise looks good to me and you should contact someone from the OWNERS file in the #forum-api-review slack channel for approval.  This will be invaluable in our CI and upgrade improvement strategy.

/lgtm

test/e2e/upgrade/dns/dns.go
suleymanakbas91: Now, I'm also printing a timestamp when it fails.

test/e2e/upgrade/dns/dns.go
```diff
@@ -0,0 +1,133 @@
...
+			line := scan.Text()
+			if strings.Contains(line, "fail") {
+				failureCount++
+			} else if ip := net.ParseIP(line); ip != nil {
+				successCount++
+			}
+		}
+
+		if successRate := (successCount / (successCount + failureCount)) * 100; successRate < 99 {
+			err = fmt.Errorf("success rate is less than 99%%: [%0.2f]", successRate)
```
deads2k, line 121: how will you know node had the problem?

deads2k:
/approve

this is a great test to add.  I've got a comment about improving the error message

/hold in case you want to address the comment.


I also suggest adding this as an invariant test for *all* openshift-tests since DNS should work during all the tests we run.

test/e2e/upgrade/dns/dns.go
suleymanakbas91: Now I'm printing the node name as well.

brandisher:
Retesting as the failures don't appear to be related to this change.

/retest

PR #26796, opened 2022-01-28
don't double report early alert failures

deads2k:
signal cleanup for a duplicated test failure.


deads2k:
no conflict, merging so I can pick.

/cherrypick 4.10

PR #26798, opened 2022-01-28
Debug missing OSUpdateStarted events

dgoodwin:
- Improve edge cases error message for OSUpdateStaged test.
- Debugging hack for missing event intervals.

An edge case we didn't really think was possible is occurring in about 0.07% of jobs. 

When this happens it appears we record an OSUpdateStaged event, but are missing an OSUpdateStarted, which logically should be there. Gather-extra sees it after the job when we oc get events, but for some reason the monitor is not recording it. Affects just one host in the job run, the others are fine.

Attempt to debug by ruling out whether or not we're getting the event at all in our watch.

### DIFF
--- a/pkg/monitor/event.go
+++ b/pkg/monitor/event.go
@@ -66,6 +66,16 @@ func startEventMonitoring(ctx context.Context, m Recorder, client kubernetes.Int
 							}
 							m.RecordResource("events", obj)
 
+							// Temporary hack by dgoodwin, we're missing events here that show up later in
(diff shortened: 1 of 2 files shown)


dgoodwin:
/hold

Waiting to see the output in prow.

deads2k:
Please be sure that you can locate instances of the problem you're chasing after this merges.

/approve

/assign @DennisPeriquet 

@DennisPeriquet owns lgtm

dgoodwin:
Depends on https://github.com/openshift/origin/pull/26808

PR #26800, opened 2022-01-28
Bug 2047790: Skip some HAProxy tests on FIPS

Miciah:
Skip HAProxy tests that use the router image's built-in default certificate, which is incompatible with FIPS.  This is a temporary measure to prevent CI from permafailing on FIPS over the weekend.

* `test/extended/router/scoped.go`:
* `test/extended/router/unprivileged.go`:
* `test/extended/router/weighted.go`: Skip on FIPS.


deads2k:
I plan to merge this on a green e2e parallel run and proceed with a pick.

deads2k:
e2e gcp mostly worked, e2e-serial failed install, but this doesn't hit that.

PR #26801, opened 2022-01-28
Bug 2047911: IBMCloud: Stop creating CRB for IBM Cloud clusters

cjschaef:
Update the IBM Cloud provider framework to no longer create a
ClusterRoleBinding for IBM Cloud clusters, as unmanaged clusters
should not need this CRB and managed clusters will need to account
for this as needed.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1825034

### DIFF
--- a/test/extended/util/ibmcloud/provider.go
+++ b/test/extended/util/ibmcloud/provider.go
@@ -1,13 +1,6 @@
 package ibmcloud
 
 import (
-	"context"
-
-	o "github.com/onsi/gomega"
-
-	rbacv1 "k8s.io/api/rbac/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
@@ -24,32 +17,4 @@ func newProvider() (framework.ProviderInterface, error) {
 // Provider is a structure to handle IBMCloud for e2e testing
 type Provider struct {
 	framework.NullProvider
-}
-
-// TODO: Determine if this can be avoided by running worker kubelets without
-// --enable-controller-attach-detach=false
-// e2e: https://bugzilla.redhat.com/show_bug.cgi?id=1825034 - Mock CSI tests fail on IBM ROKS clusters
-func (p *Provider) FrameworkBeforeEach(f *framework.Framework) {
-	_, err := f.ClientSet.RbacV1().ClusterRoleBindings().Get(context.Background(), "e2e-node-attacher", metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		rb := &rbacv1.ClusterRoleBinding{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "e2e-node-attacher",
-			},
-			Subjects: []rbacv1.Subject{
-				{
-					APIGroup: rbacv1.SchemeGroupVersion.Group,
-					Kind:     "Group",
-					Name:     "system:nodes",
-				},
-			},
-			RoleRef: rbacv1.RoleRef{
-				APIGroup: rbacv1.SchemeGroupVersion.Group,
-				Kind:     "ClusterRole",
-				Name:     "system:controller:attachdetach-controller",
-			},
-		}
-		_, err = f.ClientSet.RbacV1().ClusterRoleBindings().Create(context.Background(), rb, metav1.CreateOptions{})
-	}
-	o.Expect(err).NotTo(o.HaveOccurred())
 }


test/extended/util/ibmcloud/provider.go
```diff
@@ -25,31 +18,3 @@ func newProvider() (framework.ProviderInterface, error) {
 type Provider struct {
 	framework.NullProvider
 }
-
-// TODO: Determine if this can be avoided by running worker kubelets without
-// --enable-controller-attach-detach=false
-// e2e: https://bugzilla.redhat.com/show_bug.cgi?id=1825034 - Mock CSI tests fail on IBM ROKS clusters
-func (p *Provider) FrameworkBeforeEach(f *framework.Framework) {
-	_, err := f.ClientSet.RbacV1().ClusterRoleBindings().Get(context.Background(), "e2e-node-attacher", metav1.GetOptions{})
-	if errors.IsNotFound(err) {
```
bparees, line 34: the error reported in the BZ was an "already exists" error.  If this "IsNotFound" check is working properly, the code should not attempt to create the CRB if it already exists, so do we know why this check wasn't sufficient?

bparees:
/hold

interested in my answer to the question before this merges

cjschaef:
@bparees I wonder if this is due to tests running in parallel.
I'm not sure how the `openshift-tests` handle this, but perhaps the issue is multiple tests trying to create the CRB at nearly the same time? Although, that's just a guess at the moment, and I'm not sure how to prove that just yet.

bparees:
> @bparees I wonder if this is due to tests running in parallel.
I'm not sure how the openshift-tests handle this, but perhaps the issue is multiple tests trying to create the CRB at nearly the same time? Although, that's just a guess at the moment, and I'm not sure how to prove that just yet.

it could be, if it doesn't fail consistently.  If it fails consistently i'd be skeptical that we're getting that unlucky.

cjschaef:
>it could be, if it doesn't fail consistently. If it fails consistently i'd be skeptical that we're getting that unlucky.

I'd agree with that. And at least from the results I have seen, it did seem consistent, although I'm not 100% confident on that now looking back. Perhaps I can get some more verbosity in the test logs, or inject something more for confirmation locally.

bparees:
if this CI failure is blocking something/someone, by all means feel free to drop the `hold`, but otherwise it seems like maybe a little more investigation is warranted, maybe we can come up with a solution that still works for both platforms.

jeffnowicki:
@bparees the change won't break IBM ROKS until they move to 4.10. We've notified key devr's at IBM (they're members of our squad). We were more concerned with getting our IPI implementation to pass  for 4.10 (as it's a RH requirement) vs IBM ROKS (where it isn't a RH requirement). Ideally, the decision logic should 'assume' IPI and if ROKS is detected, perform specific setup.

jeffnowicki:
@bparees And thank you for review/feedback/assistance.

cjschaef:
Looking back through the series of collected results, I see right away this failure occur
[ocp-conformance-26.log.short.txt](https://github.com/openshift/origin/files/7981203/ocp-conformance-26.log.short.txt)


Furthermore, I see an instance where it tries to create the CRB three times and fail since it already exists.
[ocp-conformance-25.log.short.txt](https://github.com/openshift/origin/files/7981226/ocp-conformance-25.log.short.txt)

After a simple change to add a mutex, it appears the failure no longer occurs.
```
# git diff
diff --git a/test/extended/util/ibmcloud/provider.go b/test/extended/util/ibmcloud/provider.go
index 85353239d6..a3809e2178 100644
--- a/test/extended/util/ibmcloud/provider.go
+++ b/test/extended/util/ibmcloud/provider.go
@@ -2,6 +2,7 @@ package ibmcloud
 
 import (
 	"context"
+	"sync"
 
 	o "github.com/onsi/gomega"
 
@@ -26,10 +27,15 @@ type Provider struct {
 	framework.NullProvider
 }
 
+var clusterRoleBindingMutex sync.Mutex
+
 // TODO: Determine if this can be avoided by running worker kubelets without
 // --enable-controller-attach-detach=false
 // e2e: https://bugzilla.redhat.com/show_bug.cgi?id=1825034 - Mock CSI tests fail on IBM ROKS clusters
 func (p *Provider) FrameworkBeforeEach(f *framework.Framework) {
+	clusterRoleBindingMutex.Lock()
+	defer clusterRoleBindingMutex.Unlock()
+
 	_, err := f.ClientSet.RbacV1().ClusterRoleBindings().Get(context.Background(), "e2e-node-attacher", metav1.GetOptions{})
 	if errors.IsNotFound(err) {
 		rb := &rbacv1.ClusterRoleBinding{
```

As such, I will create a new PR to fix the concurrency issue that is at the root of this problem instead and then close this PR.

cjschaef:
I opened a PR to attempt to address the concurrency issue.
https://github.com/openshift/origin/pull/26806

Will close this PR in favor of that.

PR #26803, opened 2022-01-29
Bug 2047913: Fix HAProxy tests on FIPS properly

Miciah:
#### Revert "Skip some HAProxy tests on FIPS"

This reverts #26800.

#### `test/extended/util/url`: Use `curl --connect-to`

* `test/extended/util/url/url.go` (`CURL):` Add a new `ProxyHost` field.
(`Through`): Assign the given address to `ProxyHost` instead of changing the request host or headers.
(`ToShell`): Use Curl's `--connect-to` flag to specify the proxy host.

#### Fix HAProxy tests on FIPS properly

Change existing HAProxy tests to use a FIPS-compatible, 2048-bit RSA key. Add a new test to verify that using a FIPS-incompatible, 1024-bit RSA key for the router's default certificate fails on FIPS clusters and succeeds on non-FIPS clusters.

The certificate used in this new test comes from the default certificate that is currently built into the router image (which is why the tests were failing on FIPS).  The certificate used for the already existing tests was generated using the following Shell commands on a RHEL 8.4 system with OpenSSL 1.1.1g:

    openssl req -x509 -newkey rsa:2048 -days 3650 -keyout ca.key -out ca.crt -nodes -subj '/C=US/ST=SC/L=Default City/O=Default Company Ltd/OU=Test CA/CN=www.exampleca.com/emailAddress=example@example.com'
    openssl req -newkey rsa:2048 -nodes -keyout tls.key -out router.csr -subj '/CN=www.example.com/ST=SC/C=US/emailAddress=example@example.com/O=Example/OU=Example'
    printf 'basicConstraints=CA:FALSE\n' > router.cnf
    openssl x509 -req -days 3650 -in router.csr -signkey tls.key -CA ca.crt -CAcreateserial -CAkey ca.key -out tls.crt -clrext -extfile router.cnf
    cat tls.{crt,key} > default_pub_keys.pem

These commands generate a certificate that has the same parameters as the old one except for the serial number, signature algorithm, validity period, and key size.

* `test/extended/router/certs.go`: New file with the new tests.
* `test/extended/testdata/router/router-common.yaml`: Change route-1 and route-2 from non-TLS routes to edge-terminated TLS routes with `insecureEdgeTerminationPolicy: Allow`.
* `test/extended/testdata/router/router-override-domains.yaml`:
* `test/extended/testdata/router/router-override.yaml`:
* `test/extended/testdata/router/router-scoped.yaml`:
* `test/extended/testdata/router/weighted-router.yaml`: Add a parameter for the default certificate with the default value being a certificate with a FIPS-compatible, 2048-bit RSA key.
* `test/extended/testdata/bindata.go`:
* `test/extended/util/annotate/generated/zz_generated.annotations.go`: Regenerate.

#### `test/extended/router`: Dump crypto-policy for FIPS

Update the new tests that use a FIPS-incompatible key to gather and print out the content of `/etc/crypto-policies/back-ends/opensslcnf.config` so that the configured crypto-policy can be examined in case of test failures.

* `test/extended/router/certs.go`: cat `/etc/crypto-policies/back-ends/opensslcnf.config` after each test.


test/extended/util/url/url.go
```diff
@@ -362,7 +363,18 @@ func (ut *Test) ToShell(i int) string {
...
+	resolve := ""
+	if len(ut.ProxyHost) != 0 {
+		host, port, err := net.SplitHostPort(ut.Req.URL.Host)
+		if err != nil {
+			host, port = ut.Req.URL.Host, "80"
+			if ut.Req.URL.Scheme == "https" {
+				port = "443"
+			}
+		}
+		resolve = fmt.Sprintf("--resolve %s:%s:%s", host, port, ut.ProxyHost)
```
deads2k, line 375: will this function on ipv6?

deads2k:
/approve

The overall approach looks reasonable.  I'll leave lgtm with someone on network-edge.  @frobware perhaps?

test/extended/util/url/url.go
Miciah: It should, if the address is bracketed.  I've added a comment to `Through` to mention that IPv6 addresses should be bracketed.  I've also replaced `--resolve` with `--connect-to`, which handles IPv4 addresses, bracketed IPv6 addresses, and host names (the last would have failed with `--resolve`, which would cause problems on platforms such as AWS where the LB status has a host name rather than an IP address).

Miciah:
Latest push updates some additional manifests with the FIPS-compatible certificate and replaces `--resolve` with `--connect-to`.

Miciah:
The new test failed because the router started even with a FIPS-incompatible, 1024-bit key on a FIPS-enabled cluster.  I'm not sure how that could happen.  The test passed for me when I tested with cluster-bot; in manual testing, the router did _not_ start with a FIPS-incompatible key on a FIPS-enabled GCP cluster.
/test e2e-aws-fips

frobware:
Many failures.

/retest

Miciah:
The tests work on cluster-bot fips clusters.  In addition to my earlier manual testing on a FIPS-enabled GCP cluster, I used `/msg @cluster-bot test e2e openshift/origin#26803 aws,fips` and `/msg @cluster-bot test e2e openshift/origin#26803 gcp,fips` to verify that the tests work properly on FIPS clusters.  Linked below are the passing cluster-bot jobs, which show that `[sig-network][Feature:Router] when FIPS is enabled the HAProxy router should not work when configured with a 1024-bit RSA key` passes and `[sig-network][Feature:Router] when FIPS is disabled the HAProxy router should serve routes when configured with a 1024-bit RSA key` is skipped on FIPS clusters:

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws-modern/1488206794919514112

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp-modern/1488178502883610624

I'm increasingly thinking that our e2e-aws-fips and e2e-gcp-fips-serial tests are broken.

/test e2e-aws-fips
/test e2e-gcp-fips-serial

Miciah:
The e2e-gcp-fips-serial job has gone missing...
/refresh

Miciah:
```
   * could not run steps: step src failed: could not get build src: builds.build.openshift.io "src" not found 
```
/test e2e-aws-fips

PR #26809, opened 2022-02-02
Add pod securityContext sysctl whitelist tests

mmirecki:
Add tests checking that pods with whitelisted sysctl's can be created.


deads2k:
the test itself seems ok, but please get @mrunalp to check the list

/approve

PR #26774, opened 2022-01-21
Bug 2044210: [baremetal] remove wait condition when deleting extraworkers

andfasano:
Do not wait for workers to be cleaned up on teardown. Setup in any case will take care of waiting previously deleted ones (impacts only serial jobs)

PR #26810, opened 2022-02-02
Bug 2049117: Reenable wait on worker deletion and increase serial test timeout

bfournie:
With the fix to remove the extra worker deletion (https://github.com/openshift/origin/pull/26774) there have been occasional test failures when the time to delete the extra worker has exceeded the 15 minute test time. This adds back in the wait after the extra worker is deleted and increases the test time to allow the deletion to complete.

### DIFF
--- a/test/extended/baremetal/helper.go
+++ b/test/extended/baremetal/helper.go
@@ -138,6 +138,8 @@ func (b *BaremetalTestHelper) DeleteAllExtraWorkers() {
 	for _, worker := range b.extraWorkers {
 		err := b.bmcClient.Delete(context.Background(), worker.GetName(), metav1.DeleteOptions{})
 		o.Expect(err).ToNot(o.HaveOccurred())
+
+		b.waitForDeletion(worker)
 	}
 }
 
--- a/test/extended/machines/scale.go
+++ b/test/extended/machines/scale.go
@@ -216,7 +216,7 @@ var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Serial] Managed cl
 		helper.DeleteAllExtraWorkers()
 	})
 
-	g.It("grow and decrease when scaling different machineSets simultaneously", func() {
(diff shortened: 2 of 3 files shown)


andfasano:
The patch looks good to me, I'd just wait for https://github.com/openshift/origin/pull/26813 to land and have another run of `e2e-metal-ipi-serial` job

cmd/openshift-tests/e2e.go
```diff
@@ -85,6 +85,7 @@ var staticSuites = testSuites{
 				}
 				return strings.Contains(name, "[Suite:openshift/conformance/serial") || isStandardEarlyOrLateTest(name)
 			},
+			TestTimeout:         20 * time.Minute,
```
soltysh, line 88: I'm fine bringing back the extra wait, but I don't want to increase the overall test timeout for everyone. Spoke with @bfournie he'll try to make it more targetted at bare metal.

bfournie:
Latest update depends on https://github.com/openshift/origin/pull/26833 as it uses the `[Timeout:<duration>]` notation to increase the test timeout.

soltysh (approved):
/lgtm
/approve
/hold for https://github.com/openshift/origin/pull/26833

test/extended/baremetal/hosts.go
```diff
@@ -147,7 +147,7 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
 		helper.DeleteAllExtraWorkers()
 	})
 
-	g.It("skip inspection when disabled by annotation", func() {
+	g.It("skip inspection when disabled by annotation [Timeout:20m]", func() {
```
andfasano, line 150: @bfournie I think this test it's ok, it's the scale&grow that needs to be increased

bfournie: > @bfournie looks like a typo, it's the scale&grow that needs to be increased

Yeah that's what the test results look like. Updated timeout in scale and grow.

PR #26813, opened 2022-02-03
Delete extra workers before running metal high availability test

ardaguclu:
This PR adds before step to run metal high availability test.
Thanks to that, test will ensure there is no other extra workers which
statuses can be set variably.


andfasano:
In the latest serial job the first test was `: [sig-installer][Feature:baremetal][Serial] Baremetal platform should ensure cluster baremetal operator and metal3 deployment return back healthy after they are deleted [Suite:openshift/conformance/serial]` and it passed

/lgtm

PR #26817, opened 2022-02-03
Revert "Flake failed sandboxes from bug in new guard pods"

jluhrsen:
This reverts commit 333d91371a1835f499073208dbb712467921aea5.

https://bugzilla.redhat.com/show_bug.cgi?id=2038481 is resolved now and
this test workaround should no longer be needed.


deads2k:
great!

/lgtm

PR #26824, opened 2022-02-07
Break out test for OSUpdateStaged event with no OSUpdateStarted.

dgoodwin:
This is confirmed to be happening and we can see that we're missing
events in the disruption monitor watch when apiservers are upgrading.

Move this scenario out to it's own test, which will only ever flake,
while we work on a fix.

Original test will now ignore this scenario.


deads2k:
/lgtm
/hold
/approve

/cherrypick release-4.10

Please confirm you see your tests before releasing the hold.  Please get bugs assigned to allow cherrypicking to 4.10.

dgoodwin:
We very likely cannot reproduce the problem at pr testing scale, but I can confirm the test is present and showing successful in artifacts for gcp-upgrade:

```
<testcase name="[bz-monitoring][invariant] alert/PrometheusOperatorWatchErrors should not be at or above info" time="0"/>
<testcase name="[bz-Machine Config Operator] Nodes should reach OSUpdateStaged in a timely fashion" time="0"/>
```

/hold cancel

PR #26826, opened 2022-02-08
WIP: operators: add test that verify cluster operators reconciling conditions

mfojtik:
This test checks if all cluster operators properly reconcile their messages and update lastTransitionTime when the message is changed from external (the test does it here).
This test fail if one or more operators fail to change the message for condition back (or to something else) in 2 minutes.

xref: https://github.com/openshift/enhancements/pull/1025

### DIFF
--- a/test/extended/operators/clusteroperators.go
+++ b/test/extended/operators/clusteroperators.go
@@ -2,6 +2,13 @@ package operators
 
 import (
 	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/openshift/origin/pkg/synthetictests"
+
+	"k8s.io/client-go/util/retry"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


test/extended/operators/clusteroperators.go
```diff
@@ -45,13 +54,91 @@ var _ = g.Describe("[sig-arch] ClusterOperators", func() {
...
+		client, err := configclient.NewForConfig(kubeConfig)
 		o.Expect(err).ToNot(o.HaveOccurred())
-		clusterOperatorsList, err := configClient.ClusterOperators().List(context.Background(), metav1.ListOptions{})
+		clusterOperatorsClient = client.ClusterOperators()
+		clusterOperatorsList, err := clusterOperatorsClient.List(context.Background(), metav1.ListOptions{})
 		o.Expect(err).ToNot(o.HaveOccurred())
 		clusterOperators = clusterOperatorsList.Items
 	})
 
+	g.It("should reconcile the messages and update last transition time", func() {
```
wking, line 65: We're not actually expecting `lastTransitionTime` changes, as discussed [here][1].

[1]: https://github.com/openshift/enhancements/pull/1025#discussion_r799842423

PR #26831, opened 2022-02-09
bug 2053143:allow inexact matches for disruption data

deads2k:
This will help alleviate problems like https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1302/pull-ci-openshift-cluster-network-operator-master-e2e-agnostic-upgrade/1491351105987153920  where we don't have exact data matches.

### DIFF
--- a/pkg/synthetictests/allowedalerts/matches.go
+++ b/pkg/synthetictests/allowedalerts/matches.go
@@ -1,8 +1,9 @@
 package allowedalerts
 
 import (
-	"fmt"
 	"time"
+
+	"github.com/openshift/origin/pkg/synthetictests/historicaldata"
 
 	"github.com/openshift/origin/pkg/synthetictests/platformidentification"
 )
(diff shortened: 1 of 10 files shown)


pkg/synthetictests/historicaldata/next_best_guess.go
```diff
@@ -0,0 +1,163 @@
...
+		return platformidentification.JobType{}, false
+	}
+
+	ret := platformidentification.CloneJobType(in)
+	ret.FromRelease = fmt.Sprintf("%d.%d", toReleaseMajor, toReleaseMinor)
+	return ret, true
+}
+
+// PreviousReleaseUpgrade if we don't have data for the current toRelease, perhaps we have data for the congruent test
+// on the prior release.
```
dgoodwin, line 97: Example would be nice here in the godoc for each of these, i.e. x will fall back to y.

pkg/synthetictests/historicaldata/next_best_guess.go
```diff
@@ -0,0 +1,163 @@
...
+
+func getMinor(in string) int {
+	minor, err := strconv.ParseInt(strings.Split(in, ".")[1], 10, 32)
+	if err != nil {
+		panic(err)
+	}
+	return int(minor)
+}
+
+// OnOVN maybe we have data on OVN
```
dgoodwin, line 129: OnOVN and OnSDN are not super clear from the godoc, examples again would help here.

That said, jumping from SDN to equivalent job on OVN, in the context of disruption, feels like too big of a leap.

pkg/synthetictests/historicaldata/next_best_guess.go
```diff
@@ -0,0 +1,163 @@
...
+    "Platform": "azure",
+    "Network": "sdn",
+    "Topology": "ha",
+    "P95": "0.0",
+    "P99": "12.869999999999884"
+  },
+
+*/
+
+// NextBestKey returns the next best key and a bool indicating whether this guesser has an opinion.
```
dgoodwin, line 52: "next best key in the query_results.json generated from BigQuery, and a bool..."

pkg/synthetictests/allowedalerts/matches.go
```diff
@@ -41,53 +42,17 @@ type percentileAllowances struct {
...
 
 func (d *percentileAllowances) FailAfter(alertName string, jobType platformidentification.JobType) time.Duration {
-	allowed := getClosestPercentilesValues(alertName, jobType)
+	allowed, _, _ := getClosestPercentilesValues(alertName, jobType)
 	return allowed.P99
 }
 
 func (d *percentileAllowances) FlakeAfter(alertName string, jobType platformidentification.JobType) time.Duration {
-	allowed := getClosestPercentilesValues(alertName, jobType)
+	allowed, _, _ := getClosestPercentilesValues(alertName, jobType)
```
dgoodwin, line 50: Should we break out separate tests for these when they have to fallback to a fuzzy match, so we can get a clearer signal on when we're failing with exact matches vs when we're failing on fuzzy? It's getting verbose I know, but it seems worth asking.

pkg/synthetictests/historicaldata/next_best_guess.go
```diff
@@ -0,0 +1,163 @@
...
+	if in.Network == "sdn" {
+		return platformidentification.JobType{}, false
+	}
+
+	ret := platformidentification.CloneJobType(in)
+	ret.Network = "sdn"
+	return ret, true
+}
+
+func combine(nextBestKeys ...NextBestKey) NextBestKey {
```
dgoodwin, line 151: I can't follow what you're trying to do with combine, it's used above with combos of matchers we already processed, which if they returned true, would have been used. Then you "combine" them, which looks to be logic for "if any of the combined guessers returns false, return false, else return the last one that was true", but this would never run because the guessers were already processed individually and returned true?

I'm likely missing something but I've stared at it long enough to know it really needs some godoc, here on this func and probably up above in the list as well.

pkg/synthetictests/historicaldata/next_best_guess.go
```diff
@@ -0,0 +1,163 @@
...
+    "Network": "sdn",
+    "Topology": "ha",
+    "P95": "0.0",
+    "P99": "12.869999999999884"
+  },
+
+*/
+
+// NextBestKey returns the next best key and a bool indicating whether this guesser has an opinion.
+// If the bool is false, the key should not be used.
```
dgoodwin, line 53: It may help someone to clarify here that a guesser returning true does not imply we actually have said key, it's just another key to check for.

pkg/synthetictests/historicaldata/next_best_guess.go
deads2k: > That said, jumping from SDN to equivalent job on OVN, in the context of disruption, feels like too big of a leap.

you'd rather go to the prior version first or you'd rather fail?

pkg/synthetictests/allowedalerts/matches.go
deads2k: > Should we break out separate tests for these when they have to fallback to a fuzzy match,

I suggest not doing that because our periodics won't fail like this and the per PR we don't really care about beyond extreme breakage.

pkg/synthetictests/historicaldata/next_best_guess.go
dgoodwin: Fail is what I'd think.

pkg/synthetictests/historicaldata/next_best_guess.go
deads2k: combine runs the first and uses the output for the next.  I'll doc it.  That allows you to have methods for same update last release and for minor instead of micro, and combined them to have do you have a minor upgrade from last release.

PR #26832, opened 2022-02-10
WIP: Add provisioning host when virtualMediaViaExternalNetwork is true test

ardaguclu:
This PR adds new metal test namely
`Baremetal platform should successfully provision resources
when virtualMediaViaExternalNetwork is used`.

This test will set provisioning configs `virtualMediaViaExternalNetwork`
field to true and provision extra worker to assure that
it is provisioned correctly.

### DIFF
--- a/test/extended/baremetal/hosts.go
+++ b/test/extended/baremetal/hosts.go
@@ -3,12 +3,12 @@ package baremetal
 import (
 	"context"
 	"fmt"
-
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 
 	exutil "github.com/openshift/origin/test/extended/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -28,7 +28,7 @@ var _ = g.Describe("[sig-installer][Feature:baremetal] Baremetal/OpenStack/vSphe
 		c, err := e2e.LoadClientset()
 		o.Expect(err).ToNot(o.HaveOccurred())
 
-		metal3, err := c.AppsV1().Deployments("openshift-machine-api").Get(context.Background(), "metal3", metav1.GetOptions{})
+		metal3, err := c.AppsV1().Deployments(machineAPINamespace).Get(context.Background(), "metal3", metav1.GetOptions{})
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(metal3.Status.AvailableReplicas).To(o.BeEquivalentTo(1))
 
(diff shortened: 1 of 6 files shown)


ardaguclu:
Latest e2e-metal-ipi-serial failure is due to BMO is restarted.

/test e2e-metal-ipi-serial

ardaguclu:
Test passed, let's retest again to assure that consistently working;

/test e2e-metal-ipi-serial

ardaguclu:
None of the failures are related to this PR;

/test e2e-metal-ipi-serial

ardaguclu:
@andfasano @bfournie @elfosardo could you please take a look at this new test?. I think, finally it is ready for code review. Thanks.

test/extended/baremetal/hosts.go
```diff
@@ -186,4 +186,64 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
...
+		prvClient := provisioningClient(dc)
+
+		provisionings, err := prvClient.List(context.Background(), metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(provisionings.Items).NotTo(o.BeEmpty())
+
+		// expect only one provision config
+		provision := provisionings.Items[0]
+
+		realVal, found, err := unstructured.NestedBool(provision.Object, "spec", "virtualMediaViaExternalNetwork")
```
andfasano, line 207: This test seems to impact quite deeply the environment, thus I guess it'd be better to move it in the serial suite to avoid affecting other tests?

test/extended/baremetal/machine.go
```diff
@@ -0,0 +1,134 @@
...
+
+	scaleUpdate := s.DeepCopy()
+	scaleUpdate.Spec.Replicas = int32(replicas)
+	_, err = scaleClient.Scales(machineAPINamespace).Update(context.TODO(), schema.GroupResource{Group: "machine.openshift.io", Resource: "MachineSet"}, scaleUpdate, metav1.UpdateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+}
+
+// ScaleUpToExtraWorkers scales machine resources to provision
+// extra workers up to the given scale number.
+func ScaleUpToExtraWorkers(client dynamic.Interface, scale int) string {
```
andfasano, line 118: The scale up/down are interesting methods, but I'm wondering if such amount of work may be an indication that the current approach couldn't be different.
My main concern is that it seems it could be a good amount of logical overlap with the default scale&grow serial test, and obviously this functionality (scale up/down) is not a baremetal-only trait, so it could make sense to remain general.
So, given that, could make sense instead not to add a new e2e test, but a new job configuration where `virtualMediaViaExternalNetwork` is set (and run the usual serial suite)?

test/extended/baremetal/hosts.go
ardaguclu: This test is serial suite test which is running under https://github.com/openshift/origin/blob/f8e71ecca3a89c95ecbf0a11f77f434b2fe70c63/test/extended/baremetal/hosts.go#L154

test/extended/baremetal/machine.go
ardaguclu: All these scale up/down methods come here due to circular dependency between scale test and baremetal package. And I'm also not comfortable adding these just for this test. There are 2 options actually;
* After updating `virtualMediaViaExternalNetwork` to true, it already triggers all hosts re-provisioned and maybe there is no need to scale up extra worker. We can just check the current ones' statuses after update and revert back if it passes.
* Or like you said we can define this as a new job instead new test. That time, we'll only test `virtualMediaViaExternalNetwork` is true working or not. But not updating within running cluster.

I'm fine with both options.

andfasano: In this case, I'd opt for the second one. But I'd like to also to hear the opinions of @bfournie and @elfosardo on this topic

bfournie: I assume we need to move the host to *provisioned* to get any useful results, so we need to do the full scaling? (For example with the hostFirmwareSettings test I had started out doing a full scale up of the host but realized I could do a test just with the host in *Available*, but I I don't think that's the case here). If so I think its fine to duplicate the scale here (that's the same approach I originally took) but also the second option with a new job would probably be cleaner as suggested.

elfosardo: I would also go for the second option, it should be enough in this case

ardaguclu:
Thanks @andfasano @bfournie @elfosardo . This PR is deprecated, I'll introduce new environment variable in dev-scripts and define a new job configuration in release.

Thereby, I'm closing this PR.

/close

PR #26833, opened 2022-02-10
Allow a single test to override suite timeout

andfasano:
This patch adds the possibility to override the current timeout defined at the suite level for a specif test. This is required since in some specific cases a single test may require some extra time to successfully complete its execution, and it's not desirable to modify the global suite timeout since it will affect also other tests in the same suite.

The timeout is specified as part of the name using the special tag `[Timeout:<duration>]` with the required duration string, for example:

```
...
g.It("skip inspection when disabled by annotation [Timeout:300s]", func() {
...
```
In case of a malformed/incorrect duration string an error will be reported immediatly.

### DIFF
--- a/pkg/test/ginkgo/ginkgo.go
+++ b/pkg/test/ginkgo/ginkgo.go
@@ -21,7 +21,11 @@ func testsForSuite(cfg config.GinkgoConfigType) ([]*testCase, error) {
 			}
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


andfasano:
Related to https://github.com/openshift/origin/pull/26810

andfasano:
cc @soltysh @bfournie

soltysh (approved):
/lgtm
/approve

PR #26837, opened 2022-02-10
bug 2053582: fail on static pod lifecycle failures

deads2k:
related to https://bugzilla.redhat.com/show_bug.cgi?id=2053268

once the static pod operators produce messages containing "static pod lifecycle failure", this code will fail the e2e test.

This is only the easy half of the solution, but it has to be backported as well.


deads2k:
a proof PR is https://github.com/openshift/origin/pull/26839 .  That PR should fail

deads2k:
proof worked.

deads2k:
previously tested green. no conflict. merging

PR #26842, opened 2022-02-15
ip reconciler ignore

dgoodwin:
- ignore ip-reconciler failures in CI for now
- Improve ignore for ip-reconciler KubeJobFailed alert.

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -108,8 +108,15 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 			},
 		},
 		{
-			Selector: map[string]string{"alertname": "KubeJobFailed", "namespace": "openshift-multus"}, // not sure how to do a job_name prefix
+			Selector: map[string]string{"alertname": "KubeJobFailed", "namespace": "openshift-multus"},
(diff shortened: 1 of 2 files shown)


dgoodwin:
We'll add a test to isolate this as a specific test post-shiftweek as priorities allow.

dgoodwin:
Once CI is clear on this, I don't know of any way to test that it's actually working other than to merge it into the wild and watch that we stop getting the failure.

test/e2e/upgrade/alert/alert.go
```diff
@@ -107,6 +107,17 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
...
 		},
+		{
+			Selector: map[string]string{"alertname": "KubeJobFailed", "namespace": "openshift-multus"},
+			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=2054426",
+			Matches: func(sample *model.Sample) bool {
+				// Only match if the job_name label starts with ip-reconciler:
+				if !strings.HasPrefix(string(sample.Metric[model.LabelName("job_name")]), "ip-reconciler-") {
+					return false
+				}
+				return true
```
DennisPeriquet, line 118: you can probably consider this a nit, ... but, I see a double negative there.  I think it would read better if we take out the `!` and return true and then make line 118 as `return false`.

PR #26844, opened 2022-02-15
update alert thresholds after extending static pod install duration

deads2k:
https://github.com/openshift/cluster-etcd-operator/pull/738 extended the time it takes to rollout etcd.  This improved overall install and upgrade success, but it means the time over which we change leaders is extended. this means these alerts are firing for a longer period of time.

/assign @deepsm007


dgoodwin:
4.11 bug: https://bugzilla.redhat.com/show_bug.cgi?id=2055196

PR #26846, opened 2022-02-16
Bug 2054254: Rearrange test cases and gather more debugging information

slaskawi:
https://bugzilla.redhat.com/show_bug.cgi?id=2054254

This Pull Request doesn't provide a fix for the BZ mentioned the above. It provides a bit more verbose information that might be helpful for further debugging.

Furthermore, this Pull Request changes the order of test cases. This way, we should know whether this is a certificate related problem or a token one.

### DIFF
--- a/test/extended/oauth/requestheaders.go
+++ b/test/extended/oauth/requestheaders.go
@@ -134,25 +134,20 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 
 		testCases := []certAuthTest{
 			{
-				name:        "/authorize - challenging-client - valid cert: we should eventually get access token in Location header",
-				cert:        goodCert,
-				key:         goodKey,
-				endpoint:    "/oauth/authorize?client_id=openshift-challenging-client&response_type=token",
-				expectToken: true,
+				name:     "/healtz - anonymous: anyone should be able to access it",
+				endpoint: "/healthz",
 			},
 			{
-				name:          "/authorize - challenging-client - unknown CA cert: expect 302 because we never get authenticated",
-				cert:          unknownCACert,
-				key:           unknownCAKey,
-				endpoint:      "/oauth/authorize?client_id=openshift-challenging-client&response_type=token",
-				expectedError: "302 Found",
+				name:     "/healthz - valid cert",
+				cert:     goodCert,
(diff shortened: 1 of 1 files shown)


deads2k:
doesn't build

slaskawi:
Shuld be better now.

test/extended/oauth/requestheaders.go
```diff
@@ -234,6 +238,19 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 	})
 })
 
+func gatherPostMortem(oc *exutil.CLI) {
+	authn, err := oc.AdminConfigClient().ConfigV1().ClusterOperators().Get(context.Background(), "authentication", metav1.GetOptions{})
+	if err != nil {
+		e2e.Logf("Error getting authentication operator: %v\n", err)
+	}
+	e2e.Logf("Authentication %v\n", authn)
+	routerCA, err := oc.AdminKubeClient().CoreV1().ConfigMaps("openshift-config-managed").Get(context.Background(), "default-ingress-cert", metav1.GetOptions{})
```
stlaz, line 247: I'm not sure how retrieving this certificate helps, it's more or less a constant and if it were broken, many other things would be broken as well.

Maybe you may want to check:
a) the actual status of the oauth-openshift deployment
b) current config in the openshift-authentication NS
c) the logs of all the oauth-openshift deployment pods

test/extended/oauth/requestheaders.go
```diff
@@ -211,16 +211,20 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 				respDump, err := httputil.DumpResponse(resp, true)
 				o.Expect(err).NotTo(o.HaveOccurred())
 				if len(tc.expectedError) == 0 && resp.StatusCode != 200 && resp.StatusCode != 302 {
+					gatherPostMortem(oc)
```
stlaz, line 214: Don't ginkgo or gomega provide a better mechanism to perform actions upon a failure?

ibihim: Could you be more explicit? You mean adding it to some `afterEach` function or do you mean that there exists already a `gatherPostMortem` in our test suites?

PR #26849, opened 2022-02-16
Add architecture identification to job type key

deepsm007:
/assign @deads2k


pkg/synthetictests/platformidentification/types.go
```diff
@@ -6,25 +6,31 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
 )
 
+var oc = exutil.NewCLIWithoutNamespace("default")
```
deads2k, line 15: no.  This is a library type.

pkg/synthetictests/platformidentification/types.go
```diff
@@ -70,8 +76,12 @@ func GetJobType(ctx context.Context, clientConfig *rest.Config) (*JobType, error
 		platform = "ovirt"
 	case configv1.OpenStackPlatformType:
 		platform = "openstack"
+	case configv1.LibvirtPlatformType:
+		platform = "libvirt"
 	}
 
+	architecture := architecture(oc)
```
deads2k, line 83: pull this using the clientConfig

pkg/synthetictests/platformidentification/types.go
```diff
@@ -109,3 +120,26 @@ func versionFromHistory(history configv1.UpdateHistory) string {
...
+		switch node.Status.NodeInfo.Architecture {
+		case "amd64":
+			return "amd64"
+		case "arm64":
+			return "arm64"
+		case "ppc64le":
+			return "ppc64le"
+		case "s390x":
+			return "s390x"
+		default:
```
deads2k, line 141: default return amd64 please.

deads2k:
Please update the next best guess algorithm for matching to fallback to different architectures.  I suggest a method where you addMethods like `TryArchitecture("amd64"), TryArchitecture("")" to the list.

pkg/synthetictests/platformidentification/types.go
deepsm007: It is used [here](https://github.com/openshift/origin/pull/26849/files#diff-a395e190c5afd27cbc9248b8604043769a47e9ae58909aa8a69b95c8ab2a57b0R125)

deads2k:
/lgtm
/hold

This has to wait for someone with bigquery access to pull the values.  Add the commit from https://github.com/openshift/origin/pull/26882

deads2k:
hrm.  I'm not seeing libvirt there. Did the ci-tools get updated to track the new jobs?

deads2k:
do these only run once a week?  If so, you definitely need to add to the next-best-guess

deepsm007:
> hrm. I'm not seeing libvirt there. Did the ci-tools get updated to track the new jobs?

Ci-tools is not updated to use the platformtype

deepsm007:
> do these only run once a week? If so, you definitely need to add to the next-best-guess

The jobs run multiple times a day

PR #26852, opened 2022-02-21
allow slow kubelet updates for static pods, but still fail if they never succeed

deads2k:
in https://docs.google.com/document/d/1NCqpSzIcUVsIcRnsR0dLcB-ja1L20ePVWKKZM5sKw-U/edit, we found failures where the kubelet was slow in creating the new revision.  This lets it be late.


deads2k:
upgrade failed on service-load balancer disruption, not this change.  the intervenening merge was an owner's file.  merging this.

PR #26855, opened 2022-02-22
Set ovnkube-node readiness failure to flake for now.

xueqzhan:


pkg/synthetictests/networking.go
```diff
@@ -184,3 +187,34 @@ func getPodDeletionTime(events monitorapi.Intervals, podLocator string) *time.Ti
 	}
 	return nil
 }
+
+// bug is tracked here: https://bugzilla.redhat.com/show_bug.cgi?id=2057181
+func testOvnNodeReadinessProbe(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[bz-networking] ovnkube-node readiness probe should not fail repeatedly"
+	regExp := regexp.MustCompile(`ns/openshift-ovn-kubernetes pod/ovnkube-node-[a-z0-9-]+ node/[a-z0-9.-]+ - reason/Unhealthy Readiness probe failed:`)
```
deads2k, line 194: single constant used in both places please to ensure they stay in sync.

pkg/synthetictests/networking.go
```diff
@@ -16,6 +17,8 @@ type testCategorizer struct {
 	substring string
 }
 
+const ovnReadinessThreshold = 20
```
deads2k, line 20: how about a single threshold constant for both this and the original events failure.

PR #26858, opened 2022-02-24
Make a test to specifically look for image pull backoff for registry.redhat.io

DennisPeriquet:
We'd like to track cases like [this](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26854/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1496240296596344832):

```
event happened 45 times, something is wrong: ns/openshift-e2e-loki pod/loki-promtail-r6cbn node/ci-op-ipx1i5kt-db044-hw8hj-worker-b-p276c - reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
event happened 324 times, something is wrong: ns/openshift-e2e-loki pod/loki-promtail-h2zmg node/ci-op-ipx1i5kt-db044-hw8hj-master-2 - reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
```

and

```
ns/openshift-e2e-loki pod/loki-promtail-77rnl node/ip-10-0-182-230.ec2.internal container/oauth-proxy - reason/ContainerWait cause/ImagePullBackOff duration/696.91s Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
ns/openshift-e2e-loki pod/loki-promtail-7tpsj node/ip-10-0-148-202.ec2.internal container/oauth-proxy - reason/ContainerWait cause/ErrImagePull duration/671.53s rpc error: code = Unknown desc = loading manifest for target platform: reading manifest sha256:8e77bd9d4e81b4031620f32f0b5774aad843ee77fca56b555ca0a71f9ff64cd5 in registry.redhat.io/openshift4/ose-oauth-proxy: StatusCode: 404, Not found
```

### DIFF
--- a/pkg/synthetictests/image_pulls.go
+++ b/pkg/synthetictests/image_pulls.go
@@ -1,0 +1,99 @@
+package synthetictests
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+)
+
+const (
+	imagePullRedhatRegEx          = `reason/[a-zA-Z]+ .*Back-off pulling image .*registry.redhat.io`
+	imagePullRedhatFailThreshold  = 8
+	imagePullRedhatFlakeThreshold = 5
+)
+
+type eventRecognizerFunc func(event monitorapi.EventInterval) bool
+
+func matchEventForRegexOrDie(regex string) eventRecognizerFunc {
+	regExp := regexp.MustCompile(regex)
+	return func(e monitorapi.EventInterval) bool {
+		return regExp.MatchString(e.Message)
+	}
+}
+
+type singleEventCheckRegex struct {
+	testName       string
+	recognizer     eventRecognizerFunc
+	failThreshold  int
+	flakeThreshold int
+}
+
+// test goes through the events, looks for a match using the s.recognizer function,
+// if a match is found, and occurs more than s.failThreshold times, we mark it as a
+// flake (this allows us to see how often this symptom is happening).
+//
+func (s *singleEventCheckRegex) test(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	success := &junitapi.JUnitTestCase{Name: s.testName}
+	var failureOutput, flakeOutput []string
+	for _, e := range events {
+		if s.recognizer(e) {
+
+			msg := fmt.Sprintf("%s - %s", e.Locator, e.Message)
+			eventDisplayMessage, times := getTimesAnEventHappened(msg)
+			switch {
+			case times > s.failThreshold:
+				failureOutput = append(failureOutput, fmt.Sprintf("event [%s] happened %d times", eventDisplayMessage, times))
+			case times > s.flakeThreshold:
+				flakeOutput = append(flakeOutput, fmt.Sprintf("event [%s] happened %d times", eventDisplayMessage, times))
+			}
+		}
+	}
+	if len(failureOutput) > 0 {
+		totalOutput := failureOutput
+		if len(flakeOutput) >= 0 {
+			totalOutput = append(totalOutput, flakeOutput...)
+		}
+		failure := &junitapi.JUnitTestCase{
+			Name:      s.testName,
+			SystemOut: strings.Join(totalOutput, "\n"),
+			FailureOutput: &junitapi.FailureOutput{
+				Output: strings.Join(totalOutput, "\n"),
+			},
+		}
+		return []*junitapi.JUnitTestCase{failure}
+	}
+	if len(flakeOutput) > 0 {
+		failure := &junitapi.JUnitTestCase{
+			Name:      s.testName,
+			SystemOut: strings.Join(flakeOutput, "\n"),
+			FailureOutput: &junitapi.FailureOutput{
+				Output: strings.Join(flakeOutput, "\n"),
+			},
+		}
+		return []*junitapi.JUnitTestCase{failure, success}
+	}
+
+	return []*junitapi.JUnitTestCase{success}
+}
+
+func newSingleEventCheckRegex(testName, regex string) *singleEventCheckRegex {
+	return &singleEventCheckRegex{
+		testName:       testName,
+		recognizer:     matchEventForRegexOrDie(regex),
+		failThreshold:  imagePullRedhatFailThreshold,
+		flakeThreshold: imagePullRedhatFlakeThreshold,
+	}
+}
+
+// testBackoffPullingRegistryRedhatImage looks for this symptom:
+//   reason/ContainerWait ... Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+//   reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+// to happen over a certain threshold and marks it as a failure or flake accordingly.
+//
+func testBackoffPullingRegistryRedhatImage(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	testName := "[sig-arch] should not see excessive pull back-off on registry.redhat.io"
+	return newSingleEventCheckRegex(testName, imagePullRedhatRegEx).test(events)
+}
--- a/pkg/synthetictests/duplicated_events.go
+++ b/pkg/synthetictests/duplicated_events.go
@@ -97,6 +97,9 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
 
 	// we have a separate test for this
 	regexp.MustCompile(ovnReadinessRegExpStr),
+
+	// Separated out in testBackoffPullingRegistryRedhatImage
+	regexp.MustCompile(imagePullRedhatRegEx),
 }
 
 var allowedRepeatedEventFns = []isRepeatedEventOKFunc{
--- a/pkg/synthetictests/event_junits.go
+++ b/pkg/synthetictests/event_junits.go
@@ -36,6 +36,7 @@ func StableSystemEventInvariants(events monitorapi.Intervals, duration time.Dura
 	tests = append(tests, testOperatorOSUpdateStaged(events, kubeClientConfig)...)
 	tests = append(tests, testOperatorOSUpdateStartedEventRecorded(events, kubeClientConfig)...)
 	tests = append(tests, testPodNodeNameIsImmutable(events)...)
+	tests = append(tests, testBackoffPullingRegistryRedhatImage(events)...)
 
 	return tests
 }
@@ -64,6 +65,7 @@ func SystemUpgradeEventInvariants(events monitorapi.Intervals, duration time.Dur
 	tests = append(tests, testOperatorOSUpdateStaged(events, kubeClientConfig)...)
 	tests = append(tests, testOperatorOSUpdateStartedEventRecorded(events, kubeClientConfig)...)
 	tests = append(tests, testPodNodeNameIsImmutable(events)...)
+	tests = append(tests, testBackoffPullingRegistryRedhatImage(events)...)
 
 	return tests
 }
--- a/pkg/synthetictests/image_pulls_test.go
+++ b/pkg/synthetictests/image_pulls_test.go
@@ -1,0 +1,69 @@
+package synthetictests
+
+import (
+	"testing"
+	"time"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+)
+
+func Test_testBackoffPullingRegistryRedhatImage(t *testing.T) {
+	tests := []struct {
+		name      string
+		message   string
+		num_tests int
+		kind      string
+	}{
+		{
+			name:    "Test flake",
+			message: `ns/openshift-e2e-loki pod/loki-promtail-ww2rx node/ip-10-0-157-209.us-east-2.compute.internal reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest" (6 times)`,
+			kind:    "flake",
+		},
+		{
+			name:    "Test fail",
+			message: `ns/openshift-e2e-loki pod/loki-promtail-ww2rx node/ip-10-0-157-209.us-east-2.compute.internal reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest" (9 times)`,
+			kind:    "fail",
+		},
+		{
+			name:    "Test pass",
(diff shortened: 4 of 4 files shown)


pkg/synthetictests/operators.go
```diff
@@ -229,6 +230,36 @@ func testOperatorOSUpdateStartedEventRecorded(events monitorapi.Intervals, clien
...
 }
 
+// testBackoffPullingRegistryRedhatImage looks for this symptom:
+//   reason/ContainerWait ... Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+//   reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+// and marks it as a flake so we can see how often it's happening.
+//
+func testBackoffPullingRegistryRedhatImage(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+
+	image_name := "\"registry.redhat.io/openshift4"
```
deads2k, line 240: any image from registry.redhat.io

pkg/synthetictests/operators.go
```diff
@@ -229,6 +230,36 @@ func testOperatorOSUpdateStartedEventRecorded(events monitorapi.Intervals, clien
...
+	testName := "should be able to pull from registry.redhat.io"
```
deads2k, line 241: `[sig-arch]` prefix please

pkg/synthetictests/operators.go
```diff
@@ -229,6 +230,36 @@ func testOperatorOSUpdateStartedEventRecorded(events monitorapi.Intervals, clien
...
+	success := &junitapi.JUnitTestCase{Name: testName}
+
+	for _, e := range events {
+		match, _ := regexp.Match("reason/[a-zA-Z]+ .*Back-off pulling image "+image_name, []byte(e.Message))
```
deads2k, line 245: Extract the regex like Ken's PR and be sure you skip it in the "normal" test.

pkg/synthetictests/operators.go
```diff
@@ -229,6 +230,36 @@ func testOperatorOSUpdateStartedEventRecorded(events monitorapi.Intervals, clien
...
+//   reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+// and marks it as a flake so we can see how often it's happening.
+//
+func testBackoffPullingRegistryRedhatImage(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+
+	image_name := "\"registry.redhat.io/openshift4"
+	testName := "should be able to pull from registry.redhat.io"
+	success := &junitapi.JUnitTestCase{Name: testName}
+
+	for _, e := range events {
```
deads2k, line 244: we want similar threshold stuff like https://github.com/openshift/origin/pull/26855  Probably create a new struct that runs these test.

pkg/synthetictests/operators.go
bparees: in what context are we going to be running this test/job (against what payload) that we'd expect registry.redhat.io pullspecs?

i'd expect either `registry.ci.openshift.org`(image from ci payload), `quay.io`(image from nightly payload) or `image-registry.openshift-image-registry.svc:5000`(resolved via the oauth-proxy imagestream w/ pullthrough)

if we have components/test code that's setting up pods to use the registry.redhat.io image directly, we might need to be asking why.

DennisPeriquet: we have found image pull from registry.redhat.io to be unreliable.  It happens enough that it's noticeable so we'd like to track it.  The mitigation is usually to rerun the job where it happened.  In a recent case, the problem happened when trying to bring up loki during an [e2e-gcp-upgrade job in CI](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26854/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1496240296596344832).

This PR introduces a test (as a flake) so it won't affect pass/fail numbers.

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+)
+
+// testBackoffPullingRegistryRedhatImage looks for this symptom:
+//   reason/ContainerWait ... Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+//   reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+// to happen over a certain threshold and if so, marks it as a flake so we can see how often it's
+// happening.
+//
+func testBackoffPullingRegistryRedhatImage(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
```
deads2k, line 17: So this appears to match the other one, but could we do

```go
type eventRecognizerFunc func(event monitorapi.Interval) bool

type matchEventForRegexOrDie(regex string) eventRecognizerFunc{
  parse(regex)
  if err {
    panic
  }
  return func(){
    return regExp.MatchString(e.Message)
  }
}

type singleEventCheckRegex struct{
  testName string
  recognizer eventRecognizerFunc
  failThreshold int
}

func (s *singleEventCheckRegex) test(events) junittestcases{
	for _, e := range events {
		if s.recognizer(e) {

			// We found an instance of "Back-off pulling image"; if it happens more than some
			// threshold, report it as a flake.
			//
			msg := fmt.Sprintf("%s - %s", e.Locator, e.Message)
			eventDisplayMessage, times := getTimesAnEventHappened(msg)
			if times > s.failThreshold {
				output := fmt.Sprintf("event [%s] happened %d times\n", eventDisplayMessage, times)
				failure := &junitapi.JUnitTestCase{
					Name:      testName,
					SystemOut: output,
					FailureOutput: &junitapi.FailureOutput{
						Output: output,
					},
				}
				// Include a fake success so this will always be a "flake" for now.
				return []*junitapi.JUnitTestCase{failure, success}
			}
		}
	}

	return []*junitapi.JUnitTestCase{success}
}

func testBackoffPullingRegistryRedhatImage(events){
  testName := "[sig-arch] should not see excessive pull back-off on registry.redhat.io"   return 
  return newSingleEventCheckRegex(testName, regex).test(events)
}

```

deads2k: Then we can argue about a flakeThreshold (you'll need one to use here I think), and auto-exclude from the "normal" event test (this can come next week I think).

DennisPeriquet: whoa.  nice.  Thanks.  I added that with tweaks to get it to compile ok -- hopefully I interpreted what you are saying correctly (since it now seems more obvious how we can add another check using another regex with just a few lines).

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,76 @@
...
+				output := fmt.Sprintf("event [%s] happened %d times\n", eventDisplayMessage, times)
+				failure := &junitapi.JUnitTestCase{
+					Name:      s.testName,
+					SystemOut: output,
+					FailureOutput: &junitapi.FailureOutput{
+						Output: output,
+					},
+				}
+				// Include a fake success so this will always be a "flake" for now.
+				return []*junitapi.JUnitTestCase{failure, success}
```
deads2k, line 50: fail and flake are different.  If we're above the failureThreshold, it should be a straight failure, no success and return.  You can also add a flakeThreshold member and if we're above the flakeThreshold, we do this logic here.

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// test goes through the events, looks for a match using the s.recognizer function,
+// if a match is found, and occurs more than s.failThreshold times, we mark it as a
+// flake (this allows us to see how often this symptom is happening).
+//
+func (s *singleEventCheckRegex) test(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	success := &junitapi.JUnitTestCase{Name: s.testName}
+	for _, e := range events {
+		if s.recognizer(e) {
```
deads2k, line 33: You can actually match multiple events, so you cannot return early. You need to iterate over all of them and categorize into 

```
// messages for events above the fail threshold
failureMessages := []string{}
// messages for events above the flake threshold
flakeMessages := []string{}

// after this loop

if len(fails) > 0{
  return the failure only, with the output as a strings.join of failures and flakes

if len(flakes) >0{
  return the strings.join of flakes and the success like you did before.

otherwise return success
```

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,76 @@
...
+
+// testBackoffPullingRegistryRedhatImage looks for this symptom:
+//   reason/ContainerWait ... Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+//   reason/BackOff Back-off pulling image "registry.redhat.io/openshift4/ose-oauth-proxy:latest"
+// to happen over a certain threshold and if so, marks it as a flake so we can see how often it's
+// happening.
+//
+func testBackoffPullingRegistryRedhatImage(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	testName := "[sig-arch] should not see excessive pull back-off on registry.redhat.io"
+	regex := `reason/[a-zA-Z]+ .*Back-off pulling image .*registry.redhat.io`
```
deads2k, line 74: still missing the part where you skip events like this in the general event test.

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,95 @@
...
+	regExp := regexp.MustCompile(regex)
+	return func(e monitorapi.EventInterval) bool {
+		return regExp.MatchString(e.Message)
+	}
+}
+
+type singleEventCheckRegex struct {
+	testName      string
+	recognizer    eventRecognizerFunc
+	failThreshold int
```
deads2k, line 24: you want both `failThreshold int` and `flakeThreshold int`

DennisPeriquet:
Regarding thresholds for flake vs. fail.  If we go above failure threshold, we fail; if we go below flake threshold, we flake; but what do we call it when we are in between the two thresholds?

So I made the fail and flake thresholds equal (i.e., 20) so that if we exceed 20, we fail, and if we are below 20 we flake.

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,99 @@
...
+		if s.recognizer(e) {
+
+			msg := fmt.Sprintf("%s - %s", e.Locator, e.Message)
+			eventDisplayMessage, times := getTimesAnEventHappened(msg)
+			if times > s.failThreshold {
+				// Threshold exceeded, report it as a failure.
+				//
+				failureOutput = append(failureOutput, fmt.Sprintf("event [%s] happened %d times\n", eventDisplayMessage, times))
+			}
+			if times <= s.flakeThreshold {
```
deads2k, line 47: I other places we have logic like

```
switch{
case times > failThreshold:
  failure = append

case times > flake threshold
  flakes = append

default
  nothing
}

```

DennisPeriquet:
/hold
will look at [this failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26858/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1498308995176206336)

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,100 @@
...
+		if s.recognizer(e) {
+
+			msg := fmt.Sprintf("%s - %s", e.Locator, e.Message)
+			eventDisplayMessage, times := getTimesAnEventHappened(msg)
+			switch {
+			case times > s.failThreshold:
+				// Threshold exceeded, report it as a failure.
+				//
+				failureOutput = append(failureOutput, fmt.Sprintf("event [%s] happened %d times", eventDisplayMessage, times))
+			case times > 0 && times <= s.flakeThreshold:
```
deads2k, line 47: `times > flakeThreshold`

1. times==1, flakeThreshold==5, failThreshold==8 should be  a pass
2. times==6, flakeThreshold==5, failThreshold==8 should be a flake
3. times==9, flakeThreshold==5, failThreshold==8 should be a fail

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return []*junitapi.JUnitTestCase{success}
+}
+
+func newSingleEventCheckRegex(testName, regex string) *singleEventCheckRegex {
+	return &singleEventCheckRegex{
+		testName:       testName,
+		recognizer:     matchEventForRegexOrDie(regex),
+		failThreshold:  20,
+		flakeThreshold: 20,
```
deads2k, line 88: flakeThreshold=5 as a starting point.

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	return []*junitapi.JUnitTestCase{success}
+}
+
+func newSingleEventCheckRegex(testName, regex string) *singleEventCheckRegex {
+	return &singleEventCheckRegex{
+		testName:       testName,
+		recognizer:     matchEventForRegexOrDie(regex),
+		failThreshold:  8,
```
deads2k, line 82: ken made a constant. I made the example to illustrate my point.

deads2k:
please squash

/approve

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,95 @@
...
+		flakeThreshold: 5,
```
deads2k, line 83: this value is good.

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return []*junitapi.JUnitTestCase{success}
+}
+
+func newSingleEventCheckRegex(testName, regex string) *singleEventCheckRegex {
+	return &singleEventCheckRegex{
+		testName:       testName,
+		recognizer:     matchEventForRegexOrDie(regex),
+		failThreshold:  imagePullRedhatFailThreshold,
```
deads2k, line 86: set to `duplicateEventThreshold` in a followup

pkg/synthetictests/image_pulls.go
```diff
@@ -0,0 +1,99 @@
...
+	"strings"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+)
+
+const (
+	imagePullRedhatRegEx          = `reason/[a-zA-Z]+ .*Back-off pulling image .*registry.redhat.io`
+	imagePullRedhatFailThreshold  = 8
+	imagePullRedhatFlakeThreshold = 5
```
deads2k, line 15: I don't see a need for this to be a constant yet.

deads2k:
/lgtm

better than nothing, but the toleration needs tweaking.

PR #26859, opened 2022-02-24
Fix some test names that snuck in with trailing whitespace.

dgoodwin:
Caused a bug in sippy when filtering on test name equality. (fixed there
as well)


deads2k:
surely that list is generated somehow.  Does the source of that have trailing whitespace?

dgoodwin:
My bad didn't even realize, there are no other hits for ascii outside of vendor, so I'm not quite sure where these are coming from, I'll try to find out.

dgoodwin:
Found them over in the console repo. Will fix there.

/close

PR #26860, opened 2022-02-25
Fix ovnkube-node readiness test to include successes.

dgoodwin:
Test was skipping a junit on success, which appears to break aggregation
because it requires at least 3 results, and we're only getting 1 (a
failure) fairly commonly.

### DIFF
--- a/pkg/synthetictests/networking.go
+++ b/pkg/synthetictests/networking.go
@@ -205,14 +205,12 @@ func testOvnNodeReadinessProbe(events monitorapi.Intervals) []*junitapi.JUnitTes
 			}
 		}
 	}
+	test := &junitapi.JUnitTestCase{Name: testName}
 	if len(failureOutput) > 0 {
-		tests = append(tests, &junitapi.JUnitTestCase{
-			Name: testName,
-			FailureOutput: &junitapi.FailureOutput{
-				Output: failureOutput,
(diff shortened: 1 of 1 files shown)


pkg/synthetictests/networking.go
```diff
@@ -205,14 +205,12 @@ func testOvnNodeReadinessProbe(events monitorapi.Intervals) []*junitapi.JUnitTes
...
-			FailureOutput: &junitapi.FailureOutput{
-				Output: failureOutput,
-			},
-		})
-
+		test.FailureOutput = &junitapi.FailureOutput{
+			Output: failureOutput,
+		}
 	}
+	tests = append(tests, test)
```
deads2k, line 214: intentionally making this test never fail?  I suggest having a bug reference if that is your goal.

dgoodwin: Am I misreading something? tests is initialized empty, we only ever add one test to it, and that test has no failure output if successful, and will if it failed. I don't see how we make this never fail.

deads2k: > Am I misreading something? tests is initialized empty, we only ever add one test to it, and that test has no failure output if successful, and will if it failed. I don't see how we make this never fail.

yeah, my mistake.

PR #26862, opened 2022-02-28
Attempted fix for missing monitor events.

dgoodwin:
We have confirmed that events are being missed by the test monitor, which then appear in the events.json in must-gather/gather-extra steps. We believe this is due to the handling of resource version in this custom list watch code. Today is does a list of events and uses that resource version to establish the watch. When we encounter an error (which seems to be possible when apiservers are rolling), we repeat the process leaving a gap between the last event observed, and whatever the latest is when we re-establish and list again.

This is fixed by tracking the last resource version observed from incoming events, and using that to re-establish the watch.

### DIFF
--- a/pkg/monitor/event.go
+++ b/pkg/monitor/event.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -20,6 +19,9 @@ func startEventMonitoring(ctx context.Context, m Recorder, client kubernetes.Int
 	reMatchFirstQuote := regexp.MustCompile(`"([^"]+)"( in (\d+(\.\d+)?(s|ms)$))?`)
 
 	go func() {
+		// Track our last observed resource version from each event, used to re-establish the watch
+		// when it's requested rv gets too old for the server. (which may happen when apiservers cycle)
+		var rv string
 		for {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


pkg/monitor/event.go
```diff
@@ -20,6 +19,9 @@ func startEventMonitoring(ctx context.Context, m Recorder, client kubernetes.Int
 	reMatchFirstQuote := regexp.MustCompile(`"([^"]+)"( in (\d+(\.\d+)?(s|ms)$))?`)
 
 	go func() {
+		// Track our last observed resource version from each event, used to re-establish the watch
+		// when it's requested rv gets too old for the server. (which may happen when apiservers cycle)
+		var rv string
```
deads2k, line 24: how does this code handle the case where the resource version it wishes to watch from has expired and it has to re-list from the source to get a new resource version and then diff the content of the events it now sees?

pkg/monitor/event.go
```diff
@@ -64,11 +74,16 @@ func startEventMonitoring(ctx context.Context, m Recorder, client kubernetes.Int
 							if !ok {
 								continue
 							}
+							// Record the observed rv version, re-used if we lose our watch connection and re-establish
+							// to prevent missed events.
+							rv = obj.ResourceVersion
```
deads2k, line 79: this appears to be missing handling for bookmark events that are sent when a kube-apiserver shuts down.

pkg/monitor/event.go
```diff
@@ -8,175 +8,155 @@ import (
...
-			significantlyBeforeNow := time.Now().UTC().Add(-15 * time.Minute)
+	// filter out events written "now" but with significantly older start times (events
+	// created in test jobs are the most common)
+	significantlyBeforeNow := time.Now().UTC().Add(-15 * time.Minute)
 
-			events, err := client.CoreV1().Events("").List(ctx, metav1.ListOptions{Limit: 1})
-			if err != nil {
-				continue
+	listWatch := cache.NewListWatchFromClient(client.CoreV1().RESTClient(), "events", "", fields.Everything())
+	customStore := &cache.FakeCustomStore{
```
deads2k, line 26: I haven't looked in a while.  What does a delete look like?

deads2k, line 26: I suggest having a cache of event UID to resourceVersion (map[string]string) that avoids calling the record function if that UID with that resourceversion has already been passed to record once before.

dgoodwin: Good idea, done.

dgoodwin: Old code did not consider deletes at all, and that seemed intentional and correct to me as I don't think it matters when they get cleaned up to us. Let me know if there's anything I'm missing there.

dgoodwin:
/hold

Waiting for successful results.

dgoodwin:
I may be blocking execution, fix incoming once I confirm.

PR #26865, opened 2022-02-28
Implementation for basic egressfw e2e CI test

msherif1234:
Adding basic EgressFw CI test
- To manually run this test on ovnk cluster
```
./openshift-tests run-test "[sig-network][Feature:EgressFirewall] when using openshift ovn-kubernetes Should ensure egressfirewall is created [Suite:openshift/conformance/parallel]"
```
- To manually run this test on openshift-sdn cluster
```
./openshift-tests run-test "[sig-network][Feature:EgressFirewall] when using openshift-sdn Should ensure egressnetworkpolicy is created [Suite:openshift/conformance/parallel]"
```
- To run traffic in different name space
```
./openshift-tests run-test "[sig-network][Feature:EgressFirewall] EgressFirewall should have no impact outside its namespace [Suite:openshift/conformance/parallel]"
```
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>

### DIFF
--- a/test/extended/networking/egress_firewall.go
+++ b/test/extended/networking/egress_firewall.go
@@ -1,0 +1,172 @@
+package networking
+
+import (
+	"context"
+	"fmt"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/image"
+
+	kapiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+)
+
+const (
+	EgressFWTestPod      = "egressfirewall"
+	EgressFWE2E          = "egress-firewall-e2e"
+	NoEgressFWE2E        = "no-egress-firewall-e2e"
+	EgressFWTestImage    = "quay.io/redhat-developer/nfs-server:1.1"
+	OVNKManifest         = "ovnk-egressfirewall-test.yaml"
+	OpenShiftSDNManifest = "sdn-egressnetworkpolicy-test.yaml"
+)
+
+var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
+
+	egFwoc := exutil.NewCLI(EgressFWE2E)
+	egFwf := egFwoc.KubeFramework()
+
+	// The OVNKubernetes subnet plugin supports EgressFirewall objects.
+	InOVNKubernetesContext(
+		func() {
+			g.It("Should ensure egressfirewall is created", func() {
+				doEgressFwTest(egFwf, egFwoc, OVNKManifest)
+			})
+		},
+	)
+	// For Openshift SDN its supports EgressNetworkPolicy objects
+	InOpenShiftSDNContext(
+		func() {
+			g.It("Should ensure egressnetworkpolicy is created", func() {
+				doEgressFwTest(egFwf, egFwoc, OpenShiftSDNManifest)
+			})
+		},
+	)
+	noegFwoc := exutil.NewCLI(NoEgressFWE2E)
+	noegFwf := noegFwoc.KubeFramework()
+	g.It("EgressFirewall should have no impact outside its namespace", func() {
+		g.By("Creating test pod")
+		pod := "dummy"
+		o.Expect(createTestEgressFw(noegFwf, pod)).To(o.Succeed())
+		g.By("Sending traffic should all pass with no egress firewall impact")
+		_, err := noegFwoc.Run("exec").Args(pod, "--", "ping", "-c", "1", "8.8.8.8").Output()
+		expectNoError(err)
+
+		_, err = noegFwoc.Run("exec").Args(pod, "--", "ping", "-c", "1", "1.1.1.1").Output()
+		expectNoError(err)
+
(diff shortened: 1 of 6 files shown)


trozet:
@JacobTanenbaum PTAL
/assign @JacobTanenbaum

test/extended/networking/egress_firewall.go
```diff
@@ -0,0 +1,137 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
+)
+
+const (
+	EgressFWTestPod      = "egressfirewall"
+	EgressFWE2E          = "egress-firewall-e2e"
+	EgressFWTestImage    = "quay.io/redhat-developer/nfs-server:1.1"
+	OvnKManifest         = "ovnk-egressfirewall-test.yaml"
```
JacobTanenbaum, line 20: nit: initialism should be constant case: `OVNKManifest`

JacobTanenbaum:
I think we should add a test for blocking based on DNS name. Also could a sanity check be added to ensure that the egressfirewall does not apply outside the namespace in which it is created?

test/extended/testdata/bindata.go
```diff
@@ -41619,6 +41621,64 @@ func testExtendedTestdataDeploymentsTestDeploymentTestYaml() (*asset, error) {
 	return a, nil
 }
 
+var _testExtendedTestdataEgressFirewallOvnkEgressfirewallTestYaml = []byte(`apiVersion: k8s.ovn.org/v1
```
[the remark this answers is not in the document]
msherif1234: this is all auto generated code

msherif1234:
> > Also could a sanity check be added to ensure that the egressfirewall does not apply outside the namespace in which it is created?
> 
> That makes sense. Also, check the git log for bugfixes to that code for more ideas of stuff we might screw up. slightly_smiling_face

I am trying to have some basic coverage here to begin with instead of blank and we can keep growing it as we go

test/extended/testdata/bindata.go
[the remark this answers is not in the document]
JacobTanenbaum: sorry about that

JacobTanenbaum:
/LGTM could you add some JIRA cards for additional testing to egressFirewall if  you are not going to add them here?

test/extended/networking/egress_firewall.go
```diff
@@ -0,0 +1,160 @@
...
+	g.By("Sending traffic that does not matche allow cidr rule")
+	_, err = oc.Run("exec").Args(pod, "--", "ping", "-c", "1", "1.1.1.1").Output()
+	expectError(err)
+
+	// Test curl to docs.openshift.com should pass
+	// because we have allow dns rule for docs.openshift.com
+	g.By("Sending traffic that matches allow dns rule")
+	_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m1", "https://docs.openshift.com").Output()
+	expectNoError(err)
+	return nil
```
npinaeva, line 101: Maybe adding some failing curl test case can be useful here?

msherif1234: the rule in the fw policy in the yaml is to allow dns rule for the above https://docs.openshift.com  are u suggesting replace that with deny instead of allow to fail the curl ?

npinaeva: I mean adding a test similar to "Sending traffic that does not matche allow cidr rule" but for dns. 
So curl any denied dns and check that curl fails (with current fw policy).
Makes sense?

msherif1234: sure I will add another curl to different host and it should fail

test/extended/networking/egress_firewall.go
```diff
@@ -0,0 +1,169 @@
...
+func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string) error {
+	// Test ICMP / Ping to Google’s DNS IP (8.8.8.8) should pass
+	// because we have allow cidr rule for 8.8.8.8
+	g.By("Sending traffic that matches allow cidr rule")
+	_, err := oc.Run("exec").Args(pod, "--", "ping", "-c", "1", "8.8.8.8").Output()
+	expectNoError(err)
+
+	// Test ICMP / Ping to Cloudfare DNS IP (1.1.1.1) should fail
+	// because there is no allow cidr match for 1.1.1.1
+	g.By("Sending traffic that does not matche allow cidr rule")
```
npinaeva, line 95: nit: "matche" -> "match"

test/extended/networking/egress_firewall.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// Test curl to docs.openshift.com should pass
+	// because we have allow dns rule for docs.openshift.com
+	g.By("Sending traffic that matches allow dns rule")
+	_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m1", "https://docs.openshift.com").Output()
+	expectNoError(err)
+
+	// Test curl to www.google.com:80 should fail
+	// because we don't have allow dns rule for www.google.com:80
+	g.By("Sending traffic that doesn' match allow dns rule")
```
npinaeva, line 107: nit: "doesn'" -> "does not"

npinaeva (approved):
Looks good! Thanks for adding this!

dgoodwin:
Apologies but we had to revert this change due to permafailing on Azure and blocking payloads, details over in: https://github.com/openshift/origin/pull/26931

Worth noting that there's nothing on the PR that could have caught this, only aggregation did, absolutely no fault of the authors. Let us know if we can help get it back in safely.

PR #26868, opened 2022-03-01
MGMT-9440: Fix single node serial tests API crash.

paul-maidment:
It has been witnessed and reported in https://bugzilla.redhat.com/show_bug.cgi?id=2059581 that the KubeApi becomes unexpectedly unreachable for a short time after the completion of the  Serial API resiliance test. While we await the resolution of the BZ, I have introduced a 60 second wait at the end of the test to ensure that the tests after do not break.
It is our intention that this wait will be removed after the resolution of the BZ.

### DIFF
--- a/test/extended/apiserver/resiliency.go
+++ b/test/extended/apiserver/resiliency.go
@@ -94,6 +94,9 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
 		ginkgo.By("with no pods restarts during API disruption")
 		names := GetRestartedPods(c, restartingContainers)
 		gomega.Expect(len(names)).To(gomega.Equal(0), "Some pods in got restarted during kube-apiserver rollout: %s", strings.Join(names, ", "))
+
+		// Workaround for issues identified in https://bugzilla.redhat.com/show_bug.cgi?id=2059581
+		time.Sleep(60 * time.Second)
 	})
 
 })
--- a/test/extended/apiserver/resiliency.go
+++ b/test/extended/apiserver/resiliency.go
@@ -96,6 +96,7 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
 		gomega.Expect(len(names)).To(gomega.Equal(0), "Some pods in got restarted during kube-apiserver rollout: %s", strings.Join(names, ", "))
 
 		// Workaround for issues identified in https://bugzilla.redhat.com/show_bug.cgi?id=2059581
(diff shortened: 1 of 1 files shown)


test/extended/apiserver/resiliency.go
```diff
@@ -94,6 +94,9 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
 		ginkgo.By("with no pods restarts during API disruption")
 		names := GetRestartedPods(c, restartingContainers)
 		gomega.Expect(len(names)).To(gomega.Equal(0), "Some pods in got restarted during kube-apiserver rollout: %s", strings.Join(names, ", "))
+
+		// Workaround for issues identified in https://bugzilla.redhat.com/show_bug.cgi?id=2059581
```
omertuc, line 98:
```suggestion
		// Workaround for issues identified in https://bugzilla.redhat.com/show_bug.cgi?id=2059581
		// TODO: Remove this sleep once the bug above is resolved
```

test/extended/apiserver/resiliency.go
```diff
@@ -94,6 +94,10 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
...
+		// TODO: Remove this sleep once that bug is resolved
+		time.Sleep(60 * time.Second)
```
osherdp, line 100: maybe we can rather wait (for the same amount of time) with some condition to check api availability?

osherdp: another option is to check api availability right after this sleep duration (I prefer the first option, but with this we at least know if the sleep helped or not)

osherdp (changes requested):

test/extended/apiserver/resiliency.go
omertuc: >maybe we can rather wait (for the same amount of time) with some condition to check api availability?
This is already what the loop is doing. The issue is that during the period after the test, the API has fluctuating availability. It's available for a few seconds, then it isn't, then it's available again.

This is what the original test loop is already doing... checking the availability. The issue is that the API is reporting ready while it's not actually ready and has fluctuating availability, for about 10 seconds. We want to skip that period. It usually take like 10 seconds, we wait well beyond it, 60 seconds

osherdp: Ack. I didn't understand that it has fluctuating availability in that period of time

paul-maidment: Is the suggestion from @omertuc satisfactory or do we still need further actions here before we can merge?

PR #26871, opened 2022-03-01
[wip] add text to the timelines when you hover

deads2k:
similar to https://github.com/openshift/origin/pull/26870

it has done different things!


deads2k:
upgrade functioned, no conflict, just updates UI, merging.

PR #26873, opened 2022-03-02
Use common threshold constant as other event tests

DennisPeriquet:
followup to https://github.com/openshift/origin/pull/26858 re: "set to duplicateEventThreshold in a followup"

### DIFF
--- a/pkg/synthetictests/image_pulls.go
+++ b/pkg/synthetictests/image_pulls.go
@@ -11,7 +11,6 @@ import (
 
 const (
 	imagePullRedhatRegEx          = `reason/[a-zA-Z]+ .*Back-off pulling image .*registry.redhat.io`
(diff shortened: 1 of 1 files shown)


pkg/synthetictests/image_pulls.go
```diff
@@ -83,7 +83,7 @@ func newSingleEventCheckRegex(testName, regex string) *singleEventCheckRegex {
 	return &singleEventCheckRegex{
 		testName:       testName,
 		recognizer:     matchEventForRegexOrDie(regex),
-		failThreshold:  imagePullRedhatFailThreshold,
```
deads2k, line 86: delete old constant

deads2k:
/lgtm

will force merge on green build if you nag me.

deads2k:
constant swap, built and passed.

PR #26888, opened 2022-03-07
display pod lifecycle on intervals chart

deads2k:
This shows the basic pod and container lifecycle and the container readiness as reflected in the kube-api.

### DIFF
--- a/pkg/monitor/monitorapi/identification_pod.go
+++ b/pkg/monitor/monitorapi/identification_pod.go
@@ -1,0 +1,132 @@
+package monitorapi
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func LocatePod(pod *corev1.Pod) string {
+	return fmt.Sprintf("ns/%s pod/%s node/%s uid/%s", pod.Namespace, pod.Name, pod.Spec.NodeName, pod.UID)
+}
+
+func LocatePodContainer(pod *corev1.Pod, containerName string) string {
+	return fmt.Sprintf("ns/%s pod/%s node/%s uid/%s container/%s", pod.Namespace, pod.Name, pod.Spec.NodeName, pod.UID, containerName)
+}
+
+func PodFrom(locator string) PodReference {
+	parts := LocatorParts(locator)
+	namespace := NamespaceFrom(parts)
+	name := parts["pod"]
+	uid := parts["uid"]
+	if len(namespace) == 0 || len(name) == 0 || len(uid) == 0 {
+		return PodReference{}
+	}
+	return PodReference{
+		NamespacedReference: NamespacedReference{
+			Namespace: namespace,
+			Name:      name,
+			UID:       uid,
+		},
+	}
+}
+
+func ContainerFrom(locator string) ContainerReference {
+	pod := PodFrom(locator)
+	parts := LocatorParts(locator)
+	name := parts["container"]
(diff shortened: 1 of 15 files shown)


deads2k:
/hold

holding until we see this work.

pkg/monitor/pod.go
```diff
@@ -73,7 +73,39 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
 		return conditions
 	})
 
+	podScheduledFn := func(pod, oldPod *corev1.Pod) []monitorapi.Condition {
+		oldPodHasNode := oldPod != nil && len(oldPod.Spec.NodeName) > 0
+		newPodHasNode := pod != nil && len(pod.Spec.NodeName) > 0
+		if !oldPodHasNode && newPodHasNode {
+			return []monitorapi.Condition{
+				{
+					Level:   monitorapi.Warning,
```
dgoodwin, line 82: Should being scheduled be a warning? Feels like info to me.

pkg/monitor/monitorapi/identification_pod.go
```diff
@@ -0,0 +1,132 @@
...
+
+var (
+	// PodLifecycleTransitionReasons are the reasons associated with non-overlapping pod lifecycle states.
+	// A pod is logically identified by UID (I bet it's a name right now).
+	// Pods don't exist before create and don't exist after delete.
+	// Between those two states, each of these reasons can be ordered by time and used to create a contiguous view
+	// into the lifecycle of a pod.
+	PodLifecycleTransitionReasons = sets.NewString(
+		PodReasonCreated,
+		PodReasonScheduled,
```
dgoodwin, line 111: Is it feasible to cover Running as well? I found it strange looking at the html and seeing only Created/Scheduled, getting stuck there would indicate a problem.

dgoodwin (commented):
I feel like this would be dramatically more useful if we could get the intervals for a specific pod onto the same line as we see with the alerts pending->firing. It's a bit difficult to see the lifecycle for a specific pod when you're searching for matching strings in that compressed mess on the right.

Other major question is around why you skipped Running.

Otherwise looks like this will be very useful.

pkg/monitor/monitorapi/identification_pod.go
deads2k: > Is it feasible to cover Running as well? I found it strange looking at the html and seeing only Created/Scheduled, getting stuck there would indicate a problem.

Running is deceptive.  Pods have many containers and containers can be running, but the pod itself with a "running" phase doesn't tell you much useful.

dgoodwin: Does it tell us something with regards to not being stuck in Scheduled though? That the pod is doing "something", even if we don't know exactly what that is.

deads2k:
1. broke node state change colors
2. container readiness isn't showing up
3. pods we missed create for, but which *are* scheduled are not showing as dark green (this will require memory in the monitor)
4. e2e pods are not excluded from kube-apiserver view
5. possibly worth creating new e2e interval that contains the pods mixed with tests
6. containerwaiting is entirely extraneous in the alerts view now
7. division by openshift namespace is possibly a better sort than time
8. pods that have runOnce containers where all the containers are finished are in a terminated state well before deletion.  The chart will be easier to read if terminated to deleted is in a light color

deads2k:
> dramatically more useful if we could get the intervals for a specific pod onto the same line as we see with the alerts pending->firing.

I think we should hide the kubecontainerwaiting by default since we have a separate scheduling duration for pods.

deads2k:
/hold

holding until I fix the punchlist above

PR #26891, opened 2022-03-08
Push metal multiport service ipv6 test service IP further into service network.

dgoodwin:
Service network reportedly goes from fd02::0 to fd02::ffff so we're
grabbing 'face'.

This isn't failing often but we did see it twice in a row and thus
failing a payload.

### DIFF
--- a/test/extended/testdata/cmd/test/cmd/testdata/multiport-service-ipv6.yaml
+++ b/test/extended/testdata/cmd/test/cmd/testdata/multiport-service-ipv6.yaml
@@ -11,7 +11,7 @@ items:
     resourceVersion: "259"
     uid: 024d82eb-7193-11e5-b84d-080027c5bfa9
   spec:
-    clusterIP: fd02::9
+    clusterIP: fd02::face
     ports:
     - name: web
       port: 5432
--- a/test/extended/testdata/cmd/test/cmd/testdata/external-service-ipv6.yaml
+++ b/test/extended/testdata/cmd/test/cmd/testdata/external-service-ipv6.yaml
@@ -7,7 +7,7 @@ metadata:
   resourceVersion: "1"
   uid: 19cff995-5546-11e5-9f57-080027c5bfa9
 spec:
-  clusterIP: fd02::8
+  clusterIP: fd02::dace
   ports:
   - nodePort: 0
     port: 443


dgoodwin:
Failure can be seen here: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.11-e2e-metal-ipi-ovn-ipv6/1501131370355757056

test/extended/testdata/cmd/test/cmd/testdata/multiport-service-ipv6.yaml
```diff
@@ -11,7 +11,7 @@ items:
     resourceVersion: "259"
     uid: 024d82eb-7193-11e5-b84d-080027c5bfa9
   spec:
-    clusterIP: fd02::9
+    clusterIP: fd02::face
```
andfasano, line 14: Given that the `[sig-cli] oc expose can ensure the expose command is functioning as expected` test doesn't seem to have a strong dependency on the service IP, wouldn't be better to omit it at all (as it seems to happen with the ipv4 version), to avoid any potential collision with other previously created services?
(btw, the test uses also another file with an hard-coded service ip: https://github.com/openshift/origin/blob/684a48ff4b5cb9387164a103d1bae9aca2621c3d/test/extended/testdata/cmd/test/cmd/testdata/external-service-ipv6.yaml#L10)

dgoodwin: I'm a little worried that would fall back to ipv4, but I'm not clear on how this test was intended to work entirely and know very little about ipv6. @andfasano could you confirm this is not violating the spirit of the test?

I'll get the other ref updates as well.

dgoodwin: Ok ref updated.

If we remove the service IP entirely, the file is then identical with the ipv4 and I'm really suspicious that would nullify the test.

andfasano: That's a good point @dgoodwin: previously I saw that the api server service (the one initially found as a clash) didn't seem to specify it, and it got an ipv6 address - maybe it could be worht a try.
In such case, my biggest concern is that simply changing the `clusterIP` will not be enough to prevent a clash, given that any other previously created  service could get a conflicting ip.

andfasano: From a quick local test it seems fine (the service without a specific `clusterIP` gets an ipv6 address in a ipv6 context). The test also already differentiates the services created based on the current context: https://github.com/openshift/origin/blob/684a48ff4b5cb9387164a103d1bae9aca2621c3d/test/extended/cli/expose.go#L59-L69

dgoodwin: @andfasano would you or someone with knowledge be able to craft the change to the test? I am happy to close this one, and it would be nice if git history reflects the individuals with knowledge of why the change is correct and safe.

andfasano: Yep, sure happy to help. I can open a new PR with the proposed change and see if it works, if it's fine for you.

dgoodwin: Sounds good to me, thank you!

dgoodwin:
/close 

Replaced by @andfasano PR to remove the ipv6 files entirely.

PR #26892, opened 2022-03-08
Pod events 3 all

deads2k:
/hold


deads2k:
This needs to be held until run once pods (pods that only contain runonce containers, properly show end times.

and probably until the sort is by namespace.

PR #26895, opened 2022-03-09
Remove fixed clusterIP service for ipv6 contexts

andfasano:
When running in a ipv6 context, the oc expose test may fail due a conflict with a previously allocated service


andfasano:
cc @dgoodwin

dgoodwin:
Couple thoughts, you can likely delete the files, I believe they are now identical to their v4 counterparts, and the code which forks the file.
It also looks like you got bindata but not the files that source it, and I suspect that will get overridden or rejected by checks.
And could you update commit message with analysis on why it's a safe change based on the testing you did locally in the other PR.

Thanks for updating!

andfasano:
Hi @dgoodwin , could you please have another look? In the meantime I triggered the ipv6 builds, in order to verify that the patch it's working as expected.

dgoodwin:
/hold 

Waiting to see successful results.

dgoodwin:
/hold cancel

Wrong PR :)

elfosardo:
@dgoodwin @stbenjam can this be approved now? please let us know if anything is missing

PR #26897, opened 2022-03-09
Mark image pull from registry.redhat.io as flake to track but not fail

DennisPeriquet:


pkg/synthetictests/image_pulls.go
```diff
@@ -62,7 +62,9 @@ func (s *singleEventCheckRegex) test(events monitorapi.Intervals) []*junitapi.JU
 				Output: strings.Join(totalOutput, "\n"),
 			},
 		}
-		return []*junitapi.JUnitTestCase{failure}
+
+		// Even though this is a failure, we will mark it as a flake for now.
+		return []*junitapi.JUnitTestCase{failure, success}
```
deads2k, line 67: This change looks incorrect.  Don't you want to update the failThreshold to maxInt?

PR #26899, opened 2022-03-09
Add early test to check router pods are not on same node

xueqzhan:


test/extended/router/pods.go
```diff
@@ -0,0 +1,44 @@
...
+
+	g.BeforeEach(func() {
+		var err error
+		_, err = exutil.WaitForRouterServiceIP(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+	})
+
+	g.Describe("The HAProxy router pods", func() {
+		g.It("should be scheduled on different nodes", func() {
+			pods, err := oc.KubeFramework().ClientSet.CoreV1().Pods("openshift-ingress").List(context.Background(), metav1.ListOptions{})
```
deads2k, line 27: Please abstract this into a 

```go
// requirePodsOnDifferentNodesTest fails if any pods with the prefix in the namespace are scheduled to the same node
type requirePodsOnDifferentNodesTest struct{
    namespace string
    podPrefix string
}

func (p requirePodsonDifferentNodesTest) run(whatever args){
    pods, err := oc.KubeFramework().ClientSet.CoreV1().Pods("openshift-ingress").List(context.Background(), metav1.ListOptions{})
    // until the end
}

g.Describe("The HAProxy router pods", func() {
    g.It("should be scheduled on different nodes", func() {
        requirePodsOnDifferentNodesTest{namespace:, podPrefix}.run(args)
    }
}

g.Describe("The openshift-apiserver pods", func() {
    g.It("should be scheduled on different nodes", func() {
        requirePodsOnDifferentNodesTest{namespace: openshift-apiserver, podPrefix: openshift-apiserver-}.run(args)
    }
}

g.Describe("The oauth-apiserver pods", func() {
    g.It("should be scheduled on different nodes", func() {
        requirePodsOnDifferentNodesTest{namespace: oauth-apiserver, podPrefix: oauth-apiserver-}.run(args)
    }
}


```

test/extended/router/pods.go
```diff
@@ -0,0 +1,44 @@
...
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"strings"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("[sig-network][Early][Feature:Router]", func() {
```
deads2k, line 14: sig-scheduling

deads2k, line 14: not [Feature:Router] this is checking scheduling

deads2k:
need to run hack/update-generated something

deads2k:
/approve
/override ci/prow/e2e-gcp 

override because of image pull on b02

deads2k:
/assign @stbenjam 

if this appears to be working, then we'll need to make an [early] invariant check before we run any upgrade tests since upgrades will clear this problem out.

deads2k:
we'll start here and likely have to make this execute when the test first begins.  Before starting the ugprade.

/lgtm

omertuc:
This is consistently failing on single-node clusters. Please revert / fix this

omertuc:
Okay fix seems simple, no need to revert, created #26905

PR #26900, opened 2022-03-10
Allow CurrentVersionNotVerified alert

Davoska:
This pull request should add the alert `CurrentVersionNotVerified` as an allowed alert to tests that are making sure any unexpected alerts are not firing.

The `CurrentVersionNotVerified` alert is supposed to be firing when the verification of the current version of a cluster has failed and the version was forcefully applied to the cluster. Versions of the clusters used in the CI are forcefully applied and may not be signed. Thus the verification is expected to fail.

The alert `CurrentVersionNotVerified` is not as of this moment present. The alert will be added in the https://github.com/openshift/cluster-version-operator/pull/734 which is currently failing the CI tests regarding unexpected alerts firing. This pull request should solve this issue. 

Pull Request related to https://issues.redhat.com/browse/OTA-187


Davoska:
I am not aware of a method of testing these changes against the https://github.com/openshift/cluster-version-operator/pull/734 pull request without merging this pull request.

If you have any suggestions or tips, please let me know!

Davoska:
/close
With the severity level set to `info`, this pull request is no longer necessary.

PR #26904, opened 2022-03-14
MGMT-9440: Fix single node serial tests API crash

paul-maidment:
This is a follow on from https://github.com/openshift/origin/pull/26868
where we introduced a working to prevent single-node serial tests from
failing.

This patch rectfies that by removing the artificial delay and patching
the root cause, which is due to the way errors are processed when making
a call to isApiReady. The existing implementation will "bubble up"
errors from the httpClient.Get call to the caller of isApiReady.
However, at this stage, errors are undesirable, we simply want to ask
the question "is the API ready or not? -- for any reason"

So I have simplified the method to return a boolean response even if
errors are encountered.

### DIFF
--- a/test/extended/apiserver/resiliency.go
+++ b/test/extended/apiserver/resiliency.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"github.com/openshift/origin/test/extended/operators"
-	"io/ioutil"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"net/http"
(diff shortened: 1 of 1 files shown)


osherdp:
maybe premature, but figured out no harm in trying it right away

omertuc:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26904/pull-ci-openshift-origin-master-e2e-aws-single-node-serial/1503648889381588992 test is still messing up the tests after it

omertuc:
Please remove WIP

pacevedom (approved):
This PR will fix lots of flaky tests in single node, can we move it forward? Thanks!

test/extended/apiserver/resiliency.go
```diff
@@ -69,19 +68,23 @@ var _ = ginkgo.Describe("[Conformance][sig-sno][Serial] Cluster", func() {
 		forceApiRollout(oc)
 
 		ginkgo.By("Expecting API to become unavailable")
-		wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
-			ready, _, err := isApiReady(config, httpClient)
-			return !ready, err
+		err = wait.PollImmediate(time.Second, 5*time.Minute, func() (bool, error) {
```
mfojtik, line 71: why is this increased to 5 minutes?

omertuc: Turns out that it takes slightly more than a minute for the API server to go down after a rollout is triggered, and is the reason this test has been wrecking havoc on the tests following it - we never checked the return value of PollImmediate so the subsequent PollImmediate waiting for the API server to come back up would finish immediately (as it never went down in the first place, yet) then the API server would go down while the following tests execute

pacevedom: Sometimes it takes more than 1 minute for API server to stop responding (tested empirically in a single node cluster, measured over 70s on a regular basis). You want to catch that to test availability afterwards.

paul-maidment: @mfojtik as [@Omer](https://coreos.slack.com/team/U01DQKJCMCM) responded to highlight that the API server often takes longer than 1 minute to "go  down", this point also supported by @pacevedom , we need a longer time-out.

Whereas 5 minutes may be overkill, perhaps this should be 2 minutes or 3 minutes to prevent flakiness?

What are your thoughts?

PR #26907, opened 2022-03-14
Filter out ovn readiness event during installation

xueqzhan:
Attempt to filter out ovn node readiness event that happened during initial installation.


pkg/synthetictests/duplicated_events.go
```diff
@@ -389,6 +390,75 @@ func getTimesAnEventHappened(message string) (string, int) {
 	return matches[0][1], int(times)
 }
 
+func getInstallCompletionTime(kubeClientConfig *rest.Config) *metav1.Time {
+	configClient, err := configclient.NewForConfig(kubeClientConfig)
+	if err == nil {
```
deads2k, line 395: project standard is `if err != nil` with an early return to avoid nesting.

pkg/synthetictests/duplicated_events.go
```diff
@@ -389,6 +390,75 @@ func getTimesAnEventHappened(message string) (string, int) {
...
+		clusterVersion, err := configClient.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
+		if err == nil && len(clusterVersion.Status.History) != 0 {
+			return clusterVersion.Status.History[len(clusterVersion.Status.History)-1].CompletionTime
+		}
+	}
+	return nil
+}
+
+func getMatchedElementsFromMonitorEventMsg(regExp *regexp.Regexp, message string) (namespace, pod, node, reason, msg string) {
```
deads2k, line 404: no named return values.

pkg/synthetictests/duplicated_events.go
```diff
@@ -389,6 +390,75 @@ func getTimesAnEventHappened(message string) (string, int) {
...
+		return true
+	}
+	installCompletionTime := getInstallCompletionTime(kubeClientConfig)
+	if installCompletionTime == nil {
+		// default to OK
+		return true
+	}
+
+	message := fmt.Sprintf("%s - %s", monitorEvent.Locator, monitorEvent.Message)
+	namespace, pod, _, reason, msg := getMatchedElementsFromMonitorEventMsg(regExp, message)
```
deads2k, line 441: if you don't get what you need, you need to explicitly fail or default ok or something.

deads2k: > if you don't get what you need, you need to explicitly fail or default ok or something.
Reading through, I recommend a recognizeable error. When this fails, it means a rather complicated regex failed.  So we want to fail early.

pkg/synthetictests/networking.go
```diff
@@ -2,6 +2,7 @@ package synthetictests
 
 import (
 	"fmt"
+	"k8s.io/client-go/rest"
```
deads2k, line 5: nit: order. put this in a second below the standard imports.

deads2k:
This is a good idea.  I think it needs some stronger mechanism to detect coding failures.

pkg/synthetictests/duplicated_events.go
xueqzhan: Do you mean to fail the test or default to ok? For the case of allowedRepeatedEventFns, this is only used to return true/false.

pkg/synthetictests/duplicated_events.go
```diff
@@ -288,31 +290,45 @@ func (d duplicateEventsEvaluator) testDuplicatedE2ENamespaceEvents(events monito
...
 		if times > duplicateEventThreshold {
 			if allowedRepeatedEventsRegex.MatchString(eventDisplayMessage) {
 				continue
 			}
+			eInfo := eventInfo{count: times}
 			allowed := false
 			for _, allowRepeatedEventFn := range d.allowedRepeatedEventFns {
-				if allowRepeatedEventFn(event, kubeClientConfig) {
-					allowed = true
+				eInfo.err, allowed = allowRepeatedEventFn(event, kubeClientConfig)
```
deads2k, line 307: could just just define `var failures []string` above the for loop and when `allowRepeatedEventFn` had an err, simple do `failures = append(failures, err.Error)`?

pkg/synthetictests/duplicated_events.go
```diff
@@ -389,86 +411,106 @@ func getTimesAnEventHappened(message string) (string, int) {
...
 	}
 	if len(clusterVersion.Status.History) == 0 {
-		// default to OK
-		return true
+		return nil
+	}
+	return clusterVersion.Status.History[len(clusterVersion.Status.History)-1].CompletionTime
+}
+
+func getMatchedElementsFromMonitorEventMsg(regExp *regexp.Regexp, message string) (error, string, string, string, string, string) {
```
deads2k, line 429: error last

pkg/synthetictests/duplicated_events.go
```diff
@@ -389,86 +411,106 @@ func getTimesAnEventHappened(message string) (string, int) {
...
 	}
-	initialInstallHistory := clusterVersion.Status.History[len(clusterVersion.Status.History)-1]
-	if initialInstallHistory.CompletionTime == nil {
+	return nil, namespace, pod, node, reason, msg
+}
+
+// isEventDuringInstallation returns true if the monitorEvent represents a real event that happened after installation.
+// regExp defines the pattern of the monitorEvent message. Named match is used in the pattern using `(?P<>)`. The names are placed inside <>. See example below
+// `ns/(?P<NS>openshift-ovn-kubernetes) pod/(?P<POD>ovnkube-node-[a-z0-9-]+) node/(?P<NODE>[a-z0-9.-]+) - reason/(?P<REASON>Unhealthy) (?P<MSG>Readiness probe failed:.*$`
+func isEventDuringInstallation(monitorEvent monitorapi.EventInterval, kubeClientConfig *rest.Config, regExp *regexp.Regexp) (error, bool) {
```
deads2k, line 456: `bool, error`

pkg/synthetictests/duplicated_events.go
```diff
@@ -389,86 +411,106 @@ func getTimesAnEventHappened(message string) (string, int) {
...
+			node = subMatches[i]
+		case "REASON":
+			reason = subMatches[i]
+		case "MSG":
+			msg = subMatches[i]
+		}
 	}
-	initialInstallHistory := clusterVersion.Status.History[len(clusterVersion.Status.History)-1]
-	if initialInstallHistory.CompletionTime == nil {
+	return nil, namespace, pod, node, reason, msg
```
deads2k, line 450: if any of these are empty, we should fail, right?  Seems like it means the regex failed.

pkg/synthetictests/duplicated_events.go
```diff
@@ -389,86 +411,106 @@ func getTimesAnEventHappened(message string) (string, int) {
...
-		}
-		if !strings.Contains(event.Message, "connect: connection refused") {
+	for _, event := range kubeEvents.Items {
+		if event.Related.Name != pod ||
+			event.Reason != reason ||
+			!strings.Contains(event.Message, msg) {
 			continue
 		}
 
-		// if the readiness probe failure for this pod happened AFTER the initial installation was complete,
```
deads2k, line 463: this comment is still  helpful I think.

PR #26912, opened 2022-03-16
Clarify that the image mirroring requirement applies to conformance tests only

bparees:
This is my understanding after discussion with @derekwaynecarr

### DIFF
--- a/test/extended/util/image/README.md
+++ b/test/extended/util/image/README.md
@@ -1,8 +1,8 @@
 # Images used by e2e tests
 
-We limit the set of images used by e2e to reduce duplication and to allow us to provide offline mirroring of images for customers and restricted test environments. Every image used in e2e must be part of this utility package or referenced by the upstream `k8s.io/kubernetes/test/utils/image` package.
+We limit the set of images used by conformance e2e to reduce duplication and to allow us to provide offline mirroring of images for customers and restricted test environments. Every image used in e2e must be part of this utility package or referenced by the upstream `k8s.io/kubernetes/test/utils/image` package.
 
-All images used by e2e are mirrored to:
+All images used by e2e tests that are part of the `conformance` suite are mirrored to:
 
   quay.io/openshift/community-e2e-images:TAG
 


bparees:
@smarterclayton @sttts @soltysh ptal

The lack of clarity here caused some consternation recently as we discovered a bunch of build tests that consume redhat product images for exercising app building/deployment.

While it's probably undesirable that they do that, i don't see anyone investing in rewriting them to use an alternate image any time soon, nor, i imagine, do we want to mirror those product images since they aren't ours to publish.

test/extended/util/image/README.md
```diff
@@ -1,8 +1,8 @@
 # Images used by e2e tests
 
-We limit the set of images used by e2e to reduce duplication and to allow us to provide offline mirroring of images for customers and restricted test environments. Every image used in e2e must be part of this utility package or referenced by the upstream `k8s.io/kubernetes/test/utils/image` package.
+We limit the set of images used by conformance e2e to reduce duplication and to allow us to provide offline mirroring of images for customers and restricted test environments. Every image used in e2e must be part of this utility package or referenced by the upstream `k8s.io/kubernetes/test/utils/image` package.
```
soltysh, line 3: I don't think this applies to conformance tests only, the way we have it [configured](https://github.com/openshift/origin/blob/1c7bf01e164c265713888e73589dd437a61eb000/cmd/openshift-tests/openshift-tests.go#L356) we don't differentiate between conformance and non-conformance images.

bparees: pulledInvalidImages allows pulls from registry.redhat.io, among other places (for particular images), it is not enforcing that everything come from quay.io/community-e2e

bparees: in fact weirdly, that code does not allow quay.io/community-e2e which raises a whole other set of questions about how the jobs are being run: https://github.com/openshift/origin/blob/1c7bf01e164c265713888e73589dd437a61eb000/cmd/openshift-tests/images.go#L190

bparees: and as another example of the murky waters we are currently in, take this test fixture for example:
https://github.com/openshift/origin/blob/3b2752de7293a8dad552766fef2089bfbbac8b36/test/extended/testdata/deployments/deployment-example.yaml#L43-L50

technically this is ok(passed image prefix validation) because it results in a pod that pulls from the internal registry because the imagestream it references is configured with local reference policy(i.e. it uses pullthrough).  But the reality is that the image being pulled is coming from registry.redhat.io and unless someone mirrored the content and configured the samples operator to ref the mirror for the imagestreams it manages, it won't actually work in a disconnected environment.

It certainly isn't meeting the currently stated policy of "All images used by e2e are mirrored to quay.io/openshift/community-e2e-images:tag"

PR #26913, opened 2022-03-16
Migrate tests away from using samples-operator provided imagestreams

bparees:
Resolving some of the failures seen in:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-e2e-aws-no-capabilities/1503711956505202688

which runs our e2e but with optional capabilities (in this case, samples operator)
disabled.

### DIFF
--- a/test/extended/deployments/deployments.go
+++ b/test/extended/deployments/deployments.go
@@ -545,6 +545,13 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 
 		g.It("should run a successful deployment with multiple triggers", func() {
 			g.By("creating DC")
+
+			_, err := oc.Run("import-image").Args("registry.redhat.io/ubi8/ruby-30:latest", "--confirm", "--reference-policy=local").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			_, err = oc.Run("import-image").Args("registry.redhat.io/rhel8/postgresql-13:latest", "--confirm", "--reference-policy=local").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
 			dc, err := createDeploymentConfig(oc, multipleICTFixture)
 			o.Expect(err).NotTo(o.HaveOccurred())
 			o.Expect(dc.Name).To(o.Equal(dcName))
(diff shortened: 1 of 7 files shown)


bparees:
@soltysh ptal for the deployment test changes
@adambkaplan @gabemontero ptal for the template test changes
@dmage @ricardomaraschini ptal for the image command test changes

gabemontero:
/lgtm

I'll also 

/hold

to allow the others to chime in.  Of course feel free to cancel the hold when you deem it appropriate @bparees

test/extended/deployments/deployments.go
```diff
@@ -545,6 +545,13 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 
 		g.It("should run a successful deployment with multiple triggers", func() {
 			g.By("creating DC")
+
+			_, err := oc.Run("import-image").Args("registry.redhat.io/ubi8/ruby-30:latest", "--confirm", "--reference-policy=local").Output()
```
dmage, line 549: Shouldn't we use only images that are mirrored? https://github.com/openshift/origin/blob/master/test/extended/util/image/README.md

bparees: i'm trying to determine the answer to that here:

https://github.com/openshift/origin/pull/26912

because many of these tests already weren't using images that were mirrored (this one, for example, was just using the ruby imagestream provided by the samples operator and hosted on registry.redhat.io already).

so this was the minimal change to get away from using that imagestream while still using the same underlying image.

since i'm not confident we should be mirroring images like this (redhat product images), i'm hoping we are allowed use direct references for non-conformance tests.

bparees: @dmage https://github.com/openshift/origin/pull/26912 has clarified that only conformance tests need their images mirrored, these are not conformance tests.

bparees:
@soltysh any concerns on the deploymentconfig test changes?

bparees:
@soltysh bump

@dmage any other concerns?

soltysh (approved):
/lgtm
/hold cancel

PR #26916, opened 2022-03-17
Attempt multiple different token requests

arjunrn:
The token request with client header authorization does not encode the header client identifier payload with URLEscape like the other request methods. The current oauth-server implements it so but this behavior has changed in newer versions of the dependency. So now the test which makes the token requests can multiple requests to fetch the token and as long as one of them succeeds the test is considered as succeeded.

### DIFF
--- a/test/extended/oauth/client_secret.go
+++ b/test/extended/oauth/client_secret.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
-	"encoding/base64"
 	"fmt"
 	"net/http"
 	"net/http/httputil"
@@ -18,6 +17,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	oauthv1 "github.com/openshift/api/oauth/v1"
 	userv1 "github.com/openshift/api/user/v1"
+
 	exutil "github.com/openshift/origin/test/extended/util"
 
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
@@ -66,7 +66,7 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 
 			oauthDiscoveryData := getOAuthWellKnownData(oc)
 
-			for _, createTokenReq := range []func(string) *http.Request{
+			for _, createTokenReq := range []func(string) []*http.Request{
 				queryClientAuthRequest,
 				bodyClientAuthRequest,
 				headerClientAuthRequest,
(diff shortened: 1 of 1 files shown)


test/extended/oauth/client_secret.go
```diff
@@ -14,13 +14,13 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
```
slaskawi, line 17: Please leave the imports where they were. This change is not required.

slaskawi (changes requested):
Only one nitpick. Otherwise looks good.

slaskawi (approved):
/lgtm
/approve

test/extended/oauth/client_secret.go
```diff
@@ -136,17 +148,24 @@ func bodyClientAuthRequest(tokenEndpoint string) *http.Request {
...
-func headerClientAuthRequest(tokenEndpoint string) *http.Request {
-	req, err := http.NewRequest("POST", tokenEndpoint,
+func headerClientAuthRequest(tokenEndpoint string) []*http.Request {
+	reqNoURLEscape, err := http.NewRequest("POST", tokenEndpoint,
 		bytes.NewBufferString("grant_type=authorization_code&"+
 			"code="+authzToken,
 		),
 	)
 	o.Expect(err).NotTo(o.HaveOccurred())
+	reqNoURLEscape.Header.Set("Authorization", fmt.Sprintf("Basic %s", base64.StdEncoding.EncodeToString([]byte("oauth-client-with-plus:secret+with+plus"))))
```
stlaz, line 161: Please use https://pkg.go.dev/net/http#Request.SetBasicAuth for setting basic auth

test/extended/oauth/client_secret.go
```diff
@@ -136,17 +148,24 @@ func bodyClientAuthRequest(tokenEndpoint string) *http.Request {
...
 
-	req.Header.Set("Authorization", fmt.Sprintf("Basic %s", base64.StdEncoding.EncodeToString([]byte("oauth-client-with-plus:secret+with+plus"))))
-	return req
+	reqURLEscape, err := http.NewRequest("POST", tokenEndpoint,
+		bytes.NewBufferString("grant_type=authorization_code&"+
+			"code="+authzToken,
+		),
+	)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	reqURLEscape.Header.Set("Authorization", fmt.Sprintf("Basic %s", base64.StdEncoding.EncodeToString([]byte("oauth-client-with-plus:secret%2Bwith%2Bplus"))))
```
stlaz, line 169: use the library basic auth setter here too, please

PR #26917, opened 2022-03-17
moves TestScalingUpAndDownSingleNode from cluster-etcd-operator

p0lyn0mial:


test/extended/etcd/leader_changes.go
```diff
@@ -3,20 +3,54 @@ package etcd
...
+	return maxRevision, nil
+}
+
 var _ = g.Describe("[sig-etcd] etcd", func() {
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLIWithoutNamespace("etcd-leader-change").AsAdmin()
+
+	var earlyMaxRevision int
+	g.It("record the start revision of the etcd-operator [Early]", func() {
+		configMaps, err := oc.KubeClient().CoreV1().ConfigMaps("openshift-etcd").List(context.TODO(), metav1.ListOptions{})
```
deads2k, line 48: client-go has an operator client and you can read the current revision out of the etcd.operator.openshift.io status.

test/extended/etcd/vertical_scaling.go
```diff
@@ -0,0 +1,65 @@
...
+		machineClient := machineClientSet.MachineV1beta1().Machines("openshift-machine-api")
+
+		// assert the cluster state before we run the test
+		err = scalingtestinglibrary.EnsureInitialClusterState(ctx, g.GinkgoT(), etcdClientFactory, machineClient)
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		// step 1: add a new master node and wait until it is in Running state
+		machineName, err := scalingtestinglibrary.CreateNewMasterMachine(ctx, g.GinkgoT(), machineClient)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		err = scalingtestinglibrary.EnsureMasterMachine(ctx, g.GinkgoT(), machineName, machineClient)
```
hasbro17, line 37: Is this supposed to be `EnsureMasterMachineRunning()`? Or at least that's what I can see from the library-go PR.
https://github.com/openshift/library-go/pull/1334/files#diff-9e90817e01624b793c9df521af3e5a5b98b48e91fbe60ed1248d20f763cb8f19R57

vendor/github.com/openshift/library-go/test/library/etcdverticalscaling/helpers.go
```diff
@@ -0,0 +1,282 @@
...
+			t.Logf("failed to get the member list, will retry, err: %v", err)
+			return false, nil
+		}
+
+		var memberNames []string
+		for _, member := range memberList.Members {
+			memberNames = append(memberNames, member.Name)
+		}
+		if len(memberNames) != expectedMembersCount {
+			t.Logf("unexpected number of etcd members, expected exactly %d, got: %v, current members are: %v", expectedMembersCount, len(memberNames), memberNames)
```
hasbro17, line 186: Since https://github.com/openshift/cluster-etcd-operator/pull/758 will attempt to promote learners as soon as possible, I'm thinking we should make `EnsureMembersCount` => `EnsureVotingMembersCount()` to have a more strict check for voting members only. Otherwise if you have 3 voting + 1 learner, then `EnsureMembersCount()` still passes.  

That way step 2 above would be more accurate in reporting an error if say there is a failure to promote:
```Go
// step 2: wait until a new member shows up and check if it is healthy
		//         and until all kube-api servers have reached the same revision
		//         this additional step is the best-effort of ensuring they
		//         have observed the new member before disruption
		err = scalingtestinglibrary.EnsureVotingMembersCount(g.GinkgoT(), etcdClientFactory, 4)
		memberName, err := scalingtestinglibrary.MachineNameToEtcdMemberName(ctx, oc.KubeClient(), machineClient, machineName)
		err = scalingtestinglibrary.EnsureHealthyMember(g.GinkgoT(), etcdClientFactory, memberName)
		g.GinkgoT().Log("waiting for api servers to stabilize on the same revision")
		err = testlibraryapi.WaitForAPIServerToStabilizeOnTheSameRevision(g.GinkgoT(), oc.KubeClient().CoreV1().Pods("openshift-kube-apiserver"))
```

Failing at `EnsureVotingMembersCount()` tells us more accurately that a learner wasn't promoted than trying to figure out why `WaitForAPIServerToStabilizeOnTheSameRevision()` failed later.

vendor/github.com/openshift/library-go/test/library/etcdverticalscaling/helpers.go
```diff
@@ -0,0 +1,282 @@
...
+	defer etcdClient.Close() // ignore the errors
+
+	// since we have a direct connection with the member
+	// getting any response is a good sign of healthiness
+	ctx, cancel := context.WithTimeout(context.TODO(), 15*time.Second)
+	defer cancel()
+	_, err = etcdClient.Get(ctx, "health")
+	if err != nil {
+		return fmt.Errorf("failed to check healthiness condition of the %q member, err: %v", memberName, err)
+	}
```
hasbro17, line 237: This won't work if the request gets sent to a learner member which doesn't support linearized read/Get.
We have to use a serialized request for that case.
See https://github.com/openshift/cluster-etcd-operator/pull/758/files#diff-369b1a2393ad554f2b3caef6751bf1c596b99167ffacc57a3e95d2bf3e9f7f68R164-R170
```Go
if member.IsLearner {
	// Serializable (local) read request for learner members
	// since learners don't support linearized reads
	_, err = etcdClient.Get(ctx, "health", clientv3.WithSerializable())
} else {
	// Linearized read request to verify health of voting members
	_, err = etcdClient.Get(ctx, "health")
}
```

vendor/github.com/openshift/library-go/test/library/etcdverticalscaling/helpers.go
```diff
@@ -0,0 +1,282 @@
...
+
+	if machineToClone == nil {
+		return "", fmt.Errorf("unable to find a running master machine to clone")
+	}
+	// assigning a new Name and clearing ProviderID is enough
+	// for MAO to pick it up and provision a new master machine/node
+	machineToClone.Name = fmt.Sprintf("%s-clone", machineToClone.Name)
+	machineToClone.Spec.ProviderID = nil
+	machineToClone.ResourceVersion = ""
```
hasbro17, line 51: I thought we also had to remove the machine annotation and generation but maybe that doesn't matter and gets overwritten for a new machine:
```yaml
annotations:
    machine.openshift.io/instance-state: running
  ...
  generation: 2
```

vendor/github.com/openshift/library-go/test/library/etcdverticalscaling/helpers.go
hasbro17: Also given `EnsureInitialClusterState()` uses `EnsureMembersCount()`:
```Go
// EnsureInitialClusterState makes sure the cluster state is expected, that is, has only 3 running machines and exactly 3 voting members
```
This needs to be voting members only.

hasbro17 (commented):
SGTM except for how we are currently using `EnsureMembersCount()`.

vendor/github.com/openshift/library-go/test/library/etcdverticalscaling/helpers.go
p0lyn0mial: @hasbro17 I suggest we should change the test first in your PR (https://github.com/openshift/cluster-etcd-operator/pull/758) otherwise we won't be able to merge it.

p0lyn0mial:
/hold

I want @hasbro17 to review it first :)

p0lyn0mial:
/hold 

seeing sporadic disruptions, I'd like to spend more time on it.

p0lyn0mial:
the current plan is:

- step 1: update [the PR](https://github.com/openshift/origin/pull/26917) to tolerate disruption and rely on the monitor for reporting. I think this is okay since we observe disruption during the upgrade jobs too, the only difference is that during the upgrade job we don’t access the API except from the monitor.
- step 2: add a new test to the monitor so that it hits the cache and avoids contacting with etcd, that would allows us to see distinction - failed while talking to an api server
- step 3: measure LBs notification time so that we can calculate p99 and adjust
- step 4: let Abu fix client-go retry logic

PR #26918, opened 2022-03-17
Image registry redirect

deads2k:
Trying to build an e2e test to ensure that redirects directly to storage function as expected on GCP.

PR #26919, opened 2022-03-17
Image redirect 4.9

deads2k:
https://github.com/openshift/origin/pull/26918 against 4.9 which hopefully passes

/hold
just for proof


deads2k:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26919/pull-ci-openshift-origin-release-4.9-e2e-gcp/1504523745685934080

succeeded almost as expected.  `[sig-imageregistry] Image redirect should redirect an image pull on GCP [Suite:openshift/conformance/parallel] ` flaked instead of succeeding, so it failed once.  My guess is that this is related to the comment in slack from @dmage about "try to do a mirror", but I don't know how to do that in the API.

Perhaps doing the import and then doing some kind of watch until the import is complete will work?

PR #26926, opened 2022-03-23
Check token response error code

arjunrn:
Currently, the loop which tries the token requests only checks for errors in making the requests. It does not check the response status before marking the response as successful. This change moves the check for the status code to the for loop.

### DIFF
--- a/test/extended/oauth/client_secret.go
+++ b/test/extended/oauth/client_secret.go
@@ -90,8 +90,10 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 
 				g.By("querying for a token")
 
-				var reqErr []error
-				var tokenResponse *http.Response
+				var (
+					tokenResponse *http.Response
+					ok            bool
+				)
 				for _, tokenReq := range createTokenReq(oauthDiscoveryData.TokenEndpoint) {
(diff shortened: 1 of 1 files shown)


test/extended/oauth/client_secret.go
```diff
@@ -107,20 +109,19 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
 					client := &http.Client{Transport: tr}
 
 					tokenResponse, err = client.Do(tokenReq)
-					// if there was an error then append the error and continue
```
stlaz, line 110: you're mixing clauses in a sentence with a conjunctive relation, should still be `if there __was__ an error`

test/extended/oauth/client_secret.go
```diff
@@ -107,20 +109,19 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer]", func() {
...
-					if err != nil {
-						reqErr = append(reqErr, err)
+					// if there is an error or if the response was not 200 then continue
+					if err != nil || tokenResponse.StatusCode != http.StatusOK {
+						framework.Logf("%v", err)
```
stlaz, line 114: the information about an unexpected `StatusCode` is lost here, you may want to make these two separate conditions

stlaz:
/retest
looks like CI platform broke on those two verify jobs

PR #26929, opened 2022-03-23
Best matcher single node OVN

eggfoobar:
SNO disruption tests for OVN will fail due to stricter fuzzy matching on P99 historical duration data, this PR opens the fuzzy matching to allow SDN historical data to be used on OVN cluster tests.

### DIFF
--- a/pkg/synthetictests/historicaldata/next_best_guess.go
+++ b/pkg/synthetictests/historicaldata/next_best_guess.go
@@ -46,6 +46,9 @@ var nextBestGuessers = []NextBestKey{
 	combine(OnArchitecture("ppc64le"), PreviousReleaseUpgrade, MinorReleaseUpgrade),
 	combine(OnArchitecture("s390x"), PreviousReleaseUpgrade, MinorReleaseUpgrade),
 	combine(OnArchitecture("arm64"), PreviousReleaseUpgrade, MinorReleaseUpgrade),
+
+	combine(ForTopology("single"), OnSDN),
+	combine(ForTopology("single"), OnSDN, PreviousReleaseUpgrade),
 }
 
 // NextBestKey returns the next best key in the query_results.json generated from BigQuery and a bool indicating whether this guesser has an opinion.
@@ -156,6 +159,16 @@ func OnArchitecture(architecture string) func(in platformidentification.JobType)
 		ret := platformidentification.CloneJobType(in)
 		ret.Architecture = architecture
 		return ret, true
+	}
+}
+
+// ForTopology we match on exact topology
+func ForTopology(topology string) func(in platformidentification.JobType) (platformidentification.JobType, bool) {
+	return func(in platformidentification.JobType) (platformidentification.JobType, bool) {
+		if in.Topology != topology {
+			return platformidentification.JobType{}, false
+		}
+		return in, true
 	}
 }
 
--- a/pkg/synthetictests/allowedbackenddisruption/matches_test.go
+++ b/pkg/synthetictests/allowedbackenddisruption/matches_test.go
@@ -60,6 +60,38 @@ func TestGetClosestP95Value(t *testing.T) {
 			expectedDetails:  `(no exact match for historicaldata.DataKey{Name:"ingress-to-oauth-server-reused-connections", JobType:platformidentification.JobType{Release:"4.11", FromRelease:"4.11", Platform:"azure", Architecture:"amd64", Network:"sdn", Topology:"ha"}}, fell back to historicaldata.DataKey{Name:"ingress-to-oauth-server-reused-connections", JobType:platformidentification.JobType{Release:"4.11", FromRelease:"4.10", Platform:"azure", Architecture:"amd64", Network:"sdn", Topology:"ha"}})`,
 		},
 		{
+			name: "fuzzy-match-single-ovn-on-sdn",
+			args: args{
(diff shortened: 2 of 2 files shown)


pkg/synthetictests/historicaldata/types.go
```diff
@@ -120,14 +129,21 @@ func (b *bestMatcher) BestMatch(name string, jobType platformidentification.JobT
...
-	return &rawData.P99, details, err
+func (b *bestMatcher) BestMatchPercentile(p Percentile, name string, jobType platformidentification.JobType, extraMatchers ...NextBestKey) (*time.Duration, string, error) {
+	rawData, details, err := b.BestMatchDuration(name, jobType, extraMatchers...)
+	switch p {
+	case P95:
+		return &rawData.P95, details, err
+	case P99:
+		return &rawData.P99, details, err
+	default:
+		return &rawData.P99, details, fmt.Errorf("unsupported percentile of (%s)", p)
```
dgoodwin, line 145: Nit: probably would expect nil returned here for the percentile when there's an error.

test/extended/util/disruption/backend_sampler_tester.go
```diff
@@ -59,14 +60,23 @@ func (t *backendDisruptionTest) WithPostTeardown(postTearDown TearDownFunc) *bac
...
 
-	return allowedbackenddisruption.GetAllowedDisruption(backendName, *jobType)
+	extraMatchers := []historicaldata.NextBestKey{}
+	if jobType.Topology == "single" {
+		extraMatchers = append(
+			extraMatchers,
+			historicaldata.OnSDN,
+			historicaldata.Combine(historicaldata.PreviousReleaseUpgrade, historicaldata.OnSDN),
+		)
+	}
```
dgoodwin, line 77: It looks like this should be an OnTopology matcher similar to OnArchitecture (which has access to JobType), no public Combine, no extraMatchers on all function arguments. Would this work?

dgoodwin (commented):
This will definitely needs @deads2k as I am not that familiar with the code, but I'd pitch this rework first.

test/extended/util/disruption/backend_sampler_tester.go
eggfoobar: That's a good idea, I reworked this and took your suggestion in. Slight change I made was inverting the logic and wording of that `OnTopology` intention, since we want to match on `sdn` but only for single nodes. What do you think of the change?

dgoodwin: Looks much better to me, we'll need David to take a look when he's back from PTO now.

eggfoobar: That sounds great, thanks!

pkg/synthetictests/allowedbackenddisruption/matches.go
```diff
@@ -3,11 +3,18 @@ package allowedbackenddisruption
...
 // We enforce "don't get worse" for disruption by watching the aggregate data in CI over many runs.
-func GetAllowedDisruption(backendName string, jobType platformidentification.JobType) (*time.Duration, string, error) {
-	return getCurrentResults().BestMatchP99(backendName, jobType)
+func GetAllowedP99Disruption(backendName string, jobType platformidentification.JobType) (*time.Duration, string, error) {
+	return getCurrentResults().BestMatchPercentile(historicaldata.P99, backendName, jobType)
+}
+
+// GetAllowedP95Disruption uses the backend and information about the cluster to choose the best historical p95 to operate against.
+// We enforce "don't get worse" for disruption by watching the aggregate data in CI over many runs.
+func GetAllowedP95Disruption(backendName string, jobType platformidentification.JobType) (*time.Duration, string, error) {
```
deads2k, line 18: dead?

eggfoobar: Wasn't sure if we needed to expose P95 as well, going from some of the previous function names, if it's not needed I can remove it and clean things up.

deads2k: > Wasn't sure if we needed to expose P95 as well, going from some of the previous function names, if it's not needed I can remove it and clean things up.

it's making it hard for me to find the meat of this PR.  I'm not opposed to general refactoring, but probably not at the same time as a change you want that doesn't require it.

pkg/synthetictests/historicaldata/next_best_guess.go
```diff
@@ -46,6 +46,9 @@ var nextBestGuessers = []NextBestKey{
 	combine(OnArchitecture("ppc64le"), PreviousReleaseUpgrade, MinorReleaseUpgrade),
 	combine(OnArchitecture("s390x"), PreviousReleaseUpgrade, MinorReleaseUpgrade),
 	combine(OnArchitecture("arm64"), PreviousReleaseUpgrade, MinorReleaseUpgrade),
+
+	combine(ForTopology("single"), OnSDN),
```
deads2k, line 50: This change looks ok.

pkg/synthetictests/allowedbackenddisruption/matches.go
eggfoobar: That's fair, I removed the general refactoring

eggfoobar:
/rename Bug 2077935: Best matcher single node OVN

PR #26931, opened 2022-03-24
Revert "Implementation for basic egressfw e2e CI test"

dgoodwin:
Reverts openshift/origin#26865

TRT payload aggregation has uncovered that this test regresses on Azure, failing 10 out of 10 runs on two payloads, causing them to be rejected. Per agreed policy with staff engineers, we bias towards reverting to get payloads moving as quickly as possible.

TRT requests that when this is ready come back in, please revert the revert, and then layer in a commit that corrects the issue. We should verify that fix works on Azure prior to merge, we can use the /payload command to do that if there is no job configured that will do it on the repo. (which appears to be the case)

In this case I think you could add a comment with: `/payload 4.11 ci blocking`


dgoodwin:
Apologies I forgot to link the failure, here is one example: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-upgrade-from-stable-4.10-e2e-azure-upgrade/1506807144907804672

PR #26937, opened 2022-03-28
Sysctl tests: pod with sysctls not on whitelist should not start

mmirecki:


test/extended/security/sysctl.go
```diff
@@ -34,4 +34,21 @@ var _ = g.Describe("[sig-arch] [Conformance] sysctl", func() {
...
+		f := oc.KubeFramework()
+		isFailed, err := exutil.CreateExecPodWaitForStatus(f.ClientSet, f.Namespace.Name, "sysctl-pod", v1.PodFailed, func(pod *v1.Pod) {
+			pod.Spec.SecurityContext.Sysctls = []v1.Sysctl{
+				{
+					Name:  sysctl,
+					Value: value,
+				},
+			}
+		})
+		o.Expect(err).NotTo(o.HaveOccurred(), "unable to check sysctl value")
```
deads2k, line 48: If the pod shouldn't start, why would we be able to exec to it?

deads2k:
/approve

lgtm is up to someone closer to the feature

PR #26946, opened 2022-03-29
Bug 2055381: cleanup network policy ACL extended test

JacobTanenbaum:
this was not testing correct operation of network policy. it only worked because of a bug in implementation

original PR: https://github.com/openshift/origin/pull/26876

### DIFF
--- a/test/extended/networking/acl_audit_log.go
+++ b/test/extended/networking/acl_audit_log.go
@@ -109,9 +109,7 @@ func testACLLogging(f *e2e.Framework, oc *exutil.CLI, ns []string, auditOut stri
 	}
 
 	By("making \"default deny\" and \"allow from same namespace\" network policies")
-	// make two network policies, a default deny all and allow from same namespace in ns "acl-logging-on"
-	_, err = makeDenyAllPolicy(f, ns[0])
-	expectNoError(err)
+	// a make network policy that allows ingress only from the same ns "acl-logging-on"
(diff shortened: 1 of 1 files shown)


test/extended/networking/acl_audit_log.go
```diff
@@ -109,9 +109,7 @@ func testACLLogging(f *e2e.Framework, oc *exutil.CLI, ns []string, auditOut stri
 	}
 
 	By("making \"default deny\" and \"allow from same namespace\" network policies")
-	// make two network policies, a default deny all and allow from same namespace in ns "acl-logging-on"
-	_, err = makeDenyAllPolicy(f, ns[0])
-	expectNoError(err)
+	// a make network policy that allows ingress only from the same ns "acl-logging-on"
 	_, err = makeFromSameNSPolicy(f, ns[0])
```
danwinship, line 113: the `By()` message is now wrong.
(Just get rid of the comment put it all in the `By()` message; then the explanation is visible both to people reading the code and to people reading test logs)

danwinship (commented):
Not sure how this is "Clean up the namespaces" ?

The change seems fine but I'm not sure why we need/want it...

test/extended/networking/acl_audit_log.go
JacobTanenbaum: I will clean this up in a followup

PR #26952, opened 2022-03-30
Bug 2070320: IBMCloud: Move oauth External check

cjschaef:
Move the check for External topology (ROKS) for the oauth
requestheaders test, prior to any test steps being performed,
such as authentication.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2070320


cjschaef:
Failing required tests are not related to this test change, they all appear to be due to other test failures or failures during `install` steps. Attempting to retrigger.

/retest-required

PR #26954, opened 2022-03-31
BZ 2087504: Add Early test to check for CRD subresource.status

DennisPeriquet:
[TRT-158](https://issues.redhat.com/browse/TRT-158)

The purpose of this test is to enforce (at least for the ones that have it now), spec/status separation by ensuring:

* there is a "status" element in the openapiv3 spec
* there is a subresource.status defined in the CRD (which will enable the `.../status` endpoint) which can be used for the controller to write status.

The CRDs that do not pass are treated as "exceptions" for now.  We can open up a BZ for them.


test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,115 @@
...
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// check_crds returns a list of names of CRDs that don't have subresource.status
+// For now, it ignores the ones that are currently failing.
+func check_crds(crd_item_list []v1.CustomResourceDefinition) []string {
```
dgoodwin, line 20: checkCRDs is the general go format.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,115 @@
...
+		for _, item := range str_list {
+			if name == item {
+				return true
+			}
+		}
+		return false
+	}
+
+	// These 17 CRDs do not have subresource.status.
+	//
```
dgoodwin, line 32: I would elaborate here that this was at the time the test was written, we want to prevent any more from getting in.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,115 @@
...
+		"operatorpkis.network.operator.openshift.io",
+		"profiles.tuned.openshift.io",
+		"rangeallocations.security.internal.openshift.io",
+		"rolebindingrestrictions.authorization.openshift.io",
+		"securitycontextconstraints.security.openshift.io",
+		"tuneds.tuned.openshift.io",
+	}
+
+	failures := []string{}
+	for _, crd_item := range crd_item_list {
```
dgoodwin, line 54: crdItem, crdItemList typically in go. (for these and any other vars/funcs in here, see https://go.dev/doc/effective_go#mixed-caps

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,115 @@
...
+
+		// This test is interested only in CRDs that end with "openshift.io".
+		if !strings.HasSuffix(crd_item.ObjectMeta.Name, "openshift.io") {
+			continue
+		}
+
+		// Use the latest version
+		index := len(crd_item.Spec.Versions) - 1
```
dgoodwin, line 62: Is the latest version always the last in the list?

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,115 @@
...
+		g.It("should have subresource.status", func() {
+			failures := check_crds(crd_item_list)
+			if len(failures) > 0 {
+				e2e.Fail(strings.Join(failures, "\n"))
+			}
+		})
+	})
+})
+
+func get_crd_item_list(crdClient apiextensionsclientset.Clientset) ([]v1.CustomResourceDefinition, error) {
```
dgoodwin, line 108: Suggest getCRDs

test/extended/operators/check_crd_status_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+)
+
+// check_crd_subresource_status gets the list of CustomeResourceDefinitions and
+// checks if the subresource.status stanza is present.  It also does a count.
+// This is meant to be a test you can run outside of CI.
+//
```
dgoodwin, line 17: We don't want to commit this second variant correct?

It should be possible to run a specific ginko test, assuming you've oc login'd to the correct cluster under test, 

```
$ with make && ./openshift-tests run-test "[sig-arch] ClusterOperators should define valid related objects [Suite:openshift/conformance/parallel]"
```

test/extended/operators/check_crd_status.go
DennisPeriquet: Hm... I looked at the version `name` field and it can be something like `v1`, `v1beta`, etc. and so we can't tell which one is latest.  So I added a loop to check all versions.  It turns out today, there is only one version in all of the CRDs ending in `openshift.io`.

test/extended/operators/check_crd_status_test.go
DennisPeriquet: I tried the command above (but just built `openshift-tests` separately and ran it) and it works.  So I remove it.  Thanks for the tip.

test/extended/operators/check_crd_status.go
dgoodwin: Ah good, that makes sense, I recall it was a pretty huge pain to support more than one version at a time.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"strings"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	"k8s.io/kube-openapi/pkg/util/sets"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
deads2k, line 14: alias as `apiextensionsv1` for consistency

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,113 @@
...
+		if exceptionsList.Has(crdName) {
+			continue
+		}
+
+		// Iterate through all versions of the Spec.
+		var versionName string
+		for i := 0; i < len(crdItem.Spec.Versions); i++ {
+
+			versionName = crdItem.Spec.Versions[i].Name
+			if crdItem.Spec.Versions[i].Subresources == nil {
```
deads2k, line 66: if any version has a status, that will be good enough.  As opposed to requiring all versions to have a status.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,113 @@
...
+		crdName := crdItem.ObjectMeta.Name
+
+		// Skip CRDs in the exceptions list for now.
+		if exceptionsList.Has(crdName) {
+			continue
+		}
+
+		// Iterate through all versions of the Spec.
+		var versionName string
+		for i := 0; i < len(crdItem.Spec.Versions); i++ {
```
deads2k, line 63: you only need to do this check if the openapi spec has a status element.

DennisPeriquet: If `CustomResourceDefinition.Status` was a pointer, I can do `if crdItem.Status ~= nil`.

But since it's not a pointer, that syntax is illegal so I don't know how to tell if it exists or not.

dgoodwin: https://stackoverflow.com/questions/28447297/how-to-check-for-an-empty-struct should do the trick, essentially comparing the Status struct to an empty zero value'd Status struct, if equal, it's unset.

test/extended/operators/check_crd_status.go
dgoodwin: This doesn't look addressed, still looks like we fail for any version missing status.

I'd expect a bool foundVersionWithStatusSubresource, add to failures outside the loop if it never got set to true.

test/extended/operators/check_crd_status.go
DennisPeriquet: thanks @dgoodwin a straight composite literal compare wouldn't compile because of the slices and struct (with its own slice) inside `CustomResourceDefinitionStatus` so I went into the `Status` struct to check on fields that would not be empty if the `status` element was set.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,129 @@
...
+
+		crdName := crdItem.ObjectMeta.Name
+
+		// Skip CRDs in the exceptions list for now.
+		if exceptionsList.Has(crdName) {
+			continue
+		}
+
+		// Iterate through all versions of the Spec.
+		var versionName string
```
dgoodwin, line 71: No need to define this outside the loop.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,129 @@
...
+					if crdItem.Spec.Versions[i].Subresources.Status == nil {
+						failures = append(failures, fmt.Sprintf("CRD %s version %s missing subresources.status", crdName, versionName))
+						// We will pass if one of them has a subresource.status.
+						break
+					}
+				}
+			}
+		} else {
+			failures = append(failures, fmt.Sprintf("CRD %s has no Status defined", crdName))
+		}
```
dgoodwin, line 89: This still looks like it's failing if any version is missing, David's comment awhile back suggested only to fail if NO versions had a status. In this case you'd want a bool for found and failing at the end if it was still false.

DennisPeriquet: ah, fixed.

test/extended/operators/check_crd_status.go
deads2k: and these should be tidied up over time.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,125 @@
...
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// hasStatus returns true of the given crdItem has a Status defined.
+func hasStatus(crdItem apiextensionsv1.CustomResourceDefinition) bool {
```
deads2k, line 20: this looks confused.  You want to know if the schema in the CRD spec has a status element defined.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,121 @@
...
+		for i := 0; i < len(crdItem.Spec.Versions); i++ {
+			if _, ok := crdItem.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["status"]; ok {
+				if crdItem.Spec.Versions[i].Subresources != nil && crdItem.Spec.Versions[i].Subresources.Status != nil {
+					foundStatusInSchema = true
+					break
+				}
+			}
+		}
+		if !foundStatusInSchema {
+			failures = append(failures, fmt.Sprintf("CRD %s is missing subresources.status", crdName))
```
deads2k, line 80: they aren't required to have status.  Some objects don't.  But if they have status in the schema, they must have a status subresource

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,121 @@
...
+		foundStatusInSchema := false
+		for i := 0; i < len(crdItem.Spec.Versions); i++ {
+			if _, ok := crdItem.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties["status"]; ok {
+				if crdItem.Spec.Versions[i].Subresources != nil && crdItem.Spec.Versions[i].Subresources.Status != nil {
+					foundStatusInSchema = true
+					break
+				}
+			}
+		}
+		if !foundStatusInSchema {
```
deads2k, line 79: if foundStatusInSchema, they must have status enabled here as a subresource https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/types.go#L404

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,127 @@
...
+		"imagecontentpolicies.config.openshift.io",
+		"imagecontentsourcepolicies.operator.openshift.io",
+		"machineconfigs.machineconfiguration.openshift.io",
+		"netnamespaces.network.openshift.io",
+		"networks.config.openshift.io",
+		"networks.operator.openshift.io",
+		"operatorpkis.network.operator.openshift.io",
+		"profiles.tuned.openshift.io",
+		"rangeallocations.security.internal.openshift.io",
+		"rolebindingrestrictions.authorization.openshift.io",
```
deads2k, line 50: This does not have a status field.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,127 @@
...
+		"hostsubnets.network.openshift.io",
+		"imagecontentpolicies.config.openshift.io",
+		"imagecontentsourcepolicies.operator.openshift.io",
+		"machineconfigs.machineconfiguration.openshift.io",
+		"netnamespaces.network.openshift.io",
+		"networks.config.openshift.io",
+		"networks.operator.openshift.io",
+		"operatorpkis.network.operator.openshift.io",
+		"profiles.tuned.openshift.io",
+		"rangeallocations.security.internal.openshift.io",
```
deads2k, line 49: does not have a status field.

Something isn't right if you have to make exceptions for cases without status in the schema.

test/extended/operators/check_crd_status.go
```diff
@@ -0,0 +1,127 @@
...
+				break
+			}
+		}
+
+		if foundStatusInSchema {
+			if !(crdItem.Spec.Versions[i].Subresources != nil && crdItem.Spec.Versions[i].Subresources.Status != nil) {
+				failures = append(failures, fmt.Sprintf("CRD %s has a status in the schema but no subresource.status", crdName))
+			}
+		} else {
+			failures = append(failures, fmt.Sprintf("CRD %s has no 'status' element in its schema", crdName))
```
deads2k, line 86: ah, this isn't necessarily a bug.

test/extended/operators/check_crd_status.go
DennisPeriquet: correct.  `rolebindingrestrictions.authorization.openshift.io`, `rangeallocations.security.internal.openshift.io` and 16 other CRDs don't have a status field in their openapiv3 schema.

My unit test shows this:

```
CRD builds.config.openshift.io has no 'status' element in its schema
CRD clusternetworks.network.openshift.io has no 'status' element in its schema
CRD consoleclidownloads.console.openshift.io has no 'status' element in its schema
CRD consoleexternalloglinks.console.openshift.io has no 'status' element in its schema
CRD consolelinks.console.openshift.io has no 'status' element in its schema
CRD consolenotifications.console.openshift.io has no 'status' element in its schema
CRD consoleplugins.console.openshift.io has no 'status' element in its schema
CRD consolequickstarts.console.openshift.io has no 'status' element in its schema
CRD consoleyamlsamples.console.openshift.io has no 'status' element in its schema
CRD egressnetworkpolicies.network.openshift.io has no 'status' element in its schema
CRD hostsubnets.network.openshift.io has no 'status' element in its schema
CRD imagecontentpolicies.config.openshift.io has no 'status' element in its schema
CRD imagecontentsourcepolicies.operator.openshift.io has no 'status' element in its schema
CRD machineconfigs.machineconfiguration.openshift.io has no 'status' element in its schema
CRD netnamespaces.network.openshift.io has no 'status' element in its schema
CRD rangeallocations.security.internal.openshift.io has no 'status' element in its schema
CRD rolebindingrestrictions.authorization.openshift.io has no 'status' element in its schema
CRD securitycontextconstraints.security.openshift.io has no 'status' element in its schema

CRD networks.config.openshift.io has a status in the schema but no subresource.status
CRD networks.operator.openshift.io has a status in the schema but no subresource.status
CRD operatorpkis.network.operator.openshift.io has a status in the schema but no subresource.status
CRD profiles.tuned.openshift.io has a status in the schema but no subresource.status
CRD tuneds.tuned.openshift.io has a status in the schema but no subresource.status
```

and those are in the exceptions list.

test/extended/operators/check_crd_status.go
DennisPeriquet: Agreed.  But I thought the original intent of this jira/PR was to ensure that it doesn't get worse -- has that changed?

test/extended/operators/check_crd_status.go
DennisPeriquet: Agreed.  But I thought the purpose of this Jira/PR was to check for the presence of the status element in the schema.

deads2k: > Agreed. But I thought the purpose of this Jira/PR was to check for the presence of the status element in the schema.

We have two problems
1. types with status *must* have a subresource.  We've missed some in the past
2. most types should have status.

The second is a lot less severe than the first.  I suggest enforcement by separate tests and this PR just about perfect for 1 which is a lot more important.

deads2k:
we may end up relaxing that second test, but the subresource one looks good

/approve

PR #26955, opened 2022-03-31
bump: k8s.io/kubernetes

s-urbaniak:
This is necessary to pull in the newest e2e test code for upstream k8s, most notable the recent bumps in https://github.com/openshift/kubernetes/pull/1226 and https://github.com/openshift/kubernetes/pull/1212.

This is a prerequisite for https://github.com/openshift/cluster-kube-apiserver-operator/pull/1308 to pass green.

This is effectively a re-revert of https://github.com/openshift/origin/pull/26939 but it also includes https://github.com/openshift/kubernetes/pull/1228.

/cc @soltysh @stlaz @deads2k @stbenjam @danwinship @dgoodwin


dgoodwin:
/approve
/lgtm

This original issue looks ok, we're a little worried about some of the NetworkPolicy tests failing on metal ipv6, but they've been reproduced on another unrelated PR, and the job is in terrible shape to get signal out of.

We'll keep an eye out and if something goes wrong we'll revert again, but hopefully we're ok now.

PR #26957, opened 2022-03-31
NE-626: Add DNS CI coverage for golang and glibc resolver libraries

suleymanakbas91:
This PR is adding an e2e test for go and glibc resolver libraries to reduce the risk that changes to our CoreDNS configuration break DNS resolution for clients. 
- `DNS lookup using Go's DNS resolver` test creates a pod using build and deployment configs where the pod has a simple go application which uses the `net` package to create a resolver targeting CoreDNS service clusterIP. It then calls `LookupHost` function to test DNS resolution for `www.redhat.com`. 
- `DNS lookup using glibc's DNS resolver` test creates a pod using jessie image and executes `getent -s dns ahosts www.redhat.com || echo $(date) fail` ten times to test DNS resolution for `www.redhat.com`. `getent` is a tool that comes with glibc library and can be used to test DNS resolution. It does not have any functionality for us to provide a custom DNS server, but since the `resolv.conf` file has the clusterIP of CoreDNS service in the `nameserver` field, it does a lookup against our server.

### DIFF
--- a/test/extended/dns/dns_libraries.go
+++ b/test/extended/dns/dns_libraries.go
@@ -1,0 +1,115 @@
+package dns
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	kapiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	watchtools "k8s.io/client-go/tools/watch"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+// checkForPodLogFailures goes through the pod logs and determines if there was a failure
+// by looking for "fail" keyword in the logs.
+func checkForPodLogFailures(f *e2e.Framework, pod *kapiv1.Pod) {
+	By("submitting the pod to kubernetes")
+	podClient := f.ClientSet.CoreV1().Pods(f.Namespace.Name)
+	updated, err := podClient.Create(context.Background(), pod, metav1.CreateOptions{})
+	if err != nil {
+		e2e.Failf("Failed to create %s pod: %v", pod.Name, err)
+	}
+
+	w, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Watch(context.Background(), metav1.SingleObject(metav1.ObjectMeta{Name: pod.Name, ResourceVersion: updated.ResourceVersion}))
+	if err != nil {
+		e2e.Failf("Failed to watch pods: %v", err)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), e2e.PodStartTimeout)
+	defer cancel()
+	if _, err = watchtools.UntilWithoutRetry(ctx, w, PodSucceeded); err != nil {
+		e2e.Failf("Failed: %v", err)
+	}
+
+	By("retrieving the pod logs")
+	r, err := podClient.GetLogs(pod.Name, &kapiv1.PodLogOptions{Container: "querier"}).Stream(context.Background())
+	if err != nil {
+		e2e.Failf("Failed to get pod logs %s: %v", pod.Name, err)
+	}
+
+	scan := bufio.NewScanner(r)
+	for scan.Scan() {
+		line := scan.Text()
+		if strings.Contains(line, "fail") {
+			e2e.Failf("DNS resolution failed: %s", line)
+		}
+	}
+}
+
+var _ = Describe("[sig-network-edge] DNS lookup", func() {
+	f := e2e.NewDefaultFramework("dns-libraries")
+	oc := exutil.NewCLI("dns-libraries")
+	buildFixture := exutil.FixturePath("testdata", "dns", "dns_libraries_go.yaml")
+	dnsFixture := exutil.FixturePath("testdata", "dns")
+	labels := exutil.ParseLabelsOrDie("app=dns-libraries-go")
+
+	// creates a simple Pod that is using the image under /testdata/dns, which performs a DNS query to make sure Go's DNS resolver works fine with OpenShift DNS.
+	It("using Go's DNS resolver", func() {
+		dnsService, err := f.ClientSet.CoreV1().Services("openshift-dns").Get(context.Background(), "dns-default", metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("Failed to get default DNS service: %v", err)
+		}
+		By("creating build and deployment config for dns-libraries-go")
+		err = oc.Run("create").Args("-f", buildFixture).Execute()
+		if err != nil {
+			e2e.Failf("Failed to create build configs and deployment config for dns-libraries-go: %v", err)
+		}
+		By("starting the builder image build with a directory")
+		err = oc.Run("start-build").Args("dns-libraries-go", fmt.Sprintf("--from-dir=%s", dnsFixture)).Execute()
+		if err != nil {
+			e2e.Failf("Failed to start the builder image build: %v", err)
+		}
+		By("expect the builds to complete successfully and deploy a dns-libraries-go pod")
+		pods, err := exutil.WaitForPods(oc.KubeClient().CoreV1().Pods(oc.Namespace()), labels, exutil.CheckPodIsRunning, 1, 5*time.Minute)
+		if err != nil {
+			e2e.Failf("Failed to start the dns-libraries-go pod: %v", err)
+		}
+		if len(pods) != 1 {
+			e2e.Failf("Got %d pods with labels %v, expected 1", len(pods), labels)
+		}
+
+		By("expect the go application to successfully resolve DNS")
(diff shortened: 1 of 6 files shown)


brandisher:
Image pull timeouts

/retest

brandisher:
Image build errors

/retest

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,91 @@
...
+		line := scan.Text()
+		if strings.Contains(line, "fail") {
+			e2e.Failf("DNS resolution failed")
+		}
+	}
+}
+
+var _ = Describe("[sig-network-edge] DNS lookup", func() {
+	f := e2e.NewDefaultFramework("dns-libraries")
+	host := "www.redhat.com"
```
brandisher, line 59: Add a comment here noting whether or not usage of `www.redhat.com` is significant. When someone looks at this 1 year from now we want them to be confident that it's okay to change :smile:

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		_, err = r.LookupHost(context.Background(), host)
+		if err != nil {
+			e2e.Failf("Failed to look up host: %v", err)
+		}
+	})
+
+	// creates a simple Pod that is using getent to perform DNS queries, to make sure glibc's DNS resolver works fine with OpenShift DNS.
+	It("using glibc's DNS resolver", func() {
+		cmd := repeatCommand(
```
brandisher, line 84: Similarly with this, add a comment around the number of command repeats. If 10 is significant we want to call that out or if it's not significant that should also be noted.

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,91 @@
...
+		By("deleting the pod")
+		defer GinkgoRecover()
+		podClient.Delete(context.Background(), pod.Name, *metav1.NewDeleteOptions(0))
+	}()
+	updated, err := podClient.Create(context.Background(), pod, metav1.CreateOptions{})
+	if err != nil {
+		e2e.Failf("Failed to create %s pod: %v", pod.Name, err)
+	}
+
+	w, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Watch(context.Background(), metav1.SingleObject(metav1.ObjectMeta{Name: pod.Name, ResourceVersion: updated.ResourceVersion}))
```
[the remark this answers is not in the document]
suleymanakbas91: I think it shouldn't be an issue, as this code will get hit only if the pod creation succeeds [in the previous lines](https://github.com/openshift/origin/pull/26957/files#diff-3ddd2ec2bf91d7f1f1b00e352023fb86cd9c662e63b994fd995dcc93d90a85a7R27-R30). We then also wait for pod to start running [in the next few lines](https://github.com/openshift/origin/pull/26957/files#diff-3ddd2ec2bf91d7f1f1b00e352023fb86cd9c662e63b994fd995dcc93d90a85a7R36-R40).

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,91 @@
...
+
+	. "github.com/onsi/ginkgo"
+
+	kapiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	watchtools "k8s.io/client-go/tools/watch"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+func validateResults(f *e2e.Framework, pod *kapiv1.Pod) {
```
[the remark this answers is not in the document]
suleymanakbas91: Done.

test/extended/dns/dns_libraries.go
suleymanakbas91: Done.

test/extended/dns/dns_libraries.go
suleymanakbas91: Done.

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,95 @@
...
+		if strings.Contains(line, "fail") {
+			e2e.Failf("DNS resolution failed")
+		}
+	}
+}
+
+var _ = Describe("[sig-network-edge] DNS lookup", func() {
+	f := e2e.NewDefaultFramework("dns-libraries")
+	// using www.redhat.com just as a sample host for dns queries
+	host := "www.redhat.com"
```
Miciah, line 62:
```suggestion
	const host = "www.redhat.com"
```

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,95 @@
...
+	It("using Go's DNS resolver", func() {
+		dnsService, err := f.ClientSet.CoreV1().Services("openshift-dns").Get(context.Background(), "dns-default", metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("Failed to get default DNS service: %v", err)
+		}
+		r := &net.Resolver{
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				d := net.Dialer{
+					Timeout: time.Millisecond * time.Duration(10000),
```
Miciah, line 74:
```suggestion
					Timeout: 10 * time.Second,
```

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,95 @@
...
+				}
+				return d.DialContext(ctx, network, fmt.Sprintf("%s:%s", dnsService.Spec.ClusterIP, "53"))
```
Miciah, line 76:
```suggestion
				return d.DialContext(ctx, network, net.JoinHostPort(dnsService.Spec.ClusterIP, "53"))
```

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,95 @@
...
+			},
+		}
+		_, err = r.LookupHost(context.Background(), host)
```
Miciah, line 79: Isn't this code executing on the test runner?  I think you need to spawn a pod that runs inside the cluster in order to test with the service's cluster IP address.

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,95 @@
...
+
+// checkForPodLogFailures goes through the pod logs and determines if there was a failure
+// by looking for "fail" keyword in the logs.
+func checkForPodLogFailures(f *e2e.Framework, pod *kapiv1.Pod) {
+	By("submitting the pod to kubernetes")
+	podClient := f.ClientSet.CoreV1().Pods(f.Namespace.Name)
+	defer func() {
+		By("deleting the pod")
+		defer GinkgoRecover()
+		podClient.Delete(context.Background(), pod.Name, *metav1.NewDeleteOptions(0))
```
Miciah, line 27: I think the framework handles cleanup for you.  You use `f := e2e.NewDefaultFramework("dns-libraries")` at the start of the test, which creates a new namespace for the test, and then the framework deletes the namespace (and thus everything inside the namespace) after it runs.

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,95 @@
...
+	r, err := podClient.GetLogs(pod.Name, &kapiv1.PodLogOptions{Container: "querier"}).Stream(context.Background())
+	if err != nil {
+		e2e.Failf("Failed to get pod logs %s: %v", pod.Name, err)
+	}
+
+	scan := bufio.NewScanner(r)
+	for scan.Scan() {
+		line := scan.Text()
+		if strings.Contains(line, "fail") {
+			e2e.Failf("DNS resolution failed")
```
Miciah, line 54:
```suggestion
			e2e.Failf("DNS resolution failed: %s", line)
```

test/extended/dns/dns_libraries.go
suleymanakbas91: Done.

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,115 @@
...
+			e2e.Failf("Got %d pods with labels %v, expected 1", len(pods), labels)
+		}
+
+		By("expect the go application to successfully resolve DNS")
+		pod, err := oc.KubeClient().CoreV1().Pods(oc.Namespace()).Get(context.Background(), pods[0], metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("Failed to get dns-libraries-go pod: %v", err)
+		}
+		//execute "go run /go/dns_libraries.go -cluster-ip={CLUSTER_IP}" command inside the pod
+		args := []string{pod.Name, "-c", pod.Spec.Containers[0].Name, "--", "bash", "-c", fmt.Sprintf("go run /go/dns_libraries.go -cluster-ip=%s", dnsService.Spec.ClusterIP)}
```
Miciah, line 92:
```suggestion
		args := []string{pod.Name, "-c", pod.Spec.Containers[0].Name, "--", "bash", "-c", fmt.Sprintf("go run /go/dns_libraries.go -cluster-ip=%q", dnsService.Spec.ClusterIP)}
```

brandisher:
Unrelated failures.

/retest

PR #26965, opened 2022-04-01
skip alert test for etcd leaders when etcd revisions change a lot

deads2k:


pkg/synthetictests/allowedalerts/allow_on_revision_change.go
```diff
@@ -0,0 +1,63 @@
...
+	return &etcdRevisionChangeAllowance{
+		operatorClient:   operatorClient,
+		startingRevision: biggestRevision,
+	}, nil
+}
+
+func (d *etcdRevisionChangeAllowance) FailAfter(alertName string, jobType platformidentification.JobType) time.Duration {
+	etcd, err := d.operatorClient.Etcds().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return historicaldata.DurationOrDie(6.022)
```
p0lyn0mial, line 41: not `getClosestPercentilesValues` ?

pkg/synthetictests/allowedalerts/allow_on_revision_change.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	biggestRevision := 0
```
p0lyn0mial, line 43: you could move to a function, the same code is in `NewAllowedWhenEtcdRevisionChange`

pkg/synthetictests/allowedalerts/allow_on_revision_change.go
```diff
@@ -0,0 +1,63 @@
...
+	for _, nodeStatus := range etcd.Status.NodeStatuses {
+		if int(nodeStatus.CurrentRevision) > biggestRevision {
+			biggestRevision = int(nodeStatus.CurrentRevision)
+		}
+	}
+	numberOfRevisions := biggestRevision - d.startingRevision
+
+	if numberOfRevisions > 2 {
+		return time.Duration(numberOfRevisions) * 10 * time.Minute
```
p0lyn0mial, line 52: please add a comment explaining why we are doing here and the risk

pkg/synthetictests/alerts.go
```diff
@@ -14,7 +14,7 @@ import (
 func testAlerts(events monitorapi.Intervals, restConfig *rest.Config) []*junitapi.JUnitTestCase {
 	ret := []*junitapi.JUnitTestCase{}
 
-	alertTests := allowedalerts.AllAlertTests()
+	alertTests := allowedalerts.AllAlertTests(context.TODO(), restConfig)
```
p0lyn0mial, line 17: okay, this is the spot that should allows us to merge the other pr

pkg/synthetictests/allowedalerts/all.go
```diff
@@ -1,6 +1,31 @@
...
+
+	// if we have a clientConfig,  use it.
+	if clientConfig != nil {
+		operatorClient, err := operatorv1client.NewForConfig(clientConfig)
+		if err != nil {
+			panic(err)
+		}
+		etcdAllowance, err = NewAllowedWhenEtcdRevisionChange(ctx, operatorClient)
+		if err != nil {
+			panic(err)
```
p0lyn0mial, line 25: this can take down the whole suite

pkg/synthetictests/allowedalerts/all.go
```diff
@@ -21,7 +46,7 @@ func AllAlertTests() []AlertTest {
 		newAlert("etcd", "etcdInsufficientMembers").pending().neverFail(),
 		newAlert("etcd", "etcdInsufficientMembers").firing(),
 		newAlert("etcd", "etcdHighNumberOfLeaderChanges").pending().neverFail(),
```
p0lyn0mial, line 48: what about `pending` alerts?

pkg/synthetictests/allowedalerts/allow_on_revision_change.go
deads2k: > not getClosestPercentilesValues ?

When this fails, we want to know because it shouldn't fail.  and avagadros number is easy to search for and hopefully stands out to people.

pkg/synthetictests/allowedalerts/all.go
deads2k: > this can take down the whole suite

Yes.  If we cannot get the state of our etcd operator, why continue?

pkg/synthetictests/allowedalerts/all.go
deads2k: pending never fails

pkg/synthetictests/allowedalerts/all.go
p0lyn0mial: setting up a cluster is expensive, a single network error before running a single test could ruin that. Now I see we only issue a `GET` request which might be retried internally.

pkg/synthetictests/allowedalerts/all.go
p0lyn0mial: what is the purpose of this alert then?

pkg/synthetictests/allowedalerts/allow_on_revision_change.go
p0lyn0mial: why not to change the method to return an error and consume it `failOrFlake` method? I think having a clear reason is better than some constant number.

pkg/synthetictests/allowedalerts/allow_on_revision_change.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	biggestRevision := 0
+	for _, nodeStatus := range etcd.Status.NodeStatuses {
+		if int(nodeStatus.CurrentRevision) > biggestRevision {
+			biggestRevision = int(nodeStatus.CurrentRevision)
+		}
+	}
+	numberOfRevisions := biggestRevision - d.startingRevision
+
+	if numberOfRevisions > 2 {
```
p0lyn0mial, line 51: would it make sense to come up with some dynamic function that would allow X min per revision?

p0lyn0mial:
/close

https://github.com/openshift/origin/pull/27009 merged

PR #26972, opened 2022-04-05
Treat simpletest-rc-to-be-deleted pod sandbox failure as expected

DennisPeriquet:
The test that creates pod (via a ReplicationController), also deletes them (by deleting a ReplicationController).  The pods come up, and while their networks (logical ports) are in the process of being added, the pods are deleted (as part of the test).

As a result, the network cannot come up because the pod is already gone (the message in the `ovnkube-master` pod says something like `Unable to get pod e2e-gc-6231/simpletest-rc-to-be-deleted-7kz7f for pod update, most likely it was already deleted`).

The kubelet log (in the ci-op-xy9lvk43-de73b-lnh2s-worker-a-bjsmj/journal file) will say something like: `error getting pod: pods \"simpletest-rc-to-be-deleted-7kz7f\" not found`.  The `Failed to create sandbox for pod` message follows.

The job where I looked into it is [here](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-upgrade-from-stable-4.10-e2e-gcp-ovn-upgrade/1511221283516321792) and failing on `: [sig-network] pods should successfully create sandboxes by getting pod`.

The name of the test is `[sig-api-machinery] Garbage collector should not delete dependents that have both valid owner and owner that's waiting for dependents to be deleted [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]`

PR #26978, opened 2022-04-06
fix evaluation of deleted pods for pod sandbox test

deads2k:
alternative to https://github.com/openshift/origin/pull/26972/files that correctly determines deletion timestamps.

### DIFF
--- a/pkg/synthetictests/networking.go
+++ b/pkg/synthetictests/networking.go
@@ -42,7 +42,7 @@ func testPodSandboxCreation(events monitorapi.Intervals) []*junitapi.JUnitTestCa
 	networkOperatorProgressing := operatorsProgressing.Filter(func(ev monitorapi.EventInterval) bool {
 		return ev.Locator == "clusteroperator/network" || ev.Locator == "clusteroperator/machine-config"
 	})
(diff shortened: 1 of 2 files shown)


pkg/synthetictests/networking.go
```diff
@@ -183,21 +185,24 @@ func categorizeBySubset(categorizers []testCategorizer, failures, flakes []strin
...
 	}
 	return eventsByPods
 }
 
 func getPodDeletionTime(events monitorapi.Intervals, podLocator string) *time.Time {
+	partialLocator := monitorapi.NonUniquePodLocatorFrom(podLocator)
 	for _, event := range events {
-		if event.Locator == podLocator && event.Message == "reason/Deleted" {
+		currPartialLocator := monitorapi.NonUniquePodLocatorFrom(event.Locator)
+		if currPartialLocator == partialLocator && strings.Contains(event.Message, "reason/Deleted") {
```
DennisPeriquet, line 205: heh, I see you found that trailing space on the end of `"reason/Deleted "`.

deads2k:
gcp is having trouble, but this should help with spurious errors and passed on aws.  forcing.

PR #26982, opened 2022-04-06
Update limits

deads2k:
this brings in the architecture field and allows fuzzy matches across architectures so we don't start insta-failing.


deads2k:
gcp is hosed.  worked on non-gcp. We'll need to re-get these once GCP recovers, but merging this to get the architecture and current updates.

PR #26984, opened 2022-04-07
Bug 2066457: don't fail when a query returns warnings

simonpasquier:
The Thanos API may return warnings when it sends back a partial response
(e.g. some of the requests to the Thanos stores have failed). This
happens in particular when the cluster is upgraded and Prometheus pods
are rolled out because the sidecar might be up while Prometheus hasn't
finished loading the WAL.

With this commit, RunQueryAtTime() logs the warnings when they exist
instead of returning an error.

Signed-off-by: Simon Pasquier <spasquie@redhat.com>


deads2k:
gcp broken, the rest passed. This isn't platform specific and doesn't appear to collide.

PR #26987, opened 2022-04-07
Exclude more tests from disconnected environment

elfosardo:
The Feature:Router tests should really all not run in disconnected
environment.


elfosardo:
/hold
this may work even better https://github.com/openshift/origin/pull/26979

PR #26988, opened 2022-04-07
Bug 2072924: increase the threshold on the number of telemetry series

simonpasquier:
Looking at the telemetry data, 4.11 clusters send a bit more metrics on average
to telemetry than the previous versions. The threshold value needs to be
increased accordingly.

To calculate the new value, I've used the following PromQL query which averages
the cluster:telemetry_selected_series:count metric over the last 30 minutes for
all 4.11 installations and then returns the 99th percentile.

```
quantile(0.99,
  avg_over_time(
    (
      cluster:telemetry_selected_series:count
      *
      on (_id) group_left group by(_id) (cluster_version{version=~"4.11.0-0.ci.+"})
    )[30m:1m]
  )
)
```

image

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -477,7 +477,7 @@ sort_desc(
 		framework.Logf("No alerts fired during test run")
 	})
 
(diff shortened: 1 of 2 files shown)


jan--f:
/lgtm
Thanks for leaving the query in the comments!

PR #26989, opened 2022-04-07
Bug 2069201: Check resolv.conf file existence

pperiyasamy:
This fix parses the dnsRecursiveResolvConf node-config file
only if it exists on the node, otherwise ignore parsing it.

### DIFF
--- a/pkg/cmd/server/kubernetes/network/network_config.go
+++ b/pkg/cmd/server/kubernetes/network/network_config.go
@@ -3,6 +3,7 @@ package network
 import (
 	"fmt"
 	"net"
+	"os"
 	"time"
 
 	miekgdns "github.com/miekg/dns"
@@ -149,7 +150,7 @@ func New(options configapi.NodeConfig, clusterDomain string, proxyConfig *kubepr
 		for _, s := range options.DNSNameservers {
 			nameservers = append(nameservers, s)
 		}
-		if len(options.DNSRecursiveResolvConf) > 0 {
+		if len(options.DNSRecursiveResolvConf) > 0 && fileExists(options.DNSRecursiveResolvConf) {
 			c, err := miekgdns.ClientConfigFromFile(options.DNSRecursiveResolvConf)
 			if err != nil {
 				return nil, fmt.Errorf("could not start DNS, unable to read config file: %v", err)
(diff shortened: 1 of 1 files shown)


flavio-fernandes (commented):
/lgtm
Thank you, @pperiyasamy !

flavio-fernandes:
/assign @squeed 
@squeed  PTAL

pperiyasamy:
/cc @jluhrsen 

@jluhrsen Can you please help us  to find out why these two jenkins [jobs](https://github.com/openshift/origin/pull/26989#issuecomment-1099252071) always failing ?

jerpeter1:
I guess I'm not allowed to override. @deads2k can you override ci/openshift-jenkins/cmd please?

deads2k:
@DennisPeriquet since you're working on getting 3.11 working, can you advise me about overriding this job?

DennisPeriquet:
let me look at that job

DennisPeriquet:
1) your PR doesn't look like it has anything to do with the initial failures in that test.
2) one failure looks like:

```
=== BEGIN TEST CASE ===
test/cmd/migrate.sh:68: executing 'oc adm migrate image-references --include=imagestreams docker.io/mysql=my.docker.io/* --all-namespaces=false --loglevel=1 -o yaml' expecting success and text 'dockerImageReference: my.docker.io/mysql@sha256:'
FAILURE after 0.595s: test/cmd/migrate.sh:68: executing 'oc adm migrate image-references --include=imagestreams docker.io/mysql=my.docker.io/* --all-namespaces=false --loglevel=1 -o yaml' expecting success and text 'dockerImageReference: my.docker.io/mysql@sha256:': the output content test failed

status:
...
      lastTransitionTime: "2022-04-26T16:48:41Z"
      message: 'Internal error occurred: toomanyrequests: You have reached your pull
        rate limit. You may increase the limit by authenticating and upgrading: 
...
```

which suggests to me that this test has probably not passed since docker.io started throttling.
On jenkins, it seems the last successful build number is over 5000 builds ago and is no longer present on the machine.

So, I say it's safe to override this test.

flavio-fernandes:
@DennisPeriquet  @deads2k : we may need to `/override ci/openshift-jenkins/extended_clusterup ` as well?

DennisPeriquet:
/hold

To stop running those tests that will keep failing

DennisPeriquet:
I was able to get `ci/openshift-jenkins/extended_clusterup` to pass by manually copying the right `.docker/config.json` onto the VM where the commands are running.  Without that, you get the pull limit problem.

I could not figure out where to make a similar fix for `ci/openshift-jenkins/cmd`.

However, I feel like due to the limited scope of the change here (essentially, check for the existence of a file before acting on it), the second test can be overridden so this PR can land.  I would also think that given that jenkins will be shutdown in 3 months from now, we should get into the habit of leaning on QE for testing for fixes that would be more extensive.

DennisPeriquet:
Looks like there's no more need for an override.

pperiyasamy:
Thanks @DennisPeriquet !

PR #26993, opened 2022-04-07
fix(olm): remove dependency on marketplace-provided content

njhale:
Remove a test dependency on the marketplace-provided redhat-operators
catalog by introducing a new catalog test fixture. This allows the test to
pass when the marketplace capability is disabled.

Signed-off-by: Nick Hale <njohnhale@gmail.com>

See https://issues.redhat.com/browse/OLM-2542 for more details.


test/extended/olm/olm.go
```diff
@@ -240,25 +240,25 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
...
 
-		// skip test if redhat-operators is not present or disabled
-		ok, err := hasRedHatOperatorsSource(oc)
+		// Configure OperatorGroup before tests
+		og, err := oc.AsAdmin().Run("process").Args("--ignore-unknown-parameters=true", "-f", ogPath, "-p", "NAME=test-operator", fmt.Sprintf("NAMESPACE=%s", oc.Namespace())).OutputToFile("og.json")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		err = oc.AsAdmin().WithoutNamespace().Run("create").Args("-f", og).Execute()
 		o.Expect(err).NotTo(o.HaveOccurred())
-		if !ok {
-			g.Skip("redhat-operators source not found in enabled sources")
```
bparees, line 255: should we be concerned that this check was not working?  Does that mean we're not actually disabling some aspect of the redhat-operators source when marketplace is not enabled?

test/extended/testdata/olm/catalogsource.yaml
```diff
@@ -0,0 +1,17 @@
...
+  metadata:
+    name: "${NAME}"
+    namespace: "${NAMESPACE}"
+  spec:
+    sourceType: grpc
+    image: "${IMAGE}"
+parameters:
+- name: NAME
+- name: NAMESPACE
+- name: IMAGE
```
bparees, line 17: this should have triggered a requirement to update bindata.go i believe

test/extended/olm/olm.go
```diff
@@ -240,25 +240,25 @@ var _ = g.Describe("[sig-operator] an end user can use OLM", func() {
...
-		}
 
-		// configure OperatorGroup before tests
-		configFile, err := oc.AsAdmin().Run("process").Args("--ignore-unknown-parameters=true", "-f", operatorGroup, "-p", "NAME=test-operator", fmt.Sprintf("NAMESPACE=%s", oc.Namespace())).OutputToFile("config.json")
+		// Configure CatalogSource before tests
+		const image = "registry.redhat.io/redhat/redhat-operator-index:v4.10"
```
bparees, line 258: this is going to be a source of bit-rot someday.

can we identify the clusterversion and pick the version tag based on that, dynamically?

bparees: or pull the value in from elsewhere?  (i know we also have the challenge that the day we start v4.12, there may not be a v4.12 catalog published, so that could be a problem too...  maybe n-1 is the tag to use)

test/extended/testdata/olm/catalogsource.yaml
njhale: I'll run `go generate` and figure out what's going on with that.

test/extended/olm/olm.go
njhale: It was testing the OperatorHub API, which still exists as part of a separate problem that I'm currently tracking -- I'll post a PR here when there are more updates.

test/extended/olm/olm.go
bparees: yeah i guess maybe pinning to 4.10 and if someday that breaks on a v4.16 cluster, we can bump it at that point.  not ideal, but not worth holding this up over.

test/extended/testdata/olm/catalogsource.yaml
njhale: yup, you were right -- change pushed

bparees:
/hold

this change is fine, but we need to sort out why the operatorhub api is getting installed when it should not be, and right now the existing state of this test(failing) is a handy way to sort that out.

wking:
> this change is fine, but we need to sort out why the operatorhub api is getting installed when it should not be...

openshift/cluster-config-operator#245 just landed, which should be the last step in dancing the CRD over into the marketplace operator's repo.

/hold cancel

test/extended/testdata/olm/catalogsource.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: subscription-template
```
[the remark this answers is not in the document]
njhale: yup -- fixed

anik120:
closing in favor of #27302 

/close

PR #26995, opened 2022-04-08
WIP: Fix another issue with CVO upgrade ack timeout.

dgoodwin:
Previous attempt did not account for a secondary test that would fail
due to error being returned.


dgoodwin:
i'm worried this pr is stuck, opening another.

/close

PR #26998, opened 2022-04-08
Bug 2091086: cli: don't insist on metal3 CRDs being present when cluster capabilities are None

honza:
When cluster version capabilities is set to None, the baremetal platform
is disabled, and as such, the metal3 CRDs aren't present.  Here, we load
the cluster version CR, and only include the metal3 CRDs when the
capabilities is set to anything other than None.

The error in CI is:

```
: [sig-cli] oc explain should contain proper spec+status for CRDs [Suite:openshift/conformance/parallel]

fail [github.com/openshift/origin/test/extended/cli/explain.go:494]: Unexpected error:
    <*errors.errorString | 0xc001f173c0>: {
        s: "exit status 1: the server doesn't have a resource type \"baremetalhosts\"",
    }
    exit status 1: the server doesn't have a resource type "baremetalhosts"
occurred
```

### DIFF
--- a/test/extended/cli/explain.go
+++ b/test/extended/cli/explain.go
@@ -21,6 +21,7 @@ import (
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 
 	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/origin/test/extended/networking"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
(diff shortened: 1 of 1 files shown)


test/extended/cli/explain.go
```diff
@@ -480,7 +481,21 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 
 	g.It("should contain proper spec+status for CRDs", func() {
 		crdClient := apiextensionsclientset.NewForConfigOrDie(oc.AdminConfig())
-		crdTypesTest := crdTypes
+		configClient := configclient.NewForConfigOrDie(oc.AdminConfig())
+		clusterVersion, err := configClient.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
+		if err != nil {
+			panic(err)
```
wking, line 487: should this be something like `framework.Fail(err.Error())`?

test/extended/cli/explain.go
```diff
@@ -480,7 +481,21 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
...
+		}
+
+		crdTypesTest := baseCRDTypes
+		crdTypesTest = append(crdTypesTest, autoscalingTypes...)
+		crdTypesTest = append(crdTypesTest, machineTypes...)
+		crdTypesTest = append(crdTypesTest, additionalOperatorTypes...)
+
+		// Don't insist on metal3 types being present when capabilities are None
+		if clusterVersion.Spec.Capabilities.BaselineCapabilitySet != configv1.ClusterVersionCapabilitySetNone {
```
wking, line 496: Looking at `spec` is brittle.  Can we look at `status`?

```go
for _, capability := range clusterVersion.Status.Capabilities.EnabledCapabilities {
  if capability == configv1.ClusterVersionCapabilityBaremetal {
    crdTypesTest = append(crdTypesTest, metal3Types...)
  }
}
```

test/extended/cli/explain.go
```diff
@@ -480,7 +481,21 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 
 	g.It("should contain proper spec+status for CRDs", func() {
 		crdClient := apiextensionsclientset.NewForConfigOrDie(oc.AdminConfig())
-		crdTypesTest := crdTypes
```
wking, line 483: Don't we want to update `crdTypes`, wherever we set that up?  If it has any other consumers, I expect they'll want this same handling.

bparees:
/approve

this is a valid change/thing we need to do.  but @wking's comments should be addressed prior to lgtm

wking (approved):
/lgtm

PR #26999, opened 2022-04-08
EgressIP downstream tests for OVN Kubernetes

andreaskaris:
Implementation of the EgressIP downstream tests for OVN Kubernetes.
Given that these tests manipulate node and hostSubnet annotations, they must run without competition 
as part of the serial conformance suite.
Test behavior can be changed using several env variables:
EGRESSIP_TARGET_PROTOCOL, EGRESSIP_TARGET_HOST, EGRESSIP_TARGET_PORT

This is a first take on the tests and only tests IPv4 for OVNKubernetes and OpenShiftSDN on Azure, GCP and AWS. 

**Note:** The `vendor-update` label here is a legacy label that the bot added for an earlier version of this PR. The newest version does not modify the vendor folder, yet the bot still reports this.

**Note:** These tests only run on AWS, Azure and GCP and only run in serial. They skip anything baremetal and thus failing baremetal tests would not be due to these tests.
See: https://github.com/openshift/origin/pull/26999/commits/c69664a6d037b433a22036bdce9d41b5829d28bc#diff-dfcd246ff964649f05970d223aea24d306b7c748c8c9c094455fd01c34411303R109

### DIFF
--- a/test/extended/networking/egressip.go
+++ b/test/extended/networking/egressip.go
@@ -1,0 +1,837 @@
+package networking
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"time"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/framework/skipper"
+)
+
+const (
+	// for all tests
+	namespacePrefix = "egressip"
+	egressIPYaml    = "egressip.yaml"
+	probePodName    = "prober-pod"
+
+	// for tests against host networked pods
+	egressIPTargetHostPortMin = 32667
+	egressIPTargetHostPortMax = 32767
+
+	// Max time that we wait for changes to EgressIP objects
+	// to propagate to the CloudPrivateIPConfig objects.
+	// This can take a significant amount of time on Azure.
+	// BZ https://bugzilla.redhat.com/show_bug.cgi?id=2073045
+	egressUpdateTimeout = 180
+)
+
+var _ = g.Describe("[sig-network][Feature:EgressIP]", func() {
+	oc := exutil.NewCLI(namespacePrefix)
+	portAllocator := NewPortAllocator(egressIPTargetHostPortMin, egressIPTargetHostPortMax)
+
+	var (
+		networkPlugin string
+
+		clientset      kubernetes.Interface
+		tmpDirEgressIP string
+
+		workerNodesOrdered        []corev1.Node
+		workerNodesOrderedNames   []string
+		egressIPNodesOrderedNames []string
+		nonEgressIPNodeName       string
+
+		egressIPNamespace      string
+		externalNamespace      string
+		packetSnifferDaemonSet *v1.DaemonSet
+
+		ingressDomain string
+
+		cloudType configv1.PlatformType
+		hasIPv4   bool
+		hasIPv6   bool
+
+		targetProtocol string
+		targetHost     string
+		targetPort     int
+	)
+
+	// BeforeEach for OVN Kubernetes
+	g.BeforeEach(func() {
+		g.By("Verifying that this cluster uses a network plugin that is supported for this test")
+		networkPlugin = networkPluginName()
+		if networkPlugin != OVNKubernetesPluginName &&
+			networkPlugin != openshiftSDNPluginName {
+			skipper.Skipf("This cluster neither uses OVNKubernetes nor OpenShiftSDN")
+		}
+
+		g.By("Creating a temp directory")
+		var err error
+		tmpDirEgressIP, err = ioutil.TempDir("", "egressip-e2e")
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Getting the clientset")
+		f := oc.KubeFramework()
+		clientset = f.ClientSet
+
+		g.By("Determining the cloud infrastructure type")
+		infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		cloudType = infra.Spec.PlatformSpec.Type
+
+		g.By("Verifying that this is a supported cloud infrastructure platform")
+		isSupportedPlatform := false
+		supportedPlatforms := []configv1.PlatformType{
+			configv1.AWSPlatformType,
+			configv1.GCPPlatformType,
+			configv1.AzurePlatformType,
+		}
+		for _, supportedPlatform := range supportedPlatforms {
+			if cloudType == supportedPlatform {
+				isSupportedPlatform = true
+				break
+			}
+		}
+		if !isSupportedPlatform {
+			skipper.Skipf("This cloud platform (%s) is not supported for this test", cloudType)
+		}
+
+		// A supported version of OpenShift must hold the CloudPrivateIPConfig CRD.
+		// Otherwise, skip this test.
+		g.By("Verifying that this is a supported version of OpenShift")
+		isSupportedOcpVersion, err := exutil.DoesApiResourceExist(oc, "cloudprivateipconfigs")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if !isSupportedOcpVersion {
+			skipper.Skipf("This OCP version is not supported for this test (api-resource cloudprivateipconfigs not found)")
+		}
+
+		g.By("Getting all worker nodes in alphabetical order")
+		// Get all worker nodes, order them alphabetically with stable
+		// sort order.
+		workerNodesOrdered, err = getWorkerNodesOrdered(clientset)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		for _, s := range workerNodesOrdered {
+			workerNodesOrderedNames = append(workerNodesOrderedNames, s.Name)
+		}
+		if len(workerNodesOrdered) < 3 {
+			skipper.Skipf("This test requires a minimum of 3 worker nodes. However, this environment has %d worker nodes.", len(workerNodesOrdered))
+		}
+
+		g.By("Determining the cloud address families")
+		hasIPv4, hasIPv6, err = GetIPAddressFamily(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Determining the target protocol, host and port")
+		targetProtocol, targetHost, targetPort, err = getTargetProtocolHostPort(oc, hasIPv4, hasIPv6, cloudType, networkPlugin)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		framework.Logf("Testing against: CloudType: %s, Protocol %s, TargetHost: %s, TargetPort: %d",
+			cloudType,
+			targetProtocol,
+			targetHost,
+			targetPort)
+
+		g.By("Creating a project for the prober pod")
+		// Create a target project and assign source and target namespace
+		// to variables for later use.
+		// SetupProject() instead of SetupNamespace() because we need a privileged
+		// namespace: https://kubernetes.io/blog/2021/12/09/pod-security-admission-beta/
+		// and SetupProject() does that for us.
+		egressIPNamespace = f.Namespace.Name
+		externalNamespace = oc.SetupProject()
+
+		g.By("Selecting the EgressIP nodes and a non-EgressIP node")
+		nonEgressIPNodeName = workerNodesOrderedNames[0]
+		egressIPNodesOrderedNames = workerNodesOrderedNames[1:]
+
+		g.By("Setting the ingressdomain")
+		ingressDomain, err = getIngressDomain(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		if networkPluginName() == OVNKubernetesPluginName {
+			g.By("Setting the EgressIP nodes as EgressIP assignable")
+			for _, node := range egressIPNodesOrderedNames {
+				_, err = oc.AsAdmin().Run("label").Args("node", node, "k8s.ovn.org/egress-assignable=").Output()
+				o.Expect(err).NotTo(o.HaveOccurred())
+			}
+		}
+	})
+
+	// Do not check for errors in g.AfterEach as the other cleanup steps will fail, otherwise.
+	g.AfterEach(func() {
+		// Print some useful output on failure.
+		if g.CurrentGinkgoTestDescription().Failed {
+			printEgressIPState(oc)
+		}
+		if networkPluginName() == OVNKubernetesPluginName {
+			g.By("Deleting the EgressIP object if it exists for OVN Kubernetes")
+			egressIPYamlPath := tmpDirEgressIP + "/" + egressIPYaml
+			if _, err := os.Stat(egressIPYamlPath); err == nil {
+				_, _ = oc.AsAdmin().Run("delete").Args("-f", tmpDirEgressIP+"/"+egressIPYaml).Output()
+			}
+
+			g.By("Removing the EgressIP assignable annotation for OVN Kubernetes")
+			for _, nodeName := range egressIPNodesOrderedNames {
+				_, _ = oc.AsAdmin().Run("label").Args("node", nodeName, "k8s.ovn.org/egress-assignable-").Output()
+			}
+		} else {
+			g.By("Removing any hostsubnet EgressIPs for OpenShiftSDN")
+			for _, nodeName := range egressIPNodesOrderedNames {
+				_ = sdnHostsubnetFlushEgressIPs(oc, nodeName)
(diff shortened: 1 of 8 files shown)


andreaskaris:
Azure:
===============================================

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should have the assigned EgressIPs and EgressIPs can be deleted and recreated [Disabled:Unsupported]"
Result:  The test works (see notes), but Azure has issues with cloudprivateipconfigs deletion https://bugzilla.redhat.com/show_bug.cgi?id=2073045

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should keep the assigned EgressIPs when being rescheduled to another node [Disabled:Unsupported]"
Result: PASS

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should have the assigned EgressIPs and EgressIP objects can be updated [Disabled:Unsupported]"
Result: PASS

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets only pods matched by the pod selector should have the EgressIPs [Disabled:Unsupported]"
Result: PASS

GCP:
===============================================

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should have the assigned EgressIPs and EgressIPs can be deleted and recreated [Disabled:Unsupported]"
Result: PASS

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should keep the assigned EgressIPs when being rescheduled to another node [Disabled:Unsupported]"
Result: PASS

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should have the assigned EgressIPs and EgressIP objects can be updated [Disabled:Unsupported]"
Result:  PASS

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets only pods matched by the pod selector should have the EgressIPs [Disabled:Unsupported]"
Result: PASS

AWS (tests should be TCP to own route):
===============================================

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should have the assigned EgressIPs and EgressIPs can be deleted and recreated [Disabled:Unsupported]"
Result: PASS

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should keep the assigned EgressIPs when being rescheduled to another node [Disabled:Unsupported]"
Result: PASS 

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should have the assigned EgressIPs and EgressIP objects can be updated [Disabled:Unsupported]"
Result: PASS

Test: ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets only pods matched by the pod selector should have the EgressIPs [Disabled:Unsupported]"
Result: PASS

andreaskaris:
CC @jluhrsen @flavio-fernandes @kyrtapz

test/extended/networking/util.go
```diff
@@ -417,3 +419,120 @@ func InOVNKubernetesContext(body func()) {
 		},
 	)
 }
+
+// SubnetIPs enumerates all subnets in an ip net (starting with the provided IP address).
+func SubnetIPs(ipnet net.IPNet) ([]net.IP, error) {
```
jluhrsen, line 424: code, to me, seems fine. first thing I wondered was if these functions aren't already available somewhere/somehow
in some other utility or library. but, assuming you already checked and nothing obvious stood out.

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+	oc := exutil.NewCLI(namespacePrefix)
+	portAllocator := NewPortAllocator(egressIPTargetHostPortMin, egressIPTargetHostPortMax)
+
+	var (
+		clientset      kubernetes.Interface
+		tmpDirEgressIP string
+
+		workerNodesOrdered    []corev1.Node
+		workerNodesOrderedStr []string
+		// egressIPNodesOrdered    []corev1.Node
```
jluhrsen, line 50: any reason to keep these commented lines? some below too

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+			cloudType = infra.Spec.PlatformSpec.Type
+
+			g.By("Determining the cloud address families")
+			hasIPv4, hasIPv6, err = GetIPAddressFamily(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("Determining the target protocol, host and port")
+			targetProtocol, targetHost, targetPort, err = getTargetProtocolHostPort(oc, hasIPv4, hasIPv6, cloudType)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			framework.Logf("Testing against: Protocol %s, TargetHost: %s, TargetPort: %d", targetProtocol, targetHost, targetPort)
```
jluhrsen, line 93: add cloudType in this Logf?

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+			egressIPNamespace = f.Namespace.Name
+			externalNamespace = oc.SetupNamespace()
+
+			g.By("Getting all worker nodes in alphabetical order")
+			// Get all worker nodes, order them alphabetically with stable
+			// sort order.
+			workerNodesOrdered, err = getWorkerNodesOrdered(clientset)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			for _, s := range workerNodesOrdered {
+				workerNodesOrderedStr = append(workerNodesOrderedStr, s.Name)
```
jluhrsen, line 107: workerNodesOrderedNames instead of ...Str?

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+			}
+
+			g.By("Selecting the EgressIP nodes and a non-EgressIP node")
+			// nonEgressIPNode = workerNodesOrdered[0]
```
jluhrsen, line 111: rm comments

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+				_, err := oc.AsAdmin().Run("adm").Args("policy", "add-scc-to-user", "hostnetwork", fmt.Sprintf("system:serviceaccount:%s:default", externalNamespace)).Output()
+				o.Expect(err).NotTo(o.HaveOccurred())
+			})
+
+			g.It("EgressIP pods should query hostNetwork pods with the local node's SNAT", func() {
+				var targetIP string
+				var targetPort int
+
+				g.By("Selecting a single EgressIP node, and one node per source deployment")
+				// requires a total of 3 worker nodes
```
jluhrsen, line 157: are you only checking > 1, or do you want to explicitly check for 3

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+					iterations := 10
+					clientIPSet, err := probeForClientIPs(oc, externalNamespace, probePodName, routeName, targetIP, targetPort, iterations)
+					o.Expect(err).NotTo(o.HaveOccurred())
+
+					g.By("Making sure that EgressIPs were not part of the response")
+					framework.Logf("egressIPSet is: %v", egressIPSet)
+					framework.Logf("clientIPSet is: %v", clientIPSet)
+					o.Expect(len(clientIPSet)).Should(o.BeNumerically(">", 0))
+					// Note: it's a bug if we see an egressIP here:
+					// https://bugzilla.redhat.com/show_bug.cgi?id=2070929
```
jluhrsen, line 270: this bug is still marked as assigned (not fixed), so does this fail every time? I saw in your comments that most of these tests are PASS.

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+					egressIPYamlTemplateNamespaceSelector, // template yaml
+					egressIPObjectName,                    // name of EgressIP
+					egressIPsString,                       // compact yaml of egressIPs
+					fmt.Sprintf("kubernetes.io/metadata.name: %s", egressIPNamespace), // namespace selector
+				)
+				err = ioutil.WriteFile(tmpDirEgressIP+"/"+egressIPYaml, []byte(egressIPConfig), 0644)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				// Run this twice to make sure that repeated EgressIP creation and deletion works.
+				iterations := 5
```
jluhrsen, line 332: what is "iterations"? is it num packets sent for the packet sniffer work?

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+					egressIPYamlTemplatePodAndNamespaceSelector, // template yaml
+					egressIPNamespace,                       // name of EgressIP
+					egressIPsString,                         // compact yaml of egressIPs
+					fmt.Sprintf("app: %s", deployment0Name), // pod selector
+					fmt.Sprintf("kubernetes.io/metadata.name: %s", egressIPNamespace), // namespace selector
+				)
+				err = ioutil.WriteFile(tmpDirEgressIP+"/"+egressIPYaml, []byte(egressIPConfig), 0644)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				// Run this twice to make sure that repeated EgressIP creation and deletion works.
```
jluhrsen, line 475: don't see this happening twice. copy/paste from above?

test/extended/networking/egressip.go
```diff
@@ -0,0 +1,654 @@
...
+			if !exists {
+				framework.Logf("CloudPrivateIPConfig for %s not found.", eip)
+				return false
+			}
+		}
+		framework.Logf("CloudPrivateIPConfigs for %v found.", egressIPSet)
+		return true
+	}, time.Duration(timeout)*time.Second, 5*time.Second).Should(o.BeTrue())
+
+	// There's some additional delay in play here. If we probe right after the above check,
```
jluhrsen, line 630: don't like having a sleep at the end of a function. would it make sense to wrap whatever the "probe" is in some
Eventually() that checks once a second for 15s? or something along those lines.

test/extended/networking/egressip_helpers.go
```diff
@@ -0,0 +1,1330 @@
...
+spec:
+    egressIPs: %s
+    namespaceSelector:
+        matchLabels:
+            %s`
+)
+
+const (
+	nodeEgressIPConfigAnnotationKey = "cloud.network.openshift.io/egress-ipconfig"
+	// egressIPAgnImage                = "k8s.gcr.io/e2e-test-images/agnhost:2.36" // TODO: remove
```
jluhrsen, line 65: remove these commented lines?

test/extended/networking/egressip_helpers.go
```diff
@@ -0,0 +1,1330 @@
...
+
+	// The DaemonSet is not ready, but this is not because of a port conflict.
+	// This shouldn't happen and other parts of the code will likely report this error
+	// as a CI failure.
+	return ds, fmt.Errorf("Daemonset still not ready after %d tries", retries)
+}
+
+const (
+	tcpCaptureScript = `#!/bin/bash
+tcpdump -nne -i any -l -s 0  'port %d and tcp[((tcp[12:1] & 0xf0) >> 2):4] = 0x47455420' | awk '{print $9 " " $(NF-1)}'
```
jluhrsen, line 150: I know it would require a bigger refactor, but since it popped in to my head I figured to note it. I remember writing tests using tcpdump and using the -w <filename> option (as a background process), then doing a 'killall tcpdump' to finish it when it should be done writing. then tcpdump -r <filename> to read the packets back and parse what I needed.
That helped with some of the weirdness I would see when not actually capturing packets it should have.

BUT, looks like you have some good methods here as well.

jluhrsen:
in your earlier [comment](https://github.com/openshift/origin/pull/26999#issuecomment-1093053975) you note that most tests PASS. It is not reliable enough? just not fully understanding why we
want to have these skipped by default initially?  And if so, I didn't quite catch the plan to have these run some other
way, other than manually?

test/extended/networking/util.go
andreaskaris: I already checked and couldn't find anything (which doesn't mean that it's not there). I'll remove code duplication for DetermineImageFromRelease, though.

test/extended/networking/egressip.go
andreaskaris: Removed.

test/extended/networking/egressip.go
andreaskaris: Done

test/extended/networking/egressip.go
andreaskaris: Done

test/extended/networking/egressip.go
andreaskaris: I just need more than 1 target node (the one hosting the host networked pod) and 2 nodes for the egress ip tests --> a total of 3 worker nodes. I updated the comment in the new version, that'll hopefully be easier to understand.

test/extended/networking/egressip.go
andreaskaris: So to begin with, I will disable those internal tests because they fail until 2070929 is fixed. As soon as Surya fixes that bug, I will talk to her and Tim what to do with this here (I updated the comment for clarification):
~~~
269                                         // Note: my interpretation is that it's a bug if we see an egressIP here:
270                                         // We should never see egressIPs when querying internal targets:
271                                         // https://bugzilla.redhat.com/show_bug.cgi?id=2070929
272                                         // However, this was still a subject of discussion. When we enable these tests after
273                                         // we fix 2070929, decide if we want to see EgressIPs here or not and possibly remove
274                                         // this verification.
275                                         g.By("Making sure that EgressIPs were not part of the response")
276                                         framework.Logf("egressIPSet is: %v", egressIPSet)
277                                         framework.Logf("clientIPSet is: %v", clientIPSet)
278                                         o.Expect(len(clientIPSet)).Should(o.BeNumerically(">", 0))
279                                         o.Expect(AntiDeepEqual(egressIPSet, clientIPSet)).To(o.BeTrue())
~~~

I believe this is a bug, but when we talked about this, the others seemed to disagree. So there's a chance that I'll remove the above part, but for as long as the internal tests are disabled, I prefer to keep that part of the code and defer the final decision to a later point in time.

test/extended/networking/egressip.go
andreaskaris: I renamed this to: `numberOfRequestsToSend`

test/extended/networking/egressip.go
andreaskaris: yep, removed

test/extended/networking/egressip_helpers.go
andreaskaris: Done

test/extended/networking/egressip.go
andreaskaris: Done (in both occurrences where I used a sleep instead of a retry mechanism)

andreaskaris:
@jluhrsen I think this is ready for another review, I addressed your concerns, added 3 tests for OpenShift SDN, and tested against the aforementioned cloud platforms and network plugins. Baremetal is currently skipped on purpose as the tests require the cloudprivateipconfig CRD which is only available on cloud platforms (that also means that the tests are only ipv4).

jluhrsen:
> > in your earlier [comment](https://github.com/openshift/origin/pull/26999#issuecomment-1093053975) you note that most tests PASS. It is not reliable enough? just not fully understanding why we want to have these skipped by default initially? And if so, I didn't quite catch the plan to have these run some other way, other than manually?
> 
> So I had thought about setting the tests as disabled. Then, adding new test cases in github.com/openshift/release where the tests would be triggered individually in a bash script by their name, e.g.:
> 
> ```
> ./openshift-tests run-test "[sig-network][Feature:EgressIP] when using openshift ovn-kubernetes when testing to external targets pods should keep the assigned EgressIPs when being rescheduled to another node [Disabled:Unsupported]"
> ```
> 

I don't think this is a normal (or if it's anywhere) way of running tests in the release repo. I mean we could create
a whole new job that would do it, but that's a lot of overhead (new cluster, new job to maintain, etc).



> That test case in openshift/release could then be triggered by default or manually and could be marked as optional.
> 
> There are currently 2 problems that I see: a) The tests deal with node annotation and cannot run in parallel ; so the safest way of running them is [Serial]. That adds a lot of time to all unit tests, as each test takes between 1:30 and a bit more than 2 minutes to complete (that's on AWS, on Azure, I expect this to be slower) b) I'm a bit afraid that we enable these tests, and then we cause CI failure for other projects, and then we might have to roll back.

I think we should make a plan to get these un-excluded and running as part of the serial jobs. 6 new tests right? That
is 15-20m longer. If that's too long, maybe we can find a way to optimize and  shorten them, but for now we can
see if it's ok.

I think we can try to get this PR merged and then lay out a plan to get these running automatically in all
the serial suites. If there is some case (e.g., certain platform, or project) where these can't pass for some
reason, we can tweak those jobs to exclude specific tests.

Let's make some jira stories to figure this out so we don't lose track:
- run these tests multiple times on each of AWS, GCP, Azure to make sure they pass reliably
- figure out which projects have the serial suite running in their presubmits and see if we have any failures there to worry about.

jluhrsen:
@deads2k, do we need a bz for this, or can we override that bugzilla/valid-bug label requirement?
The work was tracked in jira [here](https://issues.redhat.com/projects/SDN/issues/SDN-1709?filter=allopenissues)

deads2k:
bypassing bug requirement to add tests

PR #27001, opened 2022-04-11
Fix another issue with CVO upgrade ack timeout on metal.

dgoodwin:
Edited to flake recently but the error was still being returned, which caused a secondary test to fail on metal.

### DIFF
--- a/test/extended/util/disruption/disruption.go
+++ b/test/extended/util/disruption/disruption.go
@@ -366,20 +366,20 @@ func RecordJUnit(f *framework.Framework, name string, fn func() (err error, flak
 	var failure string
 	if err != nil {
 		failure = err.Error()
-
-		if flake {
-			// Append an additional result with empty failure to trigger a flake.
-			f.TestSummaries = append(f.TestSummaries, additionalTest{
-				Name:     name,
-				Duration: duration,
-			})
-		}
 	}
 	f.TestSummaries = append(f.TestSummaries, additionalTest{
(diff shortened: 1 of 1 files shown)


dgoodwin:
In https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27001/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1513488757196591104 I can see the test flake and NOT trigger the additional failure. Removing the hacks and WIP.

test/extended/util/disruption/disruption.go
```diff
@@ -366,20 +366,22 @@ func RecordJUnit(f *framework.Framework, name string, fn func() (err error, flak
...
-
-		if flake {
+	}
+	f.TestSummaries = append(f.TestSummaries, additionalTest{
+		Name:     name,
+		Duration: duration,
+		Failure:  failure,
+	})
+	if flake {
+		if err != nil {
```
deads2k, line 376: if we have a flake, why does it matter if err ! =nil?

dgoodwin: It would show as two passes otherwise as we'd be adding the above testsummary with failure msg of "". I figured we don't want to play around with dupe successes and whatever fun that might cause. Only wanted to add this one if there was a fail.

dgoodwin:
The aws-serial fix has merged, this should have a better chance now.

PR #27003, opened 2022-04-11
Bug 2074035: Check that imagestreams exist in OpenShift namespace

coreydaley:
Ensure that imagestreams in the OpenShift namespace are available
before executing the test

### DIFF
--- a/test/extended/builds/new_app.go
+++ b/test/extended/builds/new_app.go
@@ -34,8 +34,11 @@ var _ = g.Describe("[sig-builds][Feature:Builds] oc new-app", func() {
 		})
 
 		g.JustBeforeEach(func() {
+			g.By("waiting for imagestreams in the OpenShift namespace")
+			err := exutil.WaitForOpenShiftNamespaceImageStreams(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
 			g.By("waiting on the local namespace builder/default SAs")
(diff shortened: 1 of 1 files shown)


gabemontero:
@bparees recently made changes to steer the e2e's away from depending on the sample imagestreams.

probably makes sense to cross reference with him and the failures noted in the BZ

bparees:
this test should no longer be depending on openshift imagestreams, did i miss a spot where it is?

bparees:
/hold

(we don't want general tests depending on samples operator content, because then we can't easily test clusters with the samples operator disabled)

bparees:
the failure is:
```
 fail [github.com/openshift/origin/test/extended/builds/new_app.go:85]: Expected
    <string>: error: unable to locate any images in image streams, local docker images with name "registry.redhat.io/ubi8/nodejs-16:latest"
```
this PR wouldn't even fix that failure because the new-app invocation isn't referencing an imagestream (since my update)

presumably that failure occurred because new-app couldn't contact registry.redhat.io (though the error message isn't great, as is par for the course for new-app, because it also checks for those images in the remote registry)

bparees:
(the other possibility for why it fails is that the container running the tests doesn't have valid creds to registry.redhat.io, and thus when new-app is invoked, it can't find the image there.  but if it is passing sometimes, that doesn't seem likely to be the issue)

coreydaley:
/close
in favor of #27005

PR #27004, opened 2022-04-11
getArchitecture: Fix to work with Hypershift

alvaroaleman:
The `getArchitecture` currently gets a master node to figure out the
arch of the components running on them. Hypershift doesn't have master
nodes so it breaks there.

Check if the cluster has an external controlplane topology and omit the
filter on master nodes if it is external.


deads2k:
/approve
/assign @deepsm007 

@deepsm007 owns lgtm

PR #26986, opened 2022-04-07
Bug 2072902: [release-4.9] exclude loki-promtail from duplicated events

tjungblu:
prevent the perma failing on 4.9:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.9-e2e-aws-serial/1511875308708433920

it was introduced in https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4.9.0-0.ci/release/4.9.0-0.ci-2022-04-06-133003 but the diff is sadly too large to try to bisect this, thus ignoring along with the existing BZ.

PR #27006, opened 2022-04-12
Bug 2074445: exclude loki-promtail from duplicated events

tjungblu:
#26986


bparees:
this change has already merged in 4.9, has passed all relevant PR testing here, and the lack of the fix is blocking a number of PRs elsewhere at this point, manually merging.

dgoodwin:
Good call this was blocking a LOT. Thanks!

PR #27007, opened 2022-04-12
Add loki to the known NetworkNotReady repeating events check.

dgoodwin:
We already had two flakes for this exact problem but in different namespaces. It appears the loki variant is hitting very often in aws-serial jobs: https://search.ci.openshift.org/?search=loki.*NetworkNotReady+network+is+not+ready&maxAge=48h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

This should help clean them up quite a bit.


dgoodwin:
Dupe of https://github.com/openshift/origin/pull/27006

/close

PR #27017, opened 2022-04-14
test/.../disruption: set pod security level centrally

s-urbaniak:
Currently, all disruption test frameworks, most notable upgrade tests,
are being instanciated here. This sets their pod security level.

/cc @soltysh @stlaz


soltysh (approved):
/approve

PR #27020, opened 2022-04-14
: Separate pod network creation failure from others in sandbox creation…

xueqzhan:
… test case

related to Bug 2073452 , but does not solve it.

### DIFF
--- a/pkg/synthetictests/networking.go
+++ b/pkg/synthetictests/networking.go
@@ -33,6 +33,7 @@ func testPodSandboxCreation(events monitorapi.Intervals) []*junitapi.JUnitTestCa
 		{by: " by getting pod", substring: " error getting pod: pods"},
 		{by: " by writing child", substring: "write child: broken pipe"},
 		{by: " by ovn default network ready", substring: "have you checked that your default network is ready? still waiting for readinessindicatorfile"},
+		{by: " by adding pod to network", substring: "failed (add)"},
 		{by: " by other", substring: " "}, // always matches
 	}
 


pkg/synthetictests/networking.go
```diff
@@ -33,6 +33,7 @@ func testPodSandboxCreation(events monitorapi.Intervals) []*junitapi.JUnitTestCa
 		{by: " by getting pod", substring: " error getting pod: pods"},
 		{by: " by writing child", substring: "write child: broken pipe"},
 		{by: " by ovn default network ready", substring: "have you checked that your default network is ready? still waiting for readinessindicatorfile"},
+		{by: " by creating pod network", substring: "failed to create pod network sandbox"},
```
deads2k, line 36: Can we match on the `"failed (add)"`? is that the right string?

xueqzhan: This is the whole string:

ns/openshift-dns pod/dns-default-8qwlc node/ip-10-0-218-169.us-west-1.compute.internal - never deleted - network rollout - reason/FailedCreatePodSandBox Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_dns-default-8qwlc_openshift-dns_f7973944-55fb-4ab2-ad0d-facb3f1b418f_0(091f3eff85c7f292db501b0fdfa7b8d1a49d844ca40a45f9c3d7cb1896b652ef): error adding pod openshift-dns_dns-default-8qwlc to CNI network "multus-cni-network": plugin type="multus" name="multus-cni-network" failed (add): }

I thought "failed to create pod network sandbox" makes more sense. Do you refer "failed (add)" instead?

deads2k:
need separate tests for assessing impact.  The test change appears to function as expected.

PR #27031, opened 2022-04-19
MON-2148: Handle 401 code for prometheus api route

arajkumar:
Currently prometheus apis exposed through oauth-proxy returns 403 on
unauthorized access. However the monitoring team is working on to
migrate to kube-rbac-proxy[1] which would 401 for unauthorized access.

This commit adds logic to handle both 401 and 403 status code.

A similar solution[2] has been employed already to fix other part of the relevant test.

[1] https://github.com/openshift/cluster-monitoring-operator/pull/1631
[2] https://github.com/openshift/origin/pull/26695

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>

### DIFF
--- a/test/extended/operators/routable.go
+++ b/test/extended/operators/routable.go
@@ -60,10 +60,10 @@ var _ = g.Describe("[sig-arch] Managed cluster should", func() {
 			name   string
 			scheme string
 			path   string
-			expect int
+			expect []int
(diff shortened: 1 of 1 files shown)


simonpasquier (commented):
/lgtm

PR #27035, opened 2022-04-20
test/dns: add the bits to make the Restricted PodSecurityPolicy happy

squeed:
This is just the boilerplate we need to make complying with the Restricted PodSecurityPolicy happy.


squeed:
Note: this test is only executed in e2e-metal-ipi-ovn-dualstack, so that job should pass before this merges.

squeed:
/hold
until dualstack tests pass

squeed:
Whoops, look like I need to specify a non-root user. Hold please.

squeed:
great, the test passed
/hold cancel

PR #27038, opened 2022-04-21
[release-4.9] Bug 2076256: Remove BlueOcean annotation check

coreydaley:
The BlueOcean annotation is no longer applied by the new version of
the plugin which we are upgrading to in 4.9. This test has already
been removed from 4.10 when the plugin was upgraded.

### DIFF
--- a/test/extended/builds/pipeline_origin_bld.go
+++ b/test/extended/builds/pipeline_origin_bld.go
@@ -485,10 +485,6 @@ var _ = g.Describe("[sig-builds][Feature:Builds][sig-devex][Feature:Jenkins][Slo
 					}
 					br.AssertSuccess()
 
-					g.By("confirm all the log annotations are there")
-					_, err = jenkins.ProcessLogURLAnnotations(oc, br)
-					o.Expect(err).NotTo(o.HaveOccurred())
-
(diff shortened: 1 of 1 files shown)


gabemontero:
/lgtm

but yeah you'll need an override of the bugzilla bot from someone like Ben @coreydaley 

also, the more precise explanation wrt the tests are we quit using them for the jenkins and plugin repos in 4.10, replacing them with equivalent tests in the plugin repos themselves, but the tests were not removed.  See https://github.com/openshift/origin/tree/release-4.10/test/extended/builds

Technically the test could still be run in non jenkins repos that expose e2e-aws-jenkins ... so you may need to update the 4.10 branch as well just in case

the test was actually removed in 4.11

coreydaley:
@bparees Could you please override the e2e-aws-single-node-serial test? The failure does not seem to be related to anything pertaining to jenkins or the section of the test that I removed.

bparees:
bug state is valid, bot just can't see the bug, manually applying valid-bz label

failing test is indeed unrelated
/override ci/prow/e2e-aws-single-node-serial

coreydaley:
/assign @prietyc123 @jitendar-singh 
for cherry-pick-approved label

PR #27044, opened 2022-04-22
Remove non-functional gu workload tests.

dgoodwin:
These never pass.

Also removing the non-gu workload tests flagged as broken two years ago.


fromanirh (commented):
without `test/extended/topology_manager/resourcealign.go`, the remaining code in `test/extended/topology_manager/` is dead, so if you want to go this path you may want to remove the entire directory.

dgoodwin:
Ack, thanks @fromanirh , does a delete all and restore when working sound ok to you?

dgoodwin:
Updated to remove more dead code.

DennisPeriquet:
@fromanirh do you happen to know what "gu" means? -- for example, the name of some tests mentions "gu workload".  Thanks.

DennisPeriquet:
@fromanirh thanks for the link.  TIL a new word -- "gu" :-)

PR #27046, opened 2022-04-22
Bug 1992596: e2e/cli: Migrate set-image.sh to Go test

deejross:
Migrate test from bash and fix flaky test

/assign @soltysh
/cc @ardaguclu


test/extended/cli/set-image.go
```diff
@@ -0,0 +1,128 @@
...
+
+		out, err = oc.Run("get").Args("pod/hello-openshift", "-o", "jsonpath='{.spec.containers[0].image}'").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.ContainSubstring("nginx:1.9.1"))
+
+		g.By("setting a different, valid image on multiple resources")
+		err = wait.Poll(time.Second, 2*time.Minute, func() (bool, error) {
+			// during the set image, it's possible that the temporary deployer pod existed, but the moment we invoke the patches the pod is gone, b/c it completed its task, so we retry several times
+			err := oc.Run("set").Args("image", "pods,dc", "*=ruby:2.7-ubi8", "--all", "--source=imagestreamtag").Execute()
+			return err == nil, nil
```
ardaguclu, line 114: Should we continue retry only for the error we care(pod is not found)?, or only error can be thrown is that?

test/extended/cli/set-image.go
```diff
@@ -0,0 +1,128 @@
...
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-cli] oc set image", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		deploymentConfig = exutil.FixturePath("testdata", "cmd", "test", "cmd", "testdata", "test-deployment-config.yaml")
+		imageStream      = exutil.FixturePath("testdata", "cmd", "test", "cmd", "testdata", "image-streams", "image-streams-centos7.json")
```
ardaguclu, line 20: This is out of scope of this PR but I'd like to note that one day we'll need to switch different image-streams json file other than centos7.

ardaguclu:
To ensure this PR does not break IPv6;
/test e2e-metal-ipi-ovn-ipv6

test/extended/cli/set-image.go
deejross: yes we do

test/extended/cli/set-image.go
deejross: The only other error I can think of is a networking/cluster down issue, and if that happens the next commands will fail as well.

test/extended/cli/set-image.go
soltysh: Some of those images are not centos7 based already :wink: but I guess the filename will stay the same

test/extended/cli/set-image.go
soltysh: My original suggestion was to ignore only a specific error `test-deployment-config-1-deploy not found` from https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_kubernetes/1243/pull-ci-openshift-kubernetes-master-e2e-agnostic-cmd/1516788704302075904 since the retries might still stumble on it.

soltysh (changes requested):

soltysh:
The failure from https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27046/pull-ci-openshift-origin-master-e2e-metal-ipi-ovn-ipv6/1518519556195225600 looks related to this PR, probably worth picking a different image, one of the mirrorred ones, see `k8s.io/kubernetes/test/utils/image` for those

deejross:
This needs to be redone, so closing for now

PR #27047, opened 2022-04-22
Separate out 'Back-off restarting failed container' from pathological test

DennisPeriquet:
See [TRT-201](https://issues.redhat.com/browse/TRT-201).

Example of what we want to separate out:

```


: [sig-arch] events should not repeat pathologically

2 events happened too frequently

event happened 56 times, something is wrong: ns/openshift-cluster-csi-drivers pod/gcp-pd-csi-driver-controller-6747df4b84-bc96x node/ci-op-5nx16hrj-d3bee-4jzmr-master-1 - reason/BackOff Back-off restarting failed container
event happened 78 times, something is wrong: ns/openshift-cluster-csi-drivers pod/gcp-pd-csi-driver-controller-6747df4b84-qjf5g node/ci-op-5nx16hrj-d3bee-4jzmr-master-2 - reason/BackOff Back-off restarting failed container
```

### DIFF
--- a/pkg/synthetictests/image_pulls_test.go
+++ b/pkg/synthetictests/image_pulls_test.go
@@ -131,3 +131,64 @@ func Test_testBackoffPullingRegistryRedhatImage(t *testing.T) {
 		})
 	}
 }
+
+func Test_testBackoffStartingFailedContainer(t *testing.T) {
+	tests := []struct {
+		name    string
+		message string
+		kind    string
+	}{
+		{
+			name:    "Test pass case",
+			message: "reason/BackOff Back-off restarting failed container (5 times)",
+			kind:    "pass",
+		},
+		{
+			name:    "Test failure case",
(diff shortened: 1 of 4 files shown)


pkg/synthetictests/image_pulls_test.go
```diff
@@ -131,3 +131,64 @@ func Test_testBackoffPullingRegistryRedhatImage(t *testing.T) {
...
+
+func Test_testBackoffStartingFailedContainer(t *testing.T) {
+	tests := []struct {
+		name    string
+		message string
+		kind    string
+	}{
+		{
+			name:    "Test pass case",
+			message: "reason/BackOff Back-off restarting failed container (5 times)",
```
dgoodwin, line 143: We expect a pass because it's only 5 times? Cool.

deads2k:
/hold

waiting on BZ

PR #27048, opened 2022-04-22
[release-4.10] Bug 2077995: Best matcher single node OVN

eggfoobar:
Back porting [PR](https://github.com/openshift/origin/pull/26929), This would allow fuzzy matching for SNO and OVN and should fix the identified tests: [Filtered Tests on 4.10 TestGrid](https://testgrid.k8s.io/redhat-openshift-ocp-release-4.10-informing#periodic-ci-openshift-release-master-ci-4.10-e2e-aws-upgrade-single-node&include-filter-by-regex=remains%3F%20available(.*)(with%7Cusing%7Cfor)%20(reused%7Cnew)%20connections&exclude-non-failed-tests=)


dgoodwin:
Will need the patch manager for this one, but for what it's worth:

/lgtm

PR #27050, opened 2022-04-25
updating the maven-pipeline imagestream apiVersion

jitendar-singh:
- updating the maven-pipeline apiVersion


jitendar-singh:
@jupierce @akram these test have been failing for the last couple of days, can you please merge the PR

akram:
@gabemontero can you give an advice about this one? it is a long failing PR and logs are not understandable to me, especially given the small change that is pushed.
I tried to close and reopen thinking that maybe we were using running old jobs which are not relevant for this PR anymore and failing because of that. But, it seems to still fail.

PR #27051, opened 2022-04-25
Ignore loki NetworkNotReady repeating events.

dgoodwin:
Known issue below, but no need to special case loki which we tend to
ignore in situations like these.


bparees:
/approve

does this need backporting?

dgoodwin:
It's probably worth backporting yes.

PR #27053, opened 2022-04-25
the 'should be scheduled on different nodes' were only fixed in 4.11 and later

deads2k:
When starting from previous levels we will hit a problem.

### DIFF
--- a/test/extended/scheduling/pods.go
+++ b/test/extended/scheduling/pods.go
@@ -2,8 +2,11 @@ package scheduling
 
 import (
 	"context"
+	"strings"
+
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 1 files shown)


test/extended/scheduling/pods.go
```diff
@@ -86,6 +89,33 @@ func (p requirePodsOnDifferentNodesTest) getDeploymentPods(oc *exutil.CLI) []*co
...
+				strings.HasPrefix(releaseUpdate.Version, "4.4") ||
+				strings.HasPrefix(releaseUpdate.Version, "4.5") ||
+				strings.HasPrefix(releaseUpdate.Version, "4.6") ||
+				strings.HasPrefix(releaseUpdate.Version, "4.7") ||
+				strings.HasPrefix(releaseUpdate.Version, "4.8") ||
+				strings.HasPrefix(releaseUpdate.Version, "4.9") ||
+				strings.HasPrefix(releaseUpdate.Version, "4.10"):
+				return
+			}
+			// if we got here, then we're 4.11 or after, so we run the test.  skip the rest of the history.
```
wking, line 114: This whole thing seems like a useful sort of helper, like:

```go
func mostRecentCompletedReleaseAtLeast(ctx context.Context, oc *exutil.CLI, minVersion string) bool, error
```

or some such, so we can easily soften tests that are recent tightenings when running the suite on older releases.  Which package would we put something like that in?

test/extended/scheduling/pods.go
```diff
@@ -86,6 +89,33 @@ func (p requirePodsOnDifferentNodesTest) getDeploymentPods(oc *exutil.CLI) []*co
...
+
+	if clusterVersion, err := oc.AdminConfigClient().ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{}); err == nil && len(clusterVersion.Status.History) > 0 {
+		// most recent is first.  Check the most recently installed version, if it is 4.10 or earlier, skip this test
+		for _, releaseUpdate := range clusterVersion.Status.History {
+			if releaseUpdate.State != configv1.CompletedUpdate {
+				continue
+			}
+			switch {
+			case strings.HasPrefix(releaseUpdate.Version, "4.0") ||
+				strings.HasPrefix(releaseUpdate.Version, "4.1") ||
```
wking, line 102: 4.11 has "4.1" as a prefix.  You'll want something more robust here, like feeding the version into a SemVer parser, or at least requiring a trailing period `4.1.`.

test/extended/scheduling/pods.go
deads2k: > or some such, so we can easily soften tests that are recent tightenings when running the suite on older releases. Which package would we put something like that in?

platformidentification perhaps.

PR #27054, opened 2022-04-25
test/extended/scheduling/pods: Non-fatal different-nodes flake for 4.10 and earlier

wking:
The separation is only reliable on 4.11 and later.  Still test on earlier releases, but only report violations as non-fatal flakes for those older releases.

Like #27053, but with a new `JobType.HasCompleted` helper, more robust version comparison, and a flake instead of a silent success when run on versions where the test-cases are not expected to reliably succeed.

### DIFF
--- a/pkg/synthetictests/platformidentification/types.go
+++ b/pkg/synthetictests/platformidentification/types.go
@@ -3,6 +3,8 @@ package platformidentification
 import (
 	"context"
 	"errors"
+	"fmt"
+	"strconv"
 	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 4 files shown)


test/extended/scheduling/pods.go
```diff
@@ -96,19 +101,33 @@ func (p requirePodsOnDifferentNodesTest) run(oc *exutil.CLI) {
...
+		e2e.Fail(err.Error())
+	}
 }
 
 var _ = g.Describe("[sig-scheduling][Early]", func() {
 	defer g.GinkgoRecover()
+	ctx := context.TODO()
 
 	oc := exutil.NewCLI("scheduling-pod-check")
+	jobType, err := platformidentification.GetJobType(ctx, restConfig)
```
wking, line 127: where do I get a `restConfig` from?

wking:
[verify][1]:

```
There was no error output from the command.
[ERROR] hack/lib/cmd.sh:10: `return "${return_code}"` exited with status 1.
[ERROR] hack/verify-generated.sh exited with code 1 after 00h 00m 39s
```

Not all that helpful.  Maybe a flake?

/test verify

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27054/pull-ci-openshift-origin-master-verify/1518719445512491008

test/extended/scheduling/pods.go
wking: `exutil.GetClientConfig(exutil.KubeConfigPath())` seems to work in aee3c432364.

pkg/synthetictests/platformidentification/types.go
```diff
@@ -14,22 +16,24 @@ import (
...
 
 type JobType struct {
-	Release      string
-	FromRelease  string
-	Platform     string
-	Architecture string
-	Network      string
-	Topology     string
+	Release                    string
+	MostRecentCompletedRelease string
```
deads2k, line 20: If this isn't the actual version listed in the clusterversion, give me a comment on the format please.

pkg/synthetictests/platformidentification/types.go
```diff
@@ -155,3 +204,26 @@ func getArchitecture(clientConfig *rest.Config) (string, error) {
 
 	return "amd64", errors.New("could not determine architecture from master nodes")
 }
+
+// HasCompleted returns nil if the cluster has completed the provided
+// major.minor release.  For example, HasCompleted("4.10") returns nil
+// if the cluster has completed 4.10 or a later minor version.  If the
+// cluster has not, HasCompleted returns an error mentioning the most
+// recently completed major.minor.
+func (jt *JobType) HasCompleted(version string) error {
```
deads2k, line 213: This is `IsVersionAtLeast`

pkg/synthetictests/platformidentification/types.go
wking: none of the version strings in here at the actual version listed in ClusterVersion ;).  But sure, I can add some godocs for the parts of this structure I've wrapped my head around.

pkg/synthetictests/platformidentification/types.go
wking: In aee3c43236 -> 6a0d9ba06b, I went with `MostRecentlyCompletedVersionIsAtLeast`, because it's an `error`-returning assertion, not a `bool`-returning query.

pkg/synthetictests/platformidentification/types.go
wking: In aee3c43236 -> 6a0d9ba06b, I've added godocs for all the properties in this structure.  If we have better reasoning for throwing out the configv1 enums and inventing our own competing value sets, let me know, and I'll slot that in.

test/extended/scheduling/pods.go
```diff
@@ -96,20 +100,40 @@ func (p requirePodsOnDifferentNodesTest) run(oc *exutil.CLI) {
 	}
 
 	nodeNameMap := map[string]string{}
+	var err error
 	for _, pod := range deploymentPods {
 		if podName, ok := nodeNameMap[pod.Spec.NodeName]; ok {
-			e2e.Failf("ns/%s pod %s and pod %s are running on the same node: %s", p.namespace, pod.Name, podName, pod.Spec.NodeName)
+			err = fmt.Errorf("ns/%s pod %s and pod %s are running on the same node: %s", p.namespace, pod.Name, podName, pod.Spec.NodeName)
+			if err2 := jobType.HasCompleted("4.11"); err2 != nil {
+				result.Flakef("%v, but separation was inconsistent before 4.11, and %v", err, err2)
```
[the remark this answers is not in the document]
wking: if looking for malscheduledpod event is a better way to identify this issue, why aren't we doing it that way in `master`?  The purpose of origin is detecting bugs.  Once we find a bug, we should complain.  Maybe we decide we can't fix the bug today (e.g. different-node scheduling for etcd and Co. in 4.10 and earlier), in which case we should complain as a flake.  Or maybe we want to force folks to fix the bug, in which case we fail.  But I don't see a use-case for "not even looking to see if detectable bug is there".  And the code from #27053 isn't even a skip, it's still running the test-case and claiming success, despite not actually looking (on 4.10 and earlier) to see if there is any success.

wking:
/test verify

Now that #27060 has landed to make the output more useful.

wking:
[Hrm][1]:

    panic: unable to find test [Top Level] [sig-scheduling][Early] The HAProxy router pods should be scheduled on different nodes

Presumably because of 40ec84a38a0660.  How do I make both the actual test runs and the verify presubmit happy at the same time?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27054/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1519948390429888512#1:build-log.txt%3A107

PR #27058, opened 2022-04-26
force completes upgrade failure if operator failure

neisw:
Adding explicit signal that '[sig-cluster-lifecycle] Cluster completes upgrade' failed when preceding 'Cluster version operator acknowledges upgrade' fails before it.

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -404,6 +404,11 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 			return nil, false
 		},
 	); err != nil {
+		//before returning the err force a failure for completes upgrade which follows this test
+		disruption.RecordJUnit(f, "[sig-cluster-lifecycle] Cluster completes upgrade", func() (error, bool) {
+			framework.Logf("Cluster version operator failed to acknowledge upgrade request")
+			return fmt.Errorf("Cluster did not complete %s: %s", "upgrade", "operator failed to acknowledge upgrade request"), false
+		})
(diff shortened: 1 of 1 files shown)


test/e2e/upgrade/upgrade.go
```diff
@@ -404,6 +404,11 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
 			return nil, false
 		},
 	); err != nil {
+		//before returning the err force a failure for completes upgrade which follows this test
+		disruption.RecordJUnit(f, "[sig-cluster-lifecycle] Cluster completes upgrade", func() (error, bool) {
```
dgoodwin, line 408: Lets do a const for this string so we know we're  always matching the test name below.

test/e2e/upgrade/upgrade.go
```diff
@@ -404,6 +404,11 @@ func clusterUpgrade(f *framework.Framework, c configv1client.Interface, dc dynam
...
+			framework.Logf("Cluster version operator failed to acknowledge upgrade request")
+			return fmt.Errorf("Cluster did not complete %s: %s", "upgrade", "operator failed to acknowledge upgrade request"), false
```
dgoodwin, line 410: No need for the formatting here just a static string should do the trick.

dgoodwin (commented):
This looks like what I was expecting based on the slack convo.

test/e2e/upgrade/upgrade.go
neisw: Went back and forth as errors.New wasn't imported and calling fmt without formatting felt odd too.  Will go with fmt.Errorf with a constant string for now, but can also pull in errrors if preferred.

deads2k:
need better upgrade signal

PR #27062, opened 2022-04-27
track pod IP re-use

deads2k:
This should let us catch cases where the networking for the cluster assigns the same IP to multiple pods at the same time.

### DIFF
--- a/pkg/monitor/pod.go
+++ b/pkg/monitor/pod.go
@@ -4,13 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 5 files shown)


pkg/monitor/pod_test.go
```diff
@@ -0,0 +1,221 @@
...
+		args   args
+		want   []monitorapi.Condition
+	}{
+		{
+			name: "newly-created",
+			fields: fields{
+				podIPsToCurrentPodLocators: map[string]sets.String{},
+			},
+			args: args{
+				pod: addIP(newPod("alfa", "zulu"), "10.28.0.96"),
```
dgoodwin, line 56: Suspect your code is fine but just incase anyone does anything funky in the future, it's probably worth sprinkling some ipv6 addresses in some tests.

dgoodwin (commented):
Just one comment, but this looks great otherwise.

pkg/monitor/pod.go
```diff
@@ -619,3 +628,75 @@ func conditionsForTransitioningContainer(pod *corev1.Pod, current, previous *cor
...
+	defer p.lock.Unlock()
+
+	// only consider pod network pods because host network pods will have duplicated IPs
+	if pod.Spec.HostNetwork {
+		return conditions
+	}
+
+	podLocator := monitorapi.LocatePod(pod)
+
+	if pod.DeletionTimestamp != nil {
```
trozet, line 653: if a pod is completed you also need to remove it from your tracker. There can be duplicate IPs on pods that are in completed state.

trozet (changes requested):

pkg/monitor/pod.go
deads2k: > if a pod is completed you also need to remove it from your tracker. There can be duplicate IPs on pods that are in completed state.

Please see five lines down in this method.

deads2k: I think I found it.  The .status.Conditions[name=PodCompleted]

pkg/monitor/pod.go
```diff
@@ -619,3 +628,89 @@ func conditionsForTransitioningContainer(pod *corev1.Pod, current, previous *cor
...
+	defer p.lock.Unlock()
+
+	// only consider pod network pods because host network pods will have duplicated IPs
+	if pod.Spec.HostNetwork {
+		return conditions
+	}
+
+	podLocator := monitorapi.LocatePod(pod)
+
+	if isPodIPReleased(pod) {
```
deads2k, line 653: @trozet ptal

deads2k:
added case for handling podIP being changed.  We may later restrict if we find this to be a bug, but SDN appears to do this on node restarts.

retagging the (tested) change

deads2k:
the failures aren't related to the test added, we need the signal, merging.

PR #27065, opened 2022-04-27
[release-4.10] Bug 2080759: Remove BlueOcean annotation check

coreydaley:
The BlueOcean annotation is no longer applied by the new version of
the plugin which we are upgrading to in 4.10. This test has already
been removed from 4.11 when the plugin was upgraded.


coreydaley:
/assign @jitendar-singh @prietyc123 
for cherry pick approval

PR #27070, opened 2022-04-28
improves etcdRevisionChangeAllowance calculates the number of etcd's operator rollouts

p0lyn0mial:
it turns out that `testAlerts` is evaluated after the tests are run (https://github.com/openshift/origin/blob/master/pkg/test/ginkgo/cmd_runsuite.go#L442) thus we need to change the way we calculate the number of rollouts during the tests. 

previously it calculated the number of revisions at the end of the test runs not at the beginning as it was expected.


soltysh (approved):
/lgtm
/approve

PR #27071, opened 2022-04-28
Add restricted-v2 to the default RBAC rules for authenticated users

stlaz:
/assign @s-urbaniak


stlaz:
/test e2e-aws-serial
the modified test is not even run there

PR #27073, opened 2022-04-28
Increase determinism when searching for prom service secret token

martinkennelly:
Prior to this chance, we may pick up a service account token that
isn't the one we require. This is causing rare flakes in CI runs.

Signed-off-by: Martin Kennelly <mkennell@redhat.com>

Sample CI run where flake occurred:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-e2e-aws-ovn/1514867591028412416


jan--f:
I suppose it would either pick `prometheus-operator-token` or `prometheus-adapter-token` in those flake cases? In any case
/lgtm
Thank you, nice catch!

PR #27074, opened 2022-04-28
Include timestamp when reporting observed ReusedPodIP events.

dgoodwin:
Important to see this when we start correlating which pods got which
IPs and when.


dgoodwin:
Briefly considered if we should try to get an actual interval out of this, but from what we're seeing it looks like end time isn't particularly relevant, if it happens it seems to stick around the entire run, so the observed aka start time is probably sufficient and not worth the effort to go the other route.

PR #27080, opened 2022-04-29
haproxy needs to request privilege escalation in order to work

stlaz:
This requires access to the legacy "restricted" SCC.

/assign @Miciah 
/cc @chadcrum

### DIFF
--- a/test/extended/router/h2spec.go
+++ b/test/extended/router/h2spec.go
@@ -84,6 +84,9 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 			canaryImage, err := getCanaryImage(oc)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
+			err = oc.AsAdmin().Run("adm", "policy", "add-scc-to-user").Args("restricted", "-n", oc.Namespace(), "-z", "default").Execute()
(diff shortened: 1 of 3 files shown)


test/extended/router/h2spec.go
```diff
@@ -84,6 +84,9 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 			canaryImage, err := getCanaryImage(oc)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
+			err = oc.AsAdmin().Run("adm", "policy", "add-scc-to-user").Args("restricted", "-n", oc.Namespace(), "-z", "default").Execute()
```
deads2k, line 87: how is this cleaned up at the end of the e2e run?

deads2k: > how is this cleaned up at the end of the e2e run?

How do you make sure the kube-apiserver observes this before creating the pod.

deads2k:
this fix needs adjusting, but there's a chance it improves my current state which is full broke.

test/extended/router/h2spec.go
stlaz: > how is this cleaned up at the end of the e2e run?
it's a rolebinding, it gets removed on NS removal

> How do you make sure the kube-apiserver observes this before creating the pod.
currently, we don't

PR #27081, opened 2022-04-29
don't wait for SA token secrets

stlaz:
kube 1.24 stops autogenerating SA tokens in secrets in behalf
of the projected SA tokens.

/assign @tkashem


soltysh (approved):
/lgtm
/approve

PR #27067, opened 2022-04-27
test/extended/prometheus/image_pulls: New "should be fast" test-case

wking:
To help make things [like][1]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-upgrade-from-stable-4.10-e2e-aws-ovn-upgrade/1518821292038426624/artifacts/e2e-aws-ovn-upgrade/gather-extra/artifacts/events.json | jq -r '[.items[] | select(.reason == "Pulled" and (.message | match("[0-9]m[0-9.]+s")))] | sort_by(.metadata.creationTimestamp)[] | .metadata.creationTimestamp + " " + (.involvedObject | (.namespace + " " + .name)) + " " + .reason + ": " + .message'
2022-04-26T06:11:43Z openshift-cluster-version version--xpvbz-zldzn Pulled: Successfully pulled image "registry.build04.ci.openshift.org/ci-op-9k22xqj3/release@sha256:8b4a46a564186bc9738732c573ca317b2b9d897fd8456a7dea0ed4c87f64db83" in 10m55.944899635s
```

easier to detect.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-upgrade-from-stable-4.10-e2e-aws-ovn-upgrade/1518821292038426624

PR #27082, opened 2022-04-29
image pull speed test by wking

deads2k:
make https://github.com/openshift/origin/pull/27067 a [late] test


test/extended/images/performance.go
```diff
@@ -0,0 +1,46 @@
+package images
```
wking, line 1: This directory seems to be more about ImageStreams and such.  CRI-O pull speed (which is what these tests are about) seems like it deserves a different maintainer set.  Sibling `test/extended/node` or some such?

PR #27086, opened 2022-05-02
[rebase 1.24] disable tests with time limit to facilitate rebase

tkashem:


p0lyn0mial:
/lgtm

this is what we did in the past to facilitate rebases

deads2k:
/hold

let's get a "why" for each one please.

deads2k:
> this is what we did in the past to facilitate rebases

We're working on getting better over time.

test/extended/util/annotate/rules.go
```diff
@@ -66,6 +66,11 @@ var (
 
 			// https://bugzilla.redhat.com/show_bug.cgi?id=2004074
 			`\[sig-network-edge\]\[Feature:Idling\] Unidling should work with TCP \(while idling\)`,
+
+			// TODO: to facilitate v.14 rebase, must be removed after the rebase PR lands
+			`\[sig-auth\] ServiceAccounts should allow opting out of API token automount`,
```
deads2k, line 71: This doesn't look proper either.  do we need an early pick?

test/extended/util/annotate/rules.go
```diff
@@ -66,6 +66,11 @@ var (
...
+			`\[sig-instrumentation\] Events API should ensure that an event can be fetched, patched, deleted, and listed`,
```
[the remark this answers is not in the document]
tkashem: See bug https://bugzilla.redhat.com/show_bug.cgi?id=2081084, to make sure we don't ship without fixing this. 
I think we see this test failing because origin/test (from previous version v1.23) is running against an apiserver (v1.24 with rebase PR)
So hopefully, when re re-vendor v1.24 into openshift/origin it should resolve, otherwise we will need to fix it before we ship 4.11.0

test/extended/util/annotate/rules.go
tkashem: See bug https://bugzilla.redhat.com/show_bug.cgi?id=2081087, to make sure we don't ship without fixing this.
I think we see this test failing because origin/test (from previous version v1.23) is running against an apiserver (v1.24 with rebase PR)
So hopefully, when re re-vendor v1.24 into openshift/origin it should resolve, otherwise we will need to fix it before we ship 4.11.0

deads2k: Please locate the fix and ideally fix it here.  We have previously created blocker+ bugs, which were later deferred.  Let's do better.

test/extended/util/annotate/rules.go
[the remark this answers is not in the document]
deads2k: Please locate the fix and ideally fix it here.  We have previously created blocker+ bugs, which were later deferred.  Let's do better.

test/extended/util/annotate/rules.go
```diff
@@ -66,6 +66,15 @@ var (
...
 			`\[sig-network-edge\]\[Feature:Idling\] Unidling should work with TCP \(while idling\)`,
+
+			// TODO: to facilitate v.14 rebase, must be removed after the rebase PR lands
+			//  BZs to keep track of these issues:
+			//   - [sig-api-machinery] API data in etcd should be: https://bugzilla.redhat.com/show_bug.cgi?id=2081021
+			//   - [sig-instrumentation] Events API should ensure that: https://bugzilla.redhat.com/show_bug.cgi?id=2081084
+			//   - [sig-auth\] ServiceAccounts : https//bugzilla.redhat.com/show_bug.cgi?id=2081087
+			`\[sig-auth\] ServiceAccounts should allow opting out of API token automount`,
+			`\[sig-instrumentation\] Events API should ensure that an event can be fetched, patched, deleted, and listed`,
+			`\[sig-api-machinery\] API data in etcd should be stored at the correct location and version for all resources`,
```
deads2k, line 77: find a way to skip this for 3 three weeks and  have it auto-come back.  This being a variable, seems like a clever init block could do it.  This one I can accept will fail until the new level is merged.

test/extended/util/annotate/rules.go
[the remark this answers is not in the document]
tkashem: this relates to https://github.com/kubernetes/kubernetes/pull/108047. so the old test in openshift/origin is running against a new apiserver.

test/extended/util/annotate/rules.go
tkashem: relates to upstream pr: https://github.com/kubernetes/kubernetes/pull/108309

test/extended/util/annotate/rules.go
tkashem: and the etcd data location relates to https://github.com/kubernetes/kubernetes/pull/108445. `v1` of `csistoragecapacities` has been added to 1.24.

tkashem:
/assign @soltysh 
@deads2k @p0lyn0mial added skip until support

cmd/openshift-tests/e2e.go
```diff
@@ -14,7 +14,46 @@ import (
...
+}
+
+// shouldSkipUntil allows a test to be skipped with a time limit.
+// the test should be annotated with the 'SkippedUntil' tag, as shown below.
+//  [sig-api-machinery] testing foo [SkippedUntil:05092022] [Suite:openshift]
+// The specified date should conform to the 'MMDDYYYY' format.
+// if the specified date in the tag has not passed yet, the test
+// will be skipped by the runner.
+func shouldSkipUntil(name string) bool {
+	_, after, found := cut(name, "[SkippedUntil:")
```
soltysh, line 31: How about using something like:
```
re, err := regexp.Compile(`\[SkippedUntil:(\d{6})\]`)
s := re.FindStringSubmatch(name)
```
this will return you 2-element array, where the first is the entire `[SkippedUntil:123456]` and the 2nd will be just `123456`.

test/extended/util/annotate/rules.go
```diff
@@ -329,6 +329,18 @@ var (
...
 			`\[Feature:Platform\] Managed cluster should ensure control plane operators do not make themselves unevictable`,
 		},
+
+		// TODO: to facilitate v.14 rebase, skip the following tests until May 27 2022,
+		//  the following key should be removed after the rebase PR lands
+		//  BZs to keep track of these issues:
+		//   - [sig-api-machinery] API data in etcd should be: https://bugzilla.redhat.com/show_bug.cgi?id=2081021
+		//   - [sig-instrumentation] Events API should ensure that: https://bugzilla.redhat.com/show_bug.cgi?id=2081084
+		//   - [sig-auth] ServiceAccounts : https//bugzilla.redhat.com/show_bug.cgi?id=2081087
+		"[SkippedUntil:05272022]": {
```
p0lyn0mial, line 339: since we check it in `isDisabled` we could rename to `[DisabledUntil...]`

cmd/openshift-tests/e2e.go
deads2k: `SkippedUntil:<Date>,blocker-bz/<number>` to make the requirement clear.

test/extended/util/annotate/rules.go
tkashem: so we use both `Disabled` and `Skipped` - i don't know what the semantic differences are between these two. Also i want to avoid a `DisabledUntil:` tag get misinterpreted as plain `Disabled`
```
func isDisabled(name string) bool {
	return strings.Contains(name, "[Disabled") 
}
 ```

File: cmd/openshift-tests/e2e.go [hunk 6]
```diff
@@ -14,7 +15,41 @@ import (
...
+// the test should be annotated with the 'SkippedUntil' tag, as shown below.
+//   [SkippedUntil:05092022:blocker-bz/123456]
+// - the specified date should conform to the 'MMDDYYYY' format.
+// - a valid blocker BZ must be specified
+// if the specified date in the tag has not passed yet, the test
+// will be skipped by the runner.
+func shouldSkipUntil(name string) bool {
+	re, err := regexp.Compile(`\[SkippedUntil:(\d{8}):blocker-bz\/([a-zA-Z0-9]+)\]`)
+	if err != nil {
+		// it should only happen with a programmer error and unit
```
deads2k, line 35: you're looking for regexp.MustCompile

PR #27088, opened 2022-05-02
use a reflector to drive the pod monitor to avoid watch event compression

deads2k:
Informers are very nice for level driven things, but we want to use edge triggers for our event monitor, so use the bare reflector.  This helped with events previously.

### DIFF
--- a/pkg/monitor/monitoring_reflector_store.go
+++ b/pkg/monitor/monitoring_reflector_store.go
@@ -1,0 +1,177 @@
+package monitor
+
+import (
+	"strconv"
+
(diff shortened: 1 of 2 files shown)


deads2k:
/hold

looking at results, we should try rewriting the duplicate IP detection and then carefully inspect the pod edge events in here before merge.

deads2k:
this is looking good now

/hold cancel

dgoodwin (commented):
/lgtm

deads2k:
replaced by https://github.com/openshift/origin/pull/27114

PR #27097, opened 2022-05-04
make the pod ip check a controller

deads2k:
check carefully, this is a 10-minutes-before-carpool-controller


pkg/monitor/podipcontroller/pod_ip_controller.go
```diff
@@ -0,0 +1,203 @@
...
+			if otherPod.UID == pod.UID {
+				continue
+			}
+			if isPodIPReleased(otherPod) {
+				continue
+			}
+
+			for _, otherIP := range otherPod.Status.PodIPs {
+				otherPodIP := otherIP.IP
+				if currPodIP == otherPodIP {
```
trozet, line 168: since you are going to iterate through all the pods anyway, you might as well ignore the key of the pod you get the event for and just check for duplicates across all pods. As you iterate, just store each IP as a map key with the value as the pod namespaced name. Then during iteration if you detect a collision record that one and so on.

pkg/monitor/podipcontroller/pod_ip_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		return err
+	}
+
+	podLocator := monitorapi.LocatePod(pod)
+	podIPToPodNames := map[string]sets.String{}
+	for _, currIP := range pod.Status.PodIPs {
+		currPodIP := currIP.IP
+		podIPToPodNames[currPodIP] = sets.NewString(podLocator)
+
+		// iterate through every ip of every pod.  I wonder how badly this will scale
```
trozet, line 154: level driven ftw! I don't think it will be that bad

trozet (changes requested):

dgoodwin (commented):
Makes sense and looks good to me. 

In future it might be nice to track when we see the problem clear.

pkg/monitor/podipcontroller/pod_ip_controller.go
deads2k: > level driven ftw! I don't think it will be that bad

technically it can  have gaps. practically, I don't think we'll have an issue.

jluhrsen:
@trozet , @deads2k , using cluster-bot with this PR I ran the aws-ovn-upgrade job 30 times and there were zero
failures for the "pod IPs should not be used by two pods at the same time" test case. Rough guess looking at the
job history is that we should have picked up that failure once or twice in 30 runs if this PR didn't resolve it. 11 jobs
did fail out of the 30, but all were failures we see regularly anyway. can we get this in?

deads2k:
> can we get this in?

It's already tagged.  Just working it's way through CI.

PR #27098, opened 2022-05-04
Bug 2080681: Request new token instead of using node-bootstrapper token

ravisantoshgudimetla:
Make a new token request instead of using node-bootstrappers tokens.


ravisantoshgudimetla:
When ran locally:

```
started: (0/1/1) "[sig-cluster-lifecycle] CSRs from machines that are not recognized by the cloud provider are not approved [Suite:openshift/conformance/parallel]"
Flaky invariants:

[sig-arch] Monitor cluster while tests execute

1 pass, 0 skip (36.4s)
                     
```

soltysh (approved):
/approve

PR #27099, opened 2022-05-05
Bug 2080679: fix oc related test to land k8s 1.24

soltysh:
This should solve:
- https://bugzilla.redhat.com/show_bug.cgi?id=2080679
- https://bugzilla.redhat.com/show_bug.cgi?id=2081018

### DIFF
--- a/test/extended/testdata/cmd/test/cmd/secrets.sh
+++ b/test/extended/testdata/cmd/test/cmd/secrets.sh
@@ -150,9 +150,3 @@ os::cmd::expect_success 'oc secret unlink --help'
 
 echo "secrets: ok"
 os::test::junit::declare_suite_end
-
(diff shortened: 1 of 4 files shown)


test/extended/testdata/cmd/test/cmd/secrets.sh
```diff
@@ -150,9 +150,3 @@ os::cmd::expect_success 'oc secret unlink --help'
 
 echo "secrets: ok"
 os::test::junit::declare_suite_end
-
```
soltysh, line 153: These are being deprecated in https://github.com/openshift/oc/pull/1126 and also don't make sense after we land k8s 1.24, so it's better to get rid of them entirely.

stlaz:
/hold
/test e2e-agnostic-cmd
lets see at least some reasonable results from this test

PR #27104, opened 2022-05-05
Bug 2081012: Don't wait for sa token

ravisantoshgudimetla:
Because now there is no need to wait for the tokens to be created.


ravisantoshgudimetla:
Tests passed locally.
```
./openshift-tests run all --dry-run | grep -E "TestAutomaticCreationOfPullSecrets"| ./openshift-tests run -f -
[sig-arch] Monitor cluster while tests execute

1 pass, 0 skip (4.2s)
```

PR #27106, opened 2022-05-05
add openshift-tests timeline

deads2k:
`openshift-tests timeline --type=pod-lifecycle -f '/home/deads/Downloads/e2e-events_20220502-220022 (2).json' --known-pods='/home/deads/Downloads/resource-pods_20220502-220022.zip'  --namespace=openshift-etcd > ../timeline.html`

for example.  You need the e2e events and the resource-pods to make it work.

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -13,6 +13,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/openshift/origin/pkg/monitor/monitor_cmd"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
 	"github.com/onsi/ginkgo"
 	"github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/library-go/pkg/serviceability"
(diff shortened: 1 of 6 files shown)


DennisPeriquet:
I'm trying to understand:
https://github.com/openshift/origin/blob/29b331cbd9b8d9cb8a0bf77688a1fc9643e2ec04/pkg/monitor/intervalcreation/operator.go#L11-L21

I think you're filtering to get events with those 3 conditions.  First one makes sense -- to look for OperatorAvailable events, we look for OperatorAvailable/conditionTrue.  Second and third ones do not make sense -- to look for OperatorProgressing event, we look for OperatorProgressing/conditionFalse (shouldn't it be conditionTrue)?

deads2k:
> I think you're filtering to get events with those 3 conditions. First one makes sense -- to look for OperatorAvailable events, we look for OperatorAvailable/conditionTrue. Second and third ones do not make sense -- to look for OperatorProgressing event, we look for OperatorProgressing/conditionFalse (shouldn't it be conditionTrue)?

I don't remember changing these, but that's consistently passing the "good" condition state. So it would highlight anything except that.

PR #27109, opened 2022-05-06
add general locator part matching to timeline command

deads2k:
/assign @DennisPeriquet

### DIFF
--- a/pkg/monitor/monitorapi/types.go
+++ b/pkg/monitor/monitorapi/types.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
@@ -186,6 +188,13 @@ func IsInE2ENamespace(eventInterval EventInterval) bool {
 		return true
 	}
 	return false
+}
+
+func IsInNamespaces(namespaces sets.String) EventIntervalMatchesFunc {
+	return func(eventInterval EventInterval) bool {
+		ns := NamespaceFromLocator(eventInterval.Locator)
+		return namespaces.Has(ns)
+	}
 }
 
 func And(filters ...EventIntervalMatchesFunc) EventIntervalMatchesFunc {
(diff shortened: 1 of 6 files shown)


pkg/monitor/monitorapi/types.go
```diff
@@ -197,6 +197,29 @@ func IsInNamespaces(namespaces sets.String) EventIntervalMatchesFunc {
...
 
+// ContainsAllParts ensures that all listed key match at least one of the values.
+func ContainsAllParts(matchers map[string][]string) EventIntervalMatchesFunc {
+	return func(eventInterval EventInterval) bool {
+		actualParts := LocatorParts(eventInterval.Locator)
+		for key, possibleValues := range matchers {
+			actualValue := actualParts[key]
+
+			found := false
+			for _, possibleValue := range possibleValues {
```
DennisPeriquet, line 208: How about `strings.Contains(actualValue, possibleValue)`? this way I can filter for all pods of a certain pattern or maybe make it so we can pass a regex?

pkg/monitor/monitorapi/types.go
```diff
@@ -197,6 +197,29 @@ func IsInNamespaces(namespaces sets.String) EventIntervalMatchesFunc {
...
+				if actualValue == possibleValue {
```
DennisPeriquet, line 209: How about `strings.Contains(actualValue, possibleValue)`? this way I can filter for all pods of a certain pattern or maybe make it so we can pass a regex?

deads2k: > How about strings.Contains(actualValue, possibleValue)? this way I can filter for all pods of a certain pattern or maybe make it so we can pass a regex?

A good idea.  Could you make it work with regexes and negative matching as a followup to this?

DennisPeriquet: sure.  The last thing my script has that yours doesn't is the negative match.  I can do that on a followup PR.

DennisPeriquet: @deads2k I'm ok with this as-is and I can have a PR up shortly after this lands for the regex and negative match.

PR #27110, opened 2022-05-06
test/extended/util/openshift/clusterversionoperator/adminack: Accept MultipleReasons

wking:
When there's another reason like `TechPreviewNoUpgrade`, the CVO will be complaining about both and set `MultipleReasons`.  We want the test suite to understand that that's still an `AdminAckRequired` complaint, and not fail [like][1]:

```
  : [sig-cluster-lifecycle] TestAdminAck should succeed [Suite:openshift/conformance/parallel]
    Run #0:  3m3s

    {  fail [github.com/openshift/origin/test/extended/util/openshift/clusterversionoperator/adminack.go:117]: May  6 07:24:05.028: Error while waiting for Upgradeable to go AdminAckRequired with message "Kubernetes 1.25 and therefore OpenShift 4.12 remove several APIs which require admin consideration. Please see\nthe knowledge article https://access.redhat.com/articles/6955381 for details and instructions.\n": timed out waiting for the condition
    Upgradeable: Status=False, Reason=MultipleReasons, Message="Cluster should not be upgraded between minor versions for multiple reasons: AdminAckRequired,FeatureGates_RestrictedFeatureGates_TechPreviewNoUpgrade\n* Kubernetes 1.25 and therefore OpenShift 4.12 remove several APIs which require admin consideration. Please see\nthe knowledge article https://access.redhat.com/articles/6955381 for details and instructions.\n\n* Cluster operator kube-apiserver should not be upgraded between minor versions: FeatureGatesUpgradeable: \"TechPreviewNoUpgrade\" does not allow updates".}
```

The `Contains` logic that I'm removing is from the original test-case implementation in 4f5be5a4ac (#26525), and seems to have been assuming that we would concatenate the constituent reasons, instead of replacing all of them with `MultipleReasons`.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-e2e-azure-techpreview/1522457624627384320

### DIFF
--- a/test/extended/util/openshift/clusterversionoperator/adminack.go
+++ b/test/extended/util/openshift/clusterversionoperator/adminack.go
@@ -106,7 +106,7 @@ func (t *AdminAckTest) test(ctx context.Context, exercisedGates map[string]struc
 						"false with reason AdminAckRequired and message %q.", k, v)
 				}
 				framework.Logf("Gate %s has been ack'ed. Upgradeable is "+
-					"false but not due to this gate which would set reason AdminAckRequired with message %s. %s", k, v, getUpgradeable(ctx, t.Config))
+					"false but not due to this gate which would set reason AdminAckRequired or MultipleReasons with message %s. %s", k, v, getUpgradeable(ctx, t.Config))
 			}
 			// Clear admin ack configmap gate ack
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

PR #27112, opened 2022-05-06
Correct pod timing

deads2k:
Fix some bugs with pod timing for the timeline.

### DIFF
--- a/pkg/monitor/intervalcreation/pod.go
+++ b/pkg/monitor/intervalcreation/pod.go
@@ -15,6 +15,7 @@ func CreatePodIntervalsFromInstants(input monitorapi.Intervals, recordedResource
 	sort.Stable(ByPodLifecycle(input))
 	// these *static* locators to events. These are NOT the same as the actual event locators because nodes are not consistently assigned.
 	podToStateTransitions := map[string][]monitorapi.EventInterval{}
+	allPodTransitions := map[string][]monitorapi.EventInterval{}
 	podToContainerToLifecycleTransitions := map[string][]monitorapi.EventInterval{}
 	podToContainerToReadinessTransitions := map[string][]monitorapi.EventInterval{}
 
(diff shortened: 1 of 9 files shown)


pkg/monitor/intervalcreation/pod.go
```diff
@@ -100,22 +103,37 @@ func (t simpleTimeBounder) getEndTime(locator string) time.Time {
...
-	podEvents, ok := t.podToStateTransitions[locator]
-	if !ok {
-		return t.delegate.getStartTime(locator)
-	}
+	podEvents := t.allPodTransitions[locator]
 	for _, event := range podEvents {
-		if monitorapi.ReasonFrom(event.Message) == monitorapi.PodReasonCreated {
-			return event.From
+		podCreateEventTime = &event.From
+		break
```
DennisPeriquet, line 119: is it intended that this be a loop of one iteration?

pkg/monitor/intervalcreation/pod.go
```diff
@@ -403,13 +458,13 @@ func buildTransitionsForCategory(locatorToConditions map[string][]monitorapi.Eve
 }
 
 func sanitizeTime(nextInterval monitorapi.EventInterval, startTime, endTime time.Time) monitorapi.EventInterval {
-	if nextInterval.To.After(endTime) {
+	if !endTime.IsZero() && nextInterval.To.After(endTime) {
```
DennisPeriquet, line 461: I don't know if this line is related, but, when I run this, I'm getting the `To` set to the current date.  So, when I look at a chart from 4/29, it thinks the `To` is now and the bars go all the way to the right making the other bars look very short.

deads2k: It's because some pods aren't observed to end and those have null end dates which render long durations on the intervals chart.

PR #27113, opened 2022-05-09
Fix request token ic cli/admin tests

soltysh:
Followup to https://github.com/openshift/origin/pull/27099 
/assign @tkashem

### DIFF
--- a/test/extended/cli/admin.go
+++ b/test/extended/cli/admin.go
@@ -459,12 +459,14 @@ var _ = g.Describe("[sig-cli] oc adm", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(out).To(o.ContainSubstring("serviceaccount/my-sa-name created"))
 		o.Expect(ocns.Run("get").Args("sa", "my-sa-name").Execute()).To(o.Succeed())
-		err = wait.Poll(cliInterval, cliTimeout, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


test/extended/cli/admin.go
```diff
@@ -459,12 +459,14 @@ var _ = g.Describe("[sig-cli] oc adm", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(out).To(o.ContainSubstring("serviceaccount/my-sa-name created"))
 		o.Expect(ocns.Run("get").Args("sa", "my-sa-name").Execute()).To(o.Succeed())
-		err = wait.Poll(cliInterval, cliTimeout, func() (bool, error) {
-			err := ocns.Run("sa", "get-token").Args("my-sa-name").Execute()
-			return err == nil, nil
-		})
-		o.Expect(err).NotTo(o.HaveOccurred())
 
+		// TODO (soltysh): after k8s 1.24 lands we should be able to replace this with:
```
tkashem, line 463: can we fix the test? if the fix cannot land in time then we we disable this test by using the `SkippedUntil` tag from https://github.com/openshift/origin/pull/27086?

soltysh: The other part of that test which relied on the value from that token is disabled, so I'm fine skipping the initial part as well.

PR #27115, opened 2022-05-09
key resources by UID

deads2k:
This is needed to handle static pods.


deads2k:
replaced by https://github.com/openshift/origin/pull/27114

PR #27117, opened 2022-05-10
Add regex matching, regex inverse matching, and default end date option for intervals

DennisPeriquet:
For matching, we use a regex (e.g., to match all multus pods using `-lpod=multus` or `-lpod=multus-.*`).

For inverse matching, if any item matches, it is removed (hence the option is called `--remove`) like `grep -v`.

The default end date (either by not specifying `--end-date` or `--end-date=default`) is set to 1 hour after the latest event (using the `To` field).  This will make it so that older charts don't have a long bar with the rest of the bars being very short.  You can also specify an actual date.

Also tweak the right margin so you can see more of the text (in case you want to search for it in the browser with the `...`).  Increase the width to accommodate the new margin.

### DIFF
--- a/pkg/monitor/monitor_cmd/command.go
+++ b/pkg/monitor/monitor_cmd/command.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -33,9 +34,11 @@ type TimelineOptions struct {
 	PodResourceFilename  string
 	TimelineType         string
 
-	LocatorMatchers []string
-	Namespaces      []string
-	OutputType      string
+	LocatorMatchers        []string
+	removedLocatorMatchers []string
+	Namespaces             []string
+	OutputType             string
+	EndDate                string
 
 	KnownRenderers map[string]RenderFunc
 	KnownTimelines map[string]monitorapi.EventIntervalMatchesFunc
@@ -104,7 +107,9 @@ func (o *TimelineOptions) Bind(flagset *pflag.FlagSet) error {
 	flagset.StringVarP(&o.OutputType, "output", "o", o.OutputType, fmt.Sprintf("type of output: [%s]", strings.Join(sets.StringKeySet(o.KnownRenderers).List(), ",")))
 	flagset.StringVar(&o.TimelineType, "type", o.TimelineType, "type of timeline to produce: "+strings.Join(sets.StringKeySet(o.KnownTimelines).List(), ","))
 	flagset.StringVar(&o.PodResourceFilename, "known-pods", o.PodResourceFilename, "resource-pods_<timestamp>.zip filename from openshift-tests.")
-	flagset.StringSliceVarP(&o.LocatorMatchers, "locator", "l", o.LocatorMatchers, "key=value selector for monitor event locators.  for instance -lpod=openshift-etcd-installer.  The same key listed multiple times means an OR.  Each separate key is logically ANDed")
+	flagset.StringSliceVarP(&o.LocatorMatchers, "locator", "l", o.LocatorMatchers, "key=value selector for monitor event locators (where value is a regex).  for instance -lpod=openshift-etcd-installer.  The same key listed multiple times means an OR.  Each separate key is logically ANDed")
+	flagset.StringSliceVarP(&o.removedLocatorMatchers, "remove", "r", o.removedLocatorMatchers, "key=val selector to remove monitor event locators")
+	flagset.StringVarP(&o.EndDate, "end-date", "e", o.EndDate, "End date (default is one hour after latest event)")
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/monitor/monitor_cmd/command.go
```diff
@@ -104,7 +107,9 @@ func (o *TimelineOptions) Bind(flagset *pflag.FlagSet) error {
 	flagset.StringVarP(&o.OutputType, "output", "o", o.OutputType, fmt.Sprintf("type of output: [%s]", strings.Join(sets.StringKeySet(o.KnownRenderers).List(), ",")))
 	flagset.StringVar(&o.TimelineType, "type", o.TimelineType, "type of timeline to produce: "+strings.Join(sets.StringKeySet(o.KnownTimelines).List(), ","))
 	flagset.StringVar(&o.PodResourceFilename, "known-pods", o.PodResourceFilename, "resource-pods_<timestamp>.zip filename from openshift-tests.")
-	flagset.StringSliceVarP(&o.LocatorMatchers, "locator", "l", o.LocatorMatchers, "key=value selector for monitor event locators.  for instance -lpod=openshift-etcd-installer.  The same key listed multiple times means an OR.  Each separate key is logically ANDed")
+	flagset.StringSliceVarP(&o.LocatorMatchers, "locator", "l", o.LocatorMatchers, "key=value selector for monitor event locators (where value is a regex).  for instance -lpod=openshift-etcd-installer.  The same key listed multiple times means an OR.  Each separate key is logically ANDed")
+	flagset.StringSliceVarP(&o.removedLocatorMatchers, "remove", "r", o.removedLocatorMatchers, "key=val selector to remove monitor event locators")
```
deads2k, line 111: removed or anti-match?

pkg/monitor/monitor_cmd/command.go
```diff
@@ -33,9 +34,11 @@ type TimelineOptions struct {
 	PodResourceFilename  string
 	TimelineType         string
 
-	LocatorMatchers []string
-	Namespaces      []string
-	OutputType      string
+	LocatorMatchers        []string
+	removedLocatorMatchers []string
```
deads2k, line 38: We make these public for external embedding.  kubectl is out example.

pkg/monitor/monitor_cmd/command.go
```diff
@@ -104,7 +107,9 @@ func (o *TimelineOptions) Bind(flagset *pflag.FlagSet) error {
...
 	flagset.StringVar(&o.TimelineType, "type", o.TimelineType, "type of timeline to produce: "+strings.Join(sets.StringKeySet(o.KnownTimelines).List(), ","))
 	flagset.StringVar(&o.PodResourceFilename, "known-pods", o.PodResourceFilename, "resource-pods_<timestamp>.zip filename from openshift-tests.")
-	flagset.StringSliceVarP(&o.LocatorMatchers, "locator", "l", o.LocatorMatchers, "key=value selector for monitor event locators.  for instance -lpod=openshift-etcd-installer.  The same key listed multiple times means an OR.  Each separate key is logically ANDed")
+	flagset.StringSliceVarP(&o.LocatorMatchers, "locator", "l", o.LocatorMatchers, "key=value selector for monitor event locators (where value is a regex).  for instance -lpod=openshift-etcd-installer.  The same key listed multiple times means an OR.  Each separate key is logically ANDed")
+	flagset.StringSliceVarP(&o.removedLocatorMatchers, "remove", "r", o.removedLocatorMatchers, "key=val selector to remove monitor event locators")
+	flagset.StringVarP(&o.EndDate, "end-date", "e", o.EndDate, "End date (default is one hour after latest event)")
```
deads2k, line 112: you need to specify a format

pkg/monitor/monitor_cmd/command.go
```diff
@@ -137,23 +142,49 @@ func (o *TimelineOptions) Validate() error {
...
+	for _, removedMatcher := range o.removedLocatorMatchers {
+		if !strings.Contains(removedMatcher, "=") {
+			return fmt.Errorf("invalid --remove format, must be key=value")
+		}
+	}
+
+	if len(o.EndDate) == 0 {
+		// Nothing specified for end-date so make it default.
+		o.EndDate = "default"
+	}
```
deads2k, line 154: no.  just leave it empty

pkg/monitor/monitor_cmd/command.go
```diff
@@ -165,8 +196,10 @@ type Timeline struct {
 	MonitorEventFilename string
 	PodResourceFilename  string
 
-	LocatorMatcher map[string][]string
-	Namespaces     []string
+	LocatorMatcher        map[string][]*regexp.Regexp
+	RemovedLocatorMatcher map[string][]*regexp.Regexp
+	Namespaces            []string
+	EndDate               string
```
deads2k, line 202: this should be a *time.Time with nil meaning the default

pkg/monitor/monitor_cmd/command.go
DennisPeriquet: the `--remove` option is remove in favor of using the same `--locator` option with anti-match indicating a dash before the regex value.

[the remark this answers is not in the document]
DennisPeriquet: I used dash on the front; I see cluster-debug-tools looking for dash on the front.  But can easily make it with dash on the end.

pkg/monitor/monitor_cmd/command.go
DennisPeriquet: hopefully I did that right using `time.Time{}` to denote nil and `IsZero()` to check for nil.

deads2k: >hopefully I did that right using time.Time{} to denote nil and IsZero() to check for nil.

Why not use a pointer?

pkg/monitor/monitor_cmd/command.go
```diff
@@ -33,9 +34,11 @@ type TimelineOptions struct {
 	PodResourceFilename  string
 	TimelineType         string
 
-	LocatorMatchers []string
-	Namespaces      []string
-	OutputType      string
+	LocatorMatchers        []string
+	RemovedLocatorMatchers []string
```
deads2k, line 38: cruft

pkg/monitor/monitor_cmd/command.go
```diff
@@ -137,23 +141,51 @@ func (o *TimelineOptions) Validate() error {
...
+		if strings.HasPrefix(parts[1], "-") {
+			regExp := regexp.MustCompile(parts[1][1:])
+			inverseLocatorMatcher[parts[0]] = append(inverseLocatorMatcher[parts[0]], regExp)
+		} else {
+			regExp := regexp.MustCompile(parts[1])
+			locatorMatcher[parts[0]] = append(locatorMatcher[parts[0]], regExp)
+		}
+	}
+
+	for _, matcherString := range o.RemovedLocatorMatchers {
```
deads2k, line 170: dead

pkg/monitor/monitor_cmd/command.go
DennisPeriquet: ha, I didn't see the `*` there earlier; I changed it to be `*time.Time`.

pkg/monitor/monitor_cmd/command.go
DennisPeriquet: I still use `RemovedLocatorMatchers` for inverse matchers in the `NotContainsAllParts` function.

deads2k: never has a value, right?

pkg/monitor/monitor_cmd/command.go
DennisPeriquet: you are correct, ... removed.

pkg/monitor/monitor_cmd/command.go
DennisPeriquet: cruft removed :-)

PR #27118, opened 2022-05-10
Bug 2085997: adds new test to the API monitor

p0lyn0mial:
The new tests get data from the memory cache inside the {Kube,OpenShift, OAuth}API server.
When a request fails we will be able to narrow the scope to only the connection with the server itself.
In addition, we will be able to compare it with the tests that hit the database.


aojea:
/lgtm

TIL the backend monitor consider http.status.errors

https://github.com/openshift/origin/blob/a64a4ea06e519d7bc8d0f1299b9e37bca5fb3648/pkg/monitor/backenddisruption/disruption_backend_sampler.go#L337-L351

p0lyn0mial:
`ci/prow/e2e-metal-ipi-ovn-ipv6` keeps failing on https://github.com/openshift/installer/pull/5909

PR #27120, opened 2022-05-10
Don't hardcode the SA token audience when requesting one

alvaroaleman:


test/extended/router/metrics.go
```diff
@@ -91,9 +91,7 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 				context.TODO(),
 				"prometheus-k8s",
 				&authenticationv1.TokenRequest{
-					Spec: authenticationv1.TokenRequestSpec{
-						Audiences: []string{"https://kubernetes.default.svc"},
-					},
+					Spec: authenticationv1.TokenRequestSpec{},
```
alvaroaleman, line 94: /cc @tkashem 

I am not sure if this is the correct thing to do, I am only sure that hardcoding is not the correct thing to do. I tested it locally and it seemed to work.

Miciah:
Thanks!
/approve

s-urbaniak:
According to https://kubernetes-csi.github.io/docs/token-requests.html:

> TokenRequest.Audience
> If it is empty string, the audience of the token is the APIAudiences of kube-apiserver. one of the audiences specified.

PR #27123, opened 2022-05-11
Bug 2083941: feat: add fuzzy match on micro release

eggfoobar:
This PR adds a fuzzy match on single node for previous micro upgrades when we lack data for single node minor upgrades

### DIFF
--- a/pkg/synthetictests/historicaldata/next_best_guess.go
+++ b/pkg/synthetictests/historicaldata/next_best_guess.go
@@ -49,6 +49,7 @@ var nextBestGuessers = []NextBestKey{
 
 	combine(ForTopology("single"), OnSDN),
 	combine(ForTopology("single"), OnSDN, PreviousReleaseUpgrade),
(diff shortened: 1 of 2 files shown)


pkg/synthetictests/historicaldata/next_best_guess.go
```diff
@@ -49,6 +49,7 @@ var nextBestGuessers = []NextBestKey{
 
 	combine(ForTopology("single"), OnSDN),
 	combine(ForTopology("single"), OnSDN, PreviousReleaseUpgrade),
+	combine(ForTopology("single"), OnSDN, PreviousMicroReleaseUpgrade),
```
deads2k, line 52: does `combined(forTopology, onsdn, previousreleaseupgrade, microreleaseupgrade)` work?

eggfoobar: Yup! Thank you! Updated :)

PR #27128, opened 2022-05-11
CNF-5208: Increase durations for RT Upgrades

eggfoobar:
Upgrades in 4.11 RT can take longer due to extra restarts by the worker nodes. This is due to a change in PAO. 

[Related PR](https://github.com/openshift/origin/pull/26948)


eggfoobar:
No longer needed

PR #27132, opened 2022-05-11
add test to catch single second disruption patterns

deads2k:
Many single second disruptions usually indicates a pattern where a particular node is not handling traffic properly. this will let us catch this (we think) rare pattern in CI.


jluhrsen:
I'm not the best for reviews here, but seems clean to me.
/lgtm

PR #27134, opened 2022-05-12
Bug 2085997: gives the backend monitoring an identity

p0lyn0mial:
the identity will ill be used to trace audit logs for requests that were sent by the monitor tool

the primary use case will be to find requests that arrived at an API server during its shutdown procedure.

### DIFF
--- a/pkg/monitor/backenddisruption/disruption_backend_sampler.go
+++ b/pkg/monitor/backenddisruption/disruption_backend_sampler.go
@@ -65,6 +65,9 @@ type BackendSampler struct {
 	// http status code is accepted.
 	expectRegexp *regexp.Regexp
 
+	// userAgent used to sets the User-Agent HTTP Header for all requests that are sent by this sampler
+	userAgent string
+
 	// initHTTPClient ensures we only create the http client once
 	initHTTPClient sync.Once
 	// httpClient is used to connect to the host+path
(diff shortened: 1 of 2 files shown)


pkg/monitor/backenddisruption/disruption_backend_sampler.go
```diff
@@ -282,15 +291,20 @@ func (b *BackendSampler) GetHTTPClient() (*http.Client, error) {
...
+		var err error
+		rt := http.RoundTripper(httpTransport)
+		rt, err = b.wrapWithAuth(rt)
 		if err != nil {
 			b.httpClient = nil
 			b.httpClientErr = err
 			return
 		}
+		if len(b.userAgent) > 0 {
+			rt = transport.NewUserAgentRoundTripper(b.userAgent, rt)
```
aojea, line 303: what is the default? should we add a meaningful default user agent?

p0lyn0mial: no default value per se, however, the sole user of the sampler `createAPIServerBackendSampler` sets a value.

p0lyn0mial: I don't think we should add a default user agent as it is up to the user that uses the sampler

soltysh (approved):
/approve

p0lyn0mial:
`ci/prow/e2e-metal-ipi-ovn-ipv6` keeps failing on https://github.com/openshift/installer/pull/5909

PR #27135, opened 2022-05-12
NE-626: Re-introduce the DNS CI coverage for golang and glibc resolver libraries

suleymanakbas91:
This PR re-introduces the DNS CI coverage for golang and glibc resolver libraries by reverting 4a1e80297759a22a4cae03c3deb6171c0fd382eb. 

Modifications to the previous PR: 
- Base image for Go test was changed from `golang` in Docker Hub to `registry.redhat.io/ubi8/go-toolset`.
- A test for Go 1.16 and 1.17 were added, 1.18 was removed as the `go-toolset` image does not yet support 1.18.
- A check for BareMetal environment was added to skip the test, as this environment is disconnected and does not have any suitable images on the local registry to use for different Go versions.

### DIFF
--- a/test/extended/dns/dns_libraries.go
+++ b/test/extended/dns/dns_libraries.go
@@ -1,0 +1,134 @@
+package dns
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	kapiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	watchtools "k8s.io/client-go/tools/watch"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+)
+
+// checkForPodLogFailures goes through the pod logs and determines if there was a failure
+// by looking for "fail" keyword in the logs.
+func checkForPodLogFailures(f *e2e.Framework, pod *kapiv1.Pod) {
+	By("submitting the pod to kubernetes")
+	podClient := f.ClientSet.CoreV1().Pods(f.Namespace.Name)
+	updated, err := podClient.Create(context.Background(), pod, metav1.CreateOptions{})
+	if err != nil {
+		e2e.Failf("Failed to create %s pod: %v", pod.Name, err)
+	}
+
+	w, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Watch(context.Background(), metav1.SingleObject(metav1.ObjectMeta{Name: pod.Name, ResourceVersion: updated.ResourceVersion}))
+	if err != nil {
+		e2e.Failf("Failed to watch pods: %v", err)
+	}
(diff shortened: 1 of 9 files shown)


test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,134 @@
...
+				e2e.Failf("Got %d pods with labels %v, expected 1", len(pods), labels)
+			}
+
+			By("expect the go application to successfully resolve DNS")
+			pod, err := oc.KubeClient().CoreV1().Pods(oc.Namespace()).Get(context.Background(), pods[0], metav1.GetOptions{})
+			if err != nil {
+				e2e.Failf("Failed to get dns-libraries-go pod: %v", err)
+			}
+			//execute "go run /go/dns_libraries.go -cluster-ip={CLUSTER_IP}" command inside the pod
+			args := []string{pod.Name, "-c", pod.Spec.Containers[0].Name, "--", "bash", "-c", fmt.Sprintf("go run /go/dns_libraries.go -cluster-ip=%q", dnsService.Spec.ClusterIP)}
```
gcs278, line 110: FWIW there is a `RunHostCmdOrDie` helper function that could clean this up a bit.

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,134 @@
...
+
+		dnsService, err := f.ClientSet.CoreV1().Services("openshift-dns").Get(context.Background(), "dns-default", metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("Failed to get default DNS service: %v", err)
+		}
+
+		goVersions := []string{"go-v1-16", "go-v1-17"}
+		for _, version := range goVersions {
+			By(fmt.Sprintf("creating build and deployment config for dns-libraries-%s", version))
+			err = oc.Run("create").Args("-f", exutil.FixturePath("testdata", "dns", version, "dns_libraries_go.yaml")).Execute()
```
gcs278, line 82: From what I understand, David Eads has been pushing to avoid CLI commands as much as possible (he refers to it as "shelling out"), unless you are testing the CLI. Is there a reason why `dns_libraries_go.yaml` couldn't be generated as a GO object and created via the GO API directly?

test/extended/testdata/dns/go-v1-16/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM registry.redhat.io/ubi8/go-toolset:1.16.12
+
```
gcs278, line 2: Could this file be version-agnostic? With a parameter as the version tag?

test/extended/dns/dns_libraries.go
```diff
@@ -0,0 +1,134 @@
...
+			if err != nil {
+				e2e.Failf("Failed to create build configs and deployment config for dns-libraries-go: %v", err)
+			}
+
+			By("starting the builder image build with a directory")
+			err = oc.Run("start-build").Args(fmt.Sprintf("dns-libraries-%s", version), fmt.Sprintf("--from-dir=%s", exutil.FixturePath("testdata", "dns", version))).Execute()
```
gcs278, line 88: Same comment here. I don't know how this can be done with this the kube clientset, but I'm curious if this could be a GO Object that is created via GO API directly (avoid "shelling out")

gcs278:
/close
In favor of https://github.com/openshift/origin/pull/27644

gcs278:
/close
In favor of https://github.com/openshift/origin/pull/27644

PR #27137, opened 2022-05-12
handle cases where containers restart

deads2k:
When containers restart and do not end, the container lifecycle continues.  This fixes that logic.

### DIFF
--- a/pkg/monitor/pod.go
+++ b/pkg/monitor/pod.go
@@ -205,7 +205,9 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
 			}
 
 			// if this container is not terminated, then we don't need to compute an event for it.
-			if containerStatus.LastTerminationState.Terminated == nil {
+			lastTerminated := containerStatus.LastTerminationState.Terminated != nil
+			currentTerminated := containerStatus.State.Terminated != nil
+			if !lastTerminated && !currentTerminated {
 				continue
 			}
 
@@ -215,7 +217,7 @@ func startPodMonitoring(ctx context.Context, m Recorder, client kubernetes.Inter
 				// the container exit because it will be a disconnected event that can confuse our container lifecycle ordering based
 				// on the event stream
 
-			case oldContainerStatus.LastTerminationState.Terminated == nil:
+			case lastTerminated && oldContainerStatus.LastTerminationState.Terminated == nil:
 				// if we are transitioning to a terminated state
 				if containerStatus.LastTerminationState.Terminated.ExitCode != 0 {
 					conditions = append(conditions, monitorapi.Condition{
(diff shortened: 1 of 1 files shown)


pkg/monitor/event.go
```diff
@@ -123,18 +123,7 @@ func recordAddOrUpdateEvent(
 	// special case some very common events
 	switch obj.Reason {
 	case "":
-	case "Scheduled":
```
deads2k, line 126: This was duplicated

pkg/monitor/event.go
```diff
@@ -123,18 +123,7 @@ func recordAddOrUpdateEvent(
...
-			if strings.HasPrefix(message, "Successfully assigned ") {
-				if i := strings.Index(message, " to "); i != -1 {
-					node := message[i+4:]
-					message = fmt.Sprintf("node/%s reason/%s", node, obj.Reason)
-					break
-				}
-			}
-		}
-		message = fmt.Sprintf("reason/%s %s", obj.Reason, message)
-	case "Started", "Created", "Killing":
```
deads2k, line 137: so was started and created here.

pkg/monitor/intervalcreation/pod.go
```diff
@@ -305,11 +305,11 @@ func (t containerLifecycleTimeBounder) getEndTime(inLocator string) time.Time {
 	if !ok {
 		return t.delegate.getEndTime(locator)
 	}
-	for i := len(containerEvents) - 1; i >= 0; i-- {
-		event := containerEvents[i]
-		if monitorapi.ReasonFrom(event.Message) == monitorapi.ContainerReasonContainerExit {
-			return event.From
-		}
+	// if the last event is a containerExit, then that's as long as the container lasted.
```
deads2k, line 308: this is the core fix.

pkg/monitor/pod.go
```diff
@@ -597,45 +569,14 @@ func lastContainerTimeFromStatus(current *corev1.ContainerStatus) time.Time {
 	return time.Time{}
 }
 
-func conditionsForTransitioningContainer(pod *corev1.Pod, current, previous *corev1.ContainerStatus, init bool, reason, cause, message string, currentTime time.Time, lastContainerTime time.Time) []monitorapi.Condition {
```
deads2k, line 600: the different cases here didn't provide value. They exist from before we started building a history of pod lifecycle in true intervals after the instants were reported.

pkg/monitor/intervalcreation/pod.go
```diff
@@ -305,11 +305,11 @@ func (t containerLifecycleTimeBounder) getEndTime(inLocator string) time.Time {
...
+	// if the last event isn't a containerExit, then the last time we're aware of for the container is parent.
```
DennisPeriquet, line 309: you mean `for the container's parent` right?

deads2k:
built. this improves reporting, merging.

PR #27127, opened 2022-05-11
moves TestScalingUpAndDownSingleNode from cluster-etcd-operator

p0lyn0mial:
brings back https://github.com/openshift/origin/pull/26917 and pulls https://github.com/openshift/origin/pull/27121 in

PR #27140, opened 2022-05-13
Bug 2085997: special configuration on barematal platform for the etcd vertical test is required

p0lyn0mial:
for baremetal platforms requires an extra worker to be previously deployed to allow scaling operations. Otherwise creating a new machine has no effect and is stuck on `No available BareMetalHost found`

requires https://github.com/openshift/origin/pull/27127

### DIFF
--- a/pkg/synthetictests/duplicated_events.go
+++ b/pkg/synthetictests/duplicated_events.go
@@ -569,7 +569,7 @@ func newDuplicatedEventsAllowedWhenEtcdRevisionChange(ctx context.Context, opera
 		return nil, err
 	}
 	return &etcdRevisionChangeAllowance{
-		allowedGuardProbeFailurePattern:        regexp.MustCompile(`ns/openshift-etcd pod/etcd-guard-ip-.* node/[a-z0-9.-]+ - reason/(Unhealthy|ProbeError) Readiness probe.*`),
+		allowedGuardProbeFailurePattern:        regexp.MustCompile(`ns/openshift-etcd pod/etcd-guard-.* node/[a-z0-9.-]+ - reason/(Unhealthy|ProbeError) Readiness probe.*`),
 		maxAllowedGuardProbeFailurePerRevision: 60 / 5, // 60s for starting a new pod, divided by the probe interval
 		currentRevision:                        currentRevision,
 	}, nil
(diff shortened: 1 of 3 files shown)


cmd/openshift-tests/e2e.go
```diff
@@ -120,8 +120,8 @@ var staticSuites = testSuites{
 				}
 				return strings.Contains(name, "[Suite:openshift/conformance/serial") || isStandardEarlyOrLateTest(name)
 			},
-			// doubling default test timeout (15 mins) which is needed for external ip configuration related serial tests.
-			TestTimeout:         30 * time.Minute,
+			// etcd's vertical scaling test is expensive
```
soltysh, line 123: shouldn't we move that test to entirely new suite in that case?

pkg/synthetictests/duplicated_events.go
```diff
@@ -582,7 +582,7 @@ func (a *etcdRevisionChangeAllowance) allowEtcdGuardReadinessProbeFailure(monito
 	// allow for a.maxAllowedGuardProbeFailurePerRevision * a.currentRevision failed readiness probe from the etcd-guard pods
 	// since the guards are static and the etcd pods come and go during a rollout
 	// which causes allowedGuardProbeFailurePattern to fire
-	if a.allowedGuardProbeFailurePattern.MatchString(eventMessage) && a.maxAllowedGuardProbeFailurePerRevision*a.currentRevision < times {
+	if a.allowedGuardProbeFailurePattern.MatchString(eventMessage) && a.maxAllowedGuardProbeFailurePerRevision*a.currentRevision > times {
```
soltysh, line 585: Yikes, nice catch

test/extended/etcd/helpers/client.go
```diff
@@ -0,0 +1,117 @@
...
+
+	clientv3 "go.etcd.io/etcd/client/v3"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
+)
+
+type EtcdClientCreator interface {
```
soltysh, line 20: How about doing something similar to Prometheus Client from https://github.com/openshift/cluster-kube-controller-manager-operator/pull/623 to speak directly with etcd rather than relying on port-forward?

cmd/openshift-tests/e2e.go
p0lyn0mial: nope, we want to have at least one test in the serial suite to cover a wide range of various platforms and block on promotions.

p0lyn0mial: we should create a new suite for more scaling e2e tests

test/extended/etcd/helpers/client.go
p0lyn0mial: I don't think so, the difference is that the test doesn't run inside the cluster. It needs to access the etcd cluster from the public internet.

PR #27141, opened 2022-05-13
Bug 2085997: ensure number of voting members in the cm matches the number of master machines

p0lyn0mial:
a simple test for ensuring the number of etcd voting members matches the number of master nodes.
the test will be run right after cluster installation and before running any tests.


soltysh (commented):
/approve

PR #27144, opened 2022-05-16
Add user coreydaley as an approver

coreydaley:


OWNERS_ALIASES
```diff
@@ -17,6 +17,7 @@ aliases:
   - LalatenduMohanty # Over the Air Updates
   - Miciah # Network Edge
   - adambkaplan # Build + Jenkins
+  - coreydaley # Build + Jenkins
```
bparees, line 20: remove adam while you're in here

test/extended/OWNERS
```diff
@@ -16,3 +17,4 @@ approvers:
   - ironcladlou
   - adambkaplan
   - danwinship
+  - coreydaley
```
bparees, line 20: you shouldn't need this level of approval, add yourself to the test/extended/$subdirs your team owns tests or library code in.

bparees (commented):
you should remove adam in places where you are replacing his role

coreydaley:
@bparees updated, ptal

coreydaley:
@bparees Can you skip this failing test?

bparees:
> @bparees Can you skip this failing test?

https://docs.ci.openshift.org/docs/how-tos/overriding-failing-ci-jobs/

```
Confirm that someone/team owns and is aware of the failure you are trying to override. Don’t just override something broken and let the rest of the org keep running into it until someone else reports it. You don’t have to solve it, but at least make sure there’s an active/current bug or that the team in question acknowledges the issue.
```

coreydaley:
@bparees 
Opened bug https://bugzilla.redhat.com/show_bug.cgi?id=2087285 for this test that is continuously failing.

PR #27146, opened 2022-05-17
Suppress failure on alerts if other events happened in a related time

DennisPeriquet:
Regarding [TRT-238](https://issues.redhat.com/browse/TRT-238)

* If we see KubePodNotReady alert firing and the associated pod suffered from an imagePullBackoff event within the last 10 minutes, we flake on the alert test.
* If we see a RedhatOperatorsCatalogError alert firing but there's a redhat-operator pod not in Pending state 10 minutes or more after, we flake on the alert test.

### DIFF
--- a/test/e2e/upgrade/alert/alert.go
+++ b/test/e2e/upgrade/alert/alert.go
@@ -14,7 +14,10 @@ import (
 	helper "github.com/openshift/origin/test/extended/util/prometheus"
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/common/model"
+	kapiv1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


test/e2e/upgrade/alert/alert.go
```diff
@@ -272,3 +287,73 @@ sort_desc(
 func (t *UpgradeTest) Teardown(f *framework.Framework) {
 	// rely on the namespace deletion to clean up everything
 }
+
+// redhatOperatorPodsNotPending returns true of we determined that there is a redhat-operator
+// pod not in Pending state impling that we don't need to fail on the RedhatOperatorsCatalogError
+// alert.
+func redhatOperatorPodsNotPending(alertTime time.Time, kubeClient kubernetes.Interface) bool {
```
deads2k, line 294: this is close, but you need to use the tracked resources in the monitor

DennisPeriquet:
closed in favor of https://github.com/openshift/origin/pull/27304

PR #27147, opened 2022-05-17
Bug 2085997: the machine API can be unavailable resulting in a 404 or an empty list

p0lyn0mial:
on some platforms, machine API can be totally off resulting in a 404 response, and on some, we might get an empty list.
this PR changes the etcd scaling test to take into account both scenarios.


soltysh (approved):
/lgtm
/approve

PR #27151, opened 2022-05-18
add per-namespace tests for kubepodnotready alert failures

deads2k:
this is a building block to allow allowances configured per-alert and per-namespace

/assign @DennisPeriquet


pkg/monitor/alerts.go
```diff
@@ -48,6 +61,14 @@ func whenWasAlertInState(ctx context.Context, prometheusClient prometheusv1.API,
 		return nil, err
 	}
 
+	if namespace == platformidentification.NamespaceOther {
```
DennisPeriquet, line 64: Do you want this to say `if namespace == platformidentification.NamespaceOther || len(namespace) == 0`? since the query above is identical for those two cases?

pkg/synthetictests/allowedalerts/types.go
```diff
@@ -67,3 +67,6 @@ func getCurrentResults() historicaldata.BestMatcher {
 
 	return historicalData
 }
+
+// interestingNamespaces is the list of known namespaces we have in an openshift cluster
+var interestingNamespaces = []string{}
```
DennisPeriquet, line 72: I don't see any place where this is used.

pkg/synthetictests/allowedalerts/basic_alert.go
```diff
@@ -47,67 +46,129 @@ const (
...
 
-func (a *basicAlertTest) pending() *basicAlertTest {
+func (a *alertBuilder) pending() *alertBuilder {
 	a.alertState = AlertPending
 	return a
 }
 
-func (a *basicAlertTest) firing() *basicAlertTest {
+func (a *alertBuilder) firing() *alertBuilder {
 	a.alertState = AlertInfo
```
DennisPeriquet, line 96: Is this right? or should it be something like AlertCritical?

DennisPeriquet: ah, I see line 41 and ordering so maybe this is fine.

PR #27153, opened 2022-05-19
add pod graceful termination to pod lifecycle

deads2k:
Adds a pod lifecycle state for graceful termination so the graph makes more sense and allows for early deletion when termination is done early.

### DIFF
--- a/pkg/monitor/intervalcreation/pod.go
+++ b/pkg/monitor/intervalcreation/pod.go
@@ -151,10 +151,8 @@ func (t podLifecycleTimeBounder) getEndTime(inLocator string) time.Time {
 		return *runOnceContainerTermination
 	}
 
-	// for other pod types, use the deletion time if we have it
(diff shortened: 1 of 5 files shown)


pkg/monitor/intervalcreation/pod.go
```diff
@@ -228,8 +231,11 @@ func (t podLifecycleTimeBounder) getPodDeletionTime(inLocator string) *time.Time
 		return nil
 	}
 
-	temp := pod.DeletionTimestamp
-	return &temp.Time
+	deletionTime := pod.DeletionTimestamp.Time
+	if pod.Spec.TerminationGracePeriodSeconds != nil {
+		deletionTime = deletionTime.Add(time.Duration(*pod.Spec.TerminationGracePeriodSeconds * int64(time.Second)))
```
DennisPeriquet, line 236: are we assuming that no one does `oc delete pod --grace-period=0`

PR #27157, opened 2022-05-19
Ignore tunbr NodeNetworkInterfaceFlapping alert

jluhrsen:
tunbr is an ephemeral bridge interface that can go up an down depending
on if the ephemeral network is created/deleted, we shouldnt care if that
happens. Same as we dont care if a veth interface goes up/down as that
is going to happen with CNI pod events.

Looking in to what it will take to have tunbr automatically ignored
in these alerts and not fire them. When that work is done we can
revert this commit.

https://issues.redhat.com/browse/SDN-3008

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -309,6 +309,11 @@ var _ = g.Describe("[sig-instrumentation][Late] Alerts", func() {
 				Selector: map[string]string{"alertname": "KubeJobFailed", "namespace": "openshift-multus"}, // not sure how to do a job_name prefix
 				Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=2054426",
 			},
+			{
+				Selector: map[string]string{"alertname": "NodeNetworkInterfaceFlapping", "device": "tunbr", "job": "node-exporter"},
+				Text:     "tunbr is an ephemeral bridge interface that can go up an down depending on if the ephemeral network is created/deleted",
+				// https://issues.redhat.com/browse/SDN-3008
+			},
 		}
 		allowedFiringAlerts := helper.MetricConditions{
 			{


jluhrsen:
@simonpasquier @jan--f @arajkumar , is this ok with you folks?

dcbw:
@simonpasquier does https://github.com/openshift/cluster-monitoring-operator/pull/1680 seem like the right approach?

jluhrsen:
> @simonpasquier does [openshift/cluster-monitoring-operator#1680](https://github.com/openshift/cluster-monitoring-operator/pull/1680) seem like the right approach?

@dcbw , I had this idea originally, but couldn't figure it out. I though we needed to get this 'tunbr' string in to the actual
prom rules yaml. I pushed [this PR](https://github.com/openshift/cluster-monitoring-operator/pull/1672/commits/57dae8c4ee49eff45bdd78673fe246ec3e5ffb6d) but it fails because that yaml is autogenerated from what I can tell. Then I thought the
generation came from [prom-operator/kube-prom](https://github.com/prometheus-operator/kube-prometheus) and at that point I figured it would take a long time to end up in openshift.

But if your PR works, let's get it in please. This is a failure in CI that bites us quite frequently

dcbw:
> But if your PR works, let's get it in please. This is a failure in CI that bites us quite frequently

@jluhrsen it doens't work yet, it appears we still need some fixes in prometheus-operator. Let's ignore the alert for now.

jluhrsen:
> > But if your PR works, let's get it in please. This is a failure in CI that bites us quite frequently
> 
> @jluhrsen it doens't work yet, it appears we still need some fixes in prometheus-operator. Let's ignore the alert for now.

sounds good. @simonpasquier are you ok to approve this please?

test/extended/prometheus/prometheus.go
```diff
@@ -331,6 +331,11 @@ var _ = g.Describe("[sig-instrumentation][Late] Alerts", func() {
 				Selector: map[string]string{"alertname": "ExtremelyHighIndividualControlPlaneCPU"},
 				Text:     "high CPU utilization during e2e runs is normal",
 			},
+			{
+				Selector: map[string]string{"alertname": "NodeNetworkInterfaceFlapping", "device": "tunbr", "job": "node-exporter"},
+				Text:     "tunbr is an ephemeral bridge interface that can go up an down depending on if the ephemeral network is created/deleted",
+				// https://issues.redhat.com/browse/SDN-3008
```
simonpasquier, line 337: I suggest that you move it to `firingAlertsWithBugs` instead since it already has a bugzilla.

jluhrsen: yes, thank you. just updated it. I should have noticed that the first time.

jluhrsen:
> Can we close this since [openshift/cluster-monitoring-operator#1681](https://github.com/openshift/cluster-monitoring-operator/pull/1681) has merged?

absolutely :)
Thanks @simonpasquier

PR #27161, opened 2022-05-23
don't trigger single second disruption on contiguous disruption

deads2k:
to avoid firing inaccurately in cases like https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27151/pull-ci-openshift-origin-master-e2e-aws-single-node-upgrade/1526997253254811648 where many single-second failures are contiguous with different messages.

We want to keep the separate disruption events to show these different failures because the message matters and in this case, is weird. But this particular test should not fail.


deads2k:
no conflict, merging to help CI signal

PR #27163, opened 2022-05-24
Bug 2085997: EnsureMemberRemoved tolerates some failures

p0lyn0mial:
The etcd client factory used by the method establishes a stream connection to an etcd pod.
Since a connection like this is considered a long-running connection
and isn't subjected to gracefully termination
and the fact it runs after deleting a machine gives a reason to tolerate some failures while getting a list of members.

Note that unavailability of the APIs and the etcd cluster will be reported by the monitor tool.


geliu2016 (approved):
/lgtm

soltysh (approved):
/approve

p0lyn0mial:
/close

replaced by https://github.com/openshift/origin/pull/27176

PR #27172, opened 2022-05-25
Bug 2077167: disable unidling test failing under ovn-k

candita:
Some but not all unidling tests were disabled when running on ovn-kubernetes.  This one was overlooked.


candita:
> failed: (1.3s) 2022-05-25T01:54:18 "[sig-arch][Late] operators should not create watch channels very often [Suite:openshift/conformance/parallel]"

PR #27178, opened 2022-05-26
Bug 2086092: skip tests to facilitate rebase

tkashem:
kube v1.24 update https://github.com/openshift/kubernetes/pull/1252 has landed, we need to skip these tests until we have a good clear signal with new kubelet


soltysh (approved):
/lgtm
/approve

PR #27179, opened 2022-05-26
Add disruption test to an external service, not cluster under test.

dgoodwin:
TRT has identified situations where disruption testing fails for a
batch of jobs talking to clusters in many platforms, all at the same
time, when the ci jobs were all running on same build cluster.

To help identify when we're seeing real disruption of the cluster under
test, vs the CI cluster itself losing network, we've stood up a simple
service in DPCR cluster, and added a new disruption backend that will
contiually poll that endpoint. If we see disruption here, we know any
disruption of the cluster under test is likely not real.

In future this may expand to automatically pass disruption tests that
fail if we see corresponding network problems for the CI cluster.


dgoodwin:
Testing how this behaves on PR jobs after manual testing indicates this should work.

dgoodwin:
Looks to be not causing problems now that the service is running again.

Ready for review.

deads2k:
Can we do both reused and new?

Also, please confirm we're using no-store in the cache headers.

dgoodwin:
Added reused as well, and the backing service now returns no-store.

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,98 @@
...
+// is losing networking.
+//
+// The service in question is maintained by the TRT team, and running on the DPCR cluster in
+// the trt-monitoring namespace. (ci-disruption-tester svc)
+type ciDisruptionUpgradeTest struct {
+	// filled in by pre-setup
+	jig                 *service.TestJig
+	tcpService          *v1.Service
+	unsupportedPlatform bool
+	hostGetter          *backenddisruption.SimpleHostGetter
```
deads2k, line 25: where is this used?

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,98 @@
...
+// is used to compare if we're actually observing disruption in the cluster under test, or if the CI cluster itself
+// is losing networking.
+//
+// The service in question is maintained by the TRT team, and running on the DPCR cluster in
+// the trt-monitoring namespace. (ci-disruption-tester svc)
+type ciDisruptionUpgradeTest struct {
+	// filled in by pre-setup
+	jig                 *service.TestJig
+	tcpService          *v1.Service
+	unsupportedPlatform bool
```
deads2k, line 24: where is this set?

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,98 @@
...
+
+// ciDisruptionUpgradeTest tests the actual CI cluster where tests are running is able to reach an external service. This
+// is used to compare if we're actually observing disruption in the cluster under test, or if the CI cluster itself
+// is losing networking.
+//
+// The service in question is maintained by the TRT team, and running on the DPCR cluster in
+// the trt-monitoring namespace. (ci-disruption-tester svc)
+type ciDisruptionUpgradeTest struct {
+	// filled in by pre-setup
+	jig                 *service.TestJig
```
deads2k, line 22: copy-paste/

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (t *ciDisruptionUpgradeTest) Name() string { return t.backendDisruptionTest.Name() }
+
+func (t *ciDisruptionUpgradeTest) DisplayName() string {
+	return t.backendDisruptionTest.DisplayName()
+}
+
+// RequiresKubeNamespace indicates we get an e2e-k8s- namespace so we can bind low ports.
+func (t *ciDisruptionUpgradeTest) RequiresKubeNamespace() bool {
```
deads2k, line 79: unnecessary.

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,98 @@
...
+	return false
+}
+
+// Test runs a connectivity check to the service.
+func (t *ciDisruptionUpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade upgrades.UpgradeType) {
+	if t.unsupportedPlatform {
```
deads2k, line 85: unnecessary i think

dgoodwin:
Even shorter now. :)

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,77 @@
...
+			"CI cluster where tests are running may have network issues (not the cluster under test)",
+		)
+
+	return ciDisruptTest
+}
+
+func NewCIDisruptionWithReusedConnectionsTest() upgrades.Test {
+	ciDisruptTest := &ciDisruptionUpgradeTest{}
+	backend := backenddisruption.NewSimpleBackend(
+		"https://trt-ci-disruption-tester.dptools.openshift.org",
```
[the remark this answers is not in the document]
stbenjam: There's also neverssl.com but I don't know how much is too much to be hitting that one.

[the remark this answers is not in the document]
dgoodwin: that looks great, thanks! I'll nuke the service in dpcr.

dgoodwin:
Updated to hit the static file, I'll go kill off our service on dpcr.

stbenjam:
@dgoodwin Can you file a bug or get David to add the label?

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,77 @@
...
+		"ci-cluster-network-liveness",
+		"",
+		backenddisruption.NewConnectionType)
+	allowed := 1 * time.Second
+	ciDisruptTest.backendDisruptionTest =
+		disruption.NewBackendDisruptionTestWithFixedAllowedDisruption(
+			"[sig-trt] CI cluster remains able to communicate with an external service with new connections",
+			backend,
+			&allowed, // We'll let one second slide without reporting a problem
+			"CI cluster where tests are running may have network issues (not the cluster under test)",
```
DennisPeriquet, line 35: I'd add something that indicates this message came out of the test for New connections so we can distinguish it from messages that came out from the Used connections.

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -0,0 +1,77 @@
...
+		"ci-cluster-network-liveness",
+		"/",
+		backenddisruption.NewConnectionType)
+	allowed := 1 * time.Second
+	ciDisruptTest.backendDisruptionTest =
+		disruption.NewBackendDisruptionTestWithFixedAllowedDisruption(
+			"[sig-trt] CI cluster remains able to communicate with an external service using new connections",
+			backend,
+			&allowed, // We'll let one second slide without reporting a problem
+			"CI cluster where tests are running may have network issues (not the cluster under test)",
```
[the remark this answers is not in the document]
DennisPeriquet: This message is identical to the one above it for new Connections, I suggest making it different so we can tell which test caused the message.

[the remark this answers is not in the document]
DennisPeriquet: My suggestion: make these flake for now (I agree with out that no one but TRT will know to look at these in case there are disruption failures).  But we can keep tabs on the test (with no risk of it failing payloads) when there are disruption failures to see if this test also fails too -- on our investigation, we can make the determination if the test should or should not have failed and make adjustments as necessary.

After some time, when we think the test is really reliable, we can unflake it.

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
dgoodwin: The test name reflects new/reused, does that cover us?

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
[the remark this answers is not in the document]
dgoodwin: It's merged as a fail but I will keep an eye on it. If it looks like it's firing when it shouldn't, I will adjust to a flake somehow. I'd really like it to be in people's faces when they look at a PR with disruption, but if I see any reasonable number of it failing without other disruption, I'm with you, it needs to be a flake and something we interpret behind the scenes.

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
DennisPeriquet: yes

PR #27180, opened 2022-05-26
update service load balancer test to gracefully roll out

deads2k:
This adds a readyz=false state into the service being served by the load balancer to allow time for the load balancer to realize that an endpoint is no longer ready.  During this time period, requests are still handled by the server, so we should see a reduction in the disruption as a result.

### DIFF
--- a/test/e2e/upgrade/service/service.go
+++ b/test/e2e/upgrade/service/service.go
@@ -159,13 +159,14 @@ func (t *serviceLoadBalancerUpgradeTest) loadBalancerSetup(f *framework.Framewor
 		// ensure the pod waits long enough during update for the LB to see the newly ready pod, which
 		// must be longer than the worst load balancer above (GCP at 32s)
 		rc.Spec.MinReadySeconds = 33
-		// ensure the pod waits long enough for most LBs to take it out of rotation, which has to be
-		// longer than the LB failed health check duration + 1 cycle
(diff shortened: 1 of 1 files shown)


deads2k:
```
                "message": "Back-off pulling image \"quay.io/openshift/community-e2e-images:e2e-1-k8s-gcr-io-e2e-test-images-agnhost-2-39-D3QKX7Bo93z7tlZq\"",
                "reason": "ImagePullBackOff"
```

test/extended/util/image/image.go
```diff
@@ -65,7 +65,7 @@ func init() {
 
 		// allowed upstream kube images - index and value must match upstream or
 		// tests will fail (vendor/k8s.io/kubernetes/test/utils/image/manifest.go)
-		"k8s.gcr.io/e2e-test-images/agnhost:2.33":   1,
+		"k8s.gcr.io/e2e-test-images/agnhost:2.39":   1,
```
yselkowitz, line 68: This version is defined by the kubernetes test suite, and is being updated to 2.36 as part of the k8s 1.24.0 rebase in #27181.

yselkowitz: Also, all the rest of the tests in origin that use `agnhost` also need to be updated to the same version simultaneously.

deads2k:
This is working in the sense that we go ready=false now.  We still have an outage, but it's an improvement.  To be continued once the main o/o update is done.

deads2k:
/hold cancel

this is now ready to merge.

PR #27191, opened 2022-05-31
Bug 2098094: EgressIP test: Remove print statement on test failure

andreaskaris:
Remove the print statement on test failure:
a) the print function was trying to print elements "host" instead of
"node" (this was a type)
b) the print statement is of limited value at the end of tests and tends
to pollute the logs (and release's artifacts should contain the same
and more information)

### DIFF
--- a/test/extended/networking/egressip.go
+++ b/test/extended/networking/egressip.go
@@ -1,7 +1,6 @@
 package networking
 
 import (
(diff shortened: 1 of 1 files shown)


jluhrsen:
/retest

can you also update the commit message to say why you are removing the print statement?

PR #27193, opened 2022-05-31
wait for kube-apiserver to settle after modifying network configuration

deads2k:
/assign @neisw


deads2k:
This fixed the dirty cluster problem for the external IP test.  Merging to help payloads.

PR #27194, opened 2022-05-31
clean up logic for waiting authentication operator to finish progressing

deads2k:
I happened to need a general method for networking.  See if this looks roughly equivalent.

/assign @ibihim

### DIFF
--- a/test/extended/oauth/requestheaders.go
+++ b/test/extended/oauth/requestheaders.go
@@ -17,23 +17,17 @@ import (
 	"regexp"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
-
+	configv1 "github.com/openshift/api/config/v1"
+	osinv1 "github.com/openshift/api/osin/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
-
-	configv1 "github.com/openshift/api/config/v1"
-	osinv1 "github.com/openshift/api/osin/v1"
-	clusteroperatorhelpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
-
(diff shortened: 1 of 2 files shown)


test/extended/oauth/requestheaders.go
```diff
@@ -504,32 +508,3 @@ func getTokenFromResponse(resp *http.Response) string {
...
-			return false, err
-		}
-
-		progressing := clusteroperatorhelpers.FindStatusCondition(authn.Status.Conditions, configv1.OperatorProgressing)
-		if progressing == nil || progressing.Status != expectedProgressing {
-			e2e.Logf("Waiting for progressing condition to be %q: %s", expectedProgressing, spew.Sdump(authn.Status.Conditions))
-			return false, nil
-		}
-
-		if expectedProgressing == configv1.ConditionFalse {
```
ibihim, line 522: I don't see this reflected in your changes:

```mermaid
flowchart LR
progressing["progressing: false"]
available["available: true"]
degraded["degraded: false"]
buggy["mermaid is buggy"]

progressing -->|AND| available
progressing -->|AND| degraded
```

deads2k: I don't think the degraded check was doing what you thought it was doing.  You need to be sure the level is achieved, but I wouldn't expect a degraded on a two minute momentum to have a meaningful impact for you

test/extended/util/operator.go
```diff
@@ -91,7 +94,31 @@ func waitForOperatorProgressingToBe(ctx context.Context, configClient configv1cl
...
+			operator := obj.(*configv1.ClusterOperator)
+			if desiredProgressing {
+				if v1helpers.IsStatusConditionTrue(operator.Status.Conditions, configv1.OperatorProgressing) {
+					return true, nil
+				}
+				return false, nil
+			}
+
+			if v1helpers.IsStatusConditionFalse(operator.Status.Conditions, configv1.OperatorProgressing) {
+				return true, nil
```
ibihim, line 117: In the case that the state we want to reach is `progressing=false`, we do a check for `available=true` and `degraded=false` as well.

ibihim: > I don't think the degraded check was doing what you thought it was doing. You need to be sure the level is achieved, but I wouldn't expect a degraded on a two minute momentum to have a meaningful impact for you

We can move forward without a check for degraded, but our working assumption is that tests on a degraded state are done and fail.
The inertia is set to 30 minutes in our case and not 2 minutes.

ibihim: Ok, we might not need a degraded check at all. But we would need to check available=true alongside progressing=false, right?

test/extended/oauth/requestheaders.go
ibihim: [30 minutes](https://github.com/openshift/library-go/blob/master/pkg/operator/apiserver/controllerset/apiservercontrollerset.go#L151-L168=). But anyway, After reading the docs, I tend to agree with you.

It will be interesting to see if there is any change in flakiness afterwards.

PR #27198, opened 2022-05-31
RUN-1933: test unprivileged builds

nalind:
Test builds where we set BUILD_PRIVILEGED=false in the environment to ask the build controller to ask nodes to run builds with fewer privileges.

Verify that builds run with lower privileges by looking for a diagnostic message that the builder logs at log level 2 or higher when it notices that it's running in a user namespace with ID mappings that aren't the node-level default mappings.

Add unprivileged variations to multiple tests that ensure that we can still pass secrets to the builder, and that it can read them properly.

Revise default-privileged cases that assumed the names of builds that now share a BuildConfig with an unprivileged test, since the tests, and their builds, can happen in unexpected sequence.

Depends on https://github.com/openshift/openshift-controller-manager/pull/173.


nalind:
Whoops, forgot to mark this as a WIP.
/retitle WIP: test unprivileged builds

nalind:
Test failures seem to have come down to a missing bit of SELinux policy that would allow us to use kernel overlay, and a kernel-side bug that causes an error when we try to fall back to using fuse-overlayfs.  Removing the WIP label.

coreydaley:
@nalind This needs a rebase.

nalind:
Need to weed out some errors connecting to registries, which were hopefully transient problems.
/retest

PR #27203, opened 2022-06-02
Bug 2093049: test/extended/networking: fix [pod sysctls should not affect node] to use host interfaces

dcbw:
This test was listing all interfaces on the host with 'ls /sys/class/net'
which does not return a stable order of interfaces. Thus it was sometimes
reading host-side veths of pods, which would be gone (because we run
tests in parallel and pods are starting/stopping all the time) by the time
the test wanted to read the interface's sysctls.

```
Jun  2 16:13:38.827: INFO: Error running /usr/bin/oc --namespace=e2e-test-tuning-vz9kr --kubeconfig=/tmp/secret/kubeconfig exec nodeaccess-pod-lmbdb -- cat /host/proc/sys/net/ipv4/conf/090399bc34a8bcf/send_redirects:
StdOut>
cat: can't open '/host/proc/sys/net/ipv4/conf/090399bc34a8bcf/send_redirects': No such file or directory
command terminated with exit code 1
StdErr>
cat: can't open '/host/proc/sys/net/ipv4/conf/090399bc34a8bcf/send_redirects': No such file or directory
command terminated with exit code 1
```

### DIFF
--- a/test/extended/networking/tuning.go
+++ b/test/extended/networking/tuning.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -104,12 +105,15 @@ var _ = g.Describe("[sig-network][Feature:tuning]", func() {
 		})
 
 		g.By("retrieving node interface name to create sysctl pod with interface of the same name")
-		nodeInterfaceNames, err := oc.AsAdmin().Run("exec").Args(nodePod.Name, "--", "ls", "/host/net").Output()
+		linkInfo, err := oc.AsAdmin().Run("exec").Args(nodePod.Name, "--", "bash", "-c", "ip -o link show | grep -v LOOPBACK | head -1").Output()
 		o.Expect(err).NotTo(o.HaveOccurred(), "unable to get interface names")
-		nodeInterfaceName := strings.Fields(nodeInterfaceNames)[0]
-		if nodeInterfaceName == "lo" {
-			nodeInterfaceName = strings.Fields(nodeInterfaceNames)[1]
-		}
+		fields := strings.Split(linkInfo, ": ")
+		o.Expect(len(fields)).To(o.BeNumerically(">", 2), "unexpected ip link output")
+		ifindex, err := strconv.ParseInt(fields[0], 10, 32)
+		o.Expect(err).NotTo(o.HaveOccurred(), "unable to parse interface ifindex")
+		o.Expect(ifindex).To(o.BeNumerically(">", 1), "invalid ifindex")
+		nodeInterfaceName := fields[1]
+		o.Expect(nodeInterfaceName).NotTo(o.BeEmpty(), "unable to find interface name")
 
 		g.By("getting the value of the node sysctl")
 		nodeSysctlValue, err := oc.AsAdmin().Run("exec").Args(nodePod.Name, "--", "cat", "/host"+fmt.Sprintf(sysctlPath, nodeInterfaceName)).Output()


test/extended/networking/tuning.go
```diff
@@ -43,6 +44,53 @@ var whitelistedSysctls = []SysctlVariant{
...
+	// Results will include at least INTERFACE and IFINDEX items for each interface.
+	// Ignore optional DEVTYPE.
+	sanitized := make([]string, 0)
+	for _, line := range strings.Fields(data) {
+		if strings.HasPrefix(line, interfaceTag) || strings.HasPrefix(line, ifindexTag) {
+			sanitized = append(sanitized, line)
+		}
+	}
+	if len(sanitized)%2 != 0 {
+		return "", fmt.Errorf("bad sanitized uevent data length; expected multiple of two got %d", len(sanitized))
```
jluhrsen, line 62: make sense to also output the full "data" here to help with debugging whatever problem lands us here?

test/extended/networking/tuning.go
```diff
@@ -43,6 +44,53 @@ var whitelistedSysctls = []SysctlVariant{
...
+		if err != nil {
+			return "", fmt.Errorf("failed to parse ifindex %q: %v", sanitized[i+1], err)
+		}
+		if (lowestName == "") || (ifindex < lowestIndex) {
+			lowestName = name
+			lowestIndex = ifindex
+		}
+	}
+	if lowestName == "" {
+		return "", fmt.Errorf("no usable host interface found")
```
jluhrsen, line 89: same comment as before. print "data" here too?

test/extended/networking/tuning.go
dcbw: Fixed.

dcbw:
/hold
Need to figure out why the command can't scape uevent files correctly...

test/extended/networking/tuning.go
```diff
@@ -43,6 +44,53 @@ var whitelistedSysctls = []SysctlVariant{
 	// {Sysctl: "net.ipv6.neigh.IFNAME.retrans_time_ms", Value: "1010", Path: "/proc/sys/net/ipv6/neigh/net1/retrans_time_ms"},
 }
 
+// getLowestIfindexInterfaceFromUeventData parses the output of one or more concatenated 'uevent' files
+// from /sys/class/net/*/uevent and returns the name of the interface with the lowest ifindex
+func getLowestIfindexInterfaceFromUeventData(data string) (string, error) {
```
danwinship, line 49: This seems really complicated. Any reason to not just use `ip link`?

Also, the lowest-ifindex interface is probably `lo`, which the test was trying to avoid.

`ip -o link show | grep -v LOOPBACK | head -1 | sed -e 's/:.*//'`

dcbw: Yeah, that was my next step. Wasn't sure if /sbin/ip was even in the container image.

dcbw:
@danwinship results look good with latest /sbin/ip bits:

https://prow.ci.openshift.org/log?container=test&id=1532714882443513856&job=pull-ci-openshift-origin-master-e2e-gcp
```
started: (3/2875/2875) "[sig-network][Feature:tuning] pod sysctls should not affect node [Suite:openshift/conformance/parallel]"
passed: (23s) 2022-06-03T14:57:12 "[sig-network][Feature:tuning] pod sysctls should not affect node [Suite:openshift/conformance/parallel]"
```

danwinship:
you appear to have removed the original `fix [pod sysctls should not affect node] to use host interfaces` commit from the PR...

dcbw:
@danwinship sigh... pushed the wrong branch. fixed now.

deads2k:
Thgis is perm-failing azure. See https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/aggregated-azure-ovn-upgrade-4.11-micro-release-openshift-release-analysis-aggregator/1532975545594155008

PR #27206, opened 2022-06-02
bug 2093051: use the node update finished as a backup completion time for OSStaged

deads2k:
While chasing problems with `[bz-Machine Config Operator] Nodes should reach OSUpdateStaged in a timely fashion` like in https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.11-e2e-aws-upgrade/1532346119462326272, I found that we don't have a backstop value for osupdatestaged.  If we complete the upgrade, then at some point we were OSUpdateStaged even if we didn't observe that event.

/assign @dgoodwin

### DIFF
--- a/pkg/synthetictests/operators.go
+++ b/pkg/synthetictests/operators.go
@@ -94,44 +94,73 @@ func testOperatorOSUpdateStaged(events monitorapi.Intervals, clientConfig *rest.
 	failThreshold := 10 * time.Minute
 
 	// Scan all OSUpdateStarted and OSUpdateStaged events, sort by node.
-	nodeOSUpdateTimes := map[string]*startedStaged{}
+	nodeNameToOSUpdateTimes := map[string]*startedStaged{}
 	for _, e := range events {
-		if strings.Contains(e.Message, "reason/OSUpdateStarted") {
-			// locator will be of the form: node/ci-op-j34hmfqt-253f3-cq852-master-1
-			_, ok := nodeOSUpdateTimes[e.Locator]
-			if !ok {
-				nodeOSUpdateTimes[e.Locator] = &startedStaged{}
-			}
-			// for this type of event, the from/to time are identical as this is a point in time event.
-			ss := nodeOSUpdateTimes[e.Locator]
-			ss.OSUpdateStarted = e.To
-		} else if strings.Contains(e.Message, "reason/OSUpdateStaged") {
-			// locator will be of the form: node/ci-op-j34hmfqt-253f3-cq852-master-1
-			_, ok := nodeOSUpdateTimes[e.Locator]
-			if !ok {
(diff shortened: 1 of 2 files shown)


pkg/synthetictests/operators.go
```diff
@@ -94,44 +94,67 @@ func testOperatorOSUpdateStaged(events monitorapi.Intervals, clientConfig *rest.
...
 			ss.OSUpdateStaged = e.To
+
+		case phase == "Update":
+			_, ok := nodeNameToOSUpdateTimes[nodeName]
+			if !ok {
+				nodeNameToOSUpdateTimes[nodeName] = &startedStaged{}
+			}
+			// This type of event indicates that an update completed. If an update completed, then we know that the latest
+			// possible time that it could have OSUpdateStaged is when the update is finished.  If we have not yet observed
+			// an OSUpdateStaged event, record this time as the final time.
```
dgoodwin, line 134:
```suggestion
			// an OSUpdateStaged event (which indicates we did not receive it likely due to kube API/client issues), record this time as the final time.
```

pkg/synthetictests/operators.go
```diff
@@ -94,44 +94,67 @@ func testOperatorOSUpdateStaged(events monitorapi.Intervals, clientConfig *rest.
...
+			ss := nodeNameToOSUpdateTimes[nodeName]
+			if ss.OSUpdateStaged.IsZero() {
```
dgoodwin, line 136: Thinking we should log a warning here clarifying what we did, and what it likely means.

dgoodwin (commented):
Could you --amend the commit message with some details, I am curious if you have a working theory on what's going on here, I thought we had that code pretty solid, why are we still losing events? 

I'm a little nervous about that scratch build they sent me as this was the only test that failed, though there were indications the cluster was not healthy in the flaked tests. It feels like there's a problem here that maybe only this test is catching. Admittedly this test should not fail if node updates did finish, but do we need another test monitoring api client connections more specifically?

pkg/synthetictests/operators.go
deads2k: > Thinking we should log a warning here clarifying what we did, and what it likely means. 

how?

dgoodwin: Thought just a printf but as discussed on slack, success = no output, so that won't work.

dgoodwin:
Summary from slack. I am concerned at the loss of events and patterns of client errors in the builds I looked at. David believes this is not a concern here and it's expected we lose connection because we lost a node, and nobody will bother looking at data around whether this is flaking or not so any signal on whether or not we fall back to this new method of completing the update is not useful.

/lgtm

PR #27211, opened 2022-06-03
test/extended/prometheus: clean alert exceptions

simonpasquier:
This changes removes all alert exceptions that have been fixed. It
ensures that no regression will happen for these alerts.


jan--f (approved):
/lgtm
Nice, thanks!

jan--f:
I think that's a great idea!

PR #27212, opened 2022-06-03
allow terminating pods to be ready=false

deads2k:
https://github.com/openshift/kubernetes/pull/1281 subdivides the probe error into those that fail before a pod is delete and those that fail after a pod is deleted.

### DIFF
--- a/pkg/synthetictests/duplicated_events.go
+++ b/pkg/synthetictests/duplicated_events.go
@@ -92,12 +92,16 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
 	// kube-apiserver guard probe failing due to kube-apiserver operands getting rolled out
 	// multiple times during the bootstrapping phase of a cluster installation
 	regexp.MustCompile("ns/openshift-kube-apiserver pod/kube-apiserver-guard.*ProbeError Readiness probe error"),
-	regexp.MustCompile("ns/openshift-kube-apiserver pod/kube-apiserver-guard.*Unhealthy Readiness probe failed"),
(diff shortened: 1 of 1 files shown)


pkg/synthetictests/duplicated_events.go
```diff
@@ -92,12 +92,16 @@ var allowedRepeatedEventPatterns = []*regexp.Regexp{
...
 	// multiple times during the bootstrapping phase of a cluster installation
 	regexp.MustCompile("ns/openshift-kube-apiserver pod/kube-apiserver-guard.*ProbeError Readiness probe error"),
-	regexp.MustCompile("ns/openshift-kube-apiserver pod/kube-apiserver-guard.*Unhealthy Readiness probe failed"),
 	// the same thing happens for kube-controller-manager and kube-scheduler
 	regexp.MustCompile("ns/openshift-kube-controller-manager pod/kube-controller-manager-guard.*ProbeError Readiness probe error"),
-	regexp.MustCompile("ns/openshift-kube-controller-manager pod/kube-controller-manager-guard.*Unhealthy Readiness probe failed"),
 	regexp.MustCompile("ns/openshift-kube-scheduler pod/kube-scheduler-guard.*ProbeError Readiness probe error"),
-	regexp.MustCompile("ns/openshift-kube-scheduler pod/kube-scheduler-guard.*Unhealthy Readiness probe failed"),
+
+	// this is the less specific even sent by the kubelet when a probe was executed successfully but returned false
```
martinkennelly, line 99: even -> event

PR #27214, opened 2022-06-03
[WIP] Add test cases to evaluate BestMatching

neisw:
Added test scenarios to validate best matches when disruption data is not present for specified jobtype


vrutkovs:
e2e-aws-single-node-upgrade seems to have matched the new test case, so LGTM

PR #27216, opened 2022-06-06
Bug 2085997: explicit timeout for the etcd scaling test

p0lyn0mial:
on `libvirt` the `etcd is able to vertically scale up and down with a single node` test seems to be interrupted after 15 minutes even though the global timeout for the suite is set (https://github.com/openshift/origin/blob/master/cmd/openshift-tests/e2e.go#L124). This PR sets the timeout explicitly for the test just in case something gets in the way and overwrites the global timeout for the suite.

for example:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-multiarch-master-nightly-4.11-ocp-e2e-serial-remote-libvirt-ppc64le/1533524212742885376


soltysh (approved):
/lgtm
/approve

PR #27217, opened 2022-06-06
Bug 2092961: Use http for network endpoint test.

dgoodwin:
This is tripping about 20% of the time on a tls: internal error.
Trying http to see if that makes it more stable.

### DIFF
--- a/test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
+++ b/test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
@@ -35,7 +35,7 @@ const (
 func NewCIDisruptionWithNewConnectionsTest() upgrades.Test {
 	ciDisruptTest := &ciDisruptionUpgradeTest{}
 	backend := backenddisruption.NewSimpleBackend(
-		"https://static.redhat.com/test/rhel-networkmanager.txt",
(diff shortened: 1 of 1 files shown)


dcbw:
@dgoodwin random question: is it failing on certain platforms more than others? Sometimes TLS stuff is MTU related because the large TLS handshake packets overrun the machine MTU, have to get fragmented, and fragments get lost because Reasons.

dgoodwin:
> @dgoodwin random question: is it failing on certain platforms more than others? Sometimes TLS stuff is MTU related because the large TLS handshake packets overrun the machine MTU, have to get fragmented, and fragments get lost because Reasons.

There are some 10-20% differences in some variant combos, but in terms of just platforms it's pretty consistent across everything. Worst core platform is "aws, amd64, upgrade, upgrade-micro, sdn, ha" at 75%, best is "azure, amd64, upgrade, upgrade-minor, sdn, ha" at 97%, everything is in between those. Some Azure's are near the worst numbers as well, some aws near the best. 

I was actually able to replicate the problem here locally running the disrupt sampler on my machine all day for both https and http, I saw the tls errors mid-day for https, http remained happy.

To me it looks like it's an issue with the endpoint itself.

PR #27181, opened 2022-05-26
Bug 2086092: update kube to v1.24.0

tkashem:

PR #27218, opened 2022-06-06
Add tooling to track Timeout tags

soltysh:
Requires https://github.com/openshift/kubernetes/pull/1284 and https://github.com/openshift/origin/pull/27181 to land
/hold
/assign @deads2k

### DIFF
--- a/test/extended/util/annotate/generated/timeout/OWNERS
+++ b/test/extended/util/annotate/generated/timeout/OWNERS
@@ -1,0 +1,6 @@
+approvers:
+- deads2k
+- soltysh
+
+options:
+  no_parent_owners: true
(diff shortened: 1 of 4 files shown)


test/extended/util/annotate/generated/timeout/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+- deads2k
+- soltysh
+
+options:
+  no_parent_owners: true
```
soltysh, line 6: @deads2k this will ensure that only the approvers from this file can approve changes to this directory, which is what you wanted.

deads2k: > @deads2k this will ensure that only the approvers from this file can approve changes to this directory, which is what you wanted.

Yeah, this approach looks good.  Do you have a general reviewer to watch these?  I might suggest @stbenjam.  He'll probably need some background, which I'd suggest placing in a comment at the top of the generated file or in a readme.md.

soltysh:
/hold 
wrong PR

PR #27220, opened 2022-06-06
Bug 2085997: Report late requests verification: Fix for Hypershift

alvaroaleman:
The late requests validation is based on interpreting the KAS audit
logs. The audit logs are fetched through `oc must-gather`. This works
fine for self-hosted OCP, but with an external controlplane the KAS runs
as pod in a distinct management cluster and thus it is not possible to
get its audit logs from the cluster under test.

This change fixes the test for Hypershift by introduction dedicated
audit log fetching codepaths for selfhosted/hypershift.

/assign @p0lyn0mial 

The diff gets much smaller when omitting whitespace changes.

### DIFF
--- a/test/extended/apiserver/graceful_termination.go
+++ b/test/extended/apiserver/graceful_termination.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path"
@@ -18,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/origin/pkg/test/ginkgo/result"
 	clihelpers "github.com/openshift/origin/test/extended/cli"
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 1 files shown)


geliu2016:
@wangke19 could you please take a view?

test/extended/apiserver/graceful_termination.go
```diff
@@ -143,84 +145,109 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Late]", func() {
...
+		logReaders := map[string]io.Reader{}
+		results := map[string]int{}
 
-				file, err := os.Open(path)
+		if *controlPlaneTopology == configv1.ExternalTopologyMode {
+			mgmtClusterOC := exutil.NewHypershiftManagementCLI("default").AsAdmin().WithoutNamespace()
+			pods, err := mgmtClusterOC.KubeClient().CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{LabelSelector: "hypershift.openshift.io/control-plane-component=kube-apiserver"})
+			o.Expect(err).To(o.BeNil())
+			for _, pod := range pods.Items {
+				fileName, err := mgmtClusterOC.Run("logs", "-n", pod.Namespace, pod.Name, "-c", "audit-logs").OutputToFile(pod.Name + "-audit.log")
```
p0lyn0mial, line 161: I think it only gets the latest logs, during e2e tests APIs might be restarted multiple times. The intention of this test was to validate that during a termination we didn't receive late requests. Checking all available logs increases our chances to detect incorrect behavior.

I think we would have to list all available logs and retrieve them individually. 

I am okay with what we have now we could just improve it in the future.

p0lyn0mial:
/assign @soltysh 

for approval

soltysh (approved):
/approve

test/extended/apiserver/graceful_termination.go
alvaroaleman: >  APIs might be restarted multiple times. 

Why would the api get restarted during e2e tests?

p0lyn0mial: some tests do that, i.e. `etcd is able to vertically scale up and down with a single node`

alvaroaleman:
Looks like the test works, it passed on https://github.com/openshift/origin/pull/27236
/retest

PR #27221, opened 2022-06-06
Bug 2093049: test/extended/networking: fix [pod sysctls should not affect node] to use known host interfaces

dcbw:
This test was listing all interfaces on the host with 'ls /sys/class/net'
which does not return a stable order of interfaces. Thus it was sometimes
reading host-side veths of pods, which would be gone (because we run
tests in parallel and pods are starting/stopping all the time) by the time
the test wanted to read the interface's sysctls.

```
Jun  2 16:13:38.827: INFO: Error running /usr/bin/oc --namespace=e2e-test-tuning-vz9kr --kubeconfig=/tmp/secret/kubeconfig exec nodeaccess-pod-lmbdb -- cat /host/proc/sys/net/ipv4/conf/090399bc34a8bcf/send_redirects:
StdOut>
cat: can't open '/host/proc/sys/net/ipv4/conf/090399bc34a8bcf/send_redirects': No such file or directory
command terminated with exit code 1
StdErr>
cat: can't open '/host/proc/sys/net/ipv4/conf/090399bc34a8bcf/send_redirects': No such file or directory
command terminated with exit code 1
```

Instead of grabbing random interfaces from the host that may or may not
exist when the rest of the test runs, make the test own all the interfaces
it uses by creating it's own host interface via a second Network Attachment
Definition.

Rework of https://github.com/openshift/origin/pull/27203 to fix instances where the host interface is named `eth0` like the default pod interface.

@danwinship

### DIFF
--- a/test/extended/networking/tuning.go
+++ b/test/extended/networking/tuning.go
@@ -9,6 +9,8 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 	kapiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
 	"strings"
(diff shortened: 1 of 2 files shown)


test/extended/networking/tuning.go
```diff
@@ -43,6 +45,14 @@ var whitelistedSysctls = []SysctlVariant{
 	// {Sysctl: "net.ipv6.neigh.IFNAME.retrans_time_ms", Value: "1010", Path: "/proc/sys/net/ipv6/neigh/net1/retrans_time_ms"},
 }
 
+// getPodNodeName returns the name of the node the pod is scheduled on
+func getPodNodeName(client clientset.Interface, namespace, name string) string {
+	pod, err := client.CoreV1().Pods(namespace).Get(context.TODO(), name, metav1.GetOptions{})
```
danwinship, line 50: it's not enough to just immediately re-retrieve; you have to wait for the pod to get scheduled by kube-scheduler, which might take longer-than-it-takes-you-to-re-retrieve-once.

there ought to be a helper for this...

actually, now that I think about it, the usual trick is to pick a node beforehand with `e2enode.GetRandomReadySchedulableNode()` and then set `pod.Spec.NodeName` on the first pod too

dcbw: > it's not enough to just immediately re-retrieve; you have to wait for the pod to get scheduled by kube-scheduler, which might take longer-than-it-takes-you-to-re-retrieve-once.

@danwinship this is actually getting executed after CreateExecPodOrFail() which always waits for the pod to be running.

danwinship:
ah
/lgtm

PR #27223, opened 2022-06-07
Revert "Bug 2079679: pkg/monitor: wait for Prometheus sidecars to be ready"

simonpasquier:
Reverts openshift/origin#27201 because it fails for single node jobs

image


dgoodwin:
/lgtm

Will find an approver, thanks for the quick turnaround!

PR #27225, opened 2022-06-07
Bug 2094342: bump cluster-capi-operator apirequests upper bound to 50

Fedosin:
When deploying OCP with enabled CAPI, we got this error from the tests:

"Operator \"cluster-capi-operator\" produces more watch requests than expected: watchrequestcount=77, upperbound=72, ratio=1.0694444444444444""

Bumping the upperbound to 50 should solve this issue.


alexander-demichev (approved):
/lgtm

PR #27235, opened 2022-06-08
Bug 2086092: update kube to v1.24.0

tkashem:


soltysh:
Replaced with #27244 
/close

PR #27237, opened 2022-06-09
Disable internalTrafficPolicy=Local service tests for cilium

pperiyasamy:
The Cilium CNI plugin currently does not support services with
internal traffic local policy, so skipping such relevant tests
running against environments configured with cilium as primary
network plugin.

For more information:

https://github.com/cilium/cilium/issues/17796
https://github.com/cilium/cilium/pull/17828


deads2k:
I'm not sure if this is a bug.  How does a user know this feature isn't working in their cluster?

/assign @danwinship

danwinship:
> I'm not sure if this is a bug. How does a user know this feature isn't working in their cluster?

Because they read the Cilium documentation and wouldn't have chosen to use Cilium if they needed this feature. ("But the Cilium documentation doesn't mention that they don't support this!" :man_shrugging:)

Currently third-party CNI plugin certification in OCP requires passing all networking `Conformance` tests, but we don't check non-conformance stuff, other than NetworkPolicy, which we allow them to fail some tests in, as long as they document it in their "using with openshift" docs. (We probably ought to do that with all non-conformance networking tests...)

FTR, Cilium is currently only certified against OCP 4.6 and 4.7, which are pre-`internalTrafficPolicy` anyway so this doesn't even apply to any version of Cilium+OCP that customers could currently be using.

danwinship:
@pperiyasamy was going to investigate what the purpose of this CI job is, since it's hard to say whether disabling the test is appropriate or not without knowing why the job is being run

/hold

jerpeter1:
> what the purpose of this CI job is, since it's hard to say whether disabling the test is appropriate or not without knowing why the job is being run

It looks like the job was added in [this PR](https://github.com/openshift/release/pull/17380). As described, the periodic jobs  were added to test compatibility of cilium on openshift on aws, azure & gcp.

pperiyasamy:
@danwinship as Jeremy pointed this CI job is added to test compatibility of cilium with OCP cluster. The cilium still skips running [internalTrafficPolicy=Local](https://github.com/cilium/cilium/blob/master/test/kubernetes-test.sh#L110-L111) test cases, i think we could also skip such test cases.  So are we good to review and merge this ?

test/extended/util/annotate/rules.go
```diff
@@ -304,6 +304,11 @@ var (
 			`Unidling should work with TCP`,
 			`Unidling should handle many TCP connections`,
 		},
+		// tests that don't pass under Cilium
+		"[Skipped:Network/Cilium]": {
```
danwinship, line 308: This needs to be clearer about why this section is here:
- We do not support Cilium, but we run e2e tests against it because ... why?
  - (Including "why don't we run e2e tests against other third-party plugins?")
- More info about the e2e Cilium test runs can be found at ... ?
- There are no links to BZ/Jira for the test skips here because we do not intend to fix these bugs ourselves because we don't support this configuration. (But, for some reason, we still need to run e2e tests against it?)

test/extended/util/annotate/rules.go
```diff
@@ -304,6 +304,11 @@ var (
...
+			// Cilium does not support services with internal traffic local policy
+			`internalTrafficPolicy=Local`,
```
danwinship, line 310: This should have a link to the relevant Cilium issue so people in the future can know whether this is still needed.

danwinship (changes requested):

test/extended/util/annotate/rules.go
pperiyasamy: as per this documentation https://docs.cilium.io/en/v1.12/gettingstarted/k8s-install-openshift-okd/ and https://isovalent.com/blog/post/2022-03-openshift and OCP still supports cilium as third party cni solution and this is evident from [4.11 CI](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-e2e-aws-cilium) runs as well.
We don't have BZ/Jira issue for this as it comes under CI stabilization work which is found while investigating [OCPVE-133](https://issues.redhat.com/browse/OCPVE-133).

test/extended/util/annotate/rules.go
pperiyasamy: done

test/extended/util/annotate/rules.go
danwinship: Being certified to run with OpenShift does not mean "_Red Hat_ tests that this plugin works with OpenShift", it means "_the vendor of this plugin_ tests that (specific versions of) this plugin work with (specific versions of) OpenShift" (plus stuff with support contracts and stuff).

> OCP still supports cilium as third party cni solution and this is evident from 4.11 CI runs as well.

In fact, no version of Cilium is certified against OCP 4.11. (https://access.redhat.com/articles/5436171). At any rate, if it was, that would be entirely independent of this CI job. Third party CNI plugin certification is done by having the third-party CNI plugin vendor run the e2e tests themselves and sending us the results.

> We don't have BZ/Jira issue for this as it comes under CI stabilization work which is found while investigating [OCPVE-133](https://issues.redhat.com/browse/OCPVE-133).

If there is no good reason for this job to exist, then the way to improve overall CI pass rate would be to just remove that job (or at least, remove it from the set of jobs that are counted in the overall pass rate.)

pperiyasamy: @danwinship Thanks for the info, If OpenShift doesn't intend to run its e2e tests anymore with Cilium, then we can just remove that job from running. Let me work on a separate PR for this.

pperiyasamy:
Closing the PR as per [this](https://github.com/openshift/origin/pull/27237#discussion_r957410327) comment.

PR #27234, opened 2022-06-08
Bug 2086086: update samples for 4.11

yselkowitz:
The image_ecosystem additions depend on https://github.com/openshift/cluster-samples-operator/pull/428

/cc @fbm3307

PR #27241, opened 2022-06-10
Add David Peraza as approver on samples test content

bparees:


bparees:
@coreydaley @dperaza4dustbit ptal, i'm doing this because of https://github.com/openshift/origin/pull/27234 which i think @dperaza4dustbit should probably have been the approver for.

bparees:
@coreydaley @dperaza4dustbit bump

PR #27244, opened 2022-06-10
Bug 2086092: update kube to v1.24.0

ravisantoshgudimetla:
Picking up from https://github.com/openshift/origin/pull/27235


ravisantoshgudimetla:
/payload 4.11 ci blocking
/payload 4.11 ci informing
/payload 4.11 nightly blocking
/payload 4.11 nightly informing

Not sure if it is going to stress the build farm.

ravisantoshgudimetla:
Most of the test failures are related to some file not being downloaded from github. I think running multiple tests at the same time is causing throttling at github level.

soltysh (approved):
We got green on ci blocking and nightly blocking payload
/lgtm
/approve

PR #27251, opened 2022-06-14
Request Headers: Change wait logic, add more debugging data to post mortem

ibihim:
## What

- Reuse util solution, rather than custom logic
- Add more information to state-of-the-cluster during test run and especially on failure

## Why

We are trying to identify the reason for the flakiness of the request headers idp test suite.

## Ref

https://github.com/openshift/origin/pull/27194

### DIFF
--- a/test/extended/oauth/requestheaders.go
+++ b/test/extended/oauth/requestheaders.go
@@ -56,14 +56,7 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 	var oc = exutil.NewCLI("request-headers")
 
 	g.It("test RequestHeaders IdP", func() {
-
-		// In some rare cases, CAO might be damaged when entering this test. If it is - the results
-		// of this test might flaky. This check ensures that we capture such situation early and
-		// investigate why it wasn't ready before this test.
-		e2e.Logf("Ensuring CAO is available==True, progressing==False, degraded==False")
-
-		ctx := context.Background()
-		ctx, cancel := context.WithTimeout(ctx, 20*time.Minute)
+		ctx, cancel := context.WithTimeout(context.Background(), 20*time.Minute)
 		defer cancel()
 
 		authenticationRollout := exutil.WaitForOperatorToRollout(ctx, oc.AdminConfigClient(), "authentication")
@@ -107,8 +100,7 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 		o.Expect(err).NotTo(o.HaveOccurred())
 		// clean up after ourselves
 		defer func() {
-			ctx := context.Background()
-			ctx, cancel := context.WithTimeout(ctx, 20*time.Minute)
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Minute)
 			defer cancel()
 
 			authenticationRollout := exutil.WaitForOperatorToRollout(ctx, oc.AdminConfigClient(), "authentication")
--- a/test/extended/oauth/requestheaders.go
+++ b/test/extended/oauth/requestheaders.go
@@ -56,18 +56,18 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 	var oc = exutil.NewCLI("request-headers")
 
 	g.It("test RequestHeaders IdP", func() {
-		ctx, cancel := context.WithTimeout(context.Background(), 20*time.Minute)
-		defer cancel()
-
-		authenticationRollout := exutil.WaitForOperatorToRollout(ctx, oc.AdminConfigClient(), "authentication")
-		<-authenticationRollout.StableBeforeStarting() // wait for the initial state to be stable
-
 		controlPlaneTopology, err := exutil.GetControlPlaneTopology(oc)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		if *controlPlaneTopology == configv1.ExternalTopologyMode {
 			e2eskipper.Skipf("External clusters do not allow customization of the Identity Providers for the cluster.")
 		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 20*time.Minute)
+		defer cancel()
+
+		authenticationRollout := exutil.WaitForOperatorToRollout(ctx, oc.AdminConfigClient(), "authentication")
+		<-authenticationRollout.StableBeforeStarting() // wait for the initial state to be stable
 
 		caCert, caKey := createClientCA(oc.AdminKubeClient().CoreV1())
 		defer oc.AdminKubeClient().CoreV1().ConfigMaps("openshift-config").Delete(context.Background(), clientCAName, metav1.DeleteOptions{})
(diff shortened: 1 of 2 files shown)


test/extended/oauth/requestheaders.go
```diff
@@ -288,6 +291,57 @@ func gatherPostMortem(oc *exutil.CLI) {
 		}
 		e2e.Logf("log from %s in %s: %s", pod.Name, oauthServerNamespace, logs)
 	}
+
+	operator, err := oc.AdminOperatorClient().
```
stlaz, line 295: the clusteroperator state is gathered in an aggregated form above, I don't think this is necessary

test/extended/oauth/requestheaders.go
```diff
@@ -107,6 +105,13 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 		o.Expect(err).NotTo(o.HaveOccurred())
 		// clean up after ourselves
 		defer func() {
+			g.By("clean up - wait for stability")
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Minute)
+			defer cancel()
+			authenticationRollout := exutil.WaitForOperatorToRollout(ctx, oc.AdminConfigClient(), "authentication")
```
stlaz, line 111: I don't think we need to wait here, the next test may not care about the authn operator and so we're just extending the test duration

test/extended/oauth/requestheaders.go
```diff
@@ -504,32 +558,3 @@ func getTokenFromResponse(resp *http.Response) string {
...
-		progressing := clusteroperatorhelpers.FindStatusCondition(authn.Status.Conditions, configv1.OperatorProgressing)
-		if progressing == nil || progressing.Status != expectedProgressing {
-			e2e.Logf("Waiting for progressing condition to be %q: %s", expectedProgressing, spew.Sdump(authn.Status.Conditions))
-			return false, nil
-		}
-
-		if expectedProgressing == configv1.ConditionFalse {
-			// make additional checks on availability and degraded status
-			if clusteroperatorhelpers.IsStatusConditionFalse(authn.Status.Conditions, configv1.OperatorAvailable) ||
-				clusteroperatorhelpers.IsStatusConditionTrue(authn.Status.Conditions, configv1.OperatorDegraded) {
```
stlaz, line 525: These changes are not going to improve the failure rate of the test in the referenced BZ. The problem is that the lib-go deployment logic sets Degraded=true when the deployment is taking too long but the Degraded inertia only sets the message.

Since the failure rate is in fact caused by a premature testing when the new config did not roll out yet because the test was also waiting for degraded=False, this does not help us.

test/extended/oauth/requestheaders.go
ibihim: But what if the next test cares? :smile: 

So basically your assumption is that:
- the test doesn't care
- or the test waits for `Progressing=False`?

stlaz: If the test cares then it needs to handle it itself

test/extended/oauth/requestheaders.go
stlaz: The hypothesis of not checking Degraded is wrong in this case.

You can merge these changes because I think they are good but you will have to remove the Bugzilla reference because the patch isn't fixing the issue.

test/extended/oauth/requestheaders.go
stlaz: The inertia logic only drops the "True" from Degraded state but AFAIK the messages are kept and aggregated. That being said, I think I wouldn't be against logging this.

However, too much information might also be confusing for the reader, so I would rather be a little careful about what we're posting here.

Btw - the location of these logs is wrong. It should be:
```
clusteroperator state
operator state
deployment
configmap
pods
```
==> move this to the second place

stlaz (changes requested):

test/extended/oauth/requestheaders.go
ibihim: Yes, I know that the message stays. This is why you came up with the working hypothesis, that something is in the bushes. I wanted to make it more explicit who or what is in a bad shape. The messages are very cryptic, if torn out of context.

test/extended/oauth/requestheaders.go
ibihim: Yes, this is a nice speed increase due to smart logic, but in the state of currently flaky tests, I would prefer dumb logic. 

If we have an attitude to leave the next test to handle the fallout of our tests side effects, it is hard to identify who left the mess at the end.

test/extended/oauth/requestheaders.go
ibihim: Yes, I could remove the reference to the BZ as it indeed, might not fix it and the changes themselves are not focused on it.

My general question would be, do you think we should add an availability check or later?

Why do you think it is wrong? If the deployment takes too long, progressing is true, right? Why do we need the additional signal of Degraded?

test/extended/oauth/requestheaders.go
stlaz: Like I said, I want this code removed.

test/extended/oauth/requestheaders.go
stlaz: > My general question would be, do you think we should add an availability check or later?
which availability check?

> Why do you think it is wrong? If the deployment takes too long, progressing is true, right?
no, progressing is actually false, degraded should be true but the inertia only causes the message to appear

test/extended/oauth/requestheaders.go
ibihim: hm kay

test/extended/oauth/requestheaders.go
ibihim: This is important to clarify and I think the root of our discussion:

Why do you think we will have a
- `Progressing=False` and `Degraded=True`, 
- but due to `inertia`, we see a `Degraded=False`?

If we can clarify the question, we could add a real `Degraded` check that looks into the `Operator` `Conditions` like it is in the post mortem.

With availability check I mean `Available=True`, as the state could be `Progressing=False`, but `Available=False` too.

stlaz: > Progressing=False and Degraded=True,

Because that's how the code works today, like I mentioned in our Slack discussions, one way to go around this would be to have both be set to true.

> If we can clarify the question, we could add a real Degraded check that looks into the Operator Conditions like it is in the post mortem.

Indeed we could.

> With availability check I mean Available=True, as the state could be Progressing=False, but Available=False too.

Since there are still workloads able to serve, Availability never changes to "false" in this case.

ibihim: Ok, thx for the discussion. There is another PR which should prevent a situation where a progressing = true operator is set to progressing = false as in combination with the inertia logic, it causes a weird limbo state.

[ref](https://github.com/openshift/library-go/pull/1380)

ibihim: Thx for your patience and code pointers that!

deads2k:
is this blocked on https://github.com/openshift/cluster-authentication-operator/pull/576?

ibihim:
@deads2k, this is not blocked on it. It adds your changes and detailed logging on the assumed issue.

PR #27253, opened 2022-06-14
Update to go 1.18

yselkowitz:


yselkowitz:
this will collide with the kube rebase
/hold

PR #27256, opened 2022-06-15
Poll disruption backends every 5s instead of 1s.

dgoodwin:
Possibly just a temporary change, but we're attempting to determine if
we're overloading DNS on the build clusters.

5x is chosen because we want to see a dramatic impact if there will be
one. This implies we'll miss more single second disruptions, but if we
find one we'll register 5s as much. Data will be less accurate but
perhaps more sustainable to investigate.


dgoodwin:
Ended up not being required, and not desirable long term either as we'll miss too much possible disruption.

/close

PR #27257, opened 2022-06-15
Bug 2086519: authorization: add a test checking that the restricted-v2 SCC mutates securityContext to match restricted PSa

stlaz:
This is a test to check that our SCC system by default mutates container and pod security context so that they can run with the restricted security profile

/assign @deads2k

### DIFF
--- a/test/extended/authorization/podsecurity_admission.go
+++ b/test/extended/authorization/podsecurity_admission.go
@@ -1,0 +1,39 @@
+package authorization
+
+import (
+	"context"
+
+	g "github.com/onsi/ginkgo"
+	o "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


test/extended/authorization/podsecurity_admission.go
```diff
@@ -0,0 +1,39 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "psa-testpod",
+			},
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:    "sleeper",
+						Image:   "fedora:latest",
+						Command: []string{"sleep"},
+						Args:    []string{"infinity"},
```
ibihim, line 31: We need to clean it up, don't we?

stlaz: no, the NS is removed after the test is complete

ibihim (approved):
Test Machinery removes namespaces, interesting.

/lgtm

soltysh (approved):
/approve

PR #27260, opened 2022-06-16
Bug 2098194: set sampleErr to bodyReadErr

neisw:
Believe the intention was to set bodyReadErr  in the bodyReadErr switch case


dgoodwin:
Looks like what it was intended to be to me.

/lgtm

PR #27261, opened 2022-06-16
Bug 2097297: Show DNS lookup sampler problems on the disruption intervals chart.

dgoodwin:
In #27238 we began treating DNS lookup errors as disruption warnings not
errors, as they indicate a problem in the CI cluster not the cluster
under test.

I confirmed warnings would show in the graph, but did not realize this
was based on the message. When the message later was changed for these
warnings, they no longer appeared.

Update the message logic to use a slightly safer "reason" my original PR
added, and include the new reason.

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -147,7 +147,7 @@
     }
 
     function isEndpointConnectivity(eventInterval) {
-        if (!eventInterval.message.includes("stopped responding to GET requests")){
+        if (!eventInterval.message.includes("reason/DisruptionBegan") && !eventInterval.message.includes("reason/DisruptionSamplerOutageBegan")){
             return false
         }
         if (eventInterval.locator.includes("disruption/")) {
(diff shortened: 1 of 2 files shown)


dgoodwin:
Can't get the damn thing to show up in here.

/retest

dgoodwin:
Can't reproduce in the PR, going to have to let it go to the wild. As before, nothing really broken, all we're picking up is real disruption and it's getting graphed in these pr runs.

DennisPeriquet:
you have to generate the bindata.go.  I think just do this: `make update` and it will generate it and then you can add it to the PR.  I think `verify` test will fail if you don't do that.

DennisPeriquet:
looking at [the Makefile](https://github.com/openshift/origin//blob/8bf408ce78279a227854b03534bdebeb5b2a0bff/Makefile#L54), `make update-bindata` will work too.

dgoodwin:
Thanks Dennis! Pr updated.

PR #27266, opened 2022-06-19
examples: drop OCP 3 library paths

yselkowitz:
These paths were made arch-specific in preparation for the addition of ppc64le back in 3.x, but the way samples ended up being implemented, there is no actual difference in these files.

This is needed to decommission the OCP3 arch paths in library.

Keeping as draft until 4.12 development opens.


yselkowitz:
This is a prerequisite of dropping OCP 3 paths from library
/assign @dperaza4dustbit

PR #27269, opened 2022-06-21
Bug 2086519: AUTH-133: test/extended/security/scc: comply to pod security admission 

stlaz:
Obsoletes #27154 

/assign @ibihim

### DIFF
--- a/test/extended/builds/build_pruning.go
+++ b/test/extended/builds/build_pruning.go
@@ -11,8 +11,10 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	admissionapi "k8s.io/pod-security-admission/api"
 
 	buildv1 "github.com/openshift/api/build/v1"
+
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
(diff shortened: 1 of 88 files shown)


test/extended/builds/build_pruning.go
```diff
@@ -27,7 +29,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds] prune builds based on settings
 		failedBuildConfig     = filepath.Join(buildPruningBaseDir, "failed-build-config.yaml")
 		erroredBuildConfig    = filepath.Join(buildPruningBaseDir, "errored-build-config.yaml")
 		groupBuildConfig      = filepath.Join(buildPruningBaseDir, "default-group-build-config.yaml")
-		oc                    = exutil.NewCLI("build-pruning")
+		oc                    = exutil.NewCLIWithPodSecurityLevel("build-pruning", admissionapi.LevelBaseline)
```
ibihim, line 32: In general, I am wondering if we need to expose the complexity of PodSecurityLevel to everyone. Maybe we could add `admissionapi.LevelBaseline` to every `exutil.NewCLI` by default and only set `exutil.NewCLIWithPodSecurityLevel` in the case of `admissionapi.LevelRestricted` or `admissionapi.LevelPrivileged`.

s-urbaniak: The default must be the most secure one (restricted). only once your e2e workload needs broader permissions you have to make a conscious choice (baseline or restricted). This behaviour also aligns with upstream.

PR #27270, opened 2022-06-22
Bug 2100091: Ignore FailedCreatePodSandBox events for garbage collector tests

pacevedom:
Garbage collector tests dont care about pod being available/running, they are only testing owner references. If CNI is slow to respond the pod is deleted by the time it needs it, rendering warning events for a failure to create pod sandbox.
This PR removes checks for such events, as they might happen even when the test succeeds.


dgoodwin:
/lgtm
/assign @stbenjam 

This looks good, hopefully it will help cleanup the signal coming out of this historically problematic test. Thanks!

PR #27272, opened 2022-06-22
Bug 2100439: Enable GCP PD in-tree tests

bertinatto:
Those tests have been disabled by default in k8s 1.24. See: https://github.com/kubernetes/kubernetes/pull/109541

CC @openshift/storage


Makefile
```diff
@@ -1,6 +1,11 @@
 all: build
 .PHONY: all
 
+# Since k8s 1.24 this is required to enable GCP PD tests.
+# Here it's mostly used by update targets. See:
+# https://github.com/kubernetes/kubernetes/pull/109541
+export ENABLE_STORAGE_GCE_PD_DRIVER := yes
```
bertinatto, line 7: This doesn't look like the right place for setting this env var, please advise.

bertinatto: CC @bparees 
Any ideas?

bparees: you can set it here, to put it a bit closer to where it matters:

https://github.com/openshift/origin/blob/6438b04e175f9e2ce434b7deb4b786f602eb886b/hack/update-generated.sh#L6

e.g.:
```
ENABLE_STORAGE_GCE_PD_DRIVER=yes go generate -mod vendor ./test/extended
```
should do it.

the bigger problem is you also need to set this in the actual CI jobs where you want this test to run.

hack/update-generated.sh
```diff
@@ -3,7 +3,7 @@ STARTTIME=$(date +%s)
 source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
 # Update test names
-go generate -mod vendor ./test/extended
+ENABLE_STORAGE_GCE_PD_DRIVER=yes go generate -mod vendor ./test/extended
```
bparees, line 6: include a comment/link about why this is here.

bertinatto: Added comment back.

PR #27274, opened 2022-06-24
BUG 2067323: test/extended/router: Drop host lookup for gRPC HTTP/2 h2spec tests

frobware:
Each of the tests go on to call Dial() or http.Get() and retry on
failure. By removing host lookup we also remove the association that
there's only 1 IP address returned--and that was failing for
Azure/UPI. For example:

Apr 18 20:21:37.899: INFO: host
   "http2-custom-cert-edge.e2e-test-router-http2-25wkr.apps.ci-op-00l27v7z-c3806.ci.azure.devcluster.openshift.com"
 resolves as
   [52.234.42.30]
 expecting 13.64.142.91, retrying in 1m0s...


frobware:
/hold

Need to see if this:
- breaks existing jobs and/or
- fixes Azure/UPI

PR #27275, opened 2022-06-24
WIP: try to get resourceMaps later in upgrade

DennisPeriquet:
I believe pod resource maps are saved by this time when we look at alerts.  Let's prove or disprove it.


DennisPeriquet:
no longer needed

PR #27281, opened 2022-06-27
Tune watch limits for operators

dgoodwin:
[TRT-347](https://issues.redhat.com//browse/TRT-347)

These looks to be failing a few times a week, though at a very low rate. We're in a tough spot here because no one is motivated to fix small discrepancies here.

These values were taken one year ago and measured, then given 10%. We exceed that sometimes now and the values need tuning to keep the test passing.

A better approach here might be to output a json file, ingest into one of our databases, and track against real historical data to monitor for the tests intended purpose and watch for exponential growth.

I've taken an attempted pass at finding new value, original methodology was very complex and time consuming. Instead I followed the process from a recent bug that went through this and just used search.ci to find failures, and adjust the levels to accommodate them.


dgoodwin:
Updated for a conflict. I think we should proceed with this, the vsphere bump is particularly important. We have a card for a better system for this long term.

PR #27285, opened 2022-07-02
WIP: Make chart with just image-registry,console,oauth

DennisPeriquet:
My goal: make it so that we can have a single chart with all these items in it (that tend to have disruption problems together) -- specifically, image-registry, console, oauth, router-default pods.

If this goes well, I will try to add in disruption events and node events so I don't have to make the charts manually.


DennisPeriquet:
closed in favor of a more general solution: https://github.com/openshift/origin/pull/27297

PR #27295, opened 2022-07-08
etcd storage tests: work with dynamic namespace names

stlaz:
Etcd tests were using a static namespace name and were creating
a namespace with custom logic different but similar to the namespace
setup logic used everywhere else.

Move etcd storage tests to using the namespace created by the
machinery.


soltysh (approved):
/lgtm
/approve

PR #27299, opened 2022-07-11
Convert kubelet logs into monitor events for timelines

deads2k:
kubelet logs contain the authoritative list of container status changes.  We're particularly interested in readiness at this moment, but there is general utility in this information, particularly for CNI integration information.

/hold

holding to check the data gathered and visualization of it.


deads2k:
looks to be working well.  

/hold cancel

pkg/monitor/nodedetails/node_log.go
```diff
@@ -0,0 +1,66 @@
...
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	"k8s.io/client-go/kubernetes"
+)
+
+// GetNodeLog returns logs for a particular systemd service on a given node.
+// We're count on these logs to fit into some reasonable memory size.
```
xueqzhan, line 16: Those logs are pretty big. Also, you added the MonitorEventsOptions to each run-test command. I believe a full e2e suite might spawn lots of run-tests. Will that be a problem?

deads2k:
/hold

this causes OOMKills (see https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27299/pull-ci-openshift-origin-master-e2e-gcp/1549234180578086912).  I think we'll have to find a way to stream through.

Can I get a review on the PR and I'll add a commit for the streaming separately?

deads2k:
ooms are gone

/hold cancel

pkg/monitor/intervalcreation/node.go
```diff
@@ -100,3 +114,185 @@ func IntervalsFromEvents_NodeChanges(events monitorapi.Intervals, _ monitorapi.R
...
+		return nil
+	}
+
+	failureOutputRegex.MatchString(logLine)
+	if !failureOutputRegex.MatchString(logLine) {
+		return nil
+	}
+	outputSubmatches := failureOutputRegex.FindStringSubmatch(logLine)
+	message := outputSubmatches[1]
+	message, _ = strconv.Unquote(`"` + message + `"`)
```
dgoodwin, line 189: Could use a comment, looks like you're quoting and then unquoting. Also ignored error.

pkg/monitor/intervalcreation/node.go
```diff
@@ -100,3 +114,185 @@ func IntervalsFromEvents_NodeChanges(events monitorapi.Intervals, _ monitorapi.R
...
+	ret := monitorapi.Intervals{}
+
+	scanner := bufio.NewScanner(bytes.NewBuffer(kubeletLog))
+	for scanner.Scan() {
+		currLine := scanner.Text()
+		ret = append(ret, readinessFailure(currLine)...)
+		ret = append(ret, readinessError(currLine)...)
+	}
+
+	return ret
```
dgoodwin, line 168: Suspicion here that there are a few things that could go wrong that are getting glazed over in this function and below. Should we handle errors here, at least outputting something that we could find in search.ci if suddenly we couldn't parse these the way we'd expect? (I'm ok with it just proceeding, just looking for something in build log we can find)

pkg/monitor/intervalcreation/node.go
```diff
@@ -100,3 +114,185 @@ func IntervalsFromEvents_NodeChanges(events monitorapi.Intervals, _ monitorapi.R
...
+}
+
+var kubeletTimeRegex = regexp.MustCompile(`^(?P<MONTH>\S+)\s(?P<DAY>\S+)\s(?P<TIME>\S+)`)
+
+func kubeletLogTime(logLine string) time.Time {
+	kubeletTimeRegex.MatchString(logLine)
+	if !kubeletTimeRegex.MatchString(logLine) {
+		return time.Now()
```
dgoodwin, line 271: Per below this confuses me, why would we assume time.Now(), wouldn't that mess up intervals chart?

pkg/monitor/monitoring_reflector_store.go
```diff
@@ -135,8 +135,8 @@ func newMonitoringStore(
 
 		for _, obj := range items {
 			currentUID := uidOf(obj)
-			metadata, _ := meta.Accessor(obj)
-			framework.Logf("#### replacing uid/%v pod/%v ns/%v resourceVersion/%v\n", currentUID, metadata.GetName(), metadata.GetNamespace(), metadata.GetResourceVersion())
+			//metadata, _ := meta.Accessor(obj)
+			//framework.Logf("#### replacing uid/%v pod/%v ns/%v resourceVersion/%v\n", currentUID, metadata.GetName(), metadata.GetNamespace(), metadata.GetResourceVersion())
```
dgoodwin, line 139: Looks leftover or unintentional.

pkg/monitor/nodedetails/node_log.go
```diff
@@ -0,0 +1,66 @@
...
+		Namespace("").Name(nodeName).
+		Resource("nodes").SubResource("proxy", "logs").Suffix("journal").URL().Path
+
+	req := client.CoreV1().RESTClient().Get().RequestURI(path).
+		SetHeader("Accept", "text/plain, */*")
+		//SetHeader("Accept-Encoding", "gzip")
+	req.Param("since", "-7d")
+	req.Param("unit", systemdServiceName)
+
+	fmt.Printf("path: %v\n", path)
```
dgoodwin, line 28: Intentional?

pkg/monitor/nodedetails/node_log.go
```diff
@@ -0,0 +1,66 @@
...
+
+	//decompressedReader, decompressedWriter := io.Pipe()
+	//defer decompressedWriter.Close()
+	//defer decompressedReader.Close()
```
dgoodwin, line 32: Dead code.

pkg/monitor/nodedetails/node_log.go
```diff
@@ -0,0 +1,66 @@
...
+	in, err := req.Stream(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer in.Close()
+
+	//err = optionallyDecompress(decompressedWriter, in)
+	//if err != nil {
+	//	return nil, err
+	//}
```
dgoodwin, line 43: Dead code.

pkg/monitor/nodedetails/node_log.go
```diff
@@ -0,0 +1,66 @@
...
+// GetNodeLog returns logs for a particular systemd service on a given node.
+// We're count on these logs to fit into some reasonable memory size.
+func GetNodeLog(ctx context.Context, client kubernetes.Interface, nodeName, systemdServiceName string) ([]byte, error) {
+	path := client.CoreV1().RESTClient().Get().
+		Namespace("").Name(nodeName).
+		Resource("nodes").SubResource("proxy", "logs").Suffix("journal").URL().Path
+
+	req := client.CoreV1().RESTClient().Get().RequestURI(path).
+		SetHeader("Accept", "text/plain, */*")
+		//SetHeader("Accept-Encoding", "gzip")
```
dgoodwin, line 24: Can be removed I assume.

pkg/monitor/nodedetails/node_log.go
```diff
@@ -0,0 +1,66 @@
...
+
+	//err = optionallyDecompress(decompressedWriter, in)
+	//if err != nil {
+	//	return nil, err
+	//}
+
+	return ioutil.ReadAll(in)
+}
+
+func optionallyDecompress(out io.Writer, in io.Reader) error {
```
dgoodwin, line 48: Unused func if the above is meant to be commented out / removed.

pkg/monitor/intervalcreation/node.go
deads2k: > Per below this confuses me, why would we assume time.Now(), wouldn't that mess up intervals chart?

what value would you prefer?

pkg/monitor/intervalcreation/node.go
deads2k: is it better to not render the problem at all or render it at an unusual time so someone notices.  choosing zero will destroy the chart.  Choosing now will look weird, but will not time-shift the entire chart.

dgoodwin: I guess I'd have expected logging the problematic lines and then dropping the interval. If the goal is to still display it at a weird point and assume that is more visible to someone looking, I can see that potentially helping. I would still log the problem though so we've got something definite we can look for. It somewhat depends what these particular lines are or could be.

deads2k:
pushed update.

PR #27302, opened 2022-07-12
fix(olm-test): skip test if marketplace is not enabled

anik120:
This PR skips the test `[sig-operator] an end user can use OLM can subscribe to the operator` if marketplace is disabled, since the test uses a default catalogsource included in the cluster by marketplace-operator (which when disabled, results in the absence of the default source)

Signed-off-by: Anik Bhattacharjee <anikbhattacharya93@gmail.com>

See https://issues.redhat.com/browse/OLM-2542 for more details.

### DIFF
--- a/test/extended/olm/olm.go
+++ b/test/extended/olm/olm.go
@@ -199,6 +199,23 @@ func archHasDefaultIndex(oc *exutil.CLI) bool {
 	return false
 }
 
+func marketplaceEnabled(oc *exutil.CLI) (bool, error) {
+	output, err := oc.AsAdmin().Run("get").Args("clusterversion", "version", "-o=jsonpath={.status.capabilities.enabledCapabilities}").Output()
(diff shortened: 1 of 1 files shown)


test/extended/olm/olm.go
```diff
@@ -199,6 +200,16 @@ func archHasDefaultIndex(oc *exutil.CLI) bool {
 	return false
 }
 
+func marketplaceEnabled(oc *exutil.CLI) (bool, error) {
+	_, err := oc.AsAdmin().Run("get").Args("crds", "operatorhubs.config.openshift.io").Output()
```
bparees, line 204: I think it would be to just check the clusterversion object and see if the "marketplace" capability is enabled in `status.capabilities`

bparees:
/approve

will let someone from olm team give the lgtm

PR #27312, opened 2022-07-19
Fixes OVN ACL audit log parsing

trozet:
In OVN the default deny ACL naming was changed and thus the log line
will be different for the ACL name. This changes allows for both cases.

Signed-off-by: Tim Rozet <trozet@redhat.com>

### DIFF
--- a/test/extended/networking/acl_audit_log.go
+++ b/test/extended/networking/acl_audit_log.go
@@ -206,7 +206,8 @@ func verifyAuditLogs(out string, ns []string, ips []string, ipv6 bool) (bool, bo
 			allowLogFound = true
 			continue
 		}
-		if strings.Contains(logLine, ns[0]+"_allow-from-same-ns\", verdict=drop") && strings.Contains(logLine, ipMatchSrc+ips[2]) &&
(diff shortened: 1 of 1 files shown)


test/extended/networking/acl_audit_log.go
```diff
@@ -206,7 +206,8 @@ func verifyAuditLogs(out string, ns []string, ips []string, ipv6 bool) (bool, bo
 			allowLogFound = true
 			continue
 		}
-		if strings.Contains(logLine, ns[0]+"_allow-from-same-ns\", verdict=drop") && strings.Contains(logLine, ipMatchSrc+ips[2]) &&
+		if (strings.Contains(logLine, ns[0]+"_allow-from-same-ns\", verdict=drop") && strings.Contains(logLine, ipMatchSrc+ips[2]) ||
+			strings.Contains(logLine, ns[0]+"_ingressDefaultDeny\", verdict=drop") && strings.Contains(logLine, ipMatchSrc+ips[2])) &&
```
tssurya, line 210: do we need the OR or is replacing "_allow-from-same-ns" with "_ingressDefaultDeny" good enough.

trozet: I duno I was thinking just in case we revert to the old behavior later

tssurya:
LGTM, waiting for either gcp or aws to pass

PR #27316, opened 2022-07-21
Bug 2011039: add etcd events to interval charts

tjungblu:
/hold

still figuring out whether we need to create a monitoring like controlplane.StartAllAPIMonitoring


geliu2016 (approved):
/lgtm 
/label cherry-pick-approved

tjungblu:
the events seem to only include alerts, not actual k8s events. Gotta dig a bit more where to find those...

PR #27317, opened 2022-07-22
simplify vertical scaling test

tjungblu:
The endpoints configmap does reflect voting members 1:1 (no learner is ever added there), so we can skip querying etcd altogether and ditch the port forwarding.  I also removed the revision stabilization code, while being nice it's just prolonging the test unnecessarily. 

Last but not least, I also ignored BM now given the latest data on it: http://sippy-prow.dptools.openshift.org/sippy-ng/tests/4.12/analysis?test=%5Bsig-etcd%5D%5BSerial%5D%20etcd%20is%20able%20to%20vertically%20scale%20up%20and%20down%20with%20a%20single%20node%20%5BTimeout%3A60m%5D%20%5BSuite%3Aopenshift%2Fconformance%2Fserial%5D

where it mostly times out due to the provisioned instance not coming up in due time.


deads2k:
> I also removed the revision stabilization code

serial e2e tests must end with a stable cluster (no operator progressing) to guarantee a reliable starting state.  We recently chased a bug where networking failed to do this and added the helper functions to make it easy to achieve.

PR #27318, opened 2022-07-25
Bug 2110503: EgressIP: Probes and check tcpdump logs should succeed eventually 

andreaskaris:
Run sendEgressIPProbesAndCheckPacketSnifferLogs until it eventually
succeeds of for a maximum of 2 minutes.
Before, we checked first until a single packet was EgressIP mapped, and
then ran the main test once. Instead, simply make sure that the test
eventually passed to cut down on flakes.

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

How to test with SDN:
~~~
./openshift-tests run all --print-commands  | grep EgressIP | grep external | grep -v "Skipped:Network/OpenShiftSDN" | while IFS= read -r line; do bash -c "$line" ; done
~~~

How to test with OVN:
~~~
./openshift-tests run all --print-commands  | grep EgressIP | grep external | grep -v "Skipped:Network/OVNKubernetes" | while IFS= read -r line; do bash -c "$line" ; done
~~~

Tested manually with:
* SDN + AWS (latest)
* OVN Azure (4.10)
* OVN + AWS (latest)

### DIFF
--- a/test/extended/networking/egressip.go
+++ b/test/extended/networking/egressip.go
@@ -689,25 +689,15 @@ func spawnProberSendEgressIPTrafficCheckLogs(
 	proberPod := createProberPod(oc, externalNamespace, probePodName)
 
 	// Unfortunately, even after we created the EgressIP object and the CloudPrivateIPConfig, it can take some time before everything is applied correctly.
-	// If we expect non-zero number of hits, add this 60 second retry mechanism that waits for the first log hit that contains the EgressIP.
-	// 60 seconds might seems a bit long, but for the UDP tests, agnhost's netexec needs at least 7 seconds to finish the UDP query. Add to this the log parsing
-	// retry mechanism and we are around 10 seconds minimum for a single round. On top of that, I have seen this take quite some time on Azure.
-	if expectedHits > 0 {
-		framework.Logf("Eventually (after max 60 seconds) the EgressIP setup should converge and the EgressIP configuration should be operational")
-		o.Eventually(func() bool {
-			framework.Logf("Sending a single probe and checking if it shows up inside the logs")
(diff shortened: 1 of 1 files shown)


test/extended/networking/egressip.go
```diff
@@ -689,25 +689,15 @@ func spawnProberSendEgressIPTrafficCheckLogs(
...
-	framework.Logf("Verifying that the expected number of EgressIP outbound requests can be seen in the packet sniffer logs")
-	result, err := sendEgressIPProbesAndCheckPacketSnifferLogs(oc, proberPod, routeName, targetProtocol, targetHost, targetPort, iterations, expectedHits, packetSnifferDaemonSet, egressIPSet, 10)
-	o.Expect(err).NotTo(o.HaveOccurred())
-	o.Expect(result).To(o.BeTrue())
+	// Retry this test every 30 seconds for up to 2 minutes to give the cluster time to converge - eventually, this test should pass.
+	o.Eventually(func() bool {
+		framework.Logf("Verifying that the expected number of EgressIP outbound requests can be seen in the packet sniffer logs")
+		result, err := sendEgressIPProbesAndCheckPacketSnifferLogs(oc, proberPod, routeName, targetProtocol, targetHost, targetPort, iterations, expectedHits, packetSnifferDaemonSet, egressIPSet, 10)
+		return err == nil && result
+	}, 120*time.Second, 30*time.Second).Should(o.BeTrue())
```
jluhrsen, line 697: something bulky here happening that you don't want to poll more frequently than 30s? just thinking if it converges
in time to pass at 31s, we wast 29 seconds waiting. maybe every 10s?

otherwise, seems like a good refactor and /lgtm

andreaskaris: It can potentially take quite some time; I think with UDP it's about 10 seconds to get all of the returns and 10 seconds waiting for the logs to refresh. So I could lower it to 20. Do you know what happens if I set the value to < of what's happening inside the loop?

PR #27319, opened 2022-07-25
Consider Prometheus maximum points in a time series for alert tests

DennisPeriquet:
re: [TRT-412](https://issues.redhat.com//browse/TRT-412)

When the span exceeds about 3 hours (11,000 seconds), we can't use 1s interval for a range query because it will exceed the Prometheus max.

There's still the problem of a test running for a very long time but this change allows the alert tests to still run (by using an interval of at least 2 seconds).

### DIFF
--- a/pkg/monitor/alerts.go
+++ b/pkg/monitor/alerts.go
@@ -32,10 +32,15 @@ func whenWasAlertInState(ctx context.Context, prometheusClient prometheusv1.API,
 	if alertState != "pending" && alertState != "firing" {
 		return nil, fmt.Errorf("unrecognized alertState: %v", alertState)
 	}
+
+	// Prometheus has a hardcoded maximum resolution of 11,000 points per timeseries.  The "Step"
+	// used to be 1 second but if a query exceeded 3 hours, this query would fail.  A resolution of
+	// 2 seconds is fine because the query will work for up to 6 hours and ALERTS change every 30s
+	// at most.
 	timeRange := prometheusv1.Range{
 		Start: startTime,
 		End:   time.Now(),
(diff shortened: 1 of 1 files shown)


pkg/monitor/alerts.go
```diff
@@ -32,11 +36,24 @@ func whenWasAlertInState(ctx context.Context, prometheusClient prometheusv1.API,
...
+	queryInterval := 1
+	if timeNow.Sub(startTime).Seconds() > prometheusMaxResolution {
+		queryInterval = int(timeNow.Sub(startTime).Seconds())/prometheusMaxResolution + 1
+		fmt.Printf("Warning: Alert test range exceeds %d seconds ; using range interval of %d\n", prometheusMaxResolution, queryInterval)
+	}
+
 	timeRange := prometheusv1.Range{
 		Start: startTime,
-		End:   time.Now(),
-		Step:  1 * time.Second,
```
DennisPeriquet, line 38: @deads2k or @simonpasquier looking for opinions on this.

We will exceed 11,000 seconds when a test exceeds 3 hours (and this should not happen normally).  I use a calculation to ensure the Steps fit inside the 11,000 limit but we were wondering if it makes sense to just change the Step to `2 * time.Second` in general.  This would cover the cases for anything up to 6 hours (which is well beyond Prow's 4 hour job limit) and keep the Step consistent for tests that run for less than 3 hours as well as greater than 3 hours.

What do you think? thanks.

DennisPeriquet: Thanks for the comment @simonpasquier I updated my change.

simonpasquier (commented):
/lgtm

PR #27326, opened 2022-07-28
Update authenticated users RBAC e2e test

awgreene:
Problem: OLM has historically communicated to users which operators
are available in a given namespace by creating a copy of an operator
CSV in each namespace that it is scoped to. This has caused performance
issues on large clusters with many namespaces. OLM provided users with
the means to disable Copied CSVs for operators scoped to all namespaces,
but console was unable to communicate which operators were available
globally. Console will look for csvs in the openshift namespace to
identify which operators are available globally for authenticated users.

The authenticated user group needs read permissions to view the
CSVs in the openshift namespace. The extended e2e origin test suite has
a test that ensures that authenticated users do not have unanticipated
RBAC.

Solution: This commit updates the test so that it expects the authenticated
user group to have get, list, and watch permissions on CSVs in the openshift
namespace.

### DIFF
--- a/test/extended/authorization/rbac/groups_default_rules.go
+++ b/test/extended/authorization/rbac/groups_default_rules.go
@@ -67,6 +67,9 @@ const (
 
 	// Provided as CRD via cluster-csi-snapshot-controller-operator
 	snapshotGroup = "snapshot.storage.k8s.io"
+
+	// Provided as CRD via operator-lifecycle-manager
+	operatorsCoreOSGroup = "operators.coreos.com"
 )
 
 // Do not change any of these lists without approval from the auth and master teams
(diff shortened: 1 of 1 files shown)


test/extended/authorization/rbac/groups_default_rules.go
```diff
@@ -53,6 +53,7 @@ const (
 	oauthGroup    = oauth.GroupName
 	projectGroup  = project.GroupName
 	templateGroup = template.GroupName
+	csvGroup      = "operators.coreos.com"
```
timflannagan, line 56: Note: this is an upstream OLM API, and doesn't live in o/api. It's probably best to hardcode here vs. introducing a new dependency into origin.

stlaz: name it operatorsCoreOSGroup, please

awgreene: Can do!

awgreene: Done

awgreene:
>We are adding a rule.
> Is there some link to a PR (feature / test cases) that will make tests break?

@ibihim there is an [OLM PR](https://github.com/openshift/operator-framework-olm/pull/326) that cannot merge because the testcase changed in this PR fails because authenticated users have unexpected permissions.

The OLM PR doesn't test that authenticated users have these permissions, but console will be updating it's operatorhub page to be built using the CSVs in the openshift namespace. This redesign will be tested and will fail if authenticated users lack the RBAC introduced in this PR.

I can add a testcase to the OLM e2e suite to check for these permissions, but it seems redundant given the planned console work. Let me know what you think.

ibihim:
The reference should suffice to connect the dots. Thx.

AFAI can tell, this looks legit. I wouldn't know if there is any better alternative, so I am giving a

/lgtm

but it would be good if one of the more experienced owners would give their final approve, @deads2k, @stlaz.

PR #27323, opened 2022-07-27
bump o/k for the fixed scheduling tests

stlaz:

PR #27329, opened 2022-07-29
[TRT-430] Update all agnhost references to 2.39

yselkowitz:
This was missed as part of the k8s 1.24.1 bump in #27323, and should drop the need to revert it via #27328.

https://issues.redhat.com/browse/TRT-430
https://coreos.slack.com/archives/CFP6ST0A3/p1659093112135979

### DIFF
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -34797,7 +34797,7 @@ var _testExtendedTestdataCmdTestCmdTestdataHelloOpenshiftHelloPodJson = []byte(`
     "containers": [
       {
         "name": "hello-openshift",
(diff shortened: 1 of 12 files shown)


dgoodwin:
No need for payload command here as we have ci/prow/e2e-metal-ipi-ovn-ipv6  on this PR, just need to see it pass, it's not required.

dgoodwin:
We unfortunately decided we have to proceed with revert, but this definitely looks like the likely fix that should be layered on top of the unrevert.

Thanks a bunch @yselkowitz

yselkowitz:
e2e-metal-ipi-ovn-ipv6 passed, so this is the correct fix, but the underlying rebase has been reverted.
/close

PR #27333, opened 2022-08-01
sig-api-machinery: remove project from authenticated browser APIServe…

pacevedom:
…r tests

Lose the project namespace for tests that dont require it.


soltysh (approved):
/approve

PR #27338, opened 2022-08-03
Ignore duplicated console probe events during installation

bparees:
This is a manual backport of the console bits from https://github.com/openshift/origin/pull/26924


bparees:
turns out this is just a refactor and doesn't fix anything we're aware of, so not worth backporting.

PR #27345, opened 2022-08-05
USHIFT-285: Add MicroShift version getter to framework

pacevedom:
Add a function to determine if the cluster is a MicroShift instance.

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -27,6 +27,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/apitesting"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kapierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -2105,4 +2106,40 @@ func DoesApiResourceExist(oc *CLI, apiResourceName string) (bool, error) {
 	}
 
 	return false, nil
+}
+
+var (
+	microShiftVersion         string
+	microShiftVersionNotFound bool
+	microShiftMutex           sync.Mutex
+)
+
+func getMicroShiftVersion(ocClient *CLI) (string, error) {
+	microShiftMutex.Lock()
+	defer microShiftMutex.Unlock()
+
+	if !microShiftVersionNotFound && microShiftVersion == "" {
+		cm, err := ocClient.AdminKubeClient().CoreV1().ConfigMaps("kube-public").Get(context.Background(), "microshift-version", metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				microShiftVersionNotFound = true
+				return "", nil
+			}
+			return "", fmt.Errorf("unable to get microshift version configmap: %w", err)
+		}
+		if version, ok := cm.Data["version"]; !ok || version == "" {
+			return "", fmt.Errorf("version not found")
+		} else {
+			microShiftVersion = version
+		}
+	}
+	return microShiftVersion, nil
+}
+
+func IsMicroShiftDeployment(ocClient *CLI) (bool, error) {
+	version, err := getMicroShiftVersion(ocClient)
+	if err != nil {
+		return false, err
+	}
+	return version != "", nil
 }


test/extended/util/framework.go
```diff
@@ -2106,3 +2116,47 @@ func DoesApiResourceExist(oc *CLI, apiResourceName string) (bool, error) {
...
+			return "", fmt.Errorf("major not found")
+		} else {
+			microShiftVersion = fmt.Sprintf("%s", major)
+		}
+		if minor, ok := cm.Data["minor"]; !ok {
+			return "", fmt.Errorf("minor not found")
+		} else {
+			microShiftVersion = fmt.Sprintf("%s.%s", microShiftVersion, minor)
+		}
+		if version, ok := cm.Data["version"]; !ok {
```
dhellmann, line 2147: The version string is going to be the full display version, so you could just return that instead of concatenating it with the major and minor versions.

pacevedom: Mind if we change it in a future commit? As of today the version comes from the git version. It is true though that the `microShiftVersion` is only there for caching and (maybe) future convenience to extract the concrete version to do something with it.
In this change the caller is not able to see the version, only if its microshift or not.

pacevedom: Fixed it, saw it was including the full version when compiling.

dhellmann:
The implementation looks good. Are those CI job failures related?

pacevedom:
> The implementation looks good. Are those CI job failures related?

Not really, this function is not used yet. Anyway I tmeporarily removed the call from `GetControlPlaneTopology` to inspect if we might enable bad code paths in other tests.

test/extended/util/framework.go
```diff
@@ -2106,3 +2107,37 @@ func DoesApiResourceExist(oc *CLI, apiResourceName string) (bool, error) {
...
+var (
+	microShiftVersion string
+	microShiftMutex   sync.Mutex
+)
+
+func getMicroShiftVersion(ocClient *CLI) (string, error) {
+	microShiftMutex.Lock()
+	defer microShiftMutex.Unlock()
+
+	if microShiftVersion == "" {
```
dhellmann, line 2120: This function would be simpler if this test was reversed and it returned the cached value immediately.

dhellmann (commented):
One nit that can be addressed later.

/approve

test/extended/util/framework.go
pacevedom: I thought of doing the same, but I chose to follow the existing pattern here: https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L2002-L2012

There is a kind of early return but its only under some special conditions a few lines below: https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L2030-L2042

Should I keep the style?

ggiguash:
/retitle USHIFT-285: Add MicroShift version getter to framework

test/extended/util/framework.go
dhellmann: ok, let's keep this version

dhellmann (commented):
/lgtm

test/extended/util/framework.go
```diff
@@ -2106,3 +2107,37 @@ func DoesApiResourceExist(oc *CLI, apiResourceName string) (bool, error) {
...
 	return false, nil
 }
+
+var (
+	microShiftVersion string
+	microShiftMutex   sync.Mutex
+)
+
+func getMicroShiftVersion(ocClient *CLI) (string, error) {
+	microShiftMutex.Lock()
```
aojea, line 2117: why do you need a mutex?

pacevedom: Tests run in parallel and maybe we have concurrent calls to this function. It's not something critical as the value would get reconstructed from immutable values in the cluster, but it saves calls to the API (only when starting tests, though), plus it follows the current style of other getters in the package:
https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L1982
https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L2027

aojea: just curious, :smile:

ingvagabund (changes requested):

deads2k:
/hold

let's see where this is needed instead of using existing external signals before merging.

test/extended/util/framework.go
```diff
@@ -2106,3 +2107,37 @@ func DoesApiResourceExist(oc *CLI, apiResourceName string) (bool, error) {
...
+	defer microShiftMutex.Unlock()
+
+	if microShiftVersion == "" {
+		cm, err := ocClient.AdminKubeClient().CoreV1().ConfigMaps("kube-public").Get(context.Background(), "microshift-version", metav1.GetOptions{})
```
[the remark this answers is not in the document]
pacevedom: Done

pacevedom:
> /hold
> 
> let's see where this is needed instead of using existing external signals before merging.

So it's these tests now:
```
"[sig-auth][Feature:PodSecurity] restricted-v2 SCC should mutate empty securityContext to match restricted PSa profile [Suite:openshift/conformance/parallel]"
"[sig-auth][Feature:OpenShiftAuthorization] The default cluster RBAC policy should have correct RBAC rules [Suite:openshift/conformance/parallel]"
"[sig-coreos] [Conformance] CoreOS bootimages TestBootimagesPresent [Suite:openshift/conformance/parallel/minimal]"
"[sig-devex][Feature:OpenShiftControllerManager] TestAutomaticCreationOfPullSecrets [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel]"
"[sig-devex][Feature:OpenShiftControllerManager] TestDockercfgTokenDeletedController [Suite:openshift/conformance/parallel]"
"[sig-imageregistry][Feature:ImageInfo] Image info should display information about images [Skipped:Disconnected] [Suite:openshift/conformance/parallel]"
"[sig-network][Feature:Router][apigroup:route.openshift.io] The HAProxy router converges when multiple routers are writing conflicting status [Suite:openshift/conformance/parallel]"
"[sig-network][Feature:Router][apigroup:route.openshift.io] The HAProxy router converges when multiple routers are writing status [Suite:openshift/conformance/parallel]"
```

More to come as we create/modify specific tests for microshift, I would say.

pacevedom:
Not needed anymore, as we look for dependencies in each test.

PR #27347, opened 2022-08-08
networking/acl: live-retrieve the NS instead of f.Namespace

stlaz:
f.Namespace does not actually represent the current and updated
object and so an UPDATE with such an object will cause removal
of the annotations/labels that, among others, set the PSa level
and auto-PSa label synchronization status.

/assign @openshift/networking


astoycos:
/lgtm 

thanks!

stlaz:
/retest-required
the test failure are irrelevant to the fixes as this is fixing the parallel runs

PR #27348, opened 2022-08-08
display UTC time for timeline charts

neisw:


DennisPeriquet:
You have to run `make update-bindata` to generate the bindata.go.  You cannot just update the e2e-chart-template.html.  I believe it is why the `ci/prow/verify` test is failing.

DennisPeriquet:
I have independently confirmed that with this change, the interval chart shows with UTC times.

I can give this an `/lgtm` but I'm not an owner or approver.

Thanks for making this change as it makes the debugging just a bit easier when the timestamps are all in UTC.

/lgtm

PR #27355, opened 2022-08-16
Label tests with openshift crafted API groups (round 1)

ingvagabund:
Procedure for detecting API groups in use:
- search for all Get, Update, Patch, Delete and List client invocations
  (for both direct and indirect)
- go through every e2e test code and identify all possible invocations of an api group (e.g. oc command)

### DIFF
--- a/test/extended/apiserver/root_403.go
+++ b/test/extended/apiserver/root_403.go
@@ -1,19 +1,16 @@
 package apiserver
 
 import (
-	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"io/ioutil"
 	"net/http"
-	"strings"
 
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/rest"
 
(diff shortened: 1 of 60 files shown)


test/extended/authorization/scc.go
```diff
@@ -21,7 +21,7 @@ var _ = g.Describe("[sig-auth][Feature:SCC][Early]", func() {
 
 	oc := exutil.NewCLIWithoutNamespace("working-scc-during-install")
 
-	g.It("should not have pod creation failures during install", func() {
+	g.It("should not have pod creation failures during install [apigroup:config.openshift.io]", func() {
```
deads2k, line 24: What requires this api group in this test?

test/extended/apiserver/root_403.go
```diff
@@ -25,7 +25,7 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 
 	oc := exutil.NewCLI("apiserver")
 
-	g.It("anonymous browsers should get a 403 from /", func() {
+	g.It("anonymous browsers should get a 403 from / [apigroup:config.openshift.io]", func() {
```
deads2k, line 28: this test should be rewritten to avoid the dep

test/extended/authorization/scc.go
ingvagabund: `util.AllClusterVersionsAreGTE` and https://github.com/ingvagabund/origin/blob/7826178007475886fb8878e1e60e91f14eb476b2/test/extended/util/framework.go#L1986:
```
func AllClusterVersionsAreGTE(version semver.Version, config *rest.Config) (bool, error) {
	c, err := configv1client.NewForConfig(config)
	if err != nil {
		return false, err
	}
	cv, err := c.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
	if err != nil {
		return false, err
	}
```

deads2k: Fun.  Lucky for us, we can remove that line now.

ingvagabund: Removed

test/extended/apiserver/root_403.go
ingvagabund: Removed

ingvagabund:
Just re-generating go.mod

PR #27358, opened 2022-08-17
USHIFT-285: MicroShift: allow CLI creation without OpenShift APIs

pacevedom:
Swaps project API with k8s namespaces in test/extended/util package. Automatic detection of project API to create a CLI object according to whats available in the cluster.

### DIFF
--- a/test/extended/util/client.go
+++ b/test/extended/util/client.go
@@ -251,6 +251,15 @@ func (c CLI) WithToken(token string) *CLI {
 // All resources will be then created within this project.
 // Returns the name of the new project.
 func (c *CLI) SetupProject() string {
+	exist, err := DoesApiResourceExist(c, "projects")
+	o.Expect(err).ToNot(o.HaveOccurred())
+	if exist {
+		return c.setupProject()
+	}
+	return c.setupNamespace()
+}
+
+func (c *CLI) setupProject() string {
 	requiresTestStart()
 	newNamespace := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("e2e-test-%s-", c.kubeFramework.BaseName))
 	c.SetNamespace(newNamespace).ChangeUser(fmt.Sprintf("%s-user", newNamespace))
@@ -291,7 +300,7 @@ func (c *CLI) SetupProject() string {
 	}
 	for _, sa := range DefaultServiceAccounts {
 		framework.Logf("Waiting for ServiceAccount %q to be provisioned...", sa)
-		err = WaitForServiceAccount(c.KubeClient().CoreV1().ServiceAccounts(newNamespace), sa)
+		err = WaitForServiceAccountWithSecret(c.KubeClient().CoreV1().ServiceAccounts(newNamespace), sa)
 		o.Expect(err).NotTo(o.HaveOccurred())
 	}
 
@@ -334,6 +343,47 @@ func (c *CLI) SetupProject() string {
 	WaitForProjectSCCAnnotations(c.ProjectClient().ProjectV1(), newNamespace)
 
 	framework.Logf("Project %q has been fully provisioned.", newNamespace)
+	return newNamespace
+}
+
+func (c *CLI) setupNamespace() string {
+	requiresTestStart()
+	newNamespace := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("e2e-test-%s-", c.kubeFramework.BaseName))
+	c.SetNamespace(newNamespace).ChangeUser(fmt.Sprintf("%s-user", newNamespace))
+
+	nsObject := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: newNamespace,
+		},
+	}
+	framework.Logf("Creating namespace %q", newNamespace)
+	_, err := c.AdminKubeClient().CoreV1().Namespaces().Create(context.Background(), nsObject, metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	c.kubeFramework.AddNamespacesToDelete(nsObject)
+
+	framework.Logf("Waiting on permissions in namespace %q ...", newNamespace)
+	err = WaitForSelfSAR(1*time.Second, 60*time.Second, c.AdminKubeClient(), kubeauthorizationv1.SelfSubjectAccessReviewSpec{
+		ResourceAttributes: &kubeauthorizationv1.ResourceAttributes{
+			Namespace: newNamespace,
+			Verb:      "create",
+			Group:     "",
+			Resource:  "pods",
+		},
+	})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	err = c.setupNamespacePodSecurity(newNamespace)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	serviceAccountName := "default"
+	framework.Logf("Waiting for ServiceAccount %q to be provisioned...", serviceAccountName)
+	err = WaitForServiceAccount(c.KubeClient().CoreV1().ServiceAccounts(newNamespace), serviceAccountName)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	WaitForNamespaceSCCAnnotations(c.KubeClient().CoreV1(), newNamespace)
+
+	framework.Logf("Namespace %q has been fully provisioned.", newNamespace)
 	return newNamespace
 }
 
@@ -770,6 +820,16 @@ func (c *CLI) CreateUser(prefix string) *userv1.User {
 }
 
 func (c *CLI) GetClientConfigForUser(username string) *rest.Config {
+
+	userAPIExists, err := DoesApiResourceExist(c, "users")
+	if err != nil {
+		FatalErr(err)
+	}
+
+	if !userAPIExists {
+		return turnOffRateLimiting(rest.CopyConfig(c.AdminConfig()))
+	}
+
 	ctx := context.Background()
 	userClient := c.AdminUserClient()
 
--- a/test/extended/util/client.go
+++ b/test/extended/util/client.go
@@ -3,8 +3,13 @@ package util
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
 	"crypto/sha256"
+	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/base64"
+	"encoding/pem"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -23,7 +28,9 @@ import (
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	yaml "gopkg.in/yaml.v2"
 
+	"github.com/openshift/api/annotations"
 	kubeauthorizationv1 "k8s.io/api/authorization/v1"
+	certificatesv1 "k8s.io/api/certificates/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -340,7 +347,7 @@ func (c *CLI) setupProject() string {
 		o.Expect(err).NotTo(o.HaveOccurred())
 	}
 
-	WaitForProjectSCCAnnotations(c.ProjectClient().ProjectV1(), newNamespace)
+	WaitForNamespaceSCCAnnotations(c.KubeClient().CoreV1(), newNamespace)
 
 	framework.Logf("Project %q has been fully provisioned.", newNamespace)
 	return newNamespace
@@ -349,21 +356,38 @@ func (c *CLI) setupProject() string {
 func (c *CLI) setupNamespace() string {
 	requiresTestStart()
 	newNamespace := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("e2e-test-%s-", c.kubeFramework.BaseName))
-	c.SetNamespace(newNamespace).ChangeUser(fmt.Sprintf("%s-user", newNamespace))
+	username := fmt.Sprintf("%s-user", newNamespace)
+	serviceAccountName := "default"
+	c.SetNamespace(newNamespace)
 
 	nsObject := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: newNamespace,
+			Annotations: map[string]string{
+				annotations.OpenShiftDescription: username,
+				annotations.OpenShiftDisplayName: newNamespace,
+				"openshift.io/requester":         username,
+			},
 		},
 	}
 	framework.Logf("Creating namespace %q", newNamespace)
 	_, err := c.AdminKubeClient().CoreV1().Namespaces().Create(context.Background(), nsObject, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
-
 	c.kubeFramework.AddNamespacesToDelete(nsObject)
 
+	framework.Logf("Waiting for ServiceAccount %q to be provisioned...", serviceAccountName)
+	err = WaitForServiceAccount(c.AdminKubeClient().CoreV1().ServiceAccounts(newNamespace), serviceAccountName)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	framework.Logf("Configuring kubeconfig with user %q certificates...", username)
+	c.ChangeUser(username)
+
+	framework.Logf("Waiting for RoleBinding %q to be provisioned...", username)
+	err = c.setupRoleInNamespace(username)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
 	framework.Logf("Waiting on permissions in namespace %q ...", newNamespace)
-	err = WaitForSelfSAR(1*time.Second, 60*time.Second, c.AdminKubeClient(), kubeauthorizationv1.SelfSubjectAccessReviewSpec{
+	err = WaitForSelfSAR(1*time.Second, 60*time.Second, c.KubeClient(), kubeauthorizationv1.SelfSubjectAccessReviewSpec{
 		ResourceAttributes: &kubeauthorizationv1.ResourceAttributes{
 			Namespace: newNamespace,
 			Verb:      "create",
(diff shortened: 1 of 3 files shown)


test/extended/util/client.go
```diff
@@ -337,6 +345,40 @@ func (c *CLI) SetupProject() string {
...
+			Name: newNamespace,
+		},
+	}
+	framework.Logf("Creating namespace %q", newNamespace)
+	_, err := c.AdminKubeClient().CoreV1().Namespaces().Create(context.Background(), nsObject, metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	c.kubeFramework.AddNamespacesToDelete(nsObject)
+
+	framework.Logf("Waiting on permissions in project %q ...", newNamespace)
```
dhellmann, line 364:
```suggestion
	framework.Logf("Waiting on permissions in namespace %q ...", newNamespace)
```

test/extended/util/client.go
```diff
@@ -770,52 +812,60 @@ func (c *CLI) CreateUser(prefix string) *userv1.User {
...
+			UserUID:     string(user.UID),
+			Scopes:      []string{"user:full"},
+			RedirectURI: "https://localhost:8443/oauth/token/implicit",
+		}, metav1.CreateOptions{})
+		if err != nil {
+			FatalErr(err)
+		}
+		c.AddResourceToDelete(oauthv1.GroupVersion.WithResource("oauthaccesstokens"), token)
+		userClientConfig.BearerToken = privToken
+	} else {
```
dhellmann, line 866: The change in this function would be smaller if the `if` test on line 821 was reversed and then the userClientConfig being created here was returned early. That would also make the function easier to follow, because it would be clear that most of the steps apply when the users API is present.

dhellmann (commented):
A couple of small suggestions, but this looks good generally.

test/extended/util/client.go
pacevedom: Done

test/extended/util/client.go
pacevedom: You are right, looks better now. Thanks!

ggiguash:
/retitle USHIFT-285: MicroShift: allow CLI creation without OpenShift APIs

test/extended/util/client.go
```diff
@@ -337,6 +345,40 @@ func (c *CLI) SetupProject() string {
 	return newNamespace
 }
 
+func (c *CLI) setupNamespace() string {
+	requiresTestStart()
+	newNamespace := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("e2e-test-%s-", c.kubeFramework.BaseName))
+	c.SetNamespace(newNamespace).ChangeUser(fmt.Sprintf("%s-user", newNamespace))
```
deads2k, line 351: ChangeUser requires the users.user API.

deads2k, line 351: 1. you should try to execute this and see it work without the user API
2. you could create a new ChangeUserWithClientCert that uses a client CSR to change the identity and persist in a kubeconfig file.

test/extended/util/client.go
```diff
@@ -337,6 +345,40 @@ func (c *CLI) SetupProject() string {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: newNamespace,
+		},
+	}
+	framework.Logf("Creating namespace %q", newNamespace)
+	_, err := c.AdminKubeClient().CoreV1().Namespaces().Create(context.Background(), nsObject, metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	c.kubeFramework.AddNamespacesToDelete(nsObject)
+
```
deads2k, line 363: you need to do the work of actually creating the things in teh standard openshfit project template like the proper annotations and rolebindings.

test/extended/util/client.go
```diff
@@ -337,6 +345,40 @@ func (c *CLI) SetupProject() string {
...
+	err = WaitForSelfSAR(1*time.Second, 60*time.Second, c.AdminKubeClient(), kubeauthorizationv1.SelfSubjectAccessReviewSpec{
+		ResourceAttributes: &kubeauthorizationv1.ResourceAttributes{
+			Namespace: newNamespace,
+			Verb:      "create",
+			Group:     "",
+			Resource:  "pods",
+		},
+	})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
```
deads2k, line 374: you're missing the standard "wait for these things" from the setupproject.

test/extended/util/client.go
```diff
@@ -124,7 +124,15 @@ func NewCLI(project string) *CLI {
 	cli := NewCLIWithoutNamespace(project)
 	cli.withoutNamespace = false
 	// create our own project
-	g.BeforeEach(func() { _ = cli.SetupProject() })
+	g.BeforeEach(func() {
```
deads2k, line 127: once the "wait for these things" step is common between the two impls, then I have relatively few fears about drift given the stability of the projectrequest API.

deads2k:
Currently this PR
1. still relies on openshift APIs not present in microshift
2. does not provide a mechanism to set the user as expected by the test being run
3. does not provide the user with permissions to work in the namespace
4. does not provide interoperability with the functionality it replaces
5. does not provide congruent gating guarantees for all downstream consumers


/hold until the functionality is a drop-in replacement
/assign @ingvagabund

test/extended/util/client.go
pacevedom: It works because of https://github.com/openshift/origin/pull/27358/files#diff-e21320d17fdce22cad99f368c20f8aca8b28bac96f890d69e95fda9be0ae7f93R822
Here, `AdminConfig` resolves to the kubeconfig that was used to launch the tests binary, so its the same as the `UserConfig`.
Anyway I can create a function like you say, will follow that approach.

pacevedom: This was covered here: https://github.com/openshift/origin/pull/27358/files#diff-e21320d17fdce22cad99f368c20f8aca8b28bac96f890d69e95fda9be0ae7f93R822

Even though it means that `UserConfig` == `AdminConfig`

test/extended/util/client.go
pacevedom: A project creates role bindings for cluster roles `system:image-puller`, `system:image-builder` and `system:deployer`. The first two relate only to OCP APIs, so we can skip them.
The last one I will need it if I change the kubeconfig for the `UserConfig`, as it grants permissions to operate on pods, replication controllers, events and other OCP APIs.

test/extended/util/client.go
deads2k: >This was covered here: https://github.com/openshift/origin/pull/27358/files#diff-e21320d17fdce22cad99f368c20f8aca8b28bac96f890d69e95fda9be0ae7f93R822

>Even though it means that UserConfig == AdminConfig

That is functionally incorrect.  The reason we switch users is to prove that the features function properly when run as an unprivileged user.  Given differences in admission and REST handling, using an admin user doesn't test the functionality.

test/extended/util/client.go
deads2k: > The last one I will need it if I change the kubeconfig for the UserConfig, as it grants permissions to operate on pods, replication controllers, events and other OCP APIs.

It must switch users or the testing is invalid.  The annotations are important because they change the way that consumers are able to determine ownership.

test/extended/util/client.go
pacevedom: Agree, was not familiar with the project API. Now the service account has the same permissions as a project user.

pacevedom:
Added user certificate handling for authentication.

test/extended/util/client.go
pacevedom: Done

test/extended/util/client.go
pacevedom: Done

test/extended/util/client.go
pacevedom: Fixed

test/extended/util/client.go
pacevedom: Done

test/extended/util/client.go
```diff
@@ -337,8 +352,155 @@ func (c *CLI) SetupProject() string {
...
+	_, err := c.AdminKubeClient().CoreV1().Namespaces().Create(context.Background(), nsObject, metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	c.kubeFramework.AddNamespacesToDelete(nsObject)
+
+	framework.Logf("Waiting for ServiceAccount %q to be provisioned...", serviceAccountName)
+	err = WaitForServiceAccount(c.AdminKubeClient().CoreV1().ServiceAccounts(newNamespace), serviceAccountName)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	framework.Logf("Configuring kubeconfig with user %q certificates...", username)
+	c.ChangeUser(username)
```
deads2k, line 377: I don't understand how this will work until you modify ChangeUser to tolerate missing the APIs served by the oauth-apiserver and when those APIs are missing use the CSR API to get credentials for the requested user.

deads2k: Found it below

test/extended/util/client.go
```diff
@@ -337,8 +352,160 @@ func (c *CLI) SetupProject() string {
...
+			Name:     "admin",
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "User",
+				Name:      username,
+				Namespace: c.Namespace(),
+			},
+			{
+				Kind:      "ServiceAccount",
```
deads2k, line 421: I don't recall granting this power to the service account in the project request case.  ref please?

test/extended/util/client.go
```diff
@@ -337,8 +352,160 @@ func (c *CLI) SetupProject() string {
...
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, err
+	}
+	privateKeyPem := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(privateKey),
+	})
+
+	subj := pkix.Name{CommonName: username}
```
deads2k, line 444: missing groups?  
1. `system:authenticated`
2. `system:authenticated:oauth`

test/extended/util/client.go
```diff
@@ -337,8 +352,160 @@ func (c *CLI) SetupProject() string {
...
 }
 
+func (c *CLI) setupNamespace() string {
+	requiresTestStart()
+	newNamespace := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("e2e-test-%s-", c.kubeFramework.BaseName))
+	username := fmt.Sprintf("%s-user", newNamespace)
+	serviceAccountName := "default"
+	c.SetNamespace(newNamespace)
+
+	nsObject := &corev1.Namespace{
```
deads2k, line 362: missing annotations https://github.com/openshift/openshift-apiserver/blob/master/pkg/project/apiserver/registry/projectrequest/delegated/sample_template.go#L46-L48

test/extended/util/client.go
deads2k: still missing annotations.  see https://github.com/openshift/openshift-apiserver/blob/master/pkg/project/apiserver/registry/projectrequest/delegated/sample_template.go#L46-L48

test/extended/util/client.go
```diff
@@ -337,8 +352,160 @@ func (c *CLI) SetupProject() string {
...
+			Group:     "",
+			Resource:  "pods",
+		},
+	})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	err = c.setupNamespacePodSecurity(newNamespace)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	WaitForNamespaceSCCAnnotations(c.KubeClient().CoreV1(), newNamespace)
```
deads2k, line 397: update the `WaitForProjectSCCAnnotations` call site to use the `WaitForNamespaceSCCAnnotations` method and remove the `WaitForProjectSCCAnnotations` to prevent accidents

test/extended/util/framework.go
```diff
@@ -1018,32 +1018,60 @@ func CheckBuildCancelled(b *buildv1.Build) bool {
...
+	hasDockerCfg := false
+	for _, s := range sa.ImagePullSecrets {
+		if strings.Contains(s.Name, "-dockercfg-") {
+			hasDockerCfg = true
+		}
+	}
+	return hasDockerCfg
+}
+
+// WaitForServiceAccountWithSecret waits until the named service account gets fully
```
deads2k, line 1042: wrong name

test/extended/util/framework.go
```diff
@@ -1018,32 +1018,60 @@ func CheckBuildCancelled(b *buildv1.Build) bool {
...
+// provisioned. without dockercfg secrets
 func WaitForServiceAccount(c corev1client.ServiceAccountInterface, name string) error {
 	waitFn := func() (bool, error) {
-		sc, err := c.Get(context.Background(), name, metav1.GetOptions{})
+		sa, err := getServiceAccount(c, name)
```
deads2k, line 1046: I see no reason to change this logic

```go
sc, err := c.Get(context.Background(), name, metav1.GetOptions{})
if err != nil {
	// If we can't access the service accounts, let's wait till the controller
	// create it.
	if kapierrs.IsNotFound(err) || kapierrs.IsForbidden(err) {
		e2e.Logf("Waiting for service account %q to be available: %v (will retry) ...", name, err)
		return false, nil
	}
	return false, fmt.Errorf("Failed to get service account %q: %v", name, err)
}
```

is kube-standard (versus nil, nil return) and does what you're trying to do, right?

test/extended/util/framework.go
```diff
@@ -1018,32 +1018,60 @@ func CheckBuildCancelled(b *buildv1.Build) bool {
...
+		return nil, err
+	}
+	return sc, nil
+}
+
+func serviceAccountContainsPullSecret(sa *corev1.ServiceAccount) bool {
+	hasDockerCfg := false
+	for _, s := range sa.ImagePullSecrets {
+		if strings.Contains(s.Name, "-dockercfg-") {
+			hasDockerCfg = true
```
deads2k, line 1036: if you choose to refactor like this, this should be `return true`

test/extended/util/framework.go
```diff
@@ -1018,32 +1018,60 @@ func CheckBuildCancelled(b *buildv1.Build) bool {
...
+		}
+	}
+	return hasDockerCfg
```
deads2k, line 1039: and this should be return false

test/extended/util/oc_copy.go
```diff
@@ -49,7 +49,7 @@ func getUserPartOfNickname(clientCfg *restclient.Config) (string, error) {
 		return "", err
 	}
 	userInfo, err := userClient.Users().Get(context.Background(), "~", metav1.GetOptions{})
-	if kerrors.IsNotFound(err) || kerrors.IsForbidden(err) {
+	if kerrors.IsNotFound(err) || kerrors.IsForbidden(err) || kerrors.IsUnauthorized(err) {
```
deads2k, line 52: why is this necessary?  This is an indication that an invalid token or expired or misconfigured client-certificate was used.  In either case, the error should be exposed.

test/extended/util/client.go
pacevedom: Added them.

test/extended/util/client.go
pacevedom: Done

test/extended/util/framework.go
pacevedom: Was just trying to refactor it to avoid repetition, since the same code was in two different functions. Do you think its best to have it self-contained in each of the functions? It is `WaitForServiceAccount` and `WaitForServiceAccountWithSecret`.

test/extended/util/client.go
pacevedom: It slipped through while testing some of the sig-network tests. Removed it.

test/extended/util/framework.go
pacevedom: Done

test/extended/util/framework.go
pacevedom: Done

test/extended/util/oc_copy.go
pacevedom: This was a leftover from the previous version, I removed it.

test/extended/util/client.go
```diff
@@ -331,14 +347,167 @@ func (c *CLI) SetupProject() string {
...
-	WaitForProjectSCCAnnotations(c.ProjectClient().ProjectV1(), newNamespace)
+	WaitForNamespaceSCCAnnotations(c.KubeClient().CoreV1(), newNamespace)
 
 	framework.Logf("Project %q has been fully provisioned.", newNamespace)
 	return newNamespace
 }
 
+func (c *CLI) setupNamespace() string {
+	requiresTestStart()
+	newNamespace := names.SimpleNameGenerator.GenerateName(fmt.Sprintf("e2e-test-%s-", c.kubeFramework.BaseName))
```
stlaz, line 358: why don't you use the kube framework method for namespace creation? It's likely to do most work for you.

test/extended/util/client.go
stlaz: @deads2k if these are not added in Microshift, why do we add them? The tests that require them should rightfully fail in our case, no?

test/extended/util/client.go
pacevedom: Got this wrong, let me rework it.

pacevedom: Done

test/extended/util/client.go
```diff
@@ -251,6 +257,15 @@ func (c CLI) WithToken(token string) *CLI {
 // All resources will be then created within this project.
 // Returns the name of the new project.
 func (c *CLI) SetupProject() string {
+	exist, err := DoesApiResourceExist(c, "projects")
```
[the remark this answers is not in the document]
deads2k: > Wouldn't it make more sense if this method accepted the resource and its group?

We did not instrument with individual APIs.

test/extended/util/client.go
deads2k: > @deads2k if these are not added in Microshift, why do we add them? The tests that require them should rightfully fail in our case, no?

Possibly, though I think we can say, "make an equivalent namespace"?  I could also get behind trying to tighten once we break the back of the problem.

test/extended/util/framework.go
```diff
@@ -1018,87 +1017,88 @@ func CheckBuildCancelled(b *buildv1.Build) bool {
 	return b.Status.Phase == buildv1.BuildPhaseCancelled
 }
 
+func getServiceAccount(c corev1client.ServiceAccountInterface, name string) (*corev1.ServiceAccount, error) {
+	sc, err := c.Get(context.Background(), name, metav1.GetOptions{})
+	if err != nil {
+		if kapierrs.IsNotFound(err) || kapierrs.IsForbidden(err) {
+			return nil, nil
```
deads2k, line 1024: This is a blocker.  We don't do this when working with kube APIs because it defies expectations about client handling that have existed for the past eight years.

test/extended/util/framework.go
deads2k: still a problem here.

test/extended/util/framework.go
pacevedom: Left it as it was before.

deads2k:
no conflict with the PR that merged 12h ago.

PR #27359, opened 2022-08-17
Label tests with openshift crafted API groups (round 2)

ingvagabund:
Built on top of https://github.com/openshift/origin/pull/27355

Procedure for detecting API groups in use:
- search for all Get, Update, Patch, Delete and List client invocations
  (for both direct and indirect)
- go through every e2e test code and identify all possible invocations of an api group (e.g. oc command)

### DIFF
--- a/test/extended/networking/egress_firewall.go
+++ b/test/extended/networking/egress_firewall.go
@@ -43,14 +43,14 @@ var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
 	// For Openshift SDN its supports EgressNetworkPolicy objects
 	InOpenShiftSDNContext(
 		func() {
-			g.It("should ensure egressnetworkpolicy is created", func() {
+			g.It("should ensure egressnetworkpolicy is created [apigroup:network.openshift.io]", func() {
 				doEgressFwTest(egFwf, egFwoc, openShiftSDNManifest)
 			})
 		},
 	)
 	noegFwoc := exutil.NewCLIWithPodSecurityLevel(noEgressFWE2E, admissionapi.LevelBaseline)
 	noegFwf := noegFwoc.KubeFramework()
-	g.It("egressFirewall should have no impact outside its namespace", func() {
+	g.It("egressFirewall should have no impact outside its namespace [apigroup:config.openshift.io]", func() {
 		g.By("creating test pod")
 		pod := "dummy"
 		o.Expect(createTestEgressFw(noegFwf, pod)).To(o.Succeed())
--- a/test/extended/networking/util.go
+++ b/test/extended/networking/util.go
@@ -500,7 +500,7 @@ func InOpenShiftSDNContext(body func()) {
 }
 
 func InBareMetalIPv4ClusterContext(oc *exutil.CLI, body func()) {
-	Context("when running openshift ipv4 cluster on bare metal",
+	Context("when running openshift ipv4 cluster on bare metal [apigroup:config.openshift.io]",
 		func() {
 			BeforeEach(func() {
 				pType, err := platformType(oc.AdminConfigClient())
--- a/test/extended/oauth/server_headers.go
+++ b/test/extended/oauth/server_headers.go
@@ -17,7 +17,7 @@ import (
 	"github.com/openshift/origin/test/extended/util/oauthserver"
 )
 
-var _ = g.Describe("[sig-auth][Feature:OAuthServer] [Headers]", func() {
+var _ = g.Describe("[sig-auth][Feature:OAuthServer] [Headers][apigroup:route.openshift.io]", func() {
 	var oc = exutil.NewCLIWithPodSecurityLevel("oauth-server-headers", admissionapi.LevelBaseline)
 	var transport http.RoundTripper
 	var oauthServerAddr string
--- a/test/extended/olm/olm.go
+++ b/test/extended/olm/olm.go
@@ -82,7 +82,7 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
 
 	// OCP-24061 - [bz 1685230] OLM operator should use imagePullPolicy: IfNotPresent
 	// author: bandrade@redhat.com
-	g.It("have imagePullPolicy:IfNotPresent on thier deployments", func() {
+	g.It("have imagePullPolicy:IfNotPresent on thier deployments [apigroup:config.openshift.io]", func() {
 		oc := oc.AsAdmin().WithoutNamespace()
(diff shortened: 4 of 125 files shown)


test/extended/networking/egress_firewall.go
```diff
@@ -43,14 +43,14 @@ var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
...
-			g.It("should ensure egressnetworkpolicy is created", func() {
+			g.It("should ensure egressnetworkpolicy is created [apigroup:network.openshift.io]", func() {
 				doEgressFwTest(egFwf, egFwoc, openShiftSDNManifest)
 			})
 		},
 	)
 	noegFwoc := exutil.NewCLIWithPodSecurityLevel(noEgressFWE2E, admissionapi.LevelBaseline)
 	noegFwf := noegFwoc.KubeFramework()
-	g.It("egressFirewall should have no impact outside its namespace", func() {
+	g.It("egressFirewall should have no impact outside its namespace [apigroup:config.openshift.io]", func() {
```
deads2k, line 53: @danwinship fyi, this dependency will prevent this test from functioning on microshift.  I don't know if you'd expect it to function on microshift, but if you do, you'll need to refactor.

test/extended/networking/util.go
```diff
@@ -500,7 +500,7 @@ func InOpenShiftSDNContext(body func()) {
 }
 
 func InBareMetalIPv4ClusterContext(oc *exutil.CLI, body func()) {
-	Context("when running openshift ipv4 cluster on bare metal",
+	Context("when running openshift ipv4 cluster on bare metal [apigroup:config.openshift.io]",
```
deads2k, line 503: @ingvagabund please start a doc or something with TODOs to recheck usage.  I suspect the fanout on this one may be unfortunate because microshift is baremetal.

test/extended/oauth/server_headers.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/openshift/origin/test/extended/util/oauthserver"
 )
 
-var _ = g.Describe("[sig-auth][Feature:OAuthServer] [Headers]", func() {
+var _ = g.Describe("[sig-auth][Feature:OAuthServer] [Headers][apigroup:route.openshift.io]", func() {
```
deads2k, line 20: This is going to fail for something else.  We can deal with it later, but this is going to fail because we have no oauth server.  I suspect we'll need a detection in the test based on whether we have the well_known.

test/extended/olm/olm.go
```diff
@@ -82,7 +82,7 @@ var _ = g.Describe("[sig-operator] OLM should", func() {
 
 	// OCP-24061 - [bz 1685230] OLM operator should use imagePullPolicy: IfNotPresent
 	// author: bandrade@redhat.com
-	g.It("have imagePullPolicy:IfNotPresent on thier deployments", func() {
+	g.It("have imagePullPolicy:IfNotPresent on thier deployments [apigroup:config.openshift.io]", func() {
```
deads2k, line 85: please add this to the list of followups as well.  Keep it here, I'm just thinking we'll want to go back and think about these.

test/extended/operators/routable.go
```diff
@@ -19,7 +19,7 @@ import (
 	exurl "github.com/openshift/origin/test/extended/util/url"
 )
 
-var _ = g.Describe("[sig-arch] Managed cluster should", func() {
+var _ = g.Describe("[sig-arch] Managed cluster should [apigroup:apps.openshift.io]", func() {
```
deads2k, line 22: add this to the list of followups.

test/extended/prometheus/prometheus.go
```diff
@@ -39,7 +39,7 @@ import (
 	helper "github.com/openshift/origin/test/extended/util/prometheus"
 )
 
-var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules", func() {
+var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigroup:image.openshift.io]", func() {
```
deads2k, line 42: help me remember this if/when we ever add metrics.  Also, what a strange dependency.

test/extended/router/certs.go
```diff
@@ -101,7 +101,7 @@ u3YLAbyW/lHhOCiZu2iAI8AbmXem9lW6Tr7p/97s0w==
 
 	g.When("FIPS is enabled", func() {
 		g.Describe("the HAProxy router", func() {
-			g.It("should not work when configured with a 1024-bit RSA key", func() {
+			g.It("should not work when configured with a 1024-bit RSA key [apigroup:template.openshift.io]", func() {
```
deads2k, line 104: This is not good.  Please mark for followup, and probably even a bug.  The router needs to work on microshift.

test/extended/router/config_manager.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const timeoutSeconds = 3 * 60
 
-var _ = g.Describe("[sig-network][Feature:Router]", func() {
+var _ = g.Describe("[sig-network][Feature:Router][apigroup:route.openshift.io][apigroup:config.openshift.io]", func() {
```
deads2k, line 24: this one too.

test/extended/security/scc.go
```diff
@@ -30,7 +30,7 @@ var _ = g.Describe("[sig-auth][Feature:SecurityContextConstraints] ", func() {
 	oc := exutil.NewCLIWithPodSecurityLevel("scc", admissionapi.LevelPrivileged)
 	ctx := context.Background()
 
-	g.It("TestPodUpdateSCCEnforcement", func() {
+	g.It("TestPodUpdateSCCEnforcement [apigroup:user.openshift.io][apigroup:authorization.openshift.io]", func() {
```
deads2k, line 33: This is going to bite us.  We'll need to fix these too to ensure SCC functions.  We'll need a way to handle different users.  Perhaps serviceaccounts instead of regular users.

test/extended/security/supplemental_groups.go
```diff
@@ -34,7 +34,7 @@ var _ = g.Describe("[sig-node] supplemental groups", func() {
 	)
 
 	g.Describe("Ensure supplemental groups propagate to docker", func() {
-		g.It("should propagate requested groups to the container [Local]", func() {
+		g.It("should propagate requested groups to the container [Local][apigroup:user.openshift.io][apigroup:security.openshift.io]", func() {
```
deads2k, line 37: and these.

deads2k:
we're going to have followups, but this is a good start.

/lgtm

PR #27360, opened 2022-08-17
WIP: Hard fail for non-matching network-type vs. job-name, or wrong periodic job name

DennisPeriquet:
Just trying a possible implementation of [TRT-470](https://issues.redhat.com//browse/TRT-470).

Looking at the failures, I'm not sure if this is what we want -- feel free to re-implement and close as needed or we can update this one.

### DIFF
--- a/test/extended/ci/job_names.go
+++ b/test/extended/ci/job_names.go
@@ -2,6 +2,7 @@ package ci
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


DennisPeriquet:
closed in favor of https://github.com/openshift/origin/pull/27362

PR #27361, opened 2022-08-18
USHIFT-283: microshift: skip sig-builds tests

pacevedom:
Builds API is not available in microshift.

Closes [USHIFT-283](https://issues.redhat.com//browse/USHIFT-283)

### DIFF
--- a/test/extended/builds/build_pruning.go
+++ b/test/extended/builds/build_pruning.go
@@ -189,8 +189,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds] prune builds based on settings 
 
 		})
 
-		g.It("should prune errored builds based on the failedBuildsHistoryLimit setting", func() {
-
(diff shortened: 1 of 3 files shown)


mfojtik:
/lgtm

if OAS aggregation breaks, I would expect more things will explode, so these tests won't be skipped unexpectedly

pacevedom:
> /lgtm
> 
> if OAS aggregation breaks, I would expect more things will explode, so these tests won't be skipped unexpectedly

I could picture the OCP api server not working but the aggregated APIs being configured, or nothing at all.
I am not familiar with OAS, but does it turn ready only when all aggregated APIs have been configured? If this is true then we should be ok.

pacevedom:
/hold
Need to rework.

pacevedom:
/unhold
Ready for review
/assign @ingvagabund

PR #27367, opened 2022-08-19
USHIFT-300: microshift: skip sig-imageregistry tests

pacevedom:
Skip sig-imageregistry tests if image.openshift.io API is not registered.

### DIFF
--- a/test/extended/images/info.go
+++ b/test/extended/images/info.go
@@ -23,7 +23,7 @@ var _ = g.Describe("[sig-imageregistry][Feature:ImageInfo] Image info", func() {
 
 	oc = exutil.NewCLIWithPodSecurityLevel("image-info", admissionapi.LevelBaseline)
 
-	g.It("should display information about images", func() {
+	g.It("should display information about images [apigroup:image.openshift.io]", func() {
 		ns = oc.Namespace()
 		cli := oc.KubeFramework().PodClient()
 		pod := cli.Create(cliPodWithPullSecret(oc, heredoc.Docf(`
--- a/test/extended/imageapis/imagesigning.go
+++ b/test/extended/imageapis/imagesigning.go
@@ -16,7 +16,6 @@ import (
 
 	g "github.com/onsi/ginkgo"
 	imagev1 "github.com/openshift/api/image/v1"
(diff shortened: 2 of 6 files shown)


pacevedom:
/hold
Need to rework.

pacevedom:
/unhold
Ready for review
/assign @ingvagabund

test/extended/images/info.go
```diff
@@ -21,6 +22,17 @@ var _ = g.Describe("[sig-imageregistry][Feature:ImageInfo] Image info", func() {
 		}
 	})
 
+	g.BeforeEach(func() {
+		microShift, err := exutil.IsMicroShiftDeployment(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if microShift {
+			// Test works as oc image info does not require any APIs, but the pod from
+			// where this is executed is a local image to the cluster, which is not
```
ingvagabund, line 30: Any way to pull the image to the cluster?

pacevedom: We would need to promote an image with oc inside to a public registry, not sure about that. Also, I wonder if this test should be in sig-cli, there are no checks in this one besides the pod coming up.

pacevedom:
/hold
Need to get rid of microshift checks

pacevedom:
/unhold
Using apigroups now.

test/extended/images/info.go
```diff
@@ -23,7 +23,7 @@ var _ = g.Describe("[sig-imageregistry][Feature:ImageInfo] Image info", func() {
 
 	oc = exutil.NewCLIWithPodSecurityLevel("image-info", admissionapi.LevelBaseline)
 
-	g.It("should display information about images", func() {
+	g.It("should display information about images [apigroup:image.openshift.io]", func() {
```
ingvagabund, line 26: Does the image apigroup requirement come from `cliPodWithPullSecret`? If so, is builder SA also implying build apigroup?

pacevedom: The image this test uses holds `oc` binary in it and it comes from the internal registry. This is only working when this apigroup is installed.

soltysh (approved):
/approve

PR #27372, opened 2022-08-22
Add ovnkube-node pods in the interval chart for ingress

DennisPeriquet:
Re: [TRT-475](https://issues.redhat.com//browse/TRT-475)

This adds the openshift-ovn-kubernetes namespace which will show the ovnkube-node pods.


dgoodwin:
cc @trozet as requested!

PR #27368, opened 2022-08-19
Label tests with openshift crafted API groups (round 3)

ingvagabund:
Built on top of https://github.com/openshift/origin/pull/27359

Procedure for detecting API groups in use:
- search for all Get, Update, Patch, Delete and List client invocations
  (for both direct and indirect)
- go through every e2e test code and identify all possible invocations of an api group (e.g. oc command)

jira: https://issues.redhat.com/browse/WRKLDS-488

PR #27378, opened 2022-08-25
Revert "Label tests with openshift crafted API groups (round 3)"

dgoodwin:
Reverts openshift/origin#27368

[TRT-495](https://issues.redhat.com//browse/TRT-495)

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are reverting this breaking change to get ci and/or nightly payloads flowing again.

Example: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-metal-ipi-ovn-ipv6/1562677330093019136

4.12 nightly payloads appear blocked, I believe because this change renamed HAProxy tests that are explicitly handled in the release repo, most likely here: https://github.com/openshift/release/blob/031c233818f7911be6b10ad058aadd5ff3ad1885/ci-operator/config/openshift/router/openshift-router-master.yaml#L106

Normally we'd ask you to bring this back in by unreverting this change and layering in a separate commit, but in this case the fix is likely required in the release repo. In this case the best course of action would be to unrevert, open the new PR, and run /payload 4.12 nightly blocking. The release repo may need to list *both* new and old test names for a bit so this rolls out smoothly.

CC: @deads2k @ingvagabund

### DIFF
--- a/test/extended/cli/debug.go
+++ b/test/extended/cli/debug.go
@@ -28,7 +28,7 @@ var _ = g.Describe("[sig-cli] oc debug", func() {
 	helloPod := exutil.FixturePath("..", "..", "examples", "hello-openshift", "hello-pod.json")
 	imageStreamsCentos := exutil.FixturePath("..", "..", "examples", "image-streams", "image-streams-centos7.json")
 
-	g.It("deployment configs from a build [apigroup:image.openshift.io][apigroup:apps.openshift.io]", func() {
+	g.It("deployment configs from a build", func() {
 		err := oc.Run("create").Args("-f", testCLIDebug).Execute()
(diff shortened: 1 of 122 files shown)


dgoodwin:
I'm not certain I linked to the right spot where this is happening in release repo, but based on: https://testgrid.k8s.io/redhat-openshift-ocp-release-4.12-blocking#periodic-ci-openshift-release-master-nightly-4.12-e2e-metal-ipi-ovn-ipv6&show-stale-tests=

We can see these HAProxy tests were not running prior to the rename, I am proceeding with force revert.

PR #27389, opened 2022-09-01
USHIFT-348: microshift: skip and fix sig-cli tests

pacevedom:


test/extended/cli/apiresources.go
```diff
@@ -41,6 +38,27 @@ var _ = g.Describe("[sig-cli] oc api-resources", func() {
...
 
+		out, err = oc.Run("api-versions").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(out).To(o.ContainSubstring("apps/v1"))
+		o.Expect(out).To(o.ContainSubstring("networking.k8s.io/v1"))
+		o.Expect(out).To(o.ContainSubstring("node.k8s.io/v1"))
+		o.Expect(out).To(o.ContainSubstring("storage.k8s.io/v1"))
+	})
+
+	g.It("can output expected information about openshift api-resources [apigroup:build.openshift.io][apigroup:project.openshift.io][apigroup:user.openshift.io]", func() {
```
dhellmann, line 49: I'm not sure about this part of the change. Logically it's saying that if the API exists then it should show up in the output from api-resources, but aren't we using the same query to see if it exists that oc uses to produce its output? Is that a useful test?

I know we've said we didn't want to generally, but this feels like a case where we _should_ say "If the cluster is MicroShift, skip" so that we can have a separate test that checks for the expected APIs when the cluster is MicroShift.

test/extended/cli/explain.go
```diff
@@ -54,6 +53,10 @@ var (
 		{Group: "template.openshift.io", Version: "v1", Resource: "templateinstances"},
 	}
 
+	microShiftBuiltinTypes = []schema.GroupVersionResource{
```
dhellmann, line 56: How hard would it be for us to put these microshift-specific tests in the microshift repo? If not now, eventually?

I know @bparees has been looking at splitting up the tests, so maybe we can be an early adopter of that work?

pacevedom: Have local e2e tests? I have seen other operators do it before.

pacevedom:
/hold
Need to rework.

pacevedom:
/unhold
Ready for review
/assign @ingvagabund

test/extended/cli/apiresources.go
```diff
@@ -41,6 +34,25 @@ var _ = g.Describe("[sig-cli] oc api-resources", func() {
...
```
ingvagabund, line 45: To increase the flexibility it's worth introducing new `g.It` for each apigroup. E.g.:
```go
g.It("can output expected information about openshift build api-resources [apigroup:build.openshift.io]", func() {
   ...
}
g.It("can output expected information about openshift image api-resources [apigroup:image.openshift.io]", func() {
   ...
}
g.It("can output expected information about openshift project api-resources [apigroup:project.openshift.io]", func() {
   ...
}
g.It("can output expected information about openshift user api-resources [apigroup:user.openshift.io]", func() {
   ...
}
```

test/extended/cli/observe.go
```diff
@@ -107,4 +103,10 @@ var _ = g.Describe("[sig-cli] oc observe", func() {
 		o.Expect(out).To(o.ContainSubstring("Sync should_be_passed"))
 		o.Expect(os.Unsetenv("MYENV")).NotTo(o.HaveOccurred())
 	})
+
+	g.It("works as expected with cluster operators [apigroup:config.openshift.io]", func() {
```
ingvagabund, line 107: Why separate when the original `works as expected [apigroup:config.openshift.io]` name still keeps the apigroup?

pacevedom: Done

test/extended/cli/apiresources.go
pacevedom: Done

test/extended/cli/explain.go
```diff
@@ -416,27 +447,47 @@ var (
...
+		},
+		{
+			gv:      schema.GroupVersion{Group: "security.openshift.io", Version: "v1"},
+			field:   "securitycontextconstraints",
+			pattern: `FIELDS\:.*`,
+		},
 	}
+)
 
+func getCrdTypes(oc *exutil.CLI) []schema.GroupVersionResource {
```
ingvagabund, line 470: `getCrdTypes` is invoked by two tests:
- "list uncovered GroupVersionResources"
- "should contain proper spec+status for CRDs [apigroup:config.openshift.io]"

The second already assumes existence of the config apigroup which is getting detected inside `getCrdTypes`.

`getCrdTypes` is always constructing the list of crd types from `baseCRDTypes`, `autoscalingTypes`, `machineTypes` and `additionalOperatorTypes` which belong to their corresponding groups as well. Which at least requires to label both tests with the corresponding groups as well. Making both tests as another place for refactoring and looping through through the apigroups.

@pacevedom this tests requires more refactoring than anticipated. Though, better to do it properly now than later/never. Nevertheless, very good progress here.

ingvagabund (changes requested):

test/extended/cli/basics.go
```diff
@@ -198,9 +198,19 @@ var _ = g.Describe("[sig-cli] oc basics", func() {
 			ok := rootCAs.AppendCertsFromPEM(kubeCA)
 			o.Expect(ok).To(o.BeTrue())
 
+			var certs []tls.Certificate
```
ingvagabund, line 201: Can you drop a comment here why this is needed for MicroShift?

test/extended/cli/explain.go
```diff
@@ -496,12 +551,22 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
...
-	})
+	for group, bts := range builtinTypes {
+		g.It(fmt.Sprintf("should contain spec+status for %s [apigroup:%s]", group, group), func() {
+			for _, bt := range bts {
+				o.Expect(verifySpecStatusExplain(oc, nil, bt)).NotTo(o.HaveOccurred())
+			}
+		})
+	}
+
+	for group, bts := range builtinTypes {
```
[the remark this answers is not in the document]
pacevedom: `builtinTypes` is defined as `map[string][]schema.GroupVersionResource`. This loop takes the apigroup as `group` and the list is assigned to `bts`. This is passed to `verifySpecStatusExplain` which takes a `schema.GroupVersionResource` as the last argument.

test/extended/cli/basics.go
pacevedom: Done

test/extended/cli/explain.go
pacevedom: "should contain proper spec+status for CRDs [apigroup:config.openshift.io]" requires config API because it checks for control plane topology in order to select different CRDs.
The nature of this test is to check explicitly for presence of those APIs that get built in `getCRDTypes`. If they are not present (or are poorly defined) then that should be an error. Maybe we should have a MicroShift specific test with the CRDs we install there instead? To me it is not only about checking that whatever is present is ok, it is also about specifying what should be installed.

"list uncovered GroupVersionResources" goes after API groups not covered in this file to list them. It does not perform any checks besides producing logs for those uncovered.

test/extended/cli/explain.go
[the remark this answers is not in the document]
ingvagabund: Both
```go
for group, bts := range builtinTypes {
		g.It(fmt.Sprintf("should contain spec+status for %s [apigroup:%s]", group, group), func() {
			for _, bt := range bts {
				o.Expect(verifySpecStatusExplain(oc, nil, bt)).NotTo(o.HaveOccurred())
			}
		})
	}
```
and
```go
for group, bts := range builtinTypes {
	g.It(fmt.Sprintf("should contain spec+status for %s [apigroup:%s]", group, group), func() {
			for _, bt := range bts {
				o.Expect(verifySpecStatusExplain(oc, nil, bt)).NotTo(o.HaveOccurred())
			}
		})
	}
```
are identical up to `e2e.Logf("Checking %s...", bt) line unless I am overlooking something.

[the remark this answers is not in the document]
pacevedom: Oh wait, misunderstood you. You are absolutely right, should be fixed now.

test/extended/cli/basics.go
```diff
@@ -198,9 +198,22 @@ var _ = g.Describe("[sig-cli] oc basics", func() {
...
+				cert, err := tls.X509KeyPair(certData, keyData)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				certs = []tls.Certificate{cert}
+			}
+
 			transport = &http.Transport{
 				TLSClientConfig: &tls.Config{
-					RootCAs: rootCAs,
+					RootCAs:      rootCAs,
+					Certificates: certs,
```
ingvagabund, line 216: @pacevedom does this change automatically implies OAuth tokens will no longer be used? Is there any way to detect whether OAuth tokens are used and generate list of certs only when they are not?

pacevedom: Thanks for having a look, this was a leftover from before all the changes to the oc project impersionation were complete. Now its not needed so I remove it.

test/extended/cli/explain.go
ingvagabund: `should contain proper spec+status for CRDs [apigroup:config.openshift.io]` checks presence of explanation text for corresponding CRD's spec and status.

> The nature of this test is to check explicitly for presence of those APIs that get built in getCRDTypes. If they are not present (or are poorly defined) then that should be an error.

The test does not check whether a cluster has a required list of CRDs present.  Unless I am missing the specific line.
The test checks whether CRDs listed by `getCRDTypes` have the spec and status properly described when running `oc explain` command. I.e. testing implication (is a given crd present -> has the crd spec/status properly explained). For that the list of CRDs can still be broken down based on the apigroup and each apigroup test separately. If there's a need to explicitly check for a presence of a specific list of CRDs, new test needs to be written.

> To me it is not only about checking that whatever is present is ok, it is also about specifying what should be installed.

I don't think this is the case here.

> "list uncovered GroupVersionResources" goes after API groups not covered in this file to list them. It does not perform any checks besides producing logs for those uncovered.

Ack. Reading the code a second time `resourceMap` gets populated from what's installed on a cluster minutes what's provided by `getCrdTypes`. Taking my comment back. Thanks for the pointers.

ingvagabund:
Merged through https://github.com/openshift/origin/pull/27498
/close

PR #27394, opened 2022-09-02
skip route-controller-manager scaling events in build testsuite

atiratree:
Skips events that are triggered by the build tests: https://search.ci.openshift.org/?search=ns%2Fopenshift-route-controller-manager+deployment%2Froute-controller-manager&maxAge=336h&context=1&type=junit&name=pull-ci-openshift-origin-master-e2e-gcp-builds&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job 

more background in relevant issues: 
- https://issues.redhat.com/browse/BUILD-495
- https://bugzilla.redhat.com/show_bug.cgi?id=2034984

### DIFF
--- a/pkg/synthetictests/duplicated_events.go
+++ b/pkg/synthetictests/duplicated_events.go
@@ -185,9 +185,16 @@ var knownEventsBugs = []knownProblem{
 		BZ:       "https://bugzilla.redhat.com/show_bug.cgi?id=2017435",
 		Topology: topologyPointer(v1.SingleReplicaTopologyMode),
 	},
+	// builds tests trigger many changes in the config which creates new rollouts -> event for each pod
+	// working as intended (not a bug) and needs to be tolerated
+	{
+		Regexp:    regexp.MustCompile(`ns/openshift-route-controller-manager deployment/route-controller-manager - reason/ScalingReplicaSet \(combined from similar events\): Scaled (down|up) replica set route-controller-manager-[a-z0-9-]+ to [0-9]+`),
+		TestSuite: stringPointer("openshift/build"),
+	},
+	// builds tests trigger many changes in the config which creates new rollouts -> event for each pod
+	// working as intended (not a bug) and needs to be tolerated
 	{
 		Regexp:    regexp.MustCompile(`ns/openshift-controller-manager daemonset/controller-manager - reason/SuccessfulDelete \(combined from similar events\): Deleted pod: controller-manager-[a-z0-9-]+`),
-		BZ:        "https://bugzilla.redhat.com/show_bug.cgi?id=2034984",
 		TestSuite: stringPointer("openshift/build"),
 	},
 	//{ TODO this should only be skipped for single-node


pkg/synthetictests/duplicated_events.go
```diff
@@ -185,6 +185,11 @@ var knownEventsBugs = []knownProblem{
...
 		Topology: topologyPointer(v1.SingleReplicaTopologyMode),
 	},
+	// builds tests trigger many changes in the config which creates new rollouts -> event for each pod
+	{
+		Regexp:    regexp.MustCompile(`ns/openshift-route-controller-manager deployment/route-controller-manager - reason/ScalingReplicaSet \(combined from similar events\): Scaled down replica set route-controller-manager-[a-z0-9-]+ to [0-9]+`),
+		TestSuite: stringPointer("openshift/build"),
+	},
 	{
 		Regexp:    regexp.MustCompile(`ns/openshift-controller-manager daemonset/controller-manager - reason/SuccessfulDelete \(combined from similar events\): Deleted pod: controller-manager-[a-z0-9-]+`),
 		BZ:        "https://bugzilla.redhat.com/show_bug.cgi?id=2034984",
```
atiratree, line 195: this was originally bug, but I think we could remove the BZ now

pkg/synthetictests/duplicated_events.go
```diff
@@ -185,6 +185,11 @@ var knownEventsBugs = []knownProblem{
 		BZ:       "https://bugzilla.redhat.com/show_bug.cgi?id=2017435",
 		Topology: topologyPointer(v1.SingleReplicaTopologyMode),
 	},
+	// builds tests trigger many changes in the config which creates new rollouts -> event for each pod
+	{
+		Regexp:    regexp.MustCompile(`ns/openshift-route-controller-manager deployment/route-controller-manager - reason/ScalingReplicaSet \(combined from similar events\): Scaled down replica set route-controller-manager-[a-z0-9-]+ to [0-9]+`),
+		TestSuite: stringPointer("openshift/build"),
```
atiratree, line 191: I have kept this in the `knownEventsBugs` section since it is the only one that distignuishes by `TestSuite` even though it is not a bug.

Should I do a refactoring and move it somewhere else?

bparees: i think this is fine.  You might make it clearer in the comment that this is not considered a bug, but rather "working as deisgned and needs to be tolerated" though.

Are the tests in question *only* part of the `openshift/build` suite?

atiratree: @bparees good suggestion - updated

we are observing these only in builds suite atm, but it could occur to anybody who does a lot of changes to either of

- `builds.config.openshift.io/cluster`
- `images.config.openshift.io/cluster`
- `configmap/openshift-controller-manager-images` in `openshift-controller-manager-operator` ns
- `openshiftcontrollermanagers.operator.openshift.io/cluster`

pkg/synthetictests/duplicated_events.go
bparees: given the BZ has been closed(as not a bug), we should remove it.

atiratree: removed

pkg/synthetictests/duplicated_events.go
```diff
@@ -185,9 +185,16 @@ var knownEventsBugs = []knownProblem{
...
 		Topology: topologyPointer(v1.SingleReplicaTopologyMode),
 	},
+	// builds tests trigger many changes in the config which creates new rollouts -> event for each pod
+	// working as intended (not a bug) and needs to be tolerated
+	{
+		Regexp:    regexp.MustCompile(`ns/openshift-route-controller-manager deployment/route-controller-manager - reason/ScalingReplicaSet \(combined from similar events\): Scaled (down|up) replica set route-controller-manager-[a-z0-9-]+ to [0-9]+`),
+		TestSuite: stringPointer("openshift/build"),
+	},
+	// builds tests trigger many changes in the config which creates new rollouts -> event for each pod
+	// working as intended (not a bug) and needs to be tolerated
```
deads2k, line 195: If using a daemonset that causes failures during draining as leases are not gracefully released and instead age out as pods are killed after potentially losing network access due to daemonset pods not being terminated is not a bug, what is it?  We know the better solution and fixed other components like the openshift-apiserver to use it.

bparees: What about this event implies "failures during draining" or leases not being gracefully released?

The event in question is that a pod gets deleted, which is a result of the cluster config changing(as explicitly changed by a test) and so the controller manager does a rollout in response to the config change.

The test seems to assume we should never see those pods deleted (which, at steady state w/ no config changes, is probably true)

ingvagabund: @atiratree this might be worth reporting as a bug which we might address later on.

PR #27397, opened 2022-09-05
USHIFT-344: MicroShift: fix sig-apps tests

pacevedom:
Fixes `[sig-apps][Feature:Jobs] Users should be able to create and run a job in a user project [Skipped:Disconnected] [Suite:openshift/conformance/parallel]` to also run in MicroShift.


test/extended/deployments/deployments.go
```diff
@@ -290,8 +290,14 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 		})
 
 		g.It("resolve the image pull spec", func() {
+			isMicroShift, err := exutil.IsMicroShiftDeployment(oc)
```
deads2k, line 293: same question about the list of skipped tests and why the apigroup isn't appropriate.

pacevedom:
/hold
Need to rework.

pacevedom:
/unhold
Ready for review
/assign @ingvagabund

ingvagabund:
@pacevedom quite nice improvement. Would you please share the original error that was manifesting in MicroShift?

pacevedom:
> @pacevedom quite nice improvement. Would you please share the original error that was manifesting in MicroShift?

It was a pod failing to be ready because the image was never pulled, thus a timeout in the test. It showed up in the logs for the failure.

ingvagabund:
So `image-registry.openshift-image-registry.svc:5000/openshift/tools:latest` was never available since the image registry is not deployed in MicroShift

PR #27398, opened 2022-09-05
OCPBUGS-559: update signature of mapiterinit of go 1.18

ingvagabund:
To pull https://github.com/modern-go/reflect2/pull/25. Otherwise running `oc login` produces panic when build with go >= 1.18:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x4145bc]

goroutine 1 [running]:
reflect.mapiternext(0x40?)
	/usr/lib/golang/src/runtime/map.go:1378 +0x19
github.com/openshift/origin/vendor/github.com/modern-go/reflect2.(*UnsafeMapIterator).UnsafeNext(0x0?)
	/builddir/build/BUILD/origin-20c5b86c88657888e4906ed7942b85515c650f96/_output/local/go/src/github.com/openshift/origin/vendor/github.com/modern-go/reflect2/unsafe_map.go:136 +0x32
...
```

Running `glide up` over release-3.11 head produces changes in the vendor directory. Meaning the glide has not been used for a while. Thus, updating the vendor directory without it.

### DIFF
--- a/vendor/github.com/modern-go/reflect2/unsafe_link.go
+++ b/vendor/github.com/modern-go/reflect2/unsafe_link.go
@@ -19,7 +19,7 @@ func typedslicecopy(elemType unsafe.Pointer, dst, src sliceHeader) int
 
 //go:linkname mapassign reflect.mapassign
 //go:noescape
-func mapassign(rtype unsafe.Pointer, m unsafe.Pointer, key, val unsafe.Pointer)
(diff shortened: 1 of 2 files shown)


ingvagabund:
/close

oc command in 3.11 is not expected to be supported with the latest go versions.

PR #27400, opened 2022-09-06
USHIFT-350: MicroShift: skip sig-coreos tests

pacevedom:


test/extended/coreos/coreos.go
```diff
@@ -28,6 +28,15 @@ var _ = g.Describe("[sig-coreos] [Conformance] CoreOS bootimages", func() {
...
+
+	g.BeforeEach(func() {
+		microShift, err := exutil.IsMicroShiftDeployment(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if microShift {
+			g.Skip("Skipped in MicroShift")
+		}
+	})
+
 	g.It("TestBootimagesPresent", func() {
```
deads2k, line 40: Are these present in hypershift?  I don't know what they are, but if every other form-factor of OCP has these, why shouldn't microshift?

pacevedom:
/hold
Need to rework.

test/extended/coreos/coreos.go
pacevedom: These are the RHCOS images for the node and where to get them depending on the platform. Microshift does not run on RHCOS and does not have machine config operator either, so this test may be skipped.

pacevedom:
/unhold
Ready for review. Depends on https://github.com/openshift/origin/pull/27345
/assign @ingvagabund

pacevedom:
/hold
Need to get of microshift checks

pacevedom:
/unhold
Swapped the microshift check for MCO CRDs presence, as the test is based in resources this operator creates.

test/extended/coreos/coreos.go
```diff
@@ -28,7 +28,8 @@ var _ = g.Describe("[sig-coreos] [Conformance] CoreOS bootimages", func() {
 	defer g.GinkgoRecover()
 
 	oc := exutil.NewCLIWithoutNamespace("coreos")
-	g.It("TestBootimagesPresent", func() {
+
+	g.It("TestBootimagesPresent [apigroup:machineconfiguration.openshift.io]", func() {
```
ingvagabund, line 32: `machineconfiguration` is not existing group under openshift/api

ingvagabund, line 32: Checking https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27400/pull-ci-openshift-origin-master-e2e-aws-ovn-cgroupsv2/1573331257285152768 the test is still running

pacevedom: This comes from the CRD that MCO installs. If its not there then there is certainly no MCO.

soltysh (approved):
/approve

PR #27401, opened 2022-09-06
USHIFT-349: MicroShift: skip sig-cluster-lifecycle tests

pacevedom:


test/extended/csrapprover/csrapprover.go
```diff
@@ -37,6 +37,14 @@ var _ = g.Describe("[sig-cluster-lifecycle]", func() {
 	oc := exutil.NewCLIWithPodSecurityLevel("cluster-client-cert", admissionapi.LevelBaseline)
 	defer g.GinkgoRecover()
 
+	g.BeforeEach(func() {
```
deads2k, line 40: Can you enumerate the tests being skipped and why an apigroup label isn't more appropriate?  

Also, I think `"CSRs from machines that are not recognized by the cloud provider are not approved"` should pass and if it doesn't, that seems to suggest we have a bug in microshift.

pacevedom:
/hold
Need to rework.

pacevedom:
Closing as there are no more changes required once the tests have been tagged.

PR #27402, opened 2022-09-06
USHIFT-357: MicroShift: skip sig-scheduling tests

pacevedom:


test/extended/scheduling/pods.go
```diff
@@ -140,6 +140,14 @@ var _ = g.Describe("[sig-scheduling][Early]", func() {
 
 	oc := exutil.NewCLI("scheduling-pod-check")
 
+	g.BeforeEach(func() {
```
deads2k, line 143: A more proper skip would be based on the number of nodes, right?

pacevedom:
/hold
Need to rework.

pacevedom:
/unhold
Ready for review
/assign @ingvagabund

test/extended/scheduling/pods.go
pacevedom: Changed the test to skip a `BeforeEach` if its MicroShift, as it was the cause for failure. It was checking for an ingress that is not present in the cluster, and the test was already prepared to ignore missing deployments.

ingvagabund: What about to have each `g.Describe` extended with relevant `[apigroup:XXX]` given each namespace deploys apigroup related CRD?

pacevedom: Done. Tried changing the WaitForRouterServiceIP in order to support MicroShift, but it does not help with these tests (since they are passing anyway, there is only one node). It does not help either with other tests from router API, as they have other requirements that make them fail in microshift.

ingvagabund: I don't think you need `[apigroup:operator.openshift.io]` related crd to be installed. Though, when running individual deployments/pods (e.g. `router-default`) there should be at least once primary CRD/apigroup as a requirement for running the deployment:
- `router-default` most likely assumes existence of router.openshift.io
- `apiserver` will have many apigroups (e.g. build, image, ...)
- `oauth-openshift` will be primarily oauth.openshift.io plus maybe user?
- ...

The operands do not need operator.config.io apigroup installed. A different controller might deploy the operands. So I still prefer to label individual deployment `g.Describe` bits:
- `g.Describe("The HAProxy router pods"`
- `g.Describe("The openshift-apiserver pods"`
- ...

PR #27404, opened 2022-09-07
USHIFT-346: MicroShift: skip sig-auth tests

pacevedom:


pacevedom:
/hold
Need to rework.

pacevedom:
/unhold
Ready for review. Depends on https://github.com/openshift/origin/pull/27345
/assign @ingvagabund

test/extended/util/client.go
```diff
@@ -519,10 +519,6 @@ func (c *CLI) AdminQuotaClient() quotav1client.Interface {
 	return quotav1client.NewForConfigOrDie(c.AdminConfig())
 }
 
-func (c *CLI) AdminOAuthClient() oauthv1client.Interface {
```
ingvagabund, line 522: @stlaz which of `AdminOAuthClient` and `AdminOauthClient` is the right name for the method?

test/extended/security/scc.go
```diff
@@ -271,7 +271,10 @@ var _ = g.Describe("[sig-auth][Feature:SecurityContextConstraints] ", func() {
 
 	g.It("TestPodDefaultCapabilities", func() {
 		g.By("Running a restricted pod and getting it's inherited capabilities")
-		pod, err := exutil.NewPodExecutor(oc, "restrictedcapsh", image.ShellImage())
+		// This test should use image.ShellImage but this requires having a local image
+		// registry, which not all deployment types have. Using LDAP test image guarantees
+		// capsh presence.
+		pod, err := exutil.NewPodExecutor(oc, "restrictedcapsh", image.OpenLDAPTestImage())
```
ingvagabund, line 277: @stlaz for taking a look

test/extended/oauth/server_headers.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/openshift/origin/test/extended/util/oauthserver"
 )
 
-var _ = g.Describe("[sig-auth][Feature:OAuthServer] [Headers][apigroup:route.openshift.io]", func() {
+var _ = g.Describe("[sig-auth][Feature:OAuthServer] [Headers][apigroup:route.openshift.io][apigroup:config.openshift.io]", func() {
```
ingvagabund, line 20: I see. `deployOAuthServer` -> `DeployOAuthServer` -> `getImage` -> `determineImageFromRelease` -> `DetermineImageFromRelease`

Also `deployOAuthServer` invokes `oauthserver.DeployOAuthServer` which invokes `createOAuthClient` which uses `OauthV1`. So at least `[apigroup:oauth.openshift.io]` is needed as well.

ingvagabund (changes requested):

test/extended/util/client.go
pacevedom: Original name was `AdminOauthClient`, the other one was added later and the code for both functions was exactly the same. Thought I should unify it.

test/extended/oauth/server_headers.go
pacevedom: Nice catch, this one slipped.

test/extended/util/client.go
ingvagabund: @pacevedom let's do `AdminOAuthClient`.

test/extended/oauth/http1.go
```diff
@@ -17,6 +17,15 @@ var _ = g.Describe("[sig-auth][Feature:OAuthServer] OAuth server", func() {
 
 	oc := exutil.NewCLI("oauth")
 
+	g.BeforeEach(func() {
+		microShift, err := exutil.IsMicroShiftDeployment(oc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if microShift {
+			// MicroShift does not have the oauth well known data path in the API Server
+			g.Skip("Skipped in MicroShift")
```
ingvagabund, line 25: Would it make more sense to extend `g.Describe` with `[apigroup:oauth.openshift.io]` instead?

test/extended/util/client.go
pacevedom: Done

test/extended/oauth/http1.go
pacevedom: Done

pacevedom:
/hold
Need to get of microshift checks

test/extended/authorization/rbac/groups_default_rules.go
```diff
@@ -182,47 +183,60 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] The default clust
...
+			projectGroup,
+			authzGroup,
+			security.GroupName,
+			consoleGroup,
+			"helm.openshift.io",
+			templateGroup,
+			operatorsCoreOSGroup,
+			imageGroup,
+		),
+		func() {
```
ingvagabund, line 199: Would you please put `func() {` right after the previous line as `(, func() {`? To reduce the diff. If that does not help you might move `fmt.Sprintf` next to `g.It(`.

test/extended/authorization/rbac/groups_default_rules.go
```diff
@@ -182,47 +183,60 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] The default clust
 
 	oc := exutil.NewCLI("default-rbac-policy")
 
-	g.It("should have correct RBAC rules", func() {
-		kubeInformers := informers.NewSharedInformerFactory(oc.AdminKubeClient(), 20*time.Minute)
-		ruleResolver := exutil.NewRuleResolver(kubeInformers.Rbac().V1()) // signal what informers we want to use early
+	g.It(
+		fmt.Sprintf("should have correct RBAC rules [apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s]",
```
ingvagabund, line 187: This is quite complex. `allUnauthenticatedRules` is the source of so many groups. I'd rather turn `allUnauthenticatedRules` into a api-group-key map and loop over all apigroups over `g.It("should have correct RBAC rules", func() {` rather than have all groups in a single test name.

test/extended/oauth/http1.go
```diff
@@ -12,7 +12,7 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
-var _ = g.Describe("[sig-auth][Feature:OAuthServer] OAuth server", func() {
+var _ = g.Describe("[sig-auth][Feature:OAuthServer] OAuth server [apigroup:auth.openshift.io]", func() {
```
ingvagabund, line 15: Is the requirement coming from `oc get --raw /.well-known/oauth-authorization-server`?

test/extended/security/scc.go
ingvagabund: @openshift/openshift-team-auth PTAL

ingvagabund (changes requested):

test/extended/security/scc.go
ingvagabund: @pacevedom is there a less complex image than LDAP? Something on the same level of complexity as the shell image? Complexity = minimum rpm installation/configuration inside the image. I assume you choosed OpenLDAPTestImage since it's publicly available from `quay.io/openshifttest/ldap:1.2`. I wonder whether we run these tests in a disconnected environment.

test/extended/authorization/rbac/groups_default_rules.go
pacevedom: Done

test/extended/authorization/rbac/groups_default_rules.go
pacevedom: Mind if we change this in the next PR?

ingvagabund: Making the change later would put the next work into backlog most likely. Better to do it now.

test/extended/authorization/rbac/groups_default_rules.go
```diff
@@ -182,7 +183,18 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] The default clust
 
 	oc := exutil.NewCLI("default-rbac-policy")
 
-	g.It("should have correct RBAC rules", func() {
+	g.It(fmt.Sprintf("should have correct RBAC rules [apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s][apigroup:%s]",
```
ingvagabund, line 186: This is quite complex. `allUnauthenticatedRules` is the source of so many groups. I'd rather turn `allUnauthenticatedRules` into a api-group-key map and loop over all apigroups over `g.It("should have correct RBAC rules", func() {` rather than have all groups in a single test name.

stlaz: I don't think it's safe to break the test apart. I am not sure why these groups are listed here though, because you don't actually need the API groups, you only need the RBAC group, the rest are just strings.

ingvagabund: You are right. Also, I don't see an easy way how to run the test minus missing apigroups since the test is doing set diff of both "what's currently available" and "what's expected to be available".

ingvagabund: Obvious approach is to dynamically detect the apigroups based on `exutil.DoesApiResourceExist`, constructing both `allAuthenticatedRules` and `allUnauthenticatedRules` based on the apigroup presence and then running the comparision/coverage. Though, there's more than just api group detection which needs to be taken into account. E.g. https://github.com/openshift/origin/blob/master/test/extended/authorization/rbac/groups_default_rules.go#L91-L94.

https://github.com/openshift/origin/blob/master/test/extended/authorization/rbac/groups_default_rules.go#L97-L100 can be removed. Maybe the legacy groups in https://github.com/openshift/origin/blob/master/test/extended/authorization/rbac/groups_default_rules.go#L59-L66 as well? What about https://github.com/openshift/origin/blob/master/test/extended/authorization/rbac/groups_default_rules.go#L134-L142?

pmtk:
```
{  Cluster accessed images that were not mirrored to the testing repository or already part of the cluster, see test/extended/util/image/README.md in the openshift/origin repo:
> quay.io/redhat-developer/test-build-simples2i:1.2 from pods:
  ns/e2e-test-ssc-pmk7k pod/restrictedcapsh node/ci-op-d9w0tdfn-fcf39-rqtdm-worker-a-m2nhg
}
```

ingvagabund:
`image.LocationFor("quay.io/redhat-developer/test-build-simples2i:1.2")` (from `"github.com/openshift/origin/test/extended/util/image"`) will do the trick

test/extended/authorization/rbac/groups_default_rules.go
pmtk: I'm taking a second look at this. It look like @stlaz is right, all [apigroup:*] seems wrong. Currently it complains about test data being too big which a nice mechanism except not right against microshift

pmtk: Reverted these changes as this wasn't exercising [apigroups] - it was exercising RBAC about those apigroups so the test should run on MicroShift. If it fails then we should  take a look what should we fix in MicroShift and how should we refactor test if needed

pmtk:
> /retest-required

It's no use, `ci/prow/e2e-gcp-ovn-image-ecosystem` seems to fail on many PRs on the same tests, see: https://github.com/openshift/origin/pull/27502#issuecomment-1304613482

pmtk:
/lgtm
:>

PR #27406, opened 2022-09-07
Automatically skip tests with apigroups which are not served

ingvagabund:
Excluding tests with [apigroup:GROUP] labels for apigroups which are not served by a cluster. E.g. MicroShift is not serving most of the openshift.io apigroups. Other installations might be serving only a subset of the api groups.

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -156,6 +156,25 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
 		return err
 	}
 
+	discoveryClient, err := getDiscoveryClient()
+	if err != nil {
+		if opt.DryRun {
+			fmt.Fprintf(opt.ErrOut, "Unable to get discovery client, skipping apigroup check in the dry-run mode: %v\n", err)
+		} else {
+			return err
+		}
+	} else {
+		apiGroupFilter, err := newApiGroupFilter(discoveryClient)
+		if err != nil {
+			return fmt.Errorf("unable to build api group filter: %v", err)
+		}
+
+		// Skip tests with [apigroup:GROUP] labels for apigroups which are not
+		// served by a cluster. E.g. MicroShift is not serving most of the openshift.io
+		// apigroups. Other installations might be serving only a subset of the api groups.
+		apiGroupFilter.markSkippedWhenAPIGroupNotServed(tests)
+	}
+
 	// This ensures that tests in the identified paths do not run in parallel, because
 	// the test suite reuses shared resources without considering whether another test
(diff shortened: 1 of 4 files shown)


pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -156,6 +156,21 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
 		return err
 	}
 
+	discoveryClient, err := getDiscoveryClient()
```
deads2k, line 159: need to find a way for this to not fail when running a --dry-run without a kubeconfig.  I suggest allowing all in such a case.

pkg/test/ginkgo/group_filter.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func getDiscoveryClient() (*discovery.DiscoveryClient, error) {
+	cfg := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{})
```
deads2k, line 12: huh, this is how we get our kubeconfig?  Seems weird.

pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -156,6 +156,21 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
...
+
+	apiGroupFilter, err := newApiGroupFilter(discoveryClient)
+	if err != nil {
+		return fmt.Errorf("unable to build api group filter: %v", err)
+	}
+
+	// Excluding tests with [apigroup:GROUP] labels for apigroups which are not
+	// served by a cluster. E.g. MicroShift is not serving most of the openshift.io
+	// apigroups. Other installations might be serving only a subset of the api groups.
+	tests = apiGroupFilter.filter(tests)
```
deads2k, line 172: In my ideal world, I'd get to see the test listed as skipped.  I think this is where that list gets produced https://github.com/openshift/origin/blob/master/pkg/test/ginkgo/junit.go#L30-L40 . Building on this, how much harder is it to produce that?

deads2k:
Note to self, before merge we must see the number of tests being run the same as before this PR.

ingvagabund:
> I think you need something like https://github.com/ingvagabund/origin/pull/4 to actually skip and provide details, right?

Yeap. Thanks for the pointers. I have included your commit in the PR.

deads2k:
1. gcp-csi had the correct number

deads2k:
this is looking good.

/hold cancel

PR #27411, opened 2022-09-08
WIP: Honor allowed disruption and fail instead of flaking.

dgoodwin:
[TRT-474](https://issues.redhat.com//browse/TRT-474)

Explanation forthcoming.


dgoodwin:
/close

Stopped picking up code changes from what I can see. Will open another.

PR #27415, opened 2022-09-08
Fix double disruption testing and values for most backends in intervals chart and bigquery.

dgoodwin:
[TRT-474](https://issues.redhat.com//browse/TRT-474)

We have discovered we had two parallel sets of disruption tests, both                                                             
behaving in subtly different ways, but ultimately largely duplicating                                                             
our disruption testing, and unexpectedly combining results giving us                                                              
disruption values in interval charts and likely bigquery database as                                                              
well.                                                                                                                             
                                                                                                                                  
The two testing streams:                                                                                                          
                                                                                                                                  
(1) Separate process tests, mostly of the format "remain avaialable with                                                          
new|reused connections". These tests were run in separate processes, a                                                            
couple with more complex setup/teardown procedures. During their run                                                              
they would start a monitor, and their resulting events would be written                                                           
to an AdditionalEvents json file, which we discovered was also getting                                                            
merged into the events from the main process. These tests used bigquery                                                           
data for determining what levels of disruption were acceptable, and fail                                                          
if above that. It is likely the allowed levels were double what they                                                              
should be, while the compared values were correct.                                                                                
                                                                                                                                  
(2) Disruption event invariant/synthetic tests. The main openshift-tests                                                          
binary also started virtually all monitors the above tests did, and some                                                          
more, and the results were checked after the test run via synthetic                                                               
tests. These tests never fail only flake, and did not consider allowed                                                            
disruption, any disruption was considered a flake. They were also not                                                             
wired into upgrade suites, only stable system.                                                                                    
                                                                                                                                  
This PR combines the two sets of functionality into (2), with the                                                                 
exception of a couple complex setup/teardown tests that remain in the                                                             
style of (1) and were never duplicated in (2). (image registry, service                                                           
for PDB pod). Synthetic disruption tests now use allowed disruption                                                               
data, and are wired into the upgrade suite.                                                                                       
                                                                                                                                  
The code for (1) tests remains until we can remove it from some                                                                   
presubmits where it's presently disabled, but looks like it shouldn't                                                             
be.                                                                                                                               
                                                                                                                                  
Allowed disruption data will need to be updated after this change has                                                             
merged and allowed to run for a few weeks.


pkg/synthetictests/apiserver.go
```diff
@@ -3,76 +3,11 @@ package synthetictests
...
-	"time"
-
-	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
 
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
-	"k8s.io/apimachinery/pkg/util/sets"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
 )
 
-func testServerAvailability(owner, locator string, events monitorapi.Intervals, jobRunDuration time.Duration) []*junitapi.JUnitTestCase {
```
dgoodwin, line 14: All moved to a more clear disruption.go in this module.

test/e2e/upgrade/cidisruptiontester/cidisruptiontester.go
```diff
@@ -1,90 +0,0 @@
...
-	// effectively never fail, but will flake if we experience ANY disruption. We can use this
-	// to gather data, and correlate with real disruption in graphs.
-	allowedExternalDisruption = 600 * time.Second
-
-	externalDisruptionDescription = `CI cluster where tests are running may have network issues with new ` +
-		`connections (not the cluster under test). Use intervals charts to correlate this disruption ` +
-		`against observed cluster disruption and determine if we're seeing real disruption or not.`
-)
-
-func NewCIDisruptionWithNewConnectionsTest() upgrades.Test {
```
dgoodwin, line 35: My tester for external service was only implemented in one stream of disruption testing. It is removed here and moved to a synthetic with the others.

pkg/synthetictests/disruption.go
```diff
@@ -0,0 +1,249 @@
...
+
+	ret := []*junitapi.JUnitTestCase{}
+	for _, locator := range disruptLocators.List() {
+		ret = append(ret, testServerAvailability("sig-trt", locator, allDisruptionEventsIntervals, jobRunDuration, restConfig)...)
+	}
+
+	return ret
+}
+
+func testMultipleSingleSecondDisruptions(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
```
deads2k, line 147: how much is straight move versus changes?

dgoodwin:
Analyzed the test failures, most are unrelated but one did pop up exposing that the new cache disruption tests have no data for single node, so these all fail now that they're no longer set to flake. I am going to figure out how to skip.

/hold

dgoodwin:
Egli updated the test data and we should now have entries for cache backends, this should allow the single node jobs to pass disruption tests. Ready for review.

test/e2e/upgrade/upgrade.go
```diff
@@ -54,31 +51,24 @@ func NoTests() []upgrades.Test {
 func AllTests() []upgrades.Test {
 	return []upgrades.Test{
 		&adminack.UpgradeTest{},
-		controlplane.NewKubeAvailableWithNewConnectionsTest(),
```
xueqzhan, line 57: Any impact on removing those tests? I guess just rebuilding history for them?

xueqzhan (commented):
Looking good other than a small note.

test/e2e/upgrade/upgrade.go
dgoodwin: I don't think so, it's effectively a test rename which is something that happens fairly often. Should be ok.

pkg/synthetictests/disruption.go
```diff
@@ -0,0 +1,249 @@
...
+				Name:     testName,
+				Duration: jobRunDuration.Seconds(),
+				FailureOutput: &junitapi.FailureOutput{
+					Output: fmt.Sprintf("error in getting allowed disruption: %s", err),
+				},
+			},
+		}
+	}
+	roundedAllowedDisruption := allowedDisruption.Round(time.Second)
+	if allowedDisruption.Milliseconds() == 2718 {
```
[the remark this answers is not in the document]
dgoodwin: Added a constant, thanks for looking over this ken. Latest is pushed.

deads2k:
diff of diff looks good

/approve

PR #27432, opened 2022-09-21
Check version when discovery client

ingvagabund:
`getDiscoveryClient()` does not checks whether a cluster is reachable. One needs to e.g. check cluster version to verify the connection with the cluster. A.k.a allow dry-run to skip api group filter builder when a cluster is not reachable.


deads2k:
nesting is ugly, but I'll take it

/lgtm

PR #27434, opened 2022-09-21
OCPBUGS-1052: allow frontends to tolerate 2.5% disruption during upgrades

joelanford:
Recent 4.8 to 4.9 upgrade tests fail fairly consistently due to cluster frontend ingress being unavailable. Currently the upgrade test tolerates no disruption whatsoever. This PR makes the test more lenient, allowing 2.5% disruption, which is the minimum that would have resulted in passes for the string of recent failures.

Note for reviewer: this is outside my typical domain, so if there's a more targeted fix we could make (e.g. do this only on AWS upgrades), or we think its worth root-causing and/or adding a link in the test output to help our future selves understand this decision, let me know what's appropriate.

Signed-off-by: Joe Lanford <joe.lanford@gmail.com>


wking (approved):
/lgtm

sdodson:
Accepting this test threshold change directly into 4.9 as these tests have been reimplemented in later releases.

sdodson:
This test does not seem likely to be affected by this change, we will investigate to make sure that there's a bug filed to track looking into it.

PR #27435, opened 2022-09-22
Add Kuryr exception to "pods should successfully create sandboxes" test

dulek:
Kuryr can take long to create Pods as it needs to create ports for these pods in Neutron. If a pod gets deleted while this is being processed, a warning Event will be generated. This causes the `[sig-network] pods should successfully create sandboxes by adding pod to network` to fail.

This commit adds this case as a flake.

### DIFF
--- a/pkg/synthetictests/networking.go
+++ b/pkg/synthetictests/networking.go
@@ -109,6 +109,12 @@ func testPodSandboxCreation(events monitorapi.Intervals, clientConfig *rest.Conf
 				continue
 			}
 		}
+		if strings.Contains(event.Message, "kuryr") && strings.Contains(event.Message, "deleted while processing the CNI ADD request") {
+			// This happens from time to time with Kuryr. Creating ports in Neutron for pod can take long and a controller might delete the pod before,
+			// effectively cancelling Kuryr CNI ADD request.
+			flakes = append(flakes, fmt.Sprintf("%v - pod got deleted while kuryr was still processing its creation - %v", event.Locator, event.Message))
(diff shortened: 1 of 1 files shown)


dulek:
See [1] for an example failure.

[1] https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_kuryr-kubernetes/691/pull-ci-openshift-kuryr-kubernetes-master-e2e-openstack-kuryr/1572902208108761088

MaysaMacedo:
/retest
/lgtm
It would be nice to test it somehow

dulek:
Test seems to be passing as a flake, which is what we wanted. I'm not rechecking, a different problem appeared, but this PR seems to fix the problem it intended to.

dulek:
/assign @spadgett

Please take a look, this unblocks e2e-openstack-kuryr CIs.

MaysaMacedo:
The new issue in the Kuryr CI might be related to https://issues.redhat.com/browse/OCPBUGS-2061, which is not an issue with Kuryr itself

dulek:
/test e2e-openstack-kuryr

Just out of curiosity if the problem Maysa mentioned above is fixed.

dulek:
@bparees, @spadgett, could you take a look here? It's to unblock a CI that's failing quite a lot on 4.12.

dulek:
/retest-required

Weirdly it does seem to fail on a similar set of tests. Yet I strongly doubt this patch can affect it in a negative way.

PR #27438, opened 2022-09-23
USHIFT-358: MicroShift: skip sig-storage tests

pacevedom:


jan--f:
@pacevedom Can you please add some context here? Iiuc this impacts all OpenShift runs that include this test, not just microshift.

test/extended/prometheus/storage.go
```diff
@@ -30,7 +30,7 @@ var _ = g.Describe("[sig-storage][Late] Metrics", func() {
 		var ok bool
 		url, _, bearerToken, ok = helper.LocatePrometheus(oc)
 		if !ok {
-			e2e.Failf("Prometheus could not be located on this cluster, failing prometheus test")
+			g.Skip("Prometheus could not be located on this cluster, skipping prometheus test")
```
ingvagabund, line 33: @jan--f if this check is expected to be gating there should probably be another test checking Prometheus is running for non-microshift installation. @pacevedom can you please sync with the @openshift/storage and discuss alterantives?

jsafrane: This change does not skip all storage tests, it skips prometheus tests that require storage. Is that the intention?

pacevedom: Yes, the intention is to skip anything related to Prometheus, as in MicroShift there is nothing to store/check metrics.
LocatePrometheus returns false only if there is [no service](https://github.com/openshift/origin/blob/master/test/extended/util/prometheus/helpers.go#L68) for it. Does this happen in any situation in regular OCP/HyperShift?

jsafrane: The PR title is then very confusing.

What about the other Prometheus tests in [prometheus.go](https://github.com/openshift/origin/blob/ef6536ba596817bd75acd90b56ebbd9e5d8772d3/test/extended/prometheus/prometheus.go#L92)?

jsafrane: There are more `Describe` blocks in the file... Most of the tests are behind `apigroup:config.openshift.io`, but some are always running ([`should report telemetry [Late]`](https://github.com/openshift/origin/blob/ef6536ba596817bd75acd90b56ebbd9e5d8772d3/test/extended/prometheus/prometheus.go#L449))

Anyway, from storage POV, it's OK to skip `[sig-storage][Late] Metrics` tests when running without any cloud provider. It's mostly for Azure.

/lgtm

soltysh (approved):
/approve

PR #27439, opened 2022-09-23
Check for console capability before launching it's disruption monitoring.

dgoodwin:
Adds new utilities to check for enabled and known capabilities. Check
also assumes that if the cluster doesn't know about the console
capability, it predates 4.12 and thus we should assume to do monitoring
for console as it was default then.

The test should then see no ingress- backends for console, and not fail.

[TRT-573](https://issues.redhat.com//browse/TRT-573)

### DIFF
--- a/test/extended/util/cluster/cluster.go
+++ b/test/extended/util/cluster/cluster.go
@@ -290,3 +290,20 @@ func (c *ClusterConfiguration) MatchFn() func(string) bool {
 	}
 	return matchFn
 }
+
+func HasCapability(clusterVersion *configv1.ClusterVersion, desiredCapability string) bool {
+	for _, ec := range clusterVersion.Status.Capabilities.EnabledCapabilities {
+		if string(ec) == desiredCapability {
+			return true
+		}
+	}
+	return false
+}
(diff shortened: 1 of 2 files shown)


dgoodwin:
Testing once with force disabled to ensure we don't get test failures, or disruption data, when the backend doesn't run at all.

dgoodwin:
Test confirmed that removing the backend from running entirely does not cause any tests to fail, they operate dynamically on the disruption backends seen in the monitor events, so when one disappears the test continue to work, only looking at those starting with ingress-.

PR is now updated with the full fix to skip the backend if we do not know what the Console capability is (4.11), or it is not enabled.

DennisPeriquet:
this LGTM but I have no /lgtm power

dgoodwin:
/hold cancel

aws-ovn-serial build failed to download an rpm.

dgoodwin:
Mass failures in loki pods, looks like backoff pulling: "message": "Back-off pulling image \"registry.redhat.io/openshift4/ose-oauth-proxy:latest\""

dgoodwin:
Registry issues may have been cleared.

PR #27445, opened 2022-09-28
Hypershift: skip monitoring of oauth ingress route

csrwng:
In external control plane topology clusters, the oauth server route does not exist inside the cluster under test. The endpoint is created on the control plane side.

This change skips monitoring of the oauth ingress when running on an external control plane cluster.


bparees:
> In external control plane topology clusters, the oauth server route does not exist inside the cluster under test. The endpoint is created on the control plane side.

The test code runs from outside the cluster under test, so shouldn't this still work for a hypershift guest cluster, since the route should still exist and be externally reachable?

csrwng:
> The test code runs from outside the cluster under test, so shouldn't this still work for a hypershift guest cluster, since the route should still exist and be externally reachable?

The problem is that the test code expects to find a route inside the cluster to connect to. In HyperShift clusters there is no oauth route. The oauth endpoint is served up externally from the management cluster, where it can be a route of the mgmt cluster, nodeport, or an external load balancer.

bparees:
> In HyperShift clusters there is no oauth route.

ok. "there is no route(api object)" is a different statement from "there's no route(api object) on the guest cluster".

(I assumed, much like "there's no etcd on the guest cluster"(because it's in the control plane), there was still always a route defined on the management cluster, that got you to this guest cluster's oauth endpoint).

Since you say there is at least sometimes a route defined on the mgmt cluster, is that not the case for how we run these hypershift guests in CI/testing?

csrwng:
@bparees updated to lookup the OAuth endpoint from the Kube APIServer for the external control plane topology. This doesn't require looking up anything on the management cluster side. PTAL

PR #27446, opened 2022-09-28
Bug 2033167: Remove deprecated oc extract missing dir test

ardaguclu:
Currently `oc extract` command fails when target directory is absent. However, with this PR https://github.com/openshift/oc/pull/1248, we are changing functionality and we are creating nested directories
if they are absent. 

Thus, this test becomes obsolete and this PR removes it.


ardaguclu:
It would be better to change the test(to check directory is created or not) instead removing it. But that will create circular dependency between the oc PR.

ardaguclu:
I'm closing this pr, because we are still going to use the same functionality in terms of this test.

PR #27450, opened 2022-09-28
Skip resource growth checks in upgrade-rollback jobs

jluhrsen:
The new checks don't work with upgrade rollback because there is no completion time for the upgrade. It's aborted at 99% and the clusterversion has the completion time as null.

This will still return a junit, but with [Skipped] for those cases when the completion time is null.

https://issues.redhat.com/browse/SDN-3505

### DIFF
--- a/pkg/synthetictests/resource_growth.go
+++ b/pkg/synthetictests/resource_growth.go
@@ -3,6 +3,7 @@ package synthetictests
 import (
 	"context"
 	"fmt"
+	v1 "github.com/openshift/api/config/v1"
 	"time"
 
 	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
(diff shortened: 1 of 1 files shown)


jluhrsen:
https://issues.redhat.com/browse/SDN-3505

jluhrsen:
/assign @dgoodwin 

@dgoodwin , can you review this please? a little more info in the [JIRA](https://issues.redhat.com/browse/SDN-3505) if you like.

pkg/synthetictests/resource_growth.go
```diff
@@ -123,3 +109,42 @@ func comparePostUpgradeResourceCountFromMetrics(testName, resource string) []*ju
 	}
 
 }
+
+func getAndCheckClusterVersion(testName string) (*v1.ClusterVersion, []*junitapi.JUnitTestCase) {
```
dgoodwin, line 113: Could we pass in the oc client rather than creating another one.

pkg/synthetictests/resource_growth.go
```diff
@@ -123,3 +109,42 @@ func comparePostUpgradeResourceCountFromMetrics(testName, resource string) []*ju
...
+			},
+		}
+	}
+
+	// In the case that cv completiontime is nil meaning the version change (upgrade or
+	// rollback is still in progress), skip this case. The problem is likely a bigger
+	// problem and the job will fail for that.
+	if cv.Status.History[0].CompletionTime == nil {
+		return cv, []*junitapi.JUnitTestCase{
+			{Name: "[Skipped] " + testName},
```
dgoodwin, line 145: I believe that modifying the test name in specific cases like this doesn't actually skip anything, and now has created a new test that only runs on rollback jobs, and the other original test only runs on the others. Pass rate tracking is now disjoint between the two in all our tooling. I've seen the [Skipped] syntax and I'm not super familiar with it, but from what I understand it's constant in the test name, and always skipped, or includes a condition that something internal determines is true or not, which I'm guessing we don't want to plumb through for rollback jobs.

I think the more correct thing to do in a synthetic test would be to flake it, keep the original test name, in the slice of junit test results return two entries, fail the first and say why, and pass the second. The test will now appear as a flake in the spyglass UI, sippy, and bigquery.

pkg/synthetictests/resource_growth.go
jluhrsen: yes

pkg/synthetictests/resource_growth.go
jluhrsen: @dgoodwin , please take another look. I've updated it to make it flake now.

pkg/synthetictests/resource_growth.go
dgoodwin: This was missed, I don't fully know how heavy weight the client create is, probably not much but just to be safe it would be good to pass it through. Too often there's some obscure mechanism out there, maybe assuming we use the same client throughout a test or something.

jluhrsen: dang it. my bad. just updated and tested again with this change. thanks for catching it the 2nd time.

dgoodwin:
/lgtm
/approve

No worries and thanks!

PR #27452, opened 2022-09-30
Fix missing reporting of 0s disruptions.

dgoodwin:
[TRT-584](https://issues.redhat.com//browse/TRT-584)

Not including info events caused 0s disruptions to disappear in the
reported data and bigquery since Jun 15. Broken in #27250.

In original PR I needed a way to classify DNS lookup failures as something other than disruption of the ephemeral cluster under test. We still wanted to track these failures, so I used the level of the interval event to do so moving these to a Warning, and leaving what we think is real disruption as Error. I updated the code to filter out the Warnings in appropriate places by just picking up Errors. I was unaware there were actually no disruption intervals at level Info, these then began being omitted in the backend-disruption json files and thus never making it into bigquery.

### DIFF
--- a/pkg/monitor/write_job_run_data.go
+++ b/pkg/monitor/write_job_run_data.go
@@ -70,7 +70,10 @@ func computeDisruptionData(eventIntervals monitorapi.Intervals) *BackendDisrupti
 	allDisruptionEventsIntervals := eventIntervals.Filter(
 		monitorapi.And(
 			monitorapi.IsDisruptionEvent,
-			monitorapi.IsErrorEvent, // ignore Warning events, we use these for disruption we don't actually think was from the cluster under test (i.e. DNS)
+			monitorapi.Or(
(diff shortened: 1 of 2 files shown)


dgoodwin:
WIP until I can see 0's in the backend-disruption.json files of some presubmits.

dgoodwin:
Confirmed this gets us back to 0's: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27452/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1575917476561031168/artifacts/e2e-gcp-ovn-upgrade/openshift-e2e-test/artifacts/junit/backend-disruption_20220930-192924.json

dgoodwin:
I've gone through the failures and successes, I can't see any issues related to this change and it seems to have the intended effect. Removing WIP.

dgoodwin:
Added unit tests so we don't do this again.

Then discovered we don't actually run the origin unit tests. Filed a card for this later.

PR #27454, opened 2022-10-02
Trt 579 http2 connection lost

neisw:
Adds support for extracting `http2: client connection lost` errors from kubelet logs

### DIFF
--- a/pkg/monitor/intervalcreation/node.go
+++ b/pkg/monitor/intervalcreation/node.go
@@ -165,6 +165,10 @@ func eventsFromKubeletLogs(nodeName string, kubeletLog []byte) monitorapi.Interv
 		currLine := scanner.Text()
 		ret = append(ret, readinessFailure(currLine)...)
 		ret = append(ret, readinessError(currLine)...)
+		ret = append(ret, statusHttpClientConnectionLostError(currLine)...)
+		ret = append(ret, reflectorHttpClientConnectionLostError(currLine)...)
+		ret = append(ret, kubeletNodeHttpClientConnectionLostError(currLine)...)
+
 	}
 
 	return ret
@@ -245,14 +249,18 @@ var failureOutputRegex = regexp.MustCompile(`"Probe failed" probeType="Readiness
 var errorOutputRegex = regexp.MustCompile(`"Probe errored" err="(?P<OUTPUT>.+)" probeType="Readiness"`)
 
 func probeProblemToContainerReference(logLine string) monitorapi.ContainerReference {
+	return regexToContainerReference(logLine, containerRefRegex)
+}
+
+func regexToContainerReference(logLine string, containerReferenceMatch *regexp.Regexp) monitorapi.ContainerReference {
 	ret := monitorapi.ContainerReference{}
-	containerRefRegex.MatchString(logLine)
-	if !containerRefRegex.MatchString(logLine) {
+	containerReferenceMatch.MatchString(logLine)
+	if !containerReferenceMatch.MatchString(logLine) {
 		return ret
 	}
 
-	subMatches := containerRefRegex.FindStringSubmatch(logLine)
-	subNames := containerRefRegex.SubexpNames()
+	subMatches := containerReferenceMatch.FindStringSubmatch(logLine)
+	subNames := containerReferenceMatch.SubexpNames()
 	for i, name := range subNames {
 		switch name {
 		case "NS":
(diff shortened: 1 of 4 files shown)


pkg/monitor/intervalcreation/node.go
```diff
@@ -269,6 +277,90 @@ func probeProblemToContainerReference(logLine string) monitorapi.ContainerRefere
...
+			return ""
+		}
+		return nodeRefRegex.FindStringSubmatch(logLine)[1]
+	})
+
+}
+
+func commonErrorInterval(logLine string, messageExp *regexp.Regexp, reason string, locator func() string) monitorapi.Intervals {
+
+	messageExp.MatchString(logLine)
```
dgoodwin, line 338: Extra line?

pkg/monitor/intervalcreation/node_test.go
```diff
@@ -10,6 +11,191 @@ import (
...
+				To:   time.UnixMicro(1664269199853216).In(time.UTC),
+			},
+		},
+		{
+			name:    "kubelet",
+			logLine: `Sep 27 08:59:59.853216 ci-op-747jjqn3-b3af3-f45pk-worker-centralus2-bdp5s kubenswrapper[2397]: E0927 08:59:59.849143    2397 kubelet_node_status.go:487] "Error updating node status, will retry" err="error getting node \"ci-op-747jjqn3-b3af3-f45pk-worker-centralus2-bdp5s\": Get \"https://api-int.ci-op-747jjqn3-b3af3.ci2.azure.devcluster.openshift.com:6443/api/v1/nodes/ci-op-747jjqn3-b3af3-f45pk-worker-centralus2-bdp5s?timeout=10s\": http2: client connection lost"`,
+			want: monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Level:   0,
+					Locator: "ci-op-747jjqn3-b3af3-f45pk-worker-centralus2-bdp5s",
```
dgoodwin, line 53: Should this have more locatorish syntax? Is that a node name? maybe node/ci-op...

pkg/monitor/intervalcreation/node_test.go
```diff
@@ -10,6 +11,191 @@ import (
...
+				To:   time.UnixMicro(1664269199853216).In(time.UTC),
+			},
+		},
+		{
+			name:    "simple failure",
+			logLine: `Jul 05 17:47:52.807876 ci-op-lxqqvl5x-d3bee-gl4hp-master-0 hyperkube[1495]: I0606 17:47:52.807876    1599 prober.go:121] "Probe failed" probeType="Readiness" pod="openshift-authentication/oauth-openshift-77f7b95df5-r4xf7" podUID=1af660b3-ac3a-4182-86eb-2f74725d8415 containerName="oauth-openshift" probeResult=failure output="Get \"https://10.129.0.12:6443/healthz\": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"`,
+			want: monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Level:   0,
+					Locator: "ns/openshift-authentication pod/oauth-openshift-77f7b95df5-r4xf7 uid/1af660b3-ac3a-4182-86eb-2f74725d8415 container/oauth-openshift",
```
dgoodwin, line 66: Could we slip node/ into all of these locators?

pkg/monitor/intervalcreation/node_test.go
```diff
@@ -10,6 +11,191 @@ import (
...
+	if intervals == nil {
+		t.Fatal("Invalid intervals")
+	}
+
+	if intervals.Len() != len(testcase) {
+		t.Fatal("Mismatched interval count")
+	}
+
+	for i := range intervals {
+		if !reflect.DeepEqual(intervals[i], testcase[i].want) {
```
dgoodwin, line 105: This is not consistently used here and the OGs of this repo probably think it's sacrilige, but I love the testity assert/require modules: https://github.com/stretchr/testify

They can make assertions more brief and give more consistent output, easier to view diffs, etc.

Not at all required for this PR, but can be very useful for testing in future.

pkg/synthetictests/kubelet.go
```diff
@@ -48,6 +48,35 @@ func testKubeletToAPIServerGracefulTermination(events monitorapi.Intervals) []*j
...
+	var failures []string
+	for _, event := range events {
+		if strings.Contains(event.Message, "reason/HttpClientConnectionLost") {
+			failures = append(failures, fmt.Sprintf("%v - %v", event.Locator, event.Message))
+		}
+	}
+
+	// failures during a run always fail the test suite
+	var tests []*junitapi.JUnitTestCase
+	if len(failures) > 0 {
```
dgoodwin, line 68: This will seem like a nit, but it's actually for what I think is a good reason and something we've botched in the past.

Suggest changing this to if len(failures) == 0, and add the success JunitTestCase and return immediately.

Then move the failure add out one level (no else required), and keep the success add to trigger the flake.

This is to prevent a very likely scenario where one of us goes in here to make this no longer flake, kill off line 75, and then it looks to all our tooling like we're no longer running the test unless it fails. This breaks some things and we've done if a few times before.

pkg/monitor/intervalcreation/node_test.go
neisw: I can if it is the right thing to do but for the Probe error / failures, that code existed already.  The 'status' and 'reflector' cases above are reusing that code.  For the 'kubelet' case I can definitely prefix the locator with 'node/'

pkg/monitor/intervalcreation/node_test.go
```diff
@@ -10,6 +11,191 @@ import (
...
+			name:    "status",
+			logLine: `Sep 27 08:59:59.857303 ci-op-747jjqn3-b3af3-f45pk-worker-centralus2-bdp5s kubenswrapper[2397]: I0927 08:59:59.850662    2397 status_manager.go:667] "Failed to get status for pod" podUID=a1947638-25c2-4fd8-b3c8-4dbaa666bc61 pod="openshift-monitoring/prometheus-k8s-0" err="Get \"https://api-int.ci-op-747jjqn3-b3af3.ci2.azure.devcluster.openshift.com:6443/api/v1/namespaces/openshift-monitoring/pods/prometheus-k8s-0\": http2: client connection lost"`,
+			want: monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Level:   0,
+					Locator: "ns/openshift-monitoring pod/prometheus-k8s-0 uid/a1947638-25c2-4fd8-b3c8-4dbaa666bc61 container/",
+					Message: "reason/HttpClientConnectionLost Get \"https://api-int.ci-op-747jjqn3-b3af3.ci2.azure.devcluster.openshift.com:6443/api/v1/namespaces/openshift-monitoring/pods/prometheus-k8s-0\": http2: client connection lost",
+				},
+				From: time.UnixMicro(1664269199857303).In(time.UTC),
+				To:   time.UnixMicro(1664269199857303).In(time.UTC),
```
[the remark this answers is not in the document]
neisw: Good call.  I think we would have had issues Jan 01 2023 as it was.

pkg/monitor/intervalcreation/node_test.go
neisw: The code generating that Locator value is https://github.com/openshift/origin/blob/423301d44e82ee4a199a994f3aac4fad548be1ea/pkg/monitor/monitorapi/identification_pod.go#L63-L65

dgoodwin: Ok the bare string was definitely the one that it looked most useful for above, so that's is fine with me, thinking about it today we could easily work back from a pod name to a node. Thanks!

dgoodwin:
/approve
/override ci/prow/e2e-gcp-ovn-image-ecosystem
Failures are definitely not related from what I can see.

PR #27457, opened 2022-10-03
Drastically reduce the jobs we enforce disruption on.

dgoodwin:
Our operating assumption was that we were only enforcing disruption
using P99 values, thus would expect to only fail 1% of jobs, and if you
violate that test then you likely have regressed.

We have discovered that the vast majority of job platform combinations
do not have the 100 minimum runs required to get a P99, and thus we're
actually failing jobs on as low as P50's, which causes very erratic
failures and retests.

We are now limiting the data going into the historical
query_results.json in openshift/ci-tools repo. Here we now assume that
if we do not have disruption data for the platform under test, we flake
and report why. Previously this was a failure. The flake will allow us
to find instances where this is occurring if anyone needs to.

OCPBUGS-1459
[TRT-583](https://issues.redhat.com//browse/TRT-583)

### DIFF
--- a/pkg/synthetictests/disruption.go
+++ b/pkg/synthetictests/disruption.go
@@ -57,6 +57,26 @@ func testServerAvailability(
 			},
 		}
 	}
+
+	// Check if we got an empty result, which signals we did not have historical data for this NURP and thus
+	// do not want to run the test.
+	if allowedDisruption == nil {
+		// An empty StatisticalDuration implies we did not find any data and thus do not want to run the disruption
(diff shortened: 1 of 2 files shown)


dgoodwin:
I have not yet merged the limited data going into query_results.json, as while working on this I discovered weird results that led to #27452. That needs to merge and then we should wait 3 weeks before the data will be correct.

However this change can go in, this changes such that if we don't have an exact or fuzzy match, we stop failing the test.

dgoodwin:
Back to WIP, I should add a unit test to make sure this never happens again.

PR #27458, opened 2022-10-04
Add an upgrade test to verify the migration to AllowedSourceRanges

suleymanakbas91:
This PR adds an upgrade test to verify the migration from the `service.beta.kubernetes.io/load-balancer-source-ranges` annotation usage to the new `AllowedSourceRanges` API.

Epic: https://issues.redhat.com/browse/NE-555
EP: https://github.com/openshift/enhancements/pull/1177
API: https://github.com/openshift/api/pull/1222

- `test/e2e/upgrade/ingress/ingress.go`: Implements 
[Test](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/upgrades/upgrade.go#L46) interface that includes the following steps:

1. In `Setup`, create an IngressController that requests an LB, and set the `service.beta.kubernetes.io/load-balancer-source-ranges` annotation on the resulting LoadBalancer-type service.
2. In `Test`, periodically check the service to verify that nothing updates the annotation or the `Spec.LoadBalancerSourceRanges` field on the service during the upgrade.
3. In `Test`, verify again when the upgrade is done that nothing has updated the annotation or `Spec.LoadBalancerSourceRanges`, and verify that the operator has set `Progressing=True` and `EvaluationConditionsDetected=True`.
4. Then set `AllowedSourceRanges`, verify if LoadBalancerSourceRanges is set accordingly, and the annotation is removed. Finally, verify that the operator has set `Progressing=False` and `EvaluationConditionsDetected=False`.

- `test/e2e/upgrade/upgrade.go`: Adds the new upgrade test to the test suite.


suleymanakbas91:
We don't need this anymore.

PR #27459, opened 2022-10-06
IR-258: add dockerImageLayers to all Image objects

flavianmissi:
openshift/openshift-apiserver#290 added validation for the presence of dockerImageLayers field.


flavianmissi:
after discussions with @dmage we have decided to remove the layers validation instead, since docker images are perfectly valid without any layers in them.

PR #27416, opened 2022-09-09
k8s 1.25.0

soltysh:
Replaces https://github.com/openshift/origin/pull/27376

PR #27462, opened 2022-10-10
microshift: avoid oauth and console disruption monitoring

pacevedom:
When https://github.com/openshift/origin/pull/27460 was introduced the tests in MicroShift stopped working because of checking `clusterversion.config.openshift.io`. Since this API group/resource is not available in MicroShift, we need to skip it.

/cc @dgoodwin

### DIFF
--- a/test/extended/util/disruption/frontends/known_backends.go
+++ b/test/extended/util/disruption/frontends/known_backends.go
@@ -4,46 +4,83 @@ import (
 	"context"
 
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
 	"github.com/openshift/origin/pkg/monitor"
 	"github.com/openshift/origin/pkg/monitor/backenddisruption"
+	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/cluster"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/rest"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 )
 
-func StartAllIngressMonitoring(ctx context.Context, m monitor.Recorder, clusterConfig *rest.Config) error {
-	if err := createOAuthRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
-		return err
-	}
-	if err := createOAuthRouteAvailableWithConnectionReuse().StartEndpointMonitoring(ctx, m, nil); err != nil {
-		return err
-	}
+const (
+	oauthRouteNamespace = "openshift-authentication"
+	oauthRouteName      = "oauth-openshift"
+)
 
-	// Some jobs explicitly disable the console and other features. Check if it's disabled and if so,
-	// do not run a disruption monitoring backend for it.
-	configClient, err := configclient.NewForConfig(clusterConfig)
+func StartAllIngressMonitoring(ctx context.Context, m monitor.Recorder, clusterConfig *rest.Config) error {
+	// Ingress monitoring checks for oauth and console routes to monitor healthz endpoints. Check availability
+	// before setting up any monitors.
+	routeAvailable, err := isRouteAvailable(ctx, clusterConfig, oauthRouteNamespace, oauthRouteName)
 	if err != nil {
 		return err
 	}
-	clusterVersion, err := configClient.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
-	if err != nil {
-		e2e.Failf("Failed to get cluster version: %v", err)
-	}
-	// If the cluster does not know about the Console capability, it likely predates 4.12 and we can assume
-	// it has it by default. This is to catch possible future scenarios where we upgrade 4.11 no cap to 4.12 no cap.
-	if !cluster.KnowsCapability(clusterVersion, "Console") ||
-		cluster.HasCapability(clusterVersion, "Console") {
-		if err := createConsoleRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
+	if routeAvailable {
+		if err := createOAuthRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
 			return err
 		}
-		if err := createConsoleRouteAvailableWithConnectionReuse().StartEndpointMonitoring(ctx, m, nil); err != nil {
+		if err := createOAuthRouteAvailableWithConnectionReuse().StartEndpointMonitoring(ctx, m, nil); err != nil {
 			return err
+		}
+	}
+
+	configAvailable, err := exutil.DoesApiResourceExist(clusterConfig, "clusterversions")
+	if err != nil {
+		return err
+	}
+	if configAvailable {
+		// Some jobs explicitly disable the console and other features. Check if it's disabled and if so,
+		// do not run a disruption monitoring backend for it.
+		configClient, err := configclient.NewForConfig(clusterConfig)
+		if err != nil {
+			return err
+		}
+		clusterVersion, err := configClient.ConfigV1().ClusterVersions().Get(context.TODO(), "version", metav1.GetOptions{})
+		if err != nil {
+			e2e.Failf("Failed to get cluster version: %v", err)
+		}
+		// If the cluster does not know about the Console capability, it likely predates 4.12 and we can assume
+		// it has it by default. This is to catch possible future scenarios where we upgrade 4.11 no cap to 4.12 no cap.
+		if !cluster.KnowsCapability(clusterVersion, "Console") ||
+			cluster.HasCapability(clusterVersion, "Console") {
+			if err := createConsoleRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
+				return err
+			}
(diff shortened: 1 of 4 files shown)


test/extended/util/disruption/frontends/known_backends.go
```diff
@@ -4,55 +4,109 @@ import (
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func isConfigAPIGroupAvailable(config *rest.Config) (bool, error) {
```
ingvagabund, line 86: Worth making the function more generic. E.g. `isAPIGroupAvailable(config *rest.Config, apigroup string)`. Or, more preferable to reuse `exutil.DoesApiResourceExist`.

test/extended/util/disruption/frontends/known_backends.go
```diff
@@ -4,55 +4,109 @@ import (
...
+const (
+	oauthRouteNamespace = "openshift-authentication"
+	oauthRouteName      = "oauth-openshift"
+)
+
 func StartAllIngressMonitoring(ctx context.Context, m monitor.Recorder, clusterConfig *rest.Config) error {
-	if err := createOAuthRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
+	// Ingress monitoring checks for oauth and console routes to monitor healthz endpoints. Check availability
+	// before setting up any monitors.
+	available, err := isRouteAvailable(ctx, clusterConfig, oauthRouteNamespace, oauthRouteName)
```
ingvagabund, line 27: available -> routeAvailable
to align with configAvailable

test/extended/util/disruption/frontends/known_backends.go
```diff
@@ -4,55 +4,109 @@ import (
...
-	}
-	// If the cluster does not know about the Console capability, it likely predates 4.12 and we can assume
-	// it has it by default. This is to catch possible future scenarios where we upgrade 4.11 no cap to 4.12 no cap.
-	if !cluster.KnowsCapability(clusterVersion, "Console") ||
-		cluster.HasCapability(clusterVersion, "Console") {
-		if err := createConsoleRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
+	if configAvailable {
+		// Some jobs explicitly disable the console and other features. Check if it's disabled and if so,
+		// do not run a disruption monitoring backend for it.
+		configClient, err := configclient.NewForConfig(clusterConfig)
```
ingvagabund, line 47: If `exutil.DoesApiResourceExist` gets used, the oc's config client could be used here instead of creating a new one.

test/extended/util/disruption/frontends/known_backends.go
```diff
@@ -4,55 +4,109 @@ import (
...
-	if err := createOAuthRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
+	// Ingress monitoring checks for oauth and console routes to monitor healthz endpoints. Check availability
+	// before setting up any monitors.
+	available, err := isRouteAvailable(ctx, clusterConfig, oauthRouteNamespace, oauthRouteName)
+	if err != nil {
 		return err
 	}
-	if err := createOAuthRouteAvailableWithConnectionReuse().StartEndpointMonitoring(ctx, m, nil); err != nil {
-		return err
+	if available {
```
ingvagabund, line 31: This test originally does not test for `isRouteAvailable` nor for `configAvailable`. To align with the original functionality one should test for both `isRouteAvailable` and `configAvailable` and only then create all the connections and other objects. Or, return error otherwise to fail earlier and avoid initializing half of the functionality.

ingvagabund (changes requested):

dgoodwin:
Apologies for the breakage. Looks good to me will just hold for Jan's requests and then find an approver.

test/extended/util/disruption/frontends/known_backends.go
pacevedom: Was not keen on refactoring seemingly unrelated parts of the code, but this change might be small enough to do so.

test/extended/util/disruption/frontends/known_backends.go
pacevedom: Done

test/extended/util/disruption/frontends/known_backends.go
pacevedom: This would involve creating an oc object here only to use the config client, is that an overkill? oc objects do many operations besides creating clients for APIs.

test/extended/util/disruption/frontends/known_backends.go
pacevedom: The routes used for monitoring belong to the console and the oauth api server, which are part of a regular OCP cluster. Since we don't have them in MicroShift this should be good enough to determine whether disruption tests should be created.
The one over the configAvailable comes from [this PR](https://github.com/openshift/origin/pull/27460), which introduces capability checking and a fallback if capability checking fails (which means we should still create the disruption monitor). Since we don't have the `clusterversion` in MicroShift we need to skip it.

pacevedom:
> Apologies for the breakage. Looks good to me will just hold for Jan's requests and then find an approver.

No worries! All this stuff is pretty new so it was difficult to know if there is something broken :)

test/extended/util/disruption/frontends/known_backends.go
ingvagabund: If that's the case I am taking my comment back

test/extended/util/disruption/frontends/known_backends.go
ingvagabund: I meant to refactor `isConfigAPIGroupAvailable`. Or call `exutil.DoesApiResourceExist` as it is instead of `isConfigAPIGroupAvailable`. Sorry for the confusion.

test/extended/util/disruption/frontends/known_backends.go
ingvagabund: What I meant was:
```go
	// Ingress monitoring checks for oauth and console routes to monitor healthz endpoints. Check availability
	// before setting up any monitors.
	routeAvailable, err := isRouteAvailable(ctx, clusterConfig, oauthRouteNamespace, oauthRouteName)
	if err != nil {
		return err
	}
	configAvailable, err := isConfigAPIGroupAvailable(clusterConfig)
	if err != nil {
		return err
	}

	if !routeAvailable || !configAvailable {
		return fmt.Error("Route or config not available")
	}
	if routeAvailable {
		if err := createOAuthRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
			return err
		}
		if err := createOAuthRouteAvailableWithConnectionReuse().StartEndpointMonitoring(ctx, m, nil); err != nil {
			return err
		}
	}
	if configAvailable {
		// Some jobs explicitly disable the console and other features. Check if it's disabled and if so,
		// do not run a disruption monitoring backend for it.
		configClient, err := configclient.NewForConfig(clusterConfig)
...
```

test/extended/util/disruption/frontends/known_backends.go
pacevedom: Done

test/extended/util/disruption/frontends/known_backends.go
pacevedom: The problem with that check is returning the error. This would prevent tests from running in MicroShift, as it fails to prepare before launching any of them.
If there are no routes for oauth and/or console then there is no point in setting anything up, as those services might not be installed/active, and this should mean no error.
It is a monitor device watching for any possible disruptions during the whole test run. If there are no services to watch, there cant be any disruptions, therefore the prepare phase should progress and setup nothing.

ingvagabund: > If there are no routes for oauth and/or console then there is no point in setting anything up, as those services might not be installed/active, and this should mean no error.

Valid point.

> The problem with that check is returning the error. This would prevent tests from running in MicroShift, as it fails to prepare before launching any of them.

I was under assumption running `StartEndpointMonitoring` requires existence of the route. Though, checking the code it seems it does not. `disruptionSampler.produceSamples` will produce an error over non-existing route connection check. Though, the monitor will keep running and collecting samples. Seems fair to skip running `StartEndpointMonitoring` for both cases when there's no OAuth route.

test/extended/util/disruption/frontends/known_backends.go
```diff
@@ -4,55 +4,92 @@ import (
...
-	if err := createOAuthRouteAvailableWithNewConnections().StartEndpointMonitoring(ctx, m, nil); err != nil {
+	// Ingress monitoring checks for oauth and console routes to monitor healthz endpoints. Check availability
+	// before setting up any monitors.
+	routeAvailable, err := isRouteAvailable(ctx, clusterConfig, oauthRouteNamespace, oauthRouteName)
+	if err != nil {
 		return err
 	}
-	if err := createOAuthRouteAvailableWithConnectionReuse().StartEndpointMonitoring(ctx, m, nil); err != nil {
-		return err
+	if routeAvailable {
```
ingvagabund, line 31: Is there any OpenShift installation which does not deploy routes? Given routes are deployed by default and MicroShift deploys them as well, why to perform the check?

pacevedom: There are two routes in OCP, which come from the oauth api server and the console operator. The former is always installed as it is a core operator. The latter can be disabled in 4.12+.
Neither of those exist in MicroShift, so the routes wont exist unless they are introduced in the future. In an OCP deployment there should always be (at least) a route for the oauth api server.

ingvagabund:
/hold
Let's wait for https://github.com/openshift/origin/pull/27416 to merge

PR #27465, opened 2022-10-10
OCPBUGS-1967: allow 2.5% disruption for 4.9 controlplane availability test

tkashem:


tkashem:
/hold

we need to gather some insights into why these disruption test failure has increased in 4.8 to 4.9 upgrade jobs.

tkashem:
from the discussion we had
> i think it's safe to say we have very little confidence in the data and tests in 4.9. i personally would recommend disabling those tests unless you're prepared for massive investment to bring it up to speed with the current 4.12 stack, which I suspect would be a very large effort.

/hold cancel

mfojtik:
Overriding invalid-bug because this change only affects 4.9.z

soltysh (approved):
/lgtm
/approve
/label backport-risk-assessed

PR #27467, opened 2022-10-13
wip comment out fetching events from alerts

pmtk:


pkg/monitor/alerts.go
```diff
@@ -84,6 +86,12 @@ func FetchEventIntervalsForAllAlerts(ctx context.Context, restConfig *rest.Confi
 	if err != nil {
 		return nil, err
 	}
+
+	_, err = routeClient.RouteV1().Routes("openshift-monitoring").Get(ctx, "prometheus-k8s", metav1.GetOptions{})
```
stlaz, line 90: wouldn't it be easier (and more reliable) to just check for the ns?

pmtk: well, if we'll have openshift-monitoring ns, then it's pretty sure that we'll also have prometheus - changed

PR #27468, opened 2022-10-17
disruption checking is disabled for prior releases because we lack a good baseline

deads2k:
After checking, I think this is the minimal change to disable disruption on prior releases.  We can try to remove all the code, but I think this allows us to retain data collection (such as it is), while disabling test failures.

/assign @dgoodwin

### DIFF
--- a/pkg/synthetictests/allowedbackenddisruption/matches.go
+++ b/pkg/synthetictests/allowedbackenddisruption/matches.go
@@ -9,5 +9,6 @@ import (
 // GetAllowedDisruption uses the backend and information about the cluster to choose the best historical p95 to operate against.
 // We enforce "don't get worse" for disruption by watching the aggregate data in CI over many runs.
 func GetAllowedDisruption(backendName string, jobType platformidentification.JobType) (*time.Duration, string, error) {
-	return getCurrentResults().BestMatchP99(backendName, jobType)
+	t := 6 * time.Hour
+	return &t, "disruption checking is disabled for prior releases because we lack a good baseline", nil
 }


dgoodwin:
/lgtm

The parallel set of tests are all set to perma-flake IIRC, seems like this should do the trick. Thanks.

@sdodson for awareness, we're turning off the tests for monitoring this in 4.11. Data is still collected so we can at least still see trends even if we can't trust the numbers. Just looping you in given the discussion we had recently about 4.9.

To recap the problems that I believe are likely living in 4.11:
- we're treating dns lookup timeouts in the ci cluster as ephemeral cluster disruption (which it's not), and this happens all the time.
- most backends are reporting double values (not all), because we ran two sets of disruption tests and their disruption values were getting merged. this incorrect value is not used in the individual tests but it is used (IIRC) when we submit to bigquery, which influenced the values the tests are comparing against.
- we planned to enforce failing on worse than P99's, but most combinations don't have anywhere near 100 runs, in some cases we're enforcing as low as P50 and failing if over that, which will happen all the time.

sdodson:
@dgoodwin Which of your bullet points are known to be addressed in 4.12 tests? Which of them are low effort to backport? In particular if we can easily backport the test dns client fixes it seems like we could easily account for double reporting at least in a way that would ensure that we could set a high threshold and know if we start going past that.

No objection to this going in and being backported through all supported releases, just curiosity as both 4.10 and 4.11 are pretty actively maintained and it'd be great if we could ensure that we lock in some baseline if it's easy to do so.

dgoodwin:
All are fixed in 4.12, or soon to be.

The fix for (1) was #27250 and I think #27261. Medium amount of code.
The fix for (2) was #27415, it's quite large and includes a bug that would also require #27452 which is very small.
The fix for (3) would require #27457, then the change we're going to make soon in the ci-tools repo to filter out what gets included, and someone to manually keep the allowed disruption json up to date from bigquery. We've about got this automated for 4.12 but I don't know how easy it would be to backport to 4.11. 

I can't say how cleanly they would backport, but the sum would be quite a bit of effort, and a lot of monitoring to make sure it's working properly. 

I think 4.10 is out of the question as it's likely not even using this disruption stack.

sdodson:
Ok, lets move forward with this. I'll ask OTA team if they'd like to look into tackling re-introducing disruption tests in 4.11 based on your references. Thanks.

PR #27470, opened 2022-10-18
[OCPPLAN-9509] skip storage tests if cap disabled

qJkee:
Do not run storage related tests if storage capability disabled.

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -15,8 +15,11 @@ import (
 	"time"
 
 	"github.com/onsi/ginkgo/config"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/origin/pkg/monitor"
 	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+	"github.com/openshift/origin/test/extended/util/cluster"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
@@ -284,6 +287,21 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
 		return err
 	}
 
+	cl, err := configv1client.NewForConfig(restConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create client for CVO: %v", err)
+	}
+
+	cv, err := cl.ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get cvo version object from cluster: %v", err)
+	}
+
+	storageEnabled := true
+	if cluster.KnowsCapability(cv, "Storage") {
+		storageEnabled = cluster.HasCapability(cv, "Storage")
+	}
+
 	pc.SetEvents([]string{setupEvent})
 	pc.Run(ctx)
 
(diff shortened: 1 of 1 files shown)


pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -284,6 +287,25 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
...
+		return fmt.Errorf("failed to get cvo version object from cluster: %v", err)
+	}
+
+	storageEnabled := true
+	if cluster.KnowsCapability(cv, "Storage") {
+		storageEnabled = cluster.HasCapability(cv, "Storage")
+	}
+
+	if !storageEnabled {
+		fmt.Fprintln(opt.Out, "Skipping [sig-storage] tests because Storage capability disabled")
```
bparees, line 306: 100% of sig-storage tests can't be run when this cap is disabled?  including things like emptydir storage?

wking: nit, can we move this line down to an `else` block of the newly-conditional `tests = append(tests, storageTestsCopy...)`?

```go
if storageEnabled {
  // I thought about randomizing the order of the kube, storage, and openshift tests, but storage dominates our e2e runs, so it doesn't help much.
  storageTestsCopy := copyTests(storageTests)
  q.Execute(testCtx, storageTestsCopy, max(1, parallelism/2), status.Run) // storage tests only run at half the parallelism, so we can avoid cloud provider quota problems.
  tests = append(tests, storageTestsCopy...)
} else {
  fmt.Fprintln(opt.Out, "Skipping [sig-storage] tests because Storage capability disabled")
}
```

Keeping that collected together reduces the chances that someone touches one of the branches without noticing that they may also need to adjust the other.

pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -313,6 +335,11 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
 		return strings.Contains(t.name, "[sig-cli] oc adm must-gather")
 	})
 
+	// no need to count storage tests if storage disabled
+	if !storageEnabled {
+		storageTests = nil
```
wking, line 340: Ah, clearing this here keeps `expectedTestCount` from being inflated with subsequently-not-added storage tests.  Maybe this is where we should log the `Skipping ...`.

pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -359,10 +386,13 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
 		q.Execute(testCtx, kubeTestsCopy, parallelism, status.Run)
 		tests = append(tests, kubeTestsCopy...)
 
-		// I thought about randomizing the order of the kube, storage, and openshift tests, but storage dominates our e2e runs, so it doesn't help much.
-		storageTestsCopy := copyTests(storageTests)
-		q.Execute(testCtx, storageTestsCopy, max(1, parallelism/2), status.Run) // storage tests only run at half the parallelism, so we can avoid cloud provider quota problems.
-		tests = append(tests, storageTestsCopy...)
+		// run storage tests only if storage capability enabled
+		if storageEnabled {
```
wking, line 390: With [this `nil`-ing][1] above, I don't understand why we need a conditional here.  Can we leave this block of code unchanged?

[1]: https://github.com/openshift/origin/pull/27470#discussion_r999661183

qJkee: Right
Fixed at 9e66b24935a39a9d8f6abbab925179cf4e28d4a1

pkg/test/ginkgo/cmd_runsuite.go
qJkee: Added 9e66b24935a39a9d8f6abbab925179cf4e28d4a1

qJkee:
closing because of https://github.com/openshift/origin/pull/27481

PR #27471, opened 2022-10-18
Bug 2110926: add possibility to not run invariant tests

bertinatto:
This is specially useful for the CSI test suite. Partners typically certify their CSI drivers using that test suite, which usually reports
invariant failures.

CC @openshift/storage

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -526,6 +526,7 @@ func bindOptions(opt *runOptions, flags *pflag.FlagSet) {
 
 func bindTestOptions(opt *testginkgo.Options, flags *pflag.FlagSet) {
 	flags.BoolVar(&opt.DryRun, "dry-run", opt.DryRun, "Print the tests to run without executing them.")
+	flags.BoolVar(&opt.SkipInvariants, "skip-invariants", opt.SkipInvariants, "Skip invariant tests.")
 	flags.BoolVar(&opt.PrintCommands, "print-commands", opt.PrintCommands, "Print the sub-commands that would be executed instead.")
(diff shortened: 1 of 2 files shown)


cmd/openshift-tests/openshift-tests.go
```diff
@@ -526,6 +526,7 @@ func bindOptions(opt *runOptions, flags *pflag.FlagSet) {
 
 func bindTestOptions(opt *testginkgo.Options, flags *pflag.FlagSet) {
 	flags.BoolVar(&opt.DryRun, "dry-run", opt.DryRun, "Print the tests to run without executing them.")
+	flags.BoolVar(&opt.SkipInvariants, "skip-invariants", opt.SkipInvariants, "Skip invariant tests.")
```
jsafrane, line 529: Title "Don't run invariant tests on CSI test suite" implies that this option will be `false` in `openshift/csi` suite, but I don't see any code there to do that.

bertinatto: Old title, renamed.

jsafrane:
@bertinatto I am worried about "[bz-monitoring][invariant] alert/Watchdog must have no gaps or changes" failure - that test [does not fail often in presubmits](https://sippy.dptools.openshift.org/sippy-ng/tests/Presubmits/analysis?test=%5Bbz-monitoring%5D%5Binvariant%5D%20alert%2FWatchdog%20must%20have%20no%20gaps%20or%20changes), yet it's failing pretty constantly in e2e-aws-ovn-fips and e2e-aws-ovn-serial here.

Re-running all tests to get more data...
/retest

jsafrane:
`[bz-monitoring][invariant] alert/Watchdog must have no gaps or changes` failed everywhere. This is odd. I don't know how, but this PR seems to cause it.

PR #27473, opened 2022-10-18
disruption checking is disabled for prior releases because we lack a …

deads2k:
…good baseline

After checking, I think this is the minimal change to disable disruption on 4.9. We can try to remove all the code, but I think this allows us to retain data collection (such as it is), while disabling test failures.

/assign @dgoodwin


deads2k:
As in the 4.11 PR, this is a CI change. I'm bypassing the bug and backport ordering requirement and per Scott's slack message I'm setting the other required labels.

deads2k:
/override ci/prow/e2e-aws-single-node-serial

to improve pass rates.

PR #27483, opened 2022-10-20
ignore kubepodnotready alerts firing when it happens during node upgrade

deads2k:
I think this does what I mentioned.  It doesn't address the test variant in https://github.com/openshift/origin/blob/master/test/extended/prometheus/alerts.go#L24 .  Do we still need to actually run that test or do we only use it to properly wire ignores?

/assign @dgoodwin @neisw

### DIFF
--- a/pkg/monitor/monitorapi/types.go
+++ b/pkg/monitor/monitorapi/types.go
@@ -8,9 +8,9 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/sets"
-
+	"github.com/openshift/origin/pkg/synthetictests/platformidentification"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/monitor/monitorapi/types.go
```diff
@@ -266,6 +266,78 @@ func Not(filter EventIntervalMatchesFunc) EventIntervalMatchesFunc {
...
+func AlertFiringInNamespace(alertName, namespace string) EventIntervalMatchesFunc {
+	return func(eventInterval EventInterval) bool {
+		return And(
+			func(eventInterval EventInterval) bool {
+				locatorParts := LocatorParts(eventInterval.Locator)
+				eventAlertName := AlertFrom(locatorParts)
+				if eventAlertName != alertName {
+					return false
+				}
+				if strings.Contains(eventInterval.Message, `alertstate="pending"`) {
```
neisw, line 295: are pending & firing switched? or is that intentional and I'm missing it?

deads2k: nope

deads2k:
minor fixup, relabelling

deads2k:
no conflict, merging.

PR #27484, opened 2022-10-20
Ocpbugs 2275 remove templateservicebroker test

neisw:
OCPBUGS-2275: Removing TSBClient and tests that rely on it as we seem to be permaskipping these due to the template service broker is not installed.


dgoodwin:
Looks like this needs a rebase.

dgoodwin:
/lgtm

Best way to get these linked to jira is just to mention the bug number in the pr description exactly as it is in the jira url. (i.e. OCPBUGS-2275) This comment might do it as well.

PR #27485, opened 2022-10-21
Add jUnitSuite and status to e2e-test interval locators.

dgoodwin:
Today there is no clear way to search for failed tests in the regex
search in spyglass or the html files it uses because the status of the
test is not in the locator. Searching the messages is non-trivial and
probably too slow as there can be many per row.

There are typically 3 intervals for each e2e test, start, finish, and
the end result with proper start/stop time. This change adds
status/Passed or Failed or Flakes to the latter.

In addition, I have added the jUnitSuite to the locator as well, this
will make it a little more clear what invocation of openshift-tests
we're running when we view intervals html. (today we sort of just know
instinctively by its ordering or what tests it runs) This will be useful
for a coming PR where I plan to extract the failed test names
from intervals.

### DIFF
--- a/pkg/monitor/monitorapi/identification.go
+++ b/pkg/monitor/monitorapi/identification.go
@@ -6,8 +6,8 @@ import (
 	"strings"
 )
 
-func E2ETestLocator(testName string) string {
-	return fmt.Sprintf("e2e-test/%q", testName)
+func E2ETestLocator(testName, jUnitSuiteName string) string {
+	return fmt.Sprintf("e2e-test/%q jUnitSuite/%s", testName, jUnitSuiteName)
 }
 
 func IsE2ETest(locator string) bool {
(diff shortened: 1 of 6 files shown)


dgoodwin:
The status into locator change is important for being able to efficiently regex intervals when looking for what was going on with failing tests.

I am less certain about the jUnitSuite change. I was planning to use the e2e intervals to write out a simple file of test failures so we can push to sippy for risk analysis, and I need to include suite in that. However I'm now wondering if I shouldn't just bite the bullet and parse the junit xmls, not super fun, but probably not that hard. I think we can do a workflow that uses a python script and would make this easier. 

I'm interested in thoughts on that, but is the jUnitSuite name in the locator useful in general? I think I prefer it as it makes it more clear what you're looking at.

dgoodwin:
/close

This is wildly busted, and I cannot find a reasonable way to parse the locator tags when test names can contain everything from quotes to slashes to periods. Will leave them alone today, where it's e2e-test/testname and nothing else in the locator.

PR #27487, opened 2022-10-21
move alerting tests into invariants only

deads2k:
we currently run both [late] tests and invariants.  This moves us towards an invariant-only model to avoid double counting.


neisw:
Is it needing `make update-bindata`?

dgoodwin:
Yeah looks like it does need a bindata update. Otherwise lgtm.

deads2k:
> Yeah looks like it does need a bindata update. Otherwise lgtm.

labelling per comment

/retest-required

dgoodwin:
Is the new backstop test: disruption_tests: [sig-arch] Check if alerts are firing during or after upgrade success

Note: Old one was: [sig-instrumentation][Late] Alerts shouldn't report any unexpected alerts in firing or pending state [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel]

dgoodwin:
The failures cannot be related to removing the old alerting tests. I think it's time to override and stop waiting for CI here. I started work on alerting on Friday and it would be helpful to reduce the number of classes of alerting tests to deal with asap.

/override ci/prow/e2e-gcp-ovn

deads2k:
simple rebase, relabelling.

deads2k:
simple rebase, retagging.

PR #27488, opened 2022-10-21
Add a test suite suffix to the alert test name

deads2k:
This seems unnecessary at first, but some tools like spyglass don't
properly indicate the testsuite name.  While the best fix is to address
all of those components, we have not proven good at doing that.  We can
make significant progress by providing unique names for the alert
failures to prevent false "flakes" and properly present failures on the
UI and on ci-search.

/assign @dgoodwin


pkg/synthetictests/alerts.go
```diff
@@ -22,7 +25,7 @@ func testAlerts(events monitorapi.Intervals, restConfig *rest.Config, duration t
 		junit, err := alertTest.InvariantCheck(context.TODO(), restConfig, events, *recordedResource)
 		if err != nil {
 			ret = append(ret, &junitapi.JUnitTestCase{
-				Name: alertTest.InvariantTestName(),
+				Name: alertTest.InvariantTestName() + " for suite/" + testSuite,
```
dgoodwin, line 28: I'm a bit hesitant to nullify all our past test results for these, and it's arguably setting precedent that this should be done for all tests globally.

Where was the false flake happening in our stack? 

I did open a patch upstream to show the suite name in prow but they didn't want to enable it globally and instead asked for a config option.

deads2k: > I did open a patch upstream to show the suite name in prow but they didn't want to enable it globally and instead asked for a config option.

Did the config option get set in our prow?

dgoodwin: It did not get implemented, I did not find space to figure out how to add and test prow config options.

PR #27492, opened 2022-10-24
Add monitor store to resource watch

xueqzhan:
Main changes with this PR:
1. Reorg existing monitor files (for node, pod, cluster operator etc) so that codes can be called from within monitor package and outside monitor package (e.g. for resource watch)
2. Add one monitor store for resource watch. For resources supported by monitor (currently node, pod, cluster operator and cluster version), an effort has been made to maintain the same event formats. Events is not supported yet.
3. Monitor store can be selected instead of git repo. But in a real observer, both stores can be started with two different instances of openshift-tests binary. 

[TRT-469](https://issues.redhat.com/browse/TRT-469)

### DIFF
--- a/pkg/monitor/resourcewatch/operator/starter.go
+++ b/pkg/monitor/resourcewatch/operator/starter.go
@@ -2,12 +2,16 @@ package operator
 
 import (
 	"context"
+	"fmt"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/kubernetes"
 	"os"
 	"time"
 
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apiextensionsv1informer "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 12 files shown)


pkg/monitor/resourcewatch/storage/monitor_store.go
```diff
@@ -0,0 +1,295 @@
...
+		klog.Warningf("Decoding %s failed with error: %v", objUnstructured.GetName(), err)
+	}
+	monitor.EventAddOrUpdateFunc(context.TODO(), s.monitor, s.client, s.startTime, &eventObj)
+}
+
+// OnAdd processes add event for all monitored resources
+func (s *monitorStorage) OnAdd(obj interface{}) {
+	objUnstructured, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		klog.Warningf("Object is not unstructured: %v", obj)
```
dgoodwin, line 71: Looks like you want a return here.

pkg/monitor/resourcewatch/storage/monitor_store.go
```diff
@@ -0,0 +1,295 @@
...
+	gvk := objUnstructured.GroupVersionKind()
+	switch {
+	case gvk.Group == "" && gvk.Kind == "Node":
+		{
+			nodeObj := corev1.Node{}
+			err := decodeObject(objUnstructured, &nodeObj)
+			if err != nil {
+				klog.Warningf("Decoding %s failed with error: %v", objUnstructured.GetName(), err)
+			}
+			monitor.NodeAddFunc(nodeObj)
```
dgoodwin, line 82: This will get called when we fail to decode, looks like we should return here and for all examples like this below.

pkg/monitor/resourcewatch/operator/starter.go
```diff
@@ -37,12 +69,22 @@ func RunOperator(ctx context.Context, controllerCtx *controllercmd.ControllerCon
...
-		repositoryPath = repositoryPathEnv
+	var (
+		configStore storage.ResourceWatchStore
+	)
+	if store == monitorStoreStr {
+		eventsClient, err := kubernetes.NewForConfig(controllerCtx.KubeConfig)
+		if err != nil {
+			return err
+		}
+		configStore, err = storage.NewMonitorStorage(artifactDir, eventsClient)
```
dgoodwin, line 80: Would it make any more sense to launch the startClusterOperatorMonitoring funcs, instead of an actual separate store? This is a very loose question just something to think about. Are there benefits to separate store?

Then again, maybe the startX funcs die off depending on the other question posed here about duplicated intervals.

dgoodwin (commented):
This one was challenging for me to review but I'm starting to get a handle on it. Would be good to do some discussion sometime this week and that might help me clarify more. I was struggling with the dual entrypoints, and whether or not we could simplify the process for adding a new resource type to monitor in the future, as right now it requires several changes in several places. Hopefully this will become more clear with some discussion, if it's even feasible at all.

PR #27495, opened 2022-10-25
Remove unused cmd tests

ardaguclu:
Currently, we are migrating cmd tests to go tests. However, these tests have not been used for a while due to some reasons and there is no point to migrate them or store them.

This PR removes them for better migration.


ingvagabund:
Given all 5 files are on the list since Jul 2019, there's no reason to keep them around.
/lgtm

soltysh (changes requested):
/hold
they were commented out b/c we've couldn't make them work in the normal e2e that easily, before entirely getting rid of them, I'd prefer we have a review which of the cases they cover are still relevant and not tested and try to bring them back, if possible, and which we can safely drop.

ardaguclu:
@soltysh I did some preliminary analysis for these tests;

migrate.sh: most of migrate commands were removed with this [PR](https://github.com/openshift/oc/pull/1124) and there is no need to preserve these tests.

registry.sh: [this test](https://github.com/openshift/origin/blob/135202171d05dad058e872c7f8ca8e0b39a4b5dc/test/extended/images/oc_tag.go#L74) basically covers half of registry.sh tests. [Other half](https://github.com/openshift/origin/blob/de3cdf6a1840c4c127042f547c4f8a0caacea7cb/test/extended/testdata/cmd/test/cmd/registry.sh#L15-L17 ) is about `registry login` validations which might be skipped.

newapp.sh: newapp.sh does test for different kind of oc commands rather than only new-app command. Tests in here are mostly;

- oc new-app: [this test](https://github.com/openshift/origin/blob/135202171d05dad058e872c7f8ca8e0b39a4b5dc/test/extended/builds/new_app.go#L23 ) covers, it does not have high coverage comparing to newapp.sh but according to this comment "our image is missing git, so a lot of it doesn't work", probably huge amount of tests will not work already.
- oc new-build: [this test](https://github.com/openshift/origin/blob/135202171d05dad058e872c7f8ca8e0b39a4b5dc/test/extended/cli/builds.go#L64) covers
- oc istag: it seems that build tests above ^ also covers this case
- and the others commands are covered in various tests.

login.sh: login with token is tested in [here](https://github.com/openshift/origin/blob/135202171d05dad058e872c7f8ca8e0b39a4b5dc/test/extended/images/oc_tag.go#L128) and policy tests are mostly also done in [here](https://github.com/openshift/origin/blob/135202171d05dad058e872c7f8ca8e0b39a4b5dc/test/extended/cli/admin.go#L156). the rest of the tests are mostly validations and seems useful(we should add them in go), but I'm not sure that would be a blocker dropping this file.

policy.sh: mostly already covered in [here](https://github.com/openshift/origin/blob/135202171d05dad058e872c7f8ca8e0b39a4b5dc/test/extended/cli/admin.go#L156) and a few commands are not tested anywhere else `oc policy who-can`, `oc policy scc-subject-review`, `oc policy scc-review`. Should we add test cases for these commands?

To sum up, I can not say that that current go tests fully cover these tests. But according to my analysis, in terms of testing core functionality, we have good coverage. In terms of fine-grained flag validation tests, we have less coverage but can be added independently.

soltysh:
> policy.sh: mostly already covered in [here](https://github.com/openshift/origin/blob/135202171d05dad058e872c7f8ca8e0b39a4b5dc/test/extended/cli/admin.go#L156) and a few commands are not tested anywhere else oc policy who-can, oc policy scc-subject-review, oc policy scc-review. Should we add test cases for these commands?

@ardaguclu see what's possible to be added and add it, can be in a followup.

soltysh (approved):
The above description is sufficient and we're not making the situation worse than it is, but rather cleaning stuff we've not been running for a while now.

/hold cancel
/lgtm
/approve

ardaguclu:
I did not see that there are already `who-can` tests in here https://github.com/openshift/origin/blob/9886b19a182142d2f9ce4044932b5f0e8cdb8140/test/extended/cli/admin.go#L128-L153. I'll add for `scc-subject-review` and `scc-review`.

PR #27497, opened 2022-10-25
Update etcd scaling test for CPMS supported platforms

hasbro17:
Resolves: https://issues.redhat.com/browse/ETCD-329

The vertical scaling test workflow needs to account for the presence of the ControlPlaneMachineSet(CPMS) in which case it should not manually add new machines which would get deleted by the CPMSO which reconciles the control plane to the desired number of machines in spec.replicas.

This PR adds a new workflow that utilizes the CPMSO to automatically scale-up and scale-down when the platform supports CPMS (or detects its presence).

### DIFF
--- a/test/extended/etcd/helpers/helpers.go
+++ b/test/extended/etcd/helpers/helpers.go
@@ -3,6 +3,7 @@ package helpers
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -10,8 +11,10 @@ import (
 	o "github.com/onsi/gomega"
 
 	configv1 "github.com/openshift/api/config/v1"
+	machinev1 "github.com/openshift/api/machine/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	machineclient "github.com/openshift/client-go/machine/clientset/versioned"
+	machinev1client "github.com/openshift/client-go/machine/clientset/versioned/typed/machine/v1"
 	machinev1beta1client "github.com/openshift/client-go/machine/clientset/versioned/typed/machine/v1beta1"
 
 	bmhelper "github.com/openshift/origin/test/extended/baremetal"
@@ -176,6 +179,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
 				t.Logf("successfully deleted an excessive machine %q from the API (perhaps, created by a previous test)", machine.Name)
 			}
 		}
+		return true, nil
+	})
+}
+
+func DeleteSingleMachine(ctx context.Context, t TestingT, machineClient machinev1beta1client.MachineInterface) (string, error) {
+	waitPollInterval := 15 * time.Second
+	waitPollTimeout := 5 * time.Minute
+	t.Logf("Waiting up to %s to delete a machine", waitPollTimeout.String())
+
+	machineToDelete := ""
+	err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		machineList, err := machineClient.List(ctx, metav1.ListOptions{LabelSelector: masterMachineLabelSelector})
+		if err != nil {
+			return isTransientAPIError(t, err)
+		}
+
+		// Machine names are suffixed with an index number (e.g "ci-op-xlbdrkvl-6a467-qcbkh-master-0")
+		// so we sort to pick the lowest index, e.g master-0 in this example
+		machineNames := []string{}
+		for _, m := range machineList.Items {
+			machineNames = append(machineNames, m.Name)
+		}
+		sort.Strings(machineNames)
+		machineToDelete = machineNames[0]
+		t.Logf("attempting to delete machine %q", machineToDelete)
+
+		if err := machineClient.Delete(ctx, machineToDelete, metav1.DeleteOptions{}); err != nil {
+			// The machine we just listed should be present but if not, error out
+			if apierrors.IsNotFound(err) {
+				t.Logf("machine %q was listed but not found or already deleted", machineToDelete)
+				return false, fmt.Errorf("machine %q was listed but not found or already deleted", machineToDelete)
+			}
+			return isTransientAPIError(t, err)
+		}
+		t.Logf("successfully deleted machine %q", machineToDelete)
+
+		return true, nil
+	})
+
+	return machineToDelete, err
+}
+
+// IsCPMSActive returns true if the current platform's has an active CPMS
+// Not all platforms are supported (as of 4.12 only AWS and Azure)
+// See https://github.com/openshift/cluster-control-plane-machine-set-operator/tree/main/docs/user#supported-platforms
+func IsCPMSActive(ctx context.Context, t TestingT, cpmsClient machinev1client.ControlPlaneMachineSetInterface) (bool, error) {
+	// The CPMS singleton in the "openshift-machine-api" namespace is named "cluster"
+	// https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/bba395abab62fc12de4a9b9b030700546f4b822e/pkg/controllers/controlplanemachineset/controller.go#L50-L53
+	cpms, err := cpmsClient.Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	// The CPMS state must be active in order for the platform to be supported
+	// See https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/7961d1457c6aef26d3b1dafae962da2a2aba18ef/docs/user/installation.md#anatomy-of-a-controlplanemachineset
+	if cpms.Spec.State != machinev1.ControlPlaneMachineSetStateActive {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// EnsureReadyReplicasOnCPMS checks if status.readyReplicas on the cluster CPMS is n
+// this effectively counts the number of control-plane machines with the provider state as running
+func EnsureReadyReplicasOnCPMS(ctx context.Context, t TestingT, expectedReplicaCount int, cpmsClient machinev1client.ControlPlaneMachineSetInterface) error {
+	waitPollInterval := 5 * time.Second
+	waitPollTimeout := 10 * time.Minute
+	t.Logf("Waiting up to %s for the CPMS to have status.readyReplicas = %v", waitPollTimeout.String(), expectedReplicaCount)
+
+	return wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		cpms, err := cpmsClient.Get(ctx, "cluster", metav1.GetOptions{})
+		if err != nil {
+			return isTransientAPIError(t, err)
+		}
+
+		if cpms.Status.ReadyReplicas != int32(expectedReplicaCount) {
+			t.Logf("expected %d ready replicas on CPMS, got: %v,", expectedReplicaCount, cpms.Status.ReadyReplicas)
+			return false, nil
+		}
+
+		t.Logf("CPMS has reached the desired number of ready replicas: %v,", cpms.Status.ReadyReplicas)
+
 		return true, nil
 	})
 }
--- a/test/extended/etcd/vertical_scaling.go
+++ b/test/extended/etcd/vertical_scaling.go
@@ -44,6 +44,7 @@ var _ = g.Describe("[sig-etcd][Feature:EtcdVerticalScaling][Suite:openshift/etcd
 		machineClientSet, err := machineclient.NewForConfig(oc.KubeFramework().ClientConfig())
 		o.Expect(err).ToNot(o.HaveOccurred())
 		machineClient := machineClientSet.MachineV1beta1().Machines("openshift-machine-api")
+		cpmsClient := machineClientSet.MachineV1().ControlPlaneMachineSets("openshift-machine-api")
 		kubeClient := oc.KubeClient()
 
 		// make sure it can be run on the current platform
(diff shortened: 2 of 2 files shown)


hasbro17:
/hold
Until the scaling jobs pass on all platforms

test/extended/etcd/vertical_scaling.go
```diff
@@ -54,6 +55,64 @@ var _ = g.Describe("[sig-etcd][Feature:EtcdVerticalScaling][Suite:openshift/etcd
...
+			framework.Logf("Waiting for machine %q to be replaced", machineName)
+
+			memberName, err := scalingtestinglibrary.MachineNameToEtcdMemberName(ctx, oc.KubeClient(), machineClient, machineName)
+			err = errors.Wrapf(err, "failed to get etcd member name for deleted machine: %v", machineName)
+			o.Expect(err).ToNot(o.HaveOccurred())
+
+			// step 2: wait until the CPMSO scales-up by creating a new machine
+			// We need to check the cpms' status.readyReplicas because the phase of one machine will always be Deleting
+			// so we can't use EnsureMasterMachinesAndCount() since that looks for non-Deleting machines
+			err = scalingtestinglibrary.EnsureReadyReplicasOnCPMS(ctx, g.GinkgoT(), 4, cpmsClient)
```
hasbro17, line 78: Seems like we should watch for `replicas` instead of `readyReplicas`. The latter may never surge to 4 in this case as the deleted machine's member can get removed before we scale up.

From https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27497/pull-ci-openshift-origin-master-e2e-aws-ovn-etcd-scaling/1603457840947662848


Member removed at `19:19:21`
```
2022-12-15T19:19:21.221655410Z I1215 19:19:21.221269       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"16340e00-06fa-4526-9dc4-3da772936f35", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MemberRemove' removed member with ID: 2485787822400865737
2022-12-15T19:19:21.221655410Z I1215 19:19:21.221529       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"16340e00-06fa-4526-9dc4-3da772936f35", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'ScaleDown' successfully removed member: [ url: 10.0.182.231, name: ip-10-0-182-231.ec2.internal, id: 2485787822400865737 ] from the cluster
```
Member added as learner at `19:24:50`:
```
2022-12-15T19:24:50.597610435Z I1215 19:24:50.597552       1 event.go:285] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-etcd-operator", Name:"etcd-operator", UID:"16340e00-06fa-4526-9dc4-3da772936f35", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'MemberAddAsLearner' successfully added new member https://10.0.152.48:2380
```

Looking back into the clustermemberremoval controller and in particular the PR(https://github.com/openshift/cluster-etcd-operator/pull/947) to relax the scale-down constraint for unhealthy members where I think we may have inadvertently relaxed this for a healthy member too. Need to confirm.

hasbro17: Waiting for the fix to not allow scale-down or removal of the machine pending deletion before the new one is added https://github.com/openshift/cluster-etcd-operator/pull/978

hasbro17:
Looks like the node never came up for the azure run. Retesting:
/test e2e-azure-ovn-etcd-scaling

hasbro17:
Okay so this finally works.

For CPMS supported platforms (AWS and Azure currently) the test is relying the CPMSO to create the machine while we're reverting back to the manual deletion-then-creation workflow for unsupported platforms (GCP, Vsphere).

The scaling test itself is passing on azure but tripping up on a familiar (albeit unresolved) issue:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27497/pull-ci-openshift-origin-master-e2e-azure-ovn-etcd-scaling/1615502526784737280 
```

: [sig-arch] events should not repeat pathologically expand_less | 0s
-- | --
{  1 events happened too frequently  event happened 23 times, something is wrong: ns/openshift-authentication-operator deployment/authentication-operator - reason/OpenShiftAPICheckFailed "user.openshift.io.v1" failed with an attempt failed with statusCode = 503, err = the server is currently unable to handle the request result=reject }
```

This should be good to review now.

hasbro17:
Retesting but will eventually override the azure test if it keeps failing on the known pathological events case.

/test e2e-azure-ovn-etcd-scaling
/retest-required
/unhold

test/extended/etcd/vertical_scaling.go
```diff
@@ -54,6 +55,64 @@ var _ = g.Describe("[sig-etcd][Feature:EtcdVerticalScaling][Suite:openshift/etcd
 		err = errors.Wrap(err, "pre-test: timed out waiting for initial cluster state to have 3 running machines and 3 voting members")
 		o.Expect(err).ToNot(o.HaveOccurred())
 
+		cpmsSupported, err := scalingtestinglibrary.PlatformSupportsCPMS(ctx, g.GinkgoT(), cpmsClient)
+		err = errors.Wrap(err, "pre-test: failed to determine if ControlPlaneMachineSet is present")
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		if cpmsSupported {
+
+			// TODO: Add cleanup step to recover back to 3 running machines and members if the test fails
```
tjungblu, line 64: do we still have serial jobs where this could become an issue?

test/extended/etcd/vertical_scaling.go
```diff
@@ -54,6 +55,64 @@ var _ = g.Describe("[sig-etcd][Feature:EtcdVerticalScaling][Suite:openshift/etcd
 		err = errors.Wrap(err, "pre-test: timed out waiting for initial cluster state to have 3 running machines and 3 voting members")
 		o.Expect(err).ToNot(o.HaveOccurred())
 
+		cpmsSupported, err := scalingtestinglibrary.PlatformSupportsCPMS(ctx, g.GinkgoT(), cpmsClient)
+		err = errors.Wrap(err, "pre-test: failed to determine if ControlPlaneMachineSet is present")
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		if cpmsSupported {
```
tjungblu, line 62: do you think we can move that whole if code block into a dedicated test function? have the setup code above as an initializer for both

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +182,78 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
+
+	machineToDelete := ""
+	err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		machineList, err := machineClient.List(ctx, metav1.ListOptions{LabelSelector: masterMachineLabelSelector})
+		if err != nil {
+			return isTransientAPIError(t, err)
+		}
+
+		// No need to be deterministic since it doesn't matter which machine we delete
+		machineToDelete = machineList.Items[0].Name
```
tjungblu, line 198: there might be two invocations of the wait.Poll, especially if the delete failed on some transient client/reponse error. So we _could_ end up with two deleted masters here. 

I would just remove the wait.Poll altogether and rely on the machineClient retries on a single chosen machine name.

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +182,78 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
+
+		t.Logf("attempting to delete machine %q", machineToDelete)
+
+		if err := machineClient.Delete(ctx, machineToDelete, metav1.DeleteOptions{}); err != nil {
+			return isTransientAPIError(t, err)
```
tjungblu, line 203: also regarding the polling, if it's already deleted this will error out. So I think we need another:
> if apierrors.IsNotFound(err) {

test/extended/etcd/vertical_scaling.go
hasbro17: I could move the setup code inside `BeforeEach()` so we do the common setup before each `It()` spec block.

And I would have two `It()` spec blocks in that case which would both always run except one would always get skipped based on the conditional. Just to illustrate what I'm thinking:
```Go

g.Describe("[sig-etcd][Feature:EtcdVerticalScaling][Suite:openshift/etcd/scaling] etcd [apigroup:config.openshift.io]", func() {

	var cpmsActive bool
	// Plus all the other common setup variables to pass to the It() closures below

	g.BeforeEach(func() {
		// Common setup and compute cpmsActive
		cpmsActive := ...
	})


	g.It("is able to vertically scale up and down with a single node when CPMS is active [Timeout:60m][apigroup:machine.openshift.io]", func() {
		if cpmsActive {
			Skip("CPMS is inactive so cannot use CPMS")
		}
		
		// New CPMS based workflow

	}


	g.It("is able to vertically scale up and down with a single node when CPMS is inactive [Timeout:60m][apigroup:machine.openshift.io]", func() {
		if cpmsActive {
			Skip("CPMS is active so cannot manually create machines")
		}
		
		// Old workflow
	}
	
}
```

Grammatically this is a bit awkward and not sure the conditional is too clear here.

I'm no ginkgo expert but I would think something like the following reads better with a `When` or `Describe` block which are both equivalent: 

```Go
g.When("CPMS is active", func() {

	g.It("is able to vertically scale up and down with a single node when CPMS is active [Timeout:60m][apigroup:machine.openshift.io]", func() {
	...
	}
}

g.When("CPMS is inactive", func() {

	g.It("is able to vertically scale up and down with a single node when CPMS is active [Timeout:60m][apigroup:machine.openshift.io]", func() {
	...
	}
}
```

But again I'm not sure how to conditionally run either `When()` block here so that's why I used `It()` in the former example.
But let me know if that's what you meant.

test/extended/etcd/helpers/helpers.go
hasbro17: Good catch. I'm going to make it deterministic by always picking the master name with the lowest index since all machines are suffixed e.g `...-master-0` `...-master-1` etc.

> remove the wait.Poll altogether and rely on the machineClient retries

The wait.Poll is for retrying, so not sure if you meant replacing it with another util? The machineClient won't auto retry. Or do you mean just loop the Delete until it goes through. Would need some backoff though in case of transient API errors which is what the wait.Poll does.

test/extended/etcd/vertical_scaling.go
hasbro17: Not currently no. But we may have other scaling related tests that may have to run serially in the future. But I'd prefer to leave that as a todo for when we actually have those.

test/extended/etcd/helpers/helpers.go
tjungblu: ahh, thank you. I thought that there's some retry utility baked into the clients already.  As long as you pick the master deterministically it's fine :)

test/extended/etcd/vertical_scaling.go
tjungblu: I was thinking along the lines of 

> scalingtestinglibrary.SkipIfUnsupportedPlatform(ctx, oc)

so your first description fits the best. I just want to save some refactoring steps for [ETCD-330](https://issues.redhat.com/browse/ETCD-330) - where I would also need to detect whether CPMS active or not.

The g.When() indeed reads better, but if it's taking too much time for you to figure it out you can leave this for me :)

hasbro17: Ah okay I get the context for this now. I'm still not liking the former layout of skipping one of the two `It()` blocks and haven't reviewed https://github.com/openshift/origin/pull/27461/files enough to see if we can't make that an entirely separate test case for that.

The quorum checker test would even need to disable the CPMS as well (when active) so we would have to serialize or order that so it can clean up and not be disruptive.

In the interest of landing a smaller fix temporarily (to improve pass rates and unblock other debugging efforts), I would lean towards punting the refactor for when we actually have a second test case. But not to say it's your problem :) I will refactor the test to make it more Ginkgo-readable as a follow up.

This would be easier to backport to 4.12 as well in case we have to refactor again in ETCD-330. WDYT?

tjungblu: >  and haven't reviewed https://github.com/openshift/origin/pull/27461/files enough to see if we can't make that an entirely separate test case for that.

save yourself the time, I'll likely write this again from scratch. 

> The quorum checker test would even need to disable the CPMS as well (when active) so we would have to serialize or order that so it can clean up and not be disruptive.

I personally would just have the checker test run when there is no CPMS.
 
> I would lean towards punting the refactor for when we actually have a second test case. But not to say it's your problem :) I will refactor the test to make it more Ginkgo-readable as a follow up.

All good, I'm fine with refactoring this in the other ticket. Let's get the pass rates fixed first before we complicate.

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +183,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
+		}
+		sort.Strings(machineNames)
+		machineToDelete = machineNames[0]
+		t.Logf("attempting to delete machine %q", machineToDelete)
+
+		if err := machineClient.Delete(ctx, machineToDelete, metav1.DeleteOptions{}); err != nil {
+			// The machine we just listed should be present but if not, error out
+			if apierrors.IsNotFound(err) {
+				t.Logf("machine %q was listed but not found or already deleted", machineToDelete)
+				return false, fmt.Errorf("machine %q was listed but not found or already deleted", machineToDelete)
```
tjungblu, line 212: shouldn't this be:

> return true, null

after all, the expectation of the whole function is that you want to delete the machine? :dagger:

hasbro17: Yes we want to delete the machine but having just listed the machine in L93 and then subsequently finding out the machine is no longer present means something has gone wrong.

Deleting the machine doesn't mean it should just go away immediately. You need the replacement machine to be created, added, and promoted as a member, and then have member removal and deletion hook removal etc before the machine should go away.

If at this point deleting the the machine results in an `IsNotFound ` error then the test really can't verify anything  anymore in above sequence for the vertical scaling workflow since the machine is already gone.

hasbro17:
/retest-required
/test e2e-azure-ovn-etcd-scaling
/test e2e-vsphere-ovn-etcd-scaling

Test is fine on vshpere (just the know disruption test failures again). Azure has passed with the new workflow before and looks to be an infra problem.

hasbro17:
Scaling jobs look good. Vsphere has passed before, and azure has the test passing but seeing other failures:
```
Thanos queriers not connected to all Prometheus sidecars: server_error: server error: 504
```

/retest-required

hasbro17:
Needs a backport to 4.12 as well.

/cherry-pick release-4.12

Although I'll forgot to classify this as a 4.13 bug and now need a 4.13 verified and a 4.12 bug to depend on that now ☹️

PR #27508, opened 2022-11-01
Add test risk analysis html for spyglass

xueqzhan:
[TRT-644](https://issues.redhat.com/browse/TRT-644)

### DIFF
--- a/e2echart/test-risk-analysis.html
+++ b/e2echart/test-risk-analysis.html
@@ -1,0 +1,104 @@
+<html lang="en">
+<head>
+    <title>Risk Analysis</title>
+    <script src="https://code.jquery.com/jquery-3.2.1.min.js"></script>
+    <meta name="description"
+          content="Risk analysis is performed by Sippy to attempt to determine if the failures in this job are abnormal when compared to results for similar jobs over the past week, and amidst on-going incidents in the CI infrastructure. Risk analysis API will not catch everything and is a relatively simple implementation today, please reach out to the Technical Release Team if you spot abnormalities or have suggestions.">
+</head>
+<body onLoad="buildTestCaseTable('#test_case_results')">
+<p>
+    <a target="_blank" href="TEST_RISK_ANALYSIS_SIPPY_URL_GOES_HERE">Link to Sippy</a>
+</p>
+
+<table id="test_case_results" border="1" width="100%">
+</table>
+<p>&nbsp;</p>
+
+<script>
+    var testResult = TEST_RISK_ANALYSIS_JSON_GOES_HERE
+    var testLinkPrefix = "https://sippy.dptools.openshift.org/sippy-ng/tests/"
+    var testLinkSuffix = "/analysis?test="
+
+    function buildOpenBugs(openBugs) {
+        td$ = $('<td/>')
+        if (openBugs.length > 0) {
+            cell$ = $('<ul/>')
+            for (var i = 0; i < openBugs.length; i++) {
+                li$ = $('<li>')
+                bug = "<a target=\"_blank\" href=" + openBugs[i].url + ">" + openBugs[i].id + "</a>: " + openBugs[i].summary
+                li$.append($('<li>').html(bug))
+                cell$.append(li$)
+            }
+            td$.append(cell$)
+        }
+        return td$
+    }
+
+    function buildRiskReasons(reasons) {
+        td$ = $('<td/>')
+        if (reasons.length > 0) {
+            cell$ = $('<ul/>')
+            for (var i = 0; i < reasons.length; i++) {
+                li$ = $('<li>')
+                li$.append($('<li>').html(reasons[i]))
+                cell$.append(li$)
+            }
+            td$.append(cell$)
+        }
+        return td$
+    }
+
+    function buildRiskLevel(level) {
+        td$ = $('<td/>')
+        if (level.Level >= 10) {
+            td$.css("background-color", "red");
+        } else if (level.Level >= 5) {
+            td$.css("background-color", "yellow");
+        } else {
+            td$.css("background-color", "green");
+        }
+        td$.append(level.Name)
+        return td$
+    }
+
+    // Build Test Case Table
+    function buildTestCaseTable(selector) {
+        // Add table headers
+        addColumnHeaders(selector);
+
+        // Build Overall Row
+        var row$ = $('<tr/>');
+        row$.append($('<td/>').html("Overall"));
+        row$.append(buildRiskLevel(testResult.OverallRisk.Level))
+        row$.append(buildRiskReasons(testResult.OverallRisk.Reasons))
+        row$.append(buildOpenBugs(testResult.OpenBugs))
+        $(selector).append(row$);
+
+        // First sort the tests by risk factor
+        testResult.Tests.sort(function(a, b){return b.Risk.Level.Level - a.Risk.Level.Level})
+        // Build rows for all tests
+        for (var i = 0; i < testResult.Tests.length; i++) {
+            var row$ = $('<tr/>');
+            testUrl = encodeURI(testLinkPrefix + testResult.Release + testLinkSuffix + testResult.Tests[i].Name)
+            row$.append($('<td/>').html("<a target=\"_blank\" href=" + testUrl + ">" + testResult.Tests[i].Name + "</a>"));
+            row$.append(buildRiskLevel(testResult.Tests[i].Risk.Level))
+            row$.append(buildRiskReasons(testResult.Tests[i].Risk.Reasons))
+            row$.append(buildOpenBugs(testResult.Tests[i].OpenBugs))
+            $(selector).append(row$);
+        }
+    }
+
+    function addColumnHeaders(selector) {
+        var headerTr$ = $('<tr/>');
+        headerTr$.append($('<th/>').html("Test Name"));
+        headerTr$.append($('<th/>').html("Risk Level"));
+        headerTr$.append($('<th/>').html("Risk Reason"));
+        headerTr$.append($('<th/>').html("Open Bugs"));
+
+        $(selector).append(headerTr$);
+    }
+
+</script>
+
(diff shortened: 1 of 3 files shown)


dgoodwin:
I need this asap, but I can't find any results in the presubmits that show it working, is there a non aggregated job we could use with payload-job command ken?

dgoodwin:
Wrong PR :0 ignore last comment. :)

e2echart/test-risk-analysis.html
```diff
@@ -0,0 +1,106 @@
...
+<body onLoad="buildTestCaseTable('#test_case_results')">
+<h1>
+</h1>
+<p>
+    "Risk analysis is performed by Sippy[<a href="TEST_RISK_ANALYSIS_SIPPY_URL_GOES_HERE">link</a>] to attempt to
+    determine if the failures in this job are abnormal when compared to results for similar jobs over the past week,
+    and amidst on-going incidents in the CI infrastructure. Risk analysis API will not catch everything and is a
+    relatively simple implementation today, please reach out to the Technical Release Team if you spot abnormalities
+    or have suggestions."
+</p>
```
stbenjam, line 15: Spyglass supports "info" tooltips, if you'd prefer to put most of this info except maybe the link in there.  Example:

https://github.com/openshift/release/blob/9548766e2f234917b40732d279fa1fe6cf7f6dad/ci-operator/step-registry/gather/extra/gather-extra-commands.sh#L577-L579

pkg/riskanalysis/cmd.go
```diff
@@ -91,5 +93,16 @@ func (opt *Options) Run() error {
 	}
 	fmt.Fprintf(opt.Out, "Successfully wrote: %s\n", outputFile)
 
+	// Write html file for spyglass
+	riskAnalysisHTMLTemplate := testdata.MustAsset("e2echart/test-risk-analysis.html")
+	title := fmt.Sprintf("Risk Analysis For Job: %s, ID: %d", finalProwJobRun.ProwJob, finalProwJobRun.ID)
```
stbenjam, line 98: Nit: the spyglass page already has the job name and ID prominently displayed at the top, I'd omit this and just leave it as "Risk Analysis"

e2echart/test-risk-analysis.html
xueqzhan: Does the tooltip support links? If one puts the cursor on and get the tooltip, can he click on the link to get to sippy? I just experimented and it doesn't seem to support that.

stbenjam: Right, I suggested not putting the link.  You can leave it here if you want, but it'd be more consistent if we described what the page is doing in the tooltip.

xueqzhan: Devan suggested the link. @dgoodwin do you prefer this text in tooltip without the link to Sippy, or a paragraph on the top of the page?

dgoodwin: My thought was here was just to link to top level Sippy page, so a user could see what it was. Not directly to risk analysis. I can live without either though if anyone feels strongly.

And if we keep it, I'd just do the word Sippy as the link, not a [] after it, that was just markdown slipping into my suggestions for the card.

stbenjam: Sorry if I wasn't clear, I just meant putting the link outside of the tooltip somewhere on the page would be fine.

e2echart/test-risk-analysis.html
```diff
@@ -0,0 +1,108 @@
...
+          abnormal when compared to results for similar jobs over the past week, and amidst on-going incidents
+          in the CI infrastructure. Risk analysis API will not catch everything and is a relatively simple
+          implementation today, please reach out to the Technical Release Team if you spot abnormalities or
+          have suggestions.">
+</head>
+<body onLoad="buildTestCaseTable('#test_case_results')">
+<h1>
+</h1>
+<p>
+    "<a href="TEST_RISK_ANALYSIS_SIPPY_URL_GOES_HERE">Link to Sippy</a>"
```
stbenjam, line 16: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27508/pull-ci-openshift-origin-master-e2e-aws-ovn-serial/1589660281175281664/artifacts/e2e-aws-ovn-serial/openshift-e2e-test/artifacts/junit/test-risk-analysis.html

I don't think this should be in quotes.  

This also just goes to the main page, will it eventually go somewhere else?

stbenjam:
Is there a corresponding PR to openshift/release to add this as an HTML page spyglass should show? There's no reason it can't be there now so we could see the results in the presubmits

e2echart/test-risk-analysis.html
```diff
@@ -0,0 +1,108 @@
...
+        row$.append(buildRiskReasons(testResult.OverallRisk.Reasons))
+        row$.append(buildOpenBugs(testResult.OpenBugs))
+        $(selector).append(row$);
+
+        // First sort the tests by risk factor
+        testResult.Tests.sort(function(a, b){return b.Risk.Level.Level - a.Risk.Level.Level})
+        // Build rows for all tests
+        for (var i = 0; i < testResult.Tests.length; i++) {
+            var row$ = $('<tr/>');
+            row$.append($('<td/>').html(testResult.Tests[i].Name));
```
stbenjam, line 87: Is it possible to put a link to the test analysis page in Sippy for each test? That seems more useful to me than a link to the main page of sippy

e2echart/test-risk-analysis.html
xueqzhan: There is nothing planned for this. The risk analysis API requires a body. The top page of Sippy is put here just for convenience purposes. I am open for suggestions though.

e2echart/test-risk-analysis.html
xueqzhan: > Is there a corresponding PR to openshift/release to add this as an [HTML page spyglass should show](https://github.com/openshift/release/blob/master/core-services/prow/02_config/_config.yaml#L108-L114)? There's no reason it can't be there now so we could see the results in the presubmits

@deepsm007 Deep has a PR here: https://github.com/openshift/release/pull/33731

You meant that PR can be merged now so that we can see the spyglass in our job runs. Good point.

e2echart/test-risk-analysis.html
```diff
@@ -0,0 +1,107 @@
...
+        row$.append(buildOpenBugs(testResult.OpenBugs))
+        $(selector).append(row$);
+
+        // First sort the tests by risk factor
+        testResult.Tests.sort(function(a, b){return b.Risk.Level.Level - a.Risk.Level.Level})
+        // Build rows for all tests
+        for (var i = 0; i < testResult.Tests.length; i++) {
+            var row$ = $('<tr/>');
+            testUrl = encodeURI(testLinkPrefix + testResult.Release + testLinkSuffix + testResult.Tests[i].Name)
+            row$.append($('<td/>').html("<a href=" + testUrl + ">" + testResult.Tests[i].Name + "</a>"));
```
stbenjam, line 86: Can we get this to open in a new tab?

```suggestion
            row$.append($('<td/>').html("<a target="_blank" href=" + testUrl + ">" + testResult.Tests[i].Name + "</a>"));
```

stbenjam:
It looks like the table is getting cut off, do we control the size of the iframe? The intervals chart seems to be able to be taller.

image

e2echart/test-risk-analysis.html
```diff
@@ -0,0 +1,107 @@
...
+    <meta name="description"
+          content="Risk analysis is performed by Sippy to attempt to determine if the failures in this job are
+          abnormal when compared to results for similar jobs over the past week, and amidst on-going incidents
+          in the CI infrastructure. Risk analysis API will not catch everything and is a relatively simple
+          implementation today, please reach out to the Technical Release Team if you spot abnormalities or
+          have suggestions.">
+</head>
+<body onLoad="buildTestCaseTable('#test_case_results')">
+<p>
+    <a href="TEST_RISK_ANALYSIS_SIPPY_URL_GOES_HERE">Link to Sippy</a>
```
stbenjam, line 14:
```suggestion
    <a target="_blank" href="TEST_RISK_ANALYSIS_SIPPY_URL_GOES_HERE">Link to Sippy</a>
```

e2echart/test-risk-analysis.html
```diff
@@ -0,0 +1,107 @@
...
+    var testLinkPrefix = "https://sippy.dptools.openshift.org/sippy-ng/tests/"
+    var testLinkSuffix = "/analysis?test="
+
+    function buildOpenBugs(openBugs) {
+        td$ = $('<td/>')
+        if (openBugs.length > 0) {
+            cell$ = $('<ul/>')
+            for (var i = 0; i < openBugs.length; i++) {
+                li$ = $('<li>')
+                bug = "<a href=" + openBugs[i].url + ">" + openBugs[i].id + "</a>: " + openBugs[i].summary
```
stbenjam, line 31:
```suggestion
                bug = "<a target=\"_blank\" href=" + openBugs[i].url + ">" + openBugs[i].id + "</a>: " + openBugs[i].summary
```

e2echart/test-risk-analysis.html
```diff
@@ -0,0 +1,107 @@
...
+<head>
+    <title>Risk Analysis</title>
+    <script src="https://code.jquery.com/jquery-3.2.1.min.js"></script>
+    <meta name="description"
+          content="Risk analysis is performed by Sippy to attempt to determine if the failures in this job are
+          abnormal when compared to results for similar jobs over the past week, and amidst on-going incidents
+          in the CI infrastructure. Risk analysis API will not catch everything and is a relatively simple
+          implementation today, please reach out to the Technical Release Team if you spot abnormalities or
+          have suggestions.">
+</head>
```
stbenjam, line 11: Can we get the text all on one line? Seems to not like the white space.

image

PR #27525, opened 2022-11-04
OCPBUGS-3282: Fix apigroup tag missing closing brace.

dgoodwin:
This was causing the test to be skipped:

: [sig-scheduling][Early] The openshift-console console pods [apigroup:console.openshift.io should be scheduled on different nodes [Suite:openshift/conformance/parallel]
Reason: skipped because the following required API groups are missing: console.openshift.io should be scheduled on different nodes [Suite:openshift/conformance/parallel


dgoodwin:
/jira refresh

Aparently there's target version and fix version.

PR #27528, opened 2022-11-07
Set Jakub as a Test Platform team lead

petr-muller:
cc @jmguzik


jmguzik (approved):

petr-muller:
Looks like this also needs a `/lgtm`

PR #27405, opened 2022-09-07
Add bond cni test

mmirecki:
Adding a test case for the bond-cni.
The test checks if a pod with a bond is created successfully, and then checks connectivity between two pods with bond interfaces.

PR #27536, opened 2022-11-09
[release-4.12] Add bond cni test

mmirecki:
4.12 backport of https://github.com/openshift/origin/pull/27405
Adding a test case for the bond-cni.
The test checks if a pod with a bond is created successfully, and then checks connectivity between two pods with bond interfaces.


mmirecki:
/close
This was backported as part of: https://github.com/openshift/origin/pull/27541

PR #27547, opened 2022-11-14
OCPBUGS-3633: Revert "Merge pull request #27533 from dgoodwin/merge-alert-backstops"

dgoodwin:
[TRT-674](https://issues.redhat.com//browse/TRT-674)

This reverts commit a747f94e25fcedd563822052834546a2becdf4a3, reversing
changes made to 63e7542266cce22ac87c5658e749ceb4c88fe7e6.

There appears to be a problem with recording flakes, none in 4.13 jobs since Nov 4th. There was some odd flaking code I moved here, but tried to maintain compatibility. I suspect some of the magic involved in reporting a flake for the currently running test doesn't work if you move the code for some reason. 

Due to this we're now failing payloads because we don't get enough results, I think because we usually flake this test and now are below 6 results because those don't get recorded.

4.12 is more confusing, this code should be live there, payloads are failing the same way, but I do still see some flakes getting reported in 4.12 jobs...


dgoodwin:
Bot seems a little confused. Says it's still waiting on these but they were successfully run.

/override ci/prow/e2e-aws-ovn-serial ci/prow/e2e-gcp-ovn-upgrade

PR #27552, opened 2022-11-15
Revert "refactor how tests are run"

dgoodwin:
Reverts openshift/origin#27516

We suspect this has caused problems with flakes when a certain function is used (testresult.Flakef)

Test revert.


dgoodwin:
I think we got it fixed in: https://github.com/openshift/origin/pull/27553

PR #27554, opened 2022-11-16
OCPBUGS-3819: Add bond cni test

mmirecki:
Adding a test case for the bond-cni.
The test checks if a pod with a bond is created successfully, and then checks connectivity between two pods with bond interfaces.


mmirecki:
/cc @knobunc
Could you please take a look

mmirecki:
@knobunc @deads2k @dcbw @danwinship @smarterclayton @bparees @mfojtik 
Can you please take a look at this?

mmirecki:
@dgoodwin @akram @coreydaley @dmage @gangwgr @jadhaj @jitendar-singh
Could you please apply the  backport-risk-assessed, cherry-pick-approved labels?
Thanks

mmirecki:
@kasturinarra @pamoedom @prietyc123 @stbenjam 
Can you please apply the backport-risk-assessed  label?
Thanks

pamoedom:
Hi @mmirecki, please note that from the persons you pinged above, only @stbenjam has credentials to add `backport-risk-assessed` label. You can check [here](https://github.com/openshift/release/blob/master/core-services/prow/02_config/openshift/origin/_pluginconfig.yaml) the whole list for both labels if needed.

Regards.

bparees:
i'll apply it manually.  Given this is affecting tests and not production code, the risk is low (the only real risk is that it opens a hole for a regression to be introduced elsewhere, and since it's adding tests, not changing them, that risk is also low)

PR #27553, opened 2022-11-15
OCPBUGS-3633: Fix flake reporting for certain tests.

dgoodwin:
In PR #27516 we suspect reporting of flakes broke due to a missed
assumption that test.flake accompanied test.success. Our new goal is to
more clearly have just one status set, so we're going to lean into the
new approach and properly break out the flake state into it's own case.

PR #27559, opened 2022-11-17
Flake and improve alert tests

dgoodwin:
TRT has decided with input from @deads2k that we're ok with moving the generic alert backstop tests to always flake, as the signal is not as high value. The post-upgrade alert test fails 30% of the time globally, and the post-conformance variant is also not stellar, and both disproportionately affect certain NURPs.

In sippy as of https://github.com/openshift/sippy/pull/685 we will begin tracking the reasons this test is flaking in the db by storing metadata about what alerts fired, so we'll have good insight into what may be causing issues, provided somebody remembers to look.

PR also re-introduces the refactor to merge the two backstop alert tests into a common code path, which they once were forked from. We thought this caused a regression, turned out it was something else, so this change should be good provided we start getting good payloads again after https://github.com/openshift/origin/pull/27553.

Also improves output of the alert tests for better parsing in sippy. Clearly identifies if we accept or reject the failure, and if there  is an associated bug.

### DIFF
--- a/pkg/alerts/allowed_conformance.go
+++ b/pkg/alerts/allowed_conformance.go
@@ -1,0 +1,100 @@
+package alerts
+
+import (
+	"context"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	helper "github.com/openshift/origin/test/extended/util/prometheus"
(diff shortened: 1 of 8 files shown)


dgoodwin:

The job just missed on pod sandbox.
Checked optional jobs, mostly aws install failures, which seem far more common than sippy indicates they are lately.

PR #27560, opened 2022-11-18
Trim stdout and stderr to the last 4K bytes to keep log size manageable

DennisPeriquet:
TRT-696

This will help prevent the build-log.txt file for e2e tests from getting too big.


test/extended/util/client.go
```diff
@@ -799,6 +799,17 @@ func (c *CLI) start(stdOutBuff, stdErrBuff *bytes.Buffer) (*exec.Cmd, error) {
 	return cmd, err
 }
 
+// getByteOffset calcuates a byte offset in a byte slide such that when using
+// that offset, we get the last 4096 bytes.
+func getByteOffset(byteString []byte) int {
```
deads2k, line 804: `func getStartingIndexForLastN(byteString, offset) int`

deads2k:
won't work perfectly with runes, but let's start getting things under control

/lgtm

deads2k:
/lgtm cancel

only change the part where we output on errors.

test/extended/util/client.go
```diff
@@ -818,7 +829,7 @@ func (c *CLI) outputs(stdOutBuff, stdErrBuff *bytes.Buffer) (string, string, err
 		return stdOut, stdErr, nil
 	case *exec.ExitError:
 		framework.Logf("Error running %v:\nStdOut>\n%s\nStdErr>\n%s\n", cmd, stdOut, stdErr)
-		wrappedErr := fmt.Errorf("Error running %v:\nStdOut>\n%s\nStdErr>\n%s\n%w\n", cmd, stdOut, stdErr, err)
+		wrappedErr := fmt.Errorf("Error running %v:\nStdOut>\n%s\nStdErr>\n%s\n%w\n", cmd, stdOut[getStartingIndexForLastN(stdOutBytes, 4096):], stdErr[getStartingIndexForLastN(stdErrBytes, 4096):], err)
```
deads2k, line 832: bytes don't match strings since strings are trimmed, but that probably doesn't matter neough here.

PR #27561, opened 2022-11-21
Fail on APIs removed in the next release

soltysh:
This will hard fail until we fix the problems in:
~1. api & auth & etcd https://issues.redhat.com/browse/OCPBUGS-3841~
~2. cvo https://issues.redhat.com/browse/OCPBUGS-3810~
~3. monitoring https://issues.redhat.com/browse/OCPBUGS-3924~

/assign @deads2k 
/hold

### DIFF
--- a/test/extended/apiserver/api_requests.go
+++ b/test/extended/apiserver/api_requests.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -13,11 +14,20 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	apiserverclientv1 "github.com/openshift/client-go/apiserver/clientset/versioned/typed/apiserver/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 
-	"github.com/openshift/origin/pkg/test/ginkgo/result"
 	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const unknownUser = "<unknown>"
+
+var (
+	// allowedUsers is a list of users who are allowed to make API calls to deprecated APIs
+	allowedUsers = sets.NewString("system:serviceaccount:openshift-kube-storage-version-migrator:kube-storage-version-migrator-sa")
+	// allowedUsersRE is a regular expression for a user being added in test/extended/etcd/etcd_test_runner.go
+	allowedUsersRE = regexp.MustCompile(`test-etcd-storage-path\w+`)
 )
 
 var _ = g.Describe("[sig-arch][Late]", func() {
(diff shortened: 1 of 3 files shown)


test/extended/apiserver/api_requests.go
```diff
@@ -85,8 +85,7 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 		sort.Strings(failureOutput)
 
 		if len(failureOutput) > 0 {
-			// don't insta-fail all of CI
-			result.Flakef(strings.Join(failureOutput, "\n"))
```
wking, line 89: [Unit test results][1]:

```
# github.com/openshift/origin/test/extended/apiserver
test/extended/apiserver/api_requests.go:18:2: imported and not used: "github.com/openshift/origin/pkg/test/ginkgo/result"
```

Can we drop the import, now that you're removing the only consumer here?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27561/pull-ci-openshift-origin-master-unit/1594661183284056064#1:build-log.txt%3A23

soltysh:
/hold
this is currently blocked by https://github.com/openshift/oauth-apiserver/pull/89 because oauth-apiserver is the last consumer of `flowcontrol.k8s.io/v1beta2` API.

soltysh:
One final bit needed to complete the puzzle is https://github.com/openshift/kubernetes/pull/1624

test/extended/apiserver/api_requests.go
```diff
@@ -13,13 +14,22 @@ import (
...
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
+const unknownUser = "<unknown>"
+
+var (
+	// allowedUsers is a list of users who are allowed to make API calls to deprecated APIs
+	allowedUsers = sets.NewString("system:serviceaccount:openshift-kube-storage-version-migrator:kube-storage-version-migrator-sa")
+	// allowedUsersRE is a regular expression for a user being added in test/extended/etcd/etcd_test_runner.go
+	allowedUsersRE = regexp.MustCompile(`test-etcd-storage-path\w+`)
```
benluddy, line 30: Nit:

```suggestion
	allowedUsersRE = regexp.MustCompile(`^test-etcd-storage-path\w+$`)
```

I suppose it's also safe to assume that the namespace names used to generate the suffix of the username won't contain hyphens?

soltysh: I wouldn't bet on the namespace not containing hyphens, they usually do contain, since the generated parts are usually have a hyphen between the prefix and the generated part. To some we also explicitly add `k8s-` so it's probably better not to take that far going assumption. That's why I didn't make this regexp too strict.

test/extended/apiserver/api_requests.go
```diff
@@ -73,9 +81,21 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 					}
 				}
 			}
+
+			// in case we didn't have the last 24h detailed data, at least log as a general API being accessed
+			if len(apiRequestCount.Status.Last24h) == 0 {
+				resourceToRequestCount := userToResourceToRequestCount[unknownUser]
+				if resourceToRequestCount == nil {
+					resourceToRequestCount = map[string]requestCount{}
```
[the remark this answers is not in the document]
soltysh: Good catch!

soltysh:
/hold
I think that auth test needs a second look

PR #27562, opened 2022-11-21
TRT-691: use duplicateEventThreshold for ProbeTests

neisw:
Consolidate to single threshold value


DennisPeriquet:
/hold
@dgoodwin is this right? I think we want to extract the count out of the message that says something like 'happenex x times'.  I don't think the `count` variable does that.

DennisPeriquet:
here's an example of extracting the count: https://github.com/openshift/origin//blob/master/pkg/synthetictests/duplicated_events_special.go#L54

pkg/synthetictests/kubelet.go
```diff
@@ -177,40 +173,43 @@ func testContainerFailures(events monitorapi.Intervals) []*junitapi.JUnitTestCas
...
-			count++
-		}
-	}
-	test := &junitapi.JUnitTestCase{Name: testName}
-	if count > eventFlakeThreshold {
-		// Flake for now.
-		test.FailureOutput = &junitapi.FailureOutput{
-			Output: failureOutput,
+
+			_, times := getTimesAnEventHappened(fmt.Sprintf("%s - %s", event.Locator, event.Message))
```
neisw, line 199: Was the original issue that we weren't extracting the times & the threshold was too low or just that we weren't extracting the times?  I can restore the previous thresholds if we need.  Also presume that each event / times count is compared in isolation and not accumulated then compared.  Similar to what is in duplicated_events

dgoodwin: Original issue was that this was happening all the time and we wanted to break out a specific test for the problem that we could flake, but track separately. It was working fine just failing too much.

DennisPeriquet: yes, we were failing too much because I didn't extract the count as you're doing above.  Here's the original logic: https://github.com/openshift/origin//blob/b25083353f1495ccb98b11aa78af8442dc559c66/pkg/synthetictests/duplicated_events.go#L341-L343. You can see that the threshold is 20 and not 5.

neisw: cool, so I will keep the duplicateEventThreshold update as well as this code to extract the times/count to compare against that threshold.

dgoodwin: Did the original test do anything with "number of times this line occurred"? Or was it just "how many repetitions did we see in each line". 

I'd suggest we stick precisely to whatever logic that had before. It's possible we're only interested in finding events that repeated so close to each other they got compressed by kube together, over 20x.

neisw: For the original logic it looks like the last group of events over duplicateEventThreshold wins: https://github.com/openshift/origin/blob/0bd61089e01d2c5fcbe03546c69be842fcfd6589/pkg/synthetictests/duplicated_events.go#L336-L365

eventDisplayMessage appears to be generic so the assignment

displayToCount[eventDisplayMessage] = times

ends up being last one wins. 

I thought it might be bugged but with the check for duplicateEventThreshold  if we have a case where it repeated 25 times and then later 5 times we still report the 25 case.  If we have 25 then 21 we report 21 but it will still trigger the flake so good enough.


The way it is rewritten I believe the first occurrence over duplicateEventThreshold will trigger the flake so in the case of multiple groupings {25,21} it could be slightly different message {25) but same result of a flake.

neisw:
/hold

Need to review the nodeHasNoDiskPressureEventThreshold, etc code path as well

neisw:
Realized there was another set of these tests under pkg/synthetictests/node.go with 99% duplicate logic.  Refactored to consolidate but left as separate commit in case we don't want it.

pkg/synthetictests/duplicated_events_special.go
```diff
@@ -144,3 +151,82 @@ func testErrorUpdatingEndpointSlices(events monitorapi.Intervals) []*junitapi.JU
...
+func makeNodeHasTest(testName string, events monitorapi.Intervals, regExStr string, eventFlakeThreshold int) []*junitapi.JUnitTestCase {
+	messageRegExp := regexp.MustCompile(regExStr)
+	return eventMatchThresholdTest(testName, events, func(event monitorapi.EventInterval) bool { return messageRegExp.MatchString(event.Message) }, eventFlakeThreshold)
+}
+
+func eventMatchThresholdTest(testName string, events monitorapi.Intervals, eventMatch eventRecognizerFunc, eventFlakeThreshold int) []*junitapi.JUnitTestCase {
+	
+	// attempt to force failures, should not merge!
+	eventFlakeThreshold = 1
+
```
neisw, line 201: Will remove after tests run.  Added just to try to force flakes in this area

neisw:
[Flake for reason/ProbeError Liveness probe error](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27562/pull-ci-openshift-origin-master-e2e-gcp-ovn/1597315400092094464)

neisw:
[Additional Flakes for NodeHas*](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27562/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1597315399530057728)

PR #27567, opened 2022-11-23
tests: add synthetic test that checks if test namespaces were cleaned up

mfojtik:
In https://github.com/openshift/kubernetes/pull/1383 we discovered at least one namespace leaking after the e2e tests were done. That cause the test to timeout, and debugging the timeout requires re-running the test, watching the cluster where the e2e test runs, and list namespaces AFTER the test finish.
This synthetic test aims to automate that. 
Maybe this is the wrong spot; maybe we need some "post-run" hook we can place tests like this? Maybe we need something that waits until the e2e test finishes... I am open for suggestions :-)


dgoodwin:
Could you clarify where the 2h timeout was? I'm unclear what would get blocked for 2 hours due to a namespace hanging around.

PR #27571, opened 2022-11-25
enable route creation test for microshift

atiratree:
required for https://github.com/openshift/release/pull/34392


test/extended/router/router.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/openshift/origin/test/extended/util/url"
 )
 
-var _ = g.Describe("[sig-network][Feature:Router][apigroup:operator.openshift.io][apigroup:apps.openshift.io]", func() {
```
ingvagabund, line 21: Is `WaitForRouterServiceIP` the only consumer of both api groups?

test/extended/router/router.go
```diff
@@ -50,9 +50,17 @@ var _ = g.Describe("[sig-network][Feature:Router][apigroup:operator.openshift.io
...
-		var err error
-		host, err = exutil.WaitForRouterServiceIP(oc)
+		exists, err := exutil.DoesApiResourceExist(oc.AdminConfig(), "ingresscontrollers")
 		o.Expect(err).NotTo(o.HaveOccurred())
+		if exists {
+			host, err = exutil.WaitForRouterServiceIP(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			panic(err)
+		} else {
+			// microshift
```
ingvagabund, line 60: Wherever possible the condition should be free of any MicroShift awareness.

test/extended/router/router.go
```diff
@@ -50,9 +50,17 @@ var _ = g.Describe("[sig-network][Feature:Router][apigroup:operator.openshift.io
 	oc = exutil.NewCLIWithPodSecurityLevel("router-stress", admissionapi.LevelBaseline)
 
 	g.BeforeEach(func() {
-		var err error
-		host, err = exutil.WaitForRouterServiceIP(oc)
+		exists, err := exutil.DoesApiResourceExist(oc.AdminConfig(), "ingresscontrollers")
```
ingvagabund, line 53: Does absence of `ingresscontrollers` apigroup implies existence of `router-internal-default` service?

atiratree: it does not, I guess it depends on the environment. In microshift they use this service. I suppose if it gets changed or some other env has a different name they would have to go here and customize it further

test/extended/router/router.go
atiratree: ack, will remove as per https://github.com/openshift/origin/pull/27571/files#r1034996931

test/extended/router/router.go
atiratree: you are right, I missed this one https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L1919 . I will have to add the conditional check also to there

test/extended/util/framework.go
```diff
@@ -2054,6 +2056,25 @@ func SkipIfExternalControlplaneTopology(oc *CLI, reason string) {
 	}
 }
 
+type DoesApiResourceExistFn func(apiResourceName string) bool
```
atiratree, line 2059: I have introduce this function to limit the number of resource discovery calls and support multiple invocations in a single test

test/extended/util/router.go
```diff
@@ -12,12 +12,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-func WaitForRouterInternalIP(oc *CLI) (string, error) {
-	return waitForNamedRouterServiceIP(oc, "router-internal-default")
+func WaitForRouterInternalIP(oc *CLI, doesApiResourceExistFn DoesApiResourceExistFn) (string, error) {
```
atiratree, line 15: disadvantage is that we have to pass the function around

test/extended/operators/routable.go
```diff
@@ -33,7 +33,9 @@ var _ = g.Describe("[sig-arch] Managed cluster should [apigroup:apps.openshift.i
 	)
 
 	g.BeforeEach(func() {
-		_, ns, err := exutil.GetRouterPodTemplate(oc)
+		apiResourceExistsFn, err := exutil.DoesApiResourceExistCached(oc.AdminConfig())
+		o.Expect(err).NotTo(o.HaveOccurred())
+		_, ns, err := exutil.GetRouterPodTemplate(oc, apiResourceExistsFn)
```
atiratree, line 38: in case we are not okay doing so many changes to other tests than `should serve routes that were created from an ingress`, we can always pass a `nil` in tests like these instead ,for the old behaviour of always checking `DeploymentConfigs`

test/extended/router/router.go
```diff
@@ -50,9 +50,15 @@ var _ = g.Describe("[sig-network][Feature:Router][apigroup:operator.openshift.io
 	oc = exutil.NewCLIWithPodSecurityLevel("router-stress", admissionapi.LevelBaseline)
 
 	g.BeforeEach(func() {
-		var err error
-		host, err = exutil.WaitForRouterServiceIP(oc)
+		apiResourceExistsFn, err := exutil.DoesApiResourceExistCached(oc.AdminConfig())
 		o.Expect(err).NotTo(o.HaveOccurred())
+		if apiResourceExistsFn("ingresscontrollers") {
```
atiratree, line 55: I suppose we could now move this conditional logic inside the `WaitForRouterServiceIP`, as I am not seeing any bad implications for the other tests.

Do we want to move it there or limit the changes/scope of this PR?

test/extended/scheduling/pods.go
```diff
@@ -112,7 +112,9 @@ var _ = g.Describe("[sig-scheduling][Early]", func() {
 	oc := exutil.NewCLI("scheduling-pod-check")
 
 	g.BeforeEach(func() {
```
atiratree, line 114: @ingvagabund do you recall why this test does not have `apigroup:apps.openshift.io` apigroup? And why the `The openshift-apiserver pods` test has bunch of apigroups as well as this one?

test/extended/router/router.go
atiratree: added

test/extended/util/framework.go
```diff
@@ -1912,16 +1912,18 @@ func WaitForUserBeAuthorized(oc *CLI, user string, attributes *authorizationapi.
...
 	k8sappsclient := oc.AdminKubeClient().AppsV1()
 	for _, ns := range []string{"default", "openshift-ingress", "tectonic-ingress"} {
-		dc, err := appsclient.DeploymentConfigs(ns).Get(context.Background(), "router", metav1.GetOptions{})
-		if err == nil {
-			return dc.Spec.Template, ns, nil
-		}
-		if !kapierrs.IsNotFound(err) {
-			return nil, "", err
+		if doesApiResourceExistFn == nil || doesApiResourceExistFn("deploymentconfigs") {
+			dc, err := appsclient.DeploymentConfigs(ns).Get(context.Background(), "router", metav1.GetOptions{})
```
ingvagabund, line 1920: I wonder whether any of the listed namespace is still using DC.

atiratree:
@ingvagabund I have checked and it seems router is not deployed via DeploymentConfigs anymore, so it probably should be harmless to remove the DCs from router tests and drop the `apigroup:apps.openshift.io` dependency. We are using a Deployment in the [ingress-operator](https://github.com/openshift/cluster-ingress-operator/blob/master/assets/router/deployment.yaml)

test/extended/util/framework.go
atiratree: removed and the PR is a lot of simpler now

test/extended/router/router.go
ingvagabund: @Miciah is it safe to fallback to the internal router service when ingresscontrollers CR is not available? I.e. is existence of ingresscontrollers cr always guaranteed when a cluster has other than internal router service?

ingvagabund: @dhellmann would you please help here? More elaborate on how the `router-internal-default` router service is managed? Non-existence of the `ingresscontrollers` CR is a weak condition for existence of `router-internal-default` without additional knowledge about how the service is managed. Is https://github.com/openshift/microshift/blob/main/assets/components/openshift-router/service-internal.yaml service manifest managed?

atiratree: from what I can tell the service is just created by microshift when it starts here https://github.com/openshift/microshift/blob/18491380d2e359465a47c6f23194bc1a4f1f9271/pkg/components/controllers.go#L149

PR #27569, opened 2022-11-24
Allow baremetal tests to run on AWS platform

elfosardo:
Metal3 pods can now be deployed on AWS platform.
Baremetal hosts can be booted only via virtual media. We only check for the presence of the metal3 pod to verify the baremetal deployment is available in AWS too.

PR #27573, opened 2022-11-28
Allow baremetal tests to run on Azure platform

elfosardo:
Metal3 pods can now be deployed on Azure platform. Baremetal hosts can be booted only via virtual media. We only check for the presence of the metal3 pod to verify the baremetal deployment is available in Azure too.


elfosardo:
hold until https://github.com/openshift/cluster-baremetal-operator/pull/309 merges

elfosardo:
@deads2k hi! Please have a look at this when you get a chance, thanks!

elfosardo:
this is similar to https://github.com/openshift/origin/pull/27569
 tests look ok, failures do not depend from this change

PR #27575, opened 2022-11-28
If there are no test-failure-summary files, suppress risk analysis

DennisPeriquet:
[TRT-697](https://issues.redhat.com//browse/TRT-697)

The glob operation is still valid when there are 0 matching files.  But if we didn't find any files, we can't run risk analysis.

### DIFF
--- a/pkg/riskanalysis/cmd.go
+++ b/pkg/riskanalysis/cmd.go
@@ -40,6 +40,9 @@ func (opt *Options) Run() error {
 	if err != nil {
 		return err
 	}
+	if len(resultFiles) == 0 {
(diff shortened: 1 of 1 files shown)


neisw:
I'm wondering if in considering [TRT-662](https://issues.redhat.com/browse/TRT-662) we should have sippy elevate the risk level when an invalid prowjob is received. 

Something like:
`{"OverallRisk":{"Level":{"Name":"High","Level":10},"Reasons":["Invalid ProwJob Summary"]}}`

If so then we would allow the missing test summary case to pass through as it does today

DennisPeriquet:
As part of TRT-662, we may handle an empty list in sippy.  When this is done, we may not need this PR.

DennisPeriquet:
This PR https://github.com/openshift/sippy/pull/708 handles the case at the sippy level so this PR is no longer necessary.

PR #27576, opened 2022-11-29
update vendor to pick up bug fixes

deads2k:
notably the vsphere one.

/assign @patrickdillon


deads2k:
most passed, applying label

/retest-required

deads2k:
template tests failed. if it fails that way again, something to look at.

PR #27578, opened 2022-11-30
USHIFT-646: ushift: Graceful return to disable telemetry

pacevedom:
MicroShift does not have a prometheus instance so we need to disable any tests that require it. In order to skip some of them we need a graceful return from telemetryIsEnabled function. This function calls hasPullSecret, which checks for specific pull secrets in the openshift-config/pull-secret configmap, non-existent in MicroShift. Function already returns a graceful error if there is no pull secret for a specific domain, so this commit mimicks the behavior to a situation where there is no configmap at all.


simonpasquier (commented):
/lgtm

PR #27583, opened 2022-12-01
Remove ambiguity when checking for api resources existence

pacevedom:
When using DoesApiResourceExist function from framework we might get false positives if there are two resources with the same name but difference API group. This change checks the API group (not version) with the resource name too in order to remove ambiguity.

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -1942,7 +1942,7 @@ func GetRouterPodTemplate(oc *CLI) (*corev1.PodTemplateSpec, string, error) {
 }
 
 func FindRouterImage(oc *CLI) (string, error) {
-	exists, err := DoesApiResourceExist(oc.AdminConfig(), "clusteroperators")
+	exists, err := DoesApiResourceExist(oc.AdminConfig(), "clusteroperators", "config.openshift.io")
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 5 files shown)


test/extended/util/framework.go
```diff
@@ -2066,6 +2066,11 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName string) (bool, er
 		return false, err
 	}
 	for _, apiResourceList := range allResourceList {
+		gvns := strings.Split(groupVersionName, "/")
+		gvs := strings.Split(apiResourceList.GroupVersion, "/")
```
ingvagabund, line 2070: Will the same work if you us `apiResource.Group` instead in the loop bellow? E.g.
```go
if apiResource.Name == apiResourceName && apiResource.Group == groupVersionName {
	return true, nil
}
```

ingvagabund, line 2070: Worth turning `strings.Split(groupVersionName, "/")` into a function and invoking it when comparing. E.g.
```go
func groupName(string) {
	return strings.Split(groupVersionName, "/")[0]
}
...
if groupName(groupVersionName) != groupVersionName(apiResourceList.GroupVersion) {
	continue
}
```
The function can be later extended with checks. Also, gvns probably stands for "group version name" with s as plural. Though, after splitting the gvn string the version is no longer present.

pacevedom: Done

PR #27585, opened 2022-12-01
TRT-662: include test count totals for analysis

neisw:
Include the total test count so we can compare against previous runs and detect missing tests in cases like failed installs / upgrades, etc.

On the Sippy side adding:

```
type ProwJob struct {
	gorm.Model

	Kind        ProwKind
	Name        string         `gorm:"unique"`
	Release     string         `gorm:"varchar(10)"`
	Variants    pq.StringArray `gorm:"index;type:text[]"`
	TestGridURL string
	Bugs        []Bug        `gorm:"many2many:bug_jobs;"`
	JobRuns     []ProwJobRun `gorm:"constraint:OnDelete:CASCADE;"`
	TestCount   int			 `gorm:"-"`
}
```

So we can pass the value via struct but not store in db.  Can make that a request parameter if desired.


dgoodwin:
/lgtm
/hold

This seems fine to me, holding so we can get sippy in place and actually test it out a little.

PR #27586, opened 2022-12-01
feat: support higher threshold for connection refused on sno

eggfoobar:
Added support to increase threshold for duplicate connection refused errors on single node clusters


pkg/synthetictests/duplicated_events_special.go
```diff
@@ -29,6 +29,7 @@ const (
 	nodeHasNoDiskPressureRegExpStr             = "reason/NodeHasNoDiskPressure.*status is now: NodeHasNoDiskPressure"
 	nodeHasSufficientMemoryRegExpStr           = "reason/NodeHasSufficientMemory.*status is now: NodeHasSufficientMemory"
 	nodeHasSufficientPIDRegExpStr              = "reason/NodeHasSufficientPID.*status is now: NodeHasSufficientPID"
+	singleNodeErrorConnectionRefusedRegExpStr  = "reason/.*dial tcp.*connection refused"
```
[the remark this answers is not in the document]
eggfoobar: Sure, so the common one I've seen this effect is the one below, this should only be used on connection refused for single node, if we want to limit who uses this regex, I can move it to the call site.
```
event happened 25 times, something is wrong: ns/openshift-etcd-operator deployment/etcd-operator - reason/ScriptControllerErrorUpdatingStatus Put "https://172.30.0.1:443/apis/operator.openshift.io/v1/etcds/cluster/status": dial tcp 172.30.0.1:443: connect: connection refused

event happened 29 times, something is wrong: ns/openshift-oauth-apiserver pod/apiserver-6ffcbb4974-kl8kc node/ip-10-0-187-246.us-east-2.compute.internal - reason/ProbeError Readiness probe error: Get "https://10.128.0.71:8443/readyz": dial tcp 10.128.0.71:8443: connect: connection refused
```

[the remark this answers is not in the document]
DennisPeriquet: Ah, I see -- you want it to be broad.  I think it's fine as-is since your function `isConnectionRefusedOnSingleNode` is pretty descriptive.

pkg/synthetictests/duplicated_events.go
```diff
@@ -611,6 +638,12 @@ func isOauthApiserverProbeErrorConnectionRefusedFailed(monitorEvent monitorapi.E
 	return isOperatorMatchRegexMessage(monitorEvent, "openshift-oauth-apiserver", regExp), nil
 }
 
+// isConnectionRefusedOnSingleNode returns true if the event matched has a connection refused message for single node events and is with in threshold.
+func isConnectionRefusedOnSingleNode(monitorEvent monitorapi.EventInterval, kubeClientConfig *rest.Config, count int) (bool, error) {
+	regExp := regexp.MustCompile(singleNodeErrorConnectionRefusedRegExpStr)
+	return isSingleNodeCluster(kubeClientConfig) && regExp.MatchString(monitorEvent.String()) && count < duplicateSingleNodeEventThreshold, nil
```
dgoodwin, line 644: This looks like we might be querying the Infrastructure object for every event interval, probably tens of thousands of times. Is there some precedent for a global check here in the code already for SNO?

eggfoobar: That's a good call out, I had assumed it was okay since all those allowed functions take kubeConfig as an arg, but it seems we already collect the topology info anyway, updated the code to make use of it. What do you think of the change?

dgoodwin: Looks good, sounds like trouble if anyone is actually using that kubeconfig elsewhere.

PR #27600, opened 2022-12-08
TRT-703: Fix upgrade junit results not present in risk analysis

dgoodwin:
While investigating why some tests were failing in spyglass, but not present in risk analysis, I discovered this second location where we output junit files (most notably the upgrade junit files) using test suites, which was not updated to also output the test failure summary we use for risk analysis.

Digging further I saw that this path of code was using a totally different representation of testsuite/testcase than everything else in origin (k8s TestSuite vs our ginko JUnitTestSuite). It would have been easier just to write another function to pull the test failures out of this, but I opted to get them both using the same objects as there were other subtle differences in the XML that made me nervous.

To validate I've compared junit_upgrade output from the aws-ovn-upgrade job with an older example: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27568/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1595382748753694720/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/junit_upgrade_1669214650.xml
Then ensured we have the same test counts, and the XML looks correct and consistent in all the ways we would expect.

[TRT-703](https://issues.redhat.com//browse/TRT-703)

### DIFF
--- a/test/extended/util/disruption/disruption.go
+++ b/test/extended/util/disruption/disruption.go
@@ -149,14 +149,14 @@ func runChaosmonkey(
 			fname := filepath.Join(framework.TestContext.ReportDir, fmt.Sprintf("junit_%s_%s.xml", packageName, timeSuffix))
 			f, err := os.Create(fname)
 			if err != nil {
-				// TODO: no logging?
+				fmt.Fprintf(os.Stderr, "error: Failed to write file %v: %v\n", fname, err)
 				return
 			}
 			defer f.Close()
 			xml.NewEncoder(f).Encode(testSuite)
 
 			if err := riskanalysis.WriteJobRunTestFailureSummary(framework.TestContext.ReportDir, timeSuffix, testSuite); err != nil {
-				// TODO: no logging?
+				fmt.Fprintf(os.Stderr, "error: Failed to write file %v: %v\n", fname, err)
 				return
 			}
 		}
--- a/test/extended/util/disruption/disruption.go
+++ b/test/extended/util/disruption/disruption.go
@@ -1,5 +1,8 @@
 package disruption
 
+// TODO: this testing framework is used by many upgrade tests beyond disruption, the package is somewhat misleading
+// and it should probably be relocated.
+
 import (
 	"encoding/json"
 	"encoding/xml"
@@ -14,6 +17,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/openshift/origin/pkg/riskanalysis"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
 	admissionapi "k8s.io/pod-security-admission/api"
 
 	g "github.com/onsi/ginkgo/v2"
@@ -25,7 +30,6 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework/ginkgowrapper"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	"k8s.io/kubernetes/test/e2e/upgrades"
-	"k8s.io/kubernetes/test/utils/junit"
 )
 
 const (
@@ -84,23 +88,22 @@ type TestData struct {
 // test is being executed. Description is used to populate the JUnit suite name, and testname is
 // used to define the overall test that will be run.
 func Run(f *framework.Framework, description, testname string, adapter TestData, invariants []upgrades.Test, fn func()) {
-	testSuite := &junit.TestSuite{Name: description, Package: testname}
-	test := &junit.TestCase{Name: testname, Classname: testname}
-	testSuite.TestCases = append(testSuite.TestCases, test)
+	testSuite := &junitapi.JUnitTestSuite{Name: description}
 	cm := chaosmonkey.New(func() {
 		start := time.Now()
-		defer finalizeTest(start, test, testSuite, f)
+		defer finalizeTest(start, testname, testname, testSuite, f)
 		defer g.GinkgoRecover()
 		fn()
 	})
-	runChaosmonkey(cm, adapter, invariants, testSuite)
+	runChaosmonkey(cm, adapter, invariants, testSuite, testname)
 }
 
 func runChaosmonkey(
 	cm *chaosmonkey.Chaosmonkey,
 	testData TestData,
 	tests []upgrades.Test,
-	testSuite *junit.TestSuite,
+	testSuite *junitapi.JUnitTestSuite,
+	packageName string,
 ) {
 	testFrameworks := createTestFrameworks(tests)
 	for _, t := range tests {
@@ -108,11 +111,6 @@ func runChaosmonkey(
 		if dn, ok := t.(testWithDisplayName); ok {
 			displayName = dn.DisplayName()
 		}
-		testCase := &junit.TestCase{
-			Name:      displayName,
-			Classname: "disruption_tests",
-		}
-		testSuite.TestCases = append(testSuite.TestCases, testCase)
 
 		f, ok := testFrameworks[t.Name()]
 		if !ok {
@@ -122,7 +120,8 @@ func runChaosmonkey(
 			TestData:        testData,
 			framework:       f,
 			test:            t,
-			testReport:      testCase,
+			testName:        displayName,
+			className:       "disruption_tests",
 			testSuiteReport: testSuite,
 		}
 		cm.Register(cma.Test)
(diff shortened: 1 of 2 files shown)


dgoodwin:
Seeing what happens...

test/extended/util/disruption/disruption.go
```diff
@@ -84,35 +85,29 @@ type TestData struct {
 // test is being executed. Description is used to populate the JUnit suite name, and testname is
 // used to define the overall test that will be run.
 func Run(f *framework.Framework, description, testname string, adapter TestData, invariants []upgrades.Test, fn func()) {
-	testSuite := &junit.TestSuite{Name: description, Package: testname}
-	test := &junit.TestCase{Name: testname, Classname: testname}
-	testSuite.TestCases = append(testSuite.TestCases, test)
+	testSuite := &junitapi.JUnitTestSuite{Name: description}
```
dgoodwin, line 88: Package is not present under our normal test suite XML, so I dropped it. I did keep using it for the filename however so that stays consistent in case anyone is looking for it by a specific name. Note description="Cluster upgrade", and package="upgrade" here.

test/extended/util/disruption/disruption.go
```diff
@@ -84,35 +85,29 @@ type TestData struct {
...
 	for _, t := range tests {
 		displayName := t.Name()
 		if dn, ok := t.(testWithDisplayName); ok {
 			displayName = dn.DisplayName()
 		}
-		testCase := &junit.TestCase{
-			Name:      displayName,
-			Classname: "disruption_tests",
-		}
-		testSuite.TestCases = append(testSuite.TestCases, testCase)
```
dgoodwin, line 115: Before we would implicitly create test cases at the start, and then try to figure out later if we should inject or update them for flakes/failures. I found this confusing and inconsistent with how we do things everywhere I've seen so far, so now the test cases are added *after* we've run the test and know what's going on. In practice you'll now see the pass/fail TestCases for a flake together in the xml, instead of one being appended to the end.

test/extended/util/disruption/disruption.go
```diff
@@ -122,45 +117,68 @@ func runChaosmonkey(
...
-		for _, testCase := range testSuite.TestCases {
-			allFlakes := len(testCase.Failures) > 0 && len(testCase.Errors) == 0 && len(testCase.Skipped) == 0
-			for _, failure := range testCase.Failures {
-				if failure.Type == "Flake" {
-					failure.Type = "Failure"
-				} else {
-					allFlakes = false
-				}
+
+		// Calculate NumFailed and NumSkipped
```
dgoodwin, line 130: These populate the counts in the top level <testsuite>, has to be tracked manually and I found calculating here safer than trying to increment everywhere we add a test case.

test/extended/util/disruption/disruption.go
```diff
@@ -122,45 +117,68 @@ func runChaosmonkey(
...
-					Time:      testCase.Time,
-				})
+			if tc.SkipMessage != nil {
+				testSuite.NumSkipped++
 			}
 		}
 
+		testSuite.Duration = time.Since(start).Seconds()
+
+		// I cannot find *anything* adding TestCases with type Flake, only TestSummaries.
```
dgoodwin, line 143: This is delicate, but I do not believe anything was adding k8s TestCase with type=Flake, not to be confused with TestSummaries of type flakeSummary, which was used in one place over in the alert test calling FrameworkFlakef.

test/extended/util/disruption/disruption.go
```diff
@@ -197,71 +221,76 @@ func (cma *chaosMonkeyAdapter) Test(sem *chaosmonkey.Semaphore) {
...
 		}
 	}
 
 	if r == nil {
+		// Test can be considered successful, but may have flaked, see below:
+		ts.TestCases = append(ts.TestCases, &junitapi.JUnitTestCase{
+			Name:      testName,
+			Classname: className,
+			Duration:  testDuration,
+		})
```
dgoodwin, line 260: Because we no longer pre-create TestCase objects before the test starts, this is the place where successful tests are added.

test/extended/util/disruption/disruption.go
```diff
@@ -197,71 +221,76 @@ func (cma *chaosMonkeyAdapter) Test(sem *chaosmonkey.Semaphore) {
...
-					Type:    "Flake",
-					Message: message,
-					Value:   message,
+			if message, hasFlake := hasFrameworkFlake(f); hasFlake {
+				// Add another test case with the failure, this is a flake as we already added the success testcase above:
+				ts.TestCases = append(ts.TestCases, &junitapi.JUnitTestCase{
+					Name:          testName,
+					Classname:     className,
+					Duration:      testDuration,
+					FailureOutput: &junitapi.FailureOutput{Output: message},
```
dgoodwin, line 268: And here is where flakes are added, afaik this is only used by the alert.go upgrade check for firing/pending alerts. It examines TestSummaries for those of type flakeSummary.

test/extended/util/disruption/disruption.go
```diff
@@ -197,71 +201,79 @@ func (cma *chaosMonkeyAdapter) Test(sem *chaosmonkey.Semaphore) {
...
+		testResult.SkipMessage = &junitapi.SkipMessage{Message: fmt.Sprintf("%s:%d %q", r.Filename, r.Line, r.Message)}
 	default:
-		tc.Errors = []*junit.Error{
-			{
-				Message: "Ginkgo panic encountered. See CDATA for details.",
-				Type:    "Panic",
-				Value:   fmt.Sprintf("%v\n\n%s", r, debug.Stack()),
-			},
-		}
+		testResult.FailureOutput = &junitapi.FailureOutput{Message: fmt.Sprintf("%v\n\n%s", r, debug.Stack())}
```
dgoodwin, line 270: This ran into @xueqzhan 's recent fix for the bad XML. I'm hoping this is ok still, my understanding we needed to keep the bad characters out of the message="" XML param. Our normal test cases do not use a message param, so this is gone entirely, and the full failure now resides within <failure/> tag.

test/extended/util/disruption/disruption.go
```diff
@@ -122,45 +120,45 @@ func runChaosmonkey(
...
+		testSuite.Duration = time.Since(start).Seconds()
+
 		if framework.TestContext.ReportDir != "" {
-			fname := filepath.Join(framework.TestContext.ReportDir, fmt.Sprintf("junit_%s_%d.xml", testSuite.Package, time.Now().Unix()))
+			timeSuffix := fmt.Sprintf("_%s", time.Now().UTC().Format("20060102-150405"))
+
+			fname := filepath.Join(framework.TestContext.ReportDir, fmt.Sprintf("junit_%s_%s.xml", packageName, timeSuffix))
 			f, err := os.Create(fname)
 			if err != nil {
+				// TODO: no logging?
```
neisw, line 152: Would a log entry here help in the case we had a failure and wanted to look into it?  Don't know how useful it would be so not a hard ask.

dgoodwin:
Still haven't gotten results from a new run with latest code it looks like, so I'll wait till that arrives.

As for the logging I just wasn't sure what mechanism to use, there are several in play in origin, no logging I could see anywhere in this area, I'll try to figure out what's most consistent and update those as well. Thanks for taking a look.

dgoodwin:
Commit pushed to log the errors, I found an example below my search missed last week.

dgoodwin:
/hold 

Ok to lgtm Forrest, I won't release hold until we get a look at the xml for certain.

dgoodwin:
Output looks good:

Old file: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27568/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1595382748753694720/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/junit_upgrade_1669214650.xml

New file: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27600/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1602362183100928000/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/junit_upgrade__20221212-211150.xml

The differences look as I would expect and should be more consistent with the junit e2e xml file we write out elsewhere.

/hold cancel

PR #27602, opened 2022-12-09
Test failures that flaked on retry remain counted as failures

DennisPeriquet:
In this [job](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1403/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn/1598319526313725952), the `[sig-arch] Managed cluster should ensure platform components have system-* priority class associated [Suite:openshift/conformance/parallel]` test failed, then on retry, it flaked.  In the test results, it was counted as a flake.

With this change, the test will be counted as a failure.


pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -417,6 +416,12 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
 				repeatFailures = append(repeatFailures, test)
 			}
 		}
+		for _, retry := range retries {
+			// Retry tests that flaked are omitted so that the original test is counted as a failure.
+			if !retry.flake {
+				tests = append(tests, retry)
+			}
+		}
```
dgoodwin, line 424: We loop over the retries right above, so you can move this if up into the loop above.

Otherwise this looks correct. I am a little worried if there's any unintended side effect from not being in "tests" before we run the Execute, whereas we were before. I don't see tests getting used until way later when writing XML, so to me it looks like your patch should work. We'll just want to carefully check results here, get into the logs, and make sure everything is ok. I'd be tempted to add a hack here that simulates a retry flaking (randomly, maybe 50% of the time or something), then we can check and see the outputs match what we'd expect.

Also make sure to add a log statement., something like: fmt.Fprintf(opt.Out, "Ignoring retry that returned a flake, original failure should be authoritative for test: %s", testName)

DennisPeriquet:
Here's is an example [job log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27602/pull-ci-openshift-origin-master-e2e-gcp-ovn/1602685853782511616/artifacts/e2e-gcp-ovn/openshift-e2e-test/build-log.txt) where we show `Retry count: 8` but each retry passed so we didn't have a chance to flake any retries.

For retries, I'm going to make passing retries flake on every other pass; this should allow us to see flaking retries.  The changes are a little more invasive (adding a parameter to the function that executes the tests) but the changes are easy to understand and back out because function calls with wrong number of parameters is a compile-time check.

DennisPeriquet:
Example [ci job](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27602/pull-ci-openshift-origin-master-e2e-aws-ovn-single-node-upgrade/1602732945930457088) and [job log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27602/pull-ci-openshift-origin-master-e2e-aws-ovn-single-node-upgrade/1602732945930457088/build-log.txt) where we see: `Ignoring retry that returned a flake, original failure should be authoritative for test:`

PR #27604, opened 2022-12-12
add v1alpha1 version for config.openshift.io

tremes:
This should hopefully fix https://issues.redhat.com/browse/OCPBUGS-4559


dgoodwin:
Could you clarify why this is only surfacing on techpreview + sdn?

deads2k:
/hold

this area is under refactoring https://github.com/openshift/origin/pull/27596

tremes:
@deads2k I see. So the original error message (event) from the [OCPBUGS-4559](https://issues.redhat.com/browse/OCPBUGS-4559) will disappear.  However, it's still not clear to me how this works with techpreview clusters (is a new TP resource not being watched at all?). Can I somehow run the resourcewatch locally (when running tests)?

wking:
#27625 is a robot-pick of #27596 back to 4.12.

tremes:
Fixed with https://github.com/openshift/origin/pull/27596

PR #27608, opened 2022-12-13
OCPBUGS-4731: oc status: clean job resource to prevent leaks

ardaguclu:
Currently, `oc status` test creates job resource and does not clean it after test is completed.

This PR adds clean up function to prevent any resource leakages.

### DIFF
--- a/test/extended/cli/status.go
+++ b/test/extended/cli/status.go
@@ -83,15 +83,15 @@ var _ = g.Describe("[sig-cli] oc status", func() {
 
 		err = oc.WithoutNamespace().Run("new-project").Args(projectStatus, "--display-name=my project", "--description=test project").Execute()
 		o.Expect(err).NotTo(o.HaveOccurred())
-
-		g.By("verify jobs are showing in status")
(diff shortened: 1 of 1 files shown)


test/extended/cli/status.go
```diff
@@ -85,9 +85,11 @@ var _ = g.Describe("[sig-cli] oc status", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		g.By("verify jobs are showing in status")
-		err = oc.WithoutNamespace().Run("create").Args("job", "pi", "--image=image-registry.openshift-image-registry.svc:5000/openshift/tools:latest",
-			"--", "perl", "-Mbignum=bpi", "-wle", "'print bpi(2000)'").Execute()
+		err = oc.WithoutNamespace().Run("create").Args("job", "pi", "--image=image-registry.openshift-image-registry.svc:5000/openshift/tools:latest").Execute()
```
soltysh, line 88: As long as `WithoutNamespace` makes sense in the above calls, here when you're creating namespaced resources I'd suggest creating jobs also in a particular namespace, and also checking status in that same namespace. Also, you should be cleaning all the projects being created in this test, which will handle all the resources inside them as well.

soltysh (approved):
/lgtm
/approve

PR #27613, opened 2022-12-15
OpenStack: Skip in-tree cinder provisioner tests

mandre:
In preparation for kube 1.26 rebase.

### DIFF
--- a/test/extended/util/annotate/rules.go
+++ b/test/extended/util/annotate/rules.go
@@ -88,6 +88,10 @@ var (
 			`\[sig-storage\] In-tree Volumes \[Driver: vsphere\] \[Testpattern: Dynamic PV \(delayed binding\)\] topology should provision a volume and schedule a pod with AllowedTopologies`,
 			`\[sig-storage\] In-tree Volumes \[Driver: vsphere\] \[Testpattern: Dynamic PV \(immediate binding\)\] topology should fail to schedule a pod which has topologies that conflict with AllowedTopologies`,
 			`\[sig-storage\] In-tree Volumes \[Driver: vsphere\] \[Testpattern: Dynamic PV \(immediate binding\)\] topology should provision a volume and schedule a pod with AllowedTopologies`,
+			// Skip openstack-specific storage tests in preparation for in-tree cinder provisioner removal
+			// coming with k8s 1.26. This will have to be reverted once 1.26 rebase is effective.
+			// https://issues.redhat.com/browse/OCPBUGS-5029
+			`\[sig-storage\].*\[Driver: cinder\]`,
 		},
 		// tests too slow to be part of conformance
 		"[Slow]": {},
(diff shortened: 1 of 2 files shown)


test/extended/util/annotate/rules.go
```diff
@@ -88,6 +88,9 @@ var (
 			`\[sig-storage\] In-tree Volumes \[Driver: vsphere\] \[Testpattern: Dynamic PV \(delayed binding\)\] topology should provision a volume and schedule a pod with AllowedTopologies`,
 			`\[sig-storage\] In-tree Volumes \[Driver: vsphere\] \[Testpattern: Dynamic PV \(immediate binding\)\] topology should fail to schedule a pod which has topologies that conflict with AllowedTopologies`,
 			`\[sig-storage\] In-tree Volumes \[Driver: vsphere\] \[Testpattern: Dynamic PV \(immediate binding\)\] topology should provision a volume and schedule a pod with AllowedTopologies`,
+			// Skip openstack-specific storage tests in preparation for in-tree cinder provisioner removal
+			// coming with k8s 1.26
+			`\[sig-storage\].*\[Driver: cinder\]`,
```
jsafrane, line 93: There should be a link to a bug to fix the tests.

soltysh (changes requested):
/hold
I've looked into https://github.com/openshift/kubernetes/pull/1432 which is bringing in k8s 1.26 and several random tests I've searched for I still found there being run. So I'd suggest you to compare and see which exact tests should be skipped and only disable those.

mandre:
> /hold I've looked into [openshift/kubernetes#1432](https://github.com/openshift/kubernetes/pull/1432) which is bringing in k8s 1.26 and several random tests I've searched for I still found there being run. So I'd suggest you to compare and see which exact tests should be skipped and only disable those.

Yeah, I've missed https://github.com/kubernetes/kubernetes/pull/113826 which brings back many of the tests that  https://github.com/kubernetes/kubernetes/pull/67782 removed. However we can't really fine-tune the list of tests to skip more than what we're already doing because these tests can't pass until we land the 1.26 rebase.

So in order to not have the openstack job permafailing till the 1.26 rebase, the correct way to go would be to:
- disable all cinder tests (this change)
- merge https://github.com/openshift/kubernetes/pull/1389
- land the 1.26 rebase
- revert this change

Now if I understand @jsafrane's request correctly, he's asking for a jira to track re-enabling the tests which indeed seems like the right thing to do.

I'll create a Jira and update this patch to reference it.

soltysh (approved):
/lgtm
/approve
/hold cancel

test/extended/util/annotate/rules.go
mandre: Done - https://issues.redhat.com/browse/OCPBUGS-5029.

PR #27615, opened 2022-12-16
USHIFT-647: skip non-existing resources from security.openshift.io

pacevedom:
/cc @ingvagabund

### DIFF
--- a/test/extended/cli/explain.go
+++ b/test/extended/cli/explain.go
@@ -557,6 +557,12 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 		g.It(fmt.Sprintf("should contain spec+status for %s [apigroup:%s]", groupName, groupName), func() {
 			for _, bt := range types {
 				e2e.Logf("Checking %s...", bt)
+				exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), bt.Resource, groupName)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				if !exist {
+					e2e.Logf("Resource %s does not exist, skipping", bt)
+					continue
+				}
 				o.Expect(verifySpecStatusExplain(oc, nil, bt)).NotTo(o.HaveOccurred())
 			}
 		})
@@ -586,6 +592,12 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 				e2e.Logf("Checking %s, Field=%s...", st.gv, st.field)
 				resource := strings.Split(st.field, ".")
 				gvr := st.gv.WithResource(resource[0])
+				exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), resource[0], groupName)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				if !exist {
+					e2e.Logf("Resource %s does not exist, skipping", gvr)
+					continue
+				}
 				o.Expect(verifyExplain(oc, nil, gvr,
 					st.pattern, st.field, fmt.Sprintf("--api-version=%s", st.gv))).NotTo(o.HaveOccurred())
 			}


test/extended/cli/explain.go
```diff
@@ -557,6 +557,12 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 		g.It(fmt.Sprintf("should contain spec+status for %s [apigroup:%s]", groupName, groupName), func() {
 			for _, bt := range types {
 				e2e.Logf("Checking %s...", bt)
+				exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), bt.Resource, groupName)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				if !exist {
+					e2e.Logf("Resource %s does not exist, skipping", bt)
```
ingvagabund, line 563: This will show the test as passed, not skipped. This needs to be `g.Skip` instead.

test/extended/cli/explain.go
```diff
@@ -586,6 +592,12 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 				e2e.Logf("Checking %s, Field=%s...", st.gv, st.field)
 				resource := strings.Split(st.field, ".")
 				gvr := st.gv.WithResource(resource[0])
+				exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), resource[0], groupName)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				if !exist {
+					e2e.Logf("Resource %s does not exist, skipping", gvr)
```
ingvagabund, line 598: This will show the test as passed, not skipped. This needs to be `g.Skip` instead.

ingvagabund (changes requested):

test/extended/cli/explain.go
pacevedom: Can't skip the whole test, as this is only skipping specific resources from an API group. Using `g.Skip` would mean skipping the entire API group.

test/extended/cli/explain.go
pacevedom: Can't skip the whole test, as this is only skipping specific resources from an API group. Using `g.Skip` would mean skipping the entire API group.

dhellmann:
@ingvagabund @pacevedom , what's the status of this work?

ingvagabund:
https://github.com/openshift/origin/pull/27615#discussion_r1071177180 waiting for resolution

chiragkyal:
I've opened https://github.com/openshift/origin/pull/27897 PR to address  this one.

test/extended/cli/explain.go
ingvagabund: The test is expected to test entire api group as a whole. If either of the resources is missing, an api group is not considered as fully tested. Which will break the original assumption of testing all the enumerated resources.

It is questionable whether an api group can be only partially tested. That would require properly checking each resource in each apigroup. Which is something that has not been done so far IINM. If the goal is to only test a subset of security.openshift.io group for MicroShift, then I would suggest to consider breaking this specific api group into two of the same group just different list of resources (as long as it make sense). Or figuring out a different way how to partially test this group. E.g. introducing a new test just for it.

chiragkyal: @ingvagabund I checked that all the three resources present under "security.openshift.io" apigroup are not present for MicroShift. What if we skip this entire apigroup because it's applicable for this apigroup only ? Something like 

```go
if !exist && groupName == "security.openshift.io" {
	g.Skip(fmt.Sprintf("Resource %s does not exist, skipping", bt.Resource))
}
```
This will avoid partial testing of other apigroups.

ingvagabund: https://github.com/openshift/microshift/blob/main/docs/enabled_apis.md mentions `SecurityContextConstraints` of `security.openshift.io/v1`

chiragkyal: Added a commit [here](https://github.com/openshift/origin/pull/27897/commits/ca2aa82ce0e75fe51993b59bf95a3a45c7d57d46), which skips only `security.openshift.io` group if any of its resources are not present. 

Then added a new test which checks the resources applicable to MicroShift (e.g `SecurityContextConstraints`)

Please let me know your thoughts on this.

chiragkyal:
/close

https://github.com/openshift/origin/pull/27897 PR got merged, so it's obsolete now.

PR #27619, opened 2022-12-19
USHIFT-720: skip server-side apply for rangeallocations

pacevedom:


ingvagabund:
Checking https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27619/pull-ci-openshift-origin-master-e2e-gcp-ovn/1604794800089862144 there are 17 occurrences:
- brokertemplateinstances
- buildconfigs
- builds
- deploymentconfigs
- groups
- identities
- images
- imagestreams
- oauthaccesstokens
- oauthauthorizetokens
- oauthclientauthorizations
- oauthclients
- rangeallocations
- routes
- templateinstances
- templates
- users

The same resources are enumerated in https://github.com/openshift/origin/blob/master/test/extended/etcd/etcd_storage_path.go#L39. Though, the following are not executed/skipped/failing

Missing:
- projects
- clusterrolebindings
- rolebindings
- clusterroles
- roles

test/extended/apiserver/apply.go
```diff
@@ -69,6 +70,12 @@ var _ = g.Describe("[sig-api-machinery][Feature:ServerSideApply] Server-Side App
 			// create the testing namespace
 			testNamespace := oc.SetupProject()
 
+			exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), gvr.Resource, gvr.GroupVersion().String())
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if !exist {
```
soltysh, line 75: I'm not sure that blank skipping all non-existing API might be good approach, I'd prefer having an exception list which should be consulted and printed in the logs. This will ensure that in case of a problem we will actually catch it, rather than not be aware of it.

soltysh (changes requested):
/lgtm cancel

In general that's again missing the ability to tell it's only on microshift :disappointed:

test/extended/apiserver/apply.go
ingvagabund: The current exclusion of tests (where the test name contains explicit list of apigroups) is based on invocation of `DoesApiResourceExist` under the hood and skipping tests which has at least one API group missing. Which is equivalent to what's happening right here with the change.

> I'd prefer having an exception list which should be consulted and printed in the logs. 

Would you please more elaborate on how the list should look like?

deads2k: This came up a couple times.  If we want a test for "these things must exist", such a test can be created, but it should be distinct from "do the things that exist work".  That makes it easier to write logic here and easier to interpret failures when we see them because it's not testing a side-effect.

I think it's a good idea to create a test for, "this set of resources must exist" and using the microshift subset is a good starting point.

soltysh (approved):
Discussed this on slack more, I agree about the path of having an explicit separate test ensuring that the required set of APIs is present
/lgtm
/approve

ingvagabund:
@pacevedom seems gofmt needs to be run as well

pacevedom:
> @pacevedom seems gofmt needs to be run as well

Done

pacevedom:
> Discussed this on slack more, I agree about the path of having an explicit separate test ensuring that the required set of APIs is present /lgtm /approve

Thanks! Will open a JIRA ticket so that we dont forget.

PR #27628, opened 2023-01-03
Export DeployOAuthServer function for being used by oc login tests

ardaguclu:
Most of the `oc` commands related to the login functionality can not be tested in origin because it requires a temporary oauth server. `deployOAuthFunction` seems like a very good solution to overcome that issue and this PR exports this function for being used by oc login tests.


ardaguclu:
After discussing on Slack with @stlaz, I'm going to create OAuthAccessToken directly instead deploying a temporary OAuth server.

Thus, closing this PR;

/close

PR #27629, opened 2023-01-03
Use create token instead serviceaccounts token command

ardaguclu:
This PR uses suggested `create token` command instead of `serviceaccount new-token` command as stated in the TODO.


soltysh (approved):
/lgtm
/approve

ardaguclu:
unit test failure looks like a real problem

/hold

PR #27633, opened 2023-01-05
Add tests for template cross namespace processing

ardaguclu:
This PR adds new serial suite test in order to verify template processing via `oc process` between different namespaces are working correctly. This test will not work in Disconnected environments because it requires accessing registry images. This test is in serial suite instead of parallel, because it creates temporary user and logins with that by making changes on kubeconfig.

Some steps of test will fail until https://github.com/openshift/oc/pull/1318 is merged.

### DIFF
--- a/test/extended/cli/template.go
+++ b/test/extended/cli/template.go
@@ -1,0 +1,86 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	"github.com/openshift/apiserver-library-go/pkg/authorization/scope"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-cli] templates ", func() {
+	defer g.GinkgoRecover()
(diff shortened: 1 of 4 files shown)


test/extended/cli/template.go
```diff
@@ -0,0 +1,86 @@
...
+
+var _ = g.Describe("[sig-cli] templates ", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc              = exutil.NewCLI("oc-templates")
+		appTemplatePath = exutil.FixturePath("testdata", "cmd", "test", "cmd", "testdata", "application-template-dockerbuild.json")
+	)
+
+	g.It("different namespaces [apigroup:user.openshift.io][apigroup:project.openshift.io][apigroup:template.openshift.io][apigroup:authorization.openshift.io][Serial][Skipped:Disconnected]", func() {
```
soltysh, line 27: One nit, is not to run it as serial, spoke with @ardaguclu on slack about it. The biggest obstacle seems to be switching to a different user, I've pointed him to [ChangeUser](https://github.com/openshift/origin/blob/28f31893176b88eaf3dc893a54b3b4c8dad3ffe2/test/extended/util/client.go#L210) he'll need to double check that.

ardaguclu:
Just to ensure that there is no flakiness.

/test e2e-aws-ovn-single-node-upgrade
/test e2e-openstack-ovn
/test e2e-aws-ovn-cgroupsv2
/test e2e-aws-ovn-fips
/test e2e-aws-ovn-single-node
/test e2e-gcp-ovn
/test e2e-metal-ipi-sdn

PR #27634, opened 2023-01-05
AUTH-365: add PodSecurityViolation to watched alerts

stlaz:
We don't want this to be firing in vanilla OpenShift clusters.

/cc @deads2k @ibihim @s-urbaniak


s-urbaniak:
in favor of https://github.com/openshift/origin/pull/28226

PR #27636, opened 2023-01-06
Fix missing disruption uploads.

dgoodwin:
Discovered that we are getting no data for image registry since Dec 7th
due to PR #27574 which shut off the entire monitor if we have less than
the required 100 runs to appear in the file of allowances to enforce. I
guess at this time, we did not have enough runs with image registry for
4.13. (which is unusual)

Skip the test, but let the monitor run.

[TRT-760](https://issues.redhat.com//browse/TRT-760)

### DIFF
--- a/test/extended/util/disruption/backend_sampler_tester.go
+++ b/test/extended/util/disruption/backend_sampler_tester.go
@@ -134,13 +134,6 @@ func (t *backendDisruptionTest) Test(f *framework.Framework, done <-chan struct{
 	stopCh := make(chan struct{})
 	defer close(stopCh)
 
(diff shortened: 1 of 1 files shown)


dgoodwin:
@derekhiggins can you confirm this looks ok? I believe this still avoids the nil pointer you were getting, but now allows the monitoring to run, report to datafile, and eventually end up in bigquery. We started losing data with the original change.

derekhiggins:
/lgtm

> @derekhiggins can you confirm this looks ok? I believe this still avoids the nil pointer you were getting, but now allows the monitoring to run, report to datafile, and eventually end up in bigquery. We started losing data with the original change.

yup, iirc we should be ok as long as its done before the call to "ExpectNoDisruptionForDuration"

PR #27640, opened 2023-01-09
tests: network: Prefer internal IPs first

r4f4:
Many clusters block direct requests from internal resources to the nodes external IPs as best practice. All accesses from internal resources that want to access resources running on nodes go through load balancers, nodes being on private or public subnets. Let's prefer internal IPs first, so the tests can work even when there are security group rules present blocking requests to the external IPs.

We should not require ExternalIP for Conformance, but should keep testing ExternalIPs in sig network.

This is a carry-over of https://github.com/kubernetes/kubernetes/pull/113799 so that we can enable creating clusters with public-only subnets in CI for AWS and GCP. This should save us in CI spending until we rebase on top of Kubernetes 1.27.


r4f4:
Didn't realise it was already being done https://github.com/openshift/kubernetes/pull/1446

PR #27645, opened 2023-01-10
upgrade/adminack: guarantee one admin ack check post-upgrade

petr-muller:
While looking into OCPBUGS-5505 I discovered that some 4.10->4.11
upgrade job runs perform an Admin Ack check, while some do not. 4.11 has
a `ack-4.11-kube-1.25-api-removals-in-4.12` gate, so these upgrade jobs
sometimes test that `Upgradeable` goes `false` after the ugprade, and
sometimes they do not. This is only determined by the polling race
condition: the check is executed once per 10 minutes, and we cancel the
polling after upgrade is completed. This means that in some cases we are
lucky and manage to run one check before the cancel, and sometimes we
are not and only check while still on the base version.

Example job that checked admin acks post-upgrade:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-cluster-version-operator-880-ci-4.11-upgrade-from-stable-4.10-e2e-azure-upgrade/1611444032104304640
```console
$ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/openshift-cluster-version-operator-880-ci-4.11-upgrade-from-stable-4.10-e2e-azure-upgrade/1611444032104304640/artifacts/e2e-azure-upgrade/openshift-e2e-test/artifacts/e2e.log | grep 'Waiting for Upgradeable to be AdminAckRequired'
Jan  6 21:16:40.153: INFO: Waiting for Upgradeable to be AdminAckRequired ...
```

Example job that did not check admin acks post-upgrade:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-cluster-version-operator-880-ci-4.11-upgrade-from-stable-4.10-e2e-azure-upgrade/1611444033509396480
```console
$ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/openshift-cluster-version-operator-880-ci-4.11-upgrade-from-stable-4.10-e2e-azure-upgrade/1611444033509396480/artifacts/e2e-azure-upgrade/openshift-e2e-test/artifacts/e2e.log | grep 'Waiting for Upgradeable to be AdminAckRequired'
```

Add a guaranteed single check execution after the upgrade, so that admin
ack is always checked at least once with the upgrade target version.
Doing checks after `done` is signalled has prior art in the [alert test](https://github.com/openshift/origin/blob/c081a2a8095237cee1fca893d750e7eb10975924/test/e2e/upgrade/alert/alert.go#L47-L49).

### DIFF
--- a/test/e2e/upgrade/adminack/adminack.go
+++ b/test/e2e/upgrade/adminack/adminack.go
@@ -51,6 +51,13 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 
 	adminAckTest := &clusterversionoperator.AdminAckTest{Oc: t.oc, Config: t.config, Poll: 10 * time.Minute}
 	adminAckTest.Test(ctx)
+
+	// Perform one guaranteed check after the upgrade is complete. The polled check above can be
+	// cancelled on done signal (which means, so we never know whether the poll was lucky to run at least once
+	// since the version was bumped.
+	postUpdateCtx, postUpdateCtxCancel := context.WithTimeout(context.Background(), 15*time.Minute)
+	defer postUpdateCtxCancel()
+	(&clusterversionoperator.AdminAckTest{Oc: t.oc, Config: t.config}).Test(postUpdateCtx)
 }
 
 // Teardown cleans up any remaining objects.
--- a/test/e2e/upgrade/adminack/adminack.go
+++ b/test/e2e/upgrade/adminack/adminack.go
@@ -51,13 +51,6 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 
 	adminAckTest := &clusterversionoperator.AdminAckTest{Oc: t.oc, Config: t.config, Poll: 10 * time.Minute}
 	adminAckTest.Test(ctx)
-
-	// Perform one guaranteed check after the upgrade is complete. The polled check above can be
-	// cancelled on done signal (which means, so we never know whether the poll was lucky to run at least once
-	// since the version was bumped.
-	postUpdateCtx, postUpdateCtxCancel := context.WithTimeout(context.Background(), 15*time.Minute)
-	defer postUpdateCtxCancel()
-	(&clusterversionoperator.AdminAckTest{Oc: t.oc, Config: t.config}).Test(postUpdateCtx)
 }
 
 // Teardown cleans up any remaining objects.
--- a/test/extended/util/openshift/clusterversionoperator/adminack.go
+++ b/test/extended/util/openshift/clusterversionoperator/adminack.go
@@ -40,7 +40,9 @@ var adminAckGateRegexp = regexp.MustCompile(adminAckGateFmt)
 // and verifies that the Upgradeable condition no longer complains about the ack.
 func (t *AdminAckTest) Test(ctx context.Context) {
 	if t.Poll == 0 {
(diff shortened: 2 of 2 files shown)


test/e2e/upgrade/adminack/adminack.go
```diff
@@ -47,6 +47,10 @@ func (t *UpgradeTest) Test(f *framework.Framework, done <-chan struct{}, upgrade
 	go func() {
 		<-done
 		cancel()
+		// Perform one guaranteed check after the upgrade is complete. We cancel the polled check
+		// above, so we never know whether the poll was lucky to run at least once since the version
+		// was bumped.
+		(&clusterversionoperator.AdminAckTest{Oc: t.oc, Config: t.config}).Test(context.Background())
```
wking, line 53: do we want to cap the amount of time we're willing to spend on this?   `context.WithTimeout(context.Background(), 5*time.Minute)` or similar?

petr-muller: I considered that but I checked the other place that does this one-off check and it also just does `context.Background()`:

https://github.com/openshift/origin/blob/74d8b6dc8a1803fd6153e5abbe15f96e67cc763e/test/extended/adminack/adminack.go#L23-L26

But yeah, it's probably better to cap that. I'll try to come up with some reasonable upper bound - there seems to be multiple timeouted polls inside the check, potentially even iterated, the upper bound needs to match the expected sum of these.

petr-muller: There are two 3 minute waits for each applicable gate. I don't think we ever had more than one gate but with a little future-proofing I set the timeout to 15m  to handle a potential case where we have two gates and somehowtthings go slowly enough to hit the three minute timeout on each wait (this is IMO totally pathological case).

petr-muller: Huh I guess I cannot do the check here because `Test` will terminate while we're still testing in the goroutine.

petr-muller:
The results on upgrade jobs here look sane (nothing looks broken) but there's no gate on 4.13 so the test is not doing much

test/extended/util/openshift/clusterversionoperator/adminack.go
```diff
@@ -66,11 +68,21 @@ func (t *AdminAckTest) Test(ctx context.Context) {
...
-func (t *AdminAckTest) test(ctx context.Context, exercisedGates map[string]struct{}) error {
-	exists := struct{}{}
+	// Perform one guaranteed check after the upgrade is complete. The polled check above is cancelled
+	// on done signal (so we never know whether the poll was lucky to run at least once since
+	// the version was bumped).
+	postUpdateCtx, postUpdateCtxCancel := context.WithTimeout(context.Background(), 15*time.Minute)
+	defer postUpdateCtxCancel()
+	if current := getCurrentVersion(postUpdateCtx, t.Config); current != "" && !exercisedVersions.Has(current) {
+		// We never saw the current version while polling, so lets check it now
+		if err := t.test(ctx, exercisedGates, exercisedVersions); err != nil {
```
wking, line 79: do we want `postUpdateCtx` here too?

petr-muller: thanks, good catch!

wking (approved):
/lgtm

PR #27647, opened 2023-01-10
USHIFT-741: Microshift invariant adaptations

pacevedom:


pkg/synthetictests/disruption.go
```diff
@@ -33,6 +35,11 @@ func testServerAvailability(
 	backendName := fmt.Sprintf("%s-%s-connections", disruptionName, connType)
 	jobType, err := platformidentification.GetJobType(context.TODO(), restConfig)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
```
ingvagabund, line 38: Not found error can be in some cases a transitional error. I don't think `testServerAvailability` takes that into account. Nevertheless, seeing `platformidentification.GetJobType` as a black box, it's not safe to deduce non-existence of a certain API from a NotFoundErr. Worth considering a different approach or to refactore `platformidentification.GetJobType` to provide more output.

pkg/synthetictests/networking.go
```diff
@@ -104,6 +93,19 @@ func testPodSandboxCreation(events monitorapi.Intervals, clientConfig *rest.Conf
 			continue
 		}
 		if strings.Contains(event.Message, "pinging container registry") && strings.Contains(event.Message, "i/o timeout") {
+			if platform == "" {
```
ingvagabund, line 96: `platform` will be always an empty string here.

pkg/synthetictests/disruption.go
pacevedom: Changed it to be a bit more robust now. But could that happen? The tests only start after the cluster has been deemed ready, which means all resources and componentes are ready too: https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/e2e/test/openshift-e2e-test-commands.sh#L294-L418

pkg/synthetictests/networking.go
pacevedom: It is defined [here](https://github.com/openshift/origin/pull/27647/files#diff-1fd611ecbabf8351af6690c1f2a51335b15d052c7fbdf288686335f87c9528b8R96) and then grabs a value [here](https://github.com/openshift/origin/pull/27647/files#diff-1fd611ecbabf8351af6690c1f2a51335b15d052c7fbdf288686335f87c9528b8R105) for the next for loop iterations.

ingvagabund (changes requested):

pkg/synthetictests/alerts.go
```diff
@@ -8,13 +8,34 @@ import (
...
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 	"github.com/openshift/origin/pkg/synthetictests/allowedalerts"
 )
 
 func testAlerts(events monitorapi.Intervals, restConfig *rest.Config, duration time.Duration, recordedResource *monitorapi.ResourcesMap) []*junitapi.JUnitTestCase {
 	ret := []*junitapi.JUnitTestCase{}
 
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		ret = append(ret, &junitapi.JUnitTestCase{
```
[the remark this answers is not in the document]
pacevedom: Done

pkg/synthetictests/alerts.go
```diff
@@ -8,13 +8,34 @@ import (
...
+				Output: err.Error(),
+			},
+			SystemOut: err.Error(),
+		})
+		return ret
+	}
+
+	_, err = kubeClient.CoreV1().Namespaces().Get(context.Background(), "openshift-monitoring", metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return ret
```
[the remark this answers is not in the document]
pacevedom: Intention is to be nil. If there is no monitoring then there are no alert tests, and no alert tests mean no alert failures.

pkg/synthetictests/disruption.go
```diff
@@ -26,6 +26,22 @@ func testServerAvailability(
 
 	testName := fmt.Sprintf("[%s] %s should be available throughout the test", owner, locator)
 
+	skip, err := platformidentification.CanExtractPlatform(restConfig)
```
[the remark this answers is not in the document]
ingvagabund: This change is important as well. So it's more obvious for a reader who sees the code for the first time why we check this here.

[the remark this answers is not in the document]
pacevedom: Done

pkg/synthetictests/disruption.go
```diff
@@ -26,6 +26,22 @@ func testServerAvailability(
...
+				Name:     testName,
+				Duration: jobRunDuration.Seconds(),
+				FailureOutput: &junitapi.FailureOutput{
+					Output: fmt.Sprintf("error in platform identification: %s", err),
+				},
+			},
+		}
+	}
+	if !skip {
+		return []*junitapi.JUnitTestCase{}
```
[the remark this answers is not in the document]
pacevedom: Just realized the name of the `skip` variable is actually the opposite. Changed it.

pkg/synthetictests/alerts.go
```diff
@@ -8,13 +8,35 @@ import (
...
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 	"github.com/openshift/origin/pkg/synthetictests/allowedalerts"
 )
 
 func testAlerts(events monitorapi.Intervals, restConfig *rest.Config, duration time.Duration, recordedResource *monitorapi.ResourcesMap) []*junitapi.JUnitTestCase {
+	
+
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		return &junitapi.JUnitTestCase{
```
ingvagabund, line 24: This needs to be 
```go
return []*junitapi.JUnitTestCase{
	{
		Name: "Alert setup, kube client",
		FailureOutput: &junitapi.FailureOutput{
			Output: err.Error(),
		},
		SystemOut: err.Error(),
	},
}
```

pacevedom: Oops, done.

pkg/synthetictests/alerts.go
```diff
@@ -8,13 +8,36 @@ import (
...
 	"github.com/openshift/origin/pkg/synthetictests/allowedalerts"
 )
 
 func testAlerts(events monitorapi.Intervals, restConfig *rest.Config, duration time.Duration, recordedResource *monitorapi.ResourcesMap) []*junitapi.JUnitTestCase {
+
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		return []*junitapi.JUnitTestCase{
+			{
+				Name: "Alert setup, kube client",
```
deads2k, line 25: This results in an "only failed" test.  it needs to set to "pass" in the majority of cases so that aggregation works properly.  I'm surprised we don't have clients we can pass in here.

pkg/synthetictests/disruption.go
```diff
@@ -26,6 +26,22 @@ func testServerAvailability(
 
 	testName := fmt.Sprintf("[%s] %s should be available throughout the test", owner, locator)
 
+	canDetermineJobType, err := platformidentification.CanExtractJobType(restConfig)
+	if err != nil {
+		return []*junitapi.JUnitTestCase{
```
deads2k, line 31: same "only failure" test here.

pkg/synthetictests/networking.go
deads2k, line 96: `if len(platform) == 0` is kube and openshift canonical

pkg/synthetictests/networking.go
```diff
@@ -104,6 +93,19 @@ func testPodSandboxCreation(events monitorapi.Intervals, clientConfig *rest.Conf
...
+				} else {
+					infra, err := configClient.ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+					if err != nil {
+						failures = append(failures, fmt.Sprintf("error getting cluster infrastructure: %v", err))
+					} else {
+						platform = infra.Status.PlatformStatus.Type
+					}
+				}
+			}
 			if platform == v1.AzurePlatformType {
```
deads2k, line 109: I think you may be better off always failing if you cannot determine the platform, but this amounts to the same thing with a bad message if TRT can stomach the factorization.

PR #27648, opened 2023-01-10
Separate mcd error alert

xueqzhan:
[TRT-735](https://issues.redhat.com/browse/TRT-735)

### DIFF
--- a/pkg/synthetictests/allowedalerts/all.go
+++ b/pkg/synthetictests/allowedalerts/all.go
@@ -62,6 +62,9 @@ func AllAlertTests(ctx context.Context, clientConfig *rest.Config, duration time
 	ret = append(ret, newAlert("machine config operator", "MCDDrainError").pending().neverFail().toTests()...)
 	ret = append(ret, newAlert("machine config operator", "MCDDrainError").firing().toTests()...)
 
+	ret = append(ret, newAlert("machine config operator", "MCDPivotError").pending().neverFail().toTests()...)
+	ret = append(ret, newAlert("machine config operator", "MCDPivotError").firing().toTests()...)
+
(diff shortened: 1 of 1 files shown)


pkg/synthetictests/allowedalerts/all.go
```diff
@@ -62,6 +62,9 @@ func AllAlertTests(ctx context.Context, clientConfig *rest.Config, duration time
 	ret = append(ret, newAlert("machine config operator", "MCDDrainError").pending().neverFail().toTests()...)
 	ret = append(ret, newAlert("machine config operator", "MCDDrainError").firing().toTests()...)
 
+	ret = append(ret, newAlert("machine config operator", "MCDPivotError").pending().neverFail().toTests()...)
+	ret = append(ret, newAlert("machine config operator", "MCDPivotError").firing().toTests()...)
```
dgoodwin, line 66: Looking at your presubmits I suspect this will take out payloads, can we do neverFail for the firing as well, with a TODO comment. We can track the flake percentage to know if we're getting better.

xueqzhan: I see one job with this failing. I want to rerun this job to see how often this fails.

dgoodwin: I thought I actually saw two but I can't find another now, possibly I clicked the same link twice. Rerun sounds good lets hope it's ok.

xueqzhan: In the new run, this test did not fail again.

dgoodwin: Ok lets bring it in and we can just keep an eye on payloads. 

@deepsm007 heads up on this, if we see payloads rejected on a new MCDPivotError alert test, this is the PR.

It's possible with it being a new test, aggregation will not blow up on it?

PR #27650, opened 2023-01-10
OCPBUGS-5506: DNS pod to be created on master node

deepsm007:
[sig-network-edge] DNS should answer queries using the local DNS 
endpoint [Suite:openshift/conformance/parallel] tests is perma failing for heteroenegeneous due to missing manifest list support for image registry.

Assigning the dns pod to master to avoid deploying to the incorrect node and fixing issue for  `exec format error`


deepsm007:
> /lgtm /approve

Do we need to add toleration to schedule pod on master?

deads2k:
verify needs fixing.

PR #27651, opened 2023-01-11
Fix missing disruption data again.

dgoodwin:
Previous attempt did not work, as we were still bypassing
FrameworkEventIntervals() which is needed to add the test summary, which
eventually results in the AdditionalEvents json files in artifacts,
which are then merged into the overall events and disruption data files.
Without calling it, we still get no disruption data uploaded.

Move the nil check into the ExpectNoDisruption function, and allow
FrameworkEventIntervals to run as it should.

[TRT-760](https://issues.redhat.com//browse/TRT-760)

### DIFF
--- a/test/extended/util/disruption/disruption.go
+++ b/test/extended/util/disruption/disruption.go
@@ -335,16 +335,26 @@ func createTestFrameworks(tests []upgrades.Test) map[string]*framework.Framework
 
 // ExpectNoDisruptionForDuration fails if the sum of the duration of all events exceeds allowedDisruption, reports a
 // disruption flake if any disruption occurs, and uses reason to prefix the message.
-func ExpectNoDisruptionForDuration(f *framework.Framework, allowedDisruption time.Duration, total time.Duration, events monitorapi.Intervals, reason string) {
+func ExpectNoDisruptionForDuration(f *framework.Framework, testName string, allowedDisruption *time.Duration, total time.Duration, events monitorapi.Intervals, reason string) {
+	// This step records the test summaries data we need to result in AdditionalEvents json files in
+	// the openshift-e2e-test artifacts.
 	FrameworkEventIntervals(f, events)
 	describe := events.Strings()
+
+	// Indicates there is no entry in the query_results.json data file, nor a valid fallback,
+	// we do not wish to run the test. (this likely implies we do not have the required number of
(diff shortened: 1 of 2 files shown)


dgoodwin:
I am quite confused how the previous fix failed because I could see image registry disruption being gathered in the presubmits, i.e. https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27636/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1612521262419021824/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/AdditionalEvents__sig_imageregistry_Image_registry_remains_available_using_reused_connections_2023-01-09T21:59:13Z.json

But none once it went live in similar jobs: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.13-e2e-aws-ovn-upgrade/1613061514967650304/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/

It must be a case where somehow we found a fallback on the presubmit to use, perhaps due to a different or missing FromRelease.

Regardless I think this PR will correct it.

test/extended/util/disruption/disruption.go
```diff
@@ -335,16 +335,26 @@ func createTestFrameworks(tests []upgrades.Test) map[string]*framework.Framework
 
 // ExpectNoDisruptionForDuration fails if the sum of the duration of all events exceeds allowedDisruption, reports a
 // disruption flake if any disruption occurs, and uses reason to prefix the message.
-func ExpectNoDisruptionForDuration(f *framework.Framework, allowedDisruption time.Duration, total time.Duration, events monitorapi.Intervals, reason string) {
+func ExpectNoDisruptionForDuration(f *framework.Framework, testName string, allowedDisruption *time.Duration, total time.Duration, events monitorapi.Intervals, reason string) {
+	// This step records the test summaries data we need to result in AdditionalEvents json files in
+	// the openshift-e2e-test artifacts.
 	FrameworkEventIntervals(f, events)
```
xueqzhan, line 341: I wonder if this line should be moved outside of the function. It can easily be missed. Your comment certainly helps.

PR #16408, opened 2017-09-18
test: extended: avoid using docker directly

runcom:
This test fails when running against CRI-O (https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_crio/40/console) because it uses docker.

Modify the test to check groups in container instead.

ref https://github.com/openshift/origin/issues/16403
@smarterclayton PTAL

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #27664, opened 2023-01-17
make supplemental groups test working again

atiratree:
- remove user group requirement for the test

Remove [Local] from test as it is not local anymore since https://github.com/openshift/origin/pull/16408 . So this test has not been run in the CI for a long time. And locally as well, since it has been broken because of the removal of `exec -p` feature in https://github.com/kubernetes/kubernetes/pull/76713.

### DIFF
--- a/test/extended/security/supplemental_groups.go
+++ b/test/extended/security/supplemental_groups.go
@@ -10,12 +10,16 @@ import (
 	o "github.com/onsi/gomega"
 
 	kapiv1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/util/retry"
+	rbacv1helpers "k8s.io/kubernetes/pkg/apis/rbac/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	admissionapi "k8s.io/pod-security-admission/api"
 
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/image"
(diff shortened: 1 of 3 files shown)


test/extended/security/supplemental_groups.go
```diff
@@ -28,19 +32,22 @@ const (
...
 
 	g.Describe("Ensure supplemental groups propagate to docker", func() {
-		g.It("should propagate requested groups to the container [Local][apigroup:user.openshift.io][apigroup:security.openshift.io]", func() {
+		g.It("should propagate requested groups to the container [apigroup:security.openshift.io]", func() {
 
 			fsGroup := int64(1111)
 			supGroup := int64(2222)
 
+			projectName := oc.Namespace()
+			sa := createServiceAccount(ctx, oc, projectName)
```
atiratree, line 45: reusing this func from scc.go, should I moveit to a separate file like `common.go`?

stlaz: I'm not sure whether it was me who originally reviewed the PR of that SA creation, but this wait piece of the code is weird:

https://github.com/openshift/origin/blob/79baa1b40aa634e7a29b5c01abf756d70c58c0a4/test/extended/security/scc.go#L479-L491

If you're to re-use it, replace it with
https://github.com/openshift/origin/blob/79baa1b40aa634e7a29b5c01abf756d70c58c0a4/test/extended/util/framework.go#L1023-L1024

No need to export it or move it anywhere yet, I think.

test/extended/security/supplemental_groups.go
```diff
@@ -28,19 +32,22 @@ const (
...
+			supSubject := fmt.Sprintf("system:serviceaccount:%s:%s", projectName, sa.Name)
+			createSupRoleOrDie(ctx, oc, sa)
 			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-				_, err := oc.AsAdmin().Run("adm").Args("policy", "add-scc-to-user", "anyuid", oc.Username()).Output()
+				// sa serves as a subject instead of the user
+				_, err := oc.AsAdmin().Run("adm").Args("policy", "add-scc-to-user", "anyuid", supSubject).Output()
```
stlaz, line 50: IIRC `-z <SA name> -n <namespace>` is what should be used here, right?

stlaz (changes requested):

test/extended/security/supplemental_groups.go
```diff
@@ -50,29 +57,31 @@ var _ = g.Describe("[sig-node] supplemental groups", func() {
...
 
-			// wait for the pod to run so we can inspect it.
+			// wait for the pod to run, so we can inspect it.
 			g.By("waiting for the pod to become running")
-			err = e2epod.WaitForPodNameRunningInNamespace(f.ClientSet, submittedPod.Name, f.Namespace.Name)
+			err = e2epod.WaitForPodNameRunningInNamespace(supClient, submittedPod.Name, projectName)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
-			out, stderr, err := oc.Run("exec").Args("-p", supplementalGroupsPod, "--", "/usr/bin/id", "-G").Outputs()
+			out, stderr, err := oc.Run("exec").Args(supplementalGroupsPod, "--as", supSubject, "--", "/usr/bin/id", "-G").Outputs()
```
stlaz, line 84: This should likely work even without that impersonation

test/extended/security/supplemental_groups.go
atiratree: both of these options should work, but yes this makes the intent more clearer

test/extended/security/supplemental_groups.go
atiratree: +1, using the `WaitForServiceAccount` now

test/extended/security/supplemental_groups.go
atiratree: it does not 

```
"Error from server (Forbidden): pods \"supplemental-groups\" is forbidden: exec operation is not allowed because the pod's security context exceeds your permissions: pods \"supplemental-groups\" is forbidden: unable to validate against any security context constraint: [provider \"anyuid\": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .spec.securityContext.fsGroup: Invalid value: []int64{1111}: 1111 is not an allowed group, provider \"restricted\": Forbidden: not usable by user or serviceaccount, provider \"nonroot-v2\": Forbidden: not usable by user or serviceaccount, provider \"nonroot\": Forbidden: not usable by user or serviceaccount, provider \"hostmount-anyuid\": Forbidden: not usable by user or serviceaccount, provider \"machine-api-termination-handler\": Forbidden: not usable by user or serviceaccount, provider \"hostnetwork-v2\": Forbidden: not usable by user or serviceaccount, provider \"hostnetwork\": Forbidden: not usable by user or serviceaccount, provider \"hostaccess\": Forbidden: not usable by user or serviceaccount, provider \"node-exporter\": Forbidden: not usable by user or serviceaccount, provider \"privileged\": Forbidden: not usable by user or serviceaccount]"
```

we have access to the service account which has elevated privileges.

Other option is to run this as an admin.

stlaz: I thought admin was the default user. Yep, run it like so :+1:

soltysh (approved):
/approve

PR #27665, opened 2023-01-17
WIP: Bug OCPBUGS-4425: E2E networking tests: add retry to EgressFW

martinkennelly:
EgressFW tests are flaking on metal jobs.
Its suspected the tests are failing because
when we apply an EgressFirewall obj, it does
not mean OVN has applied this policy to the whole
cluster but that the obj is applied to the K API.
EgressFirewall does not expose a status or any
other mechanism to indicate when it is applied
throughout the cluster, therefore we must retry
until we get the result we expect.

Signed-off-by: Martin Kennelly <mkennell@redhat.com>

### DIFF
--- a/test/extended/networking/egress_firewall.go
+++ b/test/extended/networking/egress_firewall.go
@@ -3,6 +3,7 @@ package networking
 import (
 	"context"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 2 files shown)


martinkennelly:
/close
Retry fixed nothing.

PR #27675, opened 2023-01-19
builds: exercise openshift-manage-dockerfile with HEALTHCHECK

nalind:
Add a HEALTHCHECK instruction to one of the Dockerfiles that we build, to ensure that openshift-manage-dockerfile doesn't create a syntax error by incorrectly reconstructing a line for that instruction.


nalind:
Now that https://github.com/openshift/builder/pull/323 has been merged, I expect this to start passing.

PR #27678, opened 2023-01-23
OCPBUGS-6503: upgrade/adminack: simplify polling and unblock "guaranteed" post-upgrade check

petr-muller:
https://github.com/openshift/origin/pull/27645 intended to add a guaranteed post-upgrade check but I have overlooked how exactly the polling is implemented and terminated, leading to the post-upgrade check never actually execute.

Previously the test used `PollImmediateWithContext` for the each-10-minutes check. The `ConditionFunc` never actually returned `true` or non-nil `err`, so the `PollImmediateWithContext` never terminated by the means of `ConditionFunc`: it was always terminated by the `ctx.Done()` that the framework does on finished upgrade (or a test timeout). This means that `PollImmediateWithContext` always terminated with `err=wait.ErrWaitTimeout` and the `Test` method immediately returned, so the "guaranteed" check code is never reached.

Given our `ConditionFunc` never terminates the polling, we can simplify and use the `wait.UntilWithContext` instead, which is a simpler version that precisely implements the desired loop (poll until context is done).

---

During testing of OCPBUGS-5505, it was discovered that even with shortening the CVO cache TTL, CVO may still only update `Upgradeable` in its sync interval, which may be as high as 4 minutes. Hence the tests needs to wait for that time (I added 5 second buffer on top of that).

### DIFF
--- a/test/extended/util/openshift/clusterversionoperator/adminack.go
+++ b/test/extended/util/openshift/clusterversionoperator/adminack.go
@@ -51,19 +51,14 @@ func (t *AdminAckTest) Test(ctx context.Context) {
 	exercisedVersions := sets.NewString()
 	success := false
 	var lastError error
-	if err := wait.PollImmediateUntilWithContext(ctx, t.Poll, func(ctx context.Context) (bool, error) {
+	wait.UntilWithContext(ctx, func(ctx context.Context) {
 		if err := t.test(ctx, exercisedGates, exercisedVersions); err != nil {
 			framework.Logf("Retriable failure to evaluate admin acks: %v", err)
 			lastError = err
 		} else {
 			success = true
 		}
-		return false, nil
-	}); err == nil || err == wait.ErrWaitTimeout {
-		return
-	} else {
-		framework.Fail(err.Error())
-	}
+	}, t.Poll)
 
 	if !success {
 		framework.Failf("Never able to evaluate admin acks.  Most recent failure: %v", lastError)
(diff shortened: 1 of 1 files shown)


petr-muller:
/retest

CI registry is unhappy:
```
error: unable to read image registry.build04.ci.openshift.org/ci-op-9xtqjngg/stable@sha256:4a6ecaf7b68d463869afb275f56d4c2cbfea2bbf295d8a599bc15399ba8e3406: received unexpected HTTP status: 500 Internal Server Error 
```

petr-muller:
GCP jobs affected by OHSS-18195, hope the AWS ones give me something to work with

petr-muller:
`ci/prow/e2e-aws-ovn-upgrade` is actually a rollback job, that's surprising :suspect: https://github.com/openshift/release/blob/8181fa1f44d0d94117a99cfd6c02dc4bad46b21c/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L206

wking (approved):
/lgtm

petr-muller:
/retest 
ci/prow/e2e-gcp-ovn-upgrade

petr-muller:
` [bz-OLM][invariant] alert/KubePodNotReady should not be at or above info in ns/openshift-marketplace` 😐 

/test e2e-aws-ovn-fips

petr-muller:
/override 
ci/prow/e2e-aws-ovn-fips

PR #27681, opened 2023-01-23
test/extended/authorization/rbac: Condition console RBAC on 'Console' capability

wking:
Clusters that disable the `Console` capability are [currently failing this test-case][1]:

```
  : [sig-auth][Feature:OpenShiftAuthorization] The default cluster RBAC policy should have correct RBAC rules [Suite:openshift/conformance/parallel] expand_less
  Run #0: Failed expand_less	3s
  {  fail [github.com/openshift/origin/test/extended/authorization/rbac/groups_default_rules.go:229]: Jan  3 13:43:14.134: test data for system:authenticated has too many unnecessary permissions:
  {APIGroups:["console.openshift.io"], Resources:["consoleclidownloads"], Verbs:["get" "list" "watch"]}
  {APIGroups:["console.openshift.io"], Resources:["consoleexternalloglinks"], Verbs:["get" "list" "watch"]}
  {APIGroups:["console.openshift.io"], Resources:["consolelinks"], Verbs:["get" "list" "watch"]}
  {APIGroups:["console.openshift.io"], Resources:["consolenotifications"], Verbs:["get" "list" "watch"]}
  {APIGroups:["console.openshift.io"], Resources:["consoleplugins"], Verbs:["get" "list" "watch"]}
  {APIGroups:["console.openshift.io"], Resources:["consolequickstarts"], Verbs:["get" "list" "watch"]}
  {APIGroups:["console.openshift.io"], Resources:["consoleyamlsamples"], Verbs:["get" "list" "watch"]}
  {APIGroups:["helm.openshift.io"], Resources:["helmchartrepositories"], Verbs:["get" "list"]}
  {APIGroups:["snapshot.storage.k8s.io"], Resources:["volumesnapshotclasses"], Verbs:["get" "list" "watch"]}
  Ginkgo exit error 1: exit with code 1}
```

This commit uses the pattern that 6d1143a0f2 (#26998) began using for CRDs to only add the console-linked rules when `Console` is enabled, and to only add the snapshot-linked rules when `CSISnapshot` is enabled.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.13-e2e-aws-sdn-no-capabilities/1610257913278894080

### DIFF
--- a/test/extended/authorization/rbac/groups_default_rules.go
+++ b/test/extended/authorization/rbac/groups_default_rules.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/openshift/api/authorization"
 	"github.com/openshift/api/build"
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/console"
 	"github.com/openshift/api/image"
 	"github.com/openshift/api/oauth"
@@ -118,18 +119,9 @@ var (
 			rbacv1helpers.NewRule("get", "list").Groups(authzGroup, legacyAuthzGroup).Resources("clusterroles").RuleOrDie(),
 			rbacv1helpers.NewRule(read...).Groups(rbacGroup).Resources("clusterroles").RuleOrDie(),
 			rbacv1helpers.NewRule("get", "list").Groups(storageGroup).Resources("storageclasses").RuleOrDie(),
-			rbacv1helpers.NewRule("get", "list", "watch").Groups(snapshotGroup).Resources("volumesnapshotclasses").RuleOrDie(),
 			rbacv1helpers.NewRule("list", "watch").Groups(projectGroup, legacyProjectGroup).Resources("projects").RuleOrDie(),
 
 			rbacv1helpers.NewRule("use").Groups(security.GroupName).Resources("securitycontextconstraints").Names("restricted-v2").RuleOrDie(),
-
-			// These custom resources are used to extend console functionality
-			// The console team is working on eliminating this exception in the near future
-			rbacv1helpers.NewRule(read...).Groups(consoleGroup).Resources("consoleclidownloads", "consolelinks", "consoleexternalloglinks", "consolenotifications", "consoleyamlsamples", "consolequickstarts", "consoleplugins").RuleOrDie(),
-
-			// HelmChartRepository instances keep Helm chart repository configuration
-			// By default users are able to browse charts from all configured repositories through console UI
-			rbacv1helpers.NewRule("get", "list").Groups("helm.openshift.io").Resources("helmchartrepositories").RuleOrDie(),
 
 			// TODO: remove when openshift-apiserver has removed these
 			rbacv1helpers.NewRule("get").URLs(
@@ -183,15 +175,52 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] The default clust
 	oc := exutil.NewCLI("default-rbac-policy")
 
 	g.It("should have correct RBAC rules", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		defer cancel()
+
+		enabledCapabilities := []configv1.ClusterVersionCapability{}
+
+		exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), "clusterversions", "config.openshift.io")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if exist {
+			clusterVersion, err := oc.AdminConfigClient().ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+			if err != nil {
+				e2e.Failf("Failed to get cluster version: %v", err)
+			}
+			enabledCapabilities = append(enabledCapabilities, clusterVersion.Status.Capabilities.EnabledCapabilities...)
+		} else {
+			e2e.Fail("Cluster version API resource does not exist")
+		}
+
+		// Conditional, capability-specific rules
+		for _, capability := range enabledCapabilities {
+			switch capability {
+			case configv1.ClusterVersionCapabilityConsole:
+				allAuthenticatedRules = append(
+					allAuthenticatedRules,
+					[]rbacv1.PolicyRule{
+						// These custom resources are used to extend console functionality
+						// The console team may eventually eliminate this exception
+						rbacv1helpers.NewRule(read...).Groups(consoleGroup).Resources("consoleclidownloads", "consolelinks", "consoleexternalloglinks", "consolenotifications", "consoleyamlsamples", "consolequickstarts", "consoleplugins").RuleOrDie(),
+
+						// HelmChartRepository instances keep Helm chart repository configuration
+						// By default users are able to browse charts from all configured repositories through console UI
+						rbacv1helpers.NewRule("get", "list").Groups("helm.openshift.io").Resources("helmchartrepositories").RuleOrDie(),
+					}...,
+				)
+
+			case configv1.ClusterVersionCapabilityCSISnapshot:
+				allAuthenticatedRules = append(
+					allAuthenticatedRules,
+					rbacv1helpers.NewRule("get", "list", "watch").Groups(snapshotGroup).Resources("volumesnapshotclasses").RuleOrDie(),
+				)
+			}
+		}
+
 		kubeInformers := informers.NewSharedInformerFactory(oc.AdminKubeClient(), 20*time.Minute)
 		ruleResolver := exutil.NewRuleResolver(kubeInformers.Rbac().V1()) // signal what informers we want to use early
 
-		stopCh := make(chan struct{})
-		defer func() { close(stopCh) }()
-		kubeInformers.Start(stopCh)
-
-		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-		defer cancel()
+		kubeInformers.Start(ctx.Done())
 
 		if ok := cache.WaitForCacheSync(ctx.Done(),
 			kubeInformers.Rbac().V1().ClusterRoles().Informer().HasSynced,
@@ -202,7 +231,7 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] The default clust
 			exutil.FatalErr("failed to sync RBAC cache")
 		}
 
-		namespaces, err := oc.AdminKubeClient().CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{})
+		namespaces, err := oc.AdminKubeClient().CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
 		if err != nil {
 			exutil.FatalErr(err)
 		}


wking:
Launching [a payload run][1] to see whether this fix works or needs further tweaks:

/payload-job periodic-ci-openshift-release-master-ci-4.13-e2e-aws-sdn-no-capabilities

/hold

[1]: https://docs.ci.openshift.org/docs/release-oversight/payload-testing/

bparees:
/approve

let's see how the test job does

wking:
/payload-job periodic-ci-openshift-release-master-ci-4.13-e2e-aws-sdn-no-capabilities

wking:
[Payload job][1]:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/logs/openshift-origin-27681-ci-4.13-e2e-aws-sdn-no-capabilities/1618038496394481664/build-log.txt | grep 'The default cluster RBAC policy should have 
correct RBAC rules'
started: 0/127/413 "[sig-auth][Feature:OpenShiftAuthorization] The default cluster RBAC policy should have correct RBAC rules [Suite:openshift/conformance/parallel]"
passed: (4.7s) 2023-01-25T01:07:26 "[sig-auth][Feature:OpenShiftAuthorization] The default cluster RBAC policy should have correct RBAC rules [Suite:openshift/conformance/parallel]"
```

/hold cancel

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-origin-27681-ci-4.13-e2e-aws-sdn-no-capabilities/1618038496394481664

test/extended/authorization/rbac/groups_default_rules.go
```diff
@@ -183,15 +175,46 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] The default clust
 	oc := exutil.NewCLI("default-rbac-policy")
 
 	g.It("should have correct RBAC rules", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		defer cancel()
+
+		exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), "clusterversions", "config.openshift.io")
```
bparees, line 181: do we ever expect this not to exist?  when?

and if so, i'm inclined to think we'd want to check all the rbac rules (because if clusterversions doesn't exist somehow, then presumably all caps should be enabled), rather than what the current logic does which is to ignore the rbac for optionalcaps when the clusterversion resource type does not exist.

wking: This portion of my code is pattern-matching from [here][1], where it looks like the goal is supporting MicroShift (#27540).  I'm not sure how reliable the "no ClusterVersion" -> "so it's MicroShift" -> "so we want all these conditional RBAC rules" chain is, so maybe we want to pivot to "if the resource the RBAC relates to is present, expect the RBAC to be present"?

[1]: https://github.com/openshift/origin/blame/46433a04abe31e43226a6eed0c6125bcd2012147/test/extended/cli/explain.go#L476-L478

bparees: hrrrrrm.  yeah, the lack of a clusterversion resource, which strongly implies microshift, would imply potentially a different subset of rbac rules to be expected.  (e.g. microshift might have some "optional capabilities" always enabled, and some other ones always disabled, so just saying "if it's microshift, assume all optional caps are disabled" doesn't seem quite right to me).

the fact that the microshift team hadn't previously found the need to disable these checks for their CI would seem to further confirm that these rbac rules do exist on microshift clusters (though admittedly that seems slightly surprising).

@dhellmann can you weigh in on what rbac rules would/would not be expected on a microshift cluster, compared to a traditional OCP cluster with all capabilities enabled?

dhellmann: There's a ConfigMap that can be used to detect a MicroShift cluster, so we don't need to make any assumptions. https://github.com/openshift/microshift/blob/main/docs/debugging_tips.md#checking-the-microshift-version

I'm having a little trouble making sense of what the test is trying to check. I see the code below is looking at `rbacv1.PolicyRule`. Is that an API type? I do not see that present in MicroShift. Is that being translated into some other type somehow?

```
$ oc api-resources | grep -i rbac
clusterrolebindings                            rbac.authorization.k8s.io/v1           false        ClusterRoleBinding
clusterroles                                   rbac.authorization.k8s.io/v1           false        ClusterRole
rolebindings                                   rbac.authorization.k8s.io/v1           true         RoleBinding
roles                                          rbac.authorization.k8s.io/v1           true         Role
```

dhellmann: @ingvagabund or @pacevedom can you help here? Is this test even relevant to MicroShift?

bparees: Doug's going to have some microshift folks weigh in on this, but my current view would be that we do not want to comingle "skipping tests based on disabled caps" with "skipping tests based on being on microshift".  Until/unless microshift decides they need to disable this test (assuming they haven't already disabled it via some other skips mechanism elsewhere), i think we should just:

1) if the clusterversion resource type is undefined, test all the rbac rules (we can't make any assumptions about what caps are enabled/disabled)
2) if the clusterversion resource type is defined, fetch the CV instance and digest it and only check the rbac that makes sense based on which caps are enabled.  (if we fail to fetch the CV instance, fail the test)

if (1) turns out to cause failures in a microshift job somewhere, they can come back and decide how they want the test to run/skip in microshift environments (and how they want to detect that they are in a microshift env).  But since it doesn't seem to have been an issue so far, i'm not inclined to proactively do something here, especially not something that comingles "optional caps" behavior/assumptions with "microshift env" behavior/assumptions.

pacevedom: We have been changing tests so they fit both OpenShift and MicroShift platforms without explicitly knowing where they are running. Instead, tests rely on available resources to determine what they are able to do.
This test seems to do exactly that, extend the RBAC rules if the resources should exist.
This one is currently failing because of the `allAuthenticatedRules` contents, which are not following the pattern introduced here/described above and mix some API groups and resources not present in MicroShift. Ref: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-microshift-main-periodic-ocp-4.12-conformance-parallel/1618760836929556480

We also have a ticket for refactoring/rewrite the test in a new approach which allows each platform test what's needed.

bparees: > This one is currently failing because of the allAuthenticatedRules contents, which are not following the pattern introduced here/described above and mix some API groups and resources not present in MicroShift. Ref: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-microshift-main-periodic-ocp-4.12-conformance-parallel/1618760836929556480

If i understand you correctly, you're saying this test is already failing in the microshift job today, and changes are needed to fix that (though fixing that was not the original intent of this PR).

I still stand by what i said in my previous comment that we should be addressing the problem of "which optional caps are enabled" separately from "this is a microshift cluster".  And @dhellmann's response implied to me that "checking if the clusterversion api is defined" isn't the right way to detect a microshift cluster.

I suspect there are more rbac rules beyond just the "optional capability related" ones that this PR is adding skip-logic for, that will be missing on a microshift cluster and thus need additional logic to properly skip in a microshift environment.

tldr:
1) checking for the api isn't the preferred way to detect a microshift cluster (per doug)
2) the set of "optional cap rbac" is (likely?) not the same as the set of "non-microshift rbac" though there will be overlap

so i'd prefer to just have this PR handle the disabled optional caps rbac scenario and let the microshift team come back w/ how they want to handle the "running on microshift" scenario.  The microshift solution may build on this, but it would be best (imho) to make the two explicitly distinct, compared to the current implementation which is "if we appear to be running on microshift, implicitly treat that as if all optional caps are disabled"  (Which, again, also probably isn't sufficient to pass on microshift anyway since microshift will be missing more rbac than just the optional cap rbac)

pacevedom: > If i understand you correctly, you're saying this test is already failing in the microshift job today, and changes are needed to fix that (though fixing that was not the original intent of this PR).

Yes, this is one of the permafailing tests we have not yet fixed. We have an ongoing effort to greenify testgrid gradually, when feature development allows/requires.

> I still stand by what i said in my previous comment that we should be addressing the problem of "which optional caps are enabled" separately from "this is a microshift cluster". And @dhellmann's response implied to me that "checking if the clusterversion api is defined" isn't the right way to detect a microshift cluster.

Checking the configmap @dhellmann mentioned is deterministic because it only exists in MicroShift. However, when we started introducing MicroShift tests we agreed with @deads2k that we should not look for that explicitly, instead looking for resources, API groups, namespaces and any other resource that might help.

> I suspect there are more rbac rules beyond just the "optional capability related" ones that this PR is adding skip-logic for, that will be missing on a microshift cluster and thus need additional logic to properly skip in a microshift environment.

Yes, many of them are missing and the test makes it difficult to split/adapt. This  approach goes inline with the rest of the `origin` tests that were updated.

> tldr:
> 
> 1. checking for the api isn't the preferred way to detect a microshift cluster (per doug)

You are never sure if its MicroShift or not, you just have a guess. As mentioned above, this was agreed with David. For example, how do you know which tests apply to MicroShift? If you scatter conditions about "is this MicroShift?" in the test body you can't know until you execute.

> 2. the set of "optional cap rbac" is (likely?) not the same as the set of "non-microshift rbac" though there will be overlap

Yes, some other API groups are not present either.

> so i'd prefer to just have this PR handle the disabled optional caps rbac scenario and let the microshift team come back w/ how they want to handle the "running on microshift" scenario. The microshift solution may build on this, but it would be best (imho) to make the two explicitly distinct, compared to the current implementation which is "if we appear to be running on microshift, implicitly treat that as if all optional caps are disabled" (Which, again, also probably isn't sufficient to pass on microshift anyway since microshift will be missing more rbac than just the optional cap rbac)

This is not the first test we find with this problem, but the suggestions to fix it are inline with the others. Since we need to rework this one I would agree with these changes.

bparees: > Since we need to rework this one I would agree with these changes.

you agree with the changes as they currently exist in the PR, or you agree w/ the changes i proposed?

pacevedom: Sorry, I meant the ones you proposed. There is no need to check if `clusterversion` is there when the test is already broken by over 20 other resources permissions.
Even if the check was kept, the test would still need to be worked on to fit it into MicroShift.

ingvagabund: From the point of enabling running all e2e tests over MicroShift this is "just" another instance where we need to improve the logic. We will need to re-iterate on all affected tests again. So +1 for what @bparees suggests.

bparees: @wking are you able to move this forward?

PR #27697, opened 2023-02-01
extended: security: do not explicitly set api audience on token request

sjenning:
Hypershift does not use `kubernetes.default.svc` as the api audience on the KAS.  It is set to the URL of the OIDC provider.  ROSA also does this so I don't imagine this test passes for it either at the moment.

Explicit setting of the `Audiences` on the `TokenRequest` is not required.  If not set, it will just default to the audiences configured in the KAS.

Causing conformance failure for hypershift
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-hypershift-main-periodics-4.13-conformance-aws-ovn/1620240601058381824


sjenning:
cc @deads2k @stlaz @s-urbaniak @ingvagabund

dgoodwin:

This job hit two known issues, one for SLB disruption that we're about to disable, and one for loki image pull problems caused by registry sig cache issues.

PR #27702, opened 2023-02-02
[release-4.12] OCPBUGS-6841: Update etcd scaling test for CPMS supported platforms

Elbehery:
This is a cherry-pick of https://github.com/openshift/origin/pull/27497

cc @hasbro17

### DIFF
--- a/test/extended/etcd/helpers/helpers.go
+++ b/test/extended/etcd/helpers/helpers.go
@@ -3,6 +3,7 @@ package helpers
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -10,8 +11,10 @@ import (
 	o "github.com/onsi/gomega"
 
 	configv1 "github.com/openshift/api/config/v1"
+	machinev1 "github.com/openshift/api/machine/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	machineclient "github.com/openshift/client-go/machine/clientset/versioned"
+	machinev1client "github.com/openshift/client-go/machine/clientset/versioned/typed/machine/v1"
 	machinev1beta1client "github.com/openshift/client-go/machine/clientset/versioned/typed/machine/v1beta1"
 
 	bmhelper "github.com/openshift/origin/test/extended/baremetal"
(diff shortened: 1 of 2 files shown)


Elbehery:
`verify` job fails on `make update-gofmt` 

i ran it locally && `make verify` and nothing changed, no error !!

/retest-required

geliu2016 (approved):
/label cherry-pick-approved

test/extended/testdata/bindata.go
```diff
@@ -51457,13 +51457,11 @@ var _bindata = map[string]func() (*asset, error){
...
 // For example if you run go-bindata on data/... and data contains the
 // following hierarchy:
-//
-//	data/
-//	  foo.txt
-//	  img/
-//	    a.png
-//	    b.png
-//
+//     data/
```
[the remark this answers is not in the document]
Elbehery: i used go1.18, it is the version used in `go.mod` for `4.12` branch 

I will remove this 👍🏽

Elbehery:
@geliu2016  can we label this please 
@hasbro17  any job needs to be overriden ?

geliu2016 (approved):
/label cherry-pick-approved

hasbro17:
Test is passing on vsphere, the job is failing due to unrelated monitoring issues:
```
: [sig-instrumentation][Late] Alerts shouldn't report any unexpected alerts in firing or pending state [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel] expand_less | 32s
-- | --
{  fail [github.com/openshift/origin/test/extended/prometheus/prometheus.go:331]: unable to check firing alerts during test Unexpected error:     <*v1.Error \| 0xc005915500>: {         Type: "server_error",         Msg: "server error: 504",         Detail: "<html><body><h1>504 Gateway Time-out</h1>\nThe server didn't respond in time.\n</body></html>\n",     }     server_error: server error: 504 occurred Ginkgo exit error 1: exit with code 1}


```

/approve
/label backport-risk-assessed

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +183,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
 	})
 }
 
+func DeleteSingleMachine(ctx context.Context, t TestingT, machineClient machinev1beta1client.MachineInterface) (string, error) {
+	waitPollInterval := 15 * time.Second
+	waitPollTimeout := 5 * time.Minute
```
deads2k, line 188: you probably want to check this every 30s or so.  The load is minimal and it helps latency.  Just fixing in master is ok with me.

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +183,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
+	t.Logf("Waiting up to %s to delete a machine", waitPollTimeout.String())
+
+	machineToDelete := ""
+	err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		machineList, err := machineClient.List(ctx, metav1.ListOptions{LabelSelector: masterMachineLabelSelector})
+		if err != nil {
+			return isTransientAPIError(t, err)
+		}
+
+		// Machine names are suffixed with an index number (e.g "ci-op-xlbdrkvl-6a467-qcbkh-master-0")
```
deads2k, line 198: not all names are ordered by number, but ordering for consistency is still ok.

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +183,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
 }
 
+func DeleteSingleMachine(ctx context.Context, t TestingT, machineClient machinev1beta1client.MachineInterface) (string, error) {
+	waitPollInterval := 15 * time.Second
+	waitPollTimeout := 5 * time.Minute
+	t.Logf("Waiting up to %s to delete a machine", waitPollTimeout.String())
+
+	machineToDelete := ""
+	err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		machineList, err := machineClient.List(ctx, metav1.ListOptions{LabelSelector: masterMachineLabelSelector})
```
deads2k, line 193: Why does is this listing repeated instead of identifying the machine-to-be-deleted once and then retrying the Delete until you get a successful call?

test/extended/etcd/helpers/helpers.go
Elbehery: thanks @deads2k 

This is a backport, maybe @hasbro17  we need a new PR with these changes to 4.13 before backporting ?

Elbehery:
@deads2k can we marge this to be aligned with master and 4.13, then we make the required changes in a new PR against master then backport ?

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +183,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// EnsureReadyReplicasOnCPMS checks if status.readyReplicas on the cluster CPMS is n
+// this effectively counts the number of control-plane machines with the provider state as running
+func EnsureReadyReplicasOnCPMS(ctx context.Context, t TestingT, expectedReplicaCount int, cpmsClient machinev1client.ControlPlaneMachineSetInterface) error {
+	waitPollInterval := 5 * time.Second
```
deads2k, line 250: poll should be faster.

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +183,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
+		}
+		sort.Strings(machineNames)
+		machineToDelete = machineNames[0]
+		t.Logf("attempting to delete machine %q", machineToDelete)
+
+		if err := machineClient.Delete(ctx, machineToDelete, metav1.DeleteOptions{}); err != nil {
+			// The machine we just listed should be present but if not, error out
+			if apierrors.IsNotFound(err) {
+				t.Logf("machine %q was listed but not found or already deleted", machineToDelete)
+				return false, fmt.Errorf("machine %q was listed but not found or already deleted", machineToDelete)
```
deads2k, line 212: this error will exist the loop as an error, rigth?

test/extended/etcd/helpers/helpers.go
```diff
@@ -180,6 +183,91 @@ func recoverClusterToInitialStateIfNeeded(ctx context.Context, t TestingT, machi
...
+	waitPollTimeout := 18 * time.Minute
+	t.Logf("Waiting up to %s for the CPMS to have status.readyReplicas = %v", waitPollTimeout.String(), expectedReplicaCount)
+
+	return wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		cpms, err := cpmsClient.Get(ctx, "cluster", metav1.GetOptions{})
+		if err != nil {
+			return isTransientAPIError(t, err)
+		}
+
+		if cpms.Status.ReadyReplicas != int32(expectedReplicaCount) {
```
deads2k, line 260: do you care about this condition or do you care this AND that you have exactly the right number of total nodes AND that the desired replicas is also expectedReplicaCount?

hasbro17: We care about this condition because we want the test to ensure that a new machine has successfully been created and is ready, which means the associated has also been created and is running.
IIUC `ReadyReplicas` should cover both conditions:
- https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/controllers/controlplanemachineset/status.go#L95-L96
- https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/machineproviders/types.go#L38-L40

I think the desired `status.replicas` will only tell us if the machine was created, and not whether it became running with a node attached.

hasbro17:
@deads2k I think we can put up a new PR to address the review comments from here and make those changes in master since this is just a backport.
However I think merging this first is still useful in fixing the test where it's broken on the platforms that have CPMS enabled.

test/extended/etcd/helpers/helpers.go
hasbro17: Yes, it will break the polling loop and exit the function with an error.

test/extended/etcd/helpers/helpers.go
hasbro17: That's to retry the List if it errors out due to a transient API error. Would we not want to retry in that case?

deads2k: > That's to retry the List if it errors out due to a transient API error. Would we not want to retry in that case?

The client retries automatically.  Failures at this level aren't expected and even in controller code aren't generally handled.

test/extended/etcd/helpers/helpers.go
deads2k: > Yes, it will break the polling loop and exit the function with an error.

Since the item was deleted, why return the error?

test/extended/etcd/helpers/helpers.go
deads2k: if expected does not match the desired (spec), the return further down looks incorrect.

deads2k: also, this looks like the correct spot to double check that the number of nodes is actually correct.

damdo:
Looks like etcd vertical scaling for 4.12 is broken until this backport merges (which is a blocking job for the control-plane-machine-set-operator).
Their 4.14 and 4.13 counterparts are working well for us, it would be nice to get this one in too.
Thanks!

Elbehery:
Thanks @damdo for your review 👍🏽 

There is another on going PR https://github.com/openshift/origin/pull/27788

Elbehery:
resolved by https://github.com/openshift/origin/pull/27907 

/close

PR #27710, opened 2023-02-07
Alert Testing on new Namespace and Level

dgoodwin:
Remove use of default pi value in alert/disruption matchers, we don't use these default values on no match now, we skip the test.

Process new level and namespace in alert query_results.json, this came into ci-tools in: https://github.com/openshift/ci-tools/pull/3280

Add unit test for basic sanity check on alert query_results.json

Update alerts query_results.json manually for new data format. We need an updated file for this PR to work, so I manually generated and included. This will be the new format when the PR comes on Friday, assuming we get this merged in time.

Don't match historical data if under 100 JobRuns. Unlike disruption, I'm no longer filtering for at least 100 runs in the query, rather including everything and letting origin decide. This is because I expect to need to make some determinations for when new alerts fire soon. Both alerts and disruption in this PR now ignore entries in the file under 100 runs. I will update ci-tools to also include everything soon so everything is consistent.

[TRT-769](https://issues.redhat.com//browse/TRT-769)


pkg/synthetictests/allowedalerts/types.go
```diff
@@ -1,41 +1,12 @@
...
-  alertName = "KubeAPIErrorBudgetBurn" or 
-  alertName = "KubeClientErrors" or 
-  alertName = "KubePersistentVolumeErrors" or 
-  alertName = "MCDDrainError" or 
-  alertName = "PrometheusOperatorWatchErrors" or
-  alertName = "VSphereOpenshiftNodeHealthFail"
-order by 
- AlertName, Release, FromRelease, Topology, Platform, Network
-`
-)
```
dgoodwin, line 37: This query was removed, it's a duplicate copy of the one in ci-tools, and has surely drifted since being recorded here.

pkg/synthetictests/historicaldata/alert_types.go
```diff
@@ -0,0 +1,141 @@
...
+	AlertName      string
+	AlertNamespace string
+	AlertLevel     string
+
+	platformidentification.JobType `json:",inline"`
+}
+
+type AlertBestMatcher struct {
+	HistoricalData map[AlertDataKey]AlertStatisticalData
+}
```
dgoodwin, line 31: I tried multiple times to re-use the old system where both alerts and disruption were mapping to the same Go types, but I couldn't pull it off. After About 3 tries I gave up and forked the code. The BestMatchers remain using the common code, which I felt was the most important to keep shared between the two.

pkg/synthetictests/historicaldata/disruption_types.go
```diff
@@ -71,48 +65,46 @@ func NewMatcher(historicalJSON []byte, defaultReturn float64) (BestMatcher, erro
...
+func (b *DisruptionBestMatcher) bestMatch(name string, jobType platformidentification.JobType) (DisruptionStatisticalData, string, error) {
 	exactMatchKey := DataKey{
-		Name:    name,
-		JobType: jobType,
+		BackendName: name,
+		JobType:     jobType,
 	}
 
-	if percentiles, ok := b.historicalData[exactMatchKey]; ok {
+	if percentiles, ok := b.historicalData[exactMatchKey]; ok && percentiles.JobRuns > minJobRuns {
```
dgoodwin, line 93: Will start sending all data over for disruption in https://github.com/openshift/ci-tools/pull/3284, but first origin needs to know to ignore it if under 100 runs.

dgoodwin:
Required gcp jobs failed due to capacity.

dgoodwin:
Alert testing looks ok here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27710/pull-ci-openshift-origin-master-e2e-aws-ovn-single-node-upgrade/1623292473084219392

pkg/synthetictests/historicaldata/alert_types.go
DennisPeriquet: where is the "common code" that you refer to?  alert_types.go and disruption_types.go look almost identical (except for the types and names).

DennisPeriquet: ah.  I'm glad you didn't use generics for now.

pkg/synthetictests/allowedalerts/types.go
DennisPeriquet: I like the idea of removing this esp. if it's outdated.  I'm assuming you are referring to the query in [ci_data_client.go](https://github.com/openshift/ci-tools/blob/82c90e69750acdd9b37b766da424bc9e860d0277/pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go#L163-L170).

In place of it, can you please add some comment that mentions that `query_results.json` is created from big query and updated using automated PRs.

PR #27714, opened 2023-02-09
WRKLDS-605: e2e: Config v1 client shim for static configuration manifests with read-only operations

ingvagabund:
Implement a config client shim that allows to intersect requests from a client and replace requests over static manifests with read-only operation over the manifests. The request replacement is no-op for object outside the of static manifest directory. In case a mutation request is required over static manifests, an error is returned.

The static manifests directory is pointed to through `STATIC_CONFIG_MANIFEST_DIR` environment variable. Currently only manifests from [infrastructure|network].config.openshift.io/v1 are supported as static manifests.

Exemplary manifests for MicroShift deployment:
```yaml
apiVersion: "config.openshift.io/v1"
kind: Infrastructure
metadata:
  name: cluster
spec:
  platformSpec:
    type: None
status:
  apiServerURL: ???
  controlPlaneTopology: SingleReplica
  infrastructureTopology: SingleReplica
  platform: None
  platformStatus:
    type: None
```

```yaml
apiVersion: config.openshift.io/v1
kind: Network
metadata:
  name: cluster
spec:
  networkType: OVNKubernetes
status:
  networkType: OVNKubernetes
```


ingvagabund:
We can either have the envs as we have now. Or, create one huge json file passed through a single env variable.
Also, probably better to rename ClusterInfrastructure -> ClusterDetails.

test/extended/util/client.go
```diff
@@ -603,7 +622,13 @@ func (c *CLI) AdminBuildClient() buildv1client.Interface {
 }
 
 func (c *CLI) AdminConfigClient() configv1client.Interface {
-	return configv1client.NewForConfigOrDie(c.AdminConfig())
+	if len(c.staticConfigManifestDir) == 0 {
+		return configv1client.NewForConfigOrDie(c.AdminConfig())
+	} else {
+		// get all config objects from the directory
+		// SetupStaticConfigsFromManifests() needs to be invoked first
+		return fakeconfigv1client.NewSimpleClientset(c.configObjects...)
```
deads2k, line 630: you need a shim for this client that causes every mutating request to fail.

test/extended/util/framework.go
```diff
@@ -2075,3 +2083,53 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, groupVersionName
...
+func collectConfigManifestsFromDir(configManifestsDir string) (error, []runtime.Object) {
+	objects := []runtime.Object{}
+	files, err := ioutil.ReadDir(configManifestsDir)
+	if err != nil {
+		return err, nil
+	}
+
+	for _, file := range files {
+		if file.IsDir() {
+			klog.V(2).Infof("%q is not a file, skipping", file.Name())
```
deads2k, line 2096: I think recursion is expected.

test/extended/util/framework.go
```diff
@@ -2075,3 +2083,53 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, groupVersionName
...
+				objects = append(objects, config)
+			case "Network":
+				config := &configv1.Network{}
+				err = yaml.Unmarshal(body, &config)
+				if err != nil {
+					return err, nil
+				}
+				objects = append(objects, config)
+			default:
+				return fmt.Errorf("unknown 'config.openshift.io/v1' kind: %v", object.Kind), nil
```
deads2k, line 2128: excellent

test/extended/util/client.go
deads2k, line 630: reading below, while you're building the shim, can you actually make the fake client only intercept requests for config objects that it has data for?

test/extended/util/framework.go
```diff
@@ -2075,3 +2083,53 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, groupVersionName
...
+
+		object := &metav1.TypeMeta{}
+		err = yaml.Unmarshal(body, &object)
+		if err != nil {
+			return err, nil
+		}
+
+		if object.APIVersion == "config.openshift.io/v1" {
+			switch object.Kind {
+			case "Infrastructure":
```
deads2k, line 2113: need to fail if you find more than one of a given group,version,kind,namespace,name

ingvagabund:
Thanks for the comments. I am updating the code as I go.

test/extended/util/client.go
ingvagabund: Done

test/extended/util/framework.go
ingvagabund: Done

deads2k:
/approve

Approving in concept if someone gets time to review in depth tomorrow before I get back to it on Friday.  The layering here is nice.  I see no reason to implement for types we don't need yet and I don't think we'll get back an investment for a generator.  I would add them as we need them.

test/extended/util/client.go
ingvagabund: Done

test/extended/util/configv1shim_test.go
```diff
@@ -0,0 +1,276 @@
...
+func TestConfigClientShimErrorOnMutation(t *testing.T) {
+	updateNotPermitted := OperationNotPermitted{Action: "update"}
+	updatestatusNotPermitted := OperationNotPermitted{Action: "updatestatus"}
+	patchNotPermitted := OperationNotPermitted{Action: "patch"}
+	applyNotPermitted := OperationNotPermitted{Action: "apply"}
+	applyStatusNotPermitted := OperationNotPermitted{Action: "applystatus"}
+	deleteNotPermitted := OperationNotPermitted{Action: "delete"}
+	deleteCollectionNotPermitted := OperationNotPermitted{Action: "deletecollection"}
+
+	object := createInfrastructureObject("cluster")
```
atiratree, line 63: maybe name them staticObject and realObject/dynamicObject? or something else for better recognition?

test/extended/util/configv1shim_test.go
```diff
@@ -0,0 +1,276 @@
...
+					diff := cmp.Diff(test.expectedWatchEvents[i], item)
+					if diff != "" {
+						t.Errorf("test '%s' failed. Results are not deep equal. mismatch (-want +got):\n%s", test.name, diff)
+					}
+					eventCounter++
+				case <-ticker.C:
+					t.Errorf("failed waiting for watch event")
+				}
+			}
+			if eventCounter < size {
```
atiratree, line 270: nit: this could also test if we got more events than expectedWatchEvents

nice fake watcher though

test/extended/util/client.go
```diff
@@ -608,7 +626,10 @@ func (c *CLI) AdminBuildClient() buildv1client.Interface {
 }
 
 func (c *CLI) AdminConfigClient() configv1client.Interface {
```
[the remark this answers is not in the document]
ingvagabund: Good point. The right course of actions is to replace the `oc get` with the code equivalent. I don't think this can be easily detected without grepping/taking a look.

test/extended/util/configv1shim_test.go
```diff
@@ -0,0 +1,276 @@
...
+		configClient,
+		[]runtime.Object{object},
+	)
+
+	_, err := client.ConfigV1().Infrastructures().Get(context.TODO(), object.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Expected no error for a Get request, got %q instead", err)
+	}
+
+	_, err = client.ConfigV1().Infrastructures().List(context.TODO(), metav1.ListOptions{})
```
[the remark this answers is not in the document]
ingvagabund: This is suitable for a separate test

test/extended/util/client.go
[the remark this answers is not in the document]
atiratree: fully replacing the logic of `oc get` with all the options is not a simple task and might get too complex. Can we opt out for something simpler like injecting forbidden groups into the CLI struct? And then just going over `commandArgs` and panic if we find these groups.

I think this might be sufficient since I am not seeing anyone using `oc` with `config.openshift.io` group. If somebody starts using it, we can either force them to use the go client or implement the oc mock logic later.

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+	}
+	if apierrors.IsNotFound(err) {
+		return c.configV1InfrastructuresClient.Patch(ctx, name, pt, data, opts, subresources...)
+	}
+	return nil, err
+}
+
+func (c *ConfigV1InfrastructuresClientShim) Apply(ctx context.Context, infrastructure *applyconfigv1.InfrastructureApplyConfiguration, opts metav1.ApplyOptions) (*apiconfigv1.Infrastructure, error) {
+	// Unable to determine existence of a static manifest
+	if infrastructure == nil || infrastructure.Name == nil {
```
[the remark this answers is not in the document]
ingvagabund: Not sure whether it's a valid use case to pass empty InfrastructureApplyConfiguration or empty Name. The opts does not have any means to point to a target object. So I presume the apply is a no-op when the object is not specified. Checking the Apply implementation:
```go

```func (c *infrastructures) Apply(ctx context.Context, infrastructure *configv1.InfrastructureApplyConfiguration, opts metav1.ApplyOptions) (result *v1.Infrastructure, err error) {
	if infrastructure == nil {
		return nil, fmt.Errorf("infrastructure provided to Apply must not be nil")
	}
	patchOpts := opts.ToPatchOptions()
	data, err := json.Marshal(infrastructure)
	if err != nil {
		return nil, err
	}
	name := infrastructure.Name
	if name == nil {
		return nil, fmt.Errorf("infrastructure.Name must be provided to Apply")
	}
	...

File: test/extended/util/configv1shim.go [hunk 10]
```diff
@@ -0,0 +1,722 @@
...
+
+	items := []apiconfigv1.Infrastructure{}
+	knownKeys := make(map[string]struct{})
+	for _, item := range staticObjList.Items {
+		items = append(items, item)
+		knownKeys[item.Name] = struct{}{}
+	}
+
+	for _, item := range objList.Items {
+		// skip objects with corresponding static manifests
```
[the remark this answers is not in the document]
ingvagabund: Yes. The static manifests takes precedence over the existing resources.

test/extended/util/framework.go
```diff
@@ -2075,3 +2082,69 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, groupVersionName
...
+
+	err := filepath.Walk(configManifestsDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		body, err := ioutil.ReadFile(path)
```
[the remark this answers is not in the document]
ingvagabund: Done

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+
+func (c *ConfigV1ClientShim) Infrastructures() configv1.InfrastructureInterface {
+	return &ConfigV1InfrastructuresClientShim{
+		fakeConfigV1InfrastructuresClient: c.fakeConfigV1Client.Infrastructures(),
+		configV1InfrastructuresClient:     c.configv1.Infrastructures(),
+	}
+}
+
+func (c *ConfigV1ClientShim) Ingresses() configv1.IngressInterface {
+	if c.v1Kinds["Ingresse"] {
```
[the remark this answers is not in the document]
ingvagabund: Renamed

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+
+func (c *ConfigV1ClientShim) Images() configv1.ImageInterface {
+	if c.v1Kinds["Image"] {
+		panic(fmt.Errorf("Image not implemented"))
+	}
+	return c.configv1.Images()
+}
+
+func (c *ConfigV1ClientShim) ImageContentPolicies() configv1.ImageContentPolicyInterface {
+	if c.v1Kinds["ImageContentPolicie"] {
```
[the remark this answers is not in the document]
ingvagabund: Renamed

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+
+func (c *ConfigV1ClientShim) Projects() configv1.ProjectInterface {
+	if c.v1Kinds["Project"] {
+		panic(fmt.Errorf("Project not implemented"))
+	}
+	return c.configv1.Projects()
+}
+
+func (c *ConfigV1ClientShim) Proxies() configv1.ProxyInterface {
+	if c.v1Kinds["Proxie"] {
```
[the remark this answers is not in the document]
ingvagabund: Renamed

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+		ListMeta: objList.ListMeta,
+		Items:    items,
+	}, nil
+}
+
+func (c *ConfigV1InfrastructuresClientShim) Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
+	// static manifests do not produce any watch event besides create
+	// If the object exists, no need to generate the ADDED watch event
+
+	// INFO: field selectors will not work
```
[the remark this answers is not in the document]
ingvagabund: Implemented

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+		return obj, nil
+	}
+	if apierrors.IsNotFound(err) {
+		return c.configV1InfrastructuresClient.Get(ctx, name, opts)
+	}
+	return nil, err
+}
+
+func (c *ConfigV1InfrastructuresClientShim) List(ctx context.Context, opts metav1.ListOptions) (*apiconfigv1.InfrastructureList, error) {
+	// INFO: field selectors will not work
```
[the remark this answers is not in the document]
ingvagabund: Implemented

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+	if err == nil {
+		return nil, &OperationNotPermitted{Action: "patch"}
+	}
+	if apierrors.IsNotFound(err) {
+		return c.configV1NetworksClient.Patch(ctx, name, pt, data, opts, subresources...)
+	}
+	return nil, err
+}
+
+func (c *ConfigV1NetworksClientShim) Apply(ctx context.Context, infrastructure *applyconfigv1.NetworkApplyConfiguration, opts metav1.ApplyOptions) (*apiconfigv1.Network, error) {
```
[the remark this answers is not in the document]
ingvagabund: Renamed

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+	if err == nil {
+		return nil, &OperationNotPermitted{Action: "apply"}
+	}
+	if apierrors.IsNotFound(err) {
+		return c.configV1NetworksClient.Apply(ctx, infrastructure, opts)
+	}
+	return nil, err
+}
+
+func (c *ConfigV1NetworksClientShim) ApplyStatus(ctx context.Context, infrastructure *applyconfigv1.NetworkApplyConfiguration, opts metav1.ApplyOptions) (*apiconfigv1.Network, error) {
```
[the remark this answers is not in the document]
ingvagabund: Renamed

test/extended/util/configv1shim_test.go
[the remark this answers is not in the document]
ingvagabund: Test implemented

test/extended/util/configv1shim_test.go
ingvagabund: Done

test/extended/util/configv1shim_test.go
ingvagabund: Done

test/extended/util/client.go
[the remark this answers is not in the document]
ingvagabund: > And then just going over commandArgs and panic if we find these groups.

Not every kind is going to be explicitly typed with the config.openshift.io group. In which case it's up to the KA to decide which group takes precedence.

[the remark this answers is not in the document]
atiratree: yes, in general it would get more complicated. I was looking for more of an temporary bandaid. Nevertheless this PR is pretty large already. Maybe the best would be to pursue this as a followup?

test/extended/util/configv1shim.go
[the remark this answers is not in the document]
atiratree: I understand, but it is a responsibility of the person who runs the test to supply valid supported objects, so I think it does not have to fail in the test, and can fail in the setup instead

Nevertheless, the current solution is okay as well

test/extended/util/configv1shim.go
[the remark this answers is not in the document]
atiratree: Thanks for making the list operations compliant.

Although eyes hurt much more after looking at all the logic duplication. Could we do at least some light refactoring? For example, for the logic around the selectors?

test/extended/util/configv1shim.go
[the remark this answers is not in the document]
atiratree: TY. No need for the comment now.

test/extended/util/configv1shim.go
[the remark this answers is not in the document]
atiratree: ok, thanks for checking

test/extended/util/configv1shim.go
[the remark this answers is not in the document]
atiratree: Thx, although I am still seeing some mistakes in other functions of `ConfigV1NetworksClientShim`

test/extended/util/configv1shim_test.go
```diff
@@ -0,0 +1,569 @@
...
+					diff := cmp.Diff(test.expectedWatchEvents[i], item)
+					if diff != "" {
+						t.Errorf("test '%s' failed. Results are not deep equal. mismatch (-want +got):\n%s", test.name, diff)
+					}
+					eventCounter++
+				case <-ticker.C:
+					t.Errorf("failed waiting for watch event")
+				}
+			}
+			if eventCounter != size {
```
atiratree, line 407: since we poll up to size at maximum we will not fail if we get more events than expected. This probably needs at least one extra poll from the ResultChan to detect such bug.

test/extended/util/configv1shim_test.go
```diff
@@ -0,0 +1,569 @@
...
+	}
+
+	if len(listItems.Items) != 2 {
+		t.Fatalf("Expected only a single item in the list, got %v instead", len(listItems.Items))
+	}
+
+	var staticObj *configv1.Infrastructure
+
+	for _, item := range listItems.Items {
+		if item.Name == "staticObject" {
```
atiratree, line 447: might be good to also check that the 2nd returned object is the real one (`realObject`)

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,722 @@
...
+// manifest present for a config v1 kind, fake client is used
+// instead of the real one.
+type ConfigClientShim struct {
+	configClient configv1client.Interface
+	v1Kinds      map[string]bool
+	fakeClient   *fakeconfigv1client.Clientset
+}
+
+func (c *ConfigClientShim) Discovery() discovery.DiscoveryInterface {
+	return c.configClient.Discovery()
```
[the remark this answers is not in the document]
ingvagabund: Implemented up to `OpenAPISchema()` method. Having it panic when invoked for now.

test/extended/util/client.go
```diff
@@ -79,23 +79,26 @@ import (
...
-	stderr             io.Writer
-	verbose            bool
-	withoutNamespace   bool
-	kubeFramework      *framework.Framework
-
+	execPath                string
+	verb                    string
+	configPath              string
+	adminConfigPath         string
+	staticConfigManifestDir string
```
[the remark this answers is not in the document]
ingvagabund: Done

test/extended/util/configv1shim_test.go
ingvagabund: Done

test/extended/util/client.go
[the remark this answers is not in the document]
ingvagabund: This will require further analysis. Nevertheless, wherever possibly I prefer to replace `oc` CLI commands with their config client equivalents when accessing the config API. Let's make it on per case basis.

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,1022 @@
...
+		Version:      configVersion,
+	}
+}
+
+func mergeResources(aList, bList []metav1.APIResource) []metav1.APIResource {
+	knownKinds := make(map[string]struct{})
+	for _, item := range aList {
+		knownKinds[item.Kind] = struct{}{}
+	}
+	resources := append([]metav1.APIResource{}, bList...)
```
atiratree, line 68: should be `aList`? Maybe we need a small test for the discovery..

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,1022 @@
...
+		ListMeta: objList.ListMeta,
+		Items:    items,
+	}, nil
+}
+
+func (c *ConfigV1InfrastructuresClientShim) Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
+	// static manifests do not produce any watch event besides create
+	// If the object exists, no need to generate the ADDED watch event
+
+	// INFO: field selectors will not work
```
atiratree, line 566: comment can be removed now

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,1022 @@
...
+		ListMeta: objList.ListMeta,
+		Items:    items,
+	}, nil
+}
+
+func (c *ConfigV1NetworksClientShim) Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
+	// static manifests do not produce any watch event besides create
+	// If the object exists, no need to generate the ADDED watch event
+
+	// INFO: field selectors will not work
```
atiratree, line 798: here as well

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,1022 @@
...
+}
+
+type ConfigV1NetworksClientShim struct {
+	fakeConfigV1NetworksClient configv1.NetworkInterface
+	configV1NetworksClient     configv1.NetworkInterface
+}
+
+var _ configv1.NetworkInterface = &ConfigV1NetworksClientShim{}
+
+func (c *ConfigV1NetworksClientShim) Create(ctx context.Context, infrastructure *apiconfigv1.Network, opts metav1.CreateOptions) (*apiconfigv1.Network, error) {
```
atiratree, line 667: nit: still wrong variable names (`infrastructure`)

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,1022 @@
...
+	return realList, err
+}
+
+func (c *ConfigV1DiscoveryClientShim) ServerGroupsAndResources() ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
+	groups, resources, err := c.configClient.Discovery().ServerGroupsAndResources()
+	if err != nil {
+		return groups, resources, err
+	}
+	if groups != nil {
+		hasConfigGroup := false
```
atiratree, line 158: can we extract this block into a function? as it is used also in `ServerGroups()`

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,1022 @@
...
+					}
+				}
+				if !hasV1Version {
+					groups[i].Versions = append(groups[i].Versions, configGroupVersionForDiscovery())
+				}
+				break
+			}
+		}
+		if !hasConfigGroup {
+			groups = append(groups, configAPIGroup())
```
atiratree, line 176: why is this one pointer? (in the duplicate code it is not)

test/extended/util/configv1shim.go
```diff
@@ -0,0 +1,1022 @@
...
+	return c.configClient.Discovery().ServerVersion()
+}
+
+func (c *ConfigV1DiscoveryClientShim) OpenAPISchema() (*openapi_v2.Document, error) {
+	// TODO(jchaloup): once needed implement this method as well
+	panic(fmt.Errorf("APIServer not implemented"))
+}
+
+func (c *ConfigV1DiscoveryClientShim) OpenAPIV3() openapi.Client {
+	return c.configClient.Discovery().OpenAPIV3()
```
atiratree, line 240: not sure if we shouldn't panic in these last 3 as well

test/extended/util/configv1shim.go
atiratree: and in other methods as well

test/extended/util/configv1shim_test.go
```diff
@@ -0,0 +1,580 @@
...
+
+			// And only then the fake clientset can be populated to generate the watch event
+			for _, obj := range test.realObjects {
+				configClient.Tracker().Add(obj)
+			}
+
+			// verify the watch events
+			ticker := time.NewTicker(500 * time.Millisecond)
+			size := len(test.expectedWatchEvents)
+			eventsSize := len(resultChan.ResultChan())
```
atiratree, line 394: this seems potentially racy, because of the processing loop. Maybe we can check that the len of the channel is 0 after the loop instead?

test/extended/util/configv1shim.go
ingvagabund: `ServerGroupsAndResources` returns `[]*metav1.APIGroup` whereas `ServerGroups` `*metav1.APIGroupList` (`[]metav1.APIGroup` instead). So the code is not totally identical. Also, the code is directly changing the list. It's more practical to inline the code twice instead.

test/extended/util/configv1shim.go
ingvagabund: `ServerGroupsAndResources` returns `[]*metav1.APIGroup` whereas `ServerGroups` `*metav1.APIGroupList` (`[]metav1.APIGroup` instead)

test/extended/util/configv1shim.go
ingvagabund: Done

test/extended/util/configv1shim.go
ingvagabund: Done

test/extended/util/configv1shim.go
ingvagabund: Done

test/extended/util/configv1shim.go
ingvagabund: Done

test/extended/util/configv1shim_test.go
ingvagabund: Checking if there's another event in the channel at the end of the test

test/extended/util/configv1shim.go
atiratree: we could make a generic function, and insert the modifications after that, but since we are trying to aim for simple shim at first, we can do that later once a need for more comprehensive refactoring  arises

test/extended/util/configv1shim.go
atiratree: thanks for adding the tests

atiratree:
@ingvagabund verify failing

test/extended/util/configv1shim.go
```diff
@@ -212,11 +216,13 @@ func (c *ConfigV1DiscoveryClientShim) ServerPreferredResources() ([]*metav1.APIR
 	for i, resource := range resources {
 		if resource.GroupVersion == configGroupVersion {
 			hasConfigGroup = true
-			resources[i].APIResources = mergeResources(fakeList.APIResources, resources[i].APIResources)
```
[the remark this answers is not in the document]
ingvagabund: Good point. Updated and returning ealy.

test/extended/util/client.go
```diff
@@ -563,6 +577,14 @@ func (c *CLI) RESTMapper() meta.RESTMapper {
 	return ret
 }
 
+func (c *CLI) setupStaticConfigsFromManifests() {
+	if len(c.staticConfigManifestDir) > 0 {
+		err, objects := collectConfigManifestsFromDir(c.staticConfigManifestDir)
+		o.Expect(err).ToNot(o.HaveOccurred())
```
pacevedom, line 583: There is a bug here.

We start with a normal dry run execution to get the test list.
```bash
$ ./openshift-tests run openshift/conformance --provider none -v 2 --dry-run  | wc -l
Unable to get server version through discovery client, skipping apigroup check in the dry-run mode: the server has asked for the client to provide credentials
3437
```
Then we export the variable
```bash
$ export STATIC_CONFIG_MANIFEST_DIR=$(pwd)/tmp
$ ls -l $STATIC_CONFIG_MANIFEST_DIR 
total 8
-rw-rw-r--. 1 pacevedo pacevedo 310 Mar 21 23:10 infra.yaml
-rw-rw-r--. 1 pacevedo pacevedo 148 Mar 21 23:10 net.yaml
$ ./openshift-tests run openshift/conformance --provider none -v 2 --dry-run  | wc -l
Unable to get server version through discovery client, skipping apigroup check in the dry-run mode: the server has asked for the client to provide credentials
213
```
There is a huge difference in the amount of tests that will be executed, and it only happens after exporting the variable, all other things equal.
The error comes from this specific line. The Expect call will check if gomega was [configured](https://github.com/openshift/origin/blob/master/vendor/github.com/onsi/gomega/gomega_dsl.go#L221).
Which checks the [configuration for the default gomega instance](https://github.com/openshift/origin/blob/master/vendor/github.com/onsi/gomega/gomega_dsl.go#L168), which is what we are using.
Which checks the [presence of special handlers](https://github.com/openshift/origin/blob/master/vendor/github.com/onsi/gomega/internal/gomega.go#L26), which are [not configured](https://github.com/openshift/origin/blob/master/vendor/github.com/onsi/gomega/internal/gomega.go#L18-L22) in the [default instance](https://github.com/openshift/origin/blob/master/vendor/github.com/onsi/gomega/gomega_dsl.go#L47).

The `run-test` command does [configure a fail handler](https://github.com/openshift/origin/blob/master/cmd/openshift-tests/openshift-tests.go#L488), for example.

pacevedom:
@ingvagabund before merging this, can we fix the bug I pointed? This will break nightly MicroShift conformance test runs.

test/extended/util/client.go
ingvagabund: My bad. I was checking error with gingko.Expect before it was properly initialized.

ingvagabund:
The last `e2e-gcp-ovn` failed with `[sig-network][Feature:EgressFirewall] when using openshift ovn-kubernetes should ensure egressfirewall is created [Suite:openshift/conformance/parallel]`. The one before with `[sig-node] static pods should start after being created`. The one before succeeded. The job is flaky.

atiratree:
failure due to https://github.com/openshift/release/pull/37909 - should be ok now
/retest-required

PR #27719, opened 2023-02-10
Chart CI cluster DNS problems in different color from disruption

dgoodwin:
[TRT-847](https://issues.redhat.com//browse/TRT-847)

Light blue for CI cluster disruption, continue with red for disruption.

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -283,6 +283,17 @@
         return [item.locator, "", "AlertCritical"]
     }
 
+    function disruptionValue(item) {
+        // We classify these disruption samples with this message if it thinks
+        // it looks like a problem in the CI cluster running the tests, not the cluster under test.
+        // (typically DNS lookup problems)
+        let ciClusterDisruption = item.message.indexOf("likely a problem in cluster running tests")
+        if (ciClusterDisruption != -1) {
+            return [item.locator, "", "CIClusterDisruption"]
+        }
+        return [item.locator, "", "Disruption"]
+    }
+
     function getDurationString(durationSeconds) {
         const seconds = durationSeconds % 60;
         const minutes = Math.floor(durationSeconds/60);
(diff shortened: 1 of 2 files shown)


stbenjam:
/lgtm

Is there a job we can test to make sure it works?

dgoodwin:
Not exactly, the CI cluster problem is not all the time, something like 15% of runs or so? Might need to let this one go out, but I can check and see what we get here, as long as we still see our disruption charted in some color we'll be good.

/hold

e2echart/e2e-chart-template.html
```diff
@@ -400,7 +411,7 @@ <h5 class="modal-title">Resource</h5>
         })
 
         timelineGroups.push({group: "endpoint-availability", data: []})
-        createTimelineData("Failed", timelineGroups[timelineGroups.length - 1].data, eventIntervals, isEndpointConnectivity, regex)
+        createTimelineData(disruptionValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isEndpointConnectivity, regex)
```
neisw, line 415: That's tricky right there passing in either a string or a function...  No issues, just made me go look.

dgoodwin: Yup made me look as well, turns out the function supports both. Good ol Javascript.

dgoodwin:
No sign of the problem in the results I checked, but we are still getting our graphs normally, should be safe to merge.

/hold cancel

dgoodwin:
Needed a bindata update, apologies. Needs another lgtm.

DennisPeriquet:
You can easily test these changes with [this job](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-origin-27694-ci-4.13-e2e-azure-ovn-upgrade/1623906725591519232) as mentioned in the description of [TRT-849](https://issues.redhat.com/browse/TRT-849).

Just need to swap in your html content with the one in the job.  So, ...

this:
Screen Shot 2023-02-13 at 4 57 06 PM

becomes this:Screen Shot 2023-02-13 at 4 56 37 PM

which looks good.

PR #27724, opened 2023-02-14
Port alert backstop test to invariant, allow running alert/disruption invariants locally for developers

dgoodwin:
This PR is another step in the work towards getting alert regression testing operational.  It does two main things:

(1) Create a copy of the late ginko alert test, instead as an invariant test. All the per-alert testing is done as invariants, moving this to an invariant allows for better code re-use and testability, and they now run closer in time than a late test vs invariant. The ginko version will eventually be removed, but I have not taken it out yet as I want to compare results.

(2) A significant amount of refactoring has been done to let the alert invariants run, without a live kube cluster. The goal was just to be able to run a new openshift-tests subcommand (see dev.go), specify your NURP details as CLI arguments, specify an e2e-events.json data file from a job run, and run all the alert or disruption invariant tests against that data. This is extremely useful for those working on this very complicated code, as we no longer have to wait hours for turnaround on our data.

Alert testing at this point is still quite broken, but all alert tests are moved to flake for some time now. More work is coming to stabilize these but I wanted to get this refactor in as it's getting quite large. The next phase will be all about updating the tests to accomplish the goals of this whole effort.

[TRT-833](https://issues.redhat.com//browse/TRT-833)


dgoodwin:
/hold

Depends on and includes #27742, if there are changes there I may have to force push here so this may not be save to build upon. Alternatively, I could do separate commits for followups over there.

@deepsm007 I don't know when I can merge this, but you could use this branch to run your disruption test I think. Just get e2e-events.json from one of those really bad jobs we looked at last week, I have some locally if you lost it but:

```
$ make build && ./openshift-tests dev run-disruption-invariants --intervals-file e2e-events-bad-disruption.json --platform aws --network sdn
```

This will run all the synthetic tests, which you'd need to add your new test to dev.go testCases, as well as the real location. (event_junits.go)

pkg/alerts/allowed_conformance.go
```diff
@@ -1,51 +1,17 @@
...
-			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=1953798",
-			Matches: func(_ *model.Sample) bool {
-				return framework.ProviderIs("gce")
-			},
-		},
-		{
-			Selector: map[string]string{"alertname": "KubeJobFailed", "namespace": "openshift-multus"}, // not sure how to do a job_name prefix
-			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=2054426",
-		},
-	}
```
dgoodwin, line 48: These were all ancient and should have been removed after 4.8 released.

pkg/alerts/allowed_upgrade.go
```diff
@@ -1,96 +1,17 @@
...
-		},
-		{
-			Selector: map[string]string{"alertname": "KubePodNotReady", "namespace": "openshift-kube-apiserver-operator"},
-			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=1939580",
-		},
-		{
-			Selector: map[string]string{"alertname": "KubePodNotReady", "namespace": "openshift-kube-apiserver-operator"},
-			Text:     "https://bugzilla.redhat.com/show_bug.cgi?id=1939580",
-		},
-	}
```
dgoodwin, line 93: Per above, all ancient and should have been removed long ago.

pkg/alerts/check.go
```diff
@@ -7,26 +7,47 @@ import (
...
 )
 
-type allowedAlertsFunc func(configclient.Interface) (allowedFiringWithBugs, allowedFiring, allowedPendingWithBugs, allowedPending helper.MetricConditions)
+type AllowedAlertsFunc func(featureSet configv1.FeatureSet) (allowedFiringWithBugs, allowedFiring, allowedPendingWithBugs, allowedPending helper.MetricConditions)
 
 // CheckAlerts will query prometheus and ensure no-unexpected alerts were pending or firing.
-// Used both post-upgrade and post-conformance, with different allowances for each.
-func CheckAlerts(allowancesFunc allowedAlertsFunc, prometheusClient prometheusv1.API, configClient configclient.Interface, testDuration time.Duration, f *framework.Framework) {
+// Used by both upgrade and conformance suites, with different allowances for each.
+func CheckAlerts(allowancesFunc AllowedAlertsFunc,
```
dgoodwin, line 28: This function is used by both ginko late tests (one for upgrade, one for conformance), this is the code that was ported to an invariant, and this version will *eventually* be removed.

pkg/synthetictests/alerts.go
```diff
@@ -32,5 +90,121 @@ func testAlerts(events monitorapi.Intervals, restConfig *rest.Config, duration t
...
+	return ret
+}
+
+// runBackstopTest will process the intervals for any alerts which do not have their own explicit test,
+// and look for any pending/firing intervals that are not within sufficient range.
+func runBackstopTest(
+	allowancesFunc alerts.AllowedAlertsFunc,
+	featureSet configv1.FeatureSet,
+	alertIntervals monitorapi.Intervals,
+	alertTests []allowedalerts.AlertTest) []*junitapi.JUnitTestCase {
```
dgoodwin, line 105: The core of the new invariant backstop test that checks all alerts that don't have their own explicit test. Above function is just for the logic around a live cluster, this is the one we can use to run locally and pass in our inputs.

pkg/synthetictests/allowedalerts/job_run_data.go
```diff
@@ -25,7 +25,7 @@ func WriteAlertDataForJobRun(artifactDir string, _ monitorapi.ResourcesMap, even
```
neisw, line 25: Not really related to your current change but more of a long term plan question (which I believe you are working on).  Can we work to at least co-locate [AlertKey](https://github.com/neisw/origin/blob/ebf4f252b21666844904e021e7feb6290e9655c1/pkg/synthetictests/allowedalerts/job_run_data.go#L54) , [AlertState](https://github.com/openshift/origin/blob/ebf4f252b21666844904e021e7feb6290e9655c1/pkg/synthetictests/allowedalerts/basic_alert.go#L37) along with [AlertLevel](https://github.com/openshift/origin/blob/ebf4f252b21666844904e021e7feb6290e9655c1/pkg/synthetictests/allowedalerts/job_run_data.go#L81) and comment on how each is 'different' such that we require separate types.  Preferably in [alert_types](https://github.com/openshift/origin/blob/ebf4f252b21666844904e021e7feb6290e9655c1/pkg/synthetictests/historicaldata/alert_types.go#L21)

neisw: Might be too much of a reach to try to tie in a common type to [prometheus/helpers](https://github.com/openshift/origin/blob/1e39f421a95fc8510dfc25780fe3cbfccf7c0b26/test/extended/util/prometheus/helpers.go#L135) and [AlertDataKey](https://github.com/openshift/origin/blob/ebf4f252b21666844904e021e7feb6290e9655c1/pkg/synthetictests/historicaldata/alert_types.go#L21)

Just seems like so much commonality dispersed throughout the code.

dgoodwin: Mmm MetricCondition is for whitelisting alerts, which is also used for any alert that got it's own test. AlertLevel is largely going away on both this MetricCondition and on AlertDataKey, alerts only have one Level, we don't need to key on it. That will help trim down a little. 

I'd like to instead just iterate all the per alert tests and say "hey do you match this interval", but we'd still need MetricCondition for explicitly whitelisting alerts that are allowed to happen. I think I can improve this in round three though, let me give it some thought and work on it there.

neisw: That's all I was going for was what can be removed / consolidated in the future.

Name & Namespace seem to be paired together throughout.  Level & State look very similar.  Good to hear AlertLevel is going away, that should tighten things up some.

dgoodwin:
/hold cancel

Thanks Forrest! Incoming @deepsm007

PR #27725, opened 2023-02-14
Add previous pathological events to spyglass charts

DennisPeriquet:
TRT-841 (but this is a bit of a draft; I'm just testing some assumptions)


pkg/monitor/event.go
```diff
@@ -18,8 +18,10 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+var ReMatchFirstQuote = regexp.MustCompile(`"([^"]+)"( in (\d+(\.\d+)?(s|ms)$))?`)
```
dgoodwin, line 21: Looks duplicated at the package level and then redeclared below in startEventMonitoring?

pkg/monitor/event.go
```diff
@@ -18,8 +18,10 @@ import (
...
+
 func startEventMonitoring(ctx context.Context, m Recorder, client kubernetes.Interface) {
-	reMatchFirstQuote := regexp.MustCompile(`"([^"]+)"( in (\d+(\.\d+)?(s|ms)$))?`)
+	ReMatchFirstQuote := regexp.MustCompile(`"([^"]+)"( in (\d+(\.\d+)?(s|ms)$))?`)
```
dgoodwin, line 24: Could you document what this is intended to match with an example. Little tough to parse.

pkg/monitor/event.go
```diff
@@ -259,7 +267,22 @@ func recordAddOrUpdateEvent(
...
+			Message   string
+			Count     int32
+		}{
+			Name:      obj.InvolvedObject.Name,
+			Namespace: obj.Namespace,
+			Reason:    obj.Reason,
+			Message:   obj.Message,
+			Count:     obj.Count,
+		}
+		fmt.Printf("processed event: %+v\nresulting new interval: %s from: %s to %s\n", summaryObj, condition.Message, pathoFrom, to)
```
dgoodwin, line 284: Suggest not declaring a struct just for the purposes of logging with %v, I'd just format the string as you wish with %s and additional params. I don't know for sure it's any kind of performance hit it just isn't something I've seen done before.

pkg/monitor/event_test.go
```diff
@@ -103,7 +104,7 @@ func Test_recordAddOrUpdateEvent(t *testing.T) {
 		}
 		t.Run(tt.name, func(t *testing.T) {
 			for _, event := range tt.args.kubeEventList.Items {
-				recordAddOrUpdateEvent(tt.args.ctx, tt.args.m, tt.args.client, tt.args.reMatchFirstQuote, tt.args.significantlyBeforeNow, &event)
+				RecordAddOrUpdateEvent(tt.args.ctx, tt.args.m, tt.args.client, tt.args.reMatchFirstQuote, tt.args.significantlyBeforeNow, &event, duplicateevents.DuplicateEventThreshold)
```
dgoodwin, line 107: Never noticed before but does this test not assert anything? Just checks that it runs and presumably a human is watching the console? Could we get some assertions in here so we get notified if something breaks this.

pkg/synthetictests/duplicated_events.go
```diff
@@ -199,13 +237,94 @@ func (d duplicateEventsEvaluator) testDuplicatedEvents(testName string, flakeOnl
...
+						coreEventFrom := coreEvent.LastTimestamp.Time
+						if coreEventFrom.IsZero() {
+							coreEventFrom = coreEvent.EventTime.Time
+						}
+						if coreEventFrom.IsZero() {
+							coreEventFrom = coreEvent.CreationTimestamp.Time
+						}
+						coreEventTo := coreEventFrom.Add(1 * time.Second)
+
+						fmt.Println("Counter: ", i)
```
dgoodwin, line 290: I'm starting to lean towards use of logrus, I don't dare try in the sub-test processes because we explicitly try to prevent logging from current tests overlapping, but for logging we know is in the main openshift-tests process it should be safe and would be helpful to see contextual logs and consistent timestamps. I'd suggest logrus.Debug and logrus.Info used here as appropriate given the amount of logging you have going on here.

pkg/test/ginkgo/options_monitor_events.go
```diff
@@ -155,6 +155,24 @@ func (o *MonitorEventsOptions) End(ctx context.Context, restConfig *rest.Config,
 	return nil
 }
 
+// AppendSyntheticTestIntervals is similar to End, except it takes list of new EventIntervals and appends
+// them to the current set of recordedEvents. Since these new EventIntervals happened within the
```
dgoodwin, line 159: Synthetic tests should not be injecting new events when they run, if so the events being moved around would be changing depending on when a test ran, later synthetics would get different results and we'd be headed for very weird and hard to debug bugs. 

We need to get these processed and added when we stop monitoring, likely in End.

pkg/synthetictests/duplicated_events.go
```diff
@@ -199,13 +237,94 @@ func (d duplicateEventsEvaluator) testDuplicatedEvents(testName string, flakeOnl
 
 		if flake || flakeOnly {
 			flakes = append(flakes, appendToFirstLine(msg, " result=allow "))
+			fmt.Println("Found patho flake")
```
dgoodwin, line 240: patho to pathological if you plan to keep this logging, and include the message would be nice and clear.

pkg/synthetictests/duplicated_events.go
```diff
@@ -199,13 +237,94 @@ func (d duplicateEventsEvaluator) testDuplicatedEvents(testName string, flakeOnl
...
+
+		if strings.Contains(pathoEvent.Message, duplicateevents.PathologicalMark) && !strings.Contains(pathoEvent.Message, duplicateevents.InterestingMark) {
+
+			fmt.Println("Processing event with pathologicalTrue but no interesting mark")
+
+			// This event occurred more than threshold but is not recognized (i.e., unknown).
+			// So we need to mark all other instances of this event so we can chart when this
+			// event started happening.
+
+			clientSet, clientErr := kubernetes.NewForConfig(kubeClientConfig)
```
dgoodwin, line 257: We should not be talking live to a cluster for this, we have all we need in the event intervals already I think. Will explain more below.

pkg/synthetictests/duplicated_events.go
```diff
@@ -134,34 +141,64 @@ func appendToFirstLine(s string, add string) string {
...
+		regexp.MustCompile(fmt.Sprintf("%s ", duplicateevents.InterestingMark)),
+		regexp.MustCompile(` \([1-9][0-9]+ times\)`),
+	}
+
+	originalMessage := messageString
+	for _, partsRegexp := range partsRe {
+		originalMessage = partsRegexp.ReplaceAllString(originalMessage, "")
+	}
+	return originalMessage
+}
```
dgoodwin, line 168: I'm hoping we do not need to reverse engineer this message, I think the key thing is that you may have focused on determining what was pathological via scanning for the flags you inserted earlier, rather than relying on the source events (which are already in the intervals) as you did originally.

dgoodwin (changes requested):
Apologies Dennis, some large requests in here. 

* We should avoid talking to a live cluster to get events, I believe we already have them, remember that the monitor is listening for kube Events and recording them as intervals.
* We should not modify the intervals when a particular test runs, it injects inconsistency in terms of what each synthetic test sees. Once we call End(), we want that to be the final list, so we need to get our code into that flow to add these additional events.
* We hopefully will not need to reverse engineer each message with all those intervals. That looks extremely fragile.

Regarding already having the kube events, when I say we already have them I mean we have intervals like this:

```
       {
            "level": "Info",
            "locator": "ns/openshift-kube-controller-manager pod/kube-controller-manager-ip-10-0-195-75.ec2.internal",
            "message": "reason/CreatedSCCRanges created SCC ranges for e2e-test-scheduling-pod-check-wmd2q namespace (2 times)",
            "from": "2023-02-20T13:32:39Z",
            "to": "2023-02-20T13:32:40Z"
        },
```

That's just the plain event recorded in the monitor, rather than the pathological flagged event you added recently. I think we can still work from this source material rather than starting with the flagged events you added and working backwards with a bunch of regexes.


In the monitoring package, as I understand it we are watching for kube events that hit a regex, or have the X times message where X is greater than 20, as they come in live from the cluster during the job run.

Essentially we need the list of messages (minus the "x times" portion), that were eventually over 20 but are not previously known via one of the regexes. There are probably two ways we could get that: (1) as we're processing incoming kube events in the monitor, same place as we watch for pathological, record the message to a new Set of strings on the monitor, or (2) in End() re-scan all events, apply same logic to determine if it's known/unknown and over 20 times, and if so save it to a Set of messages. 

Critically we need that Set in the End function. Then before we clamp/end we scan all intervals again, look for that message with either no "x times", or with "x times" but x is lower than 20, and add in an interval for them as well. It's a post-processing step taken before we finalize our intervals.

No cluster communication, no reverse engineering the message, and no morphing the intervals when a particular test runs, all synthetic tests get the same results. 

Hopefully I'm not confused myself, we can chat here or in person to work through any of this if I'm off base or something is unclear.

pkg/monitor/event.go
DennisPeriquet: I was trying to re-use it and neglected to remove it from the startEventMonitoring function.

pkg/monitor/event.go
DennisPeriquet: Looks like the original author was trying to parse something like `"something" in 5.6s`.

pkg/monitor/event_test.go
DennisPeriquet: ah you're right.  I probably just used it to step through that function.  I can tweak it to check for some kind of specific output.

pkg/synthetictests/duplicated_events.go
DennisPeriquet: Thanks for this.  Yes, I agree if we have what we need there already, no need to get it from the cluster.

pkg/synthetictests/duplicated_events.go
DennisPeriquet: Yes, logrus looks nice.

DennisPeriquet:
> Apologies Dennis, some large requests in here.
> 
> * We should avoid talking to a live cluster to get events, I believe we already have them,
> ...
> Hopefully I'm not confused myself, we can chat here or in person to work through any of this if I'm off base or something is unclear.

No problem.  Thanks much for the review and thanks for pointing out that the already processed `EventIntervals` are already there as it seems it could really simplify things.

I'm continuing the work on [PR27743](https://github.com/openshift/origin/pull/27743/files) as a bunch of what this PR went through is no longer necessary.  Let me run a few tests to see if we can get some logging to prove what I have there is doing what we need it to do.

We can sync up for a few min tomorrow.

PR #27726, opened 2023-02-14
OCPBUGS-7488: test flake: should not reconcile SC when state is Unmanaged

dobsonj:
https://issues.redhat.com/browse/OCPBUGS-7488
/cc @openshift/storage

### DIFF
--- a/test/extended/storage/storageclass.go
+++ b/test/extended/storage/storageclass.go
@@ -70,7 +70,7 @@ var _ = g.Describe("[sig-storage][Feature:DisableStorageClass][Serial]", func() 
 		sctest.SetAllowExpansion(false)
 
 		g.By("verifying the AllowVolumeExpansion reverts to true")
-		sctest.VerifyAllowExpansion(true)
+		sctest.VerifyAllowExpansion(true, nil)
 	})
 
 	g.It("should not reconcile the StorageClass when StorageClassState is Unmanaged", func() {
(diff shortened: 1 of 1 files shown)


test/extended/storage/storageclass.go
```diff
@@ -70,7 +70,7 @@ var _ = g.Describe("[sig-storage][Feature:DisableStorageClass][Serial]", func()
 		sctest.SetAllowExpansion(false)
 
 		g.By("verifying the AllowVolumeExpansion reverts to true")
-		sctest.VerifyAllowExpansion(true)
+		sctest.VerifyAllowExpansion(true, func() {})
```
jsafrane, line 73: this looks bad to me, can you pass `nil` and handle it inside `VerifyAllowExpansion`?

test/extended/storage/storageclass.go
```diff
@@ -209,6 +212,8 @@ func (d *DisableStorageClassTest) VerifyAllowExpansion(expected bool) {
 		if allowExpansion == expected {
 			e2e.Logf("AllowVolumeExpansion matched %t after %d attempts", expected, maxRetries)
 			return
+		} else {
+			retry()
```
jsafrane, line 216: Please log a retry + nr. of attempts.

test/extended/storage/storageclass.go
dobsonj: Fixed

gnufied:
Yep, this could be hard to test. If it flakes we could delete `Unmanaged` tests entirely. IMO - a failing test is worse than having no test. :-)

PR #27727, opened 2023-02-15
OCPBUGS-1125: remove reference to old guard pods

tjungblu:


tjungblu:
/assign @bparees 

Ben, could you please take a look?

PR #27729, opened 2023-02-15
Do not bail on producing artifacts when nodes are dead

dgoodwin:
Discovered that conformance test artifacts may not appear at all if
nodes are dead, due to recent work to query their logs for interval
creation.

Code modified to proceed even when errors happen here after logging the
problem.

[TRT-855](https://issues.redhat.com//browse/TRT-855)

### DIFF
--- a/pkg/test/ginkgo/options_monitor_events.go
+++ b/pkg/test/ginkgo/options_monitor_events.go
@@ -114,12 +114,12 @@ func (o *MonitorEventsOptions) End(ctx context.Context, restConfig *rest.Config,
 	// this happens before calculation because events collected here could be used to drive later calculations
 	events, err = intervalcreation.InsertIntervalsFromCluster(ctx, restConfig, events, o.recordedResources, fromTime, endTime)
 	if err != nil {
-		return fmt.Errorf("InsertIntervalsFromClusterError: %w", err)
+		fmt.Fprintf(o.ErrOut, "InsertIntervalsFromCluster error but continuing processing: %v", err)
 	}
 	// add events from alerts so we can create the intervals
 	alertEventIntervals, err := monitor.FetchEventIntervalsForAllAlerts(ctx, restConfig, *o.startTime)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/test/ginkgo/options_monitor_events.go
```diff
@@ -114,12 +114,12 @@ func (o *MonitorEventsOptions) End(ctx context.Context, restConfig *rest.Config,
 	// this happens before calculation because events collected here could be used to drive later calculations
 	events, err = intervalcreation.InsertIntervalsFromCluster(ctx, restConfig, events, o.recordedResources, fromTime, endTime)
 	if err != nil {
-		return fmt.Errorf("InsertIntervalsFromClusterError: %w", err)
+		fmt.Fprintf(o.ErrOut, "InsertIntervalsFromCluster error but continuing processing: %w", err)
 	}
```
DennisPeriquet, line 118: I know the original code used `%w` but it will result in a weird output if that Fprintf were to run.  You can use `%v`.

pkg/test/ginkgo/options_monitor_events.go
```diff
@@ -114,12 +114,12 @@ func (o *MonitorEventsOptions) End(ctx context.Context, restConfig *rest.Config,
...
 	// add events from alerts so we can create the intervals
 	alertEventIntervals, err := monitor.FetchEventIntervalsForAllAlerts(ctx, restConfig, *o.startTime)
 	if err != nil {
-		return fmt.Errorf("AlertErr: %w", err)
+		fmt.Fprintf(o.ErrOut, "FetchEventIntervalsForAllAlerts error but continuing processing: %w", err)
```
DennisPeriquet, line 122: same as above with `%v`

dgoodwin:
Thanks for heads up I didn't know that, will try to correct %w's as I see them from now on.

dgoodwin:
@DennisPeriquet sure enough the w's caused verify to fail. Needs another lgtm.

PR #27737, opened 2023-02-20
Add tap plugin test

mlguerrero12:
Test for the following PR:
https://github.com/containernetworking/plugins/pull/832

Signed-off-by: Marcelo Guerrero Viveros <marguerr@redhat.com>

### DIFF
--- a/test/extended/networking/tap.go
+++ b/test/extended/networking/tap.go
@@ -1,0 +1,114 @@
+package networking
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	kapiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	nadtypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const nodeLabelSelectorWorker = "node-role.kubernetes.io/worker"
+
+var _ = g.Describe("[sig-network][Feature:tap]", func() {
+	oc := exutil.NewCLI("tap")
+	f := oc.KubeFramework()
+	var worker *corev1.Node
+	var isCUDDisabled bool
+
+	g.BeforeEach(func() {
+		// Fetch worker nodes.
+		workerNodes, err := f.ClientSet.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{
+			LabelSelector: nodeLabelSelectorWorker,
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		if len(workerNodes.Items) == 0 {
+			e2e.Failf("cluster should have nodes")
+		}
+		worker = &workerNodes.Items[0]
+
+		// Load tun module.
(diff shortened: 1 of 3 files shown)


test/extended/networking/tap.go
```diff
@@ -0,0 +1,55 @@
...
+		nadName := "nad-tap"
+		ifName := "tap1"
+
+		g.By("creating a network attachment definition")
+		err := createNetworkAttachmentDefinition(
+			oc.AdminConfig(),
+			ns,
+			nadName,
+			fmt.Sprintf(`{"cniVersion":"0.4.0","name":"%s","type": "tap",
+"selinuxcontext": "system_u:system_r:container_t:s0", "ipam": {}}`, nadName),
```
mmirecki, line 33: Maybe make a variable out of the config,move to line 25, and format it for better readability?

mlguerrero12: done

test/extended/networking/tap.go
```diff
@@ -0,0 +1,100 @@
...
+
+		// Enable container_user_devices selinux boolean.
+		_, err = exutil.ExecCommandOnMachineConfigDaemon(f.ClientSet, oc, worker, []string{
+			"sh", "-c", "nsenter --mount=/proc/1/ns/mnt -- sh -c 'setsebool container_use_devices 1'",
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+	})
+
+	g.AfterEach(func() {
+		// Disable container_use_devices selinux boolean.
```
cgoncalves, line 52: In BeforeEach, the tun module is loaded (if not already) but here it is not unloaded. Should it be?
I guess it depends if something else on the system is using it.

test/extended/networking/tap.go
```diff
@@ -0,0 +1,100 @@
...
+		_, err := exutil.ExecCommandOnMachineConfigDaemon(f.ClientSet, oc, worker, []string{
+			"sh", "-c", "nsenter --mount=/proc/1/ns/mnt -- sh -c 'setsebool container_use_devices 0'",
```
cgoncalves, line 54: This SELinux boolean may have been already set as a dependency of some service/app on the host. Blindly disabling it might have unintended consequences. Perhaps check if it's already set (in BeforeEach) and only disable it (in AfterEach) if it was not.

test/extended/networking/tap.go
mlguerrero12: It is not loaded. The test always fails without it. I didn't unload it because it always complains that the module is in use even after removing the tap interface. So, I figured it is safe to leave it there.

test/extended/networking/tap.go
mlguerrero12: Test always fails without it. I know that it is a good practice to do what you're suggesting but I wonder if it is worth doing it in this case. No other tests modify selinux booleans.

test/extended/networking/tap.go
mmirecki: It should not be that hard to implement.
The BeforEach DS should mount a dir on the node, and note whether the bool has been set:
sh -c 'getsebool container_use_devices > /host/tmp/initial_state

and the AfterEach DS would then revert the value only if the content of this file is 0.

test/extended/networking/tap.go
cgoncalves: I don't have a strong opinion here so either way is fine.

mlguerrero12: Will it be wrong to leave it on just to see if it collides with other services/apps? This way, we would be able to detect what cannot be run together with the tap plugin.

cgoncalves: That's a good point.

mlguerrero12: done, let's see what we get

cgoncalves (approved):

mlguerrero12:
@deads2k, @bparees, could you please review/approve this pr?

PR #27740, opened 2023-02-20
OCPVE-278: feat: add rt tests package to openshift-tests

eggfoobar:
Add `rt-tests` package to `tests` image for running realtime tests on clusters with realtime worker nodes.


eggfoobar:
/unhold
/assign @stbenjam 
/cc @jeff-roche 

Hey folks, after doing some research, it seems this is the best place to contain the `rt-tests` package which will help us run realtime tests on clusters with out relying on extra dependencies. What do you think?

PR #27741, opened 2023-02-21
BUILD-64: Replacing `registry.redhat.io` images with equivalent images in `quay.io`

apoorvajagtap:
Could find mirrors for some of the images in `quay.io`, and rest were modified considering the behavior of the original image.

### DIFF
--- a/test/extended/builds/new_app.go
+++ b/test/extended/builds/new_app.go
@@ -54,7 +54,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds] oc new-app", func() {
 
 		g.It("should succeed with a --name of 58 characters [apigroup:build.openshift.io]", func() {
 			g.By("calling oc new-app")
-			err := oc.Run("new-app").Args("registry.redhat.io/ubi8/nodejs-16:latest~https://github.com/sclorg/nodejs-ex", "--name", a58, "--build-env=BUILD_LOGLEVEL=5").Execute()
+			err := oc.Run("new-app").Args("quay.io/app-sre/ubi8-nodejs-16:latest~https://github.com/sclorg/nodejs-ex", "--name", a58, "--build-env=BUILD_LOGLEVEL=5").Execute()
 			o.Expect(err).NotTo(o.HaveOccurred())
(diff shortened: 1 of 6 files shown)


test/extended/builds/new_app.go
```diff
@@ -54,7 +54,7 @@ var _ = g.Describe("[sig-builds][Feature:Builds] oc new-app", func() {
 
 		g.It("should succeed with a --name of 58 characters [apigroup:build.openshift.io]", func() {
 			g.By("calling oc new-app")
-			err := oc.Run("new-app").Args("registry.redhat.io/ubi8/nodejs-16:latest~https://github.com/sclorg/nodejs-ex", "--name", a58, "--build-env=BUILD_LOGLEVEL=5").Execute()
+			err := oc.Run("new-app").Args("quay.io/app-sre/ubi8-nodejs-16:latest~https://github.com/sclorg/nodejs-ex", "--name", a58, "--build-env=BUILD_LOGLEVEL=5").Execute()
```
bparees, line 57: this image definitely exists in registry.redhat.io, are you saying it doesn't behave the same way as the quay.io one?

bparees:
can you elaborate on the differences between the registry.redhat.io images and the ones you're replacing them with?

part of the value of using the images we were using is that those are the ones we ship imagestream references to.

apoorvajagtap:
@bparees the need as per [BUILD-64](https://issues.redhat.com/browse/BUILD-64) was to rely less on images from `registry.redhat.io` so that the tests don't fail due to `registry.redhat.io` flakes. 
So, all I did was find images in quay.io that are basically mirror of the images in registry.redhat.io. 
For example; replaced `registry.redhat.io/ubi8/nodejs-16` with [quay.io/app-sre/ubi8-nodejs-16](https://quay.io/repository/app-sre/ubi8-nodejs-16) (the exact mirror).

However, as per the slack discussion, we just realized that the availability of both registries is quite similar, so, I guess we can proceed with just closing this PR. 
Let me know in case I missed anything. Thanks!

PR #27744, opened 2023-02-22
OCPBUGS-7833: Fix pattern match for snapshot tests

jsafrane:
Round brackets must be escaped in regular expressions.

cc @openshift/storage 

I will rework the matching rules in a subsequent PR, I don't like matching exact test names that may break when upstream test is renamed or a new one is added.


jsafrane:
/payload-job periodic-ci-openshift-release-master-ci-4.13-e2e-aws-ovn-no-capabilities

bparees:
/approve

we only need one of either this PR or https://github.com/openshift/origin/pull/27746 though

jsafrane:
/close
in favor of https://github.com/openshift/origin/pull/27745

PR #27745, opened 2023-02-22
OCPBUGS-7833: Rework no optional capabilities rules

jsafrane:
Rework `[Skipped:NoOptionalCapabilities]` rules:

* Skip all tests that require `Feature:VolumeSnapshotDataSource`, it needs CSISnapshot capability.
* Skip all in-tree AWS tests, they require Storage capability.
* Don't skip tests that are skipped elsewhere.
* Enable all other tests, they should pass.

PR #27746, opened 2023-02-22
[OCPBUGS-7833] fix escapes

qJkee:
Fix regex escapes in test skips


bparees:
This needs a make update.

Also it looks like @jsafrane has another version of this pr:  https://github.com/openshift/origin/pull/27744

So I think one of the two can be closed

qJkee:
well, here i have one more line to add =)

jsafrane:
@qJkee, I'm fixing the storage + snapshot rules here: https://github.com/openshift/origin/pull/27745. I don't like skipping exact test names, because there will always be some new tests that will break it.
BTW, please tag `@openshift/storage` before skipping our tests, this could have been a simple comment on your initial PR.

bparees:
looks like https://github.com/openshift/origin/pull/27745 is going to go through (has all the labels) so to avoid a potential rebase conflict:

/hold

PR #27747, opened 2023-02-22
OCPBUGS-7519: Skip nfs tests temporarilly

sdodson:
The Centos Stream 9 kernel currently panics on nfs test suite and we'd like to disable that so that we're able to test without crashing nodes. There's a fix pending via https://bugzilla.redhat.com/show_bug.cgi?id=2169017 which we expect to receive within the next few days after which time we'll revert this change.


dobsonj:
/lgtm
FYI @openshift/storage this PR disables the `[Driver: nfs]` tests until we have the kernel fix.

PR #27748, opened 2023-02-22
OSASINFRA-3109: networking: add a test for control plane LB

EmilienM:
For on-prem platforms, check whether the HAproxy & Keepalived pods are
running, depending on the configuration of the Platform.

If using the internal LB managed by OpenShift, it should be running.
If running UPI or an External LB, it should not be running.


EmilienM:
The errors in `ci/prow/verify` will be fixed by https://github.com/openshift/origin/pull/27746

cybertron:
/lgtm

Assuming ci passes once the jobs are fixed.

EmilienM:
testing vsphere UPI via https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-vsphere-modern/1630372478419210240

PR #27755, opened 2023-02-27
record load balancer failures

deads2k:
add analysis of audit logs to show load balancer disruption.


deads2k:
https://github.com/openshift/kubernetes/pull/1456/files has morphed enough to push this out of date.  

@tkashem this should be addressed before going your own route.

/close

PR #27756, opened 2023-02-28
PSa podspec exporter e2e: don't specifically add seccompProfile to the podspec

stlaz:
The restricted-v2 SCC should mutate this instead.

/cc @openshift/openshift-team-auth @deads2k


stlaz:
This should keep failing until https://github.com/openshift/kubernetes/pull/1482 merges

PR #27757, opened 2023-02-28
OCPNODE-1257: feat: added unit test for graceful shutdown

eggfoobar:
Added the initial tests for verifying that the kubelet is respecting a pods shutdown grace period when configured for graceful shutdown, [see](https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown)

/cc @jerpeter1

### DIFF
--- a/test/extended/pods/graceful-shutdown.go
+++ b/test/extended/pods/graceful-shutdown.go
@@ -1,0 +1,321 @@
+package pods
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


jerpeter1 (approved):

rphillips:
lgtm

test/extended/pods/graceful-shutdown.go
```diff
@@ -0,0 +1,457 @@
...
+		deleteTestBed(context.Background(), oc)
+	})
+})
+
+func filePathForPod(podName string) string {
+	return fmt.Sprintf("%s/completed-%s", hostPath, podName)
+}
+
+// Wait for the node to come back up after reboot before attempting to validate results
+func waitForNode(oc *exutil.CLI, name string) error {
```
soltysh, line 161: Any particular reason you can't just use `framework.WaitForNodeToBeReady` from upstream's framework?

soltysh (approved):
/approve

test/extended/pods/graceful-shutdown.go
eggfoobar: Oh I didn't know that existed, didn't see it in the framework package. Created a new PR to use it instead, try and limit duplication https://github.com/openshift/origin/pull/27882 Thanks @soltysh !

PR #27781, opened 2023-03-07
TRT-889: Temp flake all azure disruption

neisw:
First pass at flaking all azure disruption


xueqzhan:
Do we want to disable imageregistry and service load balancer disruption tests as well?

neisw:
Got a pointer to where those tests are setup?  I can review.

dgoodwin:
We're probably ok to leave those other tests running, this problem is scoped to just the openshift/kube/oauth apiservers.

PR #27798, opened 2023-03-15
Scan etcd logs for dropped internal raft messages

dgoodwin:
If found, intervals will be created, charted in spyglass, and a test
will flake giving us insight into how often this is happening across the
fleet.


dgoodwin:

known issues.

PR #27807, opened 2023-03-16
Add (optional) dual-stack tests to the CNI certification test suite

danwinship:
Some third-party CNI vendors want to be certified as working in OpenShift with dual-stack, so add those tests to the suite.


danwinship:
/retest-required
(note that this does not change anything that any of the ci jobs run)

tssurya (approved):
/lgtm

PR #27808, opened 2023-03-21
[WIP] Added additional filtering options for e2e chart

jwforres:


jwforres:
@deads2k bulk of the change is the addition of the minified lodash like we talked about. Try this and see if it is doing what you are wanting for the OR filtering.

e2echart/e2e-chart-template.html
```diff
@@ -405,43 +613,71 @@ <h5 class="modal-title">Resource</h5>
...
+                if (ns.length > 0 && eventInterval.locatorObj.ns && eventInterval.locatorObj.ns.includes(ns)) {
+                    eventMatches = true;
+                }
+                if (category.length > 0 && eventInterval.categories[category]) {
+                    eventMatches = true;
+                }
+                return eventMatches;
+            });
+        }
+\
```
deads2k, line 658: this appears to cause the page to fail to render.  is it semantically important?

PR #27811, opened 2023-03-21
Test build.spec.source.configMaps in custom builds

nalind:
Test that build.spec.source.configMaps maps are mounted for use by custom builder images.


nalind:
I don't see us _not_ wanting to test this.
/lifecycle frozen

PR #27820, opened 2023-03-24
remove runresourcewatch git lock file on failure

deads2k:
Some resource watch repos stop recording history due to
> fatal: Unable to create '/logs/artifacts/resource-watch-store/repo/.git/index.lock': File exists.

Seen from
> bash: line 1: 270427 Bus error               (core dumped) git add "namespaces/openshift-ingress/core/pods/router-default-65c89f6859-94ttq.yaml"

Only our process should be working with git in this area, so adding a delete to the failure path so subsequent files can have intact history.

At some point, a retry for existing files may be useful.

/assign @xueqzhan


deads2k:

needed for debugging tools on periodics

PR #27824, opened 2023-03-27
OCPBUGS-11072: Add test for Egress Firewall node selector

trozet:
Adds an allow rule for EF with node selector to k8s control plane nodes then tests traffic towards kapi server.


trozet:
/retitle OCPBUGS-11072: Add test for Egress Firewall node selector

PR #27826, opened 2023-03-28
TRT-856: DNS and disruption overlap test

deepsm007:
https://issues.redhat.com/browse/TRT-856

### DIFF
--- a/pkg/synthetictests/disruption.go
+++ b/pkg/synthetictests/disruption.go
@@ -12,7 +12,6 @@ import (
 	"github.com/openshift/origin/test/extended/util/disruption"
 	"github.com/openshift/origin/test/extended/util/disruption/externalservice"
 	"github.com/sirupsen/logrus"
-
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/test/e2e/framework"
 )
@@ -308,5 +307,53 @@ func isOneSecondEvent(eventInterval monitorapi.EventInterval) bool {
 		return false
 	default:
 		return true
+	}
+}
+
+func testDNSOverlapDisruption(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[sig-network] Disruption should not overlap with DNS problems in cluster running tests"
+	failures := []string{}
+	dnsIntervals := []monitorapi.EventInterval{}
+	disruptionIntervals := []monitorapi.EventInterval{}
+	for _, event := range events {
+		// DNS outage
+		if reason := monitorapi.ReasonFrom(event.Message); strings.HasPrefix(reason, "DisruptionSamplerOutageBegan") {
+			dnsIntervals = append(dnsIntervals, event)
+		}
+		// real disruption
+		if reason := monitorapi.ReasonFrom(event.Message); strings.HasPrefix(reason, "DisruptionBegan") {
+			disruptionIntervals = append(disruptionIntervals, event)
+		}
+	}
+	errorCount := 0
+	for _, r := range disruptionIntervals {
+		for _, d := range dnsIntervals {
+			if (r.From.Before(d.To) && d.From.Before(r.To)) || (r.To.Add(10*time.Second).After(d.From) && d.To.Add(10*time.Second).After(r.From)) {
+				errorCount = errorCount + 1
+			}
+		}
+	}
+	if errorCount > 0 {
+		failures = append(failures, fmt.Sprintf("Overlap or interval within 10 seconds occured %d times.", errorCount))
+	}
+	if len(failures) == 0 {
+		return []*junitapi.JUnitTestCase{
+			{Name: testName},
+		}
+	}
+
+	output := "These failures imply disruption overlapped, or occurred in very close proximity to DNS problems in the cluster running tests.\n" + strings.Join(failures, "\n")
+
+	return []*junitapi.JUnitTestCase{
+		{
+			Name: testName,
+			FailureOutput: &junitapi.FailureOutput{
+				Output: output,
+			},
+			SystemOut: strings.Join(failures, "\n"),
+		},
+		{
+			Name: testName,
+		},
 	}
 }
--- a/pkg/synthetictests/disruption_test.go
+++ b/pkg/synthetictests/disruption_test.go
@@ -1,0 +1,140 @@
+package synthetictests
+
+import (
+	"testing"
+	"time"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_dnsOverlapDisruption(t *testing.T) {
+	events := []monitorapi.EventInterval{
+		{
+			Condition: monitorapi.Condition{
+				Locator: "disruption/openshift-api connection/new",
+				Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began",
+			},
+			From: time.Now(),
+			To:   time.Now().Add(1 * time.Minute),
+		},
+		{
+			Condition: monitorapi.Condition{
+				Locator: "disruption/openshift-api connection/new",
+				Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began",
+			},
+			From: time.Now().Add(2 * time.Minute),
+			To:   time.Now().Add(3 * time.Minute),
+		},
+		{
+			Condition: monitorapi.Condition{
+				Locator: "disruption/openshift-api connection/new",
+				Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began",
+			},
+			From: time.Now().Add(4 * time.Minute),
+			To:   time.Now().Add(5 * time.Minute),
+		},
+		{
+			Condition: monitorapi.Condition{
+				Locator: "disruption/openshift-api connection/new",
+				Message: "reason/DisruptionBegan disruption",
+			},
+			From: time.Now().Add(6 * time.Minute),
+			To:   time.Now().Add(7 * time.Minute),
+		},
+		{
+			Condition: monitorapi.Condition{
+				Locator: "disruption/openshift-api connection/new",
+				Message: "reason/DisruptionBegan disruption",
+			},
+			From: time.Now().Add(8 * time.Minute),
+			To:   time.Now().Add(9 * time.Minute),
+		},
+		{
+			Condition: monitorapi.Condition{
+				Locator: "disruption/openshift-api connection/new",
+				Message: "reason/DisruptionBegan disruption",
+			},
+			From: time.Now().Add(6 * time.Minute),
+			To:   time.Now().Add(8 * time.Minute),
+		},
+	}
+
+	testCases := []struct {
+		name      string
+		events    monitorapi.Intervals
+		expectErr bool
+	}{
+		{
+			name:      "No overlap between DNS and disruption",
+			events:    events,
+			expectErr: false,
+		},
+		{
+			name: "Partial Overlap between DNS and disruption",
+			events: append(events, monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Message: "reason/DisruptionBegan disruption",
+				},
+				From: time.Now().Add(3*time.Minute + 50*time.Second),
+				To:   time.Now().Add(4*time.Minute + 5*time.Second),
+			}),
+			expectErr: true,
+		},
+		{
+			name: "Complete Overlap between DNS and disruption",
+			events: append(events, monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began",
+				},
+				From: time.Now().Add(5 * time.Minute),
+				To:   time.Now().Add(6 * time.Minute),
+			}),
+			expectErr: true,
+		},
+		{
+			name: "Overlap within 10 seconds between DNS and disruption",
+			events: append(events, monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began",
+				},
+				From: time.Now().Add(6*time.Minute + 5*time.Second),
+				To:   time.Now().Add(6*time.Minute + 15*time.Second),
+			}),
+			expectErr: true,
+		},
+		{
+			name: "Overlap between DNS and disruption with same start time",
+			events: append(events, monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Message: "reason/DisruptionBegan disruption",
+				},
+				From: time.Now().Add(4 * time.Minute),
+				To:   time.Now().Add(4*time.Minute + 10*time.Second),
+			}),
+			expectErr: true,
+		},
+		{
+			name: "Overlap between DNS and disruption with same end time",
+			events: append(events, monitorapi.EventInterval{
+				Condition: monitorapi.Condition{
+					Message: "reason/DisruptionBegan disruption",
+				},
+				From: time.Now().Add(2*time.Minute + 45*time.Second),
+				To:   time.Now().Add(3 * time.Minute),
+			}),
+			expectErr: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			expected := 1
+			testResults := testDNSOverlapDisruption(tc.events)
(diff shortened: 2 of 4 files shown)


pkg/synthetictests/disruption.go
```diff
@@ -276,3 +277,64 @@ func isOneSecondEvent(eventInterval monitorapi.EventInterval) bool {
 		return true
 	}
 }
+
+func dnsOverlapDisruption(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
```
DennisPeriquet, line 281: ah, I just noticed the test is never run.  Call it `testDnsOverlapDisruption`, then add it to the bottom of `StableSystemEventInvariants` and `SystemUpgradeEventInvariants`.  You will see several other synthetic tests called in those two places.  All of them have the same function signature as this one.

pkg/synthetictests/disruption.go
```diff
@@ -276,3 +277,64 @@ func isOneSecondEvent(eventInterval monitorapi.EventInterval) bool {
 		return true
 	}
 }
+
+func testDnsOverlapDisruption(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[sig-network] Check DNS overlap with real disruption"
+	failures := []string{}
+	sortedEvents := mergeIntervals(events)
+
+	for i, event := range sortedEvents {
```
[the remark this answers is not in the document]
dgoodwin: Needs unit tests I fear, this is a tricky calculation.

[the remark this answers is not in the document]
dgoodwin: I don't think this handles the case where the DNS disruption is a second after the regular disruption does it?

Also do we want to expand the 1s limit a little, technically I think if you're within 5-10s I'm suspicious.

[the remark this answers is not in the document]
dgoodwin: I also don't know if this handles larger windows, like what if we had 10s of real, or DNS disruption, which I think gets lumped into one interval.

Example if we had 

t0 ------ real disruption -------t10
        t5 dns disruption t7
        
Would make great test case as would variants of the above, real within a longer DNS, etc.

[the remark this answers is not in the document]
DennisPeriquet: > Also do we want to expand the 1s limit a little, technically I think if you're within 5-10s I'm suspicious.

I thought the DNS test happens every second so if a lookup problem happened 5 seconds ago, it implies that there were 4 DNS checks that succeeded.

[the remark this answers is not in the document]
neisw: Deep we were fortunate enough to have some crazy dns disruption lately.  See if https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-machine-config-operator-3634-ci-4.14-upgrade-from-stable-4.13-e2e-azure-sdn-upgrade/1640366985034338304 is any help.

You could build a local test case off of https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/openshift-machine-config-operator-3634-ci-4.14-upgrade-from-stable-4.13-e2e-azure-sdn-upgrade/1640366985034338304/artifacts/e2e-azure-sdn-upgrade/openshift-e2e-test/artifacts/junit/e2e-events_20230327-160533.json

Then potentially collapse the JSON down to a more reasonable size if you want to use it as a source commited unit test.

There are other runs like that if we want to look at other examples too.

pkg/synthetictests/disruption.go
```diff
@@ -276,3 +277,64 @@ func isOneSecondEvent(eventInterval monitorapi.EventInterval) bool {
...
+	}
+
+	if len(failures) == 0 {
+		return []*junitapi.JUnitTestCase{
+			{Name: testName},
+		}
+	}
+
+	output := strings.Join(failures, "\n") +
+		"\nThese failures imply Disruption was preceeded by DNS failure within 1 second."
```
DennisPeriquet, line 310: Please put this string first instead of last so it appears as a title.

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,125 @@
...
+					Name: "DNS overlap disruption for openshift-api new connection",
+
+					FailureOutput: &junitapi.FailureOutput{
+						Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began for disruption/openshift-api connection/new GET requests over new connections: Get \"https://api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org:6443/apis/image.openshift.io/v1/namespaces/default/imagestreams\": dial tcp: lookup api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org on 172.30.0.10:53: read udp 10.131.172.119:52726-\u003e172.30.0.10:53: i/o timeout (likely a problem in cluster running tests, not the cluster under test)",
+					},
+				},
+			},
+		},
+		{
+			name: "DNS doesn't overlap disruption",
```
dgoodwin, line 60: more accurately, this is no dns lookup disruption, according to your input events.

dgoodwin (commented):
Not a full review just some requests for the tests, they seem to be failing for me.

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,125 @@
...
+						},
+						From: now.Add(-31 * time.Second),
+						To:   now.Add(-30 * time.Second),
+					},
+					{
+						Condition: monitorapi.Condition{
+							Locator: "ns/openshift-image-registry route/test-disruption-new disruption/image-registry connection/new",
+							Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began for ns/openshift-image-registry route/test-disruption-new disruption/image-registry connection/new GET requests over new connections: Get \"https://test-disruption-new-openshift-image-registry.apps.ci-op-v352gg5m-99751.aws-2.ci.openshift.org/healthz\": dial tcp: lookup test-disruption-new-openshift-image-registry.apps.ci-op-v352gg5m-99751.aws-2.ci.openshift.org: i/o timeout (likely a problem in cluster running tests, not the cluster under test)",
```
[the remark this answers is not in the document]
dgoodwin: Looks like this was actually me :)

reason/DisruptionSamplerOutageBegan is used for what we think looks like DNS disruption.

reason/DisruptionBegan is used for what we think looks like real disruption. 

See DisruptionBegan function in disruption_locator.go

pkg/synthetictests/disruption.go
```diff
@@ -276,3 +277,70 @@ func isOneSecondEvent(eventInterval monitorapi.EventInterval) bool {
...
+
+			if found && event.From.Sub(sortedEvents[index].From) <= 100*time.Second {
+				failures = append(failures, event.From.Format(time.RFC3339)+" "+event.Message)
+			} else if found && event.From.Sub(sortedEvents[index].From) <= 30*time.Second {
+				failures = append(failures, event.From.Format(time.RFC3339)+" "+event.Message)
+			} else if found && event.From.Sub(sortedEvents[index].From) <= 10*time.Second {
+				failures = append(failures, event.From.Format(time.RFC3339)+" "+event.Message)
+			} else if found && event.From.Sub(sortedEvents[index].From) <= 1*time.Second {
+				failures = append(failures, event.From.Format(time.RFC3339)+" "+event.Message)
+			}
```
dgoodwin, line 306: build two lists of intervals, 
if we see Reason/DisruptionSamplerOutageBegan we know it's DNS disruption
if we see reason/DisruptionBegan we know it's "real"

iterate real disruption list (r) and compare each item to every dns disruption (d).
look for:
- r.from < d.from && r.to >= d.to  (r fully overlaps d)
- r.from <= d.from <= r.to (r partially overlaps start of d)
- d.from <= r.from <= d.to (d partially overlaps start of r)
- r.from <= d.to <= r.to (r partially overlaps end of d)
- d.from <= r.to <= d.to (d partially overlaps end of r)
- r.from - d.to <= 10s
- d.from - r.to <= 10s
- r.to - d.from <= 10s
- d.to - r.from <= 10s

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,120 @@
...
+		want []*junitapi.JUnitTestCase
+	}{
+		{
+			name: "Detect disruption just after DNS lookup errors",
+			args: args{
+				events: []monitorapi.EventInterval{
+					{
+						Condition: monitorapi.Condition{
+							Locator: "disruption/openshift-api connection/new",
+							Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began for disruption/openshift-api connection/new requests over new connections: Get \"https://api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org:6443/apis/image.openshift.io/v1/namespaces/default/imagestreams\": dial tcp: lookup api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org: i/o timeout (likely a problem in cluster running tests, not the cluster under test)",
```
dgoodwin, line 30: I would extract two constants for Messages, one for DNS disruption, and one for real, will help with readability.

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,120 @@
...
+						},
+						From: now.Add(-29 * time.Second),
+						To:   now.Add(-28 * time.Second),
+					},
+				},
+			},
+			want: []*junitapi.JUnitTestCase{
+				&junitapi.JUnitTestCase{
+					FailureOutput: &junitapi.FailureOutput{
+						Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began for disruption/openshift-api connection/new GET requests over new connections: Get \"https://api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org:6443/apis/image.openshift.io/v1/namespaces/default/imagestreams\": dial tcp: lookup api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org on 172.30.0.10:53: read udp 10.131.172.119:52726-\u003e172.30.0.10:53: i/o timeout (likely a problem in cluster running tests, not the cluster under test)",
```
dgoodwin, line 49: Based on your code, shouldn't this FailureOutput include: "These failures imply Disruption overlapped DNS failure" ? If so, how is this passing? Actually locally it looks like all three tests are failing. :)

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,120 @@
...
+							Locator: "disruption/oauth-api connection/new",
+							Message: "reason/DisruptionBegan disruption/oauth-api connection/new stopped responding to GET requests over new connections: Get \"https://api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org:6443/apis/oauth.openshift.io/v1/oauthclients\": dial tcp 3.133.199.188:6443: i/o timeout",
+						},
+						From: now.Add(-31 * time.Second),
+						To:   now.Add(-30 * time.Second),
+					},
+					{
+						Condition: monitorapi.Condition{
+							Locator: "disruption/cache-openshift-api connection/new",
+							Message: "reason/DisruptionBegan disruption/cache-openshift-api connection/new stopped responding to GET requests over new connections: Get \"https://api.ci-op-v352gg5m-99751.aws-2.ci.openshift.org:6443/apis/image.openshift.io/v1/namespaces/default/imagestreams?resourceVersion=0\": dial tcp 3.133.199.188:6443: i/o timeout",
```
dgoodwin, line 69: Based on test name it looks like this Message should be DNS disruption, not real.

pkg/synthetictests/disruption_test.go
dgoodwin: I suspect you did not make it to tests yet, just got the code updated, in which case this makes more sense. I'll leave notes on the tests as review regardless.

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,120 @@
...
+						},
+						From: now.Add(-29 * time.Second),
+						To:   now.Add(-28 * time.Second),
+					},
+				},
+			},
+			want: []*junitapi.JUnitTestCase{
+				&junitapi.JUnitTestCase{
+					Name: "[sig-network] Disruption should not overlap with DNS problems in cluster running tests",
```
dgoodwin, line 78: Looks like this needs updating.

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,125 @@
...
+					{
+						Condition: monitorapi.Condition{
+							Locator: "ns/openshift-image-registry route/test-disruption-new disruption/image-registry connection/new",
+							Message: "reason/DisruptionSamplerOutageBegan DNS lookup timeouts began for ns/openshift-image-registry route/test-disruption-new disruption/image-registry connection/new GET requests over new connections: Get \"https://test-disruption-new-openshift-image-registry.apps.ci-op-v352gg5m-99751.aws-2.ci.openshift.org/healthz\": dial tcp: lookup test-disruption-new-openshift-image-registry.apps.ci-op-v352gg5m-99751.aws-2.ci.openshift.org: i/o timeout (likely a problem in cluster running tests, not the cluster under test)",
+						},
+						From: now.Add(-29 * time.Second),
+						To:   now.Add(-28 * time.Second),
+					},
+				},
+			},
```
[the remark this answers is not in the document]
dgoodwin: As above, add more tests to cover most if not all of the cases in that complex logic structure we came up with, and get some with larger than one second from->two.

neisw:
Something to try out.  When I run it currently I get 1,506,458 failures in the output string.  With some of the fixes we discussed I still get 8,648 failures.  We might want to quit processing after X failures.

[disruption_loader_test.patch.txt](https://github.com/openshift/origin/files/11324673/disruption_loader_test.patch.txt)

pkg/synthetictests/disruption.go
```diff
@@ -310,3 +310,81 @@ func isOneSecondEvent(eventInterval monitorapi.EventInterval) bool {
...
+			dnsIntervals = append(dnsIntervals, event)
+		}
+		// real disruption
+		if reason := monitorapi.ReasonFrom(event.Message); strings.HasPrefix(reason, "DisruptionBegan") {
+			disruptionIntervals = append(disruptionIntervals, event)
+		}
+	}
+	for _, r := range disruptionIntervals {
+		for _, d := range dnsIntervals {
+			if r.From.Before(d.From) && (r.To.After(d.To) || r.To == d.To) {
```
dgoodwin, line 332: ChatGPT version looked a lot better, any reason we can't use it?

pkg/synthetictests/disruption.go
```diff
@@ -310,3 +309,48 @@ func isOneSecondEvent(eventInterval monitorapi.EventInterval) bool {
...
+		}
+		// real disruption
+		if reason := monitorapi.ReasonFrom(event.Message); strings.HasPrefix(reason, "DisruptionBegan") {
+			disruptionIntervals = append(disruptionIntervals, event)
+		}
+	}
+	for _, r := range disruptionIntervals {
+		for _, d := range dnsIntervals {
+			if (r.From.Before(d.To) && d.From.Before(r.To)) || (r.To.Add(10*time.Second).After(d.From) && d.To.Add(10*time.Second).After(r.From)) {
+				failures = append(failures, "Overlap or interval within 10 seconds found.")
```
neisw, line 331: We are appending the same message over and over here.   In the patch I provided earlier that pulls in real intervals data we see 6,107 messages duplicated in the test output.  Let's change this to a count and output the message once with the number of overlaps we detected appended to it.

pkg/synthetictests/disruption_test.go
```diff
@@ -0,0 +1,109 @@
...
+				To:   time.Now().Add(6*time.Minute + 15*time.Second),
+			}),
+			expectErr: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			testResults := testDNSOverlapDisruption(tc.events)
+
```
neisw, line 99: Alternate approach here for consideration that is a little smaller

`
			// default no error
			expected := 1
			if tc.expectErr {
				expected = 2
			}
			assert.Equal(t, expected, len(testResults), "Test results did not match")`

pkg/synthetictests/disruption_test.go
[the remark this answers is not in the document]
neisw: Agree we could have more test cases here.  I preferred what you had earlier where you set now
`now := time.Now()`
and then manipulated it.  You could set times that overlapped within 9 seconds of each other and 11 seconds to test edge cases, etc.

deepsm007:
The failures doesn’t look related to the code change

PR #27828, opened 2023-03-29
add apiserver availability timeline events

deads2k:
example PR of how to scan pod logs for particular failures.  apiserver availability probably also wants to scan node logs.

### DIFF
--- a/pkg/monitor/apiserveravailability/pod_log.go
+++ b/pkg/monitor/apiserveravailability/pod_log.go
@@ -1,0 +1,148 @@
+package apiserveravailability
+
+import (
+	"bufio"
+	"context"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+)
+
+func APIServerAvailabilityIntervalsFromCluster(kubeClient kubernetes.Interface, beginning, end time.Time) (monitorapi.Intervals, error) {
+	summary, err := SummarizeInterestingPodLogs(context.TODO(), kubeClient)
+	if err != nil {
+		return nil, err
+	}
+
+	allIntervals := monitorapi.Intervals{}
+	// as more are added, append
+	allIntervals = append(allIntervals, summary.WriteOperationFailures...)
+
+	if beginning.IsZero() || end.IsZero() {
+		return allIntervals, nil
+	}
+
+	timeFiltered := monitorapi.Intervals{}
+	for i := range allIntervals {
+		interval := allIntervals[i]
+		if interval.From.After(end) || interval.To.Before(beginning) {
+			continue
+		}
+		timeFiltered = append(timeFiltered, interval)
+	}
+
+	sort.Sort(timeFiltered)
+	return timeFiltered, nil
+}
+
+var interestingNamespaces = []string{"openshift-cluster-version"}
+
+func SummarizeInterestingPodLogs(ctx context.Context, client kubernetes.Interface) (*APIServerClientAccessFailureSummary, error) {
+	summary := &APIServerClientAccessFailureSummary{}
+	wg := sync.WaitGroup{}
+	for _, namespace := range interestingNamespaces {
+		wg.Add(1)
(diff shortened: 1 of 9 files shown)


deads2k:
proof in https://github.com/openshift/origin/pull/27830

pkg/monitor/apiserveravailability/pod_log.go
```diff
@@ -0,0 +1,148 @@
...
+				defer wg.Done()
+
+				locator := monitorapi.LocatePodContainer(pod, containerName)
+				currSummary := &APIServerClientAccessFailureSummary{}
+				logOptions := &corev1.PodLogOptions{
+					Container:                    containerName,
+					Follow:                       false,
+					Previous:                     previousContainer,
+					Timestamps:                   true,
+					InsecureSkipTLSVerifyBackend: true, // set in case the kubelet doesn't have valid serving certs
```
wking, line 123: If the kubelet doesn't have valid serving certs, would we still be interested in analyzing `iptables-operation-not-permitted` and other flaky stuff from that run?  Because I'd expect top-level test-cases for "all the kubelets have valid serving certs", and that investigating any failures there would be high-priority.  If that's true, this softening here doesn't get you access to much more data, and it does open you up to sharing your creds with a man-in-the-middle impersonator, and that seems like an ambiguous risk/reward trade-off.

pkg/synthetictests/disruption.go
```diff
@@ -117,6 +117,39 @@ func testServerAvailability(
 	}
 }
 
+func TestAPIServerIPTablesAccessDisruption(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[bz-kube-apiserver] kube-apiserver should be accessible by clients using internal load balancer"
+	namespacesToCount := map[string]int{}
+	messages := []string{}
+	for _, event := range events {
+		reason := monitorapi.ReasonFrom(event.Message)
+		if !strings.Contains(reason, "iptables-operation-not-permitted") {
```
wking, line 126: This seems brittle, as we will presumably grow other concerning log lines in the future.  And there could also be false-positives where we get `iptables-operation-not-permitted` lines in logs while attempting DNS resolution of things that are not the internal Kube-API domain.  Worth having a specific slug in the `reason` for "kube-apiserver should be accessible by clients using internal load balancer"?

pkg/synthetictests/disruption.go
```diff
@@ -117,6 +117,39 @@ func testServerAvailability(
...
+	if len(messages) > 0 {
+		failureOutput := ""
+		for _, ns := range sets.StringKeySet(namespacesToCount).List() {
+			failureOutput += fmt.Sprintf("namespace/%v has %d instances of 'write: operation not permitted'\n", ns, namespacesToCount[ns])
+		}
+		failureOutput += "\n\n"
+		failureOutput += strings.Join(messages, "\n")
+
+		test.FailureOutput = &junitapi.FailureOutput{
+			Output: failureOutput,
```
wking, line 145: this looks like a non-flake, fatal error?  Current estimates of failure rates are >1%, so perhaps it's worth opening as a flake, and then tightening to fatal if/when existing releases covered by the origin dev branch (which currently includes older 4.13.z, etc, for runs testing 4.13.z -> 4.14 nightly updates) are sufficiently clean?

deads2k: I did forget unconditionally add a success.  I'll do so

pkg/synthetictests/disruption.go
deads2k: > Worth having a specific slug in the reason for "kube-apiserver should be accessible by clients using internal load balancer"?

I figure eventually they'll be motivated to invest here and tweak it however they wish.

deads2k: and reasons are already like APIs in our events.

wking: `iptables-operation-not-permitted` is not a clear API for `relevant to  internal Kube-API domain`, but :shrug: fair enough that future changes are cheap.

pkg/monitor/apiserveravailability/pod_log.go
wking: Ah, ok.  Backing that up with [docs][1]:

> This means the apiserver cannot verify the log data it is receiving came from the real kubelet.

and I agree that that's an acceptable risk :+1:

[1]: https://github.com/kubernetes/api/blob/ec59454f24f514bb472628bc0a5c8c9125218dbc/core/v1/types.go#L5776

PR #27834, opened 2023-03-31
OCPBUGS-11215: fix: add namespace annotation helper for egress cni test

eggfoobar:


eggfoobar:
Hey @dgoodwin , sorry, can I get a `lgtm` again, I needed to update the code to change the value in the initial pointer since that's the one that gets passed over to the BeforeEach block.

PR #27843, opened 2023-04-03
Revert "TRT-889: Temp flake all azure disruption"

neisw:
Reverts openshift/origin#27781

We are past the point of needing to allow azure disruption to pass through unchecked. Reverting.


dgoodwin:
/lgtm
/hold

/payload 4.14 nightly blocking

Feel free to cancel hold if the jobs look healthy.

PR #27847, opened 2023-04-03
OCPBUGS-11315: Increasing limits for Nodes OSUpdateStaged time test

deepsm007:
/cc @sdodson

### DIFF
--- a/pkg/synthetictests/operators.go
+++ b/pkg/synthetictests/operators.go
@@ -90,8 +90,8 @@ type startedStaged struct {
 func testOperatorOSUpdateStaged(events monitorapi.Intervals, clientConfig *rest.Config) []*junitapi.JUnitTestCase {
 	testName := "[bz-Machine Config Operator] Nodes should reach OSUpdateStaged in a timely fashion"
 	success := &junitapi.JUnitTestCase{Name: testName}
-	flakeThreshold := 5 * time.Minute
-	failThreshold := 10 * time.Minute
+	flakeThreshold := 7 * time.Minute
+	failThreshold := 14 * time.Minute
 
 	// Scan all OSUpdateStarted and OSUpdateStaged events, sort by node.
 	nodeNameToOSUpdateTimes := map[string]*startedStaged{}


sdodson:
We expect that the one-time move from 8.6 to 9.2 will be an exceptionally large rpm-ostree rebase. Bumping these up is fine with me.
/lgtm

sdodson:
/retitle OCPBUGS-11315: Increasing limits for Nodes OSUpdateStaged time test

sdodson:
@bparees Do you mind approving this one-off bump for 4.13. FYI there's a separate bug for a slightly more commonly occurring condition where the jobs don't simply exceed the thresholds but instead get marked as never having complete. I suspect the test may be problematic for those.

That one is https://issues.redhat.com/browse/OCPBUGS-11304 (notice the log messages are slightly different from the failed test)

bparees:
novel to the rhel8->9 update
/approve

rioliu-rh:
hold this PR for pre-merge testing 
/cc @sergiordlr

sdodson:
Most recent failure is on SCA certs permissions. Every other failure on this job has had a unique cause.
```
level=info msg=Cluster operator insights SCAAvailable is False with Forbidden: Failed to pull SCA certs from https://api.openshift.com/api/accounts_mgmt/v1/certificates: OCM API https://api.openshift.com/api/accounts_mgmt/v1/certificates returned HTTP 403: {"code":"ACCT-MGMT-11","href":"/api/accounts_mgmt/v1/errors/11","id":"11","kind":"Error","operation_id":"a0ed5ca7-2e28-4cab-9b53-961f6a60cfb4","reason":"Account with ID 2DUeKzzTD9ngfsQ6YgkzdJn1jA4 denied access to perform create on Certificate with HTTP call POST /api/accounts_mgmt/v1/certificates"}
```

@sergiordlr @rioliu-rh This should just loosen some of the thresholds in upgrades to 4.13, the primary goal is that we know 8.6 to 9.2 rpm-ostree rebase will take longer than other rebases because it's a major version upgrade. Let us know if you think we need to still hold this but we'd like to get this in sooner rather than later.

neisw:
TRT has reviewed and recommended an override on the failing gcp test as it should not be impacted by the change.

PR #27695, opened 2023-01-31
bump timeout

Elbehery:
Due to recent failures in Vertical Scaling, increasing the time out before declaring the test as failed

cc @tjungblu @hasbro17

PR #27788, opened 2023-03-09
update-etcd-scaling-test

Elbehery:
This PR addresses reviews on https://github.com/openshift/origin/pull/27702 

cc @mfojtik @deads2k @hasbro17

PR #27850, opened 2023-04-04
bump timeout

Elbehery:
This PR increased the timeout of etcd vertical scaling e2e test.

maunal cherrypick of https://github.com/openshift/origin/pull/27849

The test scenario attempts to add a new master machine, and delete another master machine. Machine provisioning requires time from the provider side, also draining and deleting a machine. CI results shows a lot of timeout during machine deletion and provisioning.

This PR attempts to increase the timeout according to observing the time required from the machine-controller logs.

see https://github.com/openshift/origin/pull/27695

cc @hasbro17 @damdo


Elbehery:
will be backported with https://github.com/openshift/origin/pull/27788

/close

PR #27853, opened 2023-04-04
Render html from monitor

deads2k:
this will give us timelines for *prior* to installation.

### DIFF
--- a/cmd/openshift-tests/openshift-tests.go
+++ b/cmd/openshift-tests/openshift-tests.go
@@ -15,19 +15,13 @@ import (
 
 	"github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/library-go/pkg/serviceability"
-	"github.com/openshift/origin/pkg/monitor"
-	"github.com/openshift/origin/pkg/monitor/apiserveravailability"
-	"github.com/openshift/origin/pkg/monitor/monitor_cmd"
-	"github.com/openshift/origin/pkg/monitor/nodedetails"
+	"github.com/openshift/origin/pkg/cmd/monitor_command"
 	"github.com/openshift/origin/pkg/monitor/resourcewatch/cmd"
 	"github.com/openshift/origin/pkg/riskanalysis"
 	testginkgo "github.com/openshift/origin/pkg/test/ginkgo"
 	"github.com/openshift/origin/pkg/version"
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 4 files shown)


cmd/openshift-tests/openshift-tests.go
```diff
@@ -109,7 +111,7 @@ func main() {
 }
 
 func newRunMonitorCommand() *cobra.Command {
```
DennisPeriquet, line 113: I think you renamed this to `monitor_command.NewRunMonitorCommand` so this code is probably no longer needed (it's definitely not called).

Same applies to `monitor_command.NewMonitorCommand` and `newMonitorCommand`.

pkg/cmd/monitor_command/run_monitor_command.go
```diff
@@ -0,0 +1,181 @@
...
+		if err := os.MkdirAll(eventDir, os.ModePerm); err != nil {
+			fmt.Printf("Failed to create monitor-events directory, err: %v\n", err)
+			return err
+		}
+
+		if err := monitor.WriteEventsForJobRun(eventDir, recordedResources, recordedEvents, timeSuffix); err != nil {
+			fmt.Printf("Failed to write event data, err: %v\n", err)
+			return err
+		}
+		if err := monitor.WriteTrackedResourcesForJobRun(eventDir, recordedResources, recordedEvents, timeSuffix); err != nil {
```
DennisPeriquet, line 152: Hm..., you are now calling `WriteTrackedResourcesForJobRun` here.  I see it's also called [here](https://github.com/openshift/origin/blob/88793278dda83b5fd5703dc8eea329f806478c20/pkg/test/ginkgo/options_monitor_events.go#L75)
Will the call on line 148 overwrite what the call in line 75 already generated?

deads2k: > Hm..., you are now calling `WriteTrackedResourcesForJobRun` here. I see it's also called [here](https://github.com/openshift/origin/blob/88793278dda83b5fd5703dc8eea329f806478c20/pkg/test/ginkgo/options_monitor_events.go#L75)
> Will the call on line 148 overwrite what the call in line 75 already generated?

those are two different types I think.  I'd be happy to see them collapsed into one, but I don't think we need that in order to get value here.

cmd/openshift-tests/openshift-tests.go
deads2k: fixed

PR #27854, opened 2023-04-04
retry getting the URL for routes until we succeed

deads2k:
fails in our e2e observer.  We need this data to determine when the route is generally available.


deads2k:
upgrade shows that we got the URL successfully.  Merging this to fix the run resource watch.

PR #27852, opened 2023-04-04
add specific test for failing cgroups path

deads2k:
This will only catch instances *after* the tests have started.  Many instances exist during installation: https://search.ci.openshift.org/?search=static+pod+lifecycle+failure+-+static+pod&maxAge=336h&context=1&type=junit&name=4.13.*gcp&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

This will help identify how many more we've got.  And given it means, "pods don't run right", I wouldn't expect otherwise successful runs to hit this.

@rphillips to be sure this isn't a normal error
@deepsm007 to be sure he thinks it will work

PR #27856, opened 2023-04-05
[4.12] add specific test for failing cgroups path

deads2k:
4.12 version of https://github.com/openshift/origin/pull/27852

this was not clean


deads2k:

unrelated, need test for determining regression or not.

PR #27857, opened 2023-04-05
add wait for ingress to be responsive before beginning upgrade

deads2k:
This adds a wait before installs start to be sure that ingress is available before beginning an upgrade so that upgrades become a cleaner signal.


dgoodwin (commented):
/lgtm

deads2k:
no conflict with last merge.  We need clean data on our ingress problem.  merging.

PR #27858, opened 2023-04-05
use v8.1 of s2i-php-container test-app

jluhrsen:
v7.0 was removed recently [0] and 8.1 is the most recent [1]

[0] https://github.com/sclorg/s2i-php-container/commit/d9eb94736dfce8bc09496503175a1d0125af6fc5
[1] https://github.com/sclorg/s2i-php-container/tree/13158390b9412f569c9e78ce8002d88b2376caeb/8.1/test


jluhrsen:
I'm guessing folks don't care about this, being 3.11 and all. I don't think the presubmits even work.

PR #27862, opened 2023-04-07
Revert "add wait for ingress to be responsive before beginning upgrade"

deepsm007:
Reverts openshift/origin#27857

We are observing upgrade failures for GCP, AWS and Azure.


deepsm007:
Found the issue, its related to the console operator.

PR #27864, opened 2023-04-10
deploy: Change imagestreams importMode to work with multi-arch compute clusters

Prashanth684:
On multi-arch compute clusters, the pods can land on any arch type of worker node. Making sure that the imagstreams import the manifestlist will help make sure the deployment succeeds on any node that it lands on.

### DIFF
--- a/test/extended/deployments/deployments.go
+++ b/test/extended/deployments/deployments.go
@@ -428,7 +428,7 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 			o.Expect(waitForSyncedConfig(oc, dcName, deploymentRunTimeout)).NotTo(o.HaveOccurred())
 
 			g.By("tagging the initial test:v1 image")
-			_, err = oc.Run("tag").Args("image-registry.openshift-image-registry.svc:5000/openshift/cli:latest", "test:v1").Output()
+			_, err = oc.Run("tag").Args("--import-mode=PreserveOriginal", "image-registry.openshift-image-registry.svc:5000/openshift/cli:latest", "test:v1").Output()
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			expectLatestVersion := func(version int) {
@@ -459,7 +459,7 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 			o.Expect(waitForSyncedConfig(oc, dcName, deploymentRunTimeout)).NotTo(o.HaveOccurred())
 
 			g.By("tagging a different image as test:v2")
-			_, err = oc.Run("tag").Args("image-registry.openshift-image-registry.svc:5000/openshift/tools:latest", "test:v2").Output()
+			_, err = oc.Run("tag").Args("--import-mode=PreserveOriginal", "image-registry.openshift-image-registry.svc:5000/openshift/tools:latest", "test:v2").Output()
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			g.By("ensuring the deployment config latest version is 2 and rollout completed")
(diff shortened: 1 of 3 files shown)


Prashanth684:
@soltysh pretty simple change - if you could approve, would be great.

test/extended/deployments/deployments.go
```diff
@@ -563,7 +563,7 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 
 		g.It("should run a successful deployment with a trigger used by different containers [apigroup:apps.openshift.io][apigroup:image.openshift.io]", func() {
 
-			_, err := oc.Run("import-image").Args("registry.redhat.io/ubi8/ruby-30:latest", "--confirm", "--reference-policy=local").Output()
+			_, err := oc.Run("import-image").Args("--import-mode=PreserveOriginal", "registry.redhat.io/ubi8/ruby-30:latest", "--confirm", "--reference-policy=local").Output()
```
deads2k, line 566: this looks improper.  Our CLI should be backwards compatible.

deads2k:
/hold

The changes here suggest that something broke compatibility-wise with our CLI and possibly our server.  I'd like to understand why these changes are necessary before merging them.  Also, very glad we had tests that demonstrate this compatibilty breakage.

jaypoulz:
@deads2k The intent of these changes was to handle the case of testing across mismatched architectures.

Our intent behind the original https://github.com/openshift/release/pull/40722 was to ensure that e2e tests running from an x86 build cluster would still be able to pass tests that were run on a target arm64 cluster.

With OpenShift 4.13, the image streams now support manifest lists. However, the default behavior when importing an image is to only import the image that matches the client architecture. This causes a problem when your test-pod architecture doesn't match your target cluster architecture because the image you've imported is incompatible.

The solution is to specify the import of the original manifest (which is a manifest list). This allows the cluster to resolve the correct image upon inspection.

@Prashanth684 do you know more about the actual changes to the oc CLI for context?

Prashanth684:
> With OpenShift 4.13, the image streams now support manifest lists. However, the default behavior when importing an image is to only import the image that matches the client architecture. This causes a problem when your test-pod architecture doesn't match your target cluster architecture because the image you've imported is incompatible.

The above reason was why the changes were introduced - when deploying a multi-arch cluster with the multi payload, we need to ensure that oc commands in general import the manifestlist rather than a single manifest so it can support any architecture.

jaypoulz:
@deads2k The functionality of the flag is itself backward compatible.

If the target image stream is a single arch manifest, it does nothing.
If the target image stream is a manifest list, it ensures the list is pulled.

The reason the arm sdn tests began to fail was that we changed the payload (a few weeks ago) to use a manifest list instead of the single-arch manifest. It used to work because even though the test pod is in an x86 cluster, the manifest it was importing was a single-arch manifest for arm, which is compatible with the target cluster.

Prashanth684:
> > However, the default behavior when importing an image is to only import the image that matches the client architecture.
> 
> Does this update indicate that we actually want the default to be based on the server architecture, not the client architecture? We want the pods using this image to succeed and those pods aren't running on the client.

That statement is incorrect - the default behavior when importing an image is to import the manifest of the architecture that matches the control plane's architecture. this is because the apiserver runs on the control plane and uses GOARCH to filter the manifest. This update makes is so that - when a manifestlist is imported, it imports the manifestlist and does not import only a single arch manifest.

deads2k:
> This update makes is so that - when a manifestlist is imported, it imports the manifestlist and does not import only a single arch manifest.

Why does this happen?  Do our clusters have masters in architecture/A and workers in arch/B?  If so, I'd like to see the test gated so that it only adds thsi argument to the heterogenous case.

jaypoulz:
> Why does this happen? Do our clusters have masters in architecture/A and workers in arch/B?

We currently have 2 jobs that run with multi-arch compute nodes (i.e. heterogeneous clusters). They currently run with 3 amd64 control-plane nodes, 3 amd64 compute nodes, and 2 arm64 compute nodes.

So this test would pass about 3/5 times without this flag.

> If so, I'd like to see the test gated so that it only adds thsi argument to the heterogenous case.
This flag doesn't affect imports unless the image stream is using manifest lists. What concern would we be addressing by not including it?

Prashanth684:
@deads2k  reviving this thread - how do we want to proceed here? given that the impact of changing importMode is minimal , i.e the images will only import manifestlists when the payload used is a multi payload with the exception of the redis and postgres image which will always be imported as manifestlists (as they are not associated to the payload).

Would it be acceptable for you to allow this?

test/extended/networking/egressip_helpers.go
```diff
@@ -334,7 +333,7 @@ func createPacketSnifferDaemonSet(oc *exutil.CLI, namespace string, scheduleOnHo
 	f := oc.KubeFramework()
 	clientset := f.ClientSet
 
-	tcpdumpImage, err := util.DetermineImageFromRelease(oc, "network-tools")
+	tcpdumpImage, err := exutil.GetDockerImageReference(oc.ImageClient().ImageV1().ImageStreams("openshift"), "network-tools", "latest")
```
Prashanth684, line 336: @tvardema could you pull this change into a separate PR an submit to origin. The basic idea of the change is to not use the `DetermineImageFromRelease` function as that spins up a pod using the cluster version's image ref. This is not desirable for a multi-arch compute cluster as the cluster version's image ref is a single arch reference (which is a problem which needs fixing by itself). But in any case, this is a simpler way to resolve an image reference rather than spinning up a pod

tvardema: Do you mean just this one change to `egressip_helpers.go`?

Prashanth684: correct

PR #27865, opened 2023-04-10
Add git retry to run resource watch

deads2k:
Two parts.
1. the easy retry on a 1s retry attempt for 15s
2. the harder part for parallelizing the onAdd (called serially for every resource type, in parallel for different resources) so that retries don't delay all other resources of the same type.

The combination should allow us 15s to retry in case it's a flushing problem.  There are many, "nothing to commit" messages in the log and then diffs in the final repo.

### DIFF
--- a/pkg/monitor/resourcewatch/storage/git_store.go
+++ b/pkg/monitor/resourcewatch/storage/git_store.go
@@ -24,9 +24,43 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+type workingSet struct {
+	currentlyWorking sets.String
+	lock             sync.RWMutex
+}
+
+func (s *workingSet) isWorkingOn(key string) bool {
+	s.lock.RLock()
+	defer s.lock.RLock()
+	return s.currentlyWorking.Has(key)
+}
+
+func (s *workingSet) reserve(key string) {
+	s.lock.Lock()
+	defer s.lock.Lock()
+	s.currentlyWorking.Insert(key)
+}
+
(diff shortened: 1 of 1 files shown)


pkg/monitor/resourcewatch/storage/git_store.go
```diff
@@ -103,34 +147,68 @@ func (s *GitStorage) handle(gvr schema.GroupVersionResource, oldObj, obj *unstru
...
+	})
 }
 
 func (s *GitStorage) OnAdd(gvr schema.GroupVersionResource, obj interface{}) {
 	objUnstructured := obj.(*unstructured.Unstructured)
-	s.handle(gvr, nil, objUnstructured, false)
+
+	// serialize updates to individual files
+	key := fmt.Sprintf("%s/%s/%s/%s/%s", gvr.Group, gvr.Version, gvr.Resource, objUnstructured.GetNamespace(), objUnstructured.GetName())
+	if err := s.currentlyRecording.waitUntilAvailable(key); err != nil {
```
xueqzhan, line 182: I don't remember the details of dynamic informers. Will OnAdd/OnUpdate/OnDelete for the same key be called from different threads?

deads2k: >I don't remember the details of dynamic informers. Will OnAdd/OnUpdate/OnDelete for the same key be called from different threads?

For each informer, OnAdd/OnUpdate/OnDelete are all called serially

PR #27875, opened 2023-04-13
ETCD-417: add a new basic etcd backup/restore test

tjungblu:
* obsoletes the older non-working tests (due to ssh bastions)
* rebuilds a slightly simpler version that only runs the backup and restore scripts with ssh from within the cluster

### DIFF
--- a/test/extended/dr/common.go
+++ b/test/extended/dr/common.go
@@ -369,7 +369,7 @@ func ensureControlPlaneSSHAccess(oc *exutil.CLI, name string, namespace string) 
 
 	var cpNodeInternalIps []string
 	for _, node := range masterNodes(oc) {
-		framework.Logf("CP Node meta: %s, addresses: %v", node.ObjectMeta.String(), node.Status.Addresses)
+		framework.Logf("CP Node: %s, addresses: %v", node.ObjectMeta.Name, node.Status.Addresses)
 		for _, address := range node.Status.Addresses {
 			if address.Type == corev1.NodeInternalIP {
 				cpNodeInternalIps = append(cpNodeInternalIps, address.Address)
@@ -691,4 +691,48 @@ func runPodAndWaitForSuccess(oc *exutil.CLI, pod *applycorev1.PodApplyConfigurat
 	// TODO(thomas): on failure it would be great to get the last logs out of the pod
 
 	return err
+}
+
+func removeMemberOfNode(oc *exutil.CLI, node *corev1.Node) error {
+	etcdEndpointsConfigMap, err := oc.AdminKubeClient().CoreV1().ConfigMaps("openshift-etcd").Get(context.Background(), "etcd-endpoints", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	memberIp := internalIP(node)
+	for memberIdentifier, votingMemberIP := range etcdEndpointsConfigMap.Data {
+		if votingMemberIP == memberIp {
+			framework.Logf("found node with IP %s, has identifier [%v]. Removing from cluster...", memberIp, memberIdentifier)
+			removeMember(oc, memberIdentifier)
+			return nil
+		}
+	}
+
+	return fmt.Errorf("removeMemberOfNode could not find any member with IP %s in current endpoint configmap: [%v]",
+		memberIp, etcdEndpointsConfigMap.Data)
+}
+
+func removeMember(oc *exutil.CLI, memberID string) {
+	nodes, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/master="})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	var healthyEtcdPod string
+	for _, node := range nodes.Items {
+		nodeReady := true
+		for _, t := range node.Spec.Taints {
+			if t.Key == "node.kubernetes.io/unreachable" {
+				nodeReady = false
+				break
+			}
+		}
+		if nodeReady {
+			healthyEtcdPod = "etcd-" + node.Name
+			break
+		}
+	}
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	member, err := oc.AsAdmin().Run("exec").Args("-n", "openshift-etcd", healthyEtcdPod, "-c", "etcdctl", "etcdctl", "member", "remove", memberID).Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(member).To(o.ContainSubstring("removed from cluster"))
 }
--- a/test/extended/dr/common.go
+++ b/test/extended/dr/common.go
@@ -601,6 +601,7 @@ func runClusterRestoreScript(oc *exutil.CLI, restoreNode *corev1.Node, backupNod
          sudo chown -R core /home/core/backup
          ls -la /home/core/backup
 
+         export MANIFEST_DIR=/etc/kubernetes/manifests
          sudo /usr/local/bin/cluster-restore.sh /home/core/backup
 
 		 exit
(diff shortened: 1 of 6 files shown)


test/extended/dr/common.go
```diff
@@ -543,3 +529,166 @@ func createPrivatePublicSSHKeySecret(oc *exutil.CLI, name string, namespace stri
...
+            ls -la /etc/kubernetes/manifests/
+            mv /etc/kubernetes/manifests/etcd-pod.yaml /tmp/etcd.pod.yaml
+            # TODO(thomas): wait for containers to stop?
+
+            source /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd.env
+            source /etc/kubernetes/static-pod-resources/etcd-certs/configmaps/etcd-scripts/etcd-common-tools
+            dl_etcdctl
+            rm -rf /var/lib/etcd
+            mkdir -p /var/lib/etcd 
+            etcdctl snapshot restore /home/core/backup/snapshot.db --data-dir=/var/lib/etcd
```
hasbro17, line 620: I'm probably missing some context but wouldn't doing an `etcdctl snapshot restore` create the new member with a new member id and cluster id?
https://etcd.io/docs/v3.5/op-guide/recovery/#restoring-a-cluster
> Restoring overwrites some snapshot metadata (specifically, the member ID and cluster ID); the member loses its former identity. This metadata overwrite prevents the new member from inadvertently joining an existing cluster. Therefore in order to start a cluster from a snapshot, the restore must start a new logical cluster.

So after this restore step we have the older 2 node cluster and a new 1 node cluster right?
So we have the same issue as the recovery script?

If that is the case, then I'm wondering if we can't just stop the static pods on the non-recovery nodes first. And then restore using the script for the recovery host to get a single node etcd cluster. And then delete and recreate the machines.

tjungblu: It depends on what we want to test here. The existing test, this PR disables, does exactly what you mention: 
* Take backup of CP Node 0
* Stop all static pods on non-recovery nodes (Node 0+2)
* Restore backup on CP Node 1, that brings up a single node etcd from snapshot
* Restore API server pods on Node 0+2

Then wait for everything to settle.

What I wanted to start with is a little simpler, akin to reverse vertical scaling:
* Take backup of CP Node 0
* Stop etcd on Node 1, restore from snapshot, restart

It's more similar to the "Replacing an unhealthy etcd member" docs.

https://docs.openshift.com/container-platform/4.12/backup_and_restore/control_plane_backup_and_restore/replacing-unhealthy-etcd-member.html

> So after this restore step we have the older 2 node cluster and a new 1 node cluster right?
> So we have the same issue as the recovery script?

Spot on. I'm simply missing to remove the etcd member before restoring. Gotta figure out how to do that from the terminal (guess with crictl exec or so..)

tjungblu: thank god there was a helper for that already. I adjusted the name of the test to be more clear:

> 

also need to roll a new revision out, which is this step in the docs:

> oc patch etcd cluster -p='{"spec": {"forceRedeploymentReason": "recovery-'"$( date --rfc-3339=ns )"'"}}' --type=merge

otherwise it's stuck because the installer installed a newer revision before the restore script put back an older one (nasty)

hasbro17: > What I wanted to start with is a little simpler, akin to reverse vertical scaling:
> - Take backup of CP Node 0
> - Stop etcd on Node 1, restore from snapshot, restart

> It's more similar to the "Replacing an unhealthy etcd member" docs.

IIUC it seems like we're scaling down the old cluster and restoring and scaling up the new one serially. Although if that's the case then what happens when we restore and restart the 2nd member of the new cluster?
Does it automatically get added to the new cluster? I'm guessing it shares the same configuration as the other member since etcd snapshot restore would also write the WAL that has the configuration.

> It depends on what we want to test here.

Ultimately we would want to test recovery from a quorum loss scenario like the old test which would mean an unavailable control-plane. Currently we're relying on a healthy control-plane to schedule the `etcd-restore-pod` but I think eventually we would need to account for that not being the case. Which from my understanding was why we have the ssh keys setup so we could do that. Or run it as a static pod.

tjungblu: > IIUC it seems like we're scaling down the old cluster and restoring and scaling up the new one serially.

Nope. It's still the same cluster. The "crashing" member is removed and replaced with the same machine, but starting off from a snapshot of another machine.

> Does it automatically get added to the new cluster?

CEO does that automatically, yes. A node with an etcd pod comes up, it's not in the member list and it will be added to the cluster again. Works like a charm, if you disable the quorum guard obviously :sweat_smile: 

> Ultimately we would want to test recovery from a quorum loss scenario like the old test which would mean an unavailable control-plane.

Agreed, let's add this in a follow-up PR. I just wanted to have an end2end that restores one of our easier scenarios to see if the ssh stuff works (it does).

hasbro17: Sounds good. We can focus on the quorum loss scenario in the next follow up to see how we want to coordinate that from within a static restore pod.

test/extended/util/annotate/generated/zz_generated.annotations.go
```diff
@@ -1831,11 +1831,11 @@ var Annotations = map[string]string{
...
 
-	"[sig-etcd][Feature:DisasterRecovery][Disruptive] [Feature:EtcdRecovery] Cluster should recover from a backup taken on one node and recovered on another [apigroup:operator.openshift.io]": " [Serial]",
+	"[sig-etcd][Feature:DisasterRecovery][Disruptive][Disabled:Broken] [Feature:EtcdRecovery] Cluster should recover from a backup taken on one node and recovered on another [apigroup:operator.openshift.io]": " [Serial]",
 
-	"[sig-etcd][Feature:DisasterRecovery][Disruptive] [Feature:EtcdRecovery] Cluster should restore itself after quorum loss [apigroup:machine.openshift.io][apigroup:operator.openshift.io]": " [Serial]",
+	"[sig-etcd][Feature:DisasterRecovery][Disruptive][Disabled:Broken] [Feature:EtcdRecovery] Cluster should restore itself after quorum loss [apigroup:machine.openshift.io][apigroup:operator.openshift.io]": " [Serial]",
 
-	"[sig-etcd][Feature:DisasterRecovery][Suite:openshift/etcd/recovery] [Feature:EtcdRecovery] should install ssh keys on CP nodes": "",
+	"[sig-etcd][Feature:DisasterRecovery][Suite:openshift/etcd/recovery] [Feature:EtcdRecovery][Disruptive] Restore snapshot from node on another single unhealthy node": " [Serial]",
```
hasbro17, line 1838: Actually, I just noticed we have the `[Serial]` tag appended here now. Not sure why it's getting added here but hopefully this doesn't mean it's going to run as part of the serial suite.

tjungblu: @dgoodwin do you happen to know how this ends up here? Is it due to the disruptive tag? 

If it's OK then could you also /approve? I think we got nobody around that has approval rights on the generated file (Michal and David are out during Kubecon).

dgoodwin: Apologies I am not sure if [serial] is from [disruptive] or not, but it seems likely, you will need to check. It seems logical that disruptive tests may require running in serial.

Could you send me a link to a presubmit that ran the test? I can't seem to find it in any.

Depending on how many jobs are expected to run this test, you should probably run some payload commands to make sure this isn't going to unleash havoc once the jira requirement to merge lifts. (payload 4.14 nightly blocking, and payload 4.14 ci blocking)

tjungblu: Thanks @dgoodwin. Test-presubmit is here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-etcd-operator/1041/pull-ci-openshift-cluster-etcd-operator-master-e2e-aws-etcd-recovery/1649350721021153280

as you can see, we disable both of the failing tests with this PR. So should get better and not worse ;-)

We only added it here: https://github.com/openshift/release/pull/37799 for etcd and the CEO repo, so shouldn't impact any payload AFAIU.

dgoodwin: Ah awesome, an aws-etcd-recovery job is perfect, ty!

PR #27876, opened 2023-04-14
Bug OCPBUGS-6778: Correct condition for rejecting connection

gryf:
Test 'Pods cannot access the /config/master API endpoint fails' assumes, that curl output should contain Connection refused, which is not always true, when cluster would be deployed on OpenStack with different CNI (like Kuryr), which is using Security Groups to block the traffic to the internal API. In that case rules created by SG would drop the packet, so there will be no response rather then refuse.

This patch is correcting the test case, where there is added additional response string and added timeout option for curl, just to save the time.


MaysaMacedo:
/retitle Bug OCPBUGS-6778: Correct condition for rejecting connection

gryf:
Hello @deads2k and @mfojtik! is there anything else I need to do regarding this patch to be merged?

PR #27878, opened 2023-04-17
[release-4.10] OCPBUGS-11874: Add Kuryr exception to "pods should successfully create sandboxes" test

dulek:
Kuryr can take long to create Pods as it needs to create ports for these pods in Neutron. If a pod gets deleted while this is being processed, a warning Event will be generated. This causes the `[sig-network] pods should successfully create sandboxes by adding pod to network` to fail.

This commit adds this case as a flake.


dulek:
Note to reviewers, @itzikb-redhat is our QE representative, so this is QE approval.

mandre (commented):
/lgtm
/label backport-risk-assessed

PR #27880, opened 2023-04-19
Handle new tls.CertificateVerificationError type in Go 1.20

ardaguclu:
In [Go 1.20](https://go-review.googlesource.com/c/go/+/449336), x509 certificates are embedded into new `tls.CertificateVerificationError` type(https://pkg.go.dev/crypto/tls#CertificateVerificationError). But currently, request_token in origin only relies onx509 certificates and when Go version is bumped to 1.20, this will no longer be a valid case. That's why, this PR also handles `tls.CertificateVerificationError` as well as other x509 certificates whose are necessary for the backwards compatibility.

But since new type only exists in Go 1.20, this PR should also bump origin images to Go 1.20.

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.19-openshift-4.14
+  tag: rhel-8-release-golang-1.20-openshift-4.14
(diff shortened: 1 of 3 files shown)


ardaguclu:
We don't require this PR currently, but when we want to switch to Go 1.20, we will.

/hold

stlaz:
/hold
I propose using #27883 instead

ardaguclu:
Since the other PR is approved, this PR became redundant;

/close

PR #27884, opened 2023-04-20
test/extended: cpu-partitioning: skip cluster infrastructure for Hypershift

sjenning:
HostedClusters i.e. Hypershift do not use MachineConfigs and do not have the CRD installed, causing `CPU Partitioning cluster infrastructure should be configured correctly` test to fail on HostedClusters.

We should skip this test for HostedClusters.


sjenning:
cc @eggfoobar

PR #27889, opened 2023-04-24
OCPBUGS-12447: Add intervals and test for ovs-vswitchd unreasonably long poll interval

dgoodwin:


dgoodwin:
/hold 

For verification of test results. Ready for review though.

pkg/monitor/intervalcreation/node.go
```diff
@@ -191,6 +200,46 @@ func eventsFromKubeletLogs(nodeName string, kubeletLog []byte) monitorapi.Interv
...
+	if !strings.Contains(logLine, "Unreasonably long") {
+		return nil
+	}
+
+	failureTime := systemdJournalLogTime(logLine)
+	message := logLine[strings.Index(logLine, "ovs-vswitchd"):]
+	return monitorapi.Intervals{
+		{
+			Condition: monitorapi.Condition{
+				Level:   monitorapi.Warning,
```
deepsm007, line 233: Does this needs to be `monitorapi.Error` ?

dgoodwin: I went with warning as that was what the author of the log line chose, I'm not sure if there's a case to call it an error here in and of itself.

deepsm007: Fair enough, lets go with what the author chose.

dgoodwin:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-gcp-ovn 3

Last pod got deleted unexpectedly. :angry:

dgoodwin:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-gcp-ovn 3

dgoodwin:
@deepsm007 need another review, I did another pass now that I see how common this is on GCP in small numbers. If over 5s, we fail, under we flake.

dgoodwin:
Presubmits do not seem to have picked up the right code, but the /payload did, and it worked, but one of the 3 had a 9s poll interval and failed, only test to fail on that job. As such I don't think I can reliably ever fail this test, moving to only flake for now. Hopefully the push makes the system pickup the new code.

dgoodwin:
/close

It can't see the correct commits, seen this problem in CI before. Have to try another PR.

dgoodwin:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-gcp-ovn 3

dgoodwin:
Same problem hit in the other PR, I don't get this, it's checking otu the right commit, and that commit has no reference to the variable the compiler is complaining about at all...

/reopen

dgoodwin:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-gcp-ovn 3

dgoodwin:
Didn't notice my push failed because I didn't include -f after a rebase. :(

/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-gcp-ovn 3

pkg/monitor/intervalcreation/node.go
```diff
@@ -192,6 +201,62 @@ func eventsFromKubeletLogs(nodeName string, kubeletLog []byte) monitorapi.Interv
 	return ret
 }
 
+// eventsFromOVSVswitchdLogs returns the produced intervals.  Any errors during this creation are logged, but
+// not returned because this is a best effort step
+func eventsFromOVSVswitchdLogs(nodeName string, ovsLogs []byte) monitorapi.Intervals {
+	nodeLocator := monitorapi.NodeLocator(nodeName)
+	ret := monitorapi.Intervals{}
+
+	scanner := bufio.NewScanner(bytes.NewBuffer(ovsLogs))
```
deepsm007, line 210: Looks good. Do you think we could use strings to iterate using range loop as we know the logs will always be in new line instead of buffer?

dgoodwin: Do you mean converting the log file byte to a string, splitting it on newlines, then iterating on range over that array?

I suspect that would be much slower and more memory intensive than the byte buffer scanning approach, if it works the way I suspect.  Looking around this seems to be the recommended approach for iterating lines out of file contents in a byte array. Let me know if I've missed anything though.

PR #27890, opened 2023-04-24
add anonymous cert detection test case

rphillips:
Detects a kubelet using only an anonymous user. Kubelet certificates rotate negotiate an OpenShift user instead of using anonymous users.

### DIFF
--- a/pkg/monitor/intervalcreation/node.go
+++ b/pkg/monitor/intervalcreation/node.go
@@ -186,6 +186,7 @@ func eventsFromKubeletLogs(nodeName string, kubeletLog []byte) monitorapi.Interv
 		ret = append(ret, startupProbeError(currLine)...)
 		ret = append(ret, errParsingSignature(currLine)...)
 		ret = append(ret, failedToDeleteCGroupsPath(nodeLocator, currLine)...)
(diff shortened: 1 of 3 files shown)


pkg/monitor/intervalcreation/node.go
```diff
@@ -434,6 +435,26 @@ func failedToDeleteCGroupsPath(nodeLocator, logLine string) monitorapi.Intervals
...
+
+	return monitorapi.Intervals{
+		{
+			Condition: monitorapi.Condition{
+				Level:   monitorapi.Error,
+				Locator: nodeLocator,
+				Message: monitorapi.ReasonedMessage("FailedToAuthenticateWithOpenShiftUser", logLine),
+			},
+			From: failureTime,
+			To:   failureTime.Add(7 * time.Minute),
```
deads2k, line 453: `1*time.Second` will still appear and let us chase these.  Since the message should be repeated regularly, I don't think there's a need for multi-minute interval.

deads2k:
/approve
/lgtm
/hold

holding until the interval is shortened, feel free to re-label and remove the hold after that.

PR #27897, opened 2023-04-26
USHIFT-647: skip non-existing resources from security.openshift.io

chiragkyal:
Closes: [USHIFT-647](https://issues.redhat.com/browse/USHIFT-647)

skip non-existing resources from security.openshift.io

### DIFF
--- a/test/extended/cli/explain.go
+++ b/test/extended/cli/explain.go
@@ -51,13 +51,18 @@ var (
 		"route.openshift.io": {
 			{Group: "route.openshift.io", Version: "v1", Resource: "routes"},
 		},
+		"template.openshift.io": {
+			{Group: "template.openshift.io", Version: "v1", Resource: "templateinstances"},
+		},
+	}
+
+	// This list holds builtin apigroups whose resources are not similar for OpenShift and MicroShift.
+	// Hence test will be performed for individual resource level.
+	builtinExceptionTypes = map[string][]schema.GroupVersionResource{
 		"security.openshift.io": {
 			{Group: "security.openshift.io", Version: "v1", Resource: "podsecuritypolicyreviews"},
 			{Group: "security.openshift.io", Version: "v1", Resource: "podsecuritypolicyselfsubjectreviews"},
 			{Group: "security.openshift.io", Version: "v1", Resource: "podsecuritypolicysubjectreviews"},
-		},
-		"template.openshift.io": {
-			{Group: "template.openshift.io", Version: "v1", Resource: "templateinstances"},
 		},
 	}
 
@@ -335,18 +340,6 @@ var (
 				pattern: `FIELD\: +displayName`,
 			},
 		},
-		"security.openshift.io": {
-			{
-				gv:      schema.GroupVersion{Group: "security.openshift.io", Version: "v1"},
-				field:   "rangeallocations.range",
-				pattern: `FIELD\: +range`,
-			},
-			{
-				gv:      schema.GroupVersion{Group: "security.openshift.io", Version: "v1"},
-				field:   "securitycontextconstraints",
-				pattern: `FIELDS\:.*`,
-			},
-		},
 		"template.openshift.io": {
 			{
 				gv:      schema.GroupVersion{Group: "template.openshift.io", Version: "v1"},
@@ -425,6 +418,23 @@ var (
 		},
 	}
 
+	// This list holds apigroups whose resources are not similar for OpenShift and MicroShift.
+	// Hence test will be performed for individual resource level.
+	specialExceptionTypes = map[string][]explainExceptions{
+		"security.openshift.io": {
+			{
+				gv:      schema.GroupVersion{Group: "security.openshift.io", Version: "v1"},
+				field:   "rangeallocations.range",
+				pattern: `FIELD\: +range`,
+			},
+			{
+				gv:      schema.GroupVersion{Group: "security.openshift.io", Version: "v1"},
+				field:   "securitycontextconstraints",
+				pattern: `FIELDS\:.*`,
+			},
+		},
+	}
+
 	specialNetworkingTypes = []explainExceptions{
 		{
 			gv:      schema.GroupVersion{Group: "network.operator.openshift.io", Version: "v1"},
@@ -450,24 +460,6 @@ var (
 			gv:      schema.GroupVersion{Group: "network.openshift.io", Version: "v1"},
 			field:   "egressnetworkpolicies.spec",
 			pattern: `DESCRIPTION\:.*`,
-		},
-	}
-
-	microShiftTypes = []explainExceptions{
-		{
-			gv:      schema.GroupVersion{Group: "authorization.openshift.io", Version: "v1"},
-			field:   "rolebindingrestrictions.spec",
-			pattern: `FIELDS\:.*`,
-		},
-		{
-			gv:      schema.GroupVersion{Group: "security.internal.openshift.io", Version: "v1"},
-			field:   "rangeallocations.range",
-			pattern: `FIELD\: +range`,
-		},
-		{
-			gv:      schema.GroupVersion{Group: "security.openshift.io", Version: "v1"},
-			field:   "securitycontextconstraints",
-			pattern: `FIELDS\:.*`,
 		},
 	}
 )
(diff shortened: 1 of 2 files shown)


ingvagabund:
@chiragkyal dropped a comment into https://github.com/openshift/origin/pull/27615/files#r1180417677

test/extended/cli/explain.go
```diff
@@ -566,7 +556,7 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 				e2e.Logf("Checking %s...", bt)
 				exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), bt.Resource, groupName)
 				o.Expect(err).NotTo(o.HaveOccurred())
-				if !exist {
+				if !exist && groupName == "security.openshift.io" {
```
ingvagabund, line 559: Skipping this group excludes three resources from testing:
- `podsecuritypolicyreviews`
- `podsecuritypolicyselfsubjectreviews`
- `podsecuritypolicysubjectreviews`

test/extended/cli/explain.go
```diff
@@ -600,14 +590,25 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 				gvr := st.gv.WithResource(resource[0])
 				exist, err := exutil.DoesApiResourceExist(oc.AdminConfig(), resource[0], groupName)
 				o.Expect(err).NotTo(o.HaveOccurred())
-				if !exist {
+				if !exist && groupName == "security.openshift.io" {
```
ingvagabund, line 593: Skipping this group skips testing of `rangeallocations.range` resources which is not tested in the new cycle for `microShiftTypes`.

ingvagabund (commented):
There are other cluster installations where the excluded resources are installed and need to be tested. The problem here is the current test is not implemented to test a subset of resources for a given API group. Also, using a simple `[security.openshift.io]` check will not work either.

test/extended/cli/explain.go
dhellmann: The change only skips if a type within the group does not exist, right? Isn't that exactly what we want in this case, since OpenShift includes more types than MicroShift?

I'm surprised to see the groupName test in the condition, though. @chiragkyal , why is that there?

test/extended/cli/explain.go
dhellmann: If we skip when the API type is not present, we shouldn't need separate lists of types for MicroShift an OpenShift any more at all.

test/extended/cli/explain.go
chiragkyal: The change will skip the entire group if any resource (type) within that group does not exist. This test expects that if a group is present, all of its resources must also be present and that is where the problem lies because for MicroShift we have only `securitycontextconstraints` under `"security.openshift.io"`. So, if we try to skip any resource, it will skip the entire apigroup. Checking for each resource in each apigroup is something has not been done so far. See [here](https://github.com/openshift/origin/pull/27615#discussion_r1180417677) and [here](https://redhat-internal.slack.com/archives/C03DP9PABNC/p1684999750574689) 


As we are having problem only with `"security.openshift.io"` group, I thought to skip this group from here and test it separately under a [new test ](https://github.com/openshift/origin/pull/27897/files#diff-f0e9098abef3f083227a769d2b891489aec36b40e6948db84b464d032c359ed2R602) with `microShiftTypes` but it will again cause problem for OpenShift with partial testing of this group. It's a deadlock situation.

test/extended/cli/explain.go
chiragkyal: It's not skipping for individual resources, but rather skipping the entire group if any of it's resource does not exist. Having separate lists of types for MicroShift and OpenShift also is not helping because the tests are unaware of the cluster type, and it will check both lists of types.

chiragkyal:
@ingvagabund I have now separated out `security.openshift.io` group and added a new test, which will check the presence of each resource and will skip only if that particular resource doesn't exist. This way, the entire group will not be skipped.

Let me know if we can take this path.

test/extended/cli/explain.go
```diff
@@ -425,6 +418,23 @@ var (
 		},
 	}
 
+	// This list holds apigroups whose resources are not similar for OpenShift and MicroShift.
```
dhellmann, line 421: It's not really clear from the comments how the items in this list are different from the items in the builtinExceptionTypes list.

test/extended/cli/explain.go
```diff
@@ -569,6 +561,23 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 		})
 	}
 
+	for group, bets := range builtinExceptionTypes {
+		groupName := group
+		types := bets
+		for _, bet := range types {
+			resourceName := bet.Resource
+			g.It(fmt.Sprintf("should contain spec+status for %s of %s [apigroup:%s]", resourceName, groupName, groupName), func() {
```
dhellmann, line 569:
```suggestion
			g.It(fmt.Sprintf("should contain spec+status for %s of %s, if the type is present [apigroup:%s]", resourceName, groupName, groupName), func() {
```

test/extended/cli/explain.go
```diff
@@ -598,6 +607,25 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 			}
 		})
 	}
+
+	for group, sets := range specialExceptionTypes {
+		groupName := group
+		types := sets
+		for _, set := range types {
+			resourceName := strings.Split(set.field, ".")[0]
+			g.It(fmt.Sprintf("should contain proper fields description for %s of %s [apigroup:%s]", resourceName, groupName, groupName), func() {
```
dhellmann, line 616:
```suggestion
			g.It(fmt.Sprintf("should contain proper fields description for %s of %s, if the type is present [apigroup:%s]", resourceName, groupName, groupName), func() {
```

dhellmann (commented):
The implementation looks OK. I have a couple of questions about details, and 2 suggestions for changing some message text. The message text changes might change the test suite output in a way we don't want, so I'll defer to other reviewers on those.

test/extended/cli/explain.go
```diff
@@ -51,14 +51,19 @@ var (
 		"route.openshift.io": {
 			{Group: "route.openshift.io", Version: "v1", Resource: "routes"},
 		},
+		"template.openshift.io": {
```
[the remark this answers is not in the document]
chiragkyal: No, I believe. There are quite a few other APIs also in this list which are not present in MicroShift. If the APIs are not at all present, entire API group will be skipped. 

The `builtinExceptionTypes` list below holds API groups which are present both in OpenShift and MicroShift but have different resources installed.

test/extended/cli/explain.go
```diff
@@ -51,14 +51,19 @@ var (
...
+			{Group: "template.openshift.io", Version: "v1", Resource: "templateinstances"},
+		},
+	}
+
+	// This list holds builtin apigroups whose resources are not similar for OpenShift and MicroShift.
```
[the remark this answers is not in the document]
chiragkyal: That is right. For that reason, test will be performed for each resource within a group.

test/extended/cli/explain.go
chiragkyal: The reasoning is same for both the list. 

As `oc explain` doc doesn't contain `spec+status` for a few apigroups, they were listed separately in `specialTypes` list, and tested against a special `field` and `pattern`. 

`specialExceptionTypes` is just taken out from `specialTypes` list due to same reason as `builtinExceptionTypes` 

If there is any suggestion, pls let me know. I'll update it.

test/extended/cli/explain.go
chiragkyal: `type` can confuse between group and resource. How about `if the resource is present` ?

test/extended/cli/explain.go
chiragkyal: added `if the resource is present`

dhellmann:
It looks like the jobs are failing to merge some auto-generated code

```
Merge conflict in test/extended/util/annotate/generated/zz_generated.annotations.go
```

test/extended/cli/explain.go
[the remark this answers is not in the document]
dhellmann: OK, that makes sense.

dhellmann (commented):
This looks OK to me. As I commented earlier, I think it needs the generated code to be updated to allow it to pass the tests, but the logic looks alright.

/lgtm

@soltysh , we discussed the approach Chirag is taking here on slack.

/assign @soltysh

chiragkyal:
@dhellmann all tests are green now.

soltysh (approved):
/approve

PR #27898, opened 2023-04-26
trt-900: log risk-analysis errors but don't pass back up

neisw:
* When we don't find any test failure summary files add a log entry and return instead of continuing to process and returning an error back up the chain.
* When we encounter an error we want to log it but don't pass back up and cause the job to fail


neisw:
/payload-job periodic-ci-openshift-release-master-ci-4.14-e2e-aws-sdn-techpreview

PR #27899, opened 2023-04-27
OCPBUGS-13379: machines: add a test which verifies number of node reboots

vrutkovs:
This adds a new tests which checks `journalctl --list-boots` to ensure no unexpected reboots occurred on the nodes.

Its similar to `should report ready nodes the entire duration of the test run` except that its applicable to SNO - when the only node reboots it won't be recorded as a change in prometheus metric.

TODO:
* [x] Calculate number of expected boots based on rendered-* MC
* [x] Make it run on SNO(+workers) clusters only

### DIFF
--- a/test/extended/machines/cluster.go
+++ b/test/extended/machines/cluster.go
@@ -1,22 +1,41 @@
 package operators
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"strings"
+	"time"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
 	"github.com/stretchr/objx"
 
+	authv1 "github.com/openshift/api/authorization/v1"
+	configv1 "github.com/openshift/api/config/v1"
+	projv1 "github.com/openshift/api/project/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	"k8s.io/utils/strings/slices"
 
 	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/image"
 	"github.com/openshift/origin/test/extended/util/prometheus"
+)
+
+const (
+	namespace          = "e2e-machines-testbed"
+	serviceAccountName = "e2e-machines"
 )
 
 var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should", func() {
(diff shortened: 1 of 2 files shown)


test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+			actual[node.Name] = nodeReboots
+		}
+
+		o.Expect(errs).To(o.BeEmpty())
+		o.Expect(actual).To(o.Equal(expected))
+	})
 })
+
+// getNodeRole reads node labels and returns either "worker" or "master"
+func getNodeRole(node *corev1.Node) string {
```
benluddy, line 191: Do SNO nodes have multiple node-role labels?

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+}
+
+func getNumberOfBootsForNode(kubeClient kubernetes.Interface, nodeName string, attempt int) (int, error) {
+
+	// Run journalctl to collect a list of boots
+	command := "exec chroot /host journalctl --list-boots"
+	isTrue := true
+	zero := int64(0)
+	ctx := context.Background()
+	name := fmt.Sprintf("list-boots-%s-%d", nodeName, attempt)
```
benluddy, line 208: Any chance of this producing a name that is too long?

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+						{
+							MountPath: "/host",
+							Name:      "host",
+						},
+					},
+				},
+			},
+		},
+	}, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
```
benluddy, line 259: If a terminated pod is deleted by the pod garbage collector, couldn't we end up with out-of-order attempt numbers (and does that matter)?

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+		return getNumberOfBootsForNode(kubeClient, nodeName, attempt+1)
```
benluddy, line 260: Is there a limit on recursive attempts?

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+	err := callRBAC(ctx, oc, false)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	err = callServiceAccount(ctx, oc, false)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	err = callProject(ctx, oc, false)
+	o.Expect(err).NotTo(o.HaveOccurred())
+}
+
+func callRBAC(ctx context.Context, oc *exutil.CLI, create bool) error {
+	obj := &authv1.ClusterRoleBinding{
```
benluddy, line 333: If this were a RoleBinding, it looks like all of the test fixtures would be namespace-scoped and you could rely on namespace finalization to clean them up.

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+	var err error
+	if create {
+		_, err = client.Create(ctx, obj, metav1.CreateOptions{})
+	} else {
+		err = client.Delete(ctx, obj.Name, metav1.DeleteOptions{})
+	}
+
+	if apierrors.IsAlreadyExists(err) || apierrors.IsNotFound(err) {
+		return nil
+	}
```
benluddy, line 409: I'm a little uncomfortable with a bool parameter toggling between create and delete as there isn't much reuse between the two paths.

Not that I imagine this specific test running on MicroShift, but with MicroShift we now have a variant that doesn't serve the project API at all. I'd expect a "create project" call to return 404 in that case.

It might be convenient to use https://github.com/openshift/origin/blob/626c3ff421d2c9f169765bb4487187cca3a627cf/test/extended/util/client.go#L134 and create only the ServiceAccount and RoleBinding fixtures.

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+		}
+
+		// List nodes, set actual and expected number of reboots
+		for _, node := range nodes.Items {
+			// Examine only the nodes which are available at the start of the test
+			if !slices.Contains(staticNodeNames, node.Name) {
+				continue
+			}
+			expectedRebootsForNodeRole, ok := expectedReboots[getNodeRole(&node)]
+			o.Expect(ok).To(o.BeTrue())
```
benluddy, line 176: `o.HaveKey` would give a more helpful failure message than a bool equality comparison.

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+
+	// Count number of boots
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		return linesInPodLogs, fmt.Errorf("failed to copy information from podLogs to buf: %v", err)
+	}
+	linesInPodLogs = strings.Count(buf.String(), "\n") - 1
+	if err != nil {
+		return linesInPodLogs, fmt.Errorf("failed to fetch boot list from %s pod logs: %v", name, err)
```
benluddy, line 305: If the line count return value is valid when a non-nil error is also returned, it's surprising that it can be a negative number. I think it would be clearer if the calling code were to omit the map entry for nodes that hit an error.

test/extended/machines/cluster.go
vrutkovs: I don't think it possible to add more labels to masters, but you can have SNO + worker which acts as infra node

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+					Name: "list-boots",
+					SecurityContext: &corev1.SecurityContext{
+						RunAsUser:  &zero,
+						Privileged: &isTrue,
+					},
+					Image: image.ShellImage(),
+					Command: []string{
+						"/bin/bash",
+						"-c",
+						command,
```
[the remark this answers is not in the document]
vrutkovs: Its probably an artifact from debugging. I'll check if plain command works

test/extended/machines/cluster.go
vrutkovs: Possibly this test could run twice, then garbage-collected pod may indeed change the attempts order.

That won't change the --list-boots output and won't affect test results

test/extended/machines/cluster.go
vrutkovs: Right, its not looking great. I think its best to remove this flag, create just the namespace-scoped objects and rely on namespace finalization instead

test/extended/machines/cluster.go
```diff
@@ -100,4 +119,293 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+	g.It("should verify that nodes have exactly two boots [Late]", func() {
+		ctx := context.Background()
+
+		// This test is applicable for SNO installations only
+		configClient, err := configv1client.NewForConfig(oc.AdminConfig())
+		o.Expect(err).NotTo(o.HaveOccurred())
+		infrastructure, err := configClient.ConfigV1().Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if infrastructure.Status.ControlPlaneTopology != configv1.SingleReplicaTopologyMode {
+			return
```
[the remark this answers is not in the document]
vrutkovs: Seems there is no "skip on HA" tag, other tests are doing similar infra checks:
* https://github.com/openshift/origin/blob/82bbfa8b3ba8e480e885baa7819764c8f9c5b978/test/extended/apiserver/resiliency.go#L40
* https://github.com/openshift/origin/blob/82bbfa8b3ba8e480e885baa7819764c8f9c5b978/test/extended/etcd/helpers/helpers.go#L487

test/extended/machines/cluster.go
vrutkovs: I don't think its possible - node names are limited to 64 chars and pod names can have 253 chars

test/extended/machines/cluster.go
vrutkovs: Good point, limited attempts to 10, checked at the top of the procedure

test/extended/machines/cluster.go
vrutkovs: good idea

test/extended/machines/cluster.go
vrutkovs: Good idea, update this to:
* return 0 if error occurred
* include pod logs in the error message
* throw error of number of boots is less than 1

deads2k:
added a commit in https://github.com/openshift/origin/pull/27993 to cleanup some comments.

vrutkovs:
Closed in favor of #27993

PR #27901, opened 2023-04-28
ETCD-423: Restore Test - Lost Quorum

tjungblu:
scenario:
* backup of node 0
* copy backup from 0 to 1
* shutdown node 0 and 2 (remove their static pods)
* restore on node 1

### DIFF
--- a/test/extended/dr/common.go
+++ b/test/extended/dr/common.go
@@ -13,10 +13,15 @@ import (
 	"text/tabwriter"
 	"time"
 
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	"github.com/openshift/library-go/test/library"
+	"github.com/openshift/origin/test/e2e/upgrade"
+	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/image"
+	"github.com/stretchr/objx"
 	xssh "golang.org/x/crypto/ssh"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -24,21 +29,17 @@ import (
 	applyappsv1 "k8s.io/client-go/applyconfigurations/apps/v1"
 	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
 	applymetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
-	"k8s.io/kubernetes/test/e2e/framework"
-
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/test/e2e/framework"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	e2essh "k8s.io/kubernetes/test/e2e/framework/ssh"
-
-	g "github.com/onsi/ginkgo/v2"
-	o "github.com/onsi/gomega"
-	"github.com/openshift/origin/test/e2e/upgrade"
-	exutil "github.com/openshift/origin/test/extended/util"
-	"github.com/stretchr/objx"
+	"k8s.io/kubernetes/test/utils"
 )
 
 const (
-	operatorWait      = 15 * time.Minute
+	operatorWait      = 25 * time.Minute
 	defaultSSHTimeout = 5 * time.Minute
 
 	sshPath = "/tmp/ssh"
@@ -141,7 +142,7 @@ func waitForOperatorsToSettle() {
 	available := make(map[string]struct{})
 	lastErr = nil
 	var lastCOs []objx.Map
-	wait.PollImmediate(30*time.Second, operatorWait, func() (bool, error) {
+	_ = wait.PollImmediate(30*time.Second, operatorWait, func() (bool, error) {
 		obj, err := coc.List(context.Background(), metav1.ListOptions{})
 		if err != nil {
 			lastErr = err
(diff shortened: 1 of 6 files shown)


test/extended/dr/common.go
```diff
@@ -337,6 +336,40 @@ func ssh(cmd string, node *corev1.Node) (*e2essh.Result, error) {
 	return e2essh.IssueSSHCommandWithResult(cmd, e2e.TestContext.Provider, node)
 }
 
+// deleteEtcdPodsOnNodes will find etcd static pods running on the nodes and remove their mirror pods from the apiserver.
+// This is used to clean left-over state, which has been stored in a snapshot and does not accurately reflect reality anymore.
+func deleteEtcdPodsOnNodes(client kubernetes.Interface, nodes []*corev1.Node) error {
```
tjungblu, line 341: this is not needed anymore since the OVN steps are added, because kubelet was missing apiserver connection to update the status.

I'll leave this here for now, in case we may need it for another test. We can do a cleanup run after the feature branch closes.

test/extended/dr/common.go
```diff
@@ -649,26 +681,162 @@ EOF
...
+          # running this process async to get a chance to finish the pod gracefully
+          ssh -i $P_KEY -o StrictHostKeyChecking=no -q core@${i} sudo systemctl restart kubelet.service &
+        done
+
+        exit`, sshKeyDance, strings.Join(nodeIPs, " "))
+
+	podSpec := applycorev1.PodSpec().WithHostNetwork(true).WithRestartPolicy(corev1.RestartPolicyOnFailure)
+	podSpec.Containers = []applycorev1.ContainerApplyConfiguration{
+		*applycorev1.Container().
+			WithName("cluster-restore").
```
hasbro17, line 789: Nit:
```suggestion
			WithName("ovn-repair").
```

test/extended/dr/common.go
```diff
@@ -692,6 +860,21 @@ func runPodAndWaitForSuccess(oc *exutil.CLI, pod *applycorev1.PodApplyConfigurat
...
 
+func runPod(oc *exutil.CLI, pod *applycorev1.PodApplyConfiguration) error {
+	err := e2epod.DeletePodWithWaitByName(oc.AdminKubeClient(), *pod.Name, *pod.Namespace)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	_, err = oc.AdminKubeClient().CoreV1().Pods(*pod.Namespace).Apply(ctx, pod, metav1.ApplyOptions{FieldManager: *pod.Name, Force: true})
```
hasbro17, line 871: Is the `Force: true` here to overwrite the pod if the previous named pod exists?

test/extended/dr/common.go
```diff
@@ -649,26 +681,162 @@ EOF
...
+func runOVNRepairCommands(oc *exutil.CLI, restoreNode *corev1.Node, nonRecoveryNodes []*corev1.Node) error {
+	const name = "etcd-ovn-repair-pod"
+	const namespace = "openshift-etcd"
+	framework.Logf("running ovn-repair script on node: %v", restoreNode.Name)
+
+	var nodeIPs []string
+	for _, n := range nonRecoveryNodes {
+		nodeIPs = append(nodeIPs, internalIP(n))
+	}
+	// adding this node last is important for kubelet restarts, as this node runs below pod
```
hasbro17, line 765: Nit: For my own clarification, this is so the ovn-repair pod can restart its underlying kubelet at the very end right?
If so, then maybe a slight clarification:
```suggestion
	// adding the restore node last is important for kubelet restarts, as the ovn-repair pod runs on top of that node
```

Also what happens if this isn't the case? The ovn-repair pod goes away after it restarts its own kubelet?

test/extended/dr/common.go
```diff
@@ -649,26 +681,162 @@ EOF
...
+		#!/bin/bash
+        set -exuo pipefail
+		
+		# ssh key dance
+		%s
+
+        NODE_IPS=( %s )
+		for i in "${NODE_IPS[@]}"; do
+		  echo "removing etcd static pod on [$i]"
+		  ssh -i $P_KEY -o StrictHostKeyChecking=no -q core@${i} sudo rm -rf /etc/kubernetes/manifests/etcd-pod.yaml
```
hasbro17, line 707: I guess this should be enough. Was thinking whether the staticpod controller would re-add the etcd pod manifest but I think that would only happen with a revision rollout.

hasbro17 (approved):

hasbro17:
Looks good overall, just a few nits and questions.
Also curious to see a test run.

/test e2e-aws-etcd-recovery

test/extended/dr/common.go
tjungblu: applied

test/extended/dr/common.go
tjungblu: I think it's not necessary, removed

test/extended/dr/common.go
tjungblu: applied. 

> Also what happens if this isn't the case? The ovn-repair pod goes away after it restarts its own kubelet?

correct. I believe we can remove the pod soon. Dan confirmed it's enough to just remove the ovnkube master pods. I'm just waiting for Allen's changes, so we can verify the watch resume more easily.

test/extended/dr/common.go
tjungblu: indeed, it does not. We force the revision rollout later in the code

tjungblu:
okay that's an odd error:
> failed: (41m5s) 2023-05-04T11:31:23 "[sig-etcd][Feature:DisasterRecovery][Suite:openshift/etcd/recovery] [Feature:EtcdRecovery][Disruptive] Recover with snapshot with two unhealthy nodes and lost quorum [Serial]"


seems the API didn't come back up after the restore. I've only tested the one in isolation against cluster-bot so far, maybe it's a bad interaction with the previous test that passed:

> passed: (13m28s) 2023-05-04T10:50:18 "[sig-etcd][Feature:DisasterRecovery][Suite:openshift/etcd/recovery] [Feature:EtcdRecovery][Disruptive] Restore snapshot from node on another single unhealthy node [Serial]"


I'll run the whole suite against one...

tjungblu:
from the couple of runs, I need to:

* make the API calls (delete+apply) for the pods more resilient, they fail on transient errors 
* at the end of the test we need to wait until the revisions stabilize, that seems to fail right now as a post test

Otherwise it looks great, the disruption tests seem quite spammy - not sure how I can make them less verbose on the api failures for this test.

tjungblu:
I'm overriding the jobs that are not really impacted by this, we only change the recovery presubmit that doesn't have any side-effects.

/override ci/prow/e2e-gcp-ovn

PR #27902, opened 2023-04-28
e2e: node: testcase to exercise device manager recovery

ffromani:
When kubelet recovers pods, there's no guarantee about order of recovering. This means a pod requesting devices can be recovered, and can start, before the device plugin governing the device is started.

The end result is that containers can be admitted and start running without the requested device actually available, causing runtime errors in the workload. This is u/s issue
https://github.com/kubernetes/kubernetes/issues/109595

With this commit, we add a e2e test, focusing on the most critical SNO scenario, which exercise the flow and verify the system behavior is fixed and workload pods requesting devices are rejected (not admitted) if they are recovered before the relevant device plugin.

### DIFF
--- a/test/extended/devices/recovery.go
+++ b/test/extended/devices/recovery.go
@@ -1,0 +1,420 @@
+package devices
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	kubeletdevicepluginv1beta1 "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	k8simage "k8s.io/kubernetes/test/utils/image"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+// This test has a very high disruption potential: it wants to bypass the node cordoning et. al.
+// and kubernetes management and trigger a reboot straight from the node.
+// Additionally, it explicitly considers SNO, where the disruption potential is even higher.
+// Because of all the above, it must opted-in explicitly by supplying the `DEVICE_RECOVERY_TARGET_NODE`
+// variable pointing to a node in the cluster, SNO or worker node.
+
+const (
+	workloadCommand = "devs=$(ls /tmp/ | egrep '^Dev-[0-9]+$'); [ -z $devs ] && { echo MISSING; exit 1; }; { echo devices=$devs; sleep 24h; }" // sleep forever in the test timescale
+
+	setupRegistrationCommandMCD = "mkdir -p /rootfs/var/lib/kubelet/device-plugins/sample && touch /rootfs/var/lib/kubelet/device-plugins/sample/registration"
+	// it's just a file plus just a directory. We reverse the step to be on the safer side and minimize the negative effects on bugs.
+	// note the file is expected to be gone, so we must tolerate the rm to fail. In any case, rmdir should never fail.
+	cleanupRegistrationCommandMCD = "rm /rootfs/var/lib/kubelet/device-plugins/sample/registration; rmdir /rootfs/var/lib/kubelet/device-plugins/sample"
+	unblockRegistrationCommandMCD = "rm /rootfs/var/lib/kubelet/device-plugins/sample/registration"
+	rebootNodeCommandMCD          = "chroot /rootfs systemctl reboot"
+
+	sampleDeviceResourceName            = "example.com/resource"
+	sampleDevicePluginPodName           = "sample-device-plugin"
+	sampleDeviceEnvVarNamePluginSockDir = "PLUGIN_SOCK_DIR"
+
+	sampleDeviceEnvVarNameControlFile = "REGISTER_CONTROL_FILE"
+	sampleDevicePluginControlFilePath = "/var/lib/kubelet/device-plugins/sample/registration"
+
+	targetNodeEnvVar = "DEVICE_RECOVERY_TARGET_NODE"
+
+	// this is the only timeout tunable because is a hard sleep. Every other long timeout is within a polling loop.
+	rebootCooldownEnvVar  = "REBOOT_COOLDOWN_TIME"
+	rebootCooldownDefault = 1 * time.Minute
+)
+
+var (
+	sampleDevicePluginPodLabel = map[string]string{
+		"device-recovery-test-pod": "device-plugin",
+	}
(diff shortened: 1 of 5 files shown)


ffromani:
/hold
still WIP

ffromani:
the PR is now completed. cc @Tal-or @swatisehgal

test/extended/devices/recovery.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+func unblockRegistrationControlOnNodeOrFail(oc *exutil.CLI, node *corev1.Node) {
+	framework.Logf("Preparing node %q for sample device plugin registration", node.Name)
+	fwk := oc.KubeFramework()
+	_, err := exutil.ExecCommandOnMachineConfigDaemon(fwk.ClientSet, oc, node, []string{
+		"sh", "-c", unblockRegistrationCommandMCD,
+	})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	framework.Logf("Prepared node %q for sample device plugin registration", node.Name)
```
swatisehgal, line 372: NIT: Log messages here need to be updated to capture that they for unblocking the device plugin registration process.

swatisehgal:
Other than the minor NIT, the PR looks good to me. Thanks for your work on this!

test/extended/devices/recovery.go
ffromani: good point. Will add to the existing log.

ffromani: my fault, I misread. Fixed the log entry.

ffromani:
addressed the review commants and functionally complete, but still on hold because further test is pending to adjust the timeouts. On cloud SNOs the timeouts are sufficient, but we want to be able to run the tests on BM, on which reboot can take long time.

test/extended/devices/recovery.go
```diff
@@ -0,0 +1,416 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/image"
+)
+
+const (
+	workloadCommand               = "devs=$(ls /tmp/ | egrep '^Dev-[0-9]+$') && echo devices=$devs && sleep 24h" // sleep forever in the test timescale
```
ffromani, line 28: this should fail hard if no devices are present.

test/extended/devices/recovery.go
```diff
@@ -0,0 +1,416 @@
...
+		}
+		framework.Logf("reboot cooldown time: %v", rebootCooldownTime)
+
+		fwk := oc.KubeFramework()
+		namespace := fwk.Namespace.Name // shortcut
+		k8sCli := oc.KubeClient()
+
+		// phase1: figure out the node and prepare it
+		node, err := k8sCli.CoreV1().Nodes().Get(context.Background(), targetNode, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred(), "error getting the target node %q", targetNode)
```
Tal-or, line 87: NIT: usually we use `ToNot` which is more readable
So we read:
Except error _to not_ have occurred other 
instead of:
Except error _not to_ have occurred

test/extended/devices/recovery.go
```diff
@@ -0,0 +1,416 @@
...
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/image"
+)
+
+const (
+	workloadCommand               = "devs=$(ls /tmp/ | egrep '^Dev-[0-9]+$') && echo devices=$devs && sleep 24h" // sleep forever in the test timescale
+	setupRegistrationCommandMCD   = "mkdir -p /rootfs/var/lib/kubelet/device-plugins/sample && touch /rootfs/var/lib/kubelet/device-plugins/sample/registration"
+	unblockRegistrationCommandMCD = "rm /rootfs/var/lib/kubelet/device-plugins/sample/registration"
```
Tal-or, line 30: let's remove also the `sample` directory because it's not part of the original content of the node

Tal-or (commented):
Looking very good.
One question I've though - in case of failure for any reason, can we make sure (or at least try) to delete the registration file and `sample` dir from the node?

test/extended/devices/recovery.go
```diff
@@ -0,0 +1,420 @@
...
+	sampleDeviceResourceName            = "example.com/resource"
+	sampleDevicePluginPodName           = "sample-device-plugin"
+	sampleDeviceEnvVarNamePluginSockDir = "PLUGIN_SOCK_DIR"
+
+	sampleDeviceEnvVarNameControlFile = "REGISTER_CONTROL_FILE"
+	sampleDevicePluginControlFilePath = "/var/lib/kubelet/device-plugins/sample/registration"
+
+	sampleDevicePluginPodDSYAML = "test/e2e/testing-manifests/sample-device-plugin/sample-device-plugin-control-registration.yaml"
+
+	targetNodeEnvVar = "DEVICE_RECOVERY_TARGET_NODE"
```
yanirq, line 45: where does the actual setup of the DEVICE_RECOVERY_TARGET_NODE env var happens ? in selective tests ?  
will the test suits running for this PR actually include that or just skip the tests?

test/extended/devices/recovery.go
ffromani: Should be fixed now, but requires verification.

test/extended/devices/recovery.go
ffromani: Absolutely, no nit at all! thanks for pointing this out. Fixed in the last upload

test/extended/devices/recovery.go
ffromani: The entity (human or program) running the test is in charge of setting the environment variable, otherwise the test will always skip. Thing is, didn't find obvious way to pick a good target automatically, so having the explicit selection requirement seem the lesser evil. I'm open to suggestions though.

ffromani:
all comments addressed. Will unhold once the tuning of the timeouts is completed.

ffromani:
/hold cancel

fixed the timeouts, further refinements are non-blocking

swatisehgal (commented):
/lgtm

ffromani:
so, running this test will cause `openshift-tests` to emit
```
[sig-arch] unknown image: registry.k8s.io/e2e-test-images/sample-device-plugin:1.5 (container/sample-device-plugin reason/Pulled image/registry.k8s.io/e2e-test-images/sample-device-plugin:1.5)
Writing JUnit report to /home/fromani/tmp/junit/junit_e2e__20230510-143327.xml

Suite run returned error: failed because an invariant was violated, 1 pass, 0 skip (5m24s)

error: failed because an invariant was violated, 1 pass, 0 skip (5m24s)
```

Which is because the added test consumes `http://registry.k8s.io/e2e-test-images/sample-device-plugin:1.5`.
I'm going to investigate further why this image doesn't have a `ImageID` u/s I can use in `test/utils/image` (per steps in `https://github.com/openshift/origin/blob/master/test/extended/util/image/README.md`). Likely this is because the image is used by `e2e_node` - not `e2e` tests. But I'll doublecheck u/s

ffromani:
/retest

PR had no changes and these lanes passed before

ffromani:
after almost 1 year, I think this is not gonna make it. Too bad.

PR #27910, opened 2023-05-09
[WIP] Add missing error check in sysctl allowlist test

mlguerrero12:
This prevents the test to pass when the pod is not running. A bug was not detected due to this missing check.


mlguerrero12:
yes, just wanted to check that the test is failing after adding the missing check. Fix got merged in master.

PR #27913, opened 2023-05-10
OCPBUGS-6586: oc idle: increase wait timeouts to 60 sec

ardaguclu:
For blocking jobs, 30 second timeout value is sufficient. However, these test fails on OpenStack by
getting timeout error. This PR increases oc idle tests timeout to 60 seconds to pass
on OpenStack as mentioned in referenced bug.


ardaguclu:
/cc @neisw 
could you please take a look at this one?. Thanks.

PR #27838, opened 2023-03-31
new disruption backend 

tkashem:

PR #27919, opened 2023-05-11
add shutdown interval for apiserver from events

tkashem:


pkg/monitor/api.go
```diff
@@ -60,6 +61,7 @@ func Start(ctx context.Context, restConfig *rest.Config, additionalEventInterval
 	startPodMonitoring(ctx, m, client)
 	startNodeMonitoring(ctx, m, client)
 	startEventMonitoring(ctx, m, client)
+	shutdown.StartMonitoringGracefulShutdownEvents(ctx, m, client)
```
dgoodwin, line 64: All the upgrade presubmits look unhealthy in an unusual way, is it this change blocking something or just bad luck?

tkashem: i had successful runs before, i pushed some changes after that, so maybe my changes or just bad luck, investigating...

tkashem: >  is it this change blocking something
yes, i didn't realize that, i just fixed it and pushed new changes

pkg/monitor/shutdown/consumer.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	Reason                       = "GracefulShutdownWindow"
+	KubeAPIServerShutdownLocator = "shutdown/apiserver"
```
tkashem, line 17: any suggestion on the locator name?

pkg/monitor/shutdown/consumer.go
```diff
@@ -0,0 +1,160 @@
...
+		condition := incomplete(event)
+		framework.Logf("GracefulShutdownEvent - Incomplete: %+v", event)
+		framework.Logf(condition.Message)
+		intervalID := c.monitor.StartInterval(startedAt, condition)
+		c.monitor.EndInterval(intervalID, startedAt.Add(time.Second))
+	}
+}
+
+func key(event *corev1.Event) string {
+	// we sue the pod namespace/name as the unique key, this ensures that:
```
xueqzhan, line 103: spelling, use not sue

pkg/monitor/shutdown/consumer.go
```diff
@@ -0,0 +1,160 @@
...
+	defer c.lock.Unlock()
+
+	var ok bool
+	if _, ok := c.processed[event.UID]; ok {
+		return
+	}
+
+	key := key(event)
+	var start, end *corev1.Event
+	start, ok = c.byHost[key]
```
xueqzhan, line 65: Any chance that the start event could be missing and then mess up the logic (treating previous end as new start)?

pkg/monitor/shutdown/consumer.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	end = event
+
+	delete(c.byHost, key)
+	c.processed[start.UID] = struct{}{}
+	c.processed[end.UID] = struct{}{}
+
+	condition := complete(start, end)
+	// these logs can be useful to debug if the windows are displayed correctly
+	framework.Logf("GracefulShutdownEvent - Start: %+v", start)
```
xueqzhan, line 78: Should this be logged when start is collected to maintain correct timestamp?

pkg/monitor/shutdown/consumer.go
xueqzhan: Locator name seems to be fine for me. Maybe rename the variable KubeAPIServerShutdownLocator to APIServerShutdownLocator since this is not specific to kube-apiserver. Nit though.

pkg/monitor/shutdown/consumer.go
tkashem: I added some comments to make it more clear, let me know if it helps

pkg/monitor/shutdown/consumer.go
tkashem: I am logging the pair next to each other so we can see how the window was constructed, and the interval timestamp should still be correct since we take the time from the  'CreationTimestamp' field of the event.

tkashem:
merge conflict, I had to rebase on top of https://github.com/openshift/origin/pull/27838, no changes to this PR, it needs an lgtm

PR #27930, opened 2023-05-19
push intervals to loki

dgoodwin:
This change is part of a process to get intervals charts, a key debugging tool, more robust and powerful. Once live, sippy will be able to pull intervals from loki instead of gcs buckets, which gives us more powerful searching, but also the ability to intermingle actual logs from the cluster. We will also be able to search intervals across all CI jobs.

Pushing intervals to loki is best effort. If we fail, we log and continue.

- Obtain bearer token for loki upload
- Push intervals to loki in batches
- Detect bad http codes
- Fix an edge case where locators parse badly for loki unpack
- Add retries for batches of intervals going to loki
- Fix discrepancies with interval creation funcs
- Final hookup to run after we write intervals to disk

[TRT-994](https://issues.redhat.com//browse/TRT-994)

### DIFF
--- a/pkg/monitor/loki_upload.go
+++ b/pkg/monitor/loki_upload.go
@@ -10,17 +10,22 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
 	batchSize = 500
 )
 
+var batchCtr int
+
 func pushBatch(client *http.Client, bearerToken string, lokiData LokiData) error {
-	logrus.Infof("pushing batch of %d intervals", len(lokiData.Streams[0].Values))
+	batchCtr++
 	url := "https://logging-loki-openshift-operators-redhat.apps.cr.j7t7.p1.openshiftapps.com/api/logs/v1/openshift-trt/loki/api/v1/push"
 	headers := map[string]string{
 		"Content-Type":  "application/json",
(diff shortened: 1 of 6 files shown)


dgoodwin:
This was caused by mistakenly base64 encoding my env vars, fixed in https://github.com/openshift/release/pull/39746

dgoodwin:
Success! https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fpull-ci-openshift-origin-master-e2e-gcp-ovn%2F1661788419493203968%5C%22,src%3D%5C%22origin-interval%5C%22%7D%20%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221685030476353%22,%22to%22:%221685066476356%22%7D%7D

pkg/monitor/loki_upload.go
```diff
@@ -0,0 +1,285 @@
...
+type Streams struct {
+	Stream map[string]string `json:"stream"`
+	Values [][]interface{}   `json:"values"` // array of arrays of [ts, jsonStr]
+}
+
+type LokiData struct {
+	Streams []Streams `json:"streams"`
+}
+
+func obtainSSOBearerToken(clientID, clientSecret string, intervals monitorapi.Intervals) (string, error) {
```
neisw, line 243: doesn't look like 'intervals' is needed

pkg/monitor/loki_upload.go
```diff
@@ -0,0 +1,285 @@
...
+
+			logLine[tag] = val
+		}
+
+		// Convert the log data to a packed json log line, where the message is in '_entry'.
+		// This allows us to search additional fields we didn't index as labels using the unpack/json
+		// directives in logQL.
+		logLineJson, err := json.Marshal(logLine)
+		if err != nil {
+			logrus.WithError(err).Errorf("unable to unmashall log line: %s", logLine)
```
neisw, line 190: Do you want to continue here if logLineJson isn't valid or intentionally append the presumably nil logline?

pkg/monitor/loki_upload.go
```diff
@@ -0,0 +1,285 @@
...
+	jobName, defined := os.LookupEnv("JOB_NAME")
+	if !defined {
+		return fmt.Errorf("JOB_NAME env var is not defined")
+	}
+	buildID, defined := os.LookupEnv("BUILD_ID")
+	if !defined {
+		return fmt.Errorf("BUILD_ID env var is not defined")
+	}
+
+	bearerToken, err := obtainSSOBearerToken(clientID, clientSecret, intervals)
```
neisw, line 121: Any concerns about calling this for each batch?  Presume we won't have 10s or 100s of batches but if we did would it be a problem / unnecessary to call SSO api in a tight(ish) loop?

pkg/monitor/loki_upload.go
```diff
@@ -0,0 +1,285 @@
...
+	// This format matches the labels used for the cluster logs in loki, and we want this
+	// to be labelled the same.
+	invoker := fmt.Sprintf("openshift-internal-ci/%s/%s", jobName, buildID)
+	if invoker == "" {
+		return fmt.Errorf("OPENSHIFT_INSTALL_INVOKER is needed for the invoker loki label")
+	}
+	labels := map[string]string{
+		"invoker": invoker,
+		"type":    "origin-interval",
+		// TODO: in future, may group intrvals in origin rather than in the js itself, if we do,
```
neisw, line 143: nit, spelling for intervals

pkg/monitor/loki_upload.go
```diff
@@ -0,0 +1,285 @@
...
+			parts := strings.Split(lp, "/")
+			tag, val := parts[0], parts[1]
+			if strings.TrimSpace(tag) == "" {
+				// Have seen this fail on: WARN[0002] unable to process locator tag: May 12 11:10:00.000 I /machine-config reason/OperatorVersionChanged clusteroperator/machine-config-operator version changed from [{operator 4.14.0-0.nightly-2023-05-03-163151}] to [{operator 4.14.0-0.nightly-2023-05-12-121801}]
+				// Would be interesting to track down where this is coming from. Should be identification.go OperatorLocator, but this is impossible.
+				logrus.Warnf("unable to process locator tag: %+v", i)
+				continue
+			}
+
+			// WAARNING: opting for a potentially risky change here, I want namespace filtering to be available as a
```
neisw, line 171: nit, extra a in warning (might be intentionally spooky and if so leave it ;-)

pkg/monitor/loki_upload.go
dgoodwin: Good catch, also should read marshall not unmarshall, thanks!

dgoodwin:
Updated for feedback, thanks @neisw.

I think this is ready, my plan is to let it merge and make then use search.ci to see if we get any hits on the error logging in here. Hoping no problems hit with rate limiting or obtaining too many sso tokens, and if they do the errors are non-fatal.

neisw:
/lgtm

/hold
remove whenever your are ready

dgoodwin:
/hold cancel

Will monitor for merge results today.

PR #27936, opened 2023-05-24
Mitigate explain output assertion regex

ardaguclu:
explain command's output was changed slightly in upstream. For example, `spec	<DeploymentSpec>` is printed instead of `spec	<Object>` (even though passing `--output=plaintext-openapiv2` still preserves the current behavior). However, our `oc explain` tests in origin have a regex expression compatible with the old behavior and they are failing in rebase 1.27 PR.

This PR mitigates regex expression to make it compatible with both versions simulateneously.


ardaguclu:
I'm closing this one in favor of the PR @atiratree has been preparing
/close

PR #27950, opened 2023-05-26
OCPBUGS-14033: statefulsets are logging numerous events but the pod does get created

rphillips:
Statefulset controller is logging numerous events for the openshift-monitoring prometheus pod. The pod does get created, so setup a filter for the test.

Related to: https://github.com/kubernetes/kubernetes/issues/89067


simonpasquier:
/hold

https://github.com/openshift/oauth-proxy/pull/255 and https://github.com/openshift/prometheus-operator/pull/234 should fix the Prometheus pods being reported as failed on eviction. Let's wait a bit and see if the failure persists.

simonpasquier:
/close

The 2 changes mentioned in my previous comment seem to fix the issue.

PR #27955, opened 2023-05-31
OCPBUGS-14321: Increase timeout in sysctl allowlist test

mlguerrero12:
This test sometimes fails because the allowlist controller takes more than 30s to update the sysctl allowlist file. This controller waits up to 60s for the pods of the DS to be in ready state.


soltysh (approved):
/approve

PR #27956, opened 2023-05-31
OCPBUGS-14342: Increase timeout in sysctl allowlist test

mlguerrero12:
This test sometimes fails because the allowlist controller takes more than 30s to update the sysctl allowlist file. This controller waits up to 60s for the pods of the DS to be in ready state.

This is a cherry-pick of #27955


mlguerrero12:
@bparees, could you please help me with the cherry-pick-approved label here? There is nothing for qe to do with this fix.

PR #27958, opened 2023-05-31
upgrade distribution to v3

flavianmissi:
the same upgrade has landed on library-go. to ensure we're keeping things compatible, we're upgrading distribution to v3 everywhere that uses it.


flavianmissi:
failures do not seem related to changes.
/retest

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/RangelReale/osincli v0.0.0-20160924135400-fababb0555f2
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/davecgh/go-spew v1.1.1
-	github.com/docker/distribution v2.8.1+incompatible
+	github.com/distribution/distribution/v3 v3.0.0-20230530204932-ba46c769b3d1
```
soltysh, line 10: I'd expect at least library-go be updated along this one, to match with the one that has docker distribution v3.

soltysh (changes requested):
/hold

go.mod
flavianmissi: gave a stab at it, let's see what breaks.

flavianmissi: Looks like all of the odd unit test failures I was seeing locally are not seen by CI 🎉

soltysh (commented):
Comments:
- squash `vendor: update openshift libraries vendored files` and `vendor: update distribution vendored files` together, those are all generated changes and it's enough to keep them in a single commit;
- squash `go.mod,go.sum: upgrade to distribution v3` and `go.mod: bump openshift libraries` together as well;

the rest looks good, but you need to rebase

flavianmissi:
rebased and squased similar commits together, PTAL.

soltysh (approved):
/lgtm
/approve

PR #27960, opened 2023-05-31
Remove sampler

deads2k:
Remove the sampler and replace with observing the beginning and end.  It eliminates extraneous events and makes it possible to have maps inside of the intervals themselves, which for annotations feels very natural.  Without this, conditions cannot have map[string]string annotations.


deads2k:
/hold

not working rigth yet.

deads2k:
node lifecycle looks right.  pending pods with starting open timeframes look wrong.

pkg/monitor/intervalcreation/node.go
```diff
@@ -82,32 +62,32 @@ func IntervalsFromEvents_NodeChanges(events monitorapi.Intervals, _ monitorapi.R
...
-			openInterval(state, "OperatingSystemUpdate", event.From)
+			intervals = append(intervals, nodeStateTracker.closeInterval(nodeLocator, drainState, simpleCondition(monitorapi.Info, monitorapi.NodeUpdateReason, fmt.Sprintf(msgPhaseDrain, roles)), event.From)...)
+			nodeStateTracker.openInterval(nodeLocator, osUpdateState, event.From)
 		case "Reboot":
-			closeInterval(state, "Drain", monitorapi.Info, event.Locator, fmt.Sprintf(msgPhaseDrain, roles), event.From)
-			closeInterval(state, "OperatingSystemUpdate", monitorapi.Info, event.Locator, fmt.Sprintf(msgPhaseOSUpdate, roles), event.From)
-			openInterval(state, "Reboot", event.From)
+			intervals = append(intervals, nodeStateTracker.closeInterval(nodeLocator, drainState, simpleCondition(monitorapi.Info, monitorapi.NodeUpdateReason, fmt.Sprintf(msgPhaseDrain, roles)), event.From)...)
+			intervals = append(intervals, nodeStateTracker.closeInterval(nodeLocator, osUpdateState, simpleCondition(monitorapi.Info, monitorapi.NodeUpdateReason, fmt.Sprintf(msgPhaseOSUpdate, roles)), event.From)...)
+			nodeStateTracker.openInterval(nodeLocator, rebootState, event.From)
```
dgoodwin, line 83: Is it necessary to have this embedded logic about one reason implying closing intervals for other reasons? Any changes in the product and we're immediately broken. Could we instead just close any matching open interval?

pkg/monitor/intervalcreation/state_tracker.go
```diff
@@ -0,0 +1,147 @@
+package intervalcreation
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
```
dgoodwin, line 10: Unit tests might be good here, I know it's getting tested at a higher level but are we certain it's enough coverage?

dgoodwin:
/lgtm

Looks to be working well enough. We can refine if needed but this will unblock #27947

PR #27964, opened 2023-06-02
USHIFT-1300: Avoid getting cluster state for TEST_PROVIDER env var

pacevedom:
/cc @soltysh 
/cc @ingvagabund

### DIFF
--- a/pkg/test/ginkgo/test_runner.go
+++ b/pkg/test/ginkgo/test_runner.go
@@ -362,9 +362,12 @@ func updateEnvVars(envs []string) []string {
 	}
 	// copied from provider.go
 	// TODO: add error handling, and maybe turn this into sharable helper?
+	config := &exutilcluster.ClusterConfiguration{}
 	clientConfig, _ := framework.LoadConfig(true)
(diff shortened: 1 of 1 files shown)


pkg/test/ginkgo/test_runner.go
```diff
@@ -362,9 +362,12 @@ func updateEnvVars(envs []string) []string {
 	}
 	// copied from provider.go
 	// TODO: add error handling, and maybe turn this into sharable helper?
+	config := &exutilcluster.ClusterConfiguration{}
 	clientConfig, _ := framework.LoadConfig(true)
 	clusterState, _ := exutilcluster.DiscoverClusterState(clientConfig)
```
ingvagabund, line 367: What if the error return by `exutilcluster.DiscoverClusterState` is valid?

soltysh (approved):
/lgtm
/approve

pkg/test/ginkgo/test_runner.go
soltysh: Yeah, we need to add proper error handling :sweat_smile: just like that TODO says

PR #27965, opened 2023-06-02
OCPBUGS-32484: chart template: limit tooltip width to first word

vrutkovs:
Two quality-of-life improvements for e2e charts:

* wrap tooltips in e2e chart to make sure that tooltips don't overflow X when a long label is used.
image
* copy segment label contents to clipboard when its clicked

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -510,6 +510,9 @@
         createTimelineData(pathologicalEvents, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isInterestingOrPathological, regex)
 
         var segmentFunc = function (segment) {
+            // Copy label to clipboard
+            navigator.clipboard.writeText(segment.labelVal);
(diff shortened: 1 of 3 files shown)


vrutkovs:
Should be resolved now: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27965/pull-ci-openshift-origin-master-e2e-agnostic-ovn-cmd/1700201695096606720/artifacts/e2e-agnostic-ovn-cmd/openshift-e2e-test/artifacts/junit/e2e-timelines_spyglass_20230908-185556.html

/retest

dgoodwin:
We need this. Can you hit the other non-spyglass template in the same directory as well? SOMEBODY came along and made a second impl and now we get to maintain two forever! :)

vrutkovs:
>Can you hit the other non-spyglass template

Done

PR #27976, opened 2023-06-10
add event interval from rest client metric broken down by source

tkashem:


pkg/monitor/clientview/gatherer.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+type gatherer struct {
+	start  time.Time
+	client prometheusv1.API
+}
+
+func (g *gatherer) query(ctx context.Context) (prometheustypes.Value, error) {
+	// TODO: should we include 5xx? "429|5..|<error>"
+	query := `rest_client_requests_total{code =~ "429|5..|<error>"}`
```
vrutkovs, line 57: I wonder if we can add DNS resolve error - or it won't be recorded?

tkashem: yes, when Michal's upstream PR merges, we can incorporate it.

dinhxuanvu:
The upstream PR: https://github.com/kubernetes/kubernetes/pull/115357

pkg/monitor/clientview/gatherer.go
tkashem: to be more clear, if dns error happens code=<error>, any other network error, code=<error> also. (we can't make the distinction today with rest_client_requests_total. Michal's PR adds a new metric just to capture dns error.

dinhxuanvu:
@p0lyn0mial @dgrisonnet Would like to have your inputs on this PR. Abu indicates this PR is ready for merge. We can either merge it and I can create subsequent PRs to address feedbacks or incorporate those changes into linked Vadim's PR.

dgrisonnet:
I think we should close this one and wrap up the work on the PR Vadim opened and improved.

PR #27986, opened 2023-06-19
add event interval from rest client metric broken down by source

vrutkovs:
Make sure locator and message are shown as is on HTML page when `rest_` increases.

Test bed for https://github.com/openshift/origin/pull/27976 + fixes

TODO:
* [ ] Metrics are scraped every 30 sec, is there a way to find how long disruptions during that period took? Seems its possible to find that out from `rest_client_request_duration_seconds_sum`?
     Currently we'll create sequential one second intervals for each error found, but exact time and duration of the disruption cannot be properly derived

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -200,6 +200,38 @@
         return false
     }
 
+    function isClientInternalLBEvent(eventInterval) {
+        if (eventInterval.locator.includes("client/APIError source/internal-lb")) {
+            return true
+        }
+
+        return false
+    }
+
+    function isClientExternalLBEvent(eventInterval) {
+        if (eventInterval.locator.includes("client/APIError source/external-lb")) {
+            return true
+        }
+
+        return false
+    }
+
+    function isClientServiceNetworkEvent(eventInterval) {
+        if (eventInterval.locator.includes("client/APIError source/service-network")) {
+            return true
+        }
+
+        return false
+    }
+
+    function isClientLocalhostEvent(eventInterval) {
+        if (eventInterval.locator.includes("client/APIError source/localhost")) {
+            return true
+        }
+
+        return false
+    }
+
     function isEndpointConnectivity(eventInterval) {
         if (!eventInterval.message.includes("reason/DisruptionBegan") && !eventInterval.message.includes("reason/DisruptionSamplerOutageBegan")){
             return false
@@ -373,6 +405,10 @@
         return [item.locator, "", "GracefulShutdownWindow"]
     }
 
+    function clientErrorEventsValue(item) {
+        return [item.locator, "", "ClientError"]
+    }
+
     function getDurationString(durationSeconds) {
         const seconds = durationSeconds % 60;
         const minutes = Math.floor(durationSeconds/60);
@@ -500,6 +536,18 @@
 
         timelineGroups.push({group: "shutdown-events", data: []})
         createTimelineData(apiserverShutdownEventsValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isAPIServerShutdownEventActivity, regex)
+
+        timelineGroups.push({group: "client-internal-lb", data: []})
+        createTimelineData(clientErrorEventsValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isClientInternalLBEvent, regex)
+
+        timelineGroups.push({group: "client-external-lb", data: []})
+        createTimelineData(clientErrorEventsValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isClientExternalLBEvent, regex)
+
+        timelineGroups.push({group: "client-service-network", data: []})
+        createTimelineData(clientErrorEventsValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isClientServiceNetworkEvent, regex)
+
+        timelineGroups.push({group: "client-localhost", data: []})
+        createTimelineData(clientErrorEventsValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isClientLocalhostEvent, regex)
 
         timelineGroups.push({group: "e2e-test-failed", data: []})
         createTimelineData("Failed", timelineGroups[timelineGroups.length - 1].data, eventIntervals, isE2EFailed, regex)
--- a/pkg/defaultmonitortests/types.go
+++ b/pkg/defaultmonitortests/types.go
@@ -26,6 +26,7 @@ import (
 	"github.com/openshift/origin/pkg/monitortests/storage/legacystoragemonitortests"
 	"github.com/openshift/origin/pkg/monitortests/testframework/additionaleventscollector"
 	"github.com/openshift/origin/pkg/monitortests/testframework/alertanalyzer"
+	"github.com/openshift/origin/pkg/monitortests/testframework/clientresterroranalyzer"
 	"github.com/openshift/origin/pkg/monitortests/testframework/clusterinfoserializer"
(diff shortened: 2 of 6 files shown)


vrutkovs:
I'm experimenting with query range to see if we can get disruptions for each second from metrics, which are scraped every 30 sec. Not sure if its possible at all, I'll add a TODO

vrutkovs:
/cc @dinhxuanvu @mfojtik

Lets merge it today so that by mid-next week we'd see if that gives valid signal

dinhxuanvu:
@tkashem PTAL

pkg/monitor/clientview/gatherer.go
```diff
@@ -0,0 +1,177 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
+	"github.com/openshift/library-go/test/library/metrics"
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+)
+
+var scrapePeriod = 30 * time.Second
+
+func FetchEventIntervalsForRestClientError(ctx context.Context, config *rest.Config, startTime time.Time) ([]monitorapi.EventInterval, error) {
```
dinhxuanvu, line 26: Look like you need `monitorserialization` package in order to use `EventInterval`.

dinhxuanvu: Look like a typo. Fixed.

dinhxuanvu:
@tkashem @dgoodwin PTAL

tkashem:
I do't see the timeline anymore: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27986/pull-ci-openshift-origin-master-e2e-gcp-ovn-rt-upgrade/1688504328878297088

dinhxuanvu:
> I do't see the timeline anymore: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/27986/pull-ci-openshift-origin-master-e2e-gcp-ovn-rt-upgrade/1688504328878297088

They are back!

dinhxuanvu:
@tkashem @dgoodwin PTAL. Would like to have this PR merge if there are no further issues.

pkg/monitor/clientview/gatherer.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	matrix := result.(prometheustypes.Matrix)
+	for _, series := range matrix {
+		if len(series.Values) <= 1 {
+			continue
+		}
+		previous := series.Values[0].Value
+		for _, current := range series.Values[1:] {
+			if !previous.Equal(current.Value) {
```
dgoodwin, line 101: Am I correct that if we're experiencing client errors over a prolonged period of time, every metric sample that we increment results in it's own interval? How far apart are the timestamps on each value? Every 15s? 

I'm concerned about bounding here, if a bad job is going to generate a million intervals? That would break a bunch of things and cost us money likely too.

For example with disruption we record a single interval for a prolonged period of disruption, we record when we start seeing disruption (and what error message), and then watch for it to stop or for the error message to change. Either event results in the interval now having a To time, so we create it.

IMO you should apply similar here such that a prolonged uninterrupted period of is just one interval. Granted this may get complicated, maybe we have one value where it doesn't change and then it starts going up again, that would make a new interval, but that would be better than nothing. Depends on how far apart the prometheus samples are as well.

pkg/invariants/testframework/alertanalyzer/invariant.go
```diff
@@ -25,6 +27,9 @@ func (w *alertSummarySerializer) StartCollection(ctx context.Context, adminRESTC
 
 func (w *alertSummarySerializer) CollectData(ctx context.Context, storageDir string, beginning, end time.Time) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error) {
 	intervals, err := fetchEventIntervalsForAllAlerts(ctx, w.adminRESTConfig, beginning)
+	clientEventIntervals, err2 := clientview.FetchEventIntervalsForRestClientError(ctx, w.adminRESTConfig, beginning)
```
dgoodwin, line 30: These are not conceptually linked to the alert analyzer tests right? How hard is it to break these out into their own InvariantTest? You just need an adminRESTConfig, should be pretty easy.

pkg/monitor/clientview/gatherer.go
vrutkovs: > every metric sample that we increment results in it's own interval? 

No, not exactly. We're going to create one interval for every scrape period (30s) if number of client errors has increased.

> I'm concerned about bounding here, if a bad job is going to generate a million intervals?

No, one interval every 30s at most.

pkg/invariants/testframework/alertanalyzer/invariant.go
vrutkovs: Right, its best to be moved its own InvariantTest

pkg/monitor/clientview/gatherer.go
dgoodwin: Ok one ever 30s, but is that also per time series returned by your promql? (it looks like that might catch a fair bit) Example if that promql returns 500 time series and we have a bad job run with a 30min problem, thats 500 * 30 * 2 or 30k intervals, which would double the number I'd expect normally. 

If the error counter for a time series increments at t, t+30s, t+60s, t+90s, I think it's probably worth the effort to make that one interval as we do for disruption. That would involve tracking start points, watching for a sample that didn't increment and using that as a trigger to terminate that interval and add it to the list, and terminating any intervals that were dangling at the end of the job run.

vrutkovs: Its per time series, yes, but I don't expect it to be significant. Its "error code" * "source namespace", so in the worst case its 10 error codes * 50 namespaces * (30 mins job / 30 secs interval) = 250 intervals for "every component firing 10 different errors throught the whole test duration)

dgoodwin: Unless I'm missing something, 10 error codes x 50 namespaces x 3 api points x 60 (30 minutes at 30s intervals) = 90,000, not 750? I know it's an extreme example by even 10% of that is a lot more for our JS UI to handle. I think you need to batch based on consecutive failed samples.

vrutkovs: Ah, you're correct. Right, this needs interval batching

vrutkovs: Implemented, now sequential intervals are merged into one:
```
  Aug 23 08:39:20.111: INFO: [client-rest-error-serializer] adding new interval Aug 23 08:36:58.784 - 1s    E client/APIError source/service-network node/10.128.0.197 namespace/openshift-apiserver component/openshift-apiserver client observed an API error - previous=1 current=5 rest_client_requests_total{apiserver="openshift-apiserver", code="503", container="openshift-apiserver", endpoint="https", host="172.30.0.1:443", instance="10.128.0.197:8443", job="api", method="GET", namespace="openshift-apiserver", pod="apiserver-5785c87bf8-9skb4", prometheus="openshift-monitoring/k8s", service="api"}
...
  Aug 23 08:39:20.111: INFO: [client-rest-error-serializer] updated existing interval Aug 23 08:36:58.784 - 30s   E client/APIError source/service-network node/10.128.0.197 namespace/openshift-apiserver component/openshift-apiserver client observed an API error - previous=1 current=5 rest_client_requests_total{apiserver="openshift-apiserver", code="503", container="openshift-apiserver", endpoint="https", host="172.30.0.1:443", instance="10.128.0.197:8443", job="api", method="GET", namespace="openshift-apiserver", pod="apiserver-5785c87bf8-9skb4", prometheus="openshift-monitoring/k8s", service="api"}
```

PR #27987, opened 2023-06-20
Changing the severity of "missing runbook_url annotation for critical alerts" test case from flaky to failure

vimalk78:
The following critical alerts do not have have runbook_url annotation added as exceptions:
  - OVNKubernetesNorthboundDatabaseClusterIDError
  - OVNKubernetesSouthboundDatabaseClusterIDError
  - OVNKubernetesNorthboundDatabaseLeaderError
  - OVNKubernetesSouthboundDatabaseLeaderError
  - OVNKubernetesNorthboundDatabaseMultipleLeadersError
  - OVNKubernetesSouthboundDatabaseMultipleLeadersError
  - OVNKubernetesNorthdInactive
  - KubeSchedulerDown
  - MultipleDefaultStorageClasses
  - MachineAPIOperatorMetricsCollectionFailing
  - MCDRebootError
  - ExtremelyHighIndividualControlPlaneMemory
  - HAProxyDown
  - ClusterOperatorDown
  - ClusterVersionOperatorDown

The check for runbook_url validity has been separated out into a new flaky test

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -67,6 +67,44 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
 		"SystemMemoryExceedsReservation",
 	)
 
+	criticalAlertsMissingRunbookURLExceptions := sets.NewString(
+		// Repository: https://github.com/openshift/cluster-network-operator
+		// Issue: https://issues.redhat.com/browse/OCPBUGS-14062
+		"OVNKubernetesNorthboundDatabaseClusterIDError",
+		"OVNKubernetesSouthboundDatabaseClusterIDError",
+		"OVNKubernetesNorthboundDatabaseLeaderError",
+		"OVNKubernetesSouthboundDatabaseLeaderError",
+		"OVNKubernetesNorthboundDatabaseMultipleLeadersError",
+		"OVNKubernetesSouthboundDatabaseMultipleLeadersError",
+		"OVNKubernetesNorthdInactive",
+
+		// Repository: https://github.com/openshift/cluster-kube-scheduler-operator
+		// Issue: https://issues.redhat.com/browse/OCPBUGS-14052
+		"KubeSchedulerDown",
+
+		// Repository; https://github.com/openshift/cluster-storage-operator
+		// Issue: https://issues.redhat.com/browse/OCPBUGS-14053
+		"MultipleDefaultStorageClasses",
+
+		// Repository: https://github.com/openshift/machine-api-operator
+		// Issue: https://issues.redhat.com/browse/OCPBUGS-14055
+		"MachineAPIOperatorMetricsCollectionFailing",
+
+		// Repository: https://github.com/openshift/machine-config-operator
+		// Issue: https://issues.redhat.com/browse/OCPBUGS-14056
+		"MCDRebootError",
+		"ExtremelyHighIndividualControlPlaneMemory",
+
+		// Repository: https://github.com/openshift/cluster-ingress-operator
+		// Issue: https://issues.redhat.com/browse/OCPBUGS-14057
+		"HAProxyDown",
+
+		// Repository: https://github.com/openshift/cluster-version-operator
+		// Issue: https://issues.redhat.com/browse/OCPBUGS-14246
+		"ClusterOperatorDown",
+		"ClusterVersionOperatorDown",
+	)
+
 	var alertingRules map[string][]promv1.AlertingRule
 	oc := exutil.NewCLIWithoutNamespace("prometheus")
 
@@ -151,6 +189,10 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
 
 	g.It("should have a runbook_url annotation if the alert is critical", func() {
 		err := helper.ForEachAlertingRule(alertingRules, func(alert promv1.AlertingRule) sets.String {
+			if criticalAlertsMissingRunbookURLExceptions.Has(alert.Name) {
+				framework.Logf("Critical alerting rule %q is known to have missing runbook_url.", alert.Name)
+				return nil
+			}
 			violations := sets.NewString()
 			severity := string(alert.Labels["severity"])
 			runbook := string(alert.Annotations["runbook_url"])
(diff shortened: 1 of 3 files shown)


vimalk78:
should the check for invalid `runbook_url` have its own list of exceptions?

test/extended/prometheus/prometheus.go
```diff
@@ -159,7 +201,22 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
...
+
+	g.It("should have valid runbook URLs", func() {
+		err := helper.ForEachAlertingRule(alertingRules, func(alert promv1.AlertingRule) sets.String {
+			violations := sets.NewString()
+			runbook := string(alert.Annotations["runbook_url"])
+
+			if runbook != "" {
 				// If there's a 'runbook_url' annotation, make sure it's a
 				// valid URL and that we can fetch the contents.
 				if err := helper.ValidateURL(runbook, 10*time.Second); err != nil {
```
simonpasquier, line 222: I would separate the test to perform 1) the semantic validation of the URL (e.g. it can be parsed by url.Parse() and the scheme is "http" or "https") and  2) the actual verification that the URL returns a 200 response code. The former can fail the test if it detects an invalid URL and the latter should only flake as the target might be temporarily unavailable.

vimalk78: The `helper.ValidateURL` function uses `url.Parse(rawURL)` and returns error if it fails to parse.
can i add check for "https" scheme in the same validation function?

https://github.com/openshift/origin/blob/master/test/extended/util/prometheus/helpers.go#L355

simonpasquier: >  can i add check for "https" scheme in the same validation function?

yes and I would split the actual request to the URL into another function (e.g. QueryURL()).

test/extended/util/prometheus/helpers.go
```diff
@@ -349,18 +349,38 @@ func FetchAlertingRules(promURL, bearerToken string) (map[string][]promv1.Alerti
...
 
-// ValidateURL takes a URL as a string and a timeout.  The URL is
+func ValidateURL(rawURL string) error {
+	var u *url.URL
+	var err error
+	if u, err = url.Parse(rawURL); err != nil {
+		return err
+	}
+
+	if u.Scheme != "https" {
```
simonpasquier, line 359: "http" should be fine too.

test/extended/util/prometheus/helpers.go
```diff
@@ -349,18 +349,38 @@ func FetchAlertingRules(promURL, bearerToken string) (map[string][]promv1.Alerti
...
+		errstr := fmt.Sprintf("ValidateURL(%q) URL scheme is not https", rawURL)
```
simonpasquier, line 360:
```suggestion
		errstr := fmt.Sprintf("%q: URL scheme is invalid, it should be 'http' or 'https'", rawURL)
```

test/extended/util/prometheus/helpers.go
```diff
@@ -349,18 +349,38 @@ func FetchAlertingRules(promURL, bearerToken string) (map[string][]promv1.Alerti
...
+		framework.Logf(errstr)
+		return fmt.Errorf(errstr)
+	}
+	if u.Host == "" {
+		errstr := fmt.Sprintf("ValidateURL(%q) URL does not have a host part", rawURL)
```
simonpasquier, line 365:
```suggestion
		errstr := fmt.Sprintf("%q: host should not be empty", rawURL)
```

test/extended/util/prometheus/helpers.go
```diff
@@ -349,18 +349,38 @@ func FetchAlertingRules(promURL, bearerToken string) (map[string][]promv1.Alerti
...
+func ValidateURL(rawURL string) error {
+	var u *url.URL
+	var err error
+	if u, err = url.Parse(rawURL); err != nil {
+		return err
+	}
+
+	if u.Scheme != "https" {
+		errstr := fmt.Sprintf("ValidateURL(%q) URL scheme is not https", rawURL)
+		framework.Logf(errstr)
```
simonpasquier, line 361: do we need to log it here since it's already done by the caller?

test/extended/util/prometheus/helpers.go
```diff
@@ -349,18 +349,38 @@ func FetchAlertingRules(promURL, bearerToken string) (map[string][]promv1.Alerti
...
+		return fmt.Errorf(errstr)
+	}
+	if u.Host == "" {
+		errstr := fmt.Sprintf("ValidateURL(%q) URL does not have a host part", rawURL)
+		framework.Logf(errstr)
```
simonpasquier, line 366: same question here

test/extended/prometheus/prometheus.go
```diff
@@ -174,8 +253,6 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
 		})
 
 		if err != nil {
-			// We are still gathering data on how many alerts need to
-			// be fixed, so this is marked as a flake for now.
 			testresult.Flakef(err.Error())
```
simonpasquier, line 256:
```suggestion
			// We can't fail the test because the runbook URLs might be temporarily unavailable.
			// At least we can manually check the CI logs to identify buggy URLs.
			testresult.Flakef(err.Error())
```

test/extended/prometheus/prometheus.go
```diff
@@ -159,13 +201,50 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
...
+
+			return violations
+		})
+
+		if err != nil {
+			e2e.Failf(err.Error())
+		}
+	})
+
+	g.It("should have valid runbook URLs", func() {
```
simonpasquier, line 214:
```suggestion
	g.It("should link to a valid URL if the runbook_url annotation is defined", func() {
```

test/extended/prometheus/prometheus.go
```diff
@@ -159,13 +201,50 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
...
+		err := helper.ForEachAlertingRule(alertingRules, func(alert promv1.AlertingRule) sets.String {
+			violations := sets.NewString()
+			runbook_url := string(alert.Annotations["runbook_url"])
+
+			if runbook_url != "" {
```
simonpasquier, line 219: (nit) easier to read.
```suggestion
			if runbook_url == "" {
				return nil
			}
```

test/extended/prometheus/prometheus.go
```diff
@@ -159,13 +201,50 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
...
+				// If there's a 'runbook_url' annotation, make sure that it is a valid URL
+				if err := helper.ValidateURL(runbook_url); err != nil {
+					violations.Insert(
+						fmt.Sprintf("WARNING: Alert %q has an 'runbook_url' annotation: %q which is not valid. Error: %v",
```
simonpasquier, line 223: ForEachAlertingRule() always include the name of the alert.
```suggestion
						fmt.Sprintf("has an 'runbook_url' annotation which is not valid: %v",
```

test/extended/prometheus/prometheus.go
```diff
@@ -159,13 +201,50 @@ var _ = g.Describe("[sig-instrumentation][Late] OpenShift alerting rules [apigro
...
+			violations := sets.NewString()
+			runbook_url := string(alert.Annotations["runbook_url"])
+
+			if runbook_url != "" {
+				// If there's a 'runbook_url' annotation, make sure that we can fetch the contents.
+				if err := helper.QueryURL(runbook_url, 10*time.Second); err != nil {
 					violations.Insert(
-						fmt.Sprintf("WARNING: Alert %q has an invalid 'runbook_url' annotation: %v",
-							alert.Name, err),
+						fmt.Sprintf("WARNING: Alert %q has an 'runbook_url' annotation: %q which cannot be fetched. Error: %v",
```
simonpasquier, line 246: same comment about the alert name already being included
```suggestion
						fmt.Sprintf("has a runbook URL which cannot be fetched: %v",
```

test/extended/prometheus/prometheus.go
simonpasquier: (nit) to be more specific and distinguish from the next test
```suggestion
	g.It("should link to an HTTP(S) location if the runbook_url annotation is defined", func() {
```

simonpasquier (commented):
one last suggestion

simonpasquier:
@vimalk78 you need to regenerate some files to fix `ci/prow/verify`.

simonpasquier (commented):
/lgtm

PR #27989, opened 2023-06-20
USHIFT-741: MicroShift invariant adaptions 

chiragkyal:
[USHIFT-741](https://issues.redhat.com/browse/USHIFT-741)


copejon:
Changes here seem good @chiragkyal

chiragkyal:
> Changes here seem good @chiragkyal

@copejon could you please add the label. Thanks!

pkg/synthetictests/disruption.go
```diff
@@ -48,6 +49,19 @@ func testServerAvailability(
 	connType := monitorapi.DisruptionConnectionTypeFrom(locatorParts)
 	backendName := fmt.Sprintf("%s-%s-connections", disruptionName, connType)
 	if jobType == nil {
+		// check for MicroShift cluster
+		kubeClient, err := framework.LoadClientset(true)
+		if err != nil {
+			panic(err)
+		}
+		isMicroShift, err := exutil.IsMicroShiftCluster(kubeClient)
```
ingvagabund, line 57: Instead of giving empty list of junit test cases, we should ask ourselves whether we need to run `testServerAvailability` at all. Checking the code:
```sh
$ grep -rn "testServerAvailability"
pkg/synthetictests/disruption.go:37:func testServerAvailability(
pkg/synthetictests/disruption.go:190:		ret = append(ret, testServerAvailability("sig-network-edge", locator, allDisruptionEventsIntervals, jobRunDuration, jobType)...)
pkg/synthetictests/disruption.go:209:		ret = append(ret, testServerAvailability("sig-trt", locator, allDisruptionEventsIntervals, jobRunDuration, jobType)...)
```
which leads to https://github.com/openshift/origin/blob/master/pkg/synthetictests/event_junits.go#L38-L39:
```go
	tests = append(tests, TestAllIngressBackendsForDisruption(events, duration, jobType)...)
	tests = append(tests, TestExternalBackendsForDisruption(events, duration, jobType)...)
```

Question here is whether it is safe to skip `TestAllIngressBackendsForDisruption` and `TestExternalBackendsForDisruption`? If the answer is not, then instead of checking whether an installation is MicroShift one should check whether the ingress/externalbackends are present.

ingvagabund:
`IsMicroShiftCluster` defined in https://github.com/openshift/origin/pull/27996/files#diff-35a89a7a7362642eebb559fb8564e857b00d6f7dd6322c3adabaf1adbd609d35R2097 should be used only in very limited cases. Where there's no other way.

pkg/synthetictests/alerts.go
```diff
@@ -59,6 +61,20 @@ func testAlerts(events monitorapi.Intervals,
...
+		if err != nil {
+			panic(err)
+		}
+	}
+	nsExist, err := exutil.IsNamespaceExist(kubeClient, "openshift-monitoring")
+	if err != nil {
+		panic(err)
+	}
+	if !nsExist {
+		return []*junitapi.JUnitTestCase{}
```
ingvagabund, line 75: Better to avoid invoking `testAlerts` completely in https://github.com/openshift/origin/blob/master/pkg/synthetictests/event_junits.go#L54-L55. Question here to ask is whether there's a sufficient replacement for absence of the monitoring stack when checking the alerts in MicroShift. Does it make sense to e.g. run an external monitoring stack (or its simplified version)?

pkg/synthetictests/disruption.go
chiragkyal: All these disruptions depend upon `JobType` and for MicroShift this will always be nil. So, I think we don't want to run any of these disruption invariants for MicroShift. 

And to make this happen, I see the possible way is either call `IsMicroShiftCluster()` inside `testServerAvailability` and return an empty list of junit test cases (which I've done above), or 

not to append the following to `tests` based on the same `IsMicroShiftCluster()` check inside `event_junits.go` 
```go
	tests = append(tests, TestAllAPIBackendsForDisruption(events, duration, jobType)...)
	tests = append(tests, TestAllIngressBackendsForDisruption(events, duration, jobType)...)
	tests = append(tests, TestExternalBackendsForDisruption(events, duration, jobType)...)
```

/cc @pacevedom @dhellmann  let me know your thoughts.

pkg/synthetictests/alerts.go
chiragkyal: >Better to avoid invoking testAlerts completely in https://github.com/openshift/origin/blob/master/pkg/synthetictests/event_junits.go#L54-L55. 

Based on the existence of  `"openshift-monitoring"`  ns right? So, IIUC the suggestion is to add this above check inside `event_junits.go` ?

>Question here to ask is whether there's a sufficient replacement for absence of the monitoring stack when checking the alerts in MicroShift. Does it make sense to e.g. run an external monitoring stack (or its simplified version)?

AFAIK there is no plan to add any monitoring stack in MicroShift as of now. So, we might want to skip them as well.

chiragkyal:
In the last 1-2 days, a lot of refactoring was done through https://github.com/openshift/origin/pull/28130 PR, and now I believe the changes suggested above will not hold true. @pacevedom should we consider accommodating these changes in your [PR](https://github.com/openshift/origin/pull/28136)?

PR #27990, opened 2023-06-21
Flake ci-cluster-network-liveness and new backend disruption failures to avoid payload rejections

DennisPeriquet:
[TRT-1095](https://issues.redhat.com//browse/TRT-1095)

We'll flake these backends to avoid failing payloads.

### DIFF
--- a/pkg/synthetictests/disruption.go
+++ b/pkg/synthetictests/disruption.go
@@ -16,6 +16,24 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
+// isExcludedDisruptionBackend returns true if any of the given backends are in the
+// disruption backend name.  Essentially, we want to test these disruption backends
+// but flake them for now to avoid failing payloads.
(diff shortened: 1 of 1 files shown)


pkg/synthetictests/disruption.go
```diff
@@ -16,6 +16,24 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
+// isExcludedDisruptionBackend returns true if any of the given backends are in the
+// disruption backend name.  Essentially, we want to test these disruption backends
+// but flake them for now to avoid failing payloads.
+func isExcludedDisruptionBackend(name string) bool {
+	excludedNames := []string{
+		"ci-cluster-network-liveness",
+		"kube-api-http1-external-lb",
```
deads2k, line 25: why is this excluded?

dgoodwin: It's a new backend, anything mentioning http is, the originals were left alone and are running along side these.

PR #27994, opened 2023-06-22
verifies number of node reboots even for HA

deads2k:
enable https://github.com/openshift/origin/pull/27993 everywhere

What if arm64 (or something else) is consistently restarting and on HA we simply recover fast enough to not notice?

/assign @vrutkovs 

This one will require a payload test first to ensure we don't straight out fail.

### DIFF
--- a/test/extended/machines/cluster.go
+++ b/test/extended/machines/cluster.go
@@ -8,8 +8,6 @@ import (
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
-	configv1 "github.com/openshift/api/config/v1"
-	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/prometheus"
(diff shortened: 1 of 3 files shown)


deads2k:
/payload-job periodic-ci-openshift-multiarch-master-nightly-4.14-ocp-e2e-aws-ovn-arm64

vrutkovs:
We have other test verifying that nodes don't reboot on HA - "should report ready nodes the entire duration of the test run" - but maybe someday we'll boot so quick that kubelet won't even change status to NotReady? Lets have another confirmation of an unexpected reboot anyway.

/lgtm

test/extended/machines/cluster.go
```diff
@@ -100,4 +112,114 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+			"master": 0,
+		}
+		dynamicClient, err := dynamic.NewForConfig(oc.KubeFramework().ClientConfig())
+		o.Expect(err).NotTo(o.HaveOccurred())
+		machineConfigGroupVersionResource := schema.GroupVersionResource{
+			Group: "machineconfiguration.openshift.io", Version: "v1", Resource: "machineconfigs",
+		}
+		mcList, err := dynamicClient.Resource(machineConfigGroupVersionResource).List(ctx, metav1.ListOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		for i := range mcList.Items {
```
romfreiman, line 141: @vrutkovs shouldnt we count the pivot reboot as well?

vrutkovs: Pivot reboot is counted too, as `rendered-` is generated for this (two boots, but just one recorded reboot)

PR #27996, opened 2023-06-23
USHIFT-647: skip sig-cli explain for CRDs in MicroShift

chiragkyal:
This change will check if a cluster is MicroShift and make changes accordingly for `[sig-cli] oc explain should contain proper spec+status for CRDs` test.

### DIFF
--- a/test/extended/cli/explain.go
+++ b/test/extended/cli/explain.go
@@ -193,6 +193,13 @@ var (
 		{Group: "performance.openshift.io", Version: "v2", Resource: "performanceprofiles"},
 	}
 
+	microshiftCRDTypes = []schema.GroupVersionResource{
+		{Group: "route.openshift.io", Version: "v1", Resource: "routes"},
+		{Group: "topolvm.io", Version: "v1", Resource: "logicalvolumes"},
+		//{Group: "security.internal.openshift.io", Version: "v1", Resource: "rangeallocations"},
+		//{Group: "security.openshift.io", Version: "v1", Resource: "securitycontextconstraints"},
+	}
+
 	specialTypes = map[string][]explainExceptions{
 		"apps.openshift.io": {
 			{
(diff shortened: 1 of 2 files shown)


test/extended/cli/explain.go
```diff
@@ -579,6 +602,9 @@ var _ = g.Describe("[sig-cli] oc explain", func() {
 		if *controlPlaneTopology == configv1.ExternalTopologyMode {
 			crdTypesTest = baseCRDTypes
 		}
+		if !doesCRDTypesExist(oc, crdTypesTest) {
```
dhellmann, line 605: Instead of skipping the test entirely, let's look for MicroShift and set crdTypesTest to the list we expect to see so we can test that explain works with them.

test/extended/cli/explain.go
```diff
@@ -193,6 +193,13 @@ var (
 		{Group: "performance.openshift.io", Version: "v2", Resource: "performanceprofiles"},
 	}
 
+	microshiftCRDTypes = []schema.GroupVersionResource{
+		{Group: "route.openshift.io", Version: "v1", Resource: "routes"},
+		{Group: "topolvm.io", Version: "v1", Resource: "logicalvolumes"},
+		//{Group: "security.internal.openshift.io", Version: "v1", Resource: "rangeallocations"},
+		//{Group: "security.openshift.io", Version: "v1", Resource: "securitycontextconstraints"},
+	}
```
chiragkyal, line 201: `oc explain` for these two resources do not have `spec` and `status` fields, and the targeted test `[sig-cli] oc explain should contain proper spec+status for CRDs` is meant to check `spec` and `status`. 

These two resources are already present in `specialTypes` list and are covered separately in another test with specific `field` and `pattern`. So, there should not be any problem if we don't keep them again here.

chiragkyal:
cc @soltysh

test/extended/cli/explain.go
dhellmann: A comment about that in the code would help future maintainers.

You say the 2 CRDs do not have spec or status fields. Shouldn't they? Does the fact that they are missing indicate some sort of problem?

chiragkyal: Added a comment 

The missing fields might be by design, as they are not present in OpenShift also. So, I think this is not a problem.

dhellmann:
@chiragkyal are those test failures known issues for this repo?

@soltysh we discussed the approach Chirag is taking in this PR on slack.

/assign @soltysh 
/lgtm

chiragkyal:
I do see some of them are failing in other PRs, also. 

/retest

soltysh (approved):
/approve

PR #27998, opened 2023-06-26
OCPBUGS-15291: [sig-cli] oc idle: get a dc name through labels instead of parsing oc create output

ingvagabund:
Avoid parsing output of `oc create` which can return additional lines of text besides the dc name


soltysh (approved):
/lgtm
/approve

PR #28000, opened 2023-06-26
Report external binary usage

soltysh:
Report external binary usage like other synthetic tests. 

/assign @bparees @ingvagabund

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -163,8 +163,10 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
 		return fmt.Errorf("failed reading origin test suites: %w", err)
 	}
 
+	var fallbackSyntheticTestResult []*junitapi.JUnitTestCase
 	if len(os.Getenv("OPENSHIFT_SKIP_EXTERNAL_TESTS")) == 0 {
-		fmt.Fprintf(opt.Out, "Attempting to pull tests from external binary...\n")
+		buf := &bytes.Buffer{}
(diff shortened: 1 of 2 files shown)


pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -166,24 +166,24 @@ func (opt *Options) Run(suite *TestSuite, junitSuiteName string) error {
...
+			// from the final lists, which contains:
+			// 1. origin tests, only
+			// 2. k8s tests, coming from external binary
+			if !strings.Contains(test.name, "[Suite:k8s]") {
+				filteredTests = append(filteredTests, test)
 			}
-			tests = append(filteredTests, externalTests...)
-			fmt.Fprintf(opt.Out, "Got %d tests from external binary\n", len(externalTests))
-		} else {
-			fmt.Fprintf(opt.Out, "Falling back to built-in suite, failed reading external test suites: %v\n", err)
```
bparees, line 185: don't we still (at least for now) need a way to fallback to the built-in suite if the person running the binary wants to do so?  that's what i meant about making it an opt-in choice.  (thinking about microshift for example, i dunno how their CI is setup, but i assume they don't have the external binary configured yet....maybe they do)

soltysh: I talked to @deads2k about it, and I'll take a different route. I'll add a flaky test which will tell us where we are actually falling back to built-in suite, and based on that we'll be able to fix the places. Only then we'll drop this fallback entirely. This way we'll minimize the risks.

bparees: works for me, thanks!

PR #28005, opened 2023-06-27
USHIFT-704: skip [sig-cli] whoami result with console for MicroShift

chiragkyal:
This change will skip 
`"[sig-cli] oc basics can show correct whoami result with console [Skipped:NoOptionalCapabilities] [Suite:openshift/conformance/parallel]"` 
test by adding NoOptionalCapabilities for MicroShift

### DIFF
--- a/cmd/openshift-tests/provider.go
+++ b/cmd/openshift-tests/provider.go
@@ -78,7 +78,23 @@ func initializeTestFramework(context *e2e.TestContextType, config *exutilcluster
 func decodeProvider(provider string, dryRun, discover bool, clusterState *exutilcluster.ClusterState) (*exutilcluster.ClusterConfiguration, error) {
 	switch provider {
 	case "none":
-		return &exutilcluster.ClusterConfiguration{ProviderName: "skeleton"}, nil
+		config := &exutilcluster.ClusterConfiguration{
+			ProviderName: "skeleton",
+		}
+		// Add NoOptionalCapabilities for MicroShift
+		coreClient, err := e2e.LoadClientset(true)
+		if err != nil {
+			return nil, err
+		}
+		isMicroShift, err := exutil.IsMicroShiftCluster(coreClient)
+		if err != nil {
+			return nil, err
+		}
+		if isMicroShift {
+			config.HasNoOptionalCapabilities = true
+		}
+
+		return config, nil
 
 	case "":
 		if _, ok := os.LookupEnv("KUBE_SSH_USER"); ok {
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -2094,9 +2094,11 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, groupVersionName
 	return false, nil
 }
 
-func IsMicroShiftCluster(oc *CLI) (bool, error) {
+// IsMicroShiftCluster returns "true" if a cluster is MicroShift,
+// "false" otherwise. It needs kube-admin client as input.
+func IsMicroShiftCluster(kubeClient k8sclient.Interface) (bool, error) {
 	// MicroShift cluster contains "microshift-version" configmap in "kube-public" namespace
-	cm, err := oc.AdminKubeClient().CoreV1().ConfigMaps("kube-public").Get(context.Background(), "microshift-version", metav1.GetOptions{})
+	cm, err := kubeClient.CoreV1().ConfigMaps("kube-public").Get(context.Background(), "microshift-version", metav1.GetOptions{})
 	if err != nil {
 		if kapierrs.IsNotFound(err) {
 			e2e.Logf("microshift-version configmap not found")
--- a/test/extended/cli/explain.go
+++ b/test/extended/cli/explain.go
@@ -475,7 +475,7 @@ var (
 )
 
 func getCrdTypes(oc *exutil.CLI) []schema.GroupVersionResource {
-	isMicroShift, err := exutil.IsMicroShiftCluster(oc)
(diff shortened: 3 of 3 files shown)


test/extended/cli/basics.go
```diff
@@ -250,6 +250,11 @@ var _ = g.Describe("[sig-cli] oc basics", func() {
 	})
 
 	g.It("can show correct whoami result with console", func() {
+		nsExist, err := exutil.IsNamespaceExist(oc, "openshift-config-managed")
```
dhellmann, line 253: How is that namespace related to this test? Is that where the console runs? Or does the whoami command use information from that namespace?

dhellmann: The console is an optional component now. What does `whoami` do if the namespace exists but the console URL is empty? Maybe there's a bug in `whoami` for this more extreme case where the console URL cannot be determined?

bparees: it only is an issue if the "--show-console" arg is specified.  If you specify that arg when you don't have the console enabled, you'll get an error but that should not be surprising.

https://github.com/openshift/oc/blob/c078876cc5aa434a02f14c28b8c2a56b9011aa13/pkg/cli/whoami/whoami.go#L149-L150

https://github.com/openshift/oc/blob/c078876cc5aa434a02f14c28b8c2a56b9011aa13/pkg/cli/whoami/whoami.go#L121-L136

i feel like we had to deal w/ this for the "nocaps" e2e job, but since this test doesn't seem to contain any logic to check for which caps are enabled, i'm not sure how we ended up doing that.

bparees: > Do we need another entry in the list in TestDecodeProvider for MicroShift?

No, that's just a test case for ensuring we detect/decode clusters correctly (so that we can pick the right set of tests).  I think you need to update the cluster discovery logic to recognize a "microshift" cluster and set the EnabledCapabilities accordingly:

https://github.com/openshift/origin/blob/5c4887ad5ff607fc70a51bcd8ccfb5f21502e0b0/cmd/openshift-tests/provider.go#L78
https://github.com/openshift/origin/blob/5c4887ad5ff607fc70a51bcd8ccfb5f21502e0b0/test/extended/util/cluster/cluster.go#L149-L153

chiragkyal: I've updated the [decodeProvider](https://github.com/openshift/origin/blob/5c4887ad5ff607fc70a51bcd8ccfb5f21502e0b0/cmd/openshift-tests/provider.go#L78) logic to recognize a "microshift" cluster and set `config.HasNoOptionalCapabilities = true`. However, the `[Skipped:NoOptionalCapabilities]` annotation doesn't seem to be getting enabled for this test. Any suggestions that I may be missing here?

cmd/openshift-tests/provider.go
```diff
@@ -139,6 +139,19 @@ func decodeProvider(provider string, dryRun, discover bool, clusterState *exutil
...
+		if err != nil {
+			return nil, err
+		}
+		isMicroShift, err := exutil.IsMicroShiftCluster(coreClient)
+		if err != nil {
+			return nil, err
+		}
+		if isMicroShift {
+			config.HasNoOptionalCapabilities = true
+		}
```
bparees, line 153: are you sure you're ending on up in the "Default" provider case when running against a microshift cluster?

should be easy enough to run locally w/ some debug statements and see what's actually happening when the detection logic runs

cmd/openshift-tests/provider.go
```diff
@@ -139,6 +139,19 @@ func decodeProvider(provider string, dryRun, discover bool, clusterState *exutil
 		if err := json.Unmarshal([]byte(provider), config); err != nil {
 			return nil, fmt.Errorf("provider must decode into the ClusterConfig object: %v", err)
 		}
+
+		coreClient, err := e2e.LoadClientset(true)
+		if err != nil {
+			return nil, err
+		}
+		isMicroShift, err := exutil.IsMicroShiftCluster(coreClient)
```
bparees, line 147: I would do this in "DiscoverClusterState" and not "decodeProvider".

The provider is irrelevant

cmd/openshift-tests/provider.go
chiragkyal: No, it should be inside "none" case. There was something wrong with my MicroShift cluster which leads all debug statements to go inside "Default" case. 

I've added it now in "none" case.

bparees:
lgtm, is there a microshift job we can run this PR against?

chiragkyal:
> lgtm, is there a microshift job we can run this PR against?

Not yet. Currently, one by one we are fixing the tests which are failing in MicroShift. Ultimately, our target is to have a working job, passing all the tests.

test/extended/util/framework.go
```diff
@@ -2094,9 +2094,9 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, groupVersionName
 	return false, nil
 }
 
-func IsMicroShiftCluster(oc *CLI) (bool, error) {
+func IsMicroShiftCluster(kubeClient k8sclient.Interface) (bool, error) {
```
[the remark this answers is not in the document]
bparees: ah, you don't have an oc client at the spot where you're calling it now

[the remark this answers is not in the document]
bparees: maybe at least godoc this function to indicate it needs an admin client

cmd/openshift-tests/provider.go
pacevedom: Agree with Ben, even though the name might be misleading because its more discovery than decoding.

PR #28006, opened 2023-06-27
OCPVE-379: fix: avoid checking resources for BestEffort pods

eggfoobar:
The reason the flakes have been occurring is due to some tests around the network-operator creating DaemonSets and Deployments that do not contain resources. Adding a skip list for resources that are okay being BestEffort and added comment to code to be explicit on the meaning of best effort in a workload partitioned cluster. This does not effect the CPU affinity for workload partitioning.

Test-Grid Flaks: https://testgrid.k8s.io/redhat-openshift-ocp-release-4.14-informing#periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-ovn-cpu-partitioning&include-filter-by-regex=CPU%20Partitioning

Ref Resources on the network operator:
[DaemonSet](https://github.com/openshift/cluster-network-operator/blob/25b457440a3cf305a75586d879542a98dde4fca2/bindata/allowlist/daemonset/daemonset.yaml)
[Deployment](https://github.com/openshift/cluster-network-operator/blob/master/bindata/egress-router/001-deployment.yaml)

/assign @deads2k

### DIFF
--- a/test/extended/cpu_partitioning/pods.go
+++ b/test/extended/cpu_partitioning/pods.go
@@ -13,7 +13,39 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/kubernetes/test/e2e/framework"
 )
+
+// When using the workload partitioning feature ( which allows cluster admins to limit the number of CPUs the platform pods have access to ),
+// with BestEffort Pods, the probability of CPU time contention goes up and resource.requests become more important for platform pods.
+// It is strongly recommended that for everything but your least important pods you should supply adequate resource.requests,
+// otherwise you might be in scenarios where your pod might be starved for CPU time since BestEffort pods get a default of 2 CPU Shares.
+// In other words under a highly utilized CPU in a CPU pinned cluster would mean your pod would only get 2/1024 * 100 = ~0.19% CPU time.
+//
+// Please be sure to verify that the deployment can function normally if these pods don't get a chance to complete their work.
+//
+// The below exclude lists are used to skip resource checks on these objects.
+// the map is string(<resource-name>):[]string(<resource-namespace-substring>).
+// Any pod resource that matches the name in the map and the substring of the namespace in the array is skipped.
+var (
+	excludedBestEffortDeployments = map[string][]string{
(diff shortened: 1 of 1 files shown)


deads2k:
The point of cpu affinity is to keep the platform from taking cpu from the workloads. Our platform pods, even best effort ones, need to be controlled for that purpose

eggfoobar:
Oh sorry @deads2k, I wasn't clear what's happening, the desired CPU affinity is not being violated here. There are two things that happen for workload partitioning, a CPU affinity and a CPU share change. 

The flake was around the CPU Share itself. Since we checked that pods containing the workload partitioning annotations also had their resource requests altered to the custom resource field, this would flake on the tests that spun up BestEffort deployments and daemonsets because the default behavior is to apply the workload annotation with the minimum 2 CPU shares and not add/change resource requests since none are present at admission time.

Unless we should require that all platform pods must contain resource requests, this would be safe to skip since the CPU affinity is still respected, and the workload get's a default CPUShare value of 2. What do you think?

eggfoobar:
Adding some more context around CPU shares. A cpu share is a relative amount of time a process is granted on a CPU, in Kubernetes 1CPU = 1024CPU shares. 

When there is no CPU time contention, nothing happens and the process is allowed to use more, however if there is CPU time contention, then the cpu share will be enforced and processes with a higher CPU share will get that proportional percentage of time on the CPU. i.e two process with a 512 cpu share will each get 50% CPU time. 

In Kubernetes when a pod has a QoS of BestEffort, the kubelet will automatically give it a default CPU share of 2, meaning it will be scheduled pretty much anywhere since it's so small and if there's contention it will get the smallest amount of time on the CPU. If you need a minimum amount, consider supplying a `resource.requests` for your pod. 

Normally this is not too much of a problem. However, when using the workload pinning feature ( which allows cluster admins to limit the number of CPUs the platform pods have access to ), the probability of CPU time contention goes up and `resource.requests` become more important for platform pods. It is strongly recommended that for everything but your least important pods you should supply adequate `resource.requests`, otherwise you might be in scenarios where your pod might be starved for CPU time with the default 2 CPU Shares.

deads2k:
the networking team ack'd these.   The CPU slice is extremely small and bugs were opened so there's long term tracking as well.

/lgtm

PR #28009, opened 2023-06-29
allow cluster-config-operator to manage featuregate upgrade block

deads2k:
required for https://github.com/openshift/cluster-config-operator/pull/325


JoelSpeed (commented):
/lgtm

PR #28015, opened 2023-06-30
USHIFT-1379: skip must-gather tests for MicroShift

chiragkyal:
As `oc adm must-gather` is not supported in MicroShift, these tests can be skipped.

### DIFF
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -30,21 +30,13 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 
 	oc := exutil.NewCLI("oc-adm-must-gather").AsAdmin()
 
-	g.BeforeEach(func() {
-		isMicroShift, err := exutil.IsMicroShiftCluster(oc)
-		o.Expect(err).NotTo(o.HaveOccurred())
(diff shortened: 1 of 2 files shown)


test/extended/cli/mustgather.go
```diff
@@ -30,6 +30,14 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 
 	oc := exutil.NewCLI("oc-adm-must-gather").AsAdmin()
 
+	g.BeforeEach(func() {
+		isMicroShift, err := exutil.IsMicroShiftCluster(oc)
```
pacevedom, line 34: must-gather uses many OCP apigroups not present in microshift, is it better to annotate the tests instead?

chiragkyal: Yes, that can be a better approach. I've added `[apigroup:config.openshift.io]` in these tests, please let me know if this looks good. Thanks!

dhellmann:
/retest-required
/lgtm

Using the config API group makes sense as an indicator that this test is tied to OpenShift.

pacevedom (commented):
/lgtm

soltysh (approved):
/approve

PR #28016, opened 2023-06-30
Update "verify that nodes have no unexpected reboots" test

vrutkovs:
Instead of assuming that every MC is causing a reboot, we should look for "rebooting" message from systemd-login

### DIFF
--- a/test/extended/machines/cluster.go
+++ b/test/extended/machines/cluster.go
@@ -12,8 +12,6 @@ import (
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/prometheus"
 	"github.com/stretchr/objx"
-	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


test/extended/machines/cluster.go
```diff
@@ -126,46 +124,18 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
-			o.Expect(expectedReboots).To(o.HaveKey(expectedRebootsForNodeRole))
-			expected[node.Name] = expectedRebootsForNodeRole
-			nodeReboots, err := getNumberOfBootsForNode(oc.KubeClient(), oc.Namespace(), node.Name)
+			nodeBoots, nodeReboots, err := getNumberOfBootsForNode(oc.KubeClient(), oc.Namespace(), node.Name)
 			if err != nil {
 				errs = append(errs, err)
 			}
-			actual[node.Name] = nodeReboots
+			expected[node.Name] = nodeBoots
+			actual[node.Name] = nodeReboots + 1
```
soltysh, line 138: Why this is +1?

vrutkovs: This compares:
* number of machine boots
* number of legitimate reboots

The former is always larger than the former (the machine boots, gets rebooted, boots again), this is why we add 1 here, so that `expected` would match `actual` later

soltysh (approved):
/lgtm

PR #28017, opened 2023-06-30
Update image stream test to create a manifest list image by default

tvardema:
This PR will fix the test failure when using a manifest-listed payload for a single-architecture deployment.

### DIFF
--- a/test/extended/images/append.go
+++ b/test/extended/images/append.go
@@ -93,6 +93,9 @@ var _ = g.Describe("[sig-imageregistry][Feature:ImageAppend] Image append", func
 		pod := cli.Create(context.TODO(), cliPodWithPullSecret(oc, heredoc.Docf(`
 			set -x
 
+			# create a copy of the single manifest image from the original image for layer count
+			oc image append --insecure --from %[3]s --to %[2]s/%[1]s/tools:singlemanifest
+
 			# create a scratch image with fixed date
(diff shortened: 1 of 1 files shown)


test/extended/images/append.go
```diff
@@ -93,6 +93,9 @@ var _ = g.Describe("[sig-imageregistry][Feature:ImageAppend] Image append", func
 		pod := cli.Create(context.TODO(), cliPodWithPullSecret(oc, heredoc.Docf(`
 			set -x
 
+			# create a copy of the single manifest image from the original image for layer count
+			oc image append --insecure --from %[3]s --to %[2]s/%[1]s/tools:singlemanifest
+
```
Prashanth684, line 98: the ideal way would have been to tag the image - but the builder role doesn't have perms for creating istags.

jaypoulz:
This looks good to me.

/lgtm

jaypoulz:
For context, what happens when you are using a manifest-listed image is that the test inspects the manifest list, and finds no references to Docker layers. To ensure that the test can still count Docker layers, what we do is copy the original shell image to a new tag without specifying that we want to preserve the original manifest. This results in the new tag pointing to a single-arch-manifest. We then leverage this new tag in the layer count check.

flavianmissi:
Hey folks, 
It looks like this test is now perma failing on 4.13 too.
Let's backport this to 4.13.
/cherry-pick release-4.13

PR #28032, opened 2023-07-07
Add logs in clusterquota test for more verbosity

ardaguclu:
This PR adds more logs in case when quota namespace labeling fails. This is basically added to investigate https://issues.redhat.com/browse/OCPBUGS-15568 bug. However, these logs should also make
easier possible regression in the future and do not generate noise(especially when we decrease timeout value
from 30 to 10 seconds back again, after investigation is finished.)


soltysh (approved):
/lgtm
/approve

PR #28037, opened 2023-07-10
Revert "disruption refinement for availability, not latency."

dgoodwin:
Reverts openshift/origin#28003

This is a test, our disruption seemed to drop to near 0 and we can't see why. This PR doesn't line up perfectly, but we want to check and make sure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,4 +29,3 @@ vendor/github.com/prometheus/procfs/fixtures/26231/exe
 
 # Ignore generated binary
 /openshift-tests
-/disruption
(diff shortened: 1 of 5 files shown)


dgoodwin:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade 5

This one for some reason showed disruption last time, I hope this isn't the cause and we're logging bad data...

dgoodwin:
I think this is sadly our culprit and the disruption improvement is not real. We probably started losing all single second disruptions...

In the payload jobs above I see consistent disruption coming back on azure whereas it's gone everywhere else.

dgoodwin:
/close

This was our culprit, we're keeping the change and restarting from new disruption data from this point onward.

PR #28044, opened 2023-07-11
OCPBUGS-15893: Update permission to incl. watch for helmchartrepositories for console users

jerolimov:
See also https://github.com/openshift/console-operator/pull/775

Change the permission check for helmchartrepositories from `get, list` to `get, list, watch`.


jerolimov:
/assign @bparees @deads2k

Hey, since @stlaz is on PTO, can you check this PR? Thanks

jhadvig (approved):
LGTM

PR #28045, opened 2023-07-11
ODC-7333: Add consolesamples to the list of console resources that all users should have read access

jerolimov:
See also https://github.com/openshift/console-operator/pull/776

Add new `consolesamples` resource to the list of resources that every (console) user could read (get, list, watch).


jhadvig (approved):
LGTM

PR #28053, opened 2023-07-14
Improvements and fixes for Loki intervals uploader

dgoodwin:
- Include durationSec when uploading intervals to loki, this allows us to deduce the "to" timestamp on the other side, and should allow logql queries such where durationSec > 10 etc.
- Refactor to fix a significant bug in what namespace intervals were getting labelled with. We were mislabeling the namespace on many intervals due to a bug in how I was batching. We now maintain streams for each namespace, push the batch when we hit the 500 limit, and then push everything remaining for each namespace at the end.
- Add filename, level, and unit tests when uploading intervals to loki.
- Add a --dry-run flag for interval upload. Just for dev testing.
- Log and warn if we don't push all the intervals we were given. (in case of any logic bugs)


pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
```diff
@@ -163,50 +191,134 @@ func UploadIntervalsToLoki(intervals monitorapi.Intervals) error {
...
+
+	// Push the final batches for each namespace:
+	logrus.Info("pushing final batches for each remaining namespace")
+	for namespace, v := range nsIntervals {
+		if len(v) > 0 {
+			err := pushBatch(client, bearerToken, invoker, namespace, nsIntervals[namespace], dryRun)
+			if err != nil {
+				return err
+			}
+			nsIntervals[namespace] = make([][]interface{}, 0)
```
DennisPeriquet, line 254: since you've pushed the rest, do we need this line?

DennisPeriquet: but it is harmless

DennisPeriquet:
> Did some more work yesterday to fix some casing issues, new sample: https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fpull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade%2F1702381233331245056%5C%22,type%3D%5C%22origin-interval%5C%22%7D%20%7C%20unpack%20%7C%20json%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221694708509858%22,%22to%22:%221694744509862%22%7D%7D

On the most recent query, I see:

```
  tempStructuredLocator_keys_e2e_test | [sig-arch][Feature:ClusterUpgrade] Cluster should be upgradeable after finishing upgrade [Late][Suite:upgrade]
-- | --
  | tempStructuredLocator_type | E2ETest
  | tempStructuredMessage_annotations_reason | E2ETestFinished
  | tempStructuredMessage_annotations_status | Passed
  | tempStructuredMessage_cause |  
  | tempStructuredMessage_humanMessage | e2e test finished finished
  | tempStructuredMessage_reason | E2ETestFinished
```

But previous one, I see (i.e., the `temp` prefix is gone):

```
  StructuredLocator_keys_e2e_test | [sig-arch][Feature:ClusterUpgrade] Cluster should be upgradeable after finishing upgrade [Late][Suite:upgrade]
-- | --
  | StructuredLocator_type | E2ETest
  | StructuredMessage_annotations_reason | E2ETestStarted
  | StructuredMessage_cause |  
  | StructuredMessage_humanMessage | started
  | StructuredMessage_reason | E2ETestStarted
```

is this ok?

pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
dgoodwin: Nope looks like you're right, probably didn't need that line but indeed harmless.

PR #28056, opened 2023-07-14
OCPVE-318: feat: add node validation for workload partitioning

eggfoobar:
This PR adds a test for validating CPU shares and CPUSets are being respected in a cluster.

The given test runs a collection script via the machine-config-daemon (this is chosen for ease of querying and it running on every node, adequate permissions and node config responsibilities), The script will generate a json payload that is a trimmed down data set of crio data and CPUSet information per pid from `taskset`. This information is then used to validate that containers in a node are correct configured.

|| Partitioned Cluster |  Regular Cluster  |
|--|-------------------|-----------------|
|CPU Set | Validate CPUSet Are Restricted or Open Depending on PAO | Validates CPUSet are wide open |
|CPU Share| Validate CPUShare match annotation at CRIO Level | Should not be needed, pre-existing tests check if annotations are present |

### DIFF
--- a/test/extended/cpu_partitioning/OWNERS
+++ b/test/extended/cpu_partitioning/OWNERS
@@ -1,0 +1,7 @@
+approvers:
+  - eggfoobar
+reviewers:
+  - yanirq
+  - ffromani
+  - Tal-or
+  - MarSik
--- a/test/extended/cpu_partitioning/crio.go
+++ b/test/extended/cpu_partitioning/crio.go
@@ -1,0 +1,338 @@
+package cpu_partitioning
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	ocpv1 "github.com/openshift/api/config/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+	"k8s.io/kubernetes/pkg/kubelet/cm/cpuset"
+)
+
+// This collection script queries CRIO for all it's containers, that data is then filtered down to only a subset
+// of information needed for validating workload partitioning and the PIDs are mapped to their corresponding host
+// CPUSet via the taskset command.
+const collectionScript = `
+#!/bin/bash
+
+container_ids=$(sudo crictl ps -q)
+container_data=$(sudo crictl inspect $container_ids)
+
+# Filter down CRIO data to relevant information only
+workload_containers=$(echo $container_data | jq -rs '[ 
+	.[] | select(.info.runtimeSpec.annotations["target.workload.openshift.io/management"]) | 
+	{
+		cpuSet: (.info.runtimeSpec.linux.resources.cpu.cpus // ""),
+        cpuShare: .info.runtimeSpec.linux.resources.cpu.shares,
+		annotations: .info.runtimeSpec.annotations,
+        podNamespace: .info.runtimeSpec.annotations["io.kubernetes.pod.namespace"],
+        podName: .info.runtimeSpec.annotations["io.kubernetes.pod.name"],
+		name: .status.metadata.name,
+		pid: .info.pid, 
+		hostname: .info.runtimeSpec.hostname 
+	}
+	]')
+
+# Get the host CPUSet for each PID, if during regular e2e tests a POD comes up and is done
+# in between us recording a CRIO process and checking taskset, we record -1 for CPUSet that PID and filter it out later.
+workload_host_cpu_set=$(echo $workload_containers | jq -r '.[]|.pid' | \
+while read i; do (taskset -cp $i 2>/dev/null || echo "pid $i's no longer runs: -1"); done | awk '{printf "{\"hostCPUSet\": \"%s\"}", $6}' | jq -rs)
+
+# Map PIDs and the CRIO data into the final results.
+final_results=$(echo "$workload_containers $workload_host_cpu_set" | jq -rs '.[0] as $file1 | .[1] as $file2 | map($file1[] + $file2[]) | unique')
+
+# Filter out results that we failed to get a CPUSet for due to failure with taskset.
+echo "$final_results" | jq -rc '[.[] | select(.hostCPUSet != "-1")]'
+`
+
+const jsonKeyCPUShare = "cpushares"
+
+// Collection data for each container in a node
+type crioContainerData struct {
+	PodNamespace string            `json:"podNamespace"`
+	PodName      string            `json:"podName"`
+	Name         string            `json:"name"`
+	CPUSet       string            `json:"cpuSet"`
+	CPUShare     int               `json:"cpuShare"`
+	Annotations  map[string]string `json:"annotations"`
+	HostCPUSet   string            `json:"hostCPUSet"`
+	Hostname     string            `json:"hostname"`
+	Pid          int               `json:"pid"`
+}
+
+// Container struct for master and worker node CPUSets
+type nodeCPUSets struct {
+	Master cpuset.CPUSet
+	Worker cpuset.CPUSet
+}
+
+// getAnnotationCPUShare Looks through the annotations to find the workload partitioning annotation and
+// returns the cpushares or error if no such annotation exists.
+func (c *crioContainerData) getAnnotationCPUShare() (int, error) {
+	anno := fmt.Sprintf("%s/%s", WorkloadAnnotationPrefix, c.Name)
+	cpuShare, ok := c.Annotations[anno]
+	if !ok {
+		return 0, fmt.Errorf("workload annotation of [%s] was expected but not found", anno)
+	}
+
+	// workload cpushare annotations have a json string that will contain `{ "cpushares": <share int> }`
+	// we parse and return that value here.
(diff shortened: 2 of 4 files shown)


deads2k:
very glad to see a test for this. Please have a teammate review, add yourself as an approver on the individual test package, and them as a reviewer.  Ping me on slack when ready.

jlojosnegros:
lgtm
but let others take a look

test/extended/cpu_partitioning/OWNERS
```diff
@@ -0,0 +1,4 @@
+approvers:
+  - eggfoobar
+reviewers:
+  - yanirq
```
ffromani, line 4: do we want more reviewers?

yanirq: please add @MarSik as well to the reviews list and others that can fit

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+while read i; do taskset -cp $i; done | awk '{printf "{\"hostCPUSet\": \"%s\"}", $6}' | jq -rs)
+
+final_results=$(echo "$workload_containers $workload_host_cpu_set" | jq -rs '.[0] as $file1 | .[1] as $file2 | map($file1[] + $file2[]) | unique')
+
+echo "$final_results"
+`
+
+// Collection data for each container in a node
+type crioContainerData struct {
+	ContainerPodNamespace string            `json:"containerPodNamespace"`
```
Tal-or, line 63: nit: I would remove the `Container` prefix from the variables names, since the struct name is `crioContainerData` it's clearly you're referring to container's data.
This would made the code cleaner and shorter.

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// workload cpushare annotations have a json string that will contain `{ "cpushares": <share int> }`
+	// we parse and return that value here.
+	temp := map[string]int{}
+	err := json.Unmarshal([]byte(cpuShare), &temp)
+	if err != nil {
+		return 0, fmt.Errorf("err parsing cpushare json annotation: %w", err)
+	}
+
+	value, ok := temp["cpushares"]
```
Tal-or, line 91: Maybe better use constant here.

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+	if !ok {
+		return 0, fmt.Errorf("err cpushare annotation was incorrect expected format {\"cpushares\": <int> }")
+	}
+
+	return value, nil
+}
+
+var _ = g.Describe("[sig-node][apigroup:config.openshift.io] CPU Partitioning node validation", func() {
+	defer g.GinkgoRecover()
```
Tal-or, line 100: `GinkgoRecover` usually called within a go routine, is this code is going to be running on a separate thread?

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+		oc                      = exutil.NewCLIWithoutNamespace("cpu-partitioning").AsAdmin()
+		ctx                     = context.Background()
+		isClusterCPUPartitioned = false
+	)
+
+	g.BeforeEach(func() {
+		isClusterCPUPartitioned = getCpuPartitionedStatus(oc) == ocpv1.CPUPartitioningAllNodes
+	})
+
+	g.It("crio containers should have correct cpuset and cpushare set", func() {
```
Tal-or, line 112: nit: more readable `It ("should have correct cpuset and cpushare set in crio containers")`

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+	pod := &pods.Items[0]
+
+	var out []byte
+	getInfo := []string{
+		"chroot",
+		"/rootfs",
+		"/bin/bash",
+		"-c", collectionScript}
+
+	ctx := context.TODO()
```
Tal-or, line 263: It's usually better to pass the context from the caller, although here it's just a test so no biggie.

Tal-or (commented):
couple of small comments from my side, but LGTM in general

test/extended/cpu_partitioning/crio.go
eggfoobar: good suggestion, changed

test/extended/cpu_partitioning/crio.go
eggfoobar: Sure thing, updated.

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+	// Default behavior is that CPUSets are empty, which means wide open to the host.
+	// There should be no CPUSet constraint if no PerformanceProfile is present.
+	nodeRoleCPUSet := map[string]string{nodeMasterLabel: "", nodeWorkerLabel: ""}
+
+	if !isClusterPartitioned {
+		return nodeRoleCPUSet, nil
+	}
+
+	// Performance profiles dictate constrained CPUSets, we gather them here to compare.
+	performanceProfiles, err := oc.AdminDynamicClient().
```
ffromani, line 197: we do we use a dynamic client here?

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+	err = exec.StreamWithContext(context.Background(), remotecommand.StreamOptions{
+		Stdout: &outputBuf,
+		Stderr: &errorBuf,
+		Tty:    false,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to run command %v: output %q; error %q; %w", command, outputBuf.String(), errorBuf.String(), err)
+	}
+
+	if errorBuf.Len() != 0 {
```
ffromani, line 314: for a general utility having messages on stderr is not necessarily a failure conditions - logs for example are usually (and better) sent on stderr

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+			// that contain workload annotations. In the event that they do, we allow the logic to continue to validate
+			// CPUSets.
+			if isClusterCPUPartitioned {
+				o.Expect(crioData).ToNot(o.BeEmpty(), "error getting crio container data, container data should not be empty")
+			}
+
+			// CPU Core counts are numbers 0-n, we minus one since Status.Capacity.Cpu count is numbered conventionally.
+			// We generate a wide open CPUSet to use for comparison.
+			totalCoreCount := node.Status.Capacity.Cpu().Value() - 1
+			parsedFullNodeCPUSet, err := cpuset.Parse(fmt.Sprintf("0-%d", totalCoreCount))
```
ffromani, line 146: I was wondering if a plain `for` loop and then using `cpuset.New` is better, but in this context it probably your approach is fine

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+					share, err := containerInfo.getAnnotationCPUShare()
+					o.Expect(err).ToNot(o.HaveOccurred(), "failed to get cpushares annotation json", err)
+					o.Expect(share).To(o.Equal(containerInfo.ContainerCPUShare), "cpushares do not match between crio config and desired")
+				}
+			}
+		}
+	})
+})
+
+func getExpectedCPUSetConstraints(ctx context.Context, oc *exutil.CLI, isClusterPartitioned bool) (map[string]string, error) {
```
ffromani, line 187: minor: nonblocking: I'd return a plain object instead of packing return values in a map.

test/extended/cpu_partitioning/crio.go
```diff
@@ -0,0 +1,314 @@
...
+				expectedCPUSet = nodeRoleCPUSet[nodeMasterLabel]
+			} else if _, isWorker := node.Labels[nodeWorkerLabel]; isWorker {
+				expectedCPUSet = nodeRoleCPUSet[nodeWorkerLabel]
+			}
+
+			g.By(fmt.Sprintf("checking for expected CPUSet %s", expectedCPUSet))
+
+			for _, containerInfo := range crioData {
+				o.Expect(containerInfo.ContainerCPUSet).To(
+					o.Equal(expectedCPUSet),
```
ffromani, line 159: do we want or need we use the canonical cpuset string representation? should we use `cpuset.Equals` to be safe and avoid flakes? Also because you seem to use cpuset objects (vs their string representation) right below

ffromani (commented):
mostly LGTM, few minorish comments inside

test/extended/cpu_partitioning/crio.go
ffromani: makes sense. Not a fan of this but not a blocker either. We can keep it like this.

test/extended/cpu_partitioning/crio.go
eggfoobar: That's fair, for this purpose though, in this exec we're just executing a script so I wouldn't expect anything to come through on stderr. If something did, like a log, then I think we would consider that a problem since it's unexpected for this call

test/extended/cpu_partitioning/crio.go
eggfoobar: That's a good call out, the idea here was since this version of the CPUSet doesn't get parsed down to crio and is a 1/1 match on what the user placed in `reserved` that we would fail if it's otherwise, but that's not robust if things change. Updating

test/extended/cpu_partitioning/crio.go
eggfoobar: That's a good point, I wasn't entirely 100% sure on how many different ways the Capacity value was going to come through, so I didn't want to cast it to an `int`. But I don't think there's harm in that, if anything the test will break so that's desired. I'll go ahead and make the update

test/extended/cpu_partitioning/crio.go
eggfoobar: That's fair, let me add a comment there that if we end up needing to do more with Performance Profile it should be imported.

test/extended/cpu_partitioning/crio.go
eggfoobar: Good idea, updated to use a struct that captures the CPUSets for masters and workers.

eggfoobar:
Updated the code with the latest suggestions and the script for collecting the CRIO data. Verified the changes manually on a partitioned cluster with and with out Performance Profiles as well as a regular cluster with no partitioning. 

@ffromani would you mind giving it another look?

ffromani (commented):
/lgtm

eggfoobar:
/unhold

Thanks @deads2k, squashed commits.

PR #28059, opened 2023-07-17
only close intervals if they are open

deads2k:
check the upgrade timeline to see if the node intervals look better.

Also, when I fixed the MCO namespace in events, that broke the "automatic" locator generated for real events.

We need to decide how to produce real locators for events to avoid lots of randomness.  Perhaps a whitelist?


pkg/monitor/monitorapi/identification.go
```diff
@@ -53,23 +45,32 @@ func IsNode(locator string) bool {
 }
 
 func NodeFromLocator(locator string) (string, bool) {
-	if !strings.HasPrefix(locator, "node/") {
```
deads2k, line 56: this made us order sensitive and this https://github.com/openshift/machine-config-operator/pull/3767 made the namespace get specified.

deads2k:
1. units fixed
2. any-order honored by rendering
3. add detail for who constructed a particular interval

PR #28062, opened 2023-07-18
Trt 1150 resource watch commit failures

neisw:
- First commit moves the locking up so we don't write a file in the middle of another commit.

- Second commit adds retry using git add . as a fallback in case we get in a bad state with multiple pending updates and no other way to recover.


neisw:
/hold 

additional review of the root issue.

PR #28070, opened 2023-07-19
USHIFT-1484: skip prometheus alerts for MicroShift

chiragkyal:
This change will first look for `"openshift-monitoring"` namespace before creating prometheus client.

If the namespace doesn't exist, it will skip the following tests
```
"[sig-instrumentation][Late] Alerts shouldn't exceed the series limit of total series sent via telemetry from each cluster": " [Suite:openshift/conformance/parallel]",
"[sig-instrumentation][Late] Alerts shouldn't report any unexpected alerts in firing or pending state": " [Suite:openshift/conformance/parallel]", 
```

[USHIFT-1484](https://issues.redhat.com/browse/USHIFT-1484)

### DIFF
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -2102,6 +2102,19 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, group string) (b
 	return false, nil
 }
 
+func IsNamespaceExist(kubeClient *kubernetes.Clientset, namespace string) (bool, error) {
+	_, err := kubeClient.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	if err != nil {
+		if kapierrs.IsNotFound(err) {
+			e2e.Logf("%s namespace not found", namespace)
+			return false, nil
+		}
+		return false, err
+	}
+
+	return true, nil
+}
+
 // IsMicroShiftCluster returns "true" if a cluster is MicroShift,
 // "false" otherwise. It needs kube-admin client as input.
 func IsMicroShiftCluster(kubeClient k8sclient.Interface) (bool, error) {
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -23,6 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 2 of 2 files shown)


copejon:
Hey @chiragkyal, these changes look reasonable to me.

test/extended/util/framework.go
```diff
@@ -2102,6 +2102,24 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, group string) (b
...
+	ns, err := kubeClient.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	if err != nil {
+		if kapierrs.IsNotFound(err) {
+			e2e.Logf("%s namespace not found", namespace)
+			return false, nil
+		}
+		e2e.Logf("error accessing %s namespace: %v", namespace, err)
+		return false, err
+	}
+	if ns == nil {
```
simonpasquier, line 2115: Can `ns` ever be nil?

test/extended/util/framework.go
```diff
@@ -2102,6 +2102,24 @@ func DoesApiResourceExist(config *rest.Config, apiResourceName, group string) (b
...
 }
 
+func IsNamespaceExist(kubeClient *kubernetes.Clientset, namespace string) (bool, error) {
+	ns, err := kubeClient.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	if err != nil {
+		if kapierrs.IsNotFound(err) {
+			e2e.Logf("%s namespace not found", namespace)
+			return false, nil
+		}
+		e2e.Logf("error accessing %s namespace: %v", namespace, err)
```
simonpasquier, line 2112: I'd avoid logging the error since it's handled by the caller already.

test/extended/util/framework.go
chiragkyal: Good catch! No, they won't be.  Updated.

test/extended/util/framework.go
chiragkyal: Done!

simonpasquier:
Don't you need the same pre-check for all Prometheus tests (for instance [here](https://github.com/openshift/origin/blob/6e977d1e39511c24af4ce17f340fbd063d4cfc74/test/extended/prometheus/prometheus.go#L54))?

chiragkyal:
>Don't you need the same pre-check for all Prometheus tests (for instance [here](https://github.com/openshift/origin/blob/6e977d1e39511c24af4ce17f340fbd063d4cfc74/test/extended/prometheus/prometheus.go#L54))?

Other tests have `[apigroup:image.openshift.io]` annotation, making them skip for MicroShift. So, as of now this pre-check is not required for them.

simonpasquier:
thanks!
/lgtm

soltysh (approved):
/approve

PR #28071, opened 2023-07-20
Set quota wait timeout to 45 seconds

ardaguclu:
After we've added a cache timeout in quota reconciliation with https://github.com/openshift/cluster-policy-controller/pull/124, we should increase timeout value greater than 30 seconds in these tests because reconciliation
can be completed anytime within 30 seconds and we have still things to do.


test/extended/quota/clusterquota.go
```diff
@@ -26,6 +26,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+// This should be greater than the timeout value in quota reconciliation controller
```
atiratree, line 29:
```suggestion
// This should be greater than the monitor's sync timeout value in cluster quota reconciliation controller (cluster-policy-controller), because it can hang for 30 seconds when a rare deadlock occurs 
```

atiratree:
/lgtm cancel

@ardaguclu  we should do this also for the normal resource quota since it should manifest the same problem

https://github.com/openshift/origin/blob/dd46b809c35b4d811aa8bf6fd6e8cf878b19d799/test/extended/quota/resourcequota.go#L93

upstream tests should be covered: https://github.com/kubernetes/kubernetes/blob/fa88c0b7796170eeff5686ae1d7d0f2f3f0df5de/test/e2e/apimachinery/resource_quota.go#L60

soltysh (approved):
/approve

PR #28065, opened 2023-07-18
trt-1150: move lock so prevent multiple file updates in single commit

neisw:
Min version of [28062](https://github.com/openshift/origin/pull/28062) 

- Just pull in the moved lock for now.  This will slow down commits, we want to evaluate these changes in isolation.

PR #28075, opened 2023-07-21
trt-1150: Set explicit filemode for resource watch 

neisw:
Extracting the resource watch git repo leads to all of the files showing modified status

`diff --git a/cluster-scoped-resources/config.openshift.io/clusteroperators/config-operator.yaml b/cluster-scoped-resources/config.openshift.io/clusteroperators/config-operator.yaml
old mode 100755
new mode 100644`

Due to a change in permissions while extracting

Git can be configured to ignore these changes:
git config core.filemode false

Or we can set the file permissions during the create / modify process.

### DIFF
--- a/pkg/monitor/resourcewatch/storage/git_store.go
+++ b/pkg/monitor/resourcewatch/storage/git_store.go
@@ -57,8 +57,9 @@ func (s *workingSet) waitUntilAvailable(key string) error {
 }
 
 type GitStorage struct {
-	repo *git.Repository
-	path string
+	repo     *git.Repository
+	path     string
+	fileMode os.FileMode
 
 	currentlyRecording workingSet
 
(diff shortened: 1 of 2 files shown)


neisw:
/payload-job periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade

pkg/monitor/resourcewatch/storage/git_store.go
```diff
@@ -90,12 +91,19 @@ func NewGitStorage(path string) (*GitStorage, error) {
 	if err != nil {
 		return nil, err
 	}
-	storage := &GitStorage{path: path, repo: repo}
+	storage := &GitStorage{path: path, repo: repo, fileMode: os.ModePerm}
```
deads2k, line 94: this is 0777 and above is 0644, why differnt?

neisw:
/hold

going with [28065](https://github.com/openshift/origin/pull/28065) for not that has more logging / single file edit/commit locking

pkg/monitor/resourcewatch/storage/git_store.go
neisw: For posterity, the original value is / was 0777.  The code change kept the original but added an opt in change to 0644.  Since we are the only ones using it we don't really need to preserve the original.  It just seemed like a nice thing to do in case others wanted the other / different value at some point.

PR #28077, opened 2023-07-21
add interface for future invariant tests

deads2k:
Adds an interface to use for tests like kube-apiserver disruption inside the cluster, pod to pod communciation failure inside the cluster, and DNS connectivity inside the cluster.

I'm going to continue wiring in the handling for these changes, but I think this interface will remain. The steps for measuring podnetwork-to-podnetwork, podnetwork-to-service, and podnetwork-to-hostnetwork will probably look something like this.

1. create generated namespace with openshift-ci-network-check-xxx (cleanup will destroy this namespace).  
2. create required permissions
3. create deployment with antiaffinity with an http server (not secured, possibly a simple on in this repo) that honors readyz.  Need a target server for pod network and host network, plus services.  
4. create a binary in this repo that takes a target service and uses its endpoint slices to run disruption checks with dynamic endpoints that log errors for connectivity failures that can be parsed out of a log.  The data needs to be stored in local volumes so it survives restarts.  This can probably start as a deployment (honors eviction) to start, but may need to extend to a daemonset (doesn't evict) at some later date.  APIServer team can provide reference of how to use the disruption checker.  Contact @tkashem 
5. add collection to the invarianttest to read the probing pods from each node and summarize results into eventintervals.  TRT can provide reference for how to shape locators and messages to get properly aggregated into the disruption report that is uploaded.  Contact @dgoodwin 

In parallel, @deads2k will wire up the invariantTests in the defaultRegistry (this PR) to be run durign e2e tests and have their methods called. 

/assign @dgoodwin 
/cc @trozet @vrutkovs

### DIFF
--- a/pkg/invariants/types.go
+++ b/pkg/invariants/types.go
@@ -1,0 +1,70 @@
+package invariants
+
+import (
+	"context"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+)
+
+type InvariantTest interface {
+	// StartCollection is responsible for setting up all resources required for collection of data on the cluster.
+	// An error will not stop execution, but will cause a junit failure that will cause the job run to fail.
+	// This allows us to know when setups fail.
+	StartCollection(ctx context.Context, adminRESTConfig *rest.Config) error
+
+	// CollectData will only be called once near the end of execution, before all Intervals are inspected.
+	// Errors reported will be indicated as junit test failure and will cause job runs to fail.
+	CollectData(ctx context.Context) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error)
+
+	// ConstructComputedIntervals is called after all InvariantTests have produced raw Intervals.
+	// Order of ConstructComputedIntervals across different InvariantTests is not guaranteed.
+	// Return *only* the constructed intervals.
+	// Errors reported will be indicated as junit test failure and will cause job runs to fail.
+	ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals) (constructedIntervals monitorapi.Intervals, err error)
+
+	// EvaluateTestsFromConstructedIntervals is called after all Intervals are known and can produce
+	// junit tests for reporting purposes.
+	// Errors reported will be indicated as junit test failure and will cause job runs to fail.
+	EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error)
+
+	// Cleanup must be idempotent and it may be called multiple times in any scenario.  Multiple defers, multi-registered
+	// abort handlers, abort handler running concurrent to planned shutdown.  Make your cleanup callable multiple times.
+	// Errors reported will cause job runs to fail to ensure cleanup functions work reliably.
+	Cleanup(ctx context.Context) error
+}
+
+type InvariantRegistry interface {
+	// AddInvariant adds an invariant test with a particular name, the name will be used to create a testsuite.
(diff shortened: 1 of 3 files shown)


pkg/invariants/types.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+)
+
+type InvariantTest interface {
```
dgoodwin, line 12: I think we need better naming here, Invariant collides hard with a whole bunch of other tests which are not represented as this InvariantTest or in this InvariantRegistry.

Would InClusterInvariantTest be better? With clear godoc that this is for a subset of complex invariant tests that require resources running in the cluster, and collecting data from them at the end of the suite.

pkg/invariants/types.go
```diff
@@ -0,0 +1,43 @@
...
+type InvariantRegistry interface {
+	// AddInvariant adds an invariant test with a particular name, the name will be used to create a testsuite.
+	// The jira component will be forced into every JunitTestCase.
+	AddInvariant(name, jiraComponent string, invariantTest InvariantTest) error
+
+	AddInvariantOrDie(name, jiraComponent string, invariantTest InvariantTest)
+
+	StartCollection(ctx context.Context, adminRESTConfig *rest.Config) ([]*junitapi.JUnitTestCase, error)
+
+	// CollectData will only be called once.  Errors reported will be indicated as junit test failure and will
```
dgoodwin, line 36: Lets clarify when it would be called that one time.

pkg/invariants/types.go
```diff
@@ -0,0 +1,43 @@
...
+	// cause job runs to fail.
+	CollectData(ctx context.Context) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error)
+
+	// Cleanup must be callable times without issue.  Errors reported will cause job runs to fail to ensure cleanup
```
dgoodwin, line 40: Please clarify why it must be callable multiple times. Suggest "Cleanup should be idempotent and it may be called multiple times in [insert scenario here]".

pkg/invariants/types.go
```diff
@@ -0,0 +1,43 @@
...
+
+type InvariantTest interface {
+	// StartCollection is responsible for setting up all resources required for collection of data on the cluster.
+	// An error will not stop execution, but will cause a junit failure that will cause the job run to fail.
+	// This allows us to know when setups fail.
+	StartCollection(ctx context.Context, adminRESTConfig *rest.Config) error
+
+	// CollectData will only be called once.  Errors reported will be indicated as junit test failure and will
+	// cause job runs to fail.
+	CollectData(ctx context.Context) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error)
```
dgoodwin, line 20: At lot of invariant tests today require other intervals from the cluster to make complex decisions. For this reason it might be more futureproof to split this into CollectData() (Intervals, error) and RunTests(allClusterIntervals) ([]JunitTestCase, error).

pkg/invariants/types.go
deads2k: > I think we need better naming here, Invariant collides hard with a whole bunch of other tests which are not represented as this InvariantTest or in this InvariantRegistry.
> 
> Would InClusterInvariantTest be better? With clear godoc that this is for a subset of complex invariant tests that require resources running in the cluster, and collecting data from them at the end of the suite.

Nothing about this limits us to in-cluster only.  It has the option of making in-cluster stuff, but it could also do nothing.

deads2k: And given that, would you like to collapse them here?

pkg/invariants/types.go
deads2k: added

PR #28085, opened 2023-07-24
convert pod watching to invariant style checking

deads2k:
built on several other commits.  Let's see how this compiles and runs.  The cleanup here is minimal and the junits aren't being written out (yet), but the intervals should come through cleanly.


deads2k:
closed in favor of https://github.com/openshift/origin/pull/28086/

PR #28086, opened 2023-07-24
add a serialization step to the monitor

deads2k:


deads2k:
/retest

pod states look good.

PR #28088, opened 2023-07-24
STOR-1315: Add e2e workflow for Secrets Store CSI Driver

dobsonj:
[STOR-1315](https://issues.redhat.com//browse/STOR-1315): Add e2e workflow for Secrets Store CSI Driver

This can be tested manually with:
```
make update && make && ./openshift-tests run all --dry-run | grep SecretsStore | ./openshift-tests run -f -
```

The change in `pkg/test/ginkgo/test_runner.go` resolves a nil pointer dereference that I hit with that command where `exutilcluster.LoadConfig(clusterState)` returns nil and then we still try to dereference `config.ProviderName` a couple lines below.

This test requires a DaemonSet that is deployed as part of the workflow in https://github.com/openshift/release/pull/41655 -- but the test is skipped in `openshift-tests` if the CSIDriver does not exist.

cc @openshift/storage

### DIFF
--- a/test/extended/storage/csi-secret-store.go
+++ b/test/extended/storage/csi-secret-store.go
@@ -1,0 +1,131 @@
+package storage
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"github.com/google/uuid"
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8simage "k8s.io/kubernetes/test/utils/image"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	"github.com/openshift/library-go/pkg/build/naming"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const (
+	csiSecretsStoreDriver = "secrets-store.csi.k8s.io"
+	testContainerName     = "test"
+	podWaitTimeout        = 2 * time.Minute
+)
+
+var _ = g.Describe("[sig-storage][Feature:SecretsStore][Serial]", func() {
+	defer g.GinkgoRecover()
+	var (
+		oc                     = exutil.NewCLIWithPodSecurityLevel("secrets-store-test-ns", admissionapi.LevelPrivileged)
+		ctx                    = context.Background()
+		baseDir                = exutil.FixturePath("testdata", "storage", "csi-secret-store")
+		e2eSecretProviderClass = filepath.Join(baseDir, "e2e-provider-secretproviderclass.yaml")
+	)
+
+	g.BeforeEach(func() {
+		exutil.PreTestDump()
+
+		// check if the CSIDriver exists, skip if not found
+		_, err := oc.AdminKubeClient().StorageV1().CSIDrivers().Get(ctx, csiSecretsStoreDriver, metav1.GetOptions{})
+		if err != nil {
+			g.Skip(fmt.Sprintf("CSIDriver %s not found", csiSecretsStoreDriver))
+		}
+
+		// Allow creation of privileged pods for this test.
+		// e2e-provider must be privileged to bind to a unix domain socket on the host.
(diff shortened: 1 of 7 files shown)


test/extended/testdata/storage/csi-secret-store/e2e-provider-installer.yaml
```diff
@@ -0,0 +1,52 @@
...
+      app: csi-secrets-store-e2e-provider
+  template:
+    metadata:
+      labels:
+        app: csi-secrets-store-e2e-provider
+    spec:
+      serviceAccountName: default
+      containers:
+        - name: e2e-provider
+          image: quay.io/jdobson/secrets-store-csi-driver-e2eprovider
```
jsafrane, line 22: There is a process how to start using a new image in tests: https://github.com/openshift/origin/blob/master/test/extended/util/image/README.md
It would be great if we did not need to go through it.

jsafrane: I assume a generic centos image + a shell script is not possible because of grpc. Then we need a new image + some openshift/ github repo to build it from it. Can we add it to the operator repo and build it in CI?

dobsonj: I was originally trying to find a way to pass an image (built from CI) to openshift-tests, but this is not the right way to do it. IMO, the e2eprovider needs to be built from the driver repo first and then deployed by a script in step-registry before openshift-tests runs.

So I have a couple of PR's now to build e2eprovider from the driver repo:
https://github.com/openshift/secrets-store-csi-driver/pull/5
https://github.com/openshift/release/pull/41693
Once both of those are merged, I think I can rework this PR to deploy e2eprovider as part of the e2e workflow:
https://github.com/openshift/release/pull/41655
And if all of that actually works, then I'll update this PR so that openshift-tests does not try to deploy this anymore.

dobsonj: > I assume a generic centos image + a shell script is not possible because of grpc.

This might be possible with a python script, but I haven't tried it. I can try that approach if the e2eprovider image turns out not to work for some reason.

test/extended/storage/csi-secret-store.go
```diff
@@ -0,0 +1,131 @@
...
+		// Allow creation of privileged pods for this test.
+		// e2e-provider must be privileged to bind to a unix domain socket on the host.
+		// The test pod must be privileged to read the files created by e2e-provider.
+		g.By("adding privileged SCC to namespace")
+		sa := fmt.Sprintf("system:serviceaccount:%s:default", oc.Namespace())
+		err = oc.AsAdmin().Run("adm").Args("policy", "add-scc-to-user", "privileged", sa).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("creating e2e-provider SecretProviderClass")
+		err = oc.AsAdmin().Run("apply").Args("-f", e2eSecretProviderClass).Execute()
```
gnufied, line 55: nit: do we need to delete providerclass?

test/extended/storage/csi-secret-store.go
```diff
@@ -0,0 +1,131 @@
...
+	defer g.GinkgoRecover()
+	var (
+		oc                     = exutil.NewCLIWithPodSecurityLevel("secrets-store-test-ns", admissionapi.LevelPrivileged)
+		ctx                    = context.Background()
+		baseDir                = exutil.FixturePath("testdata", "storage", "csi-secret-store")
+		e2eSecretProviderClass = filepath.Join(baseDir, "e2e-provider-secretproviderclass.yaml")
+	)
+
+	g.BeforeEach(func() {
+		exutil.PreTestDump()
```
gnufied, line 38: nit: `PreTestDump` doesn't seem to do anything.

dobsonj: Heh, whoops, I removed it from the storage tests.

test/extended/storage/csi-secret-store.go
dobsonj: It's installed to a temporary namespace and it gets deleted at the end of the test when the namespace is deleted.

PR #28089, opened 2023-07-24
moves and renames to make the openshift-tests execution more oc-like …

deads2k:
…and forward building

Straight moves and renames to take start taking stock of what we have.  The ultimate vision is to the commands abstracted into options that are one per-package with selective re-use.

/assign @dgoodwin


deads2k:
CI broke somewhere on this, history says unrecoverable.  Closing in favor of https://github.com/openshift/origin/pull/28090 which contains it.

PR #28094, opened 2023-07-25
wait for API sampler completion

deads2k:
Commits on top of the revert of the reverts are making sure that when the samplers are finished, we wait for completion of the consumers.  This is not done in a "pretty" way, but it's a stop gap until the overall refactor is more complete.

/hold for testing


pkg/monitor/monitor.go
```diff
@@ -1,227 +1,148 @@
...
 	}
+	m.stopFn()
+	m.stopFn = nil
+
+	// we don't want this method to return until all te additional recorders and invariants have completed processing.
+	// to do this correctly, we need closure channels or some kind of mechanism.
+	// rather than properly wire this through, we'll wait until the backend disruption consumers timeout after the producer
+	// close.
+	// TODO once we have converted the backendsamplers to invariant tests, we can properly wait for completion
+	time.Sleep(70 * time.Second)
```
deads2k, line 100: This is the meat of what I added with an explanation as to why.  We can refine/fix during the refactor, but trying to wire it all properly now will be massive.

deads2k:
if we run grep "started responding to GET requests" '/home/deads/Downloads/new-events/e2e-events_20230725-014444.json'   | grep -- "-api " on the upgrade file of e2e-events***.json, we should see all 12 types of connections listed.  That the runs without collection fail this test and runs with collections survive it.

PR #28096, opened 2023-07-25
move serialization into invariant tests

deads2k:
This gives a clear way to add serialization in a structured way so that the serialization  methods are in a single spot and consistent across all types of initialization so we can start collapsing our stop paths.


deads2k:
/payload 4.14 ci blocking

This PR makes the serialization parameters a little more straightforward, so we know where our events are coming from.  It might be necessary to *also* read the additional events from disk first now that I'm thinking about it because of the ordering

deads2k:
collapsing into https://github.com/openshift/origin/pull/28100 .  I think we have push to the other side of the refactor and https://github.com/openshift/origin/pull/28100 does that for the harder disruption checks. I'll biuld another to do the rest.

PR #28098, opened 2023-07-26
Add a test to flake if we see NetworkManager log "too many netlink events"

dgoodwin:
[TRT-1022](https://issues.redhat.com//browse/TRT-1022)


pkg/monitor/intervalcreation/node.go
```diff
@@ -238,6 +247,44 @@ func unreasonablyLongPollInterval(logLine, nodeLocator string) monitorapi.Interv
...
+
+	logTime := systemdJournalLogTime(logLine)
+
+	message := logLine[strings.Index(logLine, "NetworkManager"):]
+	return monitorapi.Intervals{
+		{
+			Condition: monitorapi.NewCondition(monitorapi.Warning).Locator(
+				nodeLocator).Message(monitorapi.NewMessage().HumanMessage(message)).Build(),
+			From: logTime,
+			To:   logTime,
```
neisw, line 283: If we ever want to show these in an interval chart we need to add a second here.  Good for now,  not sure we have developed a standard for yet but I know we do in some places.

https://github.com/openshift/origin/blob/377119ceb75412930ed6679b8d4fda738ea944f5/pkg/monitor/intervalcreation/node.go#L537

pkg/monitor/intervalcreation/node_test.go
```diff
@@ -253,13 +262,39 @@ func TestMonitorApiIntervals(t *testing.T) {
...
+					},
+					StructuredMessage: monitorapi.Message{
+						Reason:       "",
+						Cause:        "",
+						HumanMessage: "NetworkManager[1155]: <info> [1681300187.8326] platform-linux: netlink[rtnl]: read: too many netlink events. Need to resynchronize platform cache",
+						Annotations:  map[monitorapi.AnnotationKey]string{},
+					},
+				},
+				From: systemdJournalLogTime("Apr 12 11:49:49.188086"),
+				To:   systemdJournalLogTime("Apr 12 11:49:49.188086"),
```
neisw, line 288: For the unit test failure, I would expect with the recent code change it was telling us it should be 11:49:50 for the To Date but is this instead saying no To value has been set?

             	            	expected: time.Date(2023, time.April, 12, 11, 49, 49, 188086000, time.UTC)
            	            	actual  : time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)

neisw: I might have missed this in a previous review  https://github.com/dgoodwin/origin/blob/3d2edd51aaf2279f2810e33289a4125ab096dc87/pkg/monitor/monitorapi/construction.go#L59

dgoodwin:
Ready for review, looks to be working.

PR #28100, opened 2023-07-26
add collector for additionalevents__ from other processes

deads2k:
This might be the problem.  The analysis needs the data from the additionalEvents for an upgrade job.


dgoodwin:
```
 Suite run returned error: Route.route.openshift.io "test-disruption-" is invalid: metadata.name: Invalid value: "test-disruption-": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
No manifest filename passed
Shutting down SimultaneousPodIPController
error: Route.route.openshift.io "test-disruption-" is invalid: metadata.name: Invalid value: "test-disruption-": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
SimultaneousPodIPController shut down 
```

pkg/invariants/impl.go
```diff
@@ -85,77 +82,73 @@ func (r *invariantRegistry) StartCollection(ctx context.Context, adminRESTConfig
...
-					Duration: duration.Seconds(),
-					FailureOutput: &junitapi.FailureOutput{
-						Output: fmt.Sprintf("failed during collection\n%v", err),
-					},
-					SystemOut: fmt.Sprintf("failed during collection\n%v", err),
-				})
-				continue
-			}
+	for _, invariant := range r.invariantTests {
+		testName := fmt.Sprintf("jira/%q invariant test %v collection", invariant.jiraComponent, invariant.name)
```
dgoodwin, line 91: Why the different syntax here from our normal [component] prefix?

pkg/invariants/availability_image_registry/invariant.go
```diff
@@ -0,0 +1,212 @@
...
+	w.kubeClient, err = kubernetes.NewForConfig(adminRESTConfig)
+	if err != nil {
+		return err
+	}
+	w.routeClient, err = routeclient.NewForConfig(adminRESTConfig)
+	if err != nil {
+		return err
+	}
+
+	w.imageRegistryRoute, err = imageregistryutil.ExposeImageRegistryGenerateName(ctx, w.routeClient, "test-disruption-")
```
dgoodwin, line 58: This might be where an invalid route is being created. Need to drop the dashes I think?

pkg/invariants/watch_pods/invariant.go
```diff
@@ -0,0 +1,56 @@
...
+func (*podWatcher) ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals, recordedResources monitorapi.ResourcesMap, beginning, end time.Time) (monitorapi.Intervals, error) {
+	constructedIntervals := monitorapi.Intervals{}
+	constructedIntervals = append(constructedIntervals, createPodIntervalsFromInstants(startingIntervals, recordedResources, beginning, end)...)
+	constructedIntervals = append(constructedIntervals, intervalsFromEvents_PodChanges(startingIntervals, beginning, end)...)
+
+	return constructedIntervals, nil
+}
+
+func (*podWatcher) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
+	return nil, nil
```
dgoodwin, line 46: Presumably in the future, a lot of invariants from event_junits.go could move here if they were related to the intervals coming out of the pod watcher?

pkg/monitor/backenddisruption/disruption_backend_sampler.go
```diff
@@ -525,6 +545,8 @@ func (b *disruptionSampler) consumeSamples(ctx context.Context, interval time.Du
 	// when we exit this function, we want to set a final duration of failure.  We don't actually know whether it ended
 	// or how long it took to ask
 	defer func() {
+		fmt.Fprintf(os.Stderr, "%v setting the intervals as done at time %v\n", b.backendSampler.locator, time.Now())
```
dgoodwin, line 548: Why is this stderr? 

Would prefer we use logrus whenever possible, consistent structured logging would be a win for us, we're often staring at unstamped log messages wondering what happened when and how long something took.

dgoodwin (commented):
I added more on https://github.com/openshift/origin/pull/28104 which seems to have a lot of overlap with this one.

pkg/invariants/impl.go
deads2k: updated in #28104

pkg/invariants/availability_image_registry/invariant.go
deads2k: it was setting both name and prefix.  fixed in updated in #28104

pkg/invariants/watch_pods/invariant.go
deads2k: yes

deads2k:
replaced https://github.com/openshift/origin/pull/28104

PR #28101, opened 2023-07-26
Update image mirroring instructions to handle uncompressed layers

bparees:


bparees:
@soltysh @jupierce fyi

jaypoulz:
Looks fine to me. My only suggestion would be to consider updating the mirroring command to filter out the problematic image.

For a single image like this, we could suggest using grep to filter it out:
```
openshift-tests images --upstream --to-repository quay.io/openshift/community-e2e-images | grep -v "e2e-27-registry-k8s-io-pause-3-9-p9APyPDU5GsW02Rk"  | oc image mirror -f - --filter-by-os=.*
```

This is an OK way to solve this instance, but I think longer term, we should look into seeing if we can update the `openshift-tests images` command to take a flag to filter out images with uncompressed layers. It may be possible to auto-detect that with some container commands, but probably the best performance is to maintain a exclude list. I'm still interested in understanding why that command output a subset of images to mirror to community-e2e-images instead of my repository, so I can look into this if you think any of these approaches are suitable.

bparees:
> Looks fine to me. My only suggestion would be to consider updating annotated mirroring command to filter out the problematic image.

since i don't think it prevents the successful mirroring of the rest of the images, i'm going to leave it for now with the hope that this problem goes away at some point.

PR #28102, opened 2023-07-27
feat: make data gathering more robust

eggfoobar:
added retry logic to lesson chance of flakes during runs 
removed custom exec logic and leveraged framework pod package to execute commands on pods

### DIFF
--- a/test/extended/cpu_partitioning/crio.go
+++ b/test/extended/cpu_partitioning/crio.go
@@ -1,10 +1,11 @@
 package cpu_partitioning
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	g "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


test/extended/cpu_partitioning/crio.go
```diff
@@ -276,63 +275,68 @@ func collectContainerInfo(ctx context.Context, cfg *rest.Config, c kubernetes.In
...
+		PreserveWhitespace: false,
 	}
 
 	info := []crioContainerData{}
-	if err := json.Unmarshal(out, &info); err != nil {
-		return nil, fmt.Errorf("error parsing container output to json: %w", err)
-	}
+	err = retry.OnError(backoff, shouldRetryExec,
+		func() error {
+			out, _, execErr := e2epod.ExecWithOptions(oc.KubeFramework(), execOptions)
```
ffromani, line 309: the retry logic seems nice, but why using this code is better than the previous code? just avoiding code duplication or other reasons?

eggfoobar: yeah, noticed this had some nicer options (like logging the request) and figured we could avoid code duplications and hopefully less things for us to manage and worry about

test/extended/cpu_partitioning/crio.go
```diff
@@ -276,63 +275,68 @@ func collectContainerInfo(ctx context.Context, cfg *rest.Config, c kubernetes.In
 	}
 
 	pod := &pods.Items[0]
+	podName := pod.Name
+	podNamespace := pod.Namespace
+	containerName := pod.Spec.Containers[0].Name
```
ffromani, line 280: not sure we need these local vars, but it won't hurt either

ffromani (commented):
/lgtm

PR #28103, opened 2023-07-27
WIP: Sketch out ideas around completing interval model

dgoodwin:
I am struggling with classification of intervals.

I believe that intervals should have some kind of type/category information that indicates where they came from and what they represent.

We need UI groupings that are sometimes linked directly 1-1 with type information, but also sometimes require sub-classification. 

And I cannot decide if we should tie view specifics into the go code (allowing pr authors to add new intervals AND specify that they should be shown in the chart, in what color, all in one PR), or if they should remain in the JS UI (in sippy) allowing us to retroactively decide what to show for jobs that have already run and stored their intervals.

### DIFF
--- a/pkg/monitor/alerts.go
+++ b/pkg/monitor/alerts.go
@@ -381,11 +381,10 @@ func CreateEventIntervalsForAlerts(ctx context.Context, alerts prometheustypes.V
 			default:
 				level = monitorapi.Error
 			}
-			alertIntervalTemplate := monitorapi.Interval{
-				Condition: monitorapi.NewCondition(level).
-					Locator(lb).
-					Message(monitorapi.NewMessage().HumanMessage(alert.Metric.String())).Build(),
-			}
+			alertIntervalTemplate := monitorapi.NewInterval(monitorapi.IntervalCategoryAlert, level).
+				Locator(lb).
+				Message(monitorapi.NewMessage().
+					HumanMessage(alert.Metric.String())).Build()
 
 			var alertStartTime *time.Time
 			var lastTime *time.Time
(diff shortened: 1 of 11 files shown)


pkg/monitor/alerts.go
```diff
@@ -381,11 +381,10 @@ func CreateEventIntervalsForAlerts(ctx context.Context, alerts prometheustypes.V
...
 			}
-			alertIntervalTemplate := monitorapi.Interval{
-				Condition: monitorapi.NewCondition(level).
-					Locator(lb).
-					Message(monitorapi.NewMessage().HumanMessage(alert.Metric.String())).Build(),
-			}
+			alertIntervalTemplate := monitorapi.NewInterval(monitorapi.IntervalCategoryAlert, level).
+				Locator(lb).
+				Message(monitorapi.NewMessage().
+					HumanMessage(alert.Metric.String())).Build()
```
dgoodwin, line 387: Sample of interval creation that compiles with this change, we force specification of a type.

pkg/monitor/serialization/serialize.go
```diff
@@ -101,6 +109,19 @@ func EventsIntervalsToJSON(events monitorapi.Intervals) ([]byte, error) {
...
+	// to only specify ViewHint if the condition is False)
+	//
+	// Should the UI be able to "optionally" show intervals that do not have ViewHints?
+	//
+	// Or should we limit what we display to only those that this code said to? The problem there
+	// of course is that we're mixing view logic into our model, and the result is that we could
+	// not retroactively decide to display something in old job runs.
+	//
+	// It would however mean that origin authors writing new intervals could, in the same PR,
+	// specify that they should be displayed and how.
```
dgoodwin, line 123: Proposal for adding in view hints at the serialization stage.

PR #28106, opened 2023-07-27
remove apparently unused --output-file param, callers should redirect

deads2k:


deads2k:
Oh lovely, with shorthand. Mess. Ok

PR #28111, opened 2023-07-31
fix: workload partitioning - added more error handling for retry attempts

eggfoobar:
I'm noticing some continued flakes and errors crop up around this test.

Some of the flakes below seem to be empty array errors, which should not be the case, so we might be eating an error there. Added some logic to print our any `stderr` and also retry if it's empty in a cluster where cpu partitioning is expected.

https://testgrid.k8s.io/redhat-openshift-ocp-release-4.14-informing#periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-ovn-cpu-partitioning&include-filter-by-regex=CPU%20Partitioning

I also noticed some [sippy](https://sippy.dptools.openshift.org/sippy-ng/jobs/4.14/runs?filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22failed_test_names%22%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22%5Bsig-node%5D%5Bapigroup%3Aconfig.openshift.io%5D%20CPU%20Partitioning%20node%20validation%20should%20have%20correct%20cpuset%20and%20cpushare%20set%20in%20crio%20containers%20%5BSuite%3Aopenshift%2Fconformance%2Fparallel%5D%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22never-stable%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22aggregated%22%7D%5D%2C%22linkOperator%22%3A%22and%22%7D&sortField=timestamp&sort=desc) errors and flakes around connection issues and odd states such as `machine-config-daemon` not being available, this seems to be related to heterogenous clusters which alter a machine, the timing might be off when the node is coming back up and the machine-config-daemon is not running yet.

### DIFF
--- a/test/extended/cpu_partitioning/crio.go
+++ b/test/extended/cpu_partitioning/crio.go
@@ -306,9 +306,12 @@ func collectContainerInfo(ctx context.Context, oc *exutil.CLI, node corev1.Node)
 	info := []crioContainerData{}
 	err = retry.OnError(backoff, shouldRetryExec,
 		func() error {
-			out, _, execErr := e2epod.ExecWithOptions(oc.KubeFramework(), execOptions)
+			out, outStdErr, execErr := e2epod.ExecWithOptions(oc.KubeFramework(), execOptions)
 			if execErr != nil {
 				return execErr
+			}
+			if outStdErr != "" {
+				return fmt.Errorf("err execing command %s", outStdErr)
 			}
 			if err := json.Unmarshal([]byte(out), &info); err != nil {
 				return fmt.Errorf("error parsing container output to json: %w", err)
@@ -321,22 +324,29 @@ func collectContainerInfo(ctx context.Context, oc *exutil.CLI, node corev1.Node)
 
 // Due to cluster activity, sometimes execing against a pod can be flakey.
 // We retry if the below errors are received during node queries.
-//
-// JSON Syntax error:
-//
-//	This typically means that something interrupted the stream of json coming back, so we need to do the query again.
-//
-// Dialing timeout:
-//
-//	Some sort of disruption is happening causing the exec to time out, retry.
 func shouldRetryExec(err error) bool {
 	var parseErr *json.SyntaxError
+	substringErrors := []string{
+		"error dialing backend: dial tcp", // Some sort of disruption is happening causing the exec to time out, retry.
+		"container not found",             // On occasion there might be situations where the machine-config-daemon isn't running yet.
+		"err execing command",             // Error when running the command in the container
+	}
 	switch {
+	// This typically means that something interrupted the stream of json coming back, so we need to do the query again.
 	case errors.As(err, &parseErr):
 		return true
-	case strings.Contains(err.Error(), "error dialing backend: dial tcp"):
+	case substringSliceContains(substringErrors, err.Error()):
 		return true
 	default:
 		return false
 	}
+}
+
+func substringSliceContains(slice []string, s string) bool {
+	for _, substring := range slice {
+		if strings.Contains(s, substring) {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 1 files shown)


test/extended/cpu_partitioning/crio.go
```diff
@@ -306,10 +306,13 @@ func collectContainerInfo(ctx context.Context, oc *exutil.CLI, node corev1.Node)
...
 	err = retry.OnError(backoff, shouldRetryExec,
 		func() error {
-			out, _, execErr := e2epod.ExecWithOptions(oc.KubeFramework(), execOptions)
+			out, outStdErr, execErr := e2epod.ExecWithOptions(oc.KubeFramework(), execOptions)
 			if execErr != nil {
 				return execErr
 			}
+			if outStdErr != "" {
+				return fmt.Errorf("err execing command %s", outStdErr)
+			}
```
ffromani, line 315: I'm still unconvinced that having -any- data on stderr means we failed something. At very least this should be bound to the specific command we run (and still we're assuming this never changes) with a code comment.

test/extended/cpu_partitioning/crio.go
```diff
@@ -321,22 +324,29 @@ func collectContainerInfo(ctx context.Context, oc *exutil.CLI, node corev1.Node)
...
+		"error dialing backend: dial tcp", // Some sort of disruption is happening causing the exec to time out, retry.
+		"container not found",             // On occasion there might be situations where the machine-config-daemon isn't running yet.
+		"err execing command",             // Error when running the command in the container
+	}
 	switch {
+	// This typically means that something interrupted the stream of json coming back, so we need to do the query again.
 	case errors.As(err, &parseErr):
 		return true
-	case strings.Contains(err.Error(), "error dialing backend: dial tcp"):
+	case substringSliceContains(substringErrors, err.Error()):
```
ffromani, line 341: can we use `errors.As` also here? compare errors with known values? matching the error string should really be the last resort.

test/extended/cpu_partitioning/crio.go
```diff
@@ -313,6 +313,9 @@ func collectContainerInfo(ctx context.Context, oc *exutil.CLI, node corev1.Node)
 			if outStdErr != "" {
 				return fmt.Errorf("err execing command %s", outStdErr)
 			}
+			if isClusterCPUPartitioned && len(out) == 0 {
+				return fmt.Errorf("err execing command, script result was empty but expected values")
+			}
```
ffromani, line 318: we observed this happening in the wild? what was the context?

ffromani (commented):
In general I'd appreciate more context giving rationale for the changes. The changes per se look generally reasonnable, but we should aim to minimize the guesswork involved and try hard to present and record context which drove the changes.

eggfoobar:
> In general I'd appreciate more context giving rationale for the changes. The changes per se look generally reasonnable, but we should aim to minimize the guesswork involved and try hard to present and record context which drove the changes.

Ah, apologies, got a bit tunnel vision with this test. I've updated the description with some more context.

test/extended/cpu_partitioning/crio.go
eggfoobar: I might me misunderstanding then. In our scenario here we're executing this script in the container. Would it not be only errors that come through the stderr in this situation?

test/extended/cpu_partitioning/crio.go
eggfoobar: I did try and dig through the return types, unfortunately I didn't find a great way to get a guarantee on the error type being thrown there, so the string seemed to be the next sure thing.

test/extended/cpu_partitioning/crio.go
eggfoobar: Yeah, that was my bad, updated the description with info on that info. In this example job you'll see it flake with a an empty array

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-ovn-cpu-partitioning/1685720543094902784

ffromani (commented):
/lgtm

thanks for the extra context

PR #28116, opened 2023-08-01
Remove enabling multinetwork policy

trozet:
This is now enabled by default in aws serial jobs:

https://github.com/openshift/release/pull/41872

### DIFF
--- a/test/extended/networking/multinetpolicy.go
+++ b/test/extended/networking/multinetpolicy.go
@@ -11,11 +11,10 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 	operatorclientv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
 	v1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/util/retry"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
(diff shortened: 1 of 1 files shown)


test/extended/networking/multinetpolicy.go
```diff
@@ -33,19 +28,6 @@ var _ = g.Describe("[sig-network][Feature:MultiNetworkPolicy][Serial][apigroup:o
...
 
-	// Ensure that MultiNetPolicy stays in the same enabled/disabled state
-	// after the test, that it was in before the test
-	var previousUseMultiNetworkPolicy bool
-	g.BeforeEach(func() {
-		previousUseMultiNetworkPolicy = isMultinetNetworkPolicyEnabled(oc)
-	})
-
-	g.AfterEach(func() {
-		if !previousUseMultiNetworkPolicy {
```
tssurya, line 44: `previousUseMultiNetworkPolicy` should be true since https://github.com/openshift/release/pull/41872 merged.. but I see CI failing, retest the job.. let's see...

trozet:
> ideally speaking we don't have to do this PR since this is required only when we run this test in lanes that doesn't have multi-network policy enabled by default right? -> like its no-op in aws-serial lane, but tomorrow if we add this test to other non-aws-serial lanes, then we need to manually enable this flag via test so wondering it there is use in just leaving this as is? Or folks have to be careful to not add this test to any of the non-aws-ovn-serial lanes because those lanes don't have this feature enabled by default. EDIT: if we do decide to keep this stuff we should fix the wait part for CNO

we still need the PR cause the test case tries to disable the feature as well

PR #28127, opened 2023-08-03
remove extraneous layering for the monitor

deads2k:
This unwraps the monitor.  Still need to collect the invariant testing.  Need to decide how.


deads2k:
simple rebase, reapplying.

deads2k:
rebased reapplying previous lgtm.

PR #28128, opened 2023-08-03
expedient engineering: inject intervals into recorded events from out…

deads2k:
…side the monitor to serialize into timeline

https://github.com/openshift/origin/pull/28127 has the long term solution, but if it doesn't pass CI, this is small and obvious.

/assign @stbenjam 
/cc @trozet


deads2k:
/hold

https://github.com/openshift/origin/pull/28127 now testing clean and this needs retesting now

deads2k:
/hold cancel

the other one has an intermitent watchdog failure.  releasing.

PR #28130, opened 2023-08-03
move existing invariant tests

deads2k:
this is going to fail units, but I'm going to lose cell signal for a couple hours.  Let's see what it does otherwise.


deads2k:
Built on the last one that merged. Will merge on the next green payload

deads2k:
payload failed on sandbox events, not disruption.  Going for this one next.  It has just three commits on the other.

PR #28135, opened 2023-08-04
remove [sig-arch] Monitor cluster while tests execute test

deads2k:
/assign @dgoodwin 

This test provides no unique value.  The intervals are serialized to a file already and this is always true.


dgoodwin:
Checked other tooling and nothing seems to rely on this. Sippy explicitly ignores it.

/approve
/lgtm

But has conflicts.

deads2k:
merged in other PRs

PR #28144, opened 2023-08-07
 move remaining disruption tests to invariants

deads2k:
dropped new apiserver testing that Vadim has a separate PR working on.


deads2k:
looks to be working.  `ci-cluster-network-liveness` still exists in backends and would have failed intervals, but it doesn't have associated test because it never fails.

pkg/cmd/openshift-tests/dev/dev.go
```diff
@@ -153,6 +152,9 @@ a running cluster.
 `),
 
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if true {
+				return fmt.Errorf("this command got nerfed")
+			}
```
neisw, line 157: I have to at least ask, you really want this in here?

pkg/invariantlibrary/allowedbackenddisruption/matches.go
```diff
@@ -21,10 +19,5 @@ const (
 // GetAllowedDisruption uses the backend and information about the cluster to choose the best historical p95 to operate against.
 // We enforce "don't get worse" for disruption by watching the aggregate data in CI over many runs.
 func GetAllowedDisruption(backendName string, jobType platformidentification.JobType) (*time.Duration, string, error) {
-	// Special case for the liveness probe to external service:
-	if backendName == externalservice.LivenessProbeBackend {
-		aed := allowedExternalDisruption
-		return &aed, "forgiving limit for disruption to an external service", nil
-	}
```
neisw, line 28: IIRC This was in place so we didn't fail due to our check to see if the cluster running the tests was having connection issues.

pkg/cmd/openshift-tests/dev/dev.go
deads2k: What is this command trying to do and does it make sense this way with invariant tests?  Would a rewrite when we need it again make more sense?

I'm up for removing the command of you're down with it

pkg/invariantlibrary/allowedbackenddisruption/matches.go
deads2k: The test was always ok. I've left the backend check and it reports intervals and overall disruption. Do we also need the "never fail" test?

pkg/cmd/openshift-tests/dev/dev.go
neisw: As far as removing it goes I would be interested on checking with @dgoodwin to see what his original intent was and if it is still of use. If leaving it disabled then a comment regarding why it was disabled would suffice.

dgoodwin: This command lets developers work on disruption test code and run it against a given intervals file containing disruption we're interested in testing against, getting feedback in seconds rather than hours waiting for CI to run. I would prefer to keep it operational.

With your new testing, would it be possible to run through the junit generation parts of your interface, but skip the setup/generate intervals portions.

dgoodwin: It is supposed to run the "should remain available" disruption tests against the historical data on disk. It assumes it's already given an intervals file with the observed disruption. In your list, #5, plus running the tests and viewing the output to see what would fail and with what values.

Deep is going to need similar very soon for alerts.

neisw:

Failure for `[sig-apps] Daemon set [Serial] should surge pods onto nodes when spec was updated and update strategy is RollingUpdate` is not related

PR #28146, opened 2023-08-08
Disable EgressIP test termporarily due to OVN-K bug

jcaamano:
A specific EgressIP test is broken with OVN-K in interconnect mode. Disable temporarily while the bug is fixed.


tssurya:
we need to do the auto generated code changes to zz_generated.annotations.go file, there is a specific make command that does the trick for you

trozet (approved):
/lgtm

soltysh (approved):
/approve

PR #28147, opened 2023-08-08
OCPBUGS-16166: Update to Kubernetes 1.27.4

soltysh:
Goes in pair with https://github.com/openshift/kubernetes/pull/1660, but needs to be updated with correct deps after the other lands. 

/assign @bertinatto 

/hold


bertinatto (commented):
/lgtm

PR #28149, opened 2023-08-08
TRT-930: Update test to support 2 release query_results.json

deepsm007:
This is the second part to https://github.com/openshift/ci-tools/pull/3562 PR for origin test to support both releases .
Both alerts and disruption tests are updated along with query_results.json.
/cc @dgoodwin @neisw

### DIFF
--- a/pkg/monitortestlibrary/allowedalerts/matches_test.go
+++ b/pkg/monitortestlibrary/allowedalerts/matches_test.go
@@ -208,14 +208,14 @@ func TestAlertDataFileParsing(t *testing.T) {
 			foundMetalOVN = true
 		}
 	}
-
+	currentRelease = historicaldata.CurrentReleaseFromMap(releasesInQueryResults)
 	assert.Greater(t, dataOver100Runs, 5,
 		"expected at least 5 entries in query_results.json to have over 100 runs")
 	assert.True(t, foundAWSOVN, "no aws ovn job data in query_results.json")
 	assert.True(t, foundGCPOVN, "no gcp ovn job data in query_results.json")
 	assert.True(t, foundAzureOVN, "no azure ovn job data in query_results.json")
 	assert.True(t, foundMetalOVN, "no metal ovn job data in query_results.json")
-	assert.Equal(t, 1, len(releasesInQueryResults),
+	assert.Equal(t, 2, len(releasesInQueryResults),
 		"expected only one Release in query_results.json")
 
 	// Check that we get a real value for something we know should be there for every release. This alert
--- a/pkg/monitortestlibrary/allowedbackenddisruption/matches_test.go
+++ b/pkg/monitortestlibrary/allowedbackenddisruption/matches_test.go
@@ -274,14 +274,14 @@ func TestDisruptionDataFileParsing(t *testing.T) {
 			foundMetalOVN = true
 		}
 	}
-
+	currentRelease = historicaldata.CurrentReleaseFromMap(releasesInQueryResults)
 	assert.Greater(t, dataOver100Runs, 5,
 		"expected at least 5 entries in query_results.json to have over 100 runs")
 	assert.True(t, foundAWSOVN, "no aws ovn job data in query_results.json")
 	assert.True(t, foundGCPOVN, "no gcp ovn job data in query_results.json")
 	assert.True(t, foundAzureOVN, "no azure ovn job data in query_results.json")
 	assert.True(t, foundMetalOVN, "no metal ovn job data in query_results.json")
-	assert.Equal(t, 1, len(releasesInQueryResults),
+	assert.Equal(t, 2, len(releasesInQueryResults),
 		"expected only one Release in query_results.json")
 
 	// Check that we get a real value for something we know should be there for every release.
@@ -299,5 +299,4 @@ func TestDisruptionDataFileParsing(t *testing.T) {
 	// so instead we'll make sure we didn't get a msg complaining about no match:
 	assert.Equal(t, "", msg, "BestMatchDuration reported a problem finding data for kube-api-new-connections aws amd64 ovn ha")
 	assert.NoError(t, err)
-
 }
--- a/pkg/monitortestlibrary/allowedbackenddisruption/query_results.json
+++ b/pkg/monitortestlibrary/allowedbackenddisruption/query_results.json
@@ -2,15 +2,32 @@
   {
     "BackendName": "cache-kube-api-new-connections",
     "Release": "4.14",
+    "FromRelease": "4.12",
+    "Platform": "aws",
+    "Architecture": "amd64",
+    "Network": "ovn",
+    "Topology": "ha",
+    "MasterNodesUpdated": "Y",
+    "JobRuns": 2,
+    "P95": "0.0",
+    "P99": "0.0",
+    "P75": "0.0",
+    "P50": "0.0"
+  },
+  {
+    "BackendName": "cache-kube-api-new-connections",
+    "Release": "4.14",
     "FromRelease": "4.13",
     "Platform": "aws",
     "Architecture": "amd64",
     "Network": "ovn",
     "Topology": "ha",
     "MasterNodesUpdated": "Y",
-    "JobRuns": 1000,
+    "JobRuns": 1062,
     "P95": "0.0",
-    "P99": "0.0"
+    "P99": "0.0",
+    "P75": "0.0",
+    "P50": "0.0"
   },
   {
     "BackendName": "cache-kube-api-new-connections",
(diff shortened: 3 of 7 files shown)


pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
```diff
@@ -274,16 +275,20 @@ func TestDisruptionDataFileParsing(t *testing.T) {
...
 	assert.Greater(t, dataOver100Runs, 5,
 		"expected at least 5 entries in query_results.json to have over 100 runs")
 	assert.True(t, foundAWSOVN, "no aws ovn job data in query_results.json")
 	assert.True(t, foundGCPOVN, "no gcp ovn job data in query_results.json")
 	assert.True(t, foundAzureOVN, "no azure ovn job data in query_results.json")
 	assert.True(t, foundMetalOVN, "no metal ovn job data in query_results.json")
-	assert.Equal(t, 1, len(releasesInQueryResults),
-		"expected only one Release in query_results.json")
-
+	if currentRelease <= "4.14" {
```
dgoodwin, line 285: Shouldn't really do comparisons like this on string representations of versions, 4.100 <= 4.14 for example, but in your case, I think != will suffice. If it's not 4.14 specifically, we should be good, as this would not run on any past releases.

dgoodwin: That said, I think this will immediately start failing when we switch to 5.15 if we haven't updated the data file, which could take up to a week.  That would suck.

Let's just assert <= 2 all the time and be done with it. We'll need to monitor the file ourselves. In future, put a TODO that this can be made to be Equals(2) for 4.15 and onward. Hopefully someone will see it someday.

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
```diff
@@ -301,3 +306,26 @@ func TestDisruptionDataFileParsing(t *testing.T) {
...
+		releaseSlice = append(releaseSlice, struct {
+			Key   string
+			Value bool
+		}{Key: key, Value: value})
+	}
+
+	// Sort the slice in descending order
+	sort.Slice(releaseSlice, func(i, j int) bool {
+		return releaseSlice[i].Key > releaseSlice[j].Key
+	})
```
dgoodwin, line 324: Sort isn't going to work if we ever make it to 4.100 :)  But given the above I don't think this function is needed anymore.

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
neisw, line 285: Have already discussed with @deepsm007 to change these to use integer comparison logic and not strings.  Can evaluate using [getMinor](https://github.com/openshift/origin/blob/a4bbe975a9361503a929761b969bf3504225d6ae/pkg/invariantlibrary/historicaldata/next_best_guess.go#L123https://github.com/openshift/origin/blob/a4bbe975a9361503a929761b969bf3504225d6ae/pkg/invariantlibrary/historicaldata/next_best_guess.go#L123) and [getMajor](https://github.com/openshift/origin/blob/a4bbe975a9361503a929761b969bf3504225d6ae/pkg/invariantlibrary/historicaldata/next_best_guess.go#L115C9-L115C9) from next_best_guess.go

dgoodwin (changes requested):

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
dgoodwin: Just saw Forrest's comment too late. 

There should be a utility out there somewhere for comparing version strings, if you want to keep this check. 

For 4.14, it should be 1, we could test that. For 4.15, it should be 2, but only after the first week or so. The safest thing to do, if you want to keep the test specific, is to enforce 1 for that as well, it will pass when we branch, and then the first time the job submits a PR to origin, we'll see it fail, and someone will have to intervene to update the disruption data.

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
neisw: Discussed a helper function that could be used here and above that converted the versions and used int comparisons.

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
neisw: For 4.14 we will actually get 4.13 data, just for non upgrades only since the query contains ` AND (MasterNodesUpdated != "N" OR FromRelease = "")` so the null MasterNodesUpdated records get dropped, except those with no FromRelease.

I'm wondering if we shouldn't manually update the file (run the ci-tools command locally and copy it over) the first time when we are ready and update the test at that time.

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
```diff
@@ -274,13 +275,14 @@ func TestDisruptionDataFileParsing(t *testing.T) {
...
 	}
-
+	currentRelease = currentReleaseFromMap(releasesInQueryResults)
 	assert.Greater(t, dataOver100Runs, 5,
 		"expected at least 5 entries in query_results.json to have over 100 runs")
 	assert.True(t, foundAWSOVN, "no aws ovn job data in query_results.json")
 	assert.True(t, foundGCPOVN, "no gcp ovn job data in query_results.json")
 	assert.True(t, foundAzureOVN, "no azure ovn job data in query_results.json")
 	assert.True(t, foundMetalOVN, "no metal ovn job data in query_results.json")
+	// TO-DO update the assert to 2 when 4.15 branches, releaseInQueryResults will contain 2 releases after 4.15
```
neisw, line 285: We will actually see the two releases in 4.14, we can look into a manual update of the file in origin when the ci-tools code merges.

dgoodwin:
/approve

Will leave lgtm to Forrest once you guys are happy with how this will land.

neisw:
We still have to work out how it looks at previous release data, current code path will switch from minor upgrade to micro first.  We can discuss needs but seems with this addition if we don't have enough runs we should make an explicit check for the previous release.

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
neisw: `return historicaldata.GetMajor(releaseSlice[i].Key) > historicaldata.GetMinor(releaseSlice[j].Key)` doesn't look right to me.  We are comparing the Major of i to the Minor of j?  Seems we need to compare both Majors first the Minors.

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
deepsm007: we won't see 2 releases in 4.14 as 4.13 doesn't have "MASTERNODEUPDATED" column, until we have 4.15 data we should be fine with assert checking for 1 release

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
deepsm007: Added a helper function instead of GetMajor, GetMinor

pkg/invariantlibrary/allowedbackenddisruption/matches_test.go
neisw: We will see 4.13 data for the non upgrade jobs.  You can verify by running the ci-tools update locally.

pkg/invariantlibrary/historicaldata/alert_types.go
```diff
@@ -79,9 +80,15 @@ func NewAlertMatcherWithHistoricalData(data map[AlertDataKey]AlertStatisticalDat
...
+	logrus.Infof("historicalData has %d entries", len(b.HistoricalData))
 
-	if percentiles, ok := b.HistoricalData[exactMatchKey]; ok && percentiles.JobRuns >= minJobRuns {
-		return percentiles, "", nil
+	if percentiles, ok := b.HistoricalData[exactMatchKey]; ok {
+		if percentiles.JobRuns > minJobRuns {
+			logrus.Infof("found exact match: %+v", percentiles)
+			return percentiles, "", nil
+		}
+		PreviousReleaseUpgrade(key.JobType)
```
neisw, line 91: Reviewed with Deep, we aren't actually doing anything with the updated matcher here.  Need to consolidate the logic in the loop below for reusability and check it check for results here as well.  Same for disruption.

pkg/invariantlibrary/historicaldata/next_best_guess.go
```diff
@@ -189,3 +191,55 @@ func combine(nextBestKeys ...NextBestKey) NextBestKey {
...
+	}
+	return firstKey
+}
+
+func compareReleaseString(one, two string) bool {
+	major1, err := strconv.ParseInt(strings.Split(one, ".")[0], 10, 32)
+	if err != nil {
+		panic(err)
+	}
+	major2, err := strconv.ParseInt(strings.Split(two, ".")[1], 10, 32)
```
neisw, line 224: Need a fix here and some tests for CurrentReleaseFromMap.

deepsm007:
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-sdn-upgrade 10

PR #28152, opened 2023-08-08
be sure the exit code fails when the invariant tests fail

deads2k:


deads2k:
/hold

doesn't work properly with flakes.

deads2k:
no conflict, merging to get exit codes right.

PR #28154, opened 2023-08-08
Bug OCPBUGS-17483: Skip invariant load balancer test for OpenStack

MaysaMacedo:
The disruption Service tests that were previously run and skipped for OpenStack on upgrades jobs, are now run on various jobs unrelated to upgrades. This can cause the job run to fail when proxy installations are run with OpenStack given floating ip is not supported in this scenario. This commit fixes the issue by continue to skips those tests for OpenStack.


MaysaMacedo:
Note that previously it was skipped with a different reason:
`Jul 31 21:27:58.108: INFO: Skipping: [sig-network-edge] Application behind service load balancer with PDB remains available using reused connections: No historical data to calculate allowedDisruption`
But in any case, we want to continue to skip it.

PR #28161, opened 2023-08-10
OCPBUGS-17477: requestheader: wait for only 3 oauth-servers to be available

stlaz:
It seems that some of the old servers might be picking new requests even though they're shutting down. This commit should either prove or disprove that theory.

Another possibility is that the authn operator is not reporting the deployment status properly, or a revision rollout is borked.


stlaz:
/test e2e-aws-ovn-serial
infra :neutral_face:

ibihim (commented):
/lgtm

PR #28163, opened 2023-08-10
Add some tests for event interval generation

dgoodwin:
I've been trying to port these to the new interval builder, and noticed we don't really have enough test coverage to know if I broke something or not. This PR just adds some more tests for the main paths through.

Also fixes a bug I think we recently gained during the invariants refactor, it looks like we did not add the new interval we created back to the list when post-processing intervals that eventually crossed the pathological/true barrier.

### DIFF
--- a/pkg/monitor/event_test.go
+++ b/pkg/monitor/event_test.go
@@ -2,85 +2,18 @@ package monitor
 
 import (
 	"context"
-	"encoding/json"
-	"errors"
-	"fmt"
-	"os"
 	"regexp"
 	"testing"
 	"time"
 
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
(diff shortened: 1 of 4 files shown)


pkg/monitor/event_test.go
```diff
@@ -2,142 +2,133 @@ package monitor
...
-				Count:         2,
-				Reason:        "NodeHasNoDiskPressure",
-				Message:       "sample message",
-				LastTimestamp: metav1.Now(),
-			},
-		},
-	}
-	if len(kubeEvents.Items) == 0 {
-		kubeEvents.Items = append(kubeEvents.Items, smallKubeEvents.Items...)
-	}
```
dgoodwin, line 83: Killed this off, too weird to maintain a "test" that is skipped, and requires a live cluster and a file on disk.

PR #28169, opened 2023-08-11
start the watch-endpoint-slice command

deads2k:
Adds a starting point for a command `./openshift-tests disruption watch-endpoint-slice --output-file=disruption.jsonl --disruption-backend-prefix=pod-to-pod- --disruption-target-service-name=pod-network-service` that will set up disruption.

It needs refinement for naming and likely shutdown, but these are the big steps


pkg/cmd/openshift-tests/disruption/watch-endpointslice/endpointslice_controller.go
```diff
@@ -0,0 +1,272 @@
...
+	endpointSlice, err := c.endpointSliceLister.EndpointSlices(namespace).Get(name)
+	if errors.IsNotFound(err) {
+		c.removeAllWatchers()
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	watchersForCurrEndpoints := map[string]*watcher{}
```
dgoodwin, line 98: godoc

pkg/cmd/openshift-tests/disruption/watch-endpointslice/endpointslice_controller.go
```diff
@@ -0,0 +1,272 @@
...
+			if endpoint.NodeName != nil {
+				watchersForCurrEndpoints[address].nodeName = *endpoint.NodeName
+			}
+		}
+	}
+
+	c.watcherLock.Lock()
+	defer c.watcherLock.Unlock()
+
+	watchersToDelete := map[string]*watcher{}
```
dgoodwin, line 121: Lets add a comment here that this is to clean out any watchers we were holding onto that have disappeared in the current state of the EndpointSlice

pkg/cmd/openshift-tests/disruption/watch-endpointslice/endpointslice_controller.go
```diff
@@ -0,0 +1,272 @@
...
+		if _, ok := c.watchers[watcherKey]; ok {
+			continue
+		}
+		newWatcher := watchersForCurrEndpoints[watcherKey]
+		url := fmt.Sprintf("http://%s", net.JoinHostPort(newWatcher.address, newWatcher.port))
+		fmt.Fprintf(c.outFile, "Adding and starting: %v on node/%v\n", url, newWatcher.nodeName)
+
+		newWatcher.newConnectionSampler = backenddisruption.NewSimpleBackend(
+			url,
+			fmt.Sprintf("%s-new-connection-node-%v-endpoint-%v", c.backendPrefix, newWatcher.nodeName, newWatcher.address),
```
dgoodwin, line 145: Double dash here possible.

dgoodwin: mixed %s and %v a little confusing, can probably go to all v

PR #28174, opened 2023-08-11
run start and collect in parallel

deads2k:
it's painful on the CLI without this because we have several waits.  Yes the waits shoudl be cleaned up, but htis helps easily.


deads2k:
no conflicts, merging.

PR #28176, opened 2023-08-11
Invariant 40 clean stop

deads2k:


pkg/cmd/openshift-tests/disruption/watch-endpointslice/endpointslice_controller.go
```diff
@@ -8,32 +8,37 @@ import (
...
-	recorder      monitorapi.RecorderWriter
-	outFile       io.Writer
-
-	endpointSliceLister  discoverylisters.EndpointSliceLister
-	endpointSlicesSynced cache.InformerSynced
+	backendPrefix     string
+	namespaceName     string
+	serviceName       string
+	myNodeName        string
+	stopConfigMapName string
```
dgoodwin, line 35: Godoc plz, the CLI flags covers it pretty good but that's a ways from here, would be good to have it on the final stopping point. Clarify that in normal operation the configmap is created from the origin monitor test.

pkg/invariants/network/disruptionpodnetwork/pod-network-poller-deployment.yaml
```diff
@@ -48,3 +60,7 @@ spec:
                 matchLabels:
                   network.openshift.io/disruption-target: pod-network
                   network.openshift.io/disruption-actor: poller
+      volumes:
+        - hostPath:
+            path: /var/log/kube-apiserver
```
dgoodwin, line 65: Is this right? Could use a comment if so as that looks very out of place.

pkg/invariants/network/disruptionpodnetwork/invariant.go
```diff
@@ -153,8 +168,12 @@ func (pna *podNetworkAvalibility) CollectData(ctx context.Context, storageDir st
 			}
 		}
 	}
+	junits = append(junits, &junitapi.JUnitTestCase{
+		Name:      "[sig-network] poller pod logs",
```
dgoodwin, line 172: Test name looks a little generic, and a good thing to get right the first time it goes in.

"can collect pod-to-pod poller pod logs"?

pkg/invariants/network/disruptionpodnetwork/invariant.go
```diff
@@ -115,6 +113,19 @@ func (pna *podNetworkAvalibility) StartCollection(ctx context.Context, adminREST
 }
 
 func (pna *podNetworkAvalibility) CollectData(ctx context.Context, storageDir string, beginning, end time.Time) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error) {
+	// create the stop collecting configmap and wait for 30s to thing to have stopped.  the 30s is just a guess
+	if _, err := pna.kubeClient.CoreV1().ConfigMaps(pna.namespaceName).Create(ctx, &corev1.ConfigMap{
```
dgoodwin, line 117: Logging with timestamp.

pkg/invariants/network/disruptionpodnetwork/invariant.go
dgoodwin: Shouldn't this test fail if you hit an error?

pkg/invariants/network/disruptionpodnetwork/invariant.go
```diff
@@ -90,13 +92,9 @@ func (pna *podNetworkAvalibility) StartCollection(ctx context.Context, adminREST
...
-	// disable the poller until can get the image sorted out.
-	// TODO re-enable
-	if false {
-		podNetworkPollerDeployment.Spec.Replicas = &numNodes
-		if _, err = pna.kubeClient.AppsV1().Deployments(pna.namespaceName).Create(context.Background(), podNetworkPollerDeployment, metav1.CreateOptions{}); err != nil {
-			return err
-		}
+	podNetworkPollerDeployment.Spec.Replicas = &numNodes
+	if _, err = pna.kubeClient.AppsV1().Deployments(pna.namespaceName).Create(context.Background(), podNetworkPollerDeployment, metav1.CreateOptions{}); err != nil {
+		return err
```
dgoodwin, line 97: How is this going to behave on single-node/microshift etc? I'm assuming we'll just get the hairpin results as there's nothing else to talk to?

pkg/invariants/network/disruptionpodnetwork/invariant.go
dgoodwin: I know you prefer to just merge an then sort these out, but technically it would be nice to keep this test out of all the databases if it's short lived, suggest removing the junit if you want to instamerge and then bring it back when it's named in final form.

dgoodwin:
Looks like a legit problem with the deads2k image being used, it's not actually getting replaced: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/28176/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1691187500984307712

deads2k:
superceded

PR #28179, opened 2023-08-14
basic setup of podNetwork disruption tests

deads2k:
adding the skeleton of a new disruption test.


pkg/invariants/network/disruptionpodnetwork/pod-network-target-deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+        network.openshift.io/disruption-target: pod-network
+        network.openshift.io/disruption-actor: target
+    spec:
+      containers:
+        - command:
+            - /agnhost
+            - netexec
+            - --http-port=8080
+            - --delay-shutdown=30
+          image: registry.k8s.io/e2e-test-images/agnhost:2.43
```
JacobTanenbaum, line 29: Should it be mentioned that the image gets overridden as we do above for replicas?

pkg/defaultinvariants/types.go
```diff
@@ -70,6 +71,7 @@ func newDefaultInvariants() invariants.InvariantRegistry {
 
 	invariantTests.AddInvariantOrDie("apiserver-availability", "kube-apiserver", disruptionlegacyapiservers.NewAvailabilityInvariant())
 
+	invariantTests.AddInvariantOrDie(disruptionpodnetwork.InvariantName, "Network / ovn-kubernetes", disruptionpodnetwork.NewPodNetworkAvalibilityInvariant())
```
JacobTanenbaum, line 74: Nit: you use the const for the invariant name but not for the JIRA organizations name. Should it be consistent?

pkg/invariants/network/disruptionpodnetwork/invariant.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// our pods tolerate masters, so create one for each of them.
+	nodes, err := pna.kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	numNodes := int32(len(nodes.Items))
+
+	// disable the poller until can get the image sorted out.
```
JacobTanenbaum, line 100: Nit: maybe tag with a `TODO:` so it has less a chance of being lost?

deads2k: > Nit: maybe tag with a `TODO:` so it has less a chance of being lost?

I personally guarantee I won't lose it :)

pkg/invariants/network/disruptionpodnetwork/pod-network-target-deployment.yaml
JacobTanenbaum: Yeah do a followup

pkg/invariants/network/disruptionpodnetwork/invariant.go
deads2k: added

deads2k:
was lgtmed, simple rebase, tests pass, no conflict, merging.

PR #28182, opened 2023-08-14
Fix possibility of intervals without from/to by requiring when we Build()

dgoodwin:
This should help anyone forgetting to call From/To and generating an
interval without.
Using Build instead of the constructor NewInterval because there are cases where we don't really know it up front.


neisw:
Missed this earlier:

github.com/openshift/origin/pkg/monitortests/node/kubeletlogcollector
pkg/monitortests/node/kubeletlogcollector/node.go:196:4: monitorapi.NewInterval(monitorapi.SourceNetworkManagerLog, monitorapi.Warning).Locator(nodeLocator).Message(monitorapi.NewMessage().HumanMessage(message)).From undefined (type *monitorapi.IntervalBuilder has no field or method From)

deads2k:
This PR needs to be reverted.  Open time intervals for start and end are perfectly valid when you consider cases of "this pod was started before the monitor began" or "this node was still ready when this monitor ended".  Doing this limits useful expression from where we are today.  This is why the .Clamp method exists if we wish to eliminate open endedness for rendering, but doing it here will prevent us from expressing current and valid intent.

deads2k:
Ah, the PR doesn't do what the description says it.  Do you intend to stop zero value times?

PR #28187, opened 2023-08-15
Full story with the correct image.

deads2k:


deads2k:
ok, so details here confirm that the image in the payload under test is the current master, not the master in this build.  Fun.

PR #28189, opened 2023-08-15
add host-to-pod disruption checking

deads2k:
Cannot merge until the image is fully available for micro upgrades.


deads2k:
I wonder if host-to-host worked before IC.  I'll try on 4.13 tomorrow.

PR #28194, opened 2023-08-16
select poller image from payload

deads2k:


pkg/monitortests/network/disruptionpodnetwork/monitortest.go
```diff
@@ -60,15 +64,42 @@ func init() {
...
+		}
+		clusterVersion, err := configClient.ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		pna.payloadImagePullSpec = clusterVersion.Status.History[0].Image
+	}
+
+	// runImageExtract extracts src from specified image to dst
+	cmd := exec.Command("oc", "adm", "release", "info", pna.payloadImagePullSpec, "--image-for=tests")
```
deads2k, line 92: this is the meat.  not pretty, but I think it works

PR #28197, opened 2023-08-16
tolerate cluster without clusterversion

deads2k:


neisw:
/lgtm

reviewed with david

PR #28198, opened 2023-08-16
fail if no sampler output is found for the poller

deads2k:
builds on https://github.com/openshift/origin/pull/28189

### DIFF
--- a/pkg/monitortests/network/disruptionpodnetwork/host-network-to-host-network-poller-deployment.yaml
+++ b/pkg/monitortests/network/disruptionpodnetwork/host-network-to-host-network-poller-deployment.yaml
@@ -1,0 +1,67 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: host-network-to-host-network-disruption-poller
+spec:
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 34%
+      maxSurge: 0
+  # to be overridden by the number of nodes
+  replicas: 1
+  selector:
+    matchLabels:
+      network.openshift.io/disruption-target: host-to-host
+      network.openshift.io/disruption-actor: poller
+  template:
+    metadata:
+      labels:
+        network.openshift.io/disruption-target: host-to-host
+        network.openshift.io/disruption-actor: poller
+    spec:
+      containers:
+        - command:
+            - /usr/bin/openshift-tests
+            - disruption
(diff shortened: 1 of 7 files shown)


pkg/monitortests/network/disruptionpodnetwork/host-network-to-host-network-poller-deployment.yaml
```diff
@@ -0,0 +1,67 @@
...
+        - command:
+            - /usr/bin/openshift-tests
+            - disruption
+            - watch-endpoint-slice
+            - --output-file=/var/log/persistent-logs/disruption-host-to-host.jsonl
+            - --disruption-backend-prefix=host-to-host
+            - --disruption-target-service-name=host-network-service
+            - --stop-configmap=stop-collecting
+            - --my-node-name=$(MY_NODE_NAME)
+          image: docker.io/deads2k/openshift-tests:latest
```
neisw, line 33: Nit: I know these get replaced but wondering if there is value documenting that here like you did in other prs with something like `docker.io/to-be-replaced/openshift-tests:latest`

pkg/monitortests/network/disruptionpodnetwork/monitortest.go
```diff
@@ -142,6 +173,16 @@ func (pna *podNetworkAvalibility) StartCollection(ctx context.Context, adminREST
 		return err
 	}
```
neisw, line 175: Does the order of execution matter at all, any value to deploying the target and service before the pollers?

deads2k: > Does the order of execution matter at all, any value to deploying the target and service before the pollers?

It does not.  The targets don't look at anything.  The pollers use the information about what is live to decide what to sample.

pkg/monitortests/network/disruptionpodnetwork/host-network-to-host-network-poller-deployment.yaml
deads2k: > Nit: I know these get replaced but wondering if there is value documenting that here like you did in other prs with something like `docker.io/to-be-replaced/openshift-tests:latest`

Promised as an IOU

deads2k:
transitive lgtm via https://github.com/openshift/origin/pull/28191

deads2k:
no conflict with the PR that merged

PR #28165, opened 2023-08-11
Fixes openshift-tests dry-run failure asking for KUBERNETES_MASTER

Rajalakshmi-Girish:
Fixes https://github.com/openshift/origin/issues/28159

PR #28199, opened 2023-08-17
Fixes openshift-tests run-upgrade dry-run failure asking for KUBERNETES_MASTER

Rajalakshmi-Girish:
Similar to https://github.com/openshift/origin/pull/28165
This change is for `run-upgrade`

### DIFF
--- a/pkg/cmd/openshift-tests/run-upgrade/flags.go
+++ b/pkg/cmd/openshift-tests/run-upgrade/flags.go
@@ -10,6 +10,7 @@ import (
 	testginkgo "github.com/openshift/origin/pkg/test/ginkgo"
 	"github.com/spf13/pflag"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/rest"
 )
 
 // TODO collapse this with cmd_runsuite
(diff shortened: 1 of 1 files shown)


Rajalakshmi-Girish:
@vrutkovs I missed to change for `run-upgrade`.
Can you PTAL?

Rajalakshmi-Girish:
> The change looks fine, but I don't think `run-upgrade` is meant to be run in dry-mode. What's the expected result here?

The `--dry-run` flag is valid for `run-upgrade` too.
```
[root@maxwells1 origin]# ./openshift-tests run-upgrade --help
Run an upgrade test suite against an OpenShift server

 This command will run one of the following suites against a cluster identified by the current KUBECONFIG file. See the suite description for more on what actions the suite will take.

 If you specify the --dry-run argument, the actions the suite will take will be printed to the output.
---------
---------
```

vrutkovs:
It is, but what's the output you're expecting? `run-upgrade` is a single test effectively, it cannot be separated into a smaller ones. I don't think dry run makes any sense for it, but since its a global flag it would certainly apply

Rajalakshmi-Girish:
> It is, but what's the output you're expecting? `run-upgrade` is a single test effectively, it cannot be separated into a smaller ones. I don't think dry run makes any sense for it, but since its a global flag it would certainly apply

From the `./openshift-tests run-upgrade --help` command output, I see three suites available under this:
```
Available upgrade suites:

all
  Run all tests.

platform
  Run only the tests that verify the platform remains available.

none
  Don't run disruption tests.
```

We have been running below commands in our internal CI job to verify the built openshift-tests binary:
```
openshift-tests run kubernetes/conformance --dry-run
openshift-tests run-upgrade --dry-run --to-image quay.io/openshift/community-e2e-images:e2e-3-k8s-gcr-io-e2e-test-images-sample-apiserver-1-17-4-AC3A0x1ePf4fLoiv all
```
Please find the attached file for the dry-run output that we get in our CI job.
[run-upgrade-dry-run.txt](https://github.com/openshift/origin/files/12375323/run-upgrade-dry-run.txt)

Rajalakshmi-Girish:
@vrutkovs Please check the above reply.

vrutkovs:
>I see three suites available under this

Right, this starts upgrade and optionally runs some monitor tests in the background - but its pointless to run just one monitor tests separately, even if dry run would output its name.

I agree with the code change here, but I don't think it would produce anything useful - monitors running during uprgade are useless unless the upgrade is running in the background

Rajalakshmi-Girish:
> > I see three suites available under this
> 
> Right, this starts upgrade and optionally runs some monitor tests in the background - but its pointless to run just one monitor tests separately, even if dry run would output its name.
> 
> I agree with the code change here, but I don't think it would produce anything useful - monitors running during uprgade are useless unless the upgrade is running in the background

@vrutkovs do you think disabling the `--dry-run` flag for `run-upgrade` is the right thing to do?

Rajalakshmi-Girish:
We have decided to remove the `openshift-tests run-upgrade --dry-run` command from our Job.
Please close this PR if you feel it is irrelevant.

PR #28097, opened 2023-07-26
STOR-1425: Update to Kubernetes 1.28

bertinatto:

PR #28200, opened 2023-08-17
STOR-1425: Add images coming in k8s 1.28

soltysh:
Some of the metal jobs rely on information which images need to be mirrored, this information is coming in https://github.com/openshift/origin/pull/28097/ but to properly assess correctness we need those information sooner. 

This PR injects that image information directly into `./openshift-tests images` command.

/hold
Until master branch opens for OCP 4.15.

/assign @bertinatto


bertinatto (commented):
/lgtm

PR #28201, opened 2023-08-17
move graceful shutdown to monitor test

deads2k:
This is the last remaining piece in master that is living in the monitor directly and needs to move.

/assign @dgoodwin


pkg/monitortests/kubeapiserver/apiservergracefulrestart/monitortest.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	// and now close everything still open with a warning
+	for fakeLocator, startTime := range startedIntervals {
+		podRef := monitorapi.PodFrom(fakeLocator)
+		nodeName, _ := monitorapi.NodeFromLocator(fakeLocator)
+
+		computedIntervals = append(computedIntervals,
+			monitorapi.NewInterval(monitorapi.SourceKubeEvent, monitorapi.Error).
```
dgoodwin, line 110: While I'm totally unsure if Source concept will survive once we start migrating the js UI to use these, I'd expect your Constructed("graceful-shutdown-analyzer") to instead be a Source here. I was thinking source is an indicator for what code produced the Interval and a tool to clearly identify what gets grouped into a section in the UI.

pkg/monitortests/testframework/watchevents/monitortest.go
```diff
@@ -0,0 +1,55 @@
...
+		return err
+	}
+
+	startEventMonitoring(ctx, recorder, adminRESTConfig, kubeClient)
+
+	return nil
+}
+
+func (w *eventWatcher) CollectData(ctx context.Context, storageDir string, beginning, end time.Time) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error) {
+	// because we are sharing a recorder that we're streaming into, we don't need to have a separate data collection step.
```
dgoodwin, line 34: I'm curious if you're certain we need three ways a monitortest can inject intervals, CollectData, ConstructComputedIntervals (both of which make sense), but why do we need live recording as well? On the surface it feels like that could be in CollectData? 

Depending on the rationale it might be helpful to spell it out in the monitortest README when to use each.

dgoodwin (commented):
This looks mostly fine but I'm a little short on time. I have a suggestion on use of Source that you'll probably object to and want to merge anyhow so:

/lgtm
/hold


Feel free to release when you're happy, I'd like to see the job output and make sure spyglass is healthy though I suspect it was broken to begin with, we had some weird graphs and I don't recall them looking like that prior to the refactor. Perhaps Abu can confirm we've got them operational properly.

PTO next week so presumably someone else may have to help get this in fully if it isn't coming today.

pkg/monitortests/testframework/watchevents/monitortest.go
deads2k: There isn't any ordering, so the gating is required to have type A CollectData for Calculators 1 and 2

Live recording handles th e watches.

pkg/monitortests/kubeapiserver/apiservergracefulrestart/monitortest.go
deads2k: Could you generally describe what a source is?  it is a MonitorTest?

deads2k:
/hold

This needs more checking.

deads2k:
forgot to return the computed intervals.  Fixed.

pkg/monitor/monitorapi/identification_pod.go
```diff
@@ -29,9 +29,9 @@ func NonUniquePodLocatorFrom(locator string) string {
 func PodFrom(locator string) PodReference {
 	parts := LocatorParts(locator)
 	namespace := NamespaceFrom(parts)
-	name := parts["pod"]
-	uid := parts["uid"]
-	if len(namespace) == 0 || len(name) == 0 || len(uid) == 0 {
+	name := parts[string(LocatorPodKey)]
+	uid := parts[string(LocatorUIDKey)]
+	if len(namespace) == 0 || len(name) == 0 {
```
vrutkovs, line 34: uid is allowed to be empty?

deads2k: > uid is allowed to be empty?

Yeah, the events being sent from the apiservers lack it for instance, but you still want to associate with pods.  I'd be happy to see the apiserver code updated to include uid.

PR #28204, opened 2023-08-19
ignore stale condition challenge events

sanchezl:
Proposed https://github.com/openshift/cluster-kube-apiserver-operator/pull/1423:
1. Prefixes condition messages with a zero-width non-breaking space (ZWNBSP).
2. Expects operators to quickly (within 1 min) update the condition message, removing the ZWNBSP.

Many operators emit an event when they are updating conditions (thanks to common code in libary-go) , these will trigger the pathological events tests.

This PR configured the pathological events tests to ignore these "challenge response" events.

### DIFF
--- a/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
+++ b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
@@ -148,6 +148,9 @@ var AllowedRepeatedEventPatterns = []*regexp.Regexp{
 
 	// Separated out in testMarketplaceStartupProbeFailure
 	regexp.MustCompile(MarketplaceStartupProbeFailureRegExpStr),
+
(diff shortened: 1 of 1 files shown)


pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -148,6 +148,9 @@ var AllowedRepeatedEventPatterns = []*regexp.Regexp{
 
 	// Separated out in testMarketplaceStartupProbeFailure
 	regexp.MustCompile(MarketplaceStartupProbeFailureRegExpStr),
+
+	// stale condition challenge
+	regexp.MustCompile(`Requesting that the operator update this condition`),
```
deads2k, line 153: oh, I made my code emit this event, didn't I?

We should probably stop that.

PR #28209, opened 2023-08-21
fix *-to-host sampler failure calculation

deads2k:
This fixes teh current pattern which doesn't cause failure, but does count disruption improperly: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/28191/pull-ci-openshift-origin-master-e2e-gcp-ovn-rt-upgrade/1692509054028484608

### DIFF
--- a/pkg/monitor/backenddisruption/disruption_backend_sampler.go
+++ b/pkg/monitor/backenddisruption/disruption_backend_sampler.go
@@ -53,6 +53,8 @@ type BackendSampler struct {
 	// tlsConfig holds the CA bundle for verifying the server and client cert/key pair for identifying to the server.
 	tlsConfig *tls.Config
 
+	// expectedStatusCode allows status codes other than 200-399.
+	expectedStatusCode int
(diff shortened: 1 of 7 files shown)


deads2k:
install infra

/retest-required

pkg/monitor/backenddisruption/disruption_backend_sampler.go
```diff
@@ -190,6 +192,12 @@ func (b *BackendSampler) WithTLSConfig(tlsConfig *tls.Config) *BackendSampler {
 	return b
 }
 
+// WithTLSConfig sets both the CA bundle for trusting the server and the client cert/key pair for identifying to the server
+func (b *BackendSampler) WithExpectedStatusCode(statusCode int) *BackendSampler {
```
DennisPeriquet, line 196: comment not matching function name

deads2k:
lgtm'd via slack.

pkg/monitor/backenddisruption/disruption_backend_sampler.go
deads2k: > comment not matching function name

I promise to fix it in a followup.

PR #28211, opened 2023-08-21
allow failures for tests intentionally producing failures

deads2k:
I found tests in https://github.com/openshift/origin/blob/master/test/extended/networking/tuning.go that are trying to cause permission failures.  If it should result in pod sandbox creation failures, this is good to merge. If it should fail *prior* to pod sandbox creation, then that needs to be checked.


s1061123 (approved):
/lgtm

PR #28214, opened 2023-08-22
Do not fail etcd storage path test for GVRs that are not served.

benluddy:
Under the assumption that this test is of the behavior "apis are stored at the expected etcd paths," it can't make a positive or negative determination unless a given resource is served. Allowing it to pass in that case relieves the test from needing to know the complete set of served GVRs.


benluddy:
/payload-job periodic-ci-openshift-release-master-ci-4.14-e2e-gcp-sdn-techpreview-serial

dinhxuanvu (commented):
/lgtm

soltysh (approved):
/approve

PR #28215, opened 2023-08-23
make it possible to select which monitor tests to run

deads2k:
This is convenient when testing one from the CLI.  It's also pretty easy to accept and run just CollectData and SerializeResults if we want.

### DIFF
--- a/pkg/cmd/openshift-tests/monitor/run/run_monitor_command.go
+++ b/pkg/cmd/openshift-tests/monitor/run/run_monitor_command.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
@@ -25,8 +26,10 @@ import (
 )
 
 type RunMonitorFlags struct {
-	ArtifactDir    string
-	DisplayFromNow bool
+	ArtifactDir         string
+	DisplayFromNow      bool
+	ExactMonitorTests   []string
+	DisableMonitorTests []string
 
 	genericclioptions.IOStreams
 }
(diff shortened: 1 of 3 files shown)


pkg/cmd/openshift-tests/monitor/run/run_monitor_command.go
```diff
@@ -77,6 +79,8 @@ func newRunCommand(name string, streams genericclioptions.IOStreams) *cobra.Comm
 func (f *RunMonitorFlags) BindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&f.ArtifactDir, "artifact-dir", f.ArtifactDir, "The directory where monitor events will be stored.")
 	flags.BoolVar(&f.DisplayFromNow, "display-from-now", f.DisplayFromNow, "Only display intervals from at or after this comand was started.")
+	flags.StringSliceVar(&f.ExactMonitorTests, "monitor", f.ExactMonitorTests, "list of exactly which monitors to enable. All others will be disabled.")
+	flags.StringSliceVar(&f.DisableMonitorTests, "disable-monitor", f.DisableMonitorTests, "list of monitors to disable.  Defaults for others will be honored.")
```
dgoodwin, line 83: It is not clear what monitors are available without going straight to code. Would list-monitors be an acceptable cli flag?

dgoodwin (commented):
Otherwise looks fine to me.

deads2k:
added to the standard help listing where you'd need to use it.

dgoodwin:
/lgtm
/approve

Filed https://issues.redhat.com/browse/TRT-1269 to make use of

PR #28220, opened 2023-08-25
OCPVE-295: fix: add rteval to the test image

jeff-roche:
Adding the `rteval` binary to the test image to be used for realtime kernel testing. The rteval binary is a utility for measuring various aspects of realtime behavior on a system under load.


jeff-roche:
/assign @sdodson 

This is for kernel rt 4.14 work

jeff-roche:
@sdodson now that branch cut has happened are we still blocked on merging this?

sdodson:
This needs to make it into 4.14 soon, so that we have more comprehensive testing there. While this doesn't fix anything in 4.15 we'd rather not delay the backport.

PR #28226, opened 2023-08-30
AUTH-365: add PodSecurityViolation to watched alerts

s-urbaniak:
We don't want this to be firing in vanilla OpenShift clusters.

Supersedes #27634


stlaz:
/retest-required
installation/infra issues, it seems

PR #28229, opened 2023-08-30
OCPBUGS-5837: add possibility to skip invariant tests

bertinatto:
This is specially useful for the CSI test suite. Partners typically certify their CSI drivers using that test suite, which usually reports
invariant failures.

This is a replacement for https://github.com/openshift/origin/pull/27471, which I couldn't reopen.

CC @openshift/storage


bertinatto:
/hold
for 4.15 branch

PR #28239, opened 2023-09-04
OCPBUGS-18490: bump monitoring SNO bounds

jan--f:


bparees:
no QE involvement needed so labeling CPA and BRA.

PR #28240, opened 2023-09-04
NO-JIRA: Fail on APIs removed in the next release

soltysh:
Brings back https://github.com/openshift/origin/pull/27561, which was reverted in https://github.com/openshift/origin/pull/28227, tracked by [TRT-1209](https://issues.redhat.com//browse/TRT-1209).

The PR consists of 3 commits:
1. Bring back #27561 
2. Move all `ClusterVersion` related utils into a single helper file
3. Modify "[sig-arch][Late] clients should not use APIs that are removed in upcoming releases [apigroup:apiserver.openshift.io] [Suite:openshift/conformance/parallel]"` test such that it will check `ClusterVersion`, for clusters 4.14 and newer it's a hard fail, for older and microshift it will continue to flake.


soltysh:
/payload-job periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-aws-ovn-upgrade

which specifically failed last time, as mentioned in https://issues.redhat.com//browse/TRT-1209

PR #28243, opened 2023-09-05
add pod log streaming to monitor for etcd so we see all intervals

deads2k:
ought to handle

1. pod creation
2. pod deletion
3. pod restart
4. pod replacement (static pods)
5. kubelet restart
6. node restart


pkg/monitortestlibrary/podaccess/pod_logs_stream_all.go
```diff
@@ -0,0 +1,252 @@
...
+			defer wg.Done()
+			watcherToDelete.podStreamer.Stop(ctx)
+		}(ctx, watcherToDelete)
+	}
+	wg.Wait()
+
+	c.watchers = map[podKey]*watcher{}
+}
+
+// Run starts the controller and blocks until stopCh is closed.
```
hasbro17, line 212: Nit: I realize this is probably boilerplate but `stopCh` here refers to ctx>Done() right? Thought it was finishedCleanup but that's just the signal for a clean shutdown I think.

pkg/monitortestlibrary/podaccess/pod_logs_stream_all.go
```diff
@@ -0,0 +1,252 @@
...
+	logger := klog.FromContext(ctx)
+	logger.Info("Starting PodsLogStreamer")
+	defer logger.Info("Shutting down PodsLogStreamer controller")
+
+	go wait.UntilWithContext(ctx, c.runWorker, time.Second)
+
+	<-ctx.Done()
+
+	// TODO set a timeout?
+	c.removeAllWatchers(context.TODO())
```
hasbro17, line 227: 10 seconds? Should be enough for all the pods in the etcd namespace?
```
ctx, cancel := context.WithTimeout(ctx, time.Duration(time.Second*10))
defer cancel()
c.removeAllWatchers(ctx)
```

Although the cluster is going away so it may not matter if we don't finish cleanup.

pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
```diff
@@ -52,3 +154,183 @@ func (*etcdLogAnalyzer) Cleanup(ctx context.Context) error {
...
+				Message(message).
+				Build(logLine.Instant, logLine.Instant.Add(time.Second)),
+		)
+	}
+
+}
+
+var (
+	// "raft.node: 38360899e3c7337e elected leader d8a2c1adbed17efe at term 6"
+	electedLeaderRegex = regexp.MustCompile("elected leader (?P<CURR_LEADER>[a-z0-9.-]+) at term (?P<TERM>[0-9]+)")
```
hasbro17, line 284: TIL named capture groups.

hasbro17 (approved):
/lgtm

Still grasping the interval builder bits but the log streaming and matching looks good.

deads2k:
simple rebase, reapplying lgtm

PR #28245, opened 2023-09-06
Bug OCPBUGS-18718 Remove duplicate connection type from disruption name

xueqzhan:
[TRT-1219](https://issues.redhat.com/browse/TRT-1219)
[OCPBUGS-18718](https://issues.redhat.com/browse/OCPBUGS-18718)

With the refactor, we are adding connection types twice to the backend name. This PR removes the duplicate.


dgoodwin:
/approve
/lgtm

Good find Ken, thank you. Looks like no bad data in db it was purely on the lookup side.

PR #28246, opened 2023-09-07
[WIP] Refactor `test/extended/util/client.go`

soltysh:
This PR does the following things at a high level:
- cleans up `test/extended/util/client.go` to get rid of duplicate code and ensure all creation goes through the exact same path - single `NewCLI` method with a structure that allows you to configure the necessary variant
- keep explicit separation between admin and user config files to allow easily switching users, without creating new copies of the same resources, which then leads to resources not being cleaned up
- moves non-CLI helpers to their own files

### DIFF
--- a/test/extended/util/client.go
+++ b/test/extended/util/client.go
@@ -75,10 +75,11 @@ import (
 // CLI provides function to call the OpenShift CLI and Kubernetes and OpenShift
 // clients.
 type CLI struct {
-	execPath        string
-	verb            string
-	configPath      string
-	adminConfigPath string
+	execPath          string
+	verb              string
+	currentConfigPath string
+	userConfigPath    string
+	adminConfigPath   string
 
 	// directory with static manifests, each file is expected to be a single manifest
 	// manifest files can be stored under directory tree
@@ -255,9 +256,9 @@ func (c *CLI) Username() string {
 
 // AsAdmin changes current config file path to the admin config.
 func (c *CLI) AsAdmin() *CLI {
-	nc := *c
-	nc.configPath = c.adminConfigPath
-	return &nc
+	c.currentConfigPath = c.adminConfigPath
+	framework.Logf("configPath is now %q", c.currentConfigPath)
+	return c
 }
 
 // ChangeUser changes the user used by the current CLI session.
@@ -274,14 +275,15 @@ func (c *CLI) ChangeUser(name string) *CLI {
 	if err != nil {
 		fatalErr(err)
 	}
-	c.configPath = f.Name()
-	err = clientcmd.WriteToFile(*kubeConfig, c.configPath)
+	c.userConfigPath = f.Name()
+	err = clientcmd.WriteToFile(*kubeConfig, c.userConfigPath)
 	if err != nil {
 		fatalErr(err)
 	}
 
 	c.username = name
-	framework.Logf("configPath is now %q", c.configPath)
+	c.currentConfigPath = c.userConfigPath
+	framework.Logf("configPath is now %q", c.currentConfigPath)
 	return c
 }
 
@@ -302,16 +304,16 @@ func (c *CLI) SetManagedNamespace() *CLI {
 }
 
 // WithoutNamespace instructs the command should be invoked without adding --namespace parameter
-func (c CLI) WithoutNamespace() *CLI {
+func (c *CLI) WithoutNamespace() *CLI {
 	c.withoutNamespace = true
-	return &c
+	return c
 }
 
 // WithToken instructs the command should be invoked with --token rather than --kubeconfig flag
-func (c CLI) WithToken(token string) *CLI {
-	c.configPath = ""
+func (c *CLI) WithToken(token string) *CLI {
+	c.currentConfigPath = ""
 	c.token = token
-	return &c
+	return c
 }
 
 // SetupProject creates a new project and assign a random user to the project.
@@ -618,8 +620,8 @@ func (c *CLI) TeardownProject() {
 		e2edebug.DumpAllNamespaceInfo(context.TODO(), c.kubeFramework.ClientSet, c.Namespace())
 	}
 
-	if len(c.configPath) > 0 {
-		os.Remove(c.configPath)
+	if len(c.userConfigPath) > 0 {
+		os.Remove(c.userConfigPath)
 	}
 
 	dynamicClient := c.AdminDynamicClient()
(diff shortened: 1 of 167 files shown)


test/extended/util/client.go
```diff
@@ -111,62 +109,63 @@ type resourceRef struct {
...
+	if len(options.PodSecurityEnforceLevel) > 0 {
+		kubeFramework.NamespacePodSecurityEnforceLevel = options.PodSecurityEnforceLevel
+	}
+	var kubeConfig string
+	if options.HyperShift {
+		var err error
+		kubeConfig, _, err = GetHypershiftManagementClusterConfigAndNamespace()
+		o.Expect(err).NotTo(o.HaveOccurred())
+	} else {
+		kubeConfig = KubeConfigPath()
```
deads2k, line 153: CLIOptions should provide an indicator for
1. use the global
2. use this path
3. use the hypershift thing above

test/extended/util/client.go
```diff
@@ -111,62 +109,63 @@ type resourceRef struct {
...
-		username:                "admin",
-		execPath:                "oc",
-		adminConfigPath:         KubeConfigPath(),
-		staticConfigManifestDir: StaticConfigManifestDir(),
+// CliOptions allows customization of the newely created CLI resource
+type CliOptions struct {
+	// KubeFramework to use, instead of creating a brand new one
+	KubeFramework *framework.Framework
+	// prefix for the created test namespace
+	BaseName string
```
deads2k, line 117: seems like this and withoutnamespace should be mutually exclusive.

test/extended/util/client.go
```diff
@@ -111,62 +109,63 @@ type resourceRef struct {
...
+	PodSecurityEnforceLevel admissionapi.Level
+	// whether to add --namespace flag in oc invocations
+	WithoutNamespace bool
+	// whether we're dealing with HyperShift cluster
+	HyperShift bool
+}
+
+// NewCLI is responsible for creating a new CLI object for invoking OpenShift CLI,
+// CliOptions allow customization of the created object.
+func NewCLI(options CliOptions) *CLI {
```
deads2k, line 128: Perhaps you want `NewCLI(opts ...Option)`

```go
type Option func(cliOptions) cliOptions

func withTestFramework(framework) Option
func withPSALevel(level) Option
func withoutNamespace() Option

```

test/extended/util/client.go
```diff
@@ -265,46 +240,75 @@ func (c *CLI) SetManagedNamespace() *CLI {
...
+				Annotations: map[string]string{
+					annotations.OpenShiftDescription: newUsername,
+					annotations.OpenShiftDisplayName: newNamespace,
+					"openshift.io/requester":         newUsername,
+				},
+			},
+		}
+		_, err := c.AdminKubeClient().CoreV1().Namespaces().Create(context.Background(), nsObject, metav1.CreateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		c.kubeFramework.AddNamespacesToDelete(nsObject)
```
deads2k, line 299: Is the kubeframework wired in some way that requires us to contain it instead of creating an entirely separate thign?

test/extended/util/client.go
```diff
@@ -191,9 +192,9 @@ func (c *CLI) Username() string {
 
 // AsAdmin changes current config file path to the admin config.
 func (c *CLI) AsAdmin() *CLI {
-	nc := *c
-	nc.configPath = c.adminConfigPath
-	return &nc
+	c.currentConfigPath = c.adminConfigPath
```
ardaguclu, line 195: in case of multiple `CLI` modifications within a test, would it make sense to protect all these changes with mutexes?

test/extended/util/client.go
```diff
@@ -238,16 +240,16 @@ func (c *CLI) SetManagedNamespace() *CLI {
...
 	c.withoutNamespace = true
-	return &c
+	return c
 }
 
 // WithToken instructs the command should be invoked with --token rather than --kubeconfig flag
-func (c CLI) WithToken(token string) *CLI {
-	c.configPath = ""
+func (c *CLI) WithToken(token string) *CLI {
+	c.currentConfigPath = ""
```
ardaguclu, line 250: same comment above

test/extended/util/client.go
```diff
@@ -323,113 +327,58 @@ func (c *CLI) setupProject() string {
...
-		ResourceAttributes: &kubeauthorizationv1.ResourceAttributes{
-			Namespace: newNamespace,
-			Verb:      "create",
-			Group:     "",
-			Resource:  "pods",
-		},
-	})
-	o.Expect(err).NotTo(o.HaveOccurred())
-
-	err = c.setupNamespacePodSecurity(newNamespace)
```
ardaguclu, line 423: Why we don't need this anymore?. Because we are already setting it to `restricted` in somewhere?

test/extended/util/client.go
ardaguclu: Do we really need to set BaseName when withoutnamespace is true?

soltysh: BaseName is used as a UniqueName here: https://github.com/kubernetes/kubernetes/blob/24fbb13eafec665e6f4b64961930774babd8b6d1/test/e2e/framework/framework.go#L279-L282 we could not require it being set and allow it to be randomly generated when empty, especially that this `f.UniqueName` is used pretty heavily in some cases. So for now I'd prefer we just keep it in there :wink:

test/extended/util/client.go
soltysh: We have a few cases where we modify the internals of the test framework, and I didn't want to address that in this PR.

test/extended/util/client.go
soltysh: I'd leave it up to consumers to deal with that kind of changes. I can probably add a comment to `NewCLI` about not having these kind of guarantees.

test/extended/util/client.go
soltysh: I'm squashing two methods into one, to ensure we always use the same code path, and clean all the resources after the tests.

test/extended/util/client.go
soltysh: Yeah, probably this comment along with the other about using `With*` methods should go together.

test/extended/util/client.go
soltysh: Good idea, especially that we already have a lot of these `With*` methods.

PR #28249, opened 2023-09-10
[HyperShift/KubeVirt] Raise wait time between pod exec attempts

orenc1:
Currently the retry loop has a wait time of 2 nanoseconds, which can overload the API server or the newly provisioned AWS LB service. Raising it to a more reasonable value of 10 seconds.


davidvossel (commented):
/lgtm
/approve

PR #28252, opened 2023-09-11
add disruptive flag from suite

tjungblu:


tjungblu:
waiting with the acknowledge-critical-fixes-only until the gate is lifted

PR #28253, opened 2023-09-12
When --from-repository use built-in tests only

soltysh:
Alternative to https://github.com/openshift/release/pull/43229

When we're invoking tests with `--from-repository` it means we needed to mirror the images, which is problematic when we're bumping k8s, since the new images are not mirrored, yet. The long term goal would be to read the additional images from the external binary, but we don't have that functionality there, yet. This is a temporary solution to allow landing k8s bump.

/assign @bertinatto


bertinatto (commented):
During TRT standup we've agreed to go with this approach

/lgtm

PR #28259, opened 2023-09-13
test/extended: add a bit more logging to the EgressIP probe tests

dcbw:
We've seen cases where all the probes were sent but for some reason the WaitGroup Wait() did not return. We're not quite sure why, so add a bit more logging to help find out.


danwinship:
/title NO-JIRA: test/extended: add a bit more logging to the EgressIP probe tests

PR #28261, opened 2023-09-14
OCPVE-295: rteval

jeff-roche:
This implements rteval as a realtime kernel test and updates the latency thresholds to be a little more forgiving

### DIFF
--- a/test/extended/kernel/tools.go
+++ b/test/extended/kernel/tools.go
@@ -2,17 +2,20 @@ package kernel
 
 import (
 	"encoding/json"
+	"encoding/xml"
 	"fmt"
 	"strconv"
+	"time"
 
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/pkg/errors"
 )
 
 const (
-	hwlatdetectThresholdusec = 5000
-	oslatThresholdusec       = 5000
-	cyclictestThresholdusec  = 5000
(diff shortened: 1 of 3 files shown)


test/extended/kernel/tools.go
```diff
@@ -185,3 +188,54 @@ func getProcessorCount(oc *exutil.CLI) (int, error) {
...
+		Range                 int      `xml:"range"`
+		Mean                  float32  `xml:"mean"`
+		MeanAbsoluteDeviation float32  `xml:"mean_absolute_deviation"`
+		StandardDeviation     float32  `xml:"standard_deviation"`
+	} `xml:"Measurements>Profile>cyclictest>system>statistics"`
+}
+
+func runRteval(oc *exutil.CLI) error {
+	// Create a working directory
+	rtevalWorkDir := "/tmp"
```
eggfoobar, line 210: Small nit, using `os.TempDir()` would be better here to avoid someone else doing things in that dir.

jeff-roche: I'll update this comment, os.TempDir() would create a temp dir on the machine running the tests, not the machine the tests are being run against

eggfoobar: Thanks! I missed that part, lgtm to me otherwise.

jeff-roche:
@bparees are you able to approve this one?

PR #28262, opened 2023-09-15
Port the clusteroperator intervals to new structured format

dgoodwin:


dgoodwin:
/assign @DennisPeriquet 

Dennis assigning you as you will hopefully be working with this data in the future.

dgoodwin:
/hold

I think I'm seeing a bug. Some OperatorState intervals are missing. We see no Progressing during upgrades for example.

dgoodwin:
Found the issue, there was a very quiet panic preventing most cluster operator intervals from being processed.

/hold cancel

PR #28266, opened 2023-09-18
OPRUN-3014: Introduce OLM V1 Tests

awgreene:
This commit introduces a series of tests that ensure OLM V1 is behaving as expected on tech preview clusters.

The first test case ensures that Catalogd controller successfully reconciles a catalog CR and makes it's content available for install.

The second test ensures that the operator controller successfully reconciles an operator CR and installs the expected operator using RukPak.

### DIFF
--- a/test/extended/olm/olmv1.go
+++ b/test/extended/olm/olmv1.go
@@ -1,0 +1,114 @@
+package operators
+
+import (
+	"encoding/json"
+	"path/filepath"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("[sig-operator] OLM V1", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc = exutil.NewCLI("default")
+
+		buildPruningBaseDir = exutil.FixturePath("testdata", "olm")
+		catalog             = filepath.Join(buildPruningBaseDir, "catalog.yaml")
+		operator            = filepath.Join(buildPruningBaseDir, "operator.yaml")
+	)
+
+	//TODO remove this check once OLM V1 has graduated
+	if !exutil.IsTechPreviewNoUpgrade(oc) {
+		g.Skip("the test is not expected to work within Tech Preview disabled clusters")
+	}
+
+	g.It("should unpack a catalog successfully", func() {
+		oc := oc
+
+		// configure Catalog before tests
+		// TODO: Build a catalog for CI purposes
+		configFile, err := oc.AsAdmin().Run("process").Args("--ignore-unknown-parameters=true", "-f", catalog, "-p", "NAME=test-catalog", "IMAGE=quay.io/operatorhubio/catalog:latest").OutputToFile("config.json")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		err = oc.AsAdmin().WithoutNamespace().Run("create").Args("-f", configFile).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func() string {
+			output, err := oc.AsAdmin().Run("get").Args("catalog", "test-catalog", "-o=json").Output()
+			if err != nil {
+				e2e.Logf("Failed to get valid catalog, error:%v", err)
+				return ""
+			}
+			type catalogStatus struct {
+				Phase string `json:"phase"`
+			}
+			type catalog struct {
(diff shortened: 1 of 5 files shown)


awgreene:
>Warning: The referenced jira issue has an invalid target version for the target branch this PR targets: expected the story to target the "4.15.0" version, but it targets "openshift-4.15" instead.

Hmm, the "4.15.0" version doesn't exist as an option in JIRA...

test/extended/olm/olmv1.go
```diff
@@ -0,0 +1,114 @@
...
+
+		// configure Catalog before tests
+		// TODO: Build a catalog for CI purposes
+		configFile, err := oc.AsAdmin().Run("process").Args("--ignore-unknown-parameters=true", "-f", catalog, "-p", "NAME=test-catalog", "IMAGE=quay.io/operatorhubio/catalog:latest").OutputToFile("config.json")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		err = oc.AsAdmin().WithoutNamespace().Run("create").Args("-f", configFile).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func() string {
+			output, err := oc.AsAdmin().Run("get").Args("catalog", "test-catalog", "-o=json").Output()
```
joelanford, line 44: Is it standard in e2e's to use `oc` this way? Can we use a Go-based client to communicate with the api server directly?

test/extended/olm/olmv1.go
```diff
@@ -0,0 +1,114 @@
...
+					Phase: "",
+				},
+			}
+			if err := json.Unmarshal([]byte(output), &parsed); err != nil {
+				e2e.Logf("Failed to parse catalog, error:%v", err)
+				return ""
+			}
+			return parsed.Status.Phase
+			// Check every 30 seconds as it takes awhile for the catalog to unpack
+		}, 5*time.Minute, 30*time.Second).Should(o.Equal("Unpacked"))
```
joelanford, line 66: Can we use `oc wait catalog test-catalog --for=condition=Unpacked` instead of polling every 30 seconds? A 30 second poll will mean an average of 15s of wasted time across runs.

test/extended/olm/olmv1.go
```diff
@@ -0,0 +1,114 @@
...
+
+	g.It("should install an operator successfully", func() {
+		oc := oc
+
+		configFile, err := oc.AsAdmin().Run("process").Args("--ignore-unknown-parameters=true", "-f", operator, "-p", "NAME=test-operator", "VERSION=0.6.0", "PACKAGE_NAME=argocd-operator").OutputToFile("config.json")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		err = oc.AsAdmin().WithoutNamespace().Run("create").Args("-f", configFile).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func() bool {
```
joelanford, line 78: Same questions re: go client or `oc wait`.

test/extended/olm/olmv1.go
```diff
@@ -0,0 +1,114 @@
...
+
+			if !apimeta.IsStatusConditionTrue(parsed.Status.Conditions, "Installed") {
+				e2e.Logf("Operator is not installed")
+				return false
+			}
+
+			return true
+		}, 5*time.Minute, time.Second).Should(o.BeTrue())
+	})
+
```
joelanford, line 113: Can we also check to see if there are some argo CD objects actually installed on the cluster to make sure that `Installed` condition isn't lying?

test/extended/olm/olmv1.go
awgreene: [Yes it is](https://github.com/openshift/origin/blob/master/test/extended/README.md#cli-interface), all tests in the extended suite seem to be built around the oc CLI.

test/extended/olm/olmv1.go
awgreene: This seems reasonable.

test/extended/olm/olmv1.go
awgreene: Sure.

test/extended/olm/olmv1.go
joelanford: There's a `--timeout` flag we can use on the `wait` subcommand.

test/extended/olm/olmv1.go
```diff
@@ -0,0 +1,114 @@
...
+		oc = exutil.NewCLI("default")
+
+		buildPruningBaseDir = exutil.FixturePath("testdata", "olm")
+		catalog             = filepath.Join(buildPruningBaseDir, "catalog.yaml")
+		operator            = filepath.Join(buildPruningBaseDir, "operator.yaml")
+	)
+
+	//TODO remove this check once OLM V1 has graduated
+	if !exutil.IsTechPreviewNoUpgrade(oc) {
+		g.Skip("the test is not expected to work within Tech Preview disabled clusters")
```
joelanford, line 30: Nit: double negative

```suggestion
		g.Skip("the test is only expected to work within Tech Preview clusters")
```

PR #28288, opened 2023-09-28
OCPBUGS-19909: Updating parameters for build timing PushImage test

coreydaley:
Builds on bare metal look to be running faster than they used to which is making the PushImage test fail.


coreydaley:
/assign @adambkaplan @deads2k 
Per our conversation on slack.

PR #28292, opened 2023-10-02
OPNET-330: fix DualStackIPv6Primary

rbbratta:
Since [OPNET-198](https://issues.redhat.com//browse/OPNET-198) we support IPv6 Primary cluster

Correctly set IPFamily based on the first family found

### DIFF
--- a/pkg/clioptions/clusterdiscovery/cluster.go
+++ b/pkg/clioptions/clusterdiscovery/cluster.go
@@ -56,6 +56,8 @@ type ClusterConfiguration struct {
 	// and dual-stack-specific tests are run
 	HasIPv4 bool
 	HasIPv6 bool
+	// IPFamily defines default IP stack of the cluster, replaces upstream getDefaultClusterIPFamily
+	IPFamily string
 
 	// HasSCTP determines whether SCTP connectivity tests can be run in the cluster
 	HasSCTP bool
@@ -234,10 +236,17 @@ func LoadConfig(state *ClusterState) (*ClusterConfiguration, error) {
 		config.NetworkPluginMode = string(state.NetworkSpec.DefaultNetwork.OpenShiftSDNConfig.Mode)
 	}
 
+	// replaces IPFamily as determined by upstream getDefaultClusterIPFamily
+	// always IPv4 default
+	config.IPFamily = "ipv4"
 	// Determine IP configuration
-	for _, cidr := range state.NetworkSpec.ServiceNetwork {
+	for i, cidr := range state.NetworkSpec.ServiceNetwork {
 		if utilnet.IsIPv6CIDRString(cidr) {
 			config.HasIPv6 = true
+			// if the first ServiceNetwork is IPv6 we are IPFamily "ipv6"
+			if i == 0 {
+				config.IPFamily = "ipv6"
+			}
 		} else {
 			config.HasIPv4 = true
 		}
--- a/pkg/clioptions/clusterdiscovery/provider.go
+++ b/pkg/clioptions/clusterdiscovery/provider.go
@@ -65,11 +65,8 @@ func InitializeTestFramework(context *e2e.TestContextType, config *ClusterConfig
 	e2e.AfterReadingAllFlags(context)
 	context.DumpLogsOnFailure = true
 
-	// these constants are taken from kube e2e and used by tests
-	context.IPFamily = "ipv4"
-	if config.HasIPv6 && !config.HasIPv4 {
-		context.IPFamily = "ipv6"
-	}
+	// IPFamily constants are taken from kube e2e and used by tests
+	context.IPFamily = config.IPFamily
 	return nil
 }
 
--- a/pkg/clioptions/clusterdiscovery/provider_test.go
+++ b/pkg/clioptions/clusterdiscovery/provider_test.go
@@ -141,6 +141,16 @@ var ovnKubernetesConfig = &operatorv1.NetworkSpec{
 	},
 }
 
+var ovnKubernetesConfigIPv6Primary = &operatorv1.NetworkSpec{
+	DefaultNetwork: operatorv1.DefaultNetworkDefinition{
+		Type: operatorv1.NetworkTypeOVNKubernetes,
+	},
+	ServiceNetwork: []string{
+		"fd02::/112",
+		"172.30.0.0/16",
+	},
+}
+
 var e2eTests = map[string]string{
 	"everyone":              "[Skipped:Wednesday]",
 	"not-gce":               "[Skipped:gce]",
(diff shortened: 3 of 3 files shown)


rbbratta:
Failures https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/42899/rehearse-42899-pull-ci-openshift-cluster-network-operator-master-e2e-vsphere-ovn-dualstack-primaryv6/1707789324319199232

```
: [sig-network] [Feature:IPv6DualStack] should create a single stack service with cluster ip from primary service range [Suite:openshift/conformance/parallel] [Suite:k8s] expand_less 	9s
{  fail [test/e2e/network/dual_stack.go:704]: service e2e-dualstack-3953/defaultclusterip expected family IPv4 at index[0] got IPv6
Error: exit with code 1
Ginkgo exit error 1: exit with code 1}
```

rbbratta:
@mkowalski

pkg/clioptions/clusterdiscovery/cluster.go
```diff
@@ -56,6 +56,8 @@ type ClusterConfiguration struct {
 	// and dual-stack-specific tests are run
 	HasIPv4 bool
 	HasIPv6 bool
+	// IPFamily matches the kube e2e way
+	IPFamily string
```
mkowalski, line 60: If anything, I'd rather describe this variable like described upstream in https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/test_context.go#L173, e.g.

`IPFamily defines default IP stack of the cluster`

pkg/clioptions/clusterdiscovery/cluster.go
```diff
@@ -242,6 +244,16 @@ func LoadConfig(state *ClusterState) (*ClusterConfiguration, error) {
 			config.HasIPv4 = true
 		}
 	}
+	// IP order matters for DualStack.
+	for _, cidr := range state.NetworkSpec.ServiceNetwork {
```
mkowalski, line 248: Can we have `if config.HasIPv6 && config.HasIPv4` before we run this loop? If this is something for dual-stack only, then let's run it only for dual-stack

rbbratta:
Old discusssion  https://github.com/openshift/origin/pull/26140#pullrequestreview-655624069

pkg/clioptions/clusterdiscovery/cluster.go
rbbratta: Will shorten comment.  This is for single and dual.

> Plus, as we are detecting primary stack based on the 1st network, would utilnet.IsIPv6CIDRString(state.NetworkSpec.ServiceNetwork[0]) suffice?

I was trying to think of a way to combine the loops.

mkowalski: >This is for single and dual.

I miss something to understand how this added code is relevant for single-stack... For single-stack the previous loop (just above the code you add) should already correctly detect the IP stack, or no?

rbbratta: The new loop replaces the old logic

```
// these constants are taken from kube e2e and used by tests
context.IPFamily = "ipv4"
if config.HasIPv6 && !config.HasIPv4 {
        context.IPFamily = "ipv6"
}
```

mkowalski: That's fair, thanks for explaining!

rbbratta: avoided loop by assuming IPv4 default.

cybertron:
/test e2e-openstack-ovn
/test e2e-aws-ovn-single-node
/lgtm

The other jobs seem to be pretty much perma-failing, but these two seem like they should pass.

pkg/clioptions/clusterdiscovery/provider.go
```diff
@@ -65,11 +65,8 @@ func InitializeTestFramework(context *e2e.TestContextType, config *ClusterConfig
 	e2e.AfterReadingAllFlags(context)
 	context.DumpLogsOnFailure = true
 
-	// these constants are taken from kube e2e and used by tests
-	context.IPFamily = "ipv4"
-	if config.HasIPv6 && !config.HasIPv4 {
-		context.IPFamily = "ipv6"
-	}
+	// IPFamily constants are taken from kube e2e and used by tests
```
danwinship, line 68: The comment here before was explaining "the reason we have the strings `"ipv4"` and `"ipv6"` hardcoded here is because that's what kube e2e uses, but it doesn't export them as constants we can import, so we have to hardcode them". So in the context of the rewrite, you should be explaining that in `pkg/clioptions/clusterdiscovery/cluster.go`. But maybe it would be better to add local constants for them (and explain there that they have to be the same as the upstream ones)? Or else use `v1.Family` for our config, and call `strings.ToLower(string(config.PrimaryIPFamily))` here and comment that the upstream tests use the strings `"ipv4"` and `"ipv6"`, which are also the lowercasified versions of the `v1.Family` constants.

pkg/clioptions/clusterdiscovery/cluster.go
```diff
@@ -56,6 +56,8 @@ type ClusterConfiguration struct {
 	// and dual-stack-specific tests are run
 	HasIPv4 bool
 	HasIPv6 bool
+	// IPFamily defines default IP stack of the cluster, replaces upstream getDefaultClusterIPFamily
+	IPFamily string
```
danwinship, line 60: Seems like it should be `PrimaryIPFamily`?

Also, `HasIPv4`+`HasIPv6`+`IPFamily` is redundant (and allows impossible configurations like `hasIPv6: false, ipFamily: "ipv6"`). We don't want to get rid of the old fields, for compatibility, but it would probably be nicer to make this `PrimaryIPFamily string` (or `PrimaryIPFamily v1.Family`?) plus `DualStack bool` and then deprecate `HasIPv4` and `HasIPv6`?

rbbratta: `IPFamily` matches `context.IPFamily`  Changing the name would confuse.

The only use case of `context.IPFamily` in e2e seems to be`ClusterIsIPv6` which is used frequently.

```
// ClusterIsIPv6 returns true if the cluster is IPv6
func (tc TestContextType) ClusterIsIPv6() bool {
	return tc.IPFamily == "ipv6"
}
```

So, assume IPv4 unless IPFamily is IPv6.  Then to detect dualstack something else.

```
	if c.IPFamily == "ipv6" {
		skips = append(skips, "[Feature:Networking-IPv4]")
	} else {
		skips = append(skips, "[Feature:Networking-IPv6]")
	}
	if !c.HasIPv4 || !c.HasIPv6 {
		// lack of "]" is intentional; this matches multiple tags
		skips = append(skips, "[Feature:IPv6DualStack")
	
```

with consts.

rbbratta: err, well 

```go
	if !c.DualStack {
		// lack of "]" is intentional; this matches multiple tags
		skips = append(skips, "[Feature:IPv6DualStack")
		if c.IPFamily == strings.ToLower(string(corev1.IPv6Protocol)) {
			skips = append(skips, "[Feature:Networking-IPv4]")
		} else {
			skips = append(skips, "[Feature:Networking-IPv6]")
		}
	}
	// else DualStack, run all the tests.
```

rbbratta: > 
> Also, `HasIPv4`+`HasIPv6`+`IPFamily` is redundant (and allows impossible configurations like `hasIPv6: false, ipFamily: "ipv6"`). 

I'm not sure what to do in the case of no discovery.  The `tc.Context` has an IPFamily but we seem to want to skip `ipv4` tests if there is no discovery, which implies that `HasIPv4` can be false while `IPFamily == "ipv4"`


```go
		{
			name:                 "complex override without discovery",
			provider:             `{"type":"aws","region":"us-east-2","zone":"us-east-2a","multimaster":false,"multizone":true}`,
			discoveredPlatform:   nil,
			optionalCapabilities: configv1.KnownClusterVersionCapabilities,
			expectedConfig:       `{"type":"aws","ProjectID":"","Region":"us-east-2","Zone":"us-east-2a","NumNodes":0,"MultiMaster":false,"MultiZone":true,"Zones":null,"ConfigFile":"","Disconnected":false,"SingleReplicaTopology":false,"NetworkPlugin":"","HasIPv4":false,"HasIPv6":false,"HasSCTP":false,"IsProxied":false,"IsIBMROKS":false,"HasNoOptionalCapabilities":false}`,
			runTests:             sets.NewString("everyone", "not-gce", "not-sdn", "not-multitenant", "online", "requires-optional-cap"),
		},

```

PR #28295, opened 2023-10-04
Automating test "check the quota after import-image with --all option" in upstrem

gangwgr:
@soltysh Please review this PR 
Test run - https://privatebin.corp.redhat.com/?5e3a606a01e6ba72#3VwE9iHzqvht8tEagksodSgqBorG6w1LGray2t8tW88u

### DIFF
--- a/test/extended/quota/resourcequota.go
+++ b/test/extended/quota/resourcequota.go
@@ -164,6 +164,58 @@ var _ = g.Describe("[sig-api-machinery][Feature:ResourceQuota]", func() {
 			})
 			o.Expect(err).NotTo(o.HaveOccurred())
 		})
+
+		g.It("check the quota after import-image with --all option", func() {
+			testProject := oc.SetupProject()
+			testResourceQuotaName := "my-imagestream-quota-" + testProject
+			clusterAdminKubeClient := oc.AdminKubeClient()
+
+			rq := &corev1.ResourceQuota{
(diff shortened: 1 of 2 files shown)


test/extended/quota/resourcequota.go
```diff
@@ -164,6 +164,58 @@ var _ = g.Describe("[sig-api-machinery][Feature:ResourceQuota]", func() {
...
+				}
+				if !equality.Semantic.DeepEqual(actualResourceQuota.Status.Used, expectedUsedStatus) {
+					return fmt.Errorf("unexpected current total usage: actual: %#v, expected: %#v", actualResourceQuota.Status.Used, expectedUsedStatus)
+				}
+				return nil
+			})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("trying to tag a container image")
+			err = oc.AsAdmin().WithoutNamespace().Run("import-image").Args("centos", "--from=quay.io/openshifttest/alpine", "--confirm=true", "--all=true", "-n", testProject).Execute()
```
soltysh, line 198: Nit: sometime in the future if that repo gets a new tag, this might start failing, so ideally it would be good to either expect min. 2 images, rather than exactly 2. Or even better explicitly import specific 2 tags, and then check  the number. But that's a minor problem that can be fixed in a followup.

soltysh (approved):
/lgtm
/approve

PR #28297, opened 2023-10-04
e2e node label test to verify blocking forbidden labels

harche:
/hold until https://github.com/openshift/kubernetes/pull/1736 merges


test/extended/security/labels.go
```diff
@@ -0,0 +1,62 @@
...
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLIWithPodSecurityLevel("node-label-e2e", admissionapi.LevelPrivileged)
+
+	g.It("TestOpenshiftNodeLabeling", func() {
+		clusterAdminKubeClientset := oc.AdminKubeClient()
+
+		workerNodes, err := clusterAdminKubeClientset.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{
+			LabelSelector: "node-role.kubernetes.io/worker",
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
```
rphillips, line 52: There needs to be an expect here to check for len(workerNodes.Items) == 0, since the test case could return 0 items.

test/extended/security/labels.go
```diff
@@ -0,0 +1,62 @@
...
+// kubeconfig present in /var/lib/kubelet/kubeconfig. It will capture the stdout/stderr of that execution
+// and process it to verify that the forbidden labels aren't getting applied.
+func testOpenshiftNodeLabeling(oc *exutil.CLI, node *corev1.Node, forbiddenLabel string) bool {
+	// `oc debug` command below attempts to set a forbidden label. This command is expected to fail. However,
+	// during testing it is observed that sometimes even though the command fails to set the forbidden label on the node
+	// it does not return a valid error and in fact returns with exit code 0. This ends up making
+	// the test flaky. Because of this observed flakiness, we are ignoring the error returned by the
+	// `oc debug` command below, and rely on the regex matching in matchForbiddenNodeLine function
+	// to decide if the node was successfully prevented from labeling the forbidden label.
+	out, _ := oc.AsAdmin().Run("debug").Args("node/"+node.Name, "--", "chroot", "/host", "/bin/bash", "-c",
```
rphillips, line 36: I think we should use wait.ExponentialBackoff to retry on error as well.

test/extended/security/labels.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+var _ = g.Describe("[sig-node] [Conformance] Prevent openshift node labeling on update by the node", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLIWithPodSecurityLevel("node-label-e2e", admissionapi.LevelPrivileged)
+
+	g.It("TestOpenshiftNodeLabeling", func() {
+		clusterAdminKubeClientset := oc.AdminKubeClient()
+
+		workerNodes, err := clusterAdminKubeClientset.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{
```
rphillips, line 49: might need to use wait.ExponentialBackoff here too to make sure we get the Items populated.

test/extended/security/labels.go
rphillips: Or wait with the previous comment to make sure Items is populated

test/extended/security/labels.go
```diff
@@ -0,0 +1,62 @@
...
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if len(workerNodes.Items) > 0 {
+			workerNode := &workerNodes.Items[0]
+			forbiddenLabels := []string{`node-role.kubernetes.io/etcd1=""`, `node-role.kubernetes.io/etcd=""`, `node-role.kubernetes.io/master=""`}
+			for _, forbiddenLabel := range forbiddenLabels {
+				result := testOpenshiftNodeLabeling(oc, workerNode, forbiddenLabel)
+				o.Expect(result).To(o.BeTrue())
+			}
+		}
```
rphillips, line 60: if should assert if len(workerNodes.Items) == 0

harche:
Contains the code from, https://github.com/openshift/origin/pull/28296 and https://github.com/openshift/origin/pull/28499

PR #28299, opened 2023-10-05
Port Node and Pod State to Structured Intervals

dgoodwin:
Ports the code that generates the node and pod state intervals we show in spyglass to use, and generate new structured intervals. These components interpret other "source" intervals to determine what intervals they should generate. 

This was incredibly complex and time consuming to get working between the complexity and the use of raw json files as test inputs and outputs.

I discovered that Intervals were not serialized via the same logic as EventIntervals (a dupe type we use just for serialization for some reason), and this is now addressed. As a result you'll see in some of the expected.json files, a few intervals are reordered to where they should be by the new logic. (primarily consideration of the To timestamp)

There is an unfortunate hack involved where Locators cannot be used as map keys because they contain a map themselves, go does not allow this. You'll notice a couple spots where I am maintaining a map of locatorStr -> locator, for lookup later when we need the real locators. This was done because I do not want us parsing locator strings, as this was one of the problems that led to this entire effort.

Apologies for the size but it was largely unavoidable, these parts of the code are linked, one goes and they all need to.

### DIFF
--- a/pkg/monitortests/node/watchpods/compute_intervals.go
+++ b/pkg/monitortests/node/watchpods/compute_intervals.go
@@ -14,7 +14,7 @@ import (
 
 func intervalsFromEvents_PodChanges(events monitorapi.Intervals, beginning, end time.Time) monitorapi.Intervals {
 	var intervals monitorapi.Intervals
-	podStateTracker := statetracker.NewStateTracker(monitorapi.ConstructionOwnerPodLifecycle, beginning)
+	podStateTracker := statetracker.NewStateTracker(monitorapi.ConstructionOwnerPodLifecycle, monitorapi.SourcePodState, beginning)
 	locatorToMessageAnnotations := map[string]map[string]string{}
 
 	for _, event := range events {
--- a/pkg/monitortests/node/watchpods/compute_intervals.go
+++ b/pkg/monitortests/node/watchpods/compute_intervals.go
@@ -12,33 +12,55 @@ import (
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 )
 
+// reduceToPodLocator takes a locator with a Pod key, and reduces it to *only* the keys
+// relevant to a pod. This allows us to consistently map to that pod when other keys may be
+// changing, such as containers, or when it's assigned to a node. It is assumed that
+// the incoming locator has a pod key at the point this is called.
+func reduceToPodLocator(locator monitorapi.Locator) monitorapi.Locator {
+	return monitorapi.NewLocator().PodFromNames(
+		locator.Keys[monitorapi.LocatorNamespaceKey],
+		locator.Keys[monitorapi.LocatorPodKey],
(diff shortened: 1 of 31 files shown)


dgoodwin:
I think this is ready for review. The missing intervals for nodes reaching desired config is fixed, and the pod lifecycle events are now sorting together.

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/28299/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1712155920533819392/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/e2e-timelines_spyglass_20231011-180818.html

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/28299/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1712155920533819392/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/e2e-timelines_kube-apiserver_20231011-180818.html

For comparison, here is a job WITHOUT these changes, the node-state or pods sections should be identical: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/28299/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1712155920533819392/artifacts/e2e-aws-ovn-upgrade/openshift-e2e-test/artifacts/junit/e2e-timelines_kube-apiserver_20231011-180818.html

pkg/monitortests/node/watchpods/compute_intervals.go
```diff
@@ -49,39 +51,55 @@ func intervalsFromEvents_PodChanges(events monitorapi.Intervals, beginning, end
...
 
 		case isContainer && isContainerLifecycleTransition:
-			containerToLifecycleTransitions[container.ToLocator()] = append(containerToLifecycleTransitions[container.ToLocator()], event)
+			containerToLifecycleTransitions[cls] = append(containerToLifecycleTransitions[cls], event)
 
 		case isContainer && isContainerReadyTransition:
-			containerToReadinessTransitions[container.ToLocator()] = append(containerToReadinessTransitions[container.ToLocator()], event)
+			containerToReadinessTransitions[cls] = append(containerToReadinessTransitions[cls], event)
 
 		case isKubeletReadinessCheck:
```
DennisPeriquet, line 101: do we need `isContainer &&` here like the ones above that also use `cls`? or maybe `isKubeletReadinessCheck` imples that `isContainer == true`?

dgoodwin: Shouldn't no, this was it's original form, kubelet isn't run in a container (I believe it's an os level systemd service), and these intervals would not have a container key in their locator.

PR #28222, opened 2023-08-25
Cert test 01 initial sketch

deads2k:
`openshift-tests run-test "[sig-arch][Late] all tls artifacts must be known [Suite:openshift/conformance/parallel]"`

/hold for fake bump to be real

leverages https://github.com/openshift/library-go/pull/1568 to generate.  We'll need some struct updates here.

I think the path to an enforcing test is

1. library-go merges
1. make the test non-failing
1. find a way to name the json file based on the job type in a useful way (see https://github.com/openshift/origin/blob/master/pkg/monitortestlibrary/platformidentification/types.go#L152).  We likely need platform, network, topology tuples.
1. merge
1. push the starting TLS info collected in nightlies to openshift/api for "known" (this stops the hole from getting deeper)
1. add owner evaluation with whitelist of grace and start whittling down the list

This uses a pre-existing library, moved to library-go, to read the info we have.

```
######
new TLS artifacts like certificates and ca bundles must be registered
--namespace=openshift-cloud-credential-operator secret/cloud-credential-operator-serving-cert
--namespace=kube-system secret/konnectivity-agent
--namespace=openshift-cluster-machine-approver secret/machine-approver-tls
--namespace=openshift-cluster-node-tuning-operator secret/node-tuning-operator-tls
--namespace=openshift-monitoring secret/alertmanager-main-tls
--namespace=openshift-monitoring secret/cluster-monitoring-operator-tls
--namespace=openshift-cluster-storage-operator secret/cluster-storage-operator-serving-cert
--namespace=openshift-cluster-version secret/cluster-version-operator-serving-cert
--namespace=openshift-config-operator secret/config-operator-serving-cert
--namespace=openshift-console secret/console-serving-cert
--namespace=openshift-ingress secret/default-ingress-cert
--namespace=openshift-dns secret/dns-default-metrics-tls
--namespace=openshift-dns-operator secret/metrics-tls
--namespace=openshift-image-registry secret/image-registry-tls
--namespace=openshift-ingress-operator secret/metrics-tls
--namespace=openshift-kube-apiserver-operator secret/kube-control-plane-signer
--namespace=openshift-monitoring secret/federate-client-certs
--namespace=openshift-kube-controller-manager-operator secret/kube-controller-manager-operator-serving-cert
--namespace=openshift-machine-config-operator secret/mcc-proxy-tls
--namespace=openshift-machine-config-operator secret/mco-proxy-tls
--namespace=openshift-monitoring secret/kube-state-metrics-tls
[...]

PR #28305, opened 2023-10-06
Add test which verifies every certificate has necessary annotations

vrutkovs:
Rework of https://github.com/openshift/origin/pull/28222

The test would ensure known TLS artifacts have expected description and owning jira components set as annotations.

TODO: 
* [x] convert tls-artifacts-..go into jsons (easier to export)
* [x] Break into components
* [x] Add scripts to verify that TLS artifacts have owners, description and jira component set
* [x] Add scripts to generate one JSON for recert tool
* [x] Use `go:embed` instead of bindata to assemble owner jsons directly


deads2k:
closer to  https://github.com/openshift/origin/pull/28321?

hack/verify-tls-info.sh
```diff
@@ -0,0 +1,16 @@
...
+function cleanup() {
+    return_code=$?
+    os::test::junit::generate_report
+    os::util::describe_return_code "${return_code}"
+    exit "${return_code}"
+}
+trap "cleanup" EXIT
+
+os::test::junit::declare_suite_start "verify/generated"
+os::cmd::expect_success "${OS_ROOT}/hack/update-tls-info.sh"
```
deads2k, line 13: I don't speak bash.  Is this mutating files in my repo?  If so, we need a different approach that doesn't do that.  Generating to a tmp directory could work.

pkg/certs/main.go
deads2k, line 1: this filename is weird and the package is dangling

I'm supportive of `pkg/cmd/generator/certs/<command>/` that has a cobra command like other peers and a `cmd/test-generator` with a main.

pkg/certs/main.go
```diff
@@ -0,0 +1,47 @@
...
+	"strings"
+
+	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+func GetPKIRegistryInfo(platform, network string) (certgraphapi.PKIRegistryInfo, error) {
+	combined := certgraphapi.PKIRegistryInfo{}
+	chunkFileNames := exutil.FixturePaths("testdata", "tls", "ownership")
```
deads2k, line 16: avoid bindata.  Pass a path to this and read the files directly.

pkg/test/ginkgo/external.go
```diff
@@ -29,6 +29,11 @@ type serializedTest struct {
 func externalTestsForSuite(ctx context.Context) ([]*testCase, error) {
 	var tests []*testCase
 
+	// TODO get client from real client location
```
deads2k, line 32: this looks weird, why?

test/extended/util/certs/ownership/generate.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Fprintf(os.Stderr, "error: requires exactly one argument\n")
+		os.Exit(1)
+	}
+	outputDir := os.Args[1]
+
+	image.InitializeImages("")
```
deads2k, line 20: why?

test/extended/util/certs/ownership/generate.go
```diff
@@ -0,0 +1,59 @@
...
+	exutil.WithCleanup(func() {})
+
+	// TODO[vrutkovs]: read this from openshift/api
+	files := exutil.FixturePaths("testdata", "tls", "ownership")
+	for _, fileName := range files {
```
deads2k, line 25: I think we want to read from the entire rawTLS directory and transform into your struct from library-go: https://github.com/openshift/library-go/blob/master/pkg/certs/cert-inspection/certgraphapi/type_registry.go#L4

with keys based on namespace/name

test/extended/util/certs/info/generate.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Fprintf(os.Stderr, "error: requires exactly one argument\n")
+		os.Exit(1)
+	}
+	outputDir := os.Args[1]
+
+	image.InitializeImages("")
```
deads2k, line 21: remove?

test/extended/util/certs/info/generate.go
```diff
@@ -0,0 +1,59 @@
...
+	exutil.WithCleanup(func() {})
+
+	var (
+		architectures = []string{"arm64", "amd64"}
+		platforms     = []string{"aws", "azure", "gcp", "openstack", "metal"}
+		networks      = []string{"ovn", "sdn"}
+	)
+	for _, architecture := range architectures {
```
deads2k, line 29: oh, is this the command producing a serialized version of https://github.com/openshift/library-go/blob/master/pkg/certs/cert-inspection/certgraphapi/type_registry.go#L4 ?

To do so, I think we need to read all the raw TLS info content (PR producing that updated here https://github.com/openshift/origin/pull/28321), then create a `certs[InClusterSecretLocation]=PKIRegistryCertKeyPairInfo` that will fail well on conflicting information for different combinations. And perhaps after our first version we grow the library-go API to indicate which topology, arch, platform, network it exists on.

hack/verify-tls-info.sh
vrutkovs: Yes, similar to `verify-generated` - it updates the code and checks if any change appeared in `git diff`.

We know which files are generated, so I guess this script could use tempdir, yeah

pkg/certs/main.go
```diff
@@ -0,0 +1,47 @@
...
+		err := loadTLSRegistryChunk(fileName, &combined)
+		if err != nil {
+			// TODO: warn and skip instead
+			return combined, fmt.Errorf("failed to find chunk %s: %v", fileName, err)
+		}
+	}
+	return combined, nil
+}
+
+func loadTLSRegistryChunk(filename string, registry *certgraphapi.PKIRegistryInfo) error {
```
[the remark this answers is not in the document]
vrutkovs: Part of PKI registry file, grouped by owner

pkg/test/ginkgo/external.go
vrutkovs: Probably a bad rebase :/

test/extended/util/certs/info/generate.go
vrutkovs: Its necessary for `exutil.FixturePath` outside of the test. I'll rework it to avoid bindata

test/extended/util/certs/info/generate.go
vrutkovs: I've realized that we probably don't need multiple combinations. Certs don't change owner/description in different cluster topologies and in-cluster secrets can be a subset of expected secrets (to handle different secret names in older versions). 
So instead we'll have one `tls-artifacts` containing all known certs

pkg/certs/main.go
vrutkovs: Done

pkg/test/ginkgo/external.go
vrutkovs: oh, its a commit from https://github.com/openshift/origin/pull/28222 - https://github.com/openshift/origin/commit/3aec209b700d83ebf65dd48d281295fd06116d05.

Probably not necessary here

test/extended/util/certs/ownership/generate.go
vrutkovs: imo its easier to group by owner - respective teams would have to make changes in one file, which is better UX

pkg/certs/main.go
vrutkovs: Chunks grouped by owner are not stored there, but resulting TLS artifact registry is stored there (as its required to be readible by the test with no access to source)

cmd/update-tls-ownership/update-tls-ownership.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/openshift/origin/pkg/cmd/generator/certs"
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	root := &cobra.Command{
```
deads2k, line 12: Can this command produce a json file https://github.com/openshift/library-go/blob/master/pkg/certs/cert-inspection/certgraphapi/type_registry.go#L4 that is stored in /tls/owner-info?

pkg/cmd/generator/certs/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+func ValidateTLSOwnership(outputDir string) error {
+	var errors []error
+
+	owners, err := CombineAllOwnershipFiles()
+	if err != nil {
+		return err
+	}
+	for _, chunk := range owners {
+		if err := validateChunk(&chunk); err != nil {
```
deads2k, line 90: when this gets used in an e2e test, we want to be sure that existing violators (our current state) don't fail the test.  I think that requires
1. know which secrets and configmaps lack owners
2. know which secrets and confimaps in the cluster lack owners
3. subtract the known from current, and any new ones will fail.

vrutkovs: 1. `verify-tls-owners` will ensure that all registred artifacts have owners set
2 and 3. would show up as unregistered artifact:
```
 Unregistered TLS artifact: --namespace=openshift-etcd, secret/etcd-serving-ip-10-0-94-141.us-east-2.compute.internal
```

cmd/update-tls-ownership/update-tls-ownership.go
vrutkovs: Yes. Updated it to generate `tls/rawTLSInfo/tls-info.json` and `tls/ownership/known-owners.json`

deads2k:
the detection in `[sig-arch][Late] all tls artifacts must be known [Suite:openshift/conformance/parallel]` looks good. But I think formatting could be tightened to make it easier to read

vrutkovs:
>But I think formatting could be tightened to make it easier to read

Lets reformat this into:
```
Unregistered artifacts:
{
    "configMapLocation": {
        "Namespace": "openshift-foo",
        "Name": "bar"
    },
    "certificateAuthorityBundleInfo": {
        "owningJiraComponent": "",
        "description": ""
    },
    "configMapLocation": {
        "Namespace": "openshift-foo",
        "Name": "baz"
    },
    "certificateAuthorityBundleInfo": {
        "owningJiraComponent": "",
        "description": ""
    }
},
```
so that it would be easy to update `tls-ownership.json`?

pkg/certs/types.go
```diff
@@ -0,0 +1,48 @@
+package certs
+
+import (
+	"strings"
+
+	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+)
+
+type ConfigMapRefByNamespaceName []certgraphapi.InClusterConfigMapLocation
```
deads2k, line 9: followup, I'm in favor of putting these in a library-go util package.

deads2k: > ollowup, I'm in favor of putting these in a library-go util package.

err, not these, the sorting functions

tls/ownership/utils.go
```diff
@@ -0,0 +1,8 @@
+package ownership
```
deads2k, line 1: suggested followup: move this into the top of t he TLS dir

test/extended/operators/certs.go
```diff
@@ -49,6 +56,82 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+
+		nodes := map[string]int{}
+		controlPlaneLabel := labels.SelectorFromSet(map[string]string{"node-role.kubernetes.io/control-plane": ""})
+		nodeList, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{LabelSelector: controlPlaneLabel.String()})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		for i, node := range nodeList.Items {
+			nodes[node.Name] = i
+		}
+
+		actualPKIContent, err := certgraphanalysis.GatherCertsFromPlatformNamespaces(ctx, kubeClient)
```
deads2k, line 75: if you layer `certsToRegistryInfo` with a function that takes `(pkiLists ...*PKIList)` and use that for the both this and  `GetPKIInfoFromRawData`, you'll be able to look at the resulting ownership lists instead of raw values.

test/extended/operators/certs.go
```diff
@@ -49,6 +56,82 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+
+		messages := []string{}
+
+		unregisteredSecrets := map[certgraphapi.InClusterSecretLocation]certgraphapi.PKIRegistryInClusterCertKeyPair{}
+
+		for _, currCertKeyPair := range actualPKIContent.InClusterResourceData.CertKeyPairs {
+			currLocation := currCertKeyPair.SecretLocation
+			expectedSecret, err := certgraphutils.LocateCertKeyPair(currLocation, expectedPKIContent.CertKeyPairs)
+			if err != nil {
+				unregisteredSecrets[currLocation] = currCertKeyPair
```
deads2k, line 89: once our expected no longer includes revisioned secrets, this will be vulnerable.  Refactoring as suggested https://github.com/openshift/origin/pull/28305/files#r1376642834 will bring them both into the same type as `*certgraphapi.PKIRegistryInfo` which we likely want.

test/extended/operators/certs.go
```diff
@@ -49,6 +56,82 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+				continue
+			}
+			if expectedSecret.CertKeyInfo.Description == "" || expectedSecret.CertKeyInfo.OwningJiraComponent == "" {
```
deads2k, line 92: need to diff against white list in https://github.com/openshift/origin/blob/master/tls/violations/ownership/ownership-violations.json

test/extended/operators/certs.go
```diff
@@ -49,6 +56,82 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+				continue
+			}
+			errorMessage := o.Equal(expectedSecret.CertKeyInfo).FailureMessage(currCertKeyPair.CertKeyInfo)
```
deads2k, line 95: what does this line do?  Can we do without an `o.Equal` something?

test/extended/operators/certs.go
deads2k, line 92: separate the handling of description and owningjira component.

I think you'll want to eventually factor enough to make it ratcheting to enforce a particular property.

test/extended/operators/certs.go
```diff
@@ -49,6 +56,82 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 	})
 
+	g.It("all tls artifacts must be known", func() {
```
deads2k, line 59: I see three tests here which should be separated with `By`s.

1. registered/known
2. owned
3. described

deads2k:
inconsistent naming of etcd is failing in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/28305/pull-ci-openshift-origin-master-e2e-aws-ovn-fips/1719014128883863552

fixed in some of the naming remapping in library-go somewhere?

deads2k:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/28305/pull-ci-openshift-origin-master-e2e-aws-ovn-single-node-upgrade/1719014129001304064

could be fixed with revision handling.

pkg/certs/types.go
vrutkovs: Right, I'll update https://github.com/openshift/library-go/pull/1592 and remove those on library-go bump

test/extended/operators/certs.go
vrutkovs: This would generate the diff error message:
```
Expected
    <certgraphapi.PKIRegistryCertificateAuthorityInfo>: 
        description: ""
        owningJiraComponent: "Etcd"
to equal
    <certgraphapi.PKIRegistryCertificateAuthorityInfo>: 
        description: ""
        owningJiraComponent: ""
```

Perhaps `cmp` similar to what we have for ownership.md would be cleaner?

test/extended/operators/certs.go
```diff
@@ -49,6 +56,82 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+		if len(unregisteredConfigMaps) > 0 {
+			for _, cm := range unregisteredConfigMaps {
+				messages = append(messages, fmt.Sprintf("Unregistered TLS artifact: --namespace=%s, configmap/%s\n", cm.ConfigMapLocation.Namespace, cm.ConfigMapLocation.Name))
+			}
+		}
+
+		if len(messages) > 0 {
+			// TODO: uncomment when test no longer fails and enhancement is merged
+			//g.Fail(strings.Join(messages, "\n"))
+			testresult.Flakef(strings.Join(messages, "\n"))
```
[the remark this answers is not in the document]
vrutkovs: I'd rather keep it in flake longer - for about a week, so that all payload blocking / periodic jobs would run at least once and all existing secrets are registered.

test/extended/operators/certs.go
deads2k: > Perhaps `cmp` similar to what we have for ownership.md would be cleaner?

Yeah

test/extended/operators/certs.go
```diff
@@ -49,6 +57,123 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+			}
+		}
+
+		for _, currCABundle := range actualPKIContent.InClusterResourceData.CertificateAuthorityBundles {
+			currLocation := currCABundle.ConfigMapLocation
+			expectedCABundle, err := certgraphutils.LocateCertificateAuthorityBundle(currLocation, expectedPKIContent.CertificateAuthorityBundles)
+			if err != nil {
+				continue
+			}
+			errorMessage := o.Equal(expectedCABundle.CABundleInfo).FailureMessage(currCABundle.CABundleInfo)
```
deads2k, line 165: This looks overly strict.  It would fail on changes to the description, right?

Separate the checking so there is one block for checking owners and another block for checking descriptions, so that someone could copy/paste to have a new section for checking "regenerates on expired certificate"

test/extended/operators/certs.go
```diff
@@ -49,6 +57,123 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+				//g.Fail("Failed to marshal registry %#v: %v", newTLSRegistry, err)
+				testresult.Flakef("Failed to marshal registry %#v: %v", newTLSRegistry, err)
+			}
+			// TODO: uncomment when test no longer fails and enhancement is merged
+			//g.Fail(fmt.Sprintf("Unregistered TLS certificates:\n%s", registryString))
+			testresult.Flakef(fmt.Sprintf("Unregistered TLS certificates found:\n%s\nSee tls/ownership/README.md in origin repo", registryString))
+		}
+	})
+
+	g.It("all registered tls artifacts must have expected metadata", func() {
```
deads2k, line 123: either separate Its or separate Bys for handling owners separate from descriptions separate from "whatever else".

test/extended/operators/certs.go
vrutkovs: TODO: validate each property separately, maintain separate violation files, mention specific reason in markdown file

deads2k: > TODO: validate each property separately, maintain separate violation files, mention specific reason in markdown file

We already have enough to do that if you scope your PR to Owners only.

test/extended/operators/certs.go
```diff
@@ -49,6 +58,121 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+
+		messages := []string{}
+
+		for _, currCertKeyPair := range actualPKIContent.InClusterResourceData.CertKeyPairs {
+			currLocation := currCertKeyPair.SecretLocation
+			expectedSecret, err := certgraphutils.LocateCertKeyPair(currLocation, expectedPKIContent.CertKeyPairs)
+			if err != nil {
+				continue
+			}
+			if diff := cmp.Diff(expectedSecret.CertKeyInfo.OwningJiraComponent, currCertKeyPair.CertKeyInfo.OwningJiraComponent); len(diff) > 0 {
```
deads2k, line 154: followup to loosen so that any non-empty answer is acceptable.

PR #28311, opened 2023-10-09
Add debug info for interval counts and times

DennisPeriquet:
Jobs like [this](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.15-upgrade-from-stable-4.14-e2e-azure-sdn-upgrade/1710475847271452672), show intervals in the second spyglass chart that are in the upgrade phase; this is seen in the events.json and html files (what you see for spyglass).  This PR prints the first/last interval times for when those json files are written to disk to try to figure out where these extra intervals are coming from.

Also, we use the To (instead of From) since To is later in hopes of avoiding 'entry too far behind'.

The symptom does not happen consistently in CI since the sno upgrade test is the only one with two spyglass charts.  We can let this run in the wild to see if we can get more data into when and why this is happening.

### DIFF
--- a/pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
+++ b/pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
@@ -191,13 +191,11 @@ func UploadIntervalsToLoki(intervals monitorapi.Intervals) error {
 
 	for _, i := range intervals {
 		logLine := map[string]string{
-			"_entry": i.Message,
+			"_entry": fmt.Sprintf("%s, from=%s, to=%s", i.Message, i.From.String(), i.To.String()),
 		}
 
+		// We don't care to ingest intervals for e2e-tests at this time.
 		if strings.HasPrefix(i.Locator, "e2e-test/\"") {
-			startIndex := strings.Index(i.Locator, "\"") + 1
-			endIndex := strings.LastIndex(i.Locator, "\"")
-			logLine["e2e-test"] = i.Locator[startIndex:endIndex]
 			continue
 		}
 
--- a/pkg/monitor/monitor.go
+++ b/pkg/monitor/monitor.go
@@ -177,6 +177,9 @@ func (m *Monitor) SerializeResults(ctx context.Context, junitSuiteName, timeSuff
 	}
 
 	fmt.Fprintf(os.Stderr, "Writing to storage.\n")
+	fmt.Fprintf(os.Stderr, "  m.startTime = %s\n", m.startTime)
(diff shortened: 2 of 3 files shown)


pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
```diff
@@ -249,8 +249,13 @@ func UploadIntervalsToLoki(intervals monitorapi.Intervals) error {
...
 		}
 
+		// When 'intervals' was processed by Intervals.Slice(), the value of Interval.To is
+		// used to tell if the interval belongs in the slice. Since the value of Interval.To
+		// is always more recent than (or equal to) Interval.From, we use it for the
+		// timestamp here to reduce the chances of getting the 'entry too far behind' error
+		// from Loki.
 		values = append(values, []interface{}{
-			strconv.FormatInt(i.From.UnixNano(), 10),
+			strconv.FormatInt(i.To.UnixNano(), 10),
```
dgoodwin, line 258: This doesn't look right, this looks like the place we're defining the timestamp at which the "log" line took place, that is the from for an interval, if we change it to the "to", then ordering will go wonky when we query, things that started before other things will appear out of order depending on when the interval finished. from/to should not be the cause of the too old problems in loki afaict.

pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
```diff
@@ -192,12 +192,12 @@ func UploadIntervalsToLoki(intervals monitorapi.Intervals) error {
 	for _, i := range intervals {
 		logLine := map[string]string{
 			"_entry": i.Message,
+			"from":   i.From.String(),
+			"to":     i.To.String(),
```
dgoodwin, line 196: Hmm this looks like it will label the entries in loki and positively explode the cardinality, probably taking all of loki down. You could append to the something to the i.Message for now. I hope to get these logged as json sometime soon anyhow.

dgoodwin (commented):
Couple changes look needed to me, but the idea of printing out the times makes sense to debug this further.

pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
DennisPeriquet: hm... ok will put that back.

pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
DennisPeriquet: Nice catch!  I will append to the i.Message.

PR #28320, opened 2023-10-11
OCPBUGS-18776: [sig-instrumentation] tests fail due to JSON parsing error.

raptorsun:
Replace shell call of curl with net/http implementation in Prometheus helper functions.
So that timeout requests will not get truncated and reassembled responses.

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -5,9 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os/exec"
 	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
@@ -367,7 +365,7 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 
 			g.By("checking the prometheus metrics path")
 			var metrics map[string]*dto.MetricFamily
-			o.Expect(wait.PollImmediate(10*time.Second, 2*time.Minute, func() (bool, error) {
+			o.Expect(wait.PollUntilContextTimeout(context.Background(), 10*time.Second, 2*time.Minute, true, func(context.Context) (bool, error) {
 				results, err := getBearerTokenURLViaPod(ns, execPod.Name, fmt.Sprintf("%s/metrics", prometheusSvcURL), bearerToken)
 				if err != nil {
 					e2e.Logf("unable to get metrics: %v", err)
@@ -400,14 +398,16 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			g.By("verifying a service account token is able to authenticate")
-			err = expectBearerTokenURLStatusCodeExec(fmt.Sprintf("%s/api/v1/targets", queryURL), bearerToken, 200)
+
+			err = helper.ExpectURLStatusCodeExec(helper.MustJoinUrlPath(queryURL, "api/v1/targets"), bearerToken, 200)
+
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			g.By("verifying a service account token is able to access the Prometheus API")
 			// expect all endpoints within 60 seconds
 			var lastErrs []error
-			o.Expect(wait.PollImmediate(10*time.Second, 2*time.Minute, func() (bool, error) {
-				contents, err := getBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
+			o.Expect(wait.PollUntilContextTimeout(context.Background(), 10*time.Second, 2*time.Minute, true, func(context.Context) (bool, error) {
+				contents, err := helper.GetBearerTokenURL(helper.MustJoinUrlPath(prometheusURL, "api/v1/targets"), bearerToken)
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 				targets := &prometheusTargets{}
@@ -456,7 +456,7 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 
 			g.By("verifying all targets are exposing metrics over secure channel")
 			var insecureTargets []error
-			contents, err := getBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
+			contents, err := helper.GetBearerTokenURL(helper.MustJoinUrlPath(prometheusURL, "api/v1/targets"), bearerToken)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			targets := &prometheusTargets{}
@@ -579,8 +579,8 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 
 		g.It("should provide ingress metrics", func() {
 			var lastErrs []error
-			o.Expect(wait.PollImmediate(10*time.Second, 4*time.Minute, func() (bool, error) {
-				contents, err := getBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
+			o.Expect(wait.PollUntilContextTimeout(context.Background(), 10*time.Second, 4*time.Minute, true, func(ctx context.Context) (bool, error) {
+				contents, err := helper.GetBearerTokenURL(helper.MustJoinUrlPath(prometheusURL, "api/v1/targets"), bearerToken)
 				o.Expect(err).NotTo(o.HaveOccurred())
 
 				targets := &prometheusTargets{}
@@ -756,27 +756,6 @@ func findMetricLabels(f *dto.MetricFamily, labels map[string]string, match strin
 		}
 	}
 	return result
-}
-
-func expectBearerTokenURLStatusCodeExec(url, bearer string, statusCode int) error {
-	cmd := fmt.Sprintf("curl -k -s -H 'Authorization: Bearer %s' -o /dev/null -w '%%{http_code}' %q", bearer, url)
-	output, err := exec.Command("bash", "-e", "-c", cmd).Output()
-	if err != nil {
-		return fmt.Errorf("host command failed: %v\n%s", err, output)
-	}
-	if string(output) != strconv.Itoa(statusCode) {
-		return fmt.Errorf("last response from server was not %d: %s", statusCode, output)
-	}
-	return nil
-}
-
-func getBearerTokenURL(url, bearer string) (string, error) {
-	cmd := fmt.Sprintf("curl -s -k -H 'Authorization: Bearer %s' %q", bearer, url)
-	output, err := exec.Command("bash", "-e", "-c", cmd).Output()
-	if err != nil {
-		return "", fmt.Errorf("host command failed: %v\n%s", err, output)
-	}
-	return string(output), nil
 }
 
 func getBearerTokenURLViaPod(ns, execPodName, url, bearer string) (string, error) {
--- a/test/extended/util/prometheus/helpers.go
+++ b/test/extended/util/prometheus/helpers.go
@@ -2,11 +2,12 @@ package prometheus
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
-	"os/exec"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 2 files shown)


test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,34 @@ type prometheusResponseData struct {
 }
 
 // GetBearerTokenURL makes http request with bearer token
-func GetBearerTokenURL(url, bearer string) (string, error) {
-	cmd := fmt.Sprintf("curl --retry 15 --max-time 2 --retry-delay 1 -s -k -H 'Authorization: Bearer %s' %q", bearer, url)
-	output, err := exec.Command("bash", "-e", "-c", cmd).Output()
+func GetBearerTokenURL(url, bearerToken string) (string, error) {
```
simonpasquier, line 49: /hold
we need to implement a retry strategy

test/extended/util/prometheus/helpers.go
```diff
@@ -286,18 +308,25 @@ func RunQueries(ctx context.Context, prometheusClient prometheusv1.API, promQuer
 // ExpectURLStatusCodeExec attempts connection to url returning an error
 // upon failure or if status return code is not equal to any of the statusCodes.
 func ExpectURLStatusCodeExec(url string, statusCodes ...int) error {
-	cmd := fmt.Sprintf("curl -k -s -o /dev/null -w '%%{http_code}' %q", url)
-	output, err := exec.Command("bash", "-e", "-c", cmd).Output()
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{
```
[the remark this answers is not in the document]
raptorsun: added 2 second timeout in new commit.

test/extended/util/prometheus/helpers.go
raptorsun: retry up to twice in new commit :)

simonpasquier: hmm the curl command retried up to 15 times? In hindsight, we should probably retry less but with a more generous timeout.

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,49 @@ type prometheusResponseData struct {
...
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return "", err
+	}
+
+	req.Header.Add("Authorization", "Bearer "+bearerToken)
+
+	const maxRetries = 2
+	var resp *http.Response = nil
+	for retries := 0; retries < maxRetries; retries++ {
```
simonpasquier, line 67: I'd recommend using `wait.PollUntilContextTimeout()` from the `k8s.io/apimachinery/pkg/util/wait` package.

raptorsun: The retry loop is replaced by the function`wait.PollUntilContextTimeout()`. Meanwhile the timeout limit is increased to 60 seconds with retry interval of 10 seconds.

test/extended/util/prometheus/helpers.go
raptorsun: retry interval is set to 10 seconds, total timeout is set to 60 seconds in the new commit.

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,50 @@ type prometheusResponseData struct {
...
 // GetBearerTokenURL makes http request with bearer token
-func GetBearerTokenURL(url, bearer string) (string, error) {
-	cmd := fmt.Sprintf("curl --retry 15 --max-time 2 --retry-delay 1 -s -k -H 'Authorization: Bearer %s' %q", bearer, url)
-	output, err := exec.Command("bash", "-e", "-c", cmd).Output()
+func GetBearerTokenURL(url, bearerToken string) (string, error) {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{
+		Timeout:   time.Duration(10 * time.Second),
```
jan--f, line 54: Not sure we need this timeout if we retry with a timeout below?

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,50 @@ type prometheusResponseData struct {
...
+			return true, nil
+		}
+
+		if resp != nil {
+			resp.Body.Close()
+		}
+
+		return false, nil
+	}
+	err = wait.PollUntilContextTimeout(ctx, 10*time.Second, 60*time.Second, true, condition)
```
jan--f, line 80: I'm not super familiar with golang http conventions but since we create our own context should we not use a `context.WithTimeout(ctx, timeout)` and `PollUntilContextCancel()`?

raptorsun: they are equivalent according to the document of `PollUntilContextTimeout` https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout

test/extended/util/prometheus/helpers.go
raptorsun: `PollUntilContextTimeout` execute the condition function after every interval has passed, until the timeout. I put the same timeout for the HTTP client as the interval of calling the condition function. So that by the time the second HTTP request is sent, the first has timed out already.

test/extended/prometheus/prometheus.go
```diff
@@ -407,7 +407,7 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 			// expect all endpoints within 60 seconds
 			var lastErrs []error
 			o.Expect(wait.PollImmediate(10*time.Second, 2*time.Minute, func() (bool, error) {
-				contents, err := getBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
+				contents, err := helper.GetBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
```
simonpasquier, line 410: since we're at it, could you replace the fmt.Sprintf() call by path.Join()?

test/extended/prometheus/prometheus.go
```diff
@@ -456,7 +456,7 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 
 			g.By("verifying all targets are exposing metrics over secure channel")
 			var insecureTargets []error
-			contents, err := getBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
+			contents, err := helper.GetBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
```
simonpasquier, line 459: same here

test/extended/prometheus/prometheus.go
```diff
@@ -580,7 +580,7 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 		g.It("should provide ingress metrics", func() {
 			var lastErrs []error
 			o.Expect(wait.PollImmediate(10*time.Second, 4*time.Minute, func() (bool, error) {
-				contents, err := getBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
+				contents, err := helper.GetBearerTokenURL(fmt.Sprintf("%s/api/v1/targets", prometheusURL), bearerToken)
```
simonpasquier, line 583: same here

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,50 @@ type prometheusResponseData struct {
...
+
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
```
simonpasquier, line 87: I'd also move the io.ReadAll call to the retry loop.

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,45 @@ type prometheusResponseData struct {
...
+	}
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return "", err
+	}
+
+	req.Header.Add("Authorization", "Bearer "+bearerToken)
+
+	var body []byte
```
simonpasquier, line 65: we need a variable to store the last encountered error.
```suggestion
	var (
		body []byte
		lastErr error
	)
```

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,45 @@ type prometheusResponseData struct {
...
+	condition := func(ctx context.Context) (bool, error) {
+		resp, err := client.Do(req)
+		if err != nil {
+			return false, err
+		}
```
simonpasquier, line 70:
```suggestion
		resp, err := client.Do(req)
		if err != nil {
			lastErr = fmt.Errorf"%s: %w", url: err)
			return false, nil
		}
		defer resp.Body.Close()
```

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,45 @@ type prometheusResponseData struct {
...
+
+		if resp.StatusCode != http.StatusOK {
+			resp.Body.Close()
+			return false, err
+		}
```
simonpasquier, line 75:
```suggestion
		if resp.StatusCode != http.StatusOK {
			lastErr = fmt.Errorf("%s: unexpected status code: %d", url, resp.StatusCode)
			return false, nil
		}
```

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,45 @@ type prometheusResponseData struct {
...
+		body, err = io.ReadAll(resp.Body)
+		resp.Body.Close()
+
+		return err == nil, err
```
simonpasquier, line 79:
```suggestion
		body, err = io.ReadAll(resp.Body)
		if err != nil {
			lastErr = fmt.Errorf("%s: failed to read response: %w", url, err)
			return false, nil
		}

		return true, nil
```

test/extended/util/prometheus/helpers.go
```diff
@@ -286,18 +319,26 @@ func RunQueries(ctx context.Context, prometheusClient prometheusv1.API, promQuer
...
+
 	for _, statusCode := range statusCodes {
-		if string(output) == strconv.Itoa(statusCode) {
+		if resp.StatusCode == statusCode {
 			return nil
 		}
 	}
-
-	return fmt.Errorf("last response from server was not in %v: %s", statusCodes, output)
+	return fmt.Errorf("last response from server was not in %v: %d", statusCodes, resp.StatusCode)
```
simonpasquier, line 341: (nit) while we are at it
```suggestion
	return fmt.Errorf("%s: last response from server was not in %v: %d", url, statusCodes, resp.StatusCode)
```

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,45 @@ type prometheusResponseData struct {
...
+			resp.Body.Close()
+			return false, err
+		}
+		body, err = io.ReadAll(resp.Body)
+		resp.Body.Close()
+
+		return err == nil, err
+	}
+	err = wait.PollUntilContextTimeout(context.Background(), 10*time.Second, 60*time.Second, true, condition)
+
```
simonpasquier, line 82:
```suggestion
```

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,45 @@ type prometheusResponseData struct {
...
 	if err != nil {
-		return "", fmt.Errorf("host command failed: %v\n%s", err, output)
+		return "", err
```
simonpasquier, line 84:
```suggestion
		return "", fmt.Errorf("%w: %w", err, lastErr)
```

raptorsun:
Thanks for the review :D @simonpasquier 
The PR has been updated accordingly. The deprecated function `wait.PollImmediate` is replaced with `wait.PollUntilContextTimeout`, too.

raptorsun:
/retest-required

Le mar. 31 oct. 2023, 19:48, openshift-ci[bot] ***@***.***> a
écrit :

simonpasquier:
I suspect that you'll have to create the http.Request in the condition function.

raptorsun:
Thanks for the quick review :D @simonpasquier 
The problem comes from the `path.Join` function, it cut the `https://xxx` to `https:/xxx` by eliminating a slash, causing the HTTP request fails.

test/extended/util/prometheus/helpers.go
```diff
@@ -444,3 +498,9 @@ func ForEachAlertingRule(rules map[string][]promv1.AlertingRule, f func(a promv1
 
 	return fmt.Errorf("Incompliant rules detected:\n\n%s", strings.Join(allViolations.List(), "\n"))
 }
+
+// url.JoinPath but with no error checks
+func JoinUrl(base string, paths ...string) string {
+	path, _ := url.JoinPath(base, paths...)
```
simonpasquier, line 504: For safety, it would better to panic if an error is returned (which should never happen)
```suggestion
	path, err := url.JoinPath(base, paths...)
	if err != nil {
		panic(err)
	}
```

raptorsun: Right, like a `throw()` :D

test/extended/util/prometheus/helpers.go
```diff
@@ -444,3 +498,12 @@ func ForEachAlertingRule(rules map[string][]promv1.AlertingRule, f func(a promv1
 
 	return fmt.Errorf("Incompliant rules detected:\n\n%s", strings.Join(allViolations.List(), "\n"))
 }
+
+// url.JoinPath but with no error checks
+func JoinUrl(base string, paths ...string) string {
```
simonpasquier, line 503: (nit)
```suggestion
// MustJoinUrlPath behaves like url.JoinPath but it will panic in case of error.
func MustJoinUrlPath(base string, paths ...string) string {
```

```suggestion
func MustJoinUrlPath(base string, paths ...string) string {
```

raptorsun:
Thank you for the timely review. @simonpasquier 
PR updated accordingly :)

test/extended/util/prometheus/helpers.go
```diff
@@ -45,13 +46,51 @@ type prometheusResponseData struct {
...
+	req.Header.Add("Authorization", "Bearer "+bearerToken)
+
+	var (
+		body    []byte
+		lastErr error
+	)
+	condition := func(ctx context.Context) (bool, error) {
+		resp, err := client.Do(req)
+		if err != nil {
+			return false, err
```
simonpasquier, line 72: we missed here that the error should be captured in `lastErr` and the function should return `false, nil`.

PR #28331, opened 2023-10-17
Extract the cluster operator name from Down/Degraded Alerts

dgoodwin:
This will now appear in the locator and help with analysis about which
ClusterOperator is down/degraded. It may also prevent overlap problems
we could be experiencing when more than one is down.

[TRT-1235](https://issues.redhat.com//browse/TRT-1235)

### DIFF
--- a/pkg/monitortests/testframework/alertanalyzer/alerts.go
+++ b/pkg/monitortests/testframework/alertanalyzer/alerts.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	prometheustypes "github.com/prometheus/common/model"
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


pkg/monitortests/testframework/alertanalyzer/alerts.go
```diff
@@ -364,10 +358,7 @@ func createEventIntervalsForAlerts(ctx context.Context, alerts prometheustypes.V
 		}
 
 	default:
-		ret = append(ret, monitorapi.NewInterval(monitorapi.SourceAlert, monitorapi.Error).
-			Locator(monitorapi.NewLocator().AlertFromNames("all", "", "", "", "")).
-			Message(monitorapi.NewMessage().HumanMessagef("unhandled type: %v", alerts.Type())).
-			Build(startTime, time.Now()))
+		logrus.WithField("type", alerts.Type()).Warning("unhandled prometheus alert type received in alert monitor")
```
dgoodwin, line 361: Not sure what this was for, it doesn't seem to happen. Switching to a log message we can hunt for in search.ci rather than injecting an interval with little to no data.

PR #28337, opened 2023-10-19
OCPBUGS-21600: ignore alert during ci run as it can be flaky

kannon92:
This alert is not very useful for CI runs as sometimes high memory limits can be normal.  

We are going to exclude this from the list of alerts that we are concerned about.

### DIFF
--- a/pkg/alerts/allowed_conformance.go
+++ b/pkg/alerts/allowed_conformance.go
@@ -33,6 +33,10 @@ func AllowedAlertsDuringConformance(featureSet configv1.FeatureSet) (allowedFiri
 			Selector: map[string]string{"alertname": "ExtremelyHighIndividualControlPlaneCPU"},
 			Text:     "high CPU utilization during e2e runs is normal",
 		},
+		{
+			Selector: map[string]string{"alertname": "SystemMemoryExceedsReservation"},
+			Text:     "high memory utilization during e2e runs is normal",
+		},
 	}
 	pendingAlertsWithBugs := helper.MetricConditions{}
 	allowedPendingAlerts := helper.MetricConditions{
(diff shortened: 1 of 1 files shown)


harche (commented):
/lgtm

kannon92:
@spadgett or @bparees would either of you be able to look this over?

bparees:
would like TRT to weigh in on this before we drop it as i'm not sure if this is an alert they are working on driving down:
https://redhat-internal.slack.com/archives/C01CQA76KMX/p1698246527808839

dgoodwin:
This alert is extremely rare: https://grafana-loki.ci.openshift.org/d/PeFPT7XVk/ci-cluster-alerts?orgId=1&var-platform=aws&var-platform=azure&var-platform=gcp&var-platform=metal&var-platform=vsphere&var-percentile=P95&var-alerts=SystemMemoryExceedsReservation&var-namespaces=All&var-upgrade_type=micro&var-upgrade_type=minor&var-upgrade_type=none&var-networks=sdn&var-networks=ovn&var-topologies=ha&var-architectures=amd64&var-min_job_runs=1&var-lookback=1&var-level=Warning&var-level=Critical&var-min_firing_seconds=1&var-releases=4.14&var-releases=4.15

This shows 7 jobs hit this alert in the last month across all of 4.14 and 4.15. The P95 of all clusters is 0s. 

The job linked in the bug is from 4.11, our data doesn't go back that far, but even in search.ci it doesn't appear to have occurred in the last week other than two hits in an openstack job: https://search.ci.openshift.org/?search=alert+SystemMemoryExceedsReservation+fired+for&maxAge=168h&context=1&type=junit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

Alerts are something we're trying to get tighter testing on as a result of a lot of concerns raised with SD. It looks very rare, and ignoring globally could hide something we should be highlighting. Apologies but I don't think we should merge this, rather get the jobs the memory they need where this is failing. These kind of things can have an impact on other tests from what we've seen, and upping the memory could help more than is apparent.

dgoodwin:
Thanks!

PR #28343, opened 2023-10-20
trt-1320: update trt approvers

neisw:


dgoodwin:
All TRT members have made great contributions to origin and are responsible/cautious with reviews/approvals, and frankly we need more help here.

/approve
/lgtm

PR #28347, opened 2023-10-23
OCPBUGS-22276: Remove all docker.io images due to access denied

sdodson:
Observed problem is in OKD jobs of which there are none on this repo. So I'm interested in seeing at least that all the OCP jobs continue working without these image streams / tags. Then will follow up in confirming that things are fixed via rehearsals here https://github.com/openshift/release/pull/44127

### DIFF
--- a/examples/image-streams/image-streams-centos7.json
+++ b/examples/image-streams/image-streams-centos7.json
@@ -195,28 +195,6 @@
             }
           },
           {
-            "name": "2.4-el8",
-            "annotations": {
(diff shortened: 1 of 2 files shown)


wewang58:
many jobs failed, could you help to check need to update others? thanks

bparees:
@dperaza4dustbit @coreydaley heads up in case any of these are used by the image ecosystem or builds jobs.  If they are, we should update the jobs to stop using them and move to the non-docker.io hosted versions.

/approve
/lgtm

bparees:
actually looks like image ecosystem ran on this PR and passed, so we're on good shape there

PR #28350, opened 2023-10-24
port node lifecycle intervals

dgoodwin:
- Restore Node NotReady intervals
- Fix intervals mistakenly showing mass node NotReady


dgoodwin:
/close

Somehow I duped this.

https://github.com/openshift/origin/pull/28351

PR #28353, opened 2023-10-24
OCPBUGS-22358: fix: increase upper bounds for samples operator

eggfoobar:
Samples Operator added another watching triggerig the upper bound for SNO.

The upper limit was calculated to 52 via curl command below, increasing to 27*2 to buffer the change.

```
curl -L https://search.ci.openshift.org/\?search\=.\*Operator.\*cluster-samples.\*produces+more+watch+requests+than+expected\&maxAge\=168h\&context\=1\&type\=junit\&name\=aws.\*single-node\&excludeName\=\&maxMatches\=5\&maxBytes\=20971520\&groupBy\=job | grep watchrequestcount= | cut -d= -f2 | cut -d, -f1
```


dgoodwin:
Apologies for this test, we're trying to get it into a more sustainable form but it's a big change.

/lgtm

PR #28354, opened 2023-10-25
Restore APIServer graceful shutdown windows

dgoodwin:
Discovered these were mistakenly using an incorrect source
classification, so this is changed to be apiserver specific.

Also restore them on the intervals chart and collapse into one group.


dgoodwin:
APIServer shutdown intervals are back: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/28354/pull-ci-openshift-origin-master-e2e-gcp-ovn-rt-upgrade/1717208355262435328

PR #28357, opened 2023-10-25
OCPBUGS-22413: Use Centos 8 Stream mysql image in tests

vrutkovs:
This test frequently fails on OKD, as it is importing [mysql samples](https://github.com/openshift/cluster-samples-operator/blob/master/assets/operator/okd-x86_64/mysql/imagestreams/mysql-centos.json) from quay.io. This PR updates used images to explicitly import mysql based on CentOS 8 Stream image instead of relying on imagestreams imported by cluster-samples-operator

### DIFF
--- a/examples/image-streams/image-streams-centos7.json
+++ b/examples/image-streams/image-streams-centos7.json
@@ -500,6 +500,26 @@
             }
           },
           {
+            "name": "8.0-el8",
+            "annotations": {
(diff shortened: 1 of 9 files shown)


sdodson:
Thanks, I'd encountered that as well when trying to expand OKD-SCOS tests, wasn't sure where to go.
/lgtm

vrutkovs:
OKD test are using 4.13 version for now (https://github.com/openshift/release/pull/44887 would fix that)

soltysh (approved):
/lgtm
/approve

PR #28358, opened 2023-10-26
Skip FailedScheduling intervals when masters are in NodeUpdate

dgoodwin:
These events are expected for certain pods during master updates, sometimes if the scheduler retries enough times they can trip our 20x limit and be considered a pathological event, failing tests when in reality this is a fairly normal condition.

I did some more work on the structured intervals so we could detect the NodeUpdate intervals in a structured manner.

[TRT-1335](https://issues.redhat.com//browse/TRT-1335)

### DIFF
--- a/pkg/cmd/openshift-tests/monitor/timeline/timeline_command.go
+++ b/pkg/cmd/openshift-tests/monitor/timeline/timeline_command.go
@@ -247,7 +247,7 @@ func renderHTML(events monitorapi.Intervals) ([]byte, error) {
 		return nil, err
 
 	}
-	e2eChartTemplate := testdata.MustAsset("e2echart/non-spyglass-e2e-chart-template.html")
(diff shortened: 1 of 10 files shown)


pkg/cmd/openshift-tests/monitor/timeline/timeline_command.go
```diff
@@ -247,7 +247,7 @@ func renderHTML(events monitorapi.Intervals) ([]byte, error) {
 		return nil, err
 
 	}
-	e2eChartTemplate := testdata.MustAsset("e2echart/non-spyglass-e2e-chart-template.html")
+	e2eChartTemplate := testdata.MustAsset("e2echart/e2e-chart-template.html")
```
dgoodwin, line 250: This command is not used anywhere other than by devs as far as I can see. I switched it to default to the template we look at most often.

PR #28361, opened 2023-10-27
MON-3449: Ensure monitoring statefulsets are not recreated after upgr…

machine424:
…ades.

To recreate a statefulset (to change an immutable field e.g.), prometheus-operator will use foreground deletion which leads to downtime (Prometheus startup process might take minutes, depending on the amount of WAL to replay).

It's also better to avoit statefulset recreation as this may not be handeled well in some cases, leading to stuck Pods and thus more downtime.

We try as much as possible to avoid the changes that require statefulsets recreation, as we can in most cases do without them, but detecting them is not an easy task given that the statefulets are created in runtime by prometheus-operator.

Because "avoid statefulsets recreation" is the norm, it's better to have a safeguard upgrade test for it. Of course if for a certain upgrade we cannot by any means avoid the recreation of a statefulset, we can temporarily exclude it or skip the test.

### DIFF
--- a/pkg/monitortests/monitoring/statefulsetsrecreation/monitortest.go
+++ b/pkg/monitortests/monitoring/statefulsetsrecreation/monitortest.go
@@ -1,0 +1,170 @@
+package statefulsetsrecreation
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/monitortestframework"
+	"github.com/openshift/origin/pkg/monitortestlibrary/platformidentification"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	openshiftMonitoringNs = "openshift-monitoring"
+	testName              = "[sig-instrumentation] Monitoring statefulsets are not recreated after upgrade"
+)
+
+var statefulsetsToCheck = []string{"prometheus-k8s", "alertmanager-main"}
+
+type statefulsetsChecker struct {
+	statefulsetsUID map[string]string
+	kubeClient      kubernetes.Interface
+}
+
+// NewStatefulsetsChecker makes sure that some statefulsets are not recreated
+// after the upgrade to avoid downtime.
+func NewStatefulsetsChecker() monitortestframework.MonitorTest {
+	return &statefulsetsChecker{
+		statefulsetsUID: make(map[string]string),
+	}
+}
+
+func (sc *statefulsetsChecker) StartCollection(
+	ctx context.Context,
+	adminRESTConfig *rest.Config,
+	recorder monitorapi.RecorderWriter,
+) error {
+	var err error
+
+	sc.kubeClient, err = kubernetes.NewForConfig(adminRESTConfig)
+	if err != nil {
+		return err
+	}
+	sc.statefulsetsUID, err = sc.getStatefulsetsUID(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (sc *statefulsetsChecker) CollectData(
+	ctx context.Context,
+	storageDir string,
+	beginning, end time.Time,
+) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error) {
+	return nil, nil, nil
+}
+
+func (sc *statefulsetsChecker) ConstructComputedIntervals(
+	ctx context.Context,
+	startingIntervals monitorapi.Intervals,
+	recordedResources monitorapi.ResourcesMap,
+	beginning, end time.Time,
+) (monitorapi.Intervals, error) {
+	return nil, nil
+}
+
+func (sc *statefulsetsChecker) EvaluateTestsFromConstructedIntervals(
+	ctx context.Context,
+	finalIntervals monitorapi.Intervals,
+) ([]*junitapi.JUnitTestCase, error) {
+	if !platformidentification.DidUpgradeHappenDuringCollection(
+		finalIntervals,
+		time.Time{},
+		time.Time{},
+	) {
+		return nil, nil
+	}
+
+	currentStatefulsetsUID, err := sc.getStatefulsetsUID(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get the monitoring statefulsets UID: %v", err)
+	}
+	if len(currentStatefulsetsUID) != len(sc.statefulsetsUID) {
+		return nil, fmt.Errorf(
+			"got %d monitoring statefulsets' UID before but %d now, the mismatch means the test is broken",
+			len(sc.statefulsetsUID),
+			len(currentStatefulsetsUID),
+		)
+	}
+	for sts, initialUID := range sc.statefulsetsUID {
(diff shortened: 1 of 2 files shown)


test/extended/prometheus/statefulsets_stability.go
```diff
@@ -0,0 +1,60 @@
...
+	"reflect"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/upgrades"
+)
+
+// TODO: add other statefulsets
```
machine424, line 14: here

test/extended/prometheus/statefulsets_stability.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (t *StatefulsetsStabilityAfterUpgradeTest) Skip(_ upgrades.UpgradeContext) bool {
+	return false
+}
+
+func getStatefulsetsUID(ctx context.Context, f *e2e.Framework) map[string]string {
+	statefulsetsUID := make(map[string]string)
+	for _, statefulsetName := range statefulsetsToCheck {
+		statefulset, err := f.ClientSet.AppsV1().StatefulSets("openshift-monitoring").Get(ctx, statefulsetName, metav1.GetOptions{})
+		// TODO: maybe retry?
```
machine424, line 55: here

test/extended/prometheus/statefulsets_stability.go
```diff
@@ -0,0 +1,60 @@
+package prometheus
```
machine424, line 1: I assume this will also apply to old upgrades that we still test, don't know if we had statefulsets recreation back then.

test/e2e/upgrade/upgrade.go
```diff
@@ -65,6 +65,7 @@ func AllTests() []upgrades.Test {
 		&apps.DaemonSetUpgradeTest{},
 		&prometheus.ImagePullsAreFast{},
 		&prometheus.MetricsAvailableAfterUpgradeTest{},
+		&prometheus.StatefulsetsStabilityAfterUpgradeTest{},
```
machine424, line 68: maybe another name?

test/extended/prometheus/statefulsets_stability.go
```diff
@@ -0,0 +1,72 @@
...
+	ctx context.Context,
+	f *e2e.Framework,
+	done <-chan struct{},
+	upgrade upgrades.UpgradeType,
+) {
+	<-done
+
+	g.By("verifying that the statefulsets UID stay the same")
+	require.ElementsMatch(g.GinkgoT(), t.statefulsetsUID, getStatefulsetsUID(ctx, f))
+	require.ElementsMatch(
```
machine424, line 46: to be removed, just to make sure the error is clear when it fails

test/extended/prometheus/statefulsets_stability.go
simonpasquier: The latter I believe though it might not be true for jobs that test multiple upgrades (like https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.11-to-4.12-to-4.13-to-4.14-ci)? :thinking:

test/extended/prometheus/statefulsets_stability.go
```diff
@@ -0,0 +1,96 @@
...
+			ctx,
+			time.Second,
+			1*time.Minute,
+			true,
+			func(context.Context) (bool, error) {
+				statefulset, err := f.ClientSet.AppsV1().
+					StatefulSets(openshiftMonitoringNs).
+					Get(ctx, statefulsetName, metav1.GetOptions{})
+				if err != nil {
+					return false, err
```
simonpasquier, line 79: we need to save the error and return "false, nil" instead otherwise any Get error will stop the loop.

test/extended/prometheus/statefulsets_stability.go
jan--f: Maybe in `Skip()` below you can inspect the UpgradeContext and skip accordingly?

machine424: Good idea Jan, we can get that from `upgCtx.Versions[*].Version`, but I will ask around to make sure we really need this.

test/extended/prometheus/statefulsets_stability.go
machine424: removed, the code will not panic and the wrapping node will fail

test/extended/prometheus/statefulsets_stability.go
machine424: I did some digging, IIUC the whole upgrade chain is considered as a single disruption, so we only run upgrade checks at the end:

https://github.com/openshift/origin/blob/c89105f8d6f67e087f4a06abba4122d9ea46ab3d/test/e2e/upgrade/upgrade.go#L178-L192

and 

https://github.com/openshift/origin/blob/c89105f8d6f67e087f4a06abba4122d9ea46ab3d/test/extended/util/disruption/disruption.go#L75-L90

(if we don't backport, the test will only run when the final upgrade version is >=4.15)

maybe @soltysh can confirm (or decline)

machine424:
/test e2e-gcp-ovn-upgrade
/ retest

simonpasquier:
> (if we don't backport, the test will only run when the final upgrade version is >=4.15)

I don't see that we need to backport the change so this is fine by me.

jan--f:
> >  (if we don't backport, the test will only run when the final upgrade version is >=4.15)

> I don't see that we need to backport the change so this is fine by me.

Seconded.  This looks good to me.

dgoodwin:
Just wanted to pass along that there's a different way of defining tests of this type that are attempting to monitor the cluster throughout e2e testing: https://github.com/openshift/origin/blob/master/pkg/monitortestframework/types.go#L39

You could create one of these monitor tests, establish a kube watch in the StartCollection, and write the logic that no changes of any kind are seen for the StatefulSets you're interested in. Junit results could be returned from CollectData or EvaluateTestsFromConstructedIntervals.  Your could have it only return a failure if you see your StatefulSet was created at some point to accommodate any scenarios related to upgrades from old versions. 

I believe they are registered here: https://github.com/openshift/origin/blob/master/pkg/defaultmonitortests/types.go

This framework exists in 4.14 and should backport.

Not sure if this is helpful for your scenario but we thought we should mention it as it looked like it might be helpful.

pkg/monitortests/monitoring/statefulsetsrecreation/monitortest.go
```diff
@@ -0,0 +1,165 @@
...
+	beginning, end time.Time,
+) (monitorapi.Intervals, error) {
+	return nil, nil
+}
+
+func (sc *statefulsetsChecker) EvaluateTestsFromConstructedIntervals(
+	ctx context.Context,
+	finalIntervals monitorapi.Intervals,
+) ([]*junitapi.JUnitTestCase, error) {
+	if !platformidentification.DidUpgradeHappenDuringCollection(
```
machine424, line 82: I have the same question than in https://github.com/openshift/origin/pull/28361#discussion_r1380372286
(once merged WITHOUT backport, would this test only be triggered on a >=4.15 or could it be run on older versions?)

machine424:
I added a version with testmonitors, tell me which one you prefer.

jan--f:
I think I do like the monitortests approach. This seems more flexible generally speaking. I think we expect the sts's not to change in most scenarios, not just upgrades. And even if, this approach should allow us to specify exceptions. 
Also in the future, when the sts's don't exist we should be able to write sensible exceptions into this test.
Nice work @machine424 :+1:

machine424:
> I think I do like the monitortests approach. This seems more flexible generally speaking. I think we expect the sts's not to change in most scenarios, not just upgrades. And even if, this approach should allow us to specify exceptions.
Also in the future, when the sts's don't exist we should be able to write sensible exceptions into this test.

Thanks @jan--f I removed the other version, let's see.

PR #28362, opened 2023-10-27
skip revisioned checks for ownership

deads2k:
proof of https://github.com/openshift/library-go/pull/1596

Don't wait other pulls on this.  Need to gather the data first and then I'll re-run the generator to tidy up our list.


deads2k:
/hold

until its a real bump

PR #28364, opened 2023-10-30
Extract interval when multus restarts and chart all journal intervals

dgoodwin:
OCPBUGS-20479


dgoodwin:
/close

This didn't work out.

PR #28366, opened 2023-10-31
OCPBUGS-20479: Add pod sandbox failures to e2e charts

dgoodwin:
A little hacky until we revisit this for structured intervals, but
should be possible to clean up soon.

OCPBUGS-20479

### DIFF
--- a/pkg/monitor/serialization/serialize.go
+++ b/pkg/monitor/serialization/serialize.go
@@ -144,7 +144,7 @@ func IntervalsToFile(filename string, intervals monitorapi.Intervals) error {
 func EventsIntervalsToJSON(events monitorapi.Intervals) ([]byte, error) {
 	outputEvents := []EventInterval{}
 	for _, curr := range events {
-		if !curr.Display || (curr.From == curr.To && !curr.To.IsZero()) {
+		if curr.From == curr.To && !curr.To.IsZero() {
 			continue
 		}
 
--- a/pkg/monitor/serialization/serialize.go
+++ b/pkg/monitor/serialization/serialize.go
@@ -144,9 +144,10 @@ func IntervalsToFile(filename string, intervals monitorapi.Intervals) error {
 func EventsIntervalsToJSON(events monitorapi.Intervals) ([]byte, error) {
 	outputEvents := []EventInterval{}
 	for _, curr := range events {
-		if curr.From == curr.To && !curr.To.IsZero() {
+		if !curr.Display && curr.From == curr.To && !curr.To.IsZero() {
 			continue
 		}
+
 		outputEvents = append(outputEvents, monitorEventIntervalToEventInterval(curr))
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/monitor/serialization/serialize.go
```diff
@@ -144,9 +144,10 @@ func IntervalsToFile(filename string, intervals monitorapi.Intervals) error {
 func EventsIntervalsToJSON(events monitorapi.Intervals) ([]byte, error) {
 	outputEvents := []EventInterval{}
 	for _, curr := range events {
-		if curr.From == curr.To && !curr.To.IsZero() {
+		if !curr.Display && curr.From == curr.To && !curr.To.IsZero() {
```
neisw, line 147: Double checking.  Seems like we would want !curr.Display ||( curr.From == curr.To && !curr.To.IsZero() )

Currently this will allow 0 length intervals that don't have curr.Display set to false, is that the intention?

dgoodwin: Nice, yes I think you're right, thanks!

dgoodwin: pushed

pkg/monitortests/network/legacynetworkmonitortests/networking.go
```diff
@@ -92,10 +93,33 @@ func testPodSandboxCreation(events monitorapi.Intervals, clientConfig *rest.Conf
...
+
 		if !strings.Contains(event.Message, "reason/FailedCreatePodSandBox Failed to create pod sandbox") {
 			continue
 		}
+
+		// Skip pod sandbox failures when nodes are updating
+		var foundOverlap bool
+		for _, nui := range nodeNotReadyIntervals {
+			if nui.From.Before(event.From) && nui.To.After(event.To) {
+				logrus.Infof("%s was found to overlap with %s, ignoring pod sandbox error as we expect these if the node is NotReady", event, nui)
```
neisw, line 114: I'm thinking we are good but wondering if we want / need to consider partial overlap where either from and to are the same or potentially nui.to is after event.From but before event.to?  Probably not too likely so I  could go either way.

dgoodwin: Lets play by ear once this is in the wild, we might want to open it up more once we see some data and have the charts finally.

pkg/monitor/serialization/serialize.go
```diff
@@ -144,9 +144,10 @@ func IntervalsToFile(filename string, intervals monitorapi.Intervals) error {
 func EventsIntervalsToJSON(events monitorapi.Intervals) ([]byte, error) {
 	outputEvents := []EventInterval{}
 	for _, curr := range events {
-		if curr.From == curr.To && !curr.To.IsZero() {
+		if !curr.Display || (curr.From == curr.To && !curr.To.IsZero()) {
 			continue
 		}
+
```
dgoodwin, line 150: This logic is still wrong. Unported events will be Display = false by default due to go zero values. I think I need to remove Display here entirely, we keep the curr.From and curr.To check.

dgoodwin:
Looks better now, we'll keep an eye out for what's going on in the merged code, but this Display thing was a total bug in this PR. It wasn't required as I adjusted the +1s for these events as we do for pathologicals/interesting events.

PR #28372, opened 2023-10-31
"TRT-1342: Trim the intervals for loki serializer to avoid errors during ingest"

DennisPeriquet:
[TRT-1342](https://issues.redhat.com//browse/TRT-1342)

With this PR, ...

* I [revert my earlier PR](https://github.com/openshift/origin/pull/28372/commits/a62bcf00c40db061535eb859ad218756544def2e) so the second spyglass chart (for upgrade jobs) goes back to getting some intervals that started during the upgrade phase (i.e., we go back to the previous `Slice()` algorithm).
* For loki serialization/ingestion, only intervals that happened after the upgrade will be pushed (this will avoid the "entry too far behind" problem that happens when we push intervals that are too old).

### DIFF
--- a/pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
+++ b/pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -25,6 +26,7 @@ import (
 )
 
 type lokiSerializer struct {
+	monitorStartTime time.Time
 }
 
 func NewUploadSerializer() monitortestframework.MonitorTest {
@@ -32,6 +34,14 @@ func NewUploadSerializer() monitortestframework.MonitorTest {
 }
 
 func (w *lokiSerializer) StartCollection(ctx context.Context, adminRESTConfig *rest.Config, recorder monitorapi.RecorderWriter) error {
+	// Immediately before the call to monitorTestRegistry.StartCollection(),
+	// we get the current time and use that as the Monitor.startTime.
+	// MonitorTestRegistry.StartCollection calls startCollectionWithPanicProtection
+	// for each monitorTest which calls monitorTest.StartCollection, which is when
+	// this method is called. As such, our call to time.Now() captures the time when
+	// the data collection starts and is about the same time as the monitor's
+	// startTime.
+	w.monitorStartTime = time.Now()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/monitortestframework/impl.go
```diff
@@ -270,15 +271,30 @@ func (r *monitorTestRegistry) WriteContentToStorage(ctx context.Context, storage
 
 		start := time.Now()
 
-		var finalIntervalLength = len(finalIntervals)
+		// For the loki flavor of monitorTests, remove older intervals (e.g., from the upgrade)
+		// (to avoid the "entry too far behind" errors that happen when pushing intervals to
+		// loki that are too old).
+		intervalsToUse := finalIntervals
+		if monitorTest.name == "upload-to-loki-serializer" {
```
dgoodwin, line 278: This should be in the monitor test itself, not up here with a name comparison. Look to that tests WriteContentToStorage function, which is what we end up calling below. 

I think you could possibly record the start time in StartCollection there, if there's not a better way. I don't think we'd want to change the signature and pass start time to all WriteContentToStorage functions.

pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
```diff
@@ -25,13 +26,15 @@ import (
...
 type lokiSerializer struct {
+	monitorStartTime time.Time
 }
 
 func NewUploadSerializer() monitortestframework.MonitorTest {
 	return &lokiSerializer{}
 }
 
 func (w *lokiSerializer) StartCollection(ctx context.Context, adminRESTConfig *rest.Config, recorder monitorapi.RecorderWriter) error {
+	w.monitorStartTime = time.Now()
```
DennisPeriquet, line 37: Immediately before the call to [`monitorTestRegistry.StartCollection()`](https://github.com/openshift/origin/blob/11837c1460a13e900603a1dbafaea6525276aabf/pkg/monitor/monitor.go#L65), we get the current time and use that as the `Monitor.startTime`.  `MonitorTestRegistry.StartCollection` calls `startCollectionWithPanicProtection` for each `monitorTest` which calls `monitorTest.StartCollection which is when line 37 above is called.  So, `time.Now()` is when the data collection starts and about he same time as the monitor's `startTime` which makes it so I don't have to pass it.

pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
```diff
@@ -47,9 +50,22 @@ func (*lokiSerializer) EvaluateTestsFromConstructedIntervals(ctx context.Context
...
 }
 
-func (*lokiSerializer) WriteContentToStorage(ctx context.Context, storageDir, timeSuffix string, finalIntervals monitorapi.Intervals, finalResourceState monitorapi.ResourcesMap) error {
+func (w *lokiSerializer) WriteContentToStorage(ctx context.Context, storageDir, timeSuffix string, finalIntervals monitorapi.Intervals, finalResourceState monitorapi.ResourcesMap) error {
 	fmt.Fprintf(os.Stderr, "Uploading to loki.\n")
-	if err := UploadIntervalsToLoki(finalIntervals); err != nil {
+
+	// For the loki flavor of monitorTests, remove older intervals (e.g., from the upgrade)
+	// (to avoid the "entry too far behind" errors that happen when pushing intervals to
+	// loki that are too old).
```
dgoodwin, line 58: This is very delicate so I'd like to be precise with this comment, could you clarify that some monitor tests can return intervals from before we started monitoring, such as in-cluster disruption pollers that scrape pod logs which are running during the upgrade and could return quite old intervals.

pkg/monitortests/testframework/uploadtolokiserializer/monitortest.go
dgoodwin: Drop this in as a comment plz, this is good stuff.

DennisPeriquet:
/retitle "TRT-1342: Trim the intervals for loki serializer to avoid errors during ingest"

PR #28376, opened 2023-11-01
TRT-1354: Add support for intervals to have row differentiators

dgoodwin:
With a goal of having intervals be produced by origin and then immediately visible in the charts, which will soon live in sippy, we need a way to inject some kind of differentiator to force rows within one section onto different lines so they do not overlap. The clearest example of this today is the node state category where we would have the same node locator for multiple rows, but we break them out so we can clearly see the phases of update, when the node was not ready, and the overall os update. 

To do this, I've added a row key to the locator, as this feels most closely related. Changes were then required in the state tracker to support this without it getting confused.

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -480,11 +480,9 @@
 
         timelineGroups.push({group: "node-state", data: []})
         createTimelineData(nodeStateValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isNodeState, regex)
+        // Sort the node-state intervals so rows are grouped by node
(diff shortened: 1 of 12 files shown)


dgoodwin:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/28376/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1721908916658376704/artifacts/e2e-gcp-ovn-upgrade/openshift-e2e-test/artifacts/junit/e2e-timelines_spyglass_20231107-161138.html shows the new node-state category working as expected.

xueqzhan:
/lgtm

Looks good. The serial job failure shouldn't have anything to do with the change.

PR #28393, opened 2023-11-10
MON-3481: Thanos use /api/v1 endpoints for authentification test.

raptorsun:
Openshift Monitoring components are replacing Oauth proxy with kube-rbac-proxy and restrict access to path under `/api/` only. By consequence, the root URL for Thanos service will no longer be accessible and return 404. 
https://issues.redhat.com/browse/MON-3379

This is for deblocking the test like `e2e-aws-ovn`. here is an example of error: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/2136/pull-ci-openshift-cluster-monitoring-operator-master-e2e-aws-ovn/1722971201866829824


simonpasquier:
/hold
Waiting for some clarifications in https://github.com/openshift/cluster-monitoring-operator/pull/2136

raptorsun:
/close 
We no longer need to limit access to Thanos querier to `/api` pathes.

PR #28395, opened 2023-11-13
Certs 11 fix certs after break

deads2k:
The logical name removal broke consumers.  We need to restore the proxy-ca eliding because it's so long and not really useful.

proof of https://github.com/openshift/library-go/pull/1617

/hold

hold until we see it work.


deads2k:
testing changes needed to gather cert info for 4.16 planning.  labelling to merge.

PR #28398, opened 2023-11-14
HOSTEDCP-1303: hs, kubevirt: add hosted cluster node restart test

maiqueb:
Assure an hosted cluster node can be restarted, and the node eventually reach the `Ready` condition.

### DIFF
--- a/test/extended/kubevirt/nodereboot.go
+++ b/test/extended/kubevirt/nodereboot.go
@@ -1,0 +1,116 @@
+package kubevirt
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = Describe("[sig-kubevirt] node reboot", func() {
+	oc := exutil.NewCLIWithPodSecurityLevel("ns-global", admissionapi.LevelBaseline)
+
+	InKubeVirtClusterContext(oc, func() {
+		mgmtFramework := e2e.NewDefaultFramework("mgmt-framework")
+		mgmtFramework.SkipNamespaceCreation = true
+
+		hostedFramework := e2e.NewDefaultFramework("hosted-framework")
+		hostedFramework.NamespacePodSecurityEnforceLevel = admissionapi.LevelPrivileged
+
+		When("an hosted control plane worker is rebooted [Early]", func() {
+			var hostedClusterNodeToBeRebooted string
+
+			simplifiedNodeReadyCondition := func(nodeName string) (corev1.NodeCondition, error) {
+				node, err := hostedFramework.ClientSet.CoreV1().Nodes().Get(
+					context.Background(),
+					nodeName,
+					metav1.GetOptions{},
+				)
+				if err != nil {
+					return corev1.NodeCondition{}, err
+				}
+
+				for _, condition := range node.Status.Conditions {
+					if condition.Type == corev1.NodeReady {
+						return corev1.NodeCondition{
+							Type:   condition.Type,
+							Status: condition.Status,
+						}, nil
+					}
+				}
+				return corev1.NodeCondition{}, nil
+			}
+
+			BeforeEach(func() {
+				nodeList, err := hostedFramework.ClientSet.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(nodeList.Items).NotTo(BeEmpty())
+
+				hostedClusterNodeToBeRebooted = nodeList.Items[0].Name
+
+				Eventually(func() (corev1.NodeCondition, error) {
+					return simplifiedNodeReadyCondition(hostedClusterNodeToBeRebooted)
+				}).
+					WithTimeout(2*time.Minute).
+					WithPolling(5*time.Second).
+					Should(
+						Equal(corev1.NodeCondition{Type: corev1.NodeReady, Status: corev1.ConditionTrue}),
+						fmt.Sprintf(
+							"node %q should have the `Ready` condition",
+							hostedClusterNodeToBeRebooted,
+						),
+					)
+
+				setMgmtFramework(mgmtFramework)
+				expectNoError(
+					rebootWorkerNode(
+						mgmtFramework,
+						hostedClusterNodeToBeRebooted,
+					),
+				)
+				Eventually(func() (corev1.NodeCondition, error) {
+					return simplifiedNodeReadyCondition(hostedClusterNodeToBeRebooted)
+				}).
+					WithTimeout(2*time.Minute).
+					WithPolling(5*time.Second).
+					Should(
+						Equal(corev1.NodeCondition{Type: corev1.NodeReady, Status: corev1.ConditionFalse}),
+						fmt.Sprintf(
+							"node %q must reach the `NotReady` condition after reboot",
+							hostedClusterNodeToBeRebooted,
+						),
+					)
+			})
+
+			It("should maintain node readiness", func() {
+				Eventually(func() (corev1.NodeCondition, error) {
+					return simplifiedNodeReadyCondition(hostedClusterNodeToBeRebooted)
+				}).
+					WithTimeout(5*time.Minute).
+					WithPolling(5*time.Second).
+					Should(
+						Equal(corev1.NodeCondition{Type: corev1.NodeReady, Status: corev1.ConditionTrue}),
+						fmt.Sprintf("node %q should have the `Ready` condition", hostedClusterNodeToBeRebooted),
+					)
+				Consistently(func() (corev1.NodeCondition, error) {
+					return simplifiedNodeReadyCondition(hostedClusterNodeToBeRebooted)
+				}).
+					WithTimeout(2*time.Minute).
+					WithPolling(5*time.Second).
+					Should(
+						Equal(corev1.NodeCondition{Type: corev1.NodeReady, Status: corev1.ConditionTrue}),
+						fmt.Sprintf("node %q should have the `Ready` condition", hostedClusterNodeToBeRebooted),
+					)
+			})
+		})
+	})
+})
--- a/test/extended/kubevirt/util.go
+++ b/test/extended/kubevirt/util.go
@@ -33,6 +33,7 @@ import (
 )
 
 var (
+	vmGVR   = schema.GroupVersionResource{Group: "kubevirt.io", Resource: "virtualmachines", Version: "v1"}
 	vmiGVR  = schema.GroupVersionResource{Group: "kubevirt.io", Resource: "virtualmachineinstances", Version: "v1"}
 	vmimGVR = schema.GroupVersionResource{Group: "kubevirt.io", Resource: "virtualmachineinstancemigrations", Version: "v1"}
 	vmimGVK = schema.GroupVersionKind{Group: "kubevirt.io", Kind: "VirtualMachineInstanceMigration", Version: "v1"}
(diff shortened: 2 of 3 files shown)


test/extended/kubevirt/nodereboot.go
```diff
@@ -0,0 +1,85 @@
...
+	oc := exutil.NewCLIWithPodSecurityLevel("ns-global", admissionapi.LevelBaseline)
+
+	InKubeVirtClusterContext(oc, func() {
+		mgmtFramework := e2e.NewDefaultFramework("mgmt-framework")
+		mgmtFramework.SkipNamespaceCreation = true
+
+		hostedFramework := e2e.NewDefaultFramework("hosted-framework")
+		hostedFramework.NamespacePodSecurityEnforceLevel = admissionapi.LevelPrivileged
+
+		When("an hosted control plane worker is rebooted", func() {
```
qinqon, line 30:
```suggestion
		When("a hosted control plane worker is rebooted", func() {
```

test/extended/kubevirt/nodereboot.go
```diff
@@ -0,0 +1,85 @@
...
+			})
+
+			It("should maintain node readiness", func() {
+				Eventually(func() (bool, error) { return isNodeReady(hostedClusterNodeToBeRebooted.Name) }).
+					WithTimeout(2*time.Minute).
+					WithPolling(5*time.Second).
+					Should(
+						BeTrue(),
+						fmt.Sprintf("node %q should have the `Ready` condition", hostedClusterNodeToBeRebooted.Name),
+					)
```
qinqon, line 74: We can have a false possitive here, since it takes some time for the node to go NotReady. 

So we should first wait until node is NotReady, then Check that goes to Ready and then check that it stay that way.

qinqon (changes requested):

test/extended/kubevirt/nodereboot.go
maiqueb: I think the h in hosted in silent, thus this should be `an`. 

... but I'm not a native speaker . 

I'll defer to one.

test/extended/kubevirt/nodereboot.go
```diff
@@ -0,0 +1,85 @@
...
+	corev1 "k8s.io/api/core/v1"
+	k8sapierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = Describe("[sig-kubevirt] node reboot", func() {
```
nunnatsa, line 20: I think we should move to label instead of text; e.g.
```suggestion
var _ = Describe("node reboot", Label("sig-kubevirt"), func() {
```

test/extended/kubevirt/nodereboot.go
```diff
@@ -0,0 +1,85 @@
...
+		hostedFramework.NamespacePodSecurityEnforceLevel = admissionapi.LevelPrivileged
+
+		When("an hosted control plane worker is rebooted", func() {
+			var hostedClusterNodeToBeRebooted corev1.Node
+
+			isNodeReady := func(nodeName string) (bool, error) {
+				node, err := hostedFramework.ClientSet.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+				if k8sapierrors.IsNotFound(err) {
+					return false, err
+				}
```
nunnatsa, line 37: What if it's another error? It will break the rest of the code.

nunnatsa (commented):
Few inline comments

test/extended/kubevirt/nodereboot.go
maiqueb: Done.

test/extended/kubevirt/nodereboot.go
maiqueb: I assume the not found node as a non-retrieable error, while others (which I pretty much don't know what they might be) as retriable.

test/extended/kubevirt/nodereboot.go
maiqueb: Makes sense - but I just followed the pattern on the other tests here.

I don't mind moving this to labels in a follow up PR.

test/extended/kubevirt/nodereboot.go
nunnatsa: I think you will get nil pointer exception in this case, and I'm pretty sure it won't be recoverable for the next loop.

test/extended/kubevirt/nodereboot.go
```diff
@@ -0,0 +1,121 @@
...
+				}).
+					WithTimeout(2*time.Minute).
+					WithPolling(5*time.Second).
+					Should(
+						BeTrue(),
+						fmt.Sprintf(
+							"node %q must reach the `NotReady` condition after reboot",
+							hostedClusterNodeToBeRebooted.Name,
+						),
+					)
```
qinqon, line 96: This should be at the "It"

qinqon (changes requested):

test/extended/kubevirt/nodereboot.go
maiqueb: Why ? AFAIU having the node in `NotReady` is a pre-requirement of the test, not part of the test itself.

Pretty much because the test is about recovering from not ready to ready after a reboot.

test/extended/kubevirt/nodereboot.go
maiqueb: could you elaborate where would I get the nil pointer ?

maiqueb: Better now ?

test/extended/kubevirt/nodereboot.go
```diff
@@ -0,0 +1,123 @@
...
+
+		When("an hosted control plane worker is rebooted", func() {
+			var hostedClusterNodeToBeRebooted corev1.Node
+
+			isNodeReadyConditionTrue := func(condition corev1.NodeCondition) bool {
+				return condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue
+			}
+
+			isNodeReadyConditionFalse := func(condition corev1.NodeCondition) bool {
+				return condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionFalse
```
maiqueb, line 38: I wonder if it must be false or if condition unknown is enough ... 
Thoughts @qinqon ?

qinqon: Both means the node is not ready

qinqon: We have to wait for NotReady since we know it goes after NodeStatusUnknown

davidvossel (commented):
Since this test is disruptive, it will need to occur in serial (no other tests executing in parallel). How is that achieved?

Also, as we're looking at testing, we need to pay close attention to the baremetalds lane. I suspect we're going to encounter issues with monitor tests failing due to the disruption of the pods on the rebooted node.

maiqueb:
> Since this test is disruptive, it will need to occur in serial (no other tests executing in parallel). How is that achieved?

I *guess* I'll have to label it w/ the serial label. I find it strange the migration tests (from which I copied some of this) do not have that tag. @qinqon is this achieved w/ this `[Early]` tag in https://github.com/openshift/origin/blob/1325dd44a7f920872ce4bd9b7f1a7806433fe3db/test/extended/kubevirt/migration.go#L43 ?

> 
> Also, as we're looking at testing, we need to pay close attention to the baremetalds lane. I suspect we're going to encounter issues with monitor tests failing due to the disruption of the pods on the rebooted node.

How can I account for that ?

qinqon:
> > Since this test is disruptive, it will need to occur in serial (no other tests executing in parallel). How is that achieved?
> 
> I _guess_ I'll have to label it w/ the serial label. I find it strange the migration tests (from which I copied some of this) do not have that tag. @qinqon is this achieved w/ this `[Early]` tag in
> 

With this we ensure that we run live-migration before anything else.

> https://github.com/openshift/origin/blob/1325dd44a7f920872ce4bd9b7f1a7806433fe3db/test/extended/kubevirt/migration.go#L43
> 
> ?
> > Also, as we're looking at testing, we need to pay close attention to the baremetalds lane. I suspect we're going to encounter issues with monitor tests failing due to the disruption of the pods on the rebooted node.
> 
> How can I account for that ?

Not sure about serial

maiqueb:
> > > Since this test is disruptive, it will need to occur in serial (no other tests executing in parallel). How is that achieved?
> > 
> > 
> > I _guess_ I'll have to label it w/ the serial label. I find it strange the migration tests (from which I copied some of this) do not have that tag. @qinqon is this achieved w/ this `[Early]` tag in
> 
> With this we ensure that we run live-migration before anything else.
> 
> > https://github.com/openshift/origin/blob/1325dd44a7f920872ce4bd9b7f1a7806433fe3db/test/extended/kubevirt/migration.go#L43
> > 
> > ?
> > > Also, as we're looking at testing, we need to pay close attention to the baremetalds lane. I suspect we're going to encounter issues with monitor tests failing due to the disruption of the pods on the rebooted node.
> > 
> > 
> > How can I account for that ?
> 
> Not sure about serial

I meant the monitor tests @davidvossel mentioned above. 

What are those ? Are they run in parallel ? If I force this test to run sequentially will it still be impacted by those monitoring tests ?

maiqueb:
> > > > Since this test is disruptive, it will need to occur in serial (no other tests executing in parallel). How is that achieved?
> > > 
> > > 
> > > I _guess_ I'll have to label it w/ the serial label. I find it strange the migration tests (from which I copied some of this) do not have that tag. @qinqon is this achieved w/ this `[Early]` tag in
> > 
> > 
> > With this we ensure that we run live-migration before anything else.
> > > https://github.com/openshift/origin/blob/1325dd44a7f920872ce4bd9b7f1a7806433fe3db/test/extended/kubevirt/migration.go#L43
> > > 
> > > ?
> > > > Also, as we're looking at testing, we need to pay close attention to the baremetalds lane. I suspect we're going to encounter issues with monitor tests failing due to the disruption of the pods on the rebooted node.
> > > 
> > > 
> > > How can I account for that ?
> > 
> > 
> > Not sure about serial
> 
> I meant the monitor tests @davidvossel mentioned above.
> 
> What are those ? Are they run in parallel ? If I force this test to run sequentially will it still be impacted by those monitoring tests ?

@qinqon ^

maiqueb:
/test e2e-aws-ovn-kubevirt

failed before starting the suite :cry:

qinqon:
> > > > > Since this test is disruptive, it will need to occur in serial (no other tests executing in parallel). How is that achieved?
> > > > 
> > > > 
> > > > I _guess_ I'll have to label it w/ the serial label. I find it strange the migration tests (from which I copied some of this) do not have that tag. @qinqon is this achieved w/ this `[Early]` tag in
> > > 
> > > 
> > > With this we ensure that we run live-migration before anything else.
> > > > https://github.com/openshift/origin/blob/1325dd44a7f920872ce4bd9b7f1a7806433fe3db/test/extended/kubevirt/migration.go#L43
> > > > 
> > > > ?
> > > > > Also, as we're looking at testing, we need to pay close attention to the baremetalds lane. I suspect we're going to encounter issues with monitor tests failing due to the disruption of the pods on the rebooted node.
> > > > 
> > > > 
> > > > How can I account for that ?
> > > 
> > > 
> > > Not sure about serial
> > 
> > 
> > I meant the monitor tests @davidvossel mentioned above.
> > What are those ? Are they run in parallel ? If I force this test to run sequentially will it still be impacted by those monitoring tests ?
> 
> @qinqon ^

They run in parallel, but I am not sure if serializing this test will prevent them to run since they are force to run in parallel.

@maiqueb are they failing at your local runs ?

maiqueb:
The test I've added is now passing in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/28398/pull-ci-openshift-origin-master-e2e-aws-ovn-kubevirt/1725089795899658240

@qinqon / @davidvossel

qinqon:
Error from odf subscription 
```
"conditions": [
                    {
                        "lastTransitionTime": "2023-11-16T18:33:14Z",
                        "message": "all available catalogsources are healthy",
                        "reason": "AllCatalogSourcesHealthy",
                        "status": "False",
                        "type": "CatalogSourcesUnhealthy"
                    },
                    {
                        "message": "constraints not satisfiable: no operators found in package odf-operator in the catalog referenced by subscription odf-operator, subscription odf-operator exists",
                        "reason": "ConstraintsNotSatisfiable",
                        "status": "True",
                        "type": "ResolutionFailed"
                    }
                ],
```

qinqon:
@maiqueb I am checking if stable-4.14 odf channel whould fix the issue https://github.com/openshift/release/pull/45796

qinqon:
presubmit baremetalds for kuebvirt job is fixed now and installing odf works as expected, let's retry.

qinqon:
The live migration tests is failing at the baremetalds lane

PR #28399, opened 2023-11-14
patho rework

dgoodwin:
- Port creation of kube event intervals to structured
- Sketch out AllowedDuplicateEvent struct, matcher func and a first test
- Clarify a misleading function name
- Match on Message not the builder which we won't have when checking real intervals
- Clarify intended use of new matchers in event watcher
- Document the AllowedDupeEvent struct
- Return full allowed dupe matcher, we'll need it to see if we should flake a test or ignore fully
- Comment clarification
- Port more pathological regexes and tests
- Refactor to a less verbose and error prone definition of allowed dupes
- More porting
- More porting
- More porting
- Port AllowOVNReadinessProbeFailure
- Comment out problematic tests for now
- Port some regexes with their own tests
- Port a slew of complex readiness/liveness regexes
- Port some NodeHas regexes
- Legacy etcd and network pathological tests passing
- Fix more tests
- ProbeError tests now passing
- Fix event monitor tests, all tests now passing
- Port remaining regexes
- Plug the new allowances into the actual framework
- Eliminate function allowances
- Complete support for single node elevated repeate event threshold
- Document more busted stuff
- Port but mostly drop wildly out of date exceptions with known bugs
- Port core dupe event evaluator logic to work again
- Remove concept of flaking if a matching bug exists on the basis this provably doesn't work
- Use structured message for an exception


pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,569 @@ package pathologicaleventlibrary
...
 
-	DuplicateEventThreshold           = 20
-	DuplicateSingleNodeEventThreshold = 30
-	PathologicalMark                  = "pathological/true"
-	InterestingMark                   = "interesting/true"
-)
+// AllowedDupeEvent allows the definition of events that can repeat more than the threshold we allow during a job run.
+// All specified fields must match the interval for it to be allowed. If or logic is required,
+// it could be implemented within an IsRepeatedEventOKFunc.
+type AllowedDupeEvent struct {
```
dgoodwin, line 39: The new one way to define an exception.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,569 @@ package pathologicaleventlibrary
...
-	Platform *v1.PlatformType
-
-	// Topology limits the exception to a specific topology (e.g. single replica)
-	Topology *v1.TopologyMode
-
-	// TestSuite limits the exception to a specific test suite (e.g. openshift/builds)
-	TestSuite *string
-}
-
-var KnownEventsBugs = []KnownProblem{
```
dgoodwin, line 192: I checked these as I went, virtually all were years old and unresolved, some became permanent exceptions, others were no longer occurring and were removed.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,569 @@ package pathologicaleventlibrary
...
-		return false, nil
-	}
-	if !strings.Contains(monitorEvent.Locator, "pod/console-") {
-		return false, nil
-	}
-	if !strings.Contains(monitorEvent.Locator, "Readiness probe") {
-		return false, nil
-	}
-	if !strings.Contains(monitorEvent.Locator, "connect: connection refused") {
-		return false, nil
```
dgoodwin, line 307: This was broken, we're checking for things in the locator that would never be in the locator, they're from the message. If it's been in this busted for ages, we don't need it. Removed.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
```diff
@@ -26,24 +26,19 @@ import (
...
-	allowedPatterns = append(allowedPatterns, AllowedUpgradeRepeatedEventPatterns...)
+	// append upgrade allowances to the main slice
+	allowedDupeEvents := []*AllowedDupeEvent{}
+	allowedDupeEvents = append(allowedDupeEvents, AllowedRepeatedEvents...)
+	allowedDupeEvents = append(allowedDupeEvents, AllowedRepeatedUpgradeEvents...)
 
 	evaluator := duplicateEventsEvaluator{
-		allowedRepeatedEventPatterns: allowedPatterns,
-		allowedRepeatedEventFns:      AllowedRepeatedEventFns,
-		knownRepeatedEventsBugs:      KnownEventsBugs,
```
dgoodwin, line 36: This was wild, but virtually everything passed to the evaluator was never used. You'll notice these variables are never read from. Somewhere along the line the logic changed to "skip if flagged interesting", implying we don't need to check these anymore. However the code flagging interesting was *assuming* to check both upgrade and non-upgrade events, so effectively we were always allowing things we intended to only allow on upgrade.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
```diff
@@ -53,11 +48,12 @@ func TestDuplicatedEventForUpgrade(events monitorapi.Intervals, kubeClientConfig
...
 	evaluator := duplicateEventsEvaluator{
-		allowedRepeatedEventPatterns: AllowedRepeatedEventPatterns,
-		allowedRepeatedEventFns:      AllowedRepeatedEventFns,
-		knownRepeatedEventsBugs:      KnownEventsBugs,
+		allowedDupeEvents: AllowedRepeatedEvents,
 	}
 
+	/* TODO: restore, but it looks like this has been busted for ages, we append to a list of
+	functions that are never used
+
```
dgoodwin, line 56: Quite busted, see above, we were appending to the list that was never used.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
```diff
@@ -67,15 +63,12 @@ func TestDuplicatedEventForStableSystem(events monitorapi.Intervals, clientConfi
...
 	}
 	evaluator.allowedRepeatedEventFns = append(evaluator.allowedRepeatedEventFns, etcdAllowance.allowEtcdGuardReadinessProbeFailure)
+	*/
 
 	if err := evaluator.getClusterInfo(clientConfig); err != nil {
 		e2e.Logf("could not fetch cluster info: %w", err)
 	}
 
-	if evaluator.topology == v1.SingleReplicaTopologyMode {
-		evaluator.allowedRepeatedEventFns = append(evaluator.allowedRepeatedEventFns, AllowedSingleNodeRepeatedEventFns...)
```
dgoodwin, line 76: Also quite busted and removed.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
```diff
@@ -325,51 +314,42 @@ func (d duplicateEventsEvaluator) testDuplicatedEvents(testName string, flakeOnl
...
+			// Check if we have an allowance for this event. This code used to just check if it had an interesting flag,
+			// implying it matches some pattern, but that happens even for upgrade patterns occurring in non-upgrade jobs,
+			// so we were ignoring patterns that were meant to be allowed only in upgrade jobs in all jobs. The list of
+			// allowed patterns passed to this object wasn't even used.
+			if allowed, _ := MatchesAny(d.allowedDupeEvents, event.StructuredLocator, event.StructuredMessage, kubeClientConfig, &d.topology); allowed {
 				continue
 			}
 
-			eventMessageString := eventDisplayMessage + " From: " + event.From.Format("15:04:05Z") + " To: " + event.To.Format("15:04:05Z")
-			if _, ok := displayToCount[eventMessageString]; !ok {
```
dgoodwin, line 337: Pretty sure this was busted, we're using eventMessageString which contains timestamps, not eventDisplayMessage, so our attempt to put unique occurences of a message into a map key was probably broken. I think you can see this in the repeating events listed when the test fails.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,569 @@ package pathologicaleventlibrary
...
-	BZ     string
-
-	// Platform limits the exception to a specific OpenShift platform.
-	Platform *v1.PlatformType
-
-	// Topology limits the exception to a specific topology (e.g. single replica)
-	Topology *v1.TopologyMode
-
-	// TestSuite limits the exception to a specific test suite (e.g. openshift/builds)
-	TestSuite *string
```
dgoodwin, line 189: TestSuite wasn't hooked up to anything, nobody read from it. (in addition to likely being used for interesting=true and then automatically allowed as the others were.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
```diff
@@ -53,27 +57,22 @@ func TestDuplicatedEventForUpgrade(events monitorapi.Intervals, kubeClientConfig
...
-	operatorClient, err := operatorv1client.NewForConfig(clientConfig)
+	etcdAllowance, err := newDuplicatedEventsAllowedWhenEtcdRevisionChange(context.TODO(), clientConfig)
 	if err != nil {
-		panic(err)
-	}
-	etcdAllowance, err := newDuplicatedEventsAllowedWhenEtcdRevisionChange(context.TODO(), operatorClient)
-	if err != nil {
-		panic(fmt.Errorf("unable to construct duplicated events allowance for etcd, err = %v", err))
-	}
-	evaluator.allowedRepeatedEventFns = append(evaluator.allowedRepeatedEventFns, etcdAllowance.allowEtcdGuardReadinessProbeFailure)
```
dgoodwin, line 69: As most of the others, this was not hooked up to anything. I see it actively failing, so this one remains. In the new framework it boils down to just a bump to the threshold. It also looks like it was never added to the upgrade suite, which seemed wrong, it now exists in both.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,545 @@ package pathologicaleventlibrary
...
+	LocatorKeyRegexes: map[monitorapi.LocatorKey]*regexp.Regexp{
+		monitorapi.LocatorNamespaceKey: regexp.MustCompile(`openshift-ovn-kubernetes`),
+		monitorapi.LocatorPodKey:       regexp.MustCompile(`ovnkube-node-`),
+	},
+	MessageReasonRegex: regexp.MustCompile(`^Unhealthy$`),
+	MessageHumanRegex:  regexp.MustCompile(`Readiness probe failed:`),
 }
 
-var AllowedSingleNodeRepeatedEventFns = []IsRepeatedEventOKFunc{
-	isConnectionRefusedOnSingleNode,
```
dgoodwin, line 276: Pretty sure this was not active, the events were already blanket allowed by: regexp.MustCompile("Unhealthy Readiness probe failed"),

dgoodwin:
This run hit an unusual pathological event failure: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/28399/pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade/1725581927794085888

```
[sig-arch] events should not repeat pathologically for ns/openshift-kube-controller-manager-operator expand_less 	0s
{  1 events happened too frequently

event happened 25 times, something is wrong: namespace/openshift-kube-controller-manager-operator deployment/kube-controller-manager-operator hmsg/893323bb66 - reason/RequiredInstallerResourcesMissing configmaps: aggregator-client-ca,client-ca, configmaps: cluster-policy-controller-config-9,config-9,controller-manager-kubeconfig-9,kube-controller-cert-syncer-kubeconfig-9,kube-controller-manager-pod-9,recycler-config-9,service-ca-9,serviceaccount-ca-9 From: 20:09:01Z To: 20:09:02Z result=reject }
```

As far as I can tell, this was not allowed by the current code in master: https://github.com/openshift/origin/blob/master/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go. The only similar pattern was specific to etcd. Searching the bigquery database, this is the only hit I see for that message going pathological, however it does appear without breaking the 20 limit in a few cases.

```sql
SELECT prowjob_name, prowjob_build_id, from_time, payload.locator, payload.message FROM `openshift-gce-devel.ci_analysis_us.job_intervals` WHERE from_time >= DATETIME_SUB(CURRENT_DATETIME(), INTERVAL 14 DAY) 
AND message like '%reason/RequiredInstallerResourcesMissing%'
AND locator like '%namespace/openshift-kube-controller-manager-operator deployment/kube-controller-manager-operator%'
LIMIT 1000
```

Based on the data I have to assume this is a legit failure, and a very rare one.

neisw:
Looks like the node update phases are no longer grouped with the node, can still figure it out but thinking this wasn't intentional

image

pkg/monitor/monitorapi/construction.go
```diff
@@ -294,24 +295,37 @@ func (b *LocatorBuilder) withServer(serverName string) *LocatorBuilder {
...
-		b.annotations[LocatorNamespaceKey] = event.InvolvedObject.Namespace
+	if event.InvolvedObject.Kind == "Namespace" {
+		// namespace better match the event itself.
+		return b.
+			withNamespace(event.InvolvedObject.Name).
+			Build()
+	} else if event.InvolvedObject.Kind == "Node" {
+		return b.
+			withTargetType(LocatorTypeNode).
+			withNode(event.InvolvedObject.Kind).
```
neisw, line 315: Are you wanting to set "Node" here or the actual node?

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,668 @@ package pathologicaleventlibrary
...
-
-	// [sig-node] Probing container ***
-	// these tests intentionally cause repeated probe failures to ensure good handling
-	regexp.MustCompile(`ns/e2e-container-probe-[0-9]+ .* probe failed: `),
-	regexp.MustCompile(`ns/e2e-container-probe-[0-9]+ .* probe warning: `),
+	registry.AddPathologicalEventMatcherOrDie(&SimplePathologicalEventMatcher{
+		name: "E2EStatefulSetReadinessProbeFailed",
+		locatorKeyRegexes: map[monitorapi.LocatorKey]*regexp.Regexp{
+			monitorapi.LocatorNamespaceKey: regexp.MustCompile(`e2e-statefulset-[0-9]+`),
+			monitorapi.LocatorPodKey:       regexp.MustCompile(`ss-[0-9]`),
```
DennisPeriquet, line 207: looking at line 62 above, should the pattern be `ss2-[0-9]`

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,668 @@ package pathologicaleventlibrary
...
-	regexp.MustCompile("ns/openshift-kube-controller-manager pod/kube-controller-manager-guard.*ProbeError Readiness probe error"),
-	regexp.MustCompile("ns/openshift-kube-scheduler pod/kube-scheduler-guard.*ProbeError Readiness probe error"),
+
+	registry.AddPathologicalEventMatcherOrDie(&SimplePathologicalEventMatcher{
+		name: "KubeAPIReadinessProbeError",
+		locatorKeyRegexes: map[monitorapi.LocatorKey]*regexp.Regexp{
+			monitorapi.LocatorNamespaceKey: regexp.MustCompile(`openshift-kube-*`),
+			monitorapi.LocatorPodKey:       regexp.MustCompile(`^kube.*guard.*`),
+		},
+		messageReasonRegex: regexp.MustCompile(`^ProbeError$`),
```
DennisPeriquet, line 326: I'm guessing you know this is the right `reason` since I don't see any `reason/...` on lines 109-112.

DennisPeriquet: looking at master, it seems to be the right value.

pkg/monitor/monitorapi/construction.go
dgoodwin: Ooops!

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
dgoodwin: Hmm, perhaps, this came from line 58, but to be honest looking at this:

```golang
	registry.AddPathologicalEventMatcherOrDie(&SimplePathologicalEventMatcher{
		name:               "KubeletUnhealthyReadinessProbeFailed",
		messageReasonRegex: regexp.MustCompile(`^Unhealthy$`),
		messageHumanRegex:  regexp.MustCompile(`Readiness probe failed`),
	})
```

They kinda both feel pointless don't they. I guess I should remove this one too?

dgoodwin: Taking it out.

dgoodwin:
Commits pushed.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -2,419 +2,668 @@ package pathologicaleventlibrary
...
+		messageHumanRegex:  regexp.MustCompile(`multiple versions found, probably in transition`),
+	})
 
 	// etcd-quorum-guard can fail during upgrades.
-	regexp.MustCompile(`ns/openshift-etcd pod/etcd-quorum-guard-[a-z0-9-]+ node/[a-z0-9.-]+ - reason/Unhealthy Readiness probe failed: `),
-	// etcd can have unhealthy members during an upgrade
-	regexp.MustCompile(`ns/openshift-etcd-operator deployment/etcd-operator - reason/UnhealthyEtcdMember unhealthy members: .*`),
-	// etcd-operator began to version etcd-endpoints configmap in 4.10 as part of static-pod-resource. During upgrade existing revisions will not contain the resource.
-	// The condition reconciles with the next revision which the result of the upgrade. TODO(hexfusion) remove in 4.11
-	regexp.MustCompile(`ns/openshift-etcd-operator deployment/etcd-operator - reason/RequiredInstallerResourcesMissing configmaps: etcd-endpoints-[0-9]+`),
```
DennisPeriquet, line 170: I don't see where this one is recovered.  The only one I see is https://github.com/dgoodwin/origin/blob/58516b3ec699c1490d114c7580eb0a951edc3c7c/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go#L550-L555 but it doesn't have `LocatorDeploymentKey...etcd-operator`.

DennisPeriquet: ah, ignore my comment above.

dgoodwin: Yup you found it :) Now I just checked bigquery, and it's still happening. :)

```sql
SELECT prowjob_name, prowjob_build_id, from_time, payload.locator, payload.message FROM `openshift-gce-devel.ci_analysis_us.job_intervals` WHERE from_time >= DATETIME_SUB(CURRENT_DATETIME(), INTERVAL 14 DAY) 
AND message like '%pathological%reason/RequiredInstallerResourcesMissing configmaps: etcd%' 
AND locator like '%openshift-etcd%'
LIMIT 1000
```

This returned 71 hits in the last two weeks. That's not a huge hit rate, but it will mean a few failed jobs. Then again, those hits are all 4.14, none in 4.15. I say lets keep it out and move on.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
```diff
@@ -486,3 +308,30 @@ func getBiggestRevisionForEtcdOperator(ctx context.Context, operatorClient opera
...
+
+	l := monitorapi.Locator{
+		Type: monitorapi.LocatorTypePod,
+		Keys: map[monitorapi.LocatorKey]string{},
+	}
+	if namespace != "" {
+		l.Keys[monitorapi.LocatorNamespaceKey] = namespace
+	}
+	if pod != "" {
+		l.Keys[monitorapi.LocatorPodKey] = namespace
```
DennisPeriquet, line 324: should that be ` = pod`?

neisw:
/lgtm
/hold
Looks like you have worked through all the identified issues.  Holding until you are ready

PR #28408, opened 2023-11-20
OCPBUGS-20487: update egressFWTestE2E image which contains ping binary

routerhan:
This PR update the image to use a non-nfs image, since we did not see any reference using nfs image, the test just trys to ping and curl things. And it also fixes the missing ping binary issue on s390x.


djslavens (approved):
/lgtm

PR #28409, opened 2023-11-20
TRT-1362: Update disruption/alert test data and fix unit test

dgoodwin:
- Automated - Update synthetic test data
- Fix broken unit test on synthetic data

When the query last ran, it looks at just a weeks worth of data for alerts and the job key we use only had 84, below the 100 we need to be returned for testing, so the test fails. It's now over 100 if run today but just barely. Switch to a job key that has over 200, which should be more reliable.


dgoodwin:

This has sat for weeks, and another update will trigger in two days, we need to get this in.

PR #28419, opened 2023-11-27
OCPBUGS-24290: cert metadata: generate jsons and markdown for missing metadata viola…

vrutkovs:


pkg/cmd/update-tls-artifacts/metadata/owner.go
deads2k, line 1: I think this is closer to the construct you want, in a directory with all things associated with owners. Then you could direct someone who wanted to do descriptions, regeneration, revocation, etc to start by copying the directory and ownership modesl would them (and you) to get added an OWNER.

pkg/cmd/update-tls-artifacts/metadata/owner.go
```diff
@@ -0,0 +1,146 @@
...
+	v := Violation{
+		Name:     name,
+		Registry: registry,
+	}
+
+	markdown, err := generateMarkdownNoOwner(pkiInfo)
+	if err != nil {
+		return v, err
+	}
+	v.Markdown = markdown
```
deads2k, line 37: +1 for markdown summarizing violators.

pkg/cmd/update-tls-artifacts/metadata/owner.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return v, nil
+}
+
+func generateMarkdownNoOwner(pkiInfo *certgraphapi.PKIRegistryInfo) ([]byte, error) {
```
deads2k, line 42: Is the content of this file built the same way kind of violation?

violations.md

```
This is a list of violations for <insert type here>

Table of Contents

<OwningComponent> (<number of violations>)
1. ordered list of violations with descriptions associated with each item

```

tls/ownership/description.md
deads2k, line 1: I expected a new directory created for `descriptions` as peer to `ownership` if I've understood the intent correctly.

That's what we'd suggest for adding say, "supports-revocation`, right?

pkg/cmd/update-tls-artifacts/metadata/violation.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (v Violation) getJSONFilePath(parentDir string) string {
+	return filepath.Join(parentDir, fmt.Sprintf("%s-violations.json", v.Name))
+}
+
+func (v Violation) getMarkdownFilePath(parentDir string) string {
+	return filepath.Join(parentDir, fmt.Sprintf("%s.md", v.Name))
+}
+
+func (v Violation) DiffWithExistingJSON(parentDir string) error {
```
deads2k, line 29: `parentDir` means `existingTLSDirector` to point to `tls` in our case with the path being `tls/<v.Name>/something.json`?

deads2k, line 29: If you're wrapping in this way, is it more logical to simply have a `DiffWithExisting` that does all the pertinent files?

pkg/cmd/update-tls-artifacts/metadata/violation.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (v Violation) WriteJSONFile(parentDir string) error {
+	violationJSONBytes, err := json.MarshalIndent(v.Registry, "", "    ")
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(v.getJSONFilePath(parentDir), violationJSONBytes, 0644)
+}
+
+func (v Violation) WriteMarkdownFile(parentDir string) error {
```
deads2k, line 68: same about, `Write` that just writes all the files?

pkg/cmd/update-tls-artifacts/metadata/violation.go
```diff
@@ -0,0 +1,72 @@
...
+	return os.WriteFile(v.getMarkdownFilePath(parentDir), v.Markdown, 0644)
+}
+
+type ViolationFunc func(name string, pkiInfo *certgraphapi.PKIRegistryInfo) (Violation, error)
```
deads2k, line 72: interfaces over functions.  What does this actually do?

pkg/cmd/update-tls-artifacts/metadata/requirements.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+)
+
+type secretCompareFunc func(actual, expected certgraphapi.PKIRegistryCertKeyPairInfo) error
+type configMapCompareFunc func(actual, expected certgraphapi.PKIRegistryCertificateAuthorityInfo) error
+
+type Requirement struct {
+	Name                 string
+	SecretCompareFunc    secretCompareFunc
```
deads2k, line 12: where is this used?

pkg/cmd/update-tls-artifacts/metadata/requirements.go
```diff
@@ -0,0 +1,34 @@
...
+	ConfigMapCompareFunc configMapCompareFunc
```
deads2k, line 13: where is this used?

pkg/cmd/update-tls-artifacts/generate-owners/generate_owners_options.go
deads2k, line 1: I expected this file to be contained in that "copy/paste as your starting point" so we'd have `generate_DESCRIPTIONS_options.go`

pkg/cmd/update-tls-artifacts/generate-owners/generate_owners_options.go
```diff
@@ -28,51 +22,19 @@ func (o *GenerateOwnersOptions) Run() error {
...
-	if err != nil {
-		return err
-	}
-	markdown, err := GenerateOwnershipMarkdown(result)
-	if err != nil {
-		return err
-	}
-	violations := GenerateViolationJSON(result)
-	violationJSONBytes, err := json.MarshalIndent(violations, "", "    ")
+	violations, err := metadata.GenerateViolationList(result, metadata.All...)
```
deads2k, line 25: your division here I think is close. You want something like

```go

type CertificateFeature interface{
  Get<Foo>JSON([]certgraphapi.PKIList) (bytes, error)
  Get<Foo>ViolationJSON([]certgraphapi.PKIList) (bytes, error)
  Get<Foo>Markdown([]certgraphapi.PKIList) (bytes, error)
  Get<Foo>ViolationMarkdown([]certgraphapi.PKIList) (bytes, error)
}

```

And then wrap that.  I think this is close, but we want to make individuals completely responsible for what they add.

vrutkovs: I've realized this file now generates for all violations. Do we want to have separate subcommands for owners/description (and make people copy it whenever a new one is added)?
iiuc there's no harm in generating all files now

pkg/cmd/update-tls-artifacts/metadata/violation.go
vrutkovs: Correct

pkg/cmd/update-tls-artifacts/metadata/requirements.go
vrutkovs: Rebase ate my commit which used this in test. This allows us to customize diff function (not sure if it would be useful later on though)

pkg/cmd/update-tls-artifacts/metadata/violation.go
deads2k: If I have to ask, write doc and/or rename.

PR #28426, opened 2023-11-29
eliminate the closed channel race in pod_log_streamer

deads2k:
now fits in the margin.

@neisw


neisw:
[TRT-1358](https://issues.redhat.com/browse/TRT-1358)
/lgtm

PR #28427, opened 2023-11-29
[wip] Certs 13 use violations

deads2k:
Collaborating with https://github.com/openshift/origin/pull/28419

PR #28429, opened 2023-11-29
adding a new TLS requirement for auto-regeneration of certificates after expiry

deads2k:
after https://github.com/openshift/origin/pull/28427, demonstrates adding a new TLS requirement for auto-regeneration of certificates after expiry.  This gives us 

1. the list of those not compliant by owner
2. the ability to track completion one by one
3. auditability on that completion due to commit the annotations
4. e2e test ensuring that no cert is added without signing off that this functions properly
5. when the list of non-compliance is zero, we're done

/hold


vrutkovs:
Basing https://github.com/openshift/origin/pull/28430 on top of it

PR #28430, opened 2023-11-30
Certs 14 add auto regenerate remix

vrutkovs:


pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/description/requirement.go
deads2k, line 1: Seeing how this interacts with your previous.  Perhaps we create an "AnnotationBasedRequirement" that both Description and Regeneration (but **not** owners) use?

tls/description/description.md
deads2k, line 1: The auto-regeneration.md is organized by owning component in table of contents.  We should try to use that format for easy attribution.

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/description/requirement.go
vrutkovs: >AnnotationBasedRequirement

I think so far we want them separated, so that we could require owners now and have descriptions optional.

Once both are set we can squash into one

PR #28432, opened 2023-11-30
OCPBUGS-24290: cert metadata: generate jsons and markdown for missing metadata viola

deads2k:
https://github.com/openshift/origin/pull/28419 with a commit to have a library-go fix


vrutkovs:
/hold

This would make jobs fail, a bit too soon

PR #28433, opened 2023-11-30
push violation regression check into the default requirement result

deads2k:
Added a commit to do clean the starting base for requirementresult to the logical small point.

/assign @vrutkovs


vrutkovs (approved):
/lgtm

vrutkovs:
/retitle OCPBUGS-24583: push violation regression check into the default requirement result

PR #28441, opened 2023-12-05
TRT-1370: Track etcd leadership changes in etcd-leaders section of spyglass chart

DennisPeriquet:
[TRT-1370](https://issues.redhat.com//browse/TRT-1370)

This adds the `etcd-leaders` section in the spyglass charts to help us track etcd leader changes more clearly.

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -97,18 +97,17 @@
         return false
     }
 
-    // When an interval in the openshift-etcd namespace had a reason of leader-found, leader-lost,
-    // leader-elected, or leader-missing, tempSource was set to 'EtcdLeadership' otherwise it was
-    // set to 'EtcdLog'.  This function returns true if the interval is one of those.
-    function isEtcdLeadershipOrLog(eventInterval) {
-        if (eventInterval.tempSource == 'EtcdLeadership' || eventInterval.tempSource == 'EtcdLog') {
+    // When an interval in the openshift-etcd namespace had a reason of LeaderFound, LeaderLost,
+    // LeaderElected, or LeaderMissing, tempSource was set to 'EtcdLeadership'.
+    function isEtcdLeadership(eventInterval) {
+        if (eventInterval.tempSource == 'EtcdLeadership') {
             return true
         }
         return false
     }
 
     function isPodLog(eventInterval) {
-        if (isEtcdLeadershipOrLog(eventInterval)) {
+        if (isEtcdLeadership(eventInterval)) {
             return false
         }
         if (eventInterval.locator.includes("src/podLog")) {
@@ -125,7 +124,7 @@
     }
 
     function isPod(eventInterval) {
-        if (isEtcdLeadershipOrLog(eventInterval)) {
+        if (isEtcdLeadership(eventInterval)) {
             return false
         }
         // this check was added to keep the repeating events out fo the "pods" section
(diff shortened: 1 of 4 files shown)


e2echart/e2e-chart-template.html
```diff
@@ -97,13 +97,50 @@ <h5 class="modal-title">Resource</h5>
...
+    // If this is a leader found, lost, elected, or missing, return true and the reason.
+    // Otherwise, return false.
+    function isEtcdLeaderFoundOrLost(eventInterval) {
+        const { tempStructuredLocator, tempStructuredMessage } = eventInterval;
+        let reason = null;
+        if (
+            tempStructuredLocator &&
+            tempStructuredLocator.keys &&
+            tempStructuredLocator.keys.hasOwnProperty('namespace') &&
+            tempStructuredLocator.keys['namespace'].includes('openshift-etcd')
```
dgoodwin, line 109: This seems excessive, were all these checks required? tempStructuredLocator should always be there, as should keys, but it could be null. We want to be as concise as possible here so this doesn't spread, this file would be a mess if we do this every time we check a property.

e2echart/e2e-chart-template.html
```diff
@@ -97,13 +97,50 @@ <h5 class="modal-title">Resource</h5>
         return false
     }
 
+    // If this is a leader found, lost, elected, or missing, return true and the reason.
+    // Otherwise, return false.
+    function isEtcdLeaderFoundOrLost(eventInterval) {
+        const { tempStructuredLocator, tempStructuredMessage } = eventInterval;
```
dgoodwin, line 103: Whoah, how does this work?

e2echart/e2e-chart-template.html
```diff
@@ -97,13 +97,50 @@ <h5 class="modal-title">Resource</h5>
...
+        if (
+            tempStructuredLocator &&
+            tempStructuredLocator.keys &&
+            tempStructuredLocator.keys.hasOwnProperty('namespace') &&
+            tempStructuredLocator.keys['namespace'].includes('openshift-etcd')
+        ) {
+            if (tempStructuredMessage && tempStructuredMessage.hasOwnProperty('reason')) {
+                reason = tempStructuredMessage['reason']
+                if (reason.includes('leader-found') || reason.includes('leader-lost') ||
+                    reason.includes('leader-elected') || reason.includes('leader-missing')) {
```
dgoodwin, line 114: Could we instead update where these intervals are created (after #28474 merges) to have tempSource = EtcdLeadership, and eliminate the need for reason checking here in the js.

e2echart/e2e-chart-template.html
```diff
@@ -115,6 +152,13 @@ <h5 class="modal-title">Resource</h5>
     }
 
     function isPod(eventInterval) {
+        const [isLeaderFoundOrLost, notUsed] = isEtcdLeaderFoundOrLost(eventInterval)
+        if (isLeaderFoundOrLost) {
+            return false
+        }
+        if (eventInterval.locator.includes("etcd-member/")) {
+            return false
+        }
```
dgoodwin, line 161: This callout to isEtcdLeaderFoundOrLost looks still required until we fully port the file. This isPod section in particular looks susceptible to showing things because it doesn't check tempSource.

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +357,68 @@ <h5 class="modal-title">Resource</h5>
         return [item.locator, ` (${roles})`, m];
     }
 
+    function etcdLogs(item) {
```
dgoodwin, line 360: Suggest etcdLeadership

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +357,68 @@ <h5 class="modal-title">Resource</h5>
...
+                    break;
+                case "leader-lost":
+                    leaderColor = "EtcdLeaderLost";
+                    break;
+                case "leader-elected":
+                    leaderColor = "EtcdLeaderElected";
+                    break;
+                case "leader-missing":
+                    leaderColor = "EtcdLeaderMissing";
+                    break;
```
dgoodwin, line 397: I would prefer to see this logic moved to origin itself when generating the interval, as these Reason values don't match the syntax of any others, they're always CamelCase. 

Color will be tricky when we get to porting the whole file.

dgoodwin (commented):
Let's rebase this on #28463 and then do this based and use tempSource a little bit per the comments.

e2echart/e2e-chart-template.html
DennisPeriquet: it's a destructuring assignment (kind of a shorthand in javascript) -- but no longer necessary.

e2echart/e2e-chart-template.html
DennisPeriquet: understood.  Since you can guarantee those fields will be present, removing.

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +326,68 @@ <h5 class="modal-title">Resource</h5>
...
+
+            if (nodeVal.length > 0 && etcdMemberVal.length > 0) {
+                const term = item.tempStructuredMessage.annotations['term']
+                return [`Leader ${nodeVal} term/${term}`, ` etcd-member/${etcdMemberVal}`, 'EtcdLeading']
+            }
+
+            if (nodeVal == "noNode" && etcdMemberVal.length == 0) {
+                // If node is in the locator, use it.
+                nodeVal = item.tempStructuredLocator.keys['node'] || 'noNode'
+                return [`${nodeVal}`, ` (etcd-member=none)`, 'EtcdOther']
```
dgoodwin, line 384: When nodeVal is noNode, these intervals are pretty confusing. Is that showing any useful information at all?

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +326,68 @@ <h5 class="modal-title">Resource</h5>
...
+            }
+
+            const etcdMemberVal = item.tempStructuredLocator.keys['etcd-member'] || ''
+            if (nodeVal.length == 0 && etcdMemberVal.length == 0) {
+                return [`empty`, `(empty)`, 'EtcdOther']
+            }
+
+            if (nodeVal.length > 0 && etcdMemberVal.length > 0) {
+                const term = item.tempStructuredMessage.annotations['term']
+                return [`Leader ${nodeVal} term/${term}`, ` etcd-member/${etcdMemberVal}`, 'EtcdLeading']
```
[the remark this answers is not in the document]
dgoodwin: What do you think about just showing these intervals? Did someone request the per node single second intervals we get each election? The chart would look a lot cleaner if it was just the leadership lines showing that move. I find the flurry of per-second ones kinda hard to interpret and I fear others may as well.

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +326,68 @@ <h5 class="modal-title">Resource</h5>
...
+            return [`${nodeVal} term/${term}`, ` (${foundOrLost}=${nodeVal})`, leaderColor]
+        } else {
+
+            let nodeVal = 'noNode'
+            if (item.tempStructuredLocator.keys.hasOwnProperty('etcd-member') &&
+                item.tempStructuredLocator.keys.hasOwnProperty('node')) {
+                if (item.tempStructuredLocator.keys['node'].length > 0) {
+                    nodeVal = item.tempStructuredLocator.keys['node']
+                }
+            }
```
dgoodwin, line 369: Please use the cleaner syntax you use below, const = expression || 'default'

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +326,68 @@ <h5 class="modal-title">Resource</h5>
...
+                    break;
+                case "leader-missing":
+                    leaderColor = "EtcdLeaderMissing";
+                    break;
+            }
+            return [`${nodeVal} term/${term}`, ` (${foundOrLost}=${nodeVal})`, leaderColor]
+        } else {
+
+            let nodeVal = 'noNode'
+            if (item.tempStructuredLocator.keys.hasOwnProperty('etcd-member') &&
```
dgoodwin, line 364: Do you need to check etcd member here?

DennisPeriquet:
Since this [switch statement](https://github.com/openshift/origin/blob/487579bd509722ede83cacaf0b12eb03a53a9f53/pkg/monitortests/etcd/etcdloganalyzer/monitortest.go#L202-L264), shows what is present for each of those leadership related intervals, and [marking the leadership change](https://github.com/openshift/origin/blob/487579bd509722ede83cacaf0b12eb03a53a9f53/pkg/monitortests/etcd/etcdloganalyzer/monitortest.go#L116) interval, I simplified the `etcdLeadershipLogsValue` function in the javascript and fully removed that confusing `else` clause.  This allows me to get the > 1s intervals that show when a node was an etcd leader.

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +325,29 @@ <h5 class="modal-title">Resource</h5>
...
+        const nodeVal = item.tempStructuredLocator.keys['node']
+
+        // Get etcd-member value (this will be present for a leader change).
+        let etcdMemberVal = item.tempStructuredLocator.keys['etcd-member'] || ''
+        if (etcdMemberVal.length > 0) {
+            etcdMemberVal = `etcd-member/${etcdMemberVal} `
+        }
+
+        let reason = item.tempStructuredMessage.reason
+        let color = 'EtcdOther'
```
dgoodwin, line 343: color appears unused

e2echart/e2e-chart-template.html
```diff
@@ -313,6 +325,29 @@ <h5 class="modal-title">Resource</h5>
         return [item.locator, ` (${roles})`, m];
     }
 
+    function etcdLeadershipLogsValue(item) {
```
dgoodwin, line 328: Massive improvement, I can look at this and understand much easier now, thanks.

dgoodwin (commented):
Chart looking pretty good, last thing is why do a few intervals show up with node/[ip]:[port] ?

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/28441/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1745777337187700736

DennisPeriquet:
> Chart looking pretty good, last thing is why do a few intervals show up with node/[ip]:[port] ?
> 
> https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/28441/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1745777337187700736

Looking at the code, we are getting these locators/messages from the pod logs in the `openshift-etcd` namespace.  We are not manipulating the the node names.

PR #28444, opened 2023-12-07
NO-JIRA: Collect certs from disk

vrutkovs:
Testbed for https://github.com/openshift/library-go/pull/1613

### DIFF
--- a/pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
+++ b/pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
@@ -1,0 +1,186 @@
+package collectdiskcertificates
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphanalysis"
+	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+	"github.com/openshift/origin/pkg/clioptions/iooptions"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+type RunCollectDiskCertificatesFlags struct {
+	ConfigFlags *genericclioptions.ConfigFlags
+	OutputFlags *iooptions.OutputFlags
+
+	ArtifactDir string
+	Prefix      string
+	CollectDirs []string
+
+	genericclioptions.IOStreams
+}
+
+func NewRunCollectDiskCertificatesFlags(ioStreams genericclioptions.IOStreams) *RunCollectDiskCertificatesFlags {
+	return &RunCollectDiskCertificatesFlags{
+		ConfigFlags: genericclioptions.NewConfigFlags(false),
+		OutputFlags: iooptions.NewOutputOptions(),
+		IOStreams:   ioStreams,
+	}
+}
+
+func NewRunCollectDiskCertificatesCommand(ioStreams genericclioptions.IOStreams) *cobra.Command {
+	f := NewRunCollectDiskCertificatesFlags(ioStreams)
+	cmd := &cobra.Command{
+		Use:   "collect-disk-certificates",
+		Short: "Run disk certificate collector",
+		Long: templates.LongDesc(`
+		Run a colelctor which fetches information about certificates stored on disk
+		`),
+
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, cancelFn := context.WithCancel(context.Background())
+			defer cancelFn()
+			abortCh := make(chan os.Signal, 2)
+			go func() {
+				<-abortCh
+				fmt.Fprintf(f.ErrOut, "Interrupted, terminating\n")
+				cancelFn()
+
+				sig := <-abortCh
+				fmt.Fprintf(f.ErrOut, "Interrupted twice, exiting (%s)\n", sig)
+				switch sig {
+				case syscall.SIGINT:
+					os.Exit(130)
+				default:
+					os.Exit(0)
+				}
+			}()
+			signal.Notify(abortCh, syscall.SIGINT, syscall.SIGTERM)
+
+			if err := f.Validate(); err != nil {
+				return err
+			}
+
+			o, err := f.ToOptions()
+			if err != nil {
+				return err
+			}
+
+			return o.Run(ctx)
+		},
+	}
+
+	f.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (f *RunCollectDiskCertificatesFlags) AddFlags(flags *pflag.FlagSet) {
+	flags.StringArrayVar(&f.CollectDirs, "collect-dir", f.CollectDirs, "directories to collect certs in")
+	flags.StringVar(&f.Prefix, "prefix", f.Prefix, "directories prefix")
+	f.ConfigFlags.AddFlags(flags)
+	f.OutputFlags.BindFlags(flags)
+}
+
+func (f *RunCollectDiskCertificatesFlags) SetIOStreams(streams genericclioptions.IOStreams) {
+	f.IOStreams = streams
+}
+
+func (f *RunCollectDiskCertificatesFlags) Validate() error {
+	if len(f.OutputFlags.OutFile) == 0 {
+		return fmt.Errorf("output-file must be specified")
+	}
+	if len(f.CollectDirs) == 0 {
+		return fmt.Errorf("dirs must be specified")
+	}
+
+	return nil
+}
+
+func (f *RunCollectDiskCertificatesFlags) ToOptions() (*RunCollectDiskCertificatesOptions, error) {
+	originalOutStream := f.IOStreams.Out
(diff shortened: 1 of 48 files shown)


pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
```diff
@@ -0,0 +1,186 @@
...
+		IOStreams:        f.IOStreams,
+		CollectDirs:      f.CollectDirs,
+		Prefix:           f.Prefix,
+	}, nil
+}
+
+// RunCollectDiskCertificatesOptions sets options for api server disruption monitor
+type RunCollectDiskCertificatesOptions struct {
+	KubeClient       kubernetes.Interface
+	KubeClientConfig *rest.Config
```
deads2k, line 145: dead?

test/extended/operators/manifests/namespace.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  generateName: e2e-collect-certificate-
```
deads2k, line 4: yay!

test/extended/operators/certs.go
```diff
@@ -179,3 +200,153 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+			return fmt.Errorf("pod %s in namespace %s didn't complete: %v", actualPod.Name, namespace, watchErr)
+		}
+	}
+	return nil
+}
+
+func fetchNodePKIList(ctx context.Context, kubeClient kubernetes.Interface, node *corev1.Node) (*certgraphapi.PKIList, error) {
+	pkiList := &certgraphapi.PKIList{}
+
+	allBytes, err := nodeaccess.GetNodeLogFile(ctx, kubeClient, node.Name, certInspectResultFile)
```
deads2k, line 312: If memory serves, this uses a construct that raw kube clusters disallow.  While I appreciate this works, you could more generally

1. make the `collector` container into an initcontainer
2. make a `pause` container 
3. watch for `pause` to enter running state
4. make this copy pull from the pod instead of the node
5. delete the pod after a successful copy

This would
1. ensure we work on stock kube as well in the (likely) case we make this a separate command
2. avoid writing content to the host filesystem

test/extended/operators/certs.go
```diff
@@ -179,3 +200,153 @@ var _ = g.Describe("[sig-arch][Late]", func() {
...
+	err = json.Unmarshal(allBytes, pkiList)
+	if err != nil {
+		return pkiList, fmt.Errorf("failed to unmarshal file %s on node %s: %v", certInspectResultFile, node.Name, err)
+	}
+
+	return pkiList, nil
+}
+
+// GetOpenshiftTestsImagePullSpec returns the pull spec or an error.
+func getOpenshiftTestsImagePullSpec(ctx context.Context, adminRESTConfig *rest.Config) (string, error) {
```
deads2k, line 325: tbh, I don't remember how we did this for the networking check's monitortest, but let's expose that somewhere for internal consistency.

test/extended/operators/certs.go
```diff
@@ -179,3 +200,153 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 	})
 
 })
+
+func fetchOnDiskCertificates(ctx context.Context, kubeClient kubernetes.Interface, nodeList *corev1.NodeList, testPullSpec string) (*certgraphapi.PKIList, error) {
```
deads2k, line 204: I'm +1 on making this a command as well if you like.  I think we'll eventually need it to replace the cluster-debug-tools stuff.

deads2k:
a PR that shows the collected data would be great.

test/extended/operators/certs.go
vrutkovs: We use [Node Logs API](https://kubernetes.io/blog/2023/04/21/node-log-query-alpha/), its however in alpha

test/extended/operators/certs.go
vrutkovs: Updated to reuse the same function

deads2k:
update from earlier this morning (not most recent) doesn't combine

```
            {
              "Namespace": "openshift-etcd",
              "Name": "etcd-peer-client-ca"
            },

```


I haven't looked into why yet.

vrutkovs:
Picked up a new library-go, should be working now

/test e2e-gcp-ovn

pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
```diff
@@ -0,0 +1,217 @@
...
+	defer cancelFn()
+
+	pkiList := &certgraphapi.PKIList{}
+	errs := []error{}
+
+	restConfig, err := clusterinfo.GetMonitorRESTConfig()
+	if err != nil {
+		return err
+	}
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
```
deads2k, line 171: `RunCollectDiskCertificatesOptions` already has a kubeClient, why create another one instead of `o.KubeClient`?

pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
```diff
@@ -0,0 +1,217 @@
...
+	if err := wait.PollImmediate(time.Second, 60*time.Second, func() (bool, error) {
+		var listErr error
+		nodeList, listErr = kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{LabelSelector: controlPlaneLabel.String()})
+		if err != nil {
+			return false, listErr
+		}
+		return true, nil
+	}); err != nil {
+		return fmt.Errorf("failed to list nodes: %v", err)
+	}
```
deads2k, line 187:
```suggestion
		nodeList, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{LabelSelector: controlPlaneLabel.String()})
 if err != nil{
```

is sufficient.

vrutkovs: I'm not sure if that would work 100% of the time. In some cases CRB didn't get applied in time, so pod failed to start - and since its not backed with a deployment it went to Error and test failed. In actual tests the nodes will probably take time pulling the image, but I'd rather have this retry just in case

pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
```diff
@@ -0,0 +1,208 @@
...
+
+	f.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (f *RunCollectDiskCertificatesFlags) AddFlags(flags *pflag.FlagSet) {
+	flags.StringArrayVar(&f.CollectDirs, "collect-dir", f.CollectDirs, "directories to collect certs in")
+	flags.StringVar(&f.Prefix, "prefix", f.Prefix, "directories prefix")
```
deads2k, line 99: bad arg name and/or description.  Is this a replacement value maybe?  Why are we replacing this actually, why not just leave the full path?

vrutkovs: Full path starts with `/rootfs` - where we mount host FS. I suppose we can leave it as is

test/extended/operators/certs.go
deads2k: > We use [Node Logs API](https://kubernetes.io/blog/2023/04/21/node-log-query-alpha/), its however in alpha

ours is patched to expose more files.

deads2k: This is still a problem.  Let's collect from the pod

test/extended/operators/manifests/clusterrolebinding-nodelist.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: e2e-collect-certificate-nodelist
```
deads2k, line 4: generate name to avoid conflict.

pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
vrutkovs: Perhaps "root-mount-point" would be better?

test/extended/operators/certs.go
vrutkovs: Reworked to collect json in initContainer, mount it via emptyDir to the "pause" container and fetch it via `pod/exec`

pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
deads2k: what is CRB?

test/extended/util/exec.go
```diff
@@ -0,0 +1,39 @@
...
+	v1 "k8s.io/api/core/v1"
+	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+	"k8s.io/kubectl/pkg/scheme"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// commandContents fetches the result of invoking a command in the provided container from stdout.
+func ExecInPodWithResult(podClient coreclientset.CoreV1Interface, podRESTConfig *rest.Config, ns, name, containerName string, command []string) (string, error) {
```
deads2k, line 16: followup: this shouldn't live under tests.  it's used by commands.

pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
vrutkovs: ClusterRoleBinding to allow the container to list nodes. Removed retry, lets see if this happens in CI anyway

deads2k:
?!  kcm on aws fips has 30 revisions??? https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/28444/pull-ci-openshift-origin-master-e2e-aws-ovn-fips/1747189467136921600

PR #28447, opened 2023-12-07
OWNER_ALIASES: update CoreOS team leads

miabbott:
I left the CoreOS team over a year ago, so there are new co-leads for the team.


miabbott:
@mike-nguyen @travier FYI

May need an update to the CoreOS team operations docs to make sure this is kept up-to-date for onboarding/offboarding

miabbott:
@bparees could you provide the `/override` here?  Seems insane to run e2e tests on a doc change; I have a PR to try to avoid that in the future - https://github.com/openshift/release/pull/46554

PR #28450, opened 2023-12-08
update-tls-artifacts: add AnnotationRequirement

vrutkovs:
This is a wrapper around tlsmetadatainterfaces.Requirement which helps users quickly create new requirements for secret/configmap annotations

### DIFF
--- a/pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/autoregenerate_after_expiry/requirement.go
+++ b/pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/autoregenerate_after_expiry/requirement.go
@@ -1,7 +1,6 @@
 package autoregenerate_after_expiry
 
 import (
-	"encoding/json"
 	"fmt"
 
 	"github.com/openshift/origin/pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces"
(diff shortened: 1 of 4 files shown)


pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/annotation_requirement.go
```diff
@@ -0,0 +1,56 @@
+package tlsmetadatainterfaces
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+)
+
+func AnnotationInspectRequirement(o AnnotationRequirement, rawData []*certgraphapi.PKIList) (RequirementResult, error) {
```
deads2k, line 10:
```suggestion
type annotationRequirement struct{
  requirementName string
  annotationName string
}

var _ Requirement = &annotationRequirement{}

func (o &annotationRequirement) InspectRequirement(rawData []*certgraphapi.PKIList) (tlsmetadatainterfaces.RequirementResult, error) {
```

Rather than bare functions, build this as a struct that contains the details.  Something like

```go

```

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/types.go
```diff
@@ -13,6 +13,19 @@ type Requirement interface {
...
+	// GetAnnotationName returns annotation name to use
+	GetAnnotationName() string
+
+	// GenerateMarkdown returns byte representation of markdown report
+	GenerateMarkdown(pkiInfo *certgraphapi.PKIRegistryInfo) ([]byte, error)
+
+	// InspectRequirement generates and returns the result for a particular set of raw data
+	InspectRequirement(rawData []*certgraphapi.PKIList) (RequirementResult, error)
+}
+
```
deads2k, line 28:
```suggestion
type AnnotationRequirement interface {
	Requirment 
	
	// GetAnnotationName returns annotation name to use
	GetAnnotationName() string

}

```

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/autoregenerate_after_expiry/requirement.go
```diff
@@ -11,67 +10,25 @@ import (
...
-type AutoRegenerateAfterOfflineExpiryRequirement struct {
-	name string
-}
+type AutoRegenerateAfterOfflineExpiryRequirement struct{}
 
 func NewAutoRegenerateAfterOfflineExpiryRequirement() tlsmetadatainterfaces.Requirement {
-	return AutoRegenerateAfterOfflineExpiryRequirement{
-		name: "autoregenerate-after-expiry",
-	}
+	return AutoRegenerateAfterOfflineExpiryRequirement{}
```
deads2k, line 16:
```suggestion
	return NewAnnotationRequirement("autoregenerate-after-expiry", "certificates.openshift.io/auto-regenerate-after-offline-expiry")
```

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/annotation_requirement.go
vrutkovs: That'd work, but along with `requriementName` and `annotationName` I also have to pass `generateMarkdownFn` so that each class could implement its own markdown report. I could add it to the interface, but then I won't be able to have a common `InspectRequirement` func. Not sure if there is a better solution

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/autoregenerate_after_expiry/requirement.go
```diff
@@ -10,68 +9,20 @@ import (
...
 )
 
-// TODO move to openshift/api
-const AutoRegenerateAfterOfflineExpiryAnnotation = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
+const annotationName string = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
 
-type AutoRegenerateAfterOfflineExpiryRequirement struct {
-	name string
-}
+// TODO move to openshift/api
```
deads2k, line 14: comment is no longer accurate.  up on the annotation name probably?

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/autoregenerate_after_expiry/requirement.go
```diff
@@ -10,68 +9,20 @@ import (
...
-		if len(regenerates) == 0 {
-			ret.CertificateAuthorityBundles = append(ret.CertificateAuthorityBundles, curr)
-		}
-	}
-
-	return ret
+	return tlsmetadatainterfaces.NewAnnotationRequirement(
+		"autoregenerate-after-expiry",
+		annotationName,
+		generateAutoRegenerateAfterOfflineExpiryMarkdownFn,
```
deads2k, line 21: Probably just need the "what does this mean" block, right?

vrutkovs: Updated, PTAL

PR #28451, opened 2023-12-09
NO-JIRA: Update Omer to be the SNO lead

mrunalp:
@omertuc I talked with @romfreiman and he suggested you should be the lead here. 

@bparees @dhellmann fyi


bparees:
will let @omertuc lgtm this if he agree with the assignment

aravindhp:
/retitle NO-JIRA: Update Omer to be the SNO lead

PR #28465, opened 2023-12-14
OCPNODE-1892: Conditionally skip tests to help with k8s 1.29

sairameshv:
Based on the following conversation[1], this test would be skipped until a fix to the blocker bug[2] is added 

[1] - https://redhat-internal.slack.com/archives/C065R4NCLGM/p1702558716657289?thread_ts=1702481324.774039&cid=C065R4NCLGM 
[2] - https://issues.redhat.com/browse/OCPBUGS-25402

/cc @soltysh


soltysh (approved):
/lgtm
/approve
/hold cancel

PR #28466, opened 2023-12-14
add requirement for hostname change

deads2k:
Keep track of which certificates support the SNO hostname change tool and prevent new certificates from being added that have not confirmed this capability to avoid accidental regressions.

After testing for a particular TLS artifact is complete (requirements noted in the markdown), adding the annotation to a particular secret or configmap indicates that certificate is complete.  Keep in mind that 
1. existing e2e testing does not necessarily cover every certificate and CA bundle
2. this mechanism adds an e2e test that ensures new certificates must work with a hostname change prior to coming into the payload.
3. acknowledgement is lightweight, see https://github.com/openshift/cluster-etcd-operator/pull/1159/files as a similar example.
4. this has the benefit of ensuring that teams are aware of exactly how the hostname changes impact their certificates

/hold

cc @cuppett @romfreiman

### DIFF
--- a/pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/hostname_change/requirement.go
+++ b/pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/hostname_change/requirement.go
@@ -1,0 +1,39 @@
+package hostname_change
+
+import "github.com/openshift/origin/pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces"
+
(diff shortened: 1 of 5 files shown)


romfreiman:
@omertuc @mresvanis

omertuc:
In a relocation (image based upgrade/install) scenario we're changing four things:

- Hostname
- Node IP address
- Cluster name
- Cluster base domain

Is:

>Hostname

Mentioned above referring to all four things? Or is there some distinction needed here?

PR #28467, opened 2023-12-15
add test to ensure all accessible-by-default apis are stable

deads2k:
/hold

make the exclude list prior to merge

### DIFF
--- a/test/extended/operators/crd_must_be_stable.go
+++ b/test/extended/operators/crd_must_be_stable.go
@@ -1,0 +1,73 @@
+package operators
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


test/extended/operators/crd_must_be_stable.go
```diff
@@ -0,0 +1,75 @@
...
+					if !stableVersions.Has(versionSpec.Name) {
+						failures = append(failures,
+							fmt.Sprintf("crd/%v has an unstable version %q that is accessible-by-default. All CRDs accessible by default must be stable (v1, v2, etc) with guaranteed compatibility and upgradeability ~forever.",
+								crd.Name, versionSpec.Name))
+					}
+				}
+			}
+
+			if len(failures) > 0 {
+				result.Flakef(strings.Join(failures, "\n"))
```
deads2k, line 71: this makes it flake instead of fail.

PR #28478, opened 2023-12-20
OCPBUGS-25730: Kuryr: Ignore Upgradeable=False on operators tests

dulek:
Kuryr SDN is removed from 4.15 and 4.14 blocks upgrades if it's configured. This patch makes sure operator tests are ignoring such a condition.

### DIFF
--- a/test/extended/operators/operators.go
+++ b/test/extended/operators/operators.go
@@ -109,6 +109,12 @@ var _ = g.Describe("[sig-arch][Early] Managed cluster should [apigroup:config.op
 					continue
 				}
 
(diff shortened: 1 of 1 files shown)


itzikb-redhat:
Tested with 4.14.7 
The following tests passed

`"[sig-arch][Early] Managed cluster should [apigroup:config.openshift.io] start all core operators [Skipped:Disconnected] [Suite:openshift/conformance/parallel]"
"[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should have same number of Machines and Nodes [apigroup:machine.openshift.io] [Suite:openshift/conformance/parallel]"`

gryf (approved):
/lgtm

PR #28471, opened 2023-12-18
OCPBUGS-25610: fix: lane e2e-gcp-ovn-builds perma fail

eggfoobar:
Recent [changes](https://github.com/openshift/cluster-samples-operator/pull/522) to image streams removed ruby 2.7 in 4.15, updating to use ruby 3.0 in test build assets.

PR #28481, opened 2023-12-20
OCPBUGS-25610: fix: update missing ruby 3.0 ref

eggfoobar:
updated images.sh and persistent to reference 3.0 image streams

Fixes issue highlighted in agnostic-ovn-cmd, https://github.com/openshift/origin/pull/28471#issuecomment-1864489107


ardaguclu:
Thanks for the fix @eggfoobar

PR #28489, opened 2024-01-03
Update certs again

deads2k:
starting the new year right.  Skips hypershift because their certs don't align well.  It needs addressing eventually, but will bring us closer to being enforceable.


vrutkovs (approved):
/lgtm
/hold

for hypershift tests

vrutkovs:
/hold cancel

No flaking TLS tests in that kubevirt job

PR #28495, opened 2024-01-04
OCPBUGS-25142: Retry with different node port to make oc service test stable

ardaguclu:
Port number `30000` may be used in nodes by legitimate services and `oc service`
tests may incorrectly fail due to the already allocated port error. To prevent this and
have a stable test, this PR adds a simple retry with port number `30001`, if it gets
an allocated port error.

### DIFF
--- a/test/extended/cli/service.go
+++ b/test/extended/cli/service.go
@@ -1,8 +1,10 @@
 package cli
 
 import (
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/approve

PR #28500, opened 2024-01-04
adding proof for comparison

deads2k:
/hold

just to compare and see what the generate step says.

### DIFF
--- a/tls/raw-data/raw-tls-artifacts-ha-amd64-gcp-ovn.json
+++ b/tls/raw-data/raw-tls-artifacts-ha-amd64-gcp-ovn.json
@@ -351,7 +351,13 @@
           "Name": "etcd-ca-bundle"
         },
         "certificateAuthorityBundleInfo": {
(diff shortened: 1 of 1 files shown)


tls/raw-data/raw-tls-artifacts-ha-amd64-gcp-ovn.json
```diff
@@ -5252,6 +5339,488 @@
...
+          ]
+        },
+        "Status": {
+          "Errors": null
+        }
+      },
+      {
+        "LogicalName": "",
+        "Description": "",
+        "Name": "etcd-metric-signer",
```
deads2k, line 5588: Why didn't this combine with the etcd-metric-signer in the cluster?

deads2k: We need this to combine so that we have most of these auto-discover ownership, right?

PR #28510, opened 2024-01-08
demonstrate problem generating docs

deads2k:
In https://github.com/openshift/library-go/pull/1592 I indicated that it would cause problems for generating documentation.  I've had to update the chain guessing to be good enough to even see which problems we have, but have a look at https://github.com/openshift/api/tree/master/tls/docs/kube-apiserver%20Serving%20Certificates#kube-apiserver-localhost-signer .  It used to have a description.  We cannot use the description off of the in-cluster and on-disk locations because those are targeted for "what is this instance using it for", but we need an overall description to describe the thing.

To my knowledge, no replacement for this lost functionality was created, so we have a gap across dozens of these certificates.  How will we solve that?

/assign @vrutkovs


deads2k:
unit tests added.  A couple bugs fixed.  But now won't complete without public key modulus in the raw tls files, so I really want that soon.

PR #28515, opened 2024-01-10
OCPBUGS-26601: Re-enable test/extended/router/http2 tests on AWS

frobware:
It's been a long time since we disabled these tests on AWS. I have been running the http2 tests on AWS all week and I haven't run into the issue once. Let's re-enable the http2 x AWS tests for better coverage.

This PR also addresses an intermittent issue encountered in AWS environments during the router's h2spec conformance tests. The challenge involved slower hostname resolution within the cluster, resulting in frequent timeouts. Notably, AWS exhibited slower resolution times compared to Azure or GCP, hinting at potential differences in DNS handling.

The solution implemented in this PR focuses on resolving the hostname on the test host before initiating the h2spec tests within the cluster. This adjustment has resulted in a remarkable improvement in test execution speed, with the h2spec test now completing in approximately 85 seconds, a significant reduction from the previous average of over 376 seconds (just above the 5-minute mark).

While the difference in resolution times suggests environmental variations, particularly in AWS, it's important to note that this PR does not definitively attribute the issue to negative caching. Instead, it prioritises the substantial improvement achieved through the new approach. As a precaution, the polling interval and overall test timeout have been adjusted to 2 seconds and 10 minutes, respectively, to enhance test success rates across diverse cloud environments.

This PR represents a practical win in terms of improved test efficiency, while acknowledging potential environmental differences for further investigation, if needed, in the future.

Original bug: https://bugzilla.redhat.com/show_bug.cgi?id=1912413

### DIFF
--- a/test/extended/router/h2spec.go
+++ b/test/extended/router/h2spec.go
@@ -3,6 +3,7 @@ package router
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
 	"sort"
 	"strconv"
@@ -470,7 +471,7 @@ BFNBRELPe53ZdLKWpf2Sr96vRPRNw
 				return
 			}
 
-			testSuites, err := runConformanceTests(oc, host, "h2spec", 5*time.Minute)
+			testSuites, err := runConformanceTests(oc, host, "h2spec", 10*time.Minute)
 			o.Expect(err).NotTo(o.HaveOccurred())
 			o.Expect(testSuites).ShouldNot(o.BeEmpty())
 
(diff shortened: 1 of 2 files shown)


candita:
See https://github.com/openshift/origin/pull/26089

/approve
/lgtm

candita:
This probably requires https://github.com/openshift/cloud-provider-aws/pull/57

frobware:
/hold

I think the consensus was that this PR still requires https://github.com/openshift/cloud-provider-aws/pull/57.

frobware:
Slack discussion: https://redhat-internal.slack.com/archives/CBWMXQJKD/p1704908895477469.

frobware:
> /hold
> 
> I think the consensus was that this PR still requires [openshift/cloud-provider-aws#57](https://github.com/openshift/cloud-provider-aws/pull/57).

57^ has merged. 

/test all

frobware:
failed to initialize the cluster: Cluster operators authentication, console, control-plane-machine-set, image-registry, ingress, machine-api, monitoring are not available

/test e2e-gcp-ovn-upgrade

frobware:
/hold cancel

From https://redhat-internal.slack.com/archives/CBWMXQJKD/p1715681553819319?thread_ts=1704908895.477469&cid=CBWMXQJKD

> we still don't have an accepted nightly build that including the pr, but I ran flexy job and installed one aws upi cluster with 4.16.0-0.nightly-2024-05-13-102953 (rejected) today. I believe the issue is fixed, ingresscontroller as well as LB service can be deleted within about 1'20'' and no k8s rules leaking in Security Groups.

cc @lihongan

PR #28516, opened 2024-01-10
allow topology aware hints failures when pods are evicted

deads2k:
building a proper count is better, but this will allow us to get green component readiness signal since the events are expected when pods are evicted.


deads2k:
superceded by https://github.com/openshift/origin/pull/28517

PR #28521, opened 2024-01-11
NO-JIRA: add kubernetes level to openshift tests image for payload

deads2k:
This adds a kubectl version to the Dockerfile so that the release controller can better represent what level of kube is present in our payloads. As our mismatch times extend, this makes certain classes of failures obvious and ensure we don't have an accident shipping.

/assign @soltysh @wking


deads2k:
/retest

now that the `oc` PR has merged.

soltysh (approved):
/lgtm

PR #28514, opened 2024-01-10
NE-1444: test/extended/router: HAProxy 2.8 Compatibility Enhancements for h2spec tests

frobware:
This PR introduces two updates to the HAProxy configuration and the associated test suite for the h2spec tests in test/extended/router.

1. **Update to HAProxy Configuration:**
   The HAProxy configuration in `test/extended/router/h2spec.go` is updated to replace the deprecated directive `tune.ssl.capture-cipherlist-size` with `tune.ssl.capture-buffer-size`. This change is in response to deprecation warnings from HAProxy versions 2.8.5 and 2.6.13, ensuring compatibility with both versions.

2. **Removal of Known h2spec Failures for HAProxy 2.6 and 2.8:**
   The h2spec tests "http2/5.4.1.2" and "http2/6.9.1.2", previously known to fail or flake in HAProxy 2.2, have been removed from the list of known failures. The removal is based on successful test runs with HAProxy versions 2.6 and 2.8, indicating that these issues have been resolved in these newer versions. I ran the h2spec tests with an increased sample size (ROUTER_H2SPEC_SAMPLE set to 100 instead of the usual 1) and these failures were not observed.

Requires openshift/router#551.

PR #28522, opened 2024-01-12
NE-1444: test/extended/router/http2: delete skipped tests

frobware:
This PR addresses a specific issue with the HTTP/2 tests in test/extended/router. Due to current limitations in the haproxy.config configuration, two tests - `http2-default-cert-edge` and `http2-default-cert-reencrypt` - cannot accurately assert that the frontend protocol is HTTP/2.0. As a temporary measure, this PR removes these tests. It's important to emphasise that this change is only provisional, and we plan to resurrect these tests once the underlying issue with haproxy.config is resolved.

https://issues.redhat.com/browse/OCPBUGS-27210 is a blocker bug for OCP 4.16. If we haven't fixed these two tests w.r.t HAProxy 2.8 then these changes should be reverted.

### Context
Our current haproxy configuration prevents us from asserting the frontend protocol as HTTP/2.0 when testing a route that uses the default certificate (with HTTP/2 enabled on the ingress controller). This limitation has led to the skipping of the `http2-default-cert-edge` and `http2-default-cert-reencrypt` tests for a long time. The forthcoming [openshift/router#554](https://github.com/openshift/router/pull/554) aims to address this by adjusting the haproxy configuration.

### Staging Problem
We face a staging challenge: if openshift/router#554 is merged first, the GET requests in these tests will succeed, and we will no longer skip them based on the expected error. However, the test expectations for `http2-default-cert-{edge,reencrypt}` are not currently set up to handle this change, leading to test failures.

### Temporary Solution
We plan to reintroduce these tests as soon as the haproxy configuration is updated. The resurrection of these tests will occur in a separate PR ([openshift/origin#28514](https://github.com/openshift/origin/pull/28514)), where we will reinstate and adjust the tests to align with the new haproxy configuration.

### DIFF
--- a/test/extended/router/http2.go
+++ b/test/extended/router/http2.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
-	"strings"
 	"time"
 
 	g "github.com/onsi/ginkgo/v2"
@@ -391,7 +390,6 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 				backendProto      string
 				statusCode        int
 				useHTTP2Transport bool
-				expectedGetError  string
 			}{{
 				route:             "http2-custom-cert-edge",
 				frontendProto:     "HTTP/2.0",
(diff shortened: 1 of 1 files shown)


Miciah:
/approve
/lgtm
We have [OCPBUGS-27210](https://issues.redhat.com/browse/OCPBUGS-27210) to force us to revisit this test before we ship OCP 4.16 or HAProxy 2.8, so let's merge this PR to unblock NE-1444.

PR #28523, opened 2024-01-12
Bump library-go to latest master

vrutkovs:
Testbed for https://github.com/openshift/library-go/pull/1661

### DIFF
--- a/pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
+++ b/pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
@@ -14,6 +14,7 @@ import (
 	"github.com/openshift/origin/pkg/clioptions/iooptions"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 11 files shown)


pkg/cmd/openshift-tests/collect-disk-certificates/collector.go
```diff
@@ -161,27 +157,16 @@ func (o *RunCollectDiskCertificatesOptions) Run(ctx context.Context) error {
...
-		masters = append(masters, &nodeList.Items[i])
-	}
-
 	for _, srcDir := range o.CollectDirs {
-		dirPKIList, err := certgraphanalysis.GatherCertsFromDisk(ctx, o.KubeClient, o.Prefix, srcDir,
+		dirPKIList, err := certgraphanalysis.GatherCertsFromDisk(ctx, o.KubeClient, srcDir,
 			certgraphanalysis.ElideProxyCADetails,
 			certgraphanalysis.SkipRevisioned,
 			certgraphanalysis.SkipHashed,
-			certgraphanalysis.RewriteNodeIPs(masters))
```
deads2k, line 180: expected to continue rewriting node ips

vrutkovs: They would be rewritten in the final report - but not in the report sent from the node. This would allow us to avoid requiring node access from the pod and be able to process it once at deduplication stage

PR #28539, opened 2024-01-22
update generators to handle ondisk resources

deads2k:
generator isn't yet using the on-disk certificates.  Types are missing from library-go


deads2k:
/hold

really want to remove the on-disk revisions.

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
```diff
@@ -22,6 +22,19 @@ func AnnotationValue(whitelistedAnnotations []certgraphapi.AnnotationValue, key
 	return "", false
 }
 
+func DescriptionFor(in []certgraphapi.AnnotationValue) string {
+	ret, _ := AnnotationValue(in, annotations.OpenShiftDescription)
+	return ret
+}
+
+func OwnerFor(in []certgraphapi.AnnotationValue) string {
```
vrutkovs, line 30: That seems to treat "Unknown" as string, so markdown report now doesn't have those listed first, which is a handy feature.

Imo it should return `(string, bool)` and have those listed separate from violations

PR #28543, opened 2024-01-23
TRT-1452: Separate excessive Back-off restarting tests via namespace

DennisPeriquet:
This introduces `NamespacedTest()` which can be used where the `Test()` method is used if we want to split out more pathological tests by namespace.

To test, we need this test to fail on multiple namespaces:

```
[sig-cluster-lifecycle] pathological event should not see excessive Back-off restarting failed containers
```

### DIFF
--- a/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
+++ b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
@@ -141,20 +141,16 @@ func generateJUnitTestCasesCoreNamespaces(testName string, nsResults map[string]
 	for namespace := range namespaces {
 		jUnitName := getJUnitName(testName, namespace)
 		if result, ok := nsResults[namespace]; ok {
-			if len(result.failures) == 0 && len(result.flakes) == 0 {
+			output := generateFailureOutput(result.failures, result.flakes)
+			tests = append(tests, &junitapi.JUnitTestCase{
+				Name: jUnitName,
+				FailureOutput: &junitapi.FailureOutput{
+					Output: output,
+				},
+			})
+			// Add a success for flakes
+			if len(result.failures) == 0 && len(result.flakes) > 0 {
 				tests = append(tests, &junitapi.JUnitTestCase{Name: jUnitName})
-			} else {
-				output := generateFailureOutput(result.failures, result.flakes)
-				tests = append(tests, &junitapi.JUnitTestCase{
-					Name: jUnitName,
-					FailureOutput: &junitapi.FailureOutput{
-						Output: output,
-					},
-				})
-				// Add a success for flakes
-				if len(result.failures) == 0 && len(result.flakes) > 0 {
-					tests = append(tests, &junitapi.JUnitTestCase{Name: jUnitName})
-				}
 			}
 		} else {
 			tests = append(tests, &junitapi.JUnitTestCase{Name: jUnitName})
--- a/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
+++ b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
@@ -141,16 +141,20 @@ func generateJUnitTestCasesCoreNamespaces(testName string, nsResults map[string]
 	for namespace := range namespaces {
 		jUnitName := getJUnitName(testName, namespace)
 		if result, ok := nsResults[namespace]; ok {
-			output := generateFailureOutput(result.failures, result.flakes)
-			tests = append(tests, &junitapi.JUnitTestCase{
-				Name: jUnitName,
-				FailureOutput: &junitapi.FailureOutput{
-					Output: output,
-				},
-			})
-			// Add a success for flakes
-			if len(result.failures) == 0 && len(result.flakes) > 0 {
+			if len(result.failures) == 0 && len(result.flakes) == 0 {
 				tests = append(tests, &junitapi.JUnitTestCase{Name: jUnitName})
+			} else {
+				output := generateFailureOutput(result.failures, result.flakes)
+				tests = append(tests, &junitapi.JUnitTestCase{
+					Name: jUnitName,
+					FailureOutput: &junitapi.FailureOutput{
+						Output: output,
+					},
+				})
+				// Add a success for flakes
+				if len(result.failures) == 0 && len(result.flakes) > 0 {
+					tests = append(tests, &junitapi.JUnitTestCase{Name: jUnitName})
+				}
 			}
 		} else {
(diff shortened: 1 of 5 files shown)


pkg/monitortests/node/legacynodemonitortests/pathological_events_test.go
```diff
@@ -108,23 +110,34 @@ func Test_testBackoffStartingFailedContainer(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			e := monitorapi.Intervals{tt.interval}
 			junits := testBackoffStartingFailedContainer(e)
+
+			// Find the junit with the namespace of openshift-etcd-operator int the testname
+			var testJunits []*junitapi.JUnitTestCase
+			for _, j := range junits {
+				if strings.Contains(j.Name, "openshift-etcd-operator") {
```
neisw, line 117: Probably better to pass in the namespace you want to find as part of the tests struct in case there is a want / need to test for other namespaces later on

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
```diff
@@ -141,16 +141,20 @@ func generateJUnitTestCasesCoreNamespaces(testName string, nsResults map[string]
...
-			output := generateFailureOutput(result.failures, result.flakes)
-			tests = append(tests, &junitapi.JUnitTestCase{
-				Name: jUnitName,
-				FailureOutput: &junitapi.FailureOutput{
-					Output: output,
-				},
-			})
-			// Add a success for flakes
-			if len(result.failures) == 0 && len(result.flakes) > 0 {
+			if len(result.failures) == 0 && len(result.flakes) == 0 {
```
neisw, line 144: Curious about this change, [generateFailureOutput](https://github.com/DennisPeriquet/origin/blob/a4f1e13bbff4eb91494779f68ad5e0414449eb5b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go#L124) looks like it returns an empty string if no flakes or failures, did you see failures recorded when there weren't any?

pkg/monitortests/node/legacynodemonitortests/pathological_events_test.go
DennisPeriquet: that makes sense. thx

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go
DennisPeriquet: yes, it returns a "" which means that when we create the `JUnitTestCase` in https://github.com/openshift/origin/blob/a4f1e13bbff4eb91494779f68ad5e0414449eb5b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events.go#L148C1-L153C7, we end up with a `FailureOutput` which results in a failed `JUnitTestCase` as noted in my first set of test runs including [this job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/28543/pull-ci-openshift-origin-master-e2e-agnostic-ovn-cmd/1749923124394070016).  This is why I have the line at 145 to create a success `JUnitTestCase` where the FailureOutput will be nil.  This way, the test will show as a success.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_events_special.go
```diff
@@ -60,6 +61,45 @@ func (s *singleEventThresholdCheck) Test(events monitorapi.Intervals) []*junitap
...
+			switch {
+			case s.failThreshold > 0 && times > s.failThreshold:
+				nsResults[namespace].failures = append(nsResults[namespace].failures, fmt.Sprintf("event [%s] happened %d times", msg, times))
+			case times > s.flakeThreshold:
+				nsResults[namespace].flakes = append(nsResults[namespace].flakes, fmt.Sprintf("event [%s] happened %d times", msg, times))
+			}
+		}
+	}
+
+	for ns, result := range nsResults {
```
neisw, line 93: Wondering if you can do it without double processing the loop, initializing all namespaces then removing the empty ones

If you had a function like below could you call
nsResults[namespace] = addEventResult(nsResults[namespace], fmt.Sprintf("event [%s] happened %d times", msg, times), true)

etc.

func addEventResult(nsResult *eventResult{}, msg string, flake bool) *eventResult {
  if nsResult == nil {
     nsResult = &eventResult{}
  }

 if flake {
  nsResult.flakes = append(nsResult.flakes, msg)
} else {
  nsResult.failures = append(nsResult.failures, msg)
}
return nsResult

}

neisw: That snippet won't work as is, you would probably have to pass in the map and the namespace and then do the check in the function.  Just doesn't feel right initializing it all and then removing the empty ones.. 

if _, ok := nsResults[namespace]; !ok {
			tmp := &eventResult{}
			nsResults[namespace] = tmp
		}

neisw: [addEventResult_function.txt](https://github.com/openshift/origin/files/14058922/addEventResult_function.txt)

wonder if this patch would do it

DennisPeriquet: >> Just doesn't feel right initializing it all and then removing the empty ones.

Good instinct.

I pushed a commit that will track only namespaces with failures or flakes so that we only traverse the loop once.

pkg/monitortests/node/legacynodemonitortests/pathological_events_test.go
```diff
@@ -108,23 +113,34 @@ func Test_testBackoffStartingFailedContainer(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			e := monitorapi.Intervals{tt.interval}
 			junits := testBackoffStartingFailedContainer(e)
+
+			// Find the junit with the namespace of openshift-etcd-operator int the testname
+			var testJunits []*junitapi.JUnitTestCase
+			for _, j := range junits {
+				if strings.Contains(j.Name, namespace) {
```
neisw, line 120: Here what I was thinking was the test would have `namespace: "openshift-etcd-operator"` and the logic would be

if strings.Contains(j.Name, tt.namespace) 

Then if the test was updated later on to test another namespace it would handle it.

But no need to hold this up for that, if / when there is a need it is easy enough to refactor.

DennisPeriquet:
FYI: I'm not saying this is a critical fix; I'm only trying to add enough label so this PR can get merged.  This PR predates the enactment of the `acknowledge-critical-fixes` so it should not matter.

PR #28551, opened 2024-01-25
WIP: adjust types to accomodate for ondisk-only certs/cabundles

vrutkovs:


pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/ownership/requirement.go
```diff
@@ -52,14 +52,22 @@ func generateViolationJSON(pkiInfo *certgraphapi.PKIRegistryInfo) *certgraphapi.
...
+		}
+		owner := certKeyInfo.OwningJiraComponent
 		if len(owner) == 0 || owner == tlsmetadatainterfaces.UnknownOwner {
 			ret.CertKeyPairs = append(ret.CertKeyPairs, curr)
 		}
 	}
 	for i := range pkiInfo.CertificateAuthorityBundles {
 		curr := pkiInfo.CertificateAuthorityBundles[i]
-		owner := curr.CABundleInfo.OwningJiraComponent
+		caBundleInfo, err := tlsmetadatainterfaces.CertificateAuthorityInfoForCABundle(curr)
```
deads2k, line 66: don't like this block with an error.  We know the metadata at this point.  I resolved with https://github.com/openshift/origin/pull/28539/files#diff-b2792dc75c2beb82a9699af3c941303e12704132ebf6a6f102ccc7069a408a7dR18

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
```diff
@@ -23,36 +23,137 @@ func AnnotationValue(whitelistedAnnotations []certgraphapi.AnnotationValue, key
...
+						isAlsoInCluster = len(caBundle.Spec.ConfigMapLocations) > 0
+						break
+					}
+				}
+				if found {
+					break
+				}
+			}
+			if found {
+				if !isAlsoInCluster {
```
deads2k, line 93: I don't see why it matters whether it's also in the cluster.

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
```diff
@@ -23,36 +23,137 @@ func AnnotationValue(whitelistedAnnotations []certgraphapi.AnnotationValue, key
...
 		return nil, utilerrors.NewAggregate(errs)
 	}
 
-	return certs.CertsToRegistryInfo(certKeyPairs, caBundles), nil
+	return certs.CertsToRegistryInfo(inClusterCertKeyPairs, onDiskCertsWithoutSecrets, inClusterCABundles, onDiskCABundlesWithoutConfigMaps), nil
+}
+
+// TODO[vrutkovs]: move this to /api?
+
+func CertInfoForCertKeyPair(curr certgraphapi.PKIRegistryCertKeyPair) (certgraphapi.PKIRegistryCertKeyPairInfo, error) {
```
deads2k, line 114: for printing purposes. these are very different and incomparable values.  Notice how the division is kept through to the printing layer here https://github.com/openshift/origin/pull/28539/files#diff-b2792dc75c2beb82a9699af3c941303e12704132ebf6a6f102ccc7069a408a7dR178 to avoid losing the information early.

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
vrutkovs: on-disk-but-also-in-the-cluster artifacts have metadata set in annotations. on-disk-only artifacts need to be looked up in `onDiskCertsWithoutSecrets` / `onDiskCABundlesWithoutConfigMaps` maps. This set needs to be kept minimal as its going to be stored in `origin` (even `api` maybe?)

deads2k:
is this still a WIP?

pkg/certs/types.go
```diff
@@ -6,10 +6,21 @@ import (
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
 )
 
+// PKIRegistryInfo holds information about TLS artifacts stored in etcd. This includes object location and metadata based on object annotations
+type PKIRegistryInfo struct {
```
deads2k, line 10: Can this type be entirely contained to origin since we don't produce it raw and we're going to generate docs from here?

vendor/github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi/type_registry.go
```diff
@@ -40,6 +53,19 @@ type PKIRegistryInClusterCABundle struct {
 	CABundleInfo PKIRegistryCertificateAuthorityInfo `json:"certificateAuthorityBundleInfo"`
 }
 
+type PKIRegistryCABundle struct {
```
deads2k, line 56: missing union discriminator.

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
deads2k: > on-disk-but-also-in-the-cluster artifacts have metadata set in annotations. on-disk-only artifacts need to be looked

I don't think this will work in practice since we see duplicate CA bundles owned by different components.  I suspect we'll end up needing to remove this.

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
```diff
@@ -21,38 +23,299 @@ func AnnotationValue(whitelistedAnnotations []certgraphapi.AnnotationValue, key
...
+	onDiskCertificateAuthorities = []certgraphapi.PKIRegistryOnDiskCABundle{
+		{
+			OnDiskLocation: certgraphapi.OnDiskLocationWithMetadata{
+				OnDiskLocation: certgraphapi.OnDiskLocation{
+					Path: "/etc/kubernetes/ca.crt",
+				},
+				User:           "root",
+				Group:          "root",
+				Permissions:    "-rw-r--r--",
+				SELinuxOptions: "system_u:object_r:kubernetes_file_t:s0",
```
deads2k, line 125: this information should be coming from the raw tls data.

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
```diff
@@ -21,38 +23,299 @@ func AnnotationValue(whitelistedAnnotations []certgraphapi.AnnotationValue, key
...
 		return nil, utilerrors.NewAggregate(errs)
 	}
 
-	return certs.CertsToRegistryInfo(certKeyPairs, caBundles), nil
+	return certs.CertsToRegistryInfo(inClusterCertKeyPairs, onDiskCertsWithoutSecrets, inClusterCABundles, onDiskCABundlesWithoutConfigMaps), nil
+}
+
+// TODO[vrutkovs]: move this to /api?
+var (
+	onDiskCertificateAuthorities = []certgraphapi.PKIRegistryOnDiskCABundle{
```
deads2k, line 116: Is this trying to be a logical map from `ondisk.path` to `SelectedCertMetadataAnnotations []AnnotationValue`?

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/annotation_requirement.go
```diff
@@ -63,38 +64,44 @@ func (o annotationRequirement) InspectRequirement(rawData []*certgraphapi.PKILis
...
+	violatingCABundlesByOwner := complianceIntermediate.ViolatingCABundlesByOwner
 
 	md := NewMarkdown(o.title)
 	md.Title(2, "How to meet the requirement")
 	md.ExactText(o.explanationMD)
 
+	certsWithoutOwners := violatingCertsByOwner[UnknownOwner]
+	caBundlesWithoutOwners := violatingCABundlesByOwner[UnknownOwner]
+
+	if len(certsWithoutOwners) > 0 || len(caBundlesWithoutOwners) > 0 {
```
deads2k, line 81: This makes sense in ownership, but I think we want to see the main sort remain, compliant or non-compliant in places like auto-regenerate-after-expiry

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/markdown_rendering.go
```diff
@@ -0,0 +1,188 @@
...
+	case InCluster:
+		md.Textf("ns/%v %v/%v\n", location.InClusterLocation.Namespace, location.InClusterLocation.ResourceType, location.InClusterLocation.Name)
+		md.Textf("**Description:** %v", DescriptionFor(location.InClusterLocation.ResourceMetadata))
+		md.Text("\n")
+	case OnDisk:
+		md.Textf("%v\n", location.OnDiskLocation.FileMetadata.Path)
+		// TODO where to store descriptions
+		md.Textf("**Permission:** %v\n", location.OnDiskLocation.FileMetadata.Permissions)
+		md.Textf("**User:** %v\n", location.OnDiskLocation.FileMetadata.User)
+		md.Textf("**Group %v\n", location.OnDiskLocation.FileMetadata.Group)
```
deads2k, line 184:
```suggestion
		md.Textf("**Group** %v\n", location.OnDiskLocation.FileMetadata.Group)
```

pkg/certs/types.go
vrutkovs: We could `PKIList` there instead

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
vrutkovs: This is the expected file properties, raw TLS data will contain actual data

pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadatainterfaces/helpers.go
vrutkovs: Pretty much,  yes. Do we want to represent these attributed as synthetic annotations?

vrutkovs: That would make rendering and comparison a bit easier, but in 
```
type CertKeyPairInfoByOnDiskLocation map[certgraphapi.OnDiskLocationWithMetadata]certgraphapi.PKIRegistryCertKeyPairInfo
```
it cannot be used as a comparable key, so filtering would become much more complicated

vrutkovs:
Included in https://github.com/openshift/origin/pull/28657

PR #28556, opened 2024-01-26
Live migration: if managed namespace doesnt exist, create it

martinkennelly:
Live migration has added an extra requirement that it may only begin if the cluster is a managed cluster (SD).
This is to prevent self managed clusters from using this func as we do not support it.

See PR: https://github.com/openshift/cluster-network-operator/pull/2225

and slack conversation https://redhat-internal.slack.com/archives/CFJD1NZFT/p1706207602864769?thread_ts=1706194037.470999&cid=CFJD1NZFT

/cc @pliurh

### DIFF
--- a/test/extended/networking/cnimigration.go
+++ b/test/extended/networking/cnimigration.go
@@ -10,10 +10,12 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	networkopernames "github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/origin/test/extended/util"
 	"github.com/openshift/origin/test/extended/util/operator"
 
 	corev1 "k8s.io/api/core/v1"
 	eventsv1 "k8s.io/api/events/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


deads2k:
Live migration runs from a separate job in CI, right?  Add the namespace creation here to the job set up script.

martinkennelly:
/retest 

" Pod scheduling timeout."

martinkennelly:
> Live migration runs from a separate job in CI, right? Add the namespace creation here to the job set up script.

@deads2k Yes it runs as a seperate job. You want the namespace creation to be part of the job itself and not here?
This PR creates the NS if it doesnt exist.

deads2k:
> @deads2k Yes it runs as a seperate job. You want the namespace creation to be part of the job itself and not here?
> This PR creates the NS if it doesnt exist.

Yes, I think that would be better.  Our expected state will be having this functionality ungated when we're ready to allow it for all clusters.

martinkennelly:
> > @deads2k Yes it runs as a seperate job. You want the namespace creation to be part of the job itself and not here?
> > This PR creates the NS if it doesnt exist.
> 
> Yes, I think that would be better. Our expected state will be having this functionality ungated when we're ready to allow it for all clusters.

Ack - i agree. Let me update this PR to skip if the namespace isnt present.

PR #28557, opened 2024-01-26
ETCD-512: Increase thresholds for etcd RequiredInstallerResourcesMissing

tjungblu:
This is to avoid failing nightlies after we merge the first cert rotation stories.


tjungblu:
/hold

may have a fix in our operator already

PR #28559, opened 2024-01-26
Live migration: dont rollback within the same test as migration

martinkennelly:
After merging the code and applying the jobs for migration and rollback, we noticed that monitors have a toleration for disruption. This toleration was configured for upgrade. For our job, we perform migration and rollback.
Its expected the disruption time will be larger.

Instead of configuring the monitors to have a
custom disruption duration, simply call the test twice for migration and rollback and have new instances of the monitors.

This PR performs the migration of the CNI when called. Within the CI jobs, we can simply call this test twice to perform migration and rollback.

A point to note is that this test is active within the 'all' suite.

### DIFF
--- a/test/extended/networking/cnimigration.go
+++ b/test/extended/networking/cnimigration.go
@@ -3,7 +3,6 @@ package networking
 import (
 	"context"
 	"fmt"
-	"os"
 	"strings"
 	"time"
 
@@ -12,6 +11,8 @@ import (
 	networkopernames "github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/origin/test/extended/util/operator"
 
+	g "github.com/onsi/ginkgo/v2"
+	"github.com/pborman/uuid"
 	corev1 "k8s.io/api/core/v1"
 	eventsv1 "k8s.io/api/events/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


martinkennelly:
/pj-rehearse pull-ci-openshift-ovn-kubernetes-release-4.16-e2e-aws-live-migration-sdn-ovn

dgoodwin:
Jumping ahead, I think we need to find a way to (a) disable the disruption testing for your job based on something we can identify in https://github.com/openshift/origin/blob/master/pkg/monitortestlibrary/disruptionlibrary/disruption_invariant_adapter.go#L224 to skip those tests.

This will allow us to still collect data for your job so we could answer the question, how much disruption would a customer expect going through this process. 

Then (b) we need to filter our your jobs via a consistent naming scheme in our scheduled queries that report disruption data daily. This will keep your results out of the dashboard and regression detection.

These tests can't be a part of the all suite afaict, they would need to be their own. Does this sound reasonable?

martinkennelly:
> Can you paste links to job runs where you are noticing increased disruption?

https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-e2e-aws-live-migration-sdn-ovn-rollback

> Does the team feel it's acceptable to be taking this increased hit in disruption during this migration?

This increased disruption sometimes happens (1/4 or 1/5) when we migrate to a CNI and then rollback. So essentially its two migrations. We do not see it during a single migration. All is good there. 

> Creating the monitors multiple times is unlikely to help as the data is still aggregated for the job run and submitted to the bigquery database, which later feeds into the limits the tests will fail on.


I see. Thanks for this info. I thought every time we exec openshift-tests run .. within the job, it would aggregate and calculate disruption for each invocation.

martinkennelly:
> These tests can't be a part of the all suite afaict, they would need to be their own. Does this sound reasonable?

yes

PR #28560, opened 2024-01-26
OCPBUGS-23115: Test builds with subscription content

adambkaplan:
Add end to end tests to verify we can build images that install
RHEL subscription content via `yum/dnf install`. These tests
need a fully subscribed cluster to function properly - for example,
a ROSA cluster. If the cluster is not subscribed/does not have its
SCA certificates synced via the Insights Operator, the tests are
skipped.

The test verifies the instructions in the official OpenShift
documentation, which includes copying the entitlement keys to
the build's namespace. Docs are currently being drafted in https://github.com/openshift/openshift-docs/pull/70897

### DIFF
--- a/test/extended/builds/subscription_content.go
+++ b/test/extended/builds/subscription_content.go
@@ -1,0 +1,93 @@
+package builds
+
+import (
+	"path/filepath"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


adambkaplan:
@bparees @deads2k who owns the contributor experience for openshift/origin these days?

I'm running into a thorny Ginkgo issue - at https://github.com/openshift/origin/pull/28560/commits/7bc0f7f4927cb91ed7af56758becf65ca430c9c3, my new specs register and get pulled into the openshift/build suite.

But once I start implementing actual logic (current branch state), the specs disappear and don't show up in any test suite. The fundamental structure of my test doesn't seem to have changed - what broke here?

adambkaplan:
A follow up pull request should verify the procedure for using entitlements via the shared resource CSI driver.

PR #28561, opened 2024-01-29
deploy: change imagestreams importMode to preserveOriginal only when working with multi-arch compute clusters

Prashanth684:
Following discussions around #27864, rather than changing importMode unconditionally, change it to import manifestlists only when the  installed cluster has nodes of differing architectures

### DIFF
--- a/test/extended/deployments/deployments.go
+++ b/test/extended/deployments/deployments.go
@@ -108,6 +108,23 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 		imageChangeTriggerFixture       = exutil.FixturePath("testdata", "deployments", "deployment-trigger.yaml")
 	)
 
+	// Check if there are nodes of different architectures
+	isMultiArchCluster := false
+	output, err := oc.WithoutNamespace().AsAdmin().Run("get").Args("nodes", "-o=jsonpath={.items[*].status.nodeInfo.architecture}").Output()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


Prashanth684:
> /cc @deads2k this follows up on [this ](https://github.com/openshift/origin/pull/27864#issuecomment-1611764300) discussion to add the importMode option only when the cluster in question uses the multi payload (which is used by heterogeneous clusters today)

I have modified this now to actually check if the cluster has differing arch nodes to make this decision

test/extended/deployments/deployments.go
```diff
@@ -108,6 +108,23 @@ var _ = g.Describe("[sig-apps][Feature:DeploymentConfig] deploymentconfigs", fun
 		imageChangeTriggerFixture       = exutil.FixturePath("testdata", "deployments", "deployment-trigger.yaml")
 	)
 
+	// Check if there are nodes of different architectures
+	isMultiArchCluster := false
+	output, err := oc.WithoutNamespace().AsAdmin().Run("get").Args("nodes", "-o=jsonpath={.items[*].status.nodeInfo.architecture}").Output()
```
deads2k, line 113: do this part with an `AdminKubeClient` and `.Core().Nodes().List()`

deads2k:
I'm still not clear on why it's a good thing that our documentation, existing blogs, and existing scripts will not be portable to heterogeneous clusters.  Can we update the description to explain why this is better than making our existing collateral "just work"?

PR #28584, opened 2024-02-09
SPLAT-1439: Ignore edge workers when listing nodes

jcpowermac:


mtulio (commented):
I'd suggest remove from the list as it seems valid for run on edge workloads:
- test/extended/kernel/common.go
- test/extended/machines/workers.go
- test/extended/olm/olm.go
- test/extended/security/labels.go
- test/extended/security/fips.go

Other than that it seems to be fine, WDYT?

jcpowermac:
> I'd suggest remove from the list as it seems valid for run on edge workloads:
> 
>     * test/extended/kernel/common.go
> 
>     * test/extended/machines/workers.go
> 
>     * test/extended/olm/olm.go
> 
>     * test/extended/security/labels.go
> 
>     * test/extended/security/fips.go
> 
> 
> Other than that it seems to be fine, WDYT?

sgtm, updated

mtulio:
Both e2e for AWS and edge zones are passing as expected.
/lgtm

PR #28586, opened 2024-02-12
TRT-1499: Revert #28571 gcp-network-liveness until gcp resources are whitelisted

DennisPeriquet:
Until https://issues.redhat.com/browse/DPP-14288 is addressed, the resources created in our gcp account for the liveness endpoint on will get reaped.  As a result, the tests will fail.

For now, we remove the test.


dgoodwin:
/retitle TRT-1499: Revert #28571 gcp-network-liveness until gcp resources are whitelisted

PR #28589, opened 2024-02-12
AUTH-483: Add a monitor test to check the existence of required-scc annotation in platform workloads

liouk:
This PR adds a monitor test that requires the existence of the `openshift.io/required-scc` annotation on all platform workloads (workloads that belong to `openshift*`, `kube-*` and `default` namespaces).

If the annotation is missing, the test suggests an SCC to be pinned using the annotation, based on the following:
- suggested SCC is the one that was used to admit the workload, provided that it is a default SCC and not a custom one
- if it's a custom SCC, the test cannot suggest any; the developer must determine the appropriate SCC
- since SCC admission is disabled for runlevel 0/1 namespaces, the test won't suggest any SCC for their workloads; the developer must determine the appropriate one

At its current state, the test is implemented as a flake, until we've pinned SCCs to all platform workloads (see https://issues.redhat.com/browse/AUTH-482).

### DIFF
--- a/pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
+++ b/pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
@@ -1,0 +1,149 @@
+package requiredsccmonitortests
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	securityv1 "github.com/openshift/api/security/v1"
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/monitortestframework"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+var defaultSCCs = sets.NewString(
+	"anyuid",
+	"hostaccess",
+	"hostmount-anyuid",
+	"hostnetwork",
+	"hostnetwork-v2",
+	"nonroot",
+	"nonroot-v2",
+	"privileged",
+	"restricted",
+	"restricted-v2",
+)
+
+type requiredSCCAnnotationChecker struct {
+	kubeClient kubernetes.Interface
+}
+
+func NewAnalyzer() monitortestframework.MonitorTest {
+	return &requiredSCCAnnotationChecker{}
+}
+
+func (w *requiredSCCAnnotationChecker) StartCollection(ctx context.Context, adminRESTConfig *rest.Config, recorder monitorapi.RecorderWriter) error {
+	var err error
+	w.kubeClient, err = kubernetes.NewForConfig(adminRESTConfig)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageDir string, beginning, end time.Time) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error) {
+	if w.kubeClient == nil {
+		return nil, nil, nil
+	}
+
+	namespaces, err := w.kubeClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	junits := []*junitapi.JUnitTestCase{}
+	for _, ns := range namespaces.Items {
+		if !strings.HasPrefix(ns.Name, "openshift") && !strings.HasPrefix(ns.Name, "kube-") && ns.Name != "default" {
+			continue
+		}
+
+		pods, err := w.kubeClient.CoreV1().Pods(ns.Name).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return nil, nil, err
+		}
+
+		failures := make([]string, 0)
+		for _, pod := range pods.Items {
+			if _, exists := pod.Annotations[securityv1.RequiredSCCAnnotation]; exists {
+				continue
+			}
+
+			suggestedSCC := suggestSCC(&pod)
+			owners := ownerReferences(&pod)
+			failures = append(failures, fmt.Sprintf("annotation missing from pod '%s'%s; %s", pod.Name, owners, suggestedSCC))
+		}
+
+		testName := fmt.Sprintf("[sig-auth] all workloads in ns/%s must set the '%s' annotation", ns.Name, securityv1.RequiredSCCAnnotation)
+		if len(failures) == 0 {
+			junits = append(junits, &junitapi.JUnitTestCase{Name: testName})
+			continue
+		}
+
+		failureMsg := strings.Join(failures, "\n")
+		junits = append(junits,
+			&junitapi.JUnitTestCase{
+				Name:          testName,
+				SystemOut:     failureMsg,
+				FailureOutput: &junitapi.FailureOutput{Output: failureMsg},
+			},
+
+			// add a successful test with the same name to cause a flake
+			&junitapi.JUnitTestCase{
+				Name: testName,
+			},
+		)
+	}
+
+	return nil, junits, nil
+}
+
+func (w *requiredSCCAnnotationChecker) ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals, recordedResources monitorapi.ResourcesMap, beginning, end time.Time) (monitorapi.Intervals, error) {
+	return nil, nil
+}
+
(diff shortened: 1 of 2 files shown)


pkg/monitortests/clusterversionoperator/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,131 @@
...
+			msg := fmt.Sprintf("annotation missing from pod '%s'; suggested required-scc: '%s'", pod.Name, suggestedSCC)
+			if len(ownerRefs) > 0 {
+				msg += fmt.Sprintf("; owners: %s", strings.Join(ownerRefs, ", "))
+			}
+			failures = append(failures, msg)
+		}
+
+		testName := fmt.Sprintf("[sig-auth] all workloads in ns/%s must set the '%s' annotation", ns.Name, securityv1.RequiredSCCAnnotation)
+		junit := &junitapi.JUnitTestCase{Name: testName}
+		junits = append(junits, junit)
```
deads2k, line 103: please add after you're done mutating. Makes it easier to read

deads2k: oh, wow.  You're trying to do success adn then failure.  Yeah, definitely refactor.  Add a different instance entirely.

pkg/monitortests/clusterversionoperator/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,131 @@
+package requiredsccmonitortests
```
deads2k, line 1: pkg/monitortests/authentication/requiredsccmonitortests/monitoretest.go

pkg/monitortests/clusterversionoperator/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,131 @@
...
+	junits := []*junitapi.JUnitTestCase{}
+	for _, ns := range namespaces.Items {
+		if !strings.HasPrefix(ns.Name, "openshift") && !strings.HasPrefix(ns.Name, "kube-") && ns.Name != "default" {
+			continue
+		}
+
+		suggestedSCC := ""
+		switch ns.Labels["openshift.io/run-level"] {
+		case "0", "1":
+			suggestedSCC = "privileged"
```
deads2k, line 70: Use this only as a last resort if in 0 or 1.  Generally use the validatedSCC.  For example: kas-o should be restricted, not privileged even though it's got privileged powers.

liouk: Since it's not entirely clear on what SCC to suggest for runlevel NSes (as there's no validated SCC), I've chosen to drop the suggestion in this case.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,135 @@
...
+
+		failures := make([]string, 0)
+		for _, pod := range pods.Items {
+			if _, exists := pod.Annotations[securityv1.RequiredSCCAnnotation]; exists {
+				continue
+			}
+
+			// suggest the assigned SCC only if it belongs to the default set of SCCs
+			// pods in runlevel 0/1 namespaces won't have any assigned SCC as SCC admission is disabled
+			suggestedSCC := ""
```
ibihim, line 80: Nit:
It could be useful to have a fallback value, such that someone reading that understands what `annotation missing from pod 'my-pod'; suggested requried-scc: ''` it means.

Maybe adjust the message to, "can't suggest SCC (non validated)", when `suggestedSCC == ""`, WDYT?

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,135 @@
...
+
+		failureMsg := strings.Join(failures, "\n")
+		junits = append(junits,
+			&junitapi.JUnitTestCase{
+				Name:          testName,
+				SystemOut:     failureMsg,
+				FailureOutput: &junitapi.FailureOutput{Output: failureMsg},
+			},
+
+			// add a successful test with the same name to cause a flake
```
ibihim, line 111: Useful comment :smile:

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,148 @@
...
+	return nil
+}
+
+// getSuggestedSCC suggests the assigned SCC only if it belongs to the default set of SCCs
+// pods in runlevel 0/1 namespaces won't have any assigned SCC as SCC admission is disabled
+func getSuggestedSCC(pod *v1.Pod) string {
+	if len(pod.Annotations[securityv1.ValidatedSCCAnnotation]) == 0 {
+		return "cannot suggest required-scc, no validated SCC on pod"
+
+	} else if defaultSCCs.Has(pod.Annotations[securityv1.ValidatedSCCAnnotation]) {
```
ibihim, line 130: nit: `if` is sufficient, you don't need to `else if`.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,148 @@
...
+	return nil
+}
+
+func (w *requiredSCCAnnotationChecker) Cleanup(ctx context.Context) error {
+	return nil
+}
+
+// getSuggestedSCC suggests the assigned SCC only if it belongs to the default set of SCCs
+// pods in runlevel 0/1 namespaces won't have any assigned SCC as SCC admission is disabled
+func getSuggestedSCC(pod *v1.Pod) string {
```
ibihim, line 126: nit: you could name it `suggestSCC` :smile:

ibihim (commented):
Looks amazing

/lgtm

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,149 @@
...
+
+		failureMsg := strings.Join(failures, "\n")
+		junits = append(junits,
+			&junitapi.JUnitTestCase{
+				Name:          testName,
+				SystemOut:     failureMsg,
+				FailureOutput: &junitapi.FailureOutput{Output: failureMsg},
+			},
+
+			// add a successful test with the same name to cause a flake
```
stlaz, line 98: why are you interested in causing a flake rather than a proper failure?

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -0,0 +1,149 @@
...
+			if _, exists := pod.Annotations[securityv1.RequiredSCCAnnotation]; exists {
+				continue
+			}
+
+			suggestedSCC := suggestSCC(&pod)
+			owners := ownerReferences(&pod)
+			failures = append(failures, fmt.Sprintf("annotation missing from pod '%s'%s; %s", pod.Name, owners, suggestedSCC))
+		}
+
+		testName := fmt.Sprintf("[sig-auth] all workloads in ns/%s must set the '%s' annotation", ns.Name, securityv1.RequiredSCCAnnotation)
```
stlaz, line 84: shouldn't the test name be stable, and we'd list the failures in the outputs?

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
liouk: This way we'll be able to merge this before fixing all namespaces/workloads, and once we're done with pinning we can turn this into a proper failure.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
liouk: I opted for a test per NS as I felt it would make any failures more easily visible in the test output. Is it more common in similar cases to have a single test and enumerate all failures?

liouk: Also, I believe that using the specific namespace in the test name as `ns/<name>` helps assign any failures to namespace owners.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
stlaz: Ok, that makes sense. In the past we used to just fix everything before test merge but I don't think we had a choice. I can see the advantage of this proposed kind of approach.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
stlaz: I see this is actually a pattern in our tests now, feel free to keep your testnames.

deads2k: per ns is perfect. that results in an auto-assign to the correct component in component readiness.

deads2k:
/lgtm
/approve

I'd like to hear an update every week of how many unique namespaces are still failing and how many have been fixed so far.

PR #28591, opened 2024-02-13
WIP: Add oc adm release new test to check kubectl sanity

ardaguclu:


ardaguclu:
This test takes approximately 30 minutes, that means it is not a feasible test. I think, it is better to fix this problem permanently in oc rather than running a 30 minutes test each pre-submit.

/close

PR #28594, opened 2024-02-14
TRT-1493: Re-introduce the GCP liveness endpoint test

DennisPeriquet:
Revert of this revert: https://github.com/openshift/origin/pull/28586
Plus a commit to add the new GCP LB -- this one should not remain there.

- TODO
  - [x] confirm there is no junit for the endpoint liveness test
    - I do NOT see a test in the Passed section on the prow job for `gcp-network-liveness` (but I also didn't see one in the tests for https://github.com/openshift/origin/pull/28571)
  - [x] run /payload to ensure we don't take out payloads

### DIFF
--- a/pkg/defaultmonitortests/types.go
+++ b/pkg/defaultmonitortests/types.go
@@ -3,6 +3,7 @@ package defaultmonitortests
 import (
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


DennisPeriquet:
the metal failure has nothing to do with the GCP VM/LB

PR #28598, opened 2024-02-15
MULTIARCH-4160: Stop using the DetermineFromRelease function

tvardema:
Multi-arch compute clusters have an issue where the cluster version's image ref is single arch, so this change resolves the image ref without spinning up a pod.


trozet (approved):
/approve

PR #28606, opened 2024-02-21
NO-JIRA: Add tail to node-logs test to reduce the log size

ardaguclu:
This PR passes `--tail=1000` flag to `adm node-logs` test to reduce the unnecessary log size.


soltysh (approved):
/lgtm
/approve

PR #28607, opened 2024-02-21
NO-JIRA: Pass tail flag in adm node-logs to reduce log size

ardaguclu:
This PR adds `--tail=100` flag in `adm node-logs --boot=0` test to reduce the generated log size that
may put a pressure on CI pod's memory.


soltysh (approved):
/lgtm
/approve

PR #28627, opened 2024-02-28
TRT-1539: Do not let loki alerts fail tests

dgoodwin:
We saw yesterday with loki in full outage, this single test would fail
reporting an alert firing in openshift-e2e-loki due to daemon set
rollout stuck. Promtail pods would run but would never go ready because they
couldn't communicate with loki.

Filter out any alerts from openshift-e2e-loki in this test so we can
pass all tests even when loki is down. We were very close otherwise as
no other problems popped up.

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -551,7 +551,8 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 			}
 
 			tests := map[string]bool{
-				fmt.Sprintf(`ALERTS{alertname!~"%s",alertstate="firing",severity!="info"} >= 1`, strings.Join(allowedAlertNames, "|")): false,
+				// openshift-e2e-loki alerts should never fail this test, we've seen this happen on daemon set rollout stuck when CI loki was down.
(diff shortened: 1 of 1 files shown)


test/extended/prometheus/prometheus.go
```diff
@@ -551,7 +551,8 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 			}
 
 			tests := map[string]bool{
-				fmt.Sprintf(`ALERTS{alertname!~"%s",alertstate="firing",severity!="info"} >= 1`, strings.Join(allowedAlertNames, "|")): false,
+				// openshift-e2e-loki alerts should never fail this test, we've seen this happen on daemon set rollout stuck when CI loki was down.
+				fmt.Sprintf(`ALERTS{alertname!~"%s",alertstate="firing",severity!="info",namespace!="openshift-e2e-loki"} >= 1`, strings.Join(allowedAlertNames, "|")): false,
```
dgoodwin, line 555: I have confirmed that namespace!="openshift-e2e-loki" will still match alerts that do not have a namespace with some testing on our alertmanager / prom instance.

machine424: yes, not having a label is like having it set to empty string.
nit: It was out of date before this change, but if you could adjust the desc (`g.It("shouldn't report any alerts in firing state apart...`) of the test, it'd be great.

machine424: Good to know, thanks.

PR #28628, opened 2024-02-28
MON-1157: Revive prometheus metrics best practices

jan--f:


jan--f:
Fwiw this is for now just an attempt to revive code that Damien already wrote. I have not checked if revived code make sense in the current test setup. This might need significantly more work.

dgrisonnet:
The initial concern with that test was that it might create disruption in CI if not done properly. I don't really know what would be the safest way to introduce it, but at least you would need to update the exception list and then maybe make it flaky for a bit, or notify TRT to have them watch the test and revert in case something bad happens.

Another concern was that e2e tests might not catch all the metrics & labels since it depends on whether the scenario that triggers the generation of a particular timeserie is tested or not. But having at least some of the metrics checked is already better than nothing, so I don't think it make sense anymore.

PR #28631, opened 2024-02-29
NO-JIRA: Handle release info output streams separately

ardaguclu:
This PR handles release info command's output streams separately instead of merging them. Because
test does not need the data in `errOut`.


soltysh (approved):
/lgtm
/approve

PR #28639, opened 2024-03-06
API-1802: watchevents: add cert rotation events

vrutkovs:
library-go emits events when cert rotation happens, these should be displayed as interesting events.

[Example](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/release-openshift-origin-installer-launch-gcp-modern/1780541177255170048/artifacts/launch/openshift-e2e-test/artifacts/junit/e2e-timelines_everything_20240417-115106.html), filter by Interesting Events

### DIFF
--- a/pkg/monitortests/testframework/watchevents/event.go
+++ b/pkg/monitortests/testframework/watchevents/event.go
@@ -150,6 +150,8 @@ func recordAddOrUpdateEvent(
 				}
 			}
 		}
+	case "CABundleUpdateRequired", "SignerUpdateRequired", "TargetUpdateRequired", "CertificateUpdated", "CertificateRemoved", "CertificateUpdateFailed":
+		message = message.WithAnnotation(monitorapi.AnnotationInteresting, "true")
 	default:
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/monitortests/testframework/watchevents/event.go
```diff
@@ -150,6 +150,10 @@ func recordAddOrUpdateEvent(
 				}
 			}
 		}
+	case "CABundleUpdateRequired", "SignerUpdateRequired", "TargetUpdateRequired", "CertificateUpdated", "CertificateRemoved":
+		message = message.WithAnnotation(monitorapi.AnnotationInteresting, "true")
+	case "CertificateUpdateFailed":
+		message = message.WithAnnotation(monitorapi.AnnotationPathological, "true")
```
dgoodwin, line 156: AnnotationPathological is for events that re-occur over the threshold of 20x in a job run, it's not intended as a mechanism for severity. In your case it almost feels like you may want to add an invariant test that looks for these CertificateUpdateFailed events and fails, or perhaps just flakes, a test.

vrutkovs: so, instead of displaying these events, we should have a synthetic test which ensures these events didn't happen? Yeah, that would work I guess

dgoodwin: You can absolutely display them, just mark them interesting, not pathological.

Based on your changeover to using patho I assumed you consider this event quite serious, and if you did write a test that flakes if it sees one of those, it would help you see fleet wide how often that is happening.

vrutkovs: Okay, added it to interesting list. Will create a flaking test to find these events as a followup

soltysh (approved):
/lgtm
/approve

PR #28655, opened 2024-03-13
NO-JIRA: Use bound prometheus-k8s API tokens

sanchezl:
Change GetPrometheusSABearerToken to use TokenRequest API instead of expecting an already existing API token secret.


soltysh (approved):
/lgtm
/approve

PR #28657, opened 2024-03-14
OCPBUGS-34346: Regenerate TLS info

vrutkovs:
This updates TLS info, which includes new secrets from etcd

Requires https://github.com/openshift/library-go/pull/1717


test/extended/operators/certs.go
```diff
@@ -111,6 +142,10 @@ var _ = g.Describe(fmt.Sprintf("[sig-arch][Late][Jira:%q]", "kube-apiserver"), g
 			masters = append(masters, &nodeList.Items[i])
 		}
 
+		// etcd operator doesn't immediately remove the certificate for boostrap
+		err = cleanupEtcdCertificates(ctx, kubeClient, masters)
```
deads2k, line 146: openshift-tests cannot make lasting change to a cluster. Why is this necessary?

tls/descriptions/descriptions.md
```diff
@@ -69,754 +70,777 @@ To create a description, set the `openshift.io/description` annotation to the ma
...
 
       **Description:** 
       
 
-19. ns/openshift-kube-apiserver-operator secret/node-system-admin-signer
 
-      **Description:** 
+
+## Items That DO Meet the Requirement (124)
+### etcd (36)
```
deads2k, line 657: @hasbro17 @soltysh These are some pretty weak descriptions.  

A good descriptions is in the style of API documentation and includes

1. Which connections a CA bundle can be used to verify.
2. What kind of certificates a signer will sign for.
3. Which names and IPs a serving certificate terminates.
4. Which subject (user and group) a client certificate is created for.
5. Which binary and flags is this certificate wired to.


[ref](https://github.com/openshift/origin/blob/master/tls/descriptions/descriptions.md#how-to-meet-the-requirement)

deads2k:
glad to see an update, but we cannot mutate the secrets of the cluster we're inspecting.

test/extended/operators/certs.go
vrutkovs: Etcd-operator now keeps previous certkeys in preparation to signer/ca/key rotation, so this code will remove bootstrap certkey. It has a "garbage collector", but there is no way to trigger it immediately.

Perhaps etcd-operator should have a way to trigger it immediately - and eventually do this automatically after bootstrap?

vrutkovs: TODO: RewriteNodeIPs should include bootstrap IP (fetch it from secrets in openshift-etcd namespace)

tls/descriptions/descriptions.md
hasbro17: Noted. I think we overlooked that the descriptions had these requirements when we recently updated our cert controller.
Created a TODO for us to bring the descriptions up to spec.
https://issues.redhat.com/browse/ETCD-574

test/extended/operators/certs.go
```diff
@@ -79,6 +78,7 @@ func gatherCertsFromPlatformNamespaces(ctx context.Context, kubeClient kubernete
 	return certgraphanalysis.GatherCertsFromPlatformNamespaces(ctx, kubeClient,
 		certgraphanalysis.SkipRevisioned,
 		certgraphanalysis.SkipHashed,
+		certgraphanalysis.SkipBootstrapCerts(bootstrapIP),
```
deads2k, line 81: at some point we'll need to unskip these.

deads2k:
/lgtm
/approve
/hold

holding for acknowledgement of https://github.com/openshift/origin/pull/28657#discussion_r1576866645

test/extended/operators/certs.go
vrutkovs: Created https://issues.redhat.com/browse/API-1801 to follow up

vrutkovs:
Needs to pull in k8s 1.30 rebase to make it buildable

soltysh (approved):
/lgtm

soltysh (approved):
/lgtm

PR #28662, opened 2024-03-20
OCPBUGS-31074: fix: add check for static container

eggfoobar:
added a check for static containers during the cpu affinity comparison between the host and expected value, this is done to account for the fact that static pods do not get altered by the workload webhook and on realtime deployments we might get CPU affinity changes for static pods that do not go through the workload partitioning feature

### DIFF
--- a/test/extended/cpu_partitioning/crio.go
+++ b/test/extended/cpu_partitioning/crio.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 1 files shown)


eggfoobar:
/hold

Waiting for successful rt upgrade from clusterbot

eggfoobar:
/payload-job  periodic-ci-openshift-release-master-ci-4.16-upgrade-from-stable-4.15-e2e-gcp-ovn-rt-upgrade

eggfoobar:
/unhold

The payload test passed, the static pod `kube-rbac-proxy-crio` was present with `"target.workload.openshift.io/management": "{\"effect\": \"PreferredDuringScheduling\"}"` being present as well. This fix should be good

I updated the comment to reflect what's going on

PR #28667, opened 2024-03-26
OCPBUGS-30267: Clarify a misleading message in patho event failures

dgoodwin:
We were outputting From: To: timestamps, which was always 1s implying
the event happened 20+ times in a second, which was entirely untrue. We
don't really know the start time for the event, and even the observed
time is a little sketchy. Switch to just showing the time we received
the event with the count being displayed for now.


neisw:
/lgtm

At some point we may want to at least preserve / record the original From time on the incoming kube event.  But doesn't have to be now.

dgoodwin:
Yes that would be nice, it's going to get a little tricky I think, it could have occurred before we were even watching, but we could modify this whole stack to track each unique message, each time it came in, the count number, and then subtract the largest count from the initial count we saw and do something with that perhaps.

PR #28670, opened 2024-03-27
NO-JIRA: add conditional execution of featuregated tests

deads2k:
By adding `[OCPFeatureGate:MyFeatureGate]` to a test description, openshift-tests will filter the list using the list contained in FeatureGate.  This allows writing the test once and it will run when enabled using any FeatureSet or ClusterProfile.

As a side-effect, it now becomes possible to count how many tests we have for a give feature.

/assign @dgoodwin


deads2k:
looks to work correctly.  Devan was ok in concept.

deads2k:
z-networking][invariant] alert/OVNKubernetesResourceRetryFailure should not be at or above info again

deads2k:

failed on unrelated

PR #28673, opened 2024-03-27
NO-JIRA: Demonstrate trampoline pod write fix

deads2k:
By using the serviceaccount node claim and validatingadmissionpolicy it is possible to restrict the ability of a serviceaccount to write particular resources to only those instances of resource/foo that  have `name == node-name` or `.spec.nodeName == node-name` while allowing impeded access for other users.  If it is using a serviceaccount token (most do), this requires no modification to the workload being restricted.

I can write up a more detailed enhancement/blog post if desired, but this demonstrates how it can be done with today's TechPreview technology (may require https://github.com/openshift/api/pull/1831).

Once https://github.com/openshift/origin/pull/28670 merges, this will automatically only run on TechPreview and be skipped on Default installations.

per request
/cc @derekwaynecarr 

likely interest
/cc @mrunalp @knobunc

### DIFF
--- a/test/extended/authorization/per_node_update.go
+++ b/test/extended/authorization/per_node_update.go
@@ -1,0 +1,141 @@
+package authorization
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+	o "github.com/onsi/gomega"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	"k8s.io/apiserver/pkg/authentication/serviceaccount"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubernetes/test/e2e/framework/pod"
+	imageutils "k8s.io/kubernetes/test/utils/image"
+	"strings"
+
+	g "github.com/onsi/ginkgo/v2"
+	exutil "github.com/openshift/origin/test/extended/util"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	//go:embed per_node_pod.yaml
+	perNodeCheckPod string
+	//go:embed per_node_rolebinding.yaml
+	perNodeRoleBinding string
+	//go:embed per_node_validatingadmissionpolicy.yaml
+	perNodeCheckValidatingAdmissionPolicy string
+	//go:embed per_node_validatingadmissionpolicybinding.yaml
+	perNodeCheckValidatingAdmissionPolicyBinding string
+)
+
+var _ = g.Describe("[sig-auth][Feature:ServiceAccountTokenNodeBinding][OCPFeatureGate:ValidatingAdmissionPolicy] per-node SA tokens", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("by-node-access")
+
+	g.It(fmt.Sprintf("can restrict access by-node"), func() {
+		ctx := context.Background()
+		podYaml := strings.ReplaceAll(perNodeCheckPod, "e2e-ns", oc.Namespace())
+		podYaml = strings.ReplaceAll(podYaml, "registry.build03.ci.openshift.org/ci-ln-i4f498b/stable@sha256:9f9772bb3afa8877a2d58de06d35e75e213bc75df692ee2223746ff3cdf9ced1", imageutils.GetE2EImage(imageutils.Agnhost))
+		podToCreate := resourceread.ReadPodV1OrDie([]byte(podYaml))
+		rolebindingYaml := strings.ReplaceAll(perNodeRoleBinding, "e2e-ns", oc.Namespace())
+		roleBindingToCreate := resourceread.ReadRoleBindingV1OrDie([]byte(rolebindingYaml))
+		admission := strings.ReplaceAll(perNodeCheckValidatingAdmissionPolicy, "e2e-ns", oc.Namespace())
+		admissionToCreate := resourceread.ReadValidatingAdmissionPolicyV1beta1OrDie([]byte(admission))
+		admissionBinding := strings.ReplaceAll(perNodeCheckValidatingAdmissionPolicyBinding, "e2e-ns", oc.Namespace())
+		admissionBindingToCreate := resourceread.ReadValidatingAdmissionPolicyBindingV1beta1OrDie([]byte(admissionBinding))
+
+		defer func() {
+			oc.AdminKubeClient().AdmissionregistrationV1beta1().ValidatingAdmissionPolicies().Delete(context.Background(), admissionToCreate.Name, metav1.DeleteOptions{})
+			oc.AdminKubeClient().AdmissionregistrationV1beta1().ValidatingAdmissionPolicyBindings().Delete(context.Background(), admissionBindingToCreate.Name, metav1.DeleteOptions{})
+		}()
+
+		var err error
+		_, err = oc.AdminKubeClient().AdmissionregistrationV1beta1().ValidatingAdmissionPolicies().Create(context.Background(), admissionToCreate, metav1.CreateOptions{})
+		o.Expect(err).ToNot(o.HaveOccurred())
+		_, err = oc.AdminKubeClient().AdmissionregistrationV1beta1().ValidatingAdmissionPolicyBindings().Create(context.Background(), admissionBindingToCreate, metav1.CreateOptions{})
+		o.Expect(err).ToNot(o.HaveOccurred())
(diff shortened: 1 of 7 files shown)


test/extended/authorization/per_node_update.go
```diff
@@ -0,0 +1,119 @@
...
+
+		// now create so we can see the update cases
+		actualDisallowedConfigMap, err := oc.AdminKubeClient().CoreV1().ConfigMaps(oc.Namespace()).Create(ctx, disallowedConfigMap, metav1.CreateOptions{})
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		actualAllowedConfigMap, err = nodeScopedClient.CoreV1().ConfigMaps(oc.Namespace()).Update(ctx, actualAllowedConfigMap, metav1.UpdateOptions{})
+		o.Expect(err).ToNot(o.HaveOccurred())
+		_, err = nodeScopedClient.CoreV1().ConfigMaps(oc.Namespace()).Update(ctx, actualDisallowedConfigMap, metav1.UpdateOptions{})
+		o.Expect(err).To(o.HaveOccurred())
+		o.Expect(err.Error()).To(o.ContainSubstring("this user may only modify configmaps that belong to the node the pod is running on"))
```
deads2k, line 117: This is the meat.  The restricted serviceaccount token can only modify its own node's configmap, but other users are still unconstrained.

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+        # it probably shouldn't have create powers either, but we can control that
+        operations:  ["CREATE", "UPDATE"]
+        resources:   ["configmaps"]
+  variables:
+    - name: hasNodeName
+      expression: ('authentication.kubernetes.io/node-name' in request.userInfo.extra)
+    - name: isPartitionedServiceAccount
+      expression: request.userInfo.username == "system:serviceaccount:e2e-ns:default"
+  validations:
+    - expression: "!variables.isPartitionedServiceAccount || (!variables.hasNodeName) || object.metadata.name == request.userInfo.extra[\"authentication.kubernetes.io/node-name\"][0]"
```
deads2k, line 21: this is the expression that prevents access.

test/extended/authorization/per_node_validatingadmissionpolicybinding.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: "only-allow-name-matching-node-configmaps"
+spec:
+  policyName: "only-allow-name-matching-node-configmaps"
+  validationActions: [Deny]
+  matchResources:
+    namespaceSelector:
+      matchLabels:
+        kubernetes.io/metadata.name: "e2e-ns"
```
deads2k, line 11: this is currently bound to a single namespace, but you could bind to anything or all.  Just kept it small for the test.

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+        apiVersions: ["v1"]
+        # notice you cannot restrict DELETEs.  Your per-node actor should not have deletion powers.
+        # it probably shouldn't have create powers either, but we can control that
+        operations:  ["CREATE", "UPDATE"]
+        resources:   ["configmaps"]
+  variables:
+    - name: hasNodeName
+      expression: ('authentication.kubernetes.io/node-name' in request.userInfo.extra)
+    - name: isPartitionedServiceAccount
+      expression: request.userInfo.username == "system:serviceaccount:e2e-ns:default"
```
deads2k, line 19: This is what restricts the permissions check to only the subdivided identity (similar to the "are you a node check" in the node authorizer)

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
deads2k, line 21: object.metadata.name could be any field you want and both the oldObject and the newObject.

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "only-allow-name-matching-node-configmaps"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   [""]
+        apiVersions: ["v1"]
+        # notice you cannot restrict DELETEs.  Your per-node actor should not have deletion powers.
```
deads2k, line 11: this detail on delete is worth being sure is clear.

deads2k:
Checking that skip now

/retest

test/extended/authorization/per_node_update.go
```diff
@@ -0,0 +1,119 @@
...
+	perNodeCheckPod string
+	//go:embed per_node_rolebinding.yaml
+	perNodeRoleBinding string
+	//go:embed per_node_validatingadmissionpolicy.yaml
+	perNodeCheckValidatingAdmissionPolicy string
+	//go:embed per_node_validatingadmissionpolicybinding.yaml
+	perNodeCheckValidatingAdmissionPolicyBinding string
+)
+
+var _ = g.Describe("[sig-auth][Feature:ServiceAccountTokenNodeBinding][OCPFeatureGate:ValidatingAdmissionPolicy] per-node SA tokens", func() {
```
deads2k, line 34: @everettraven example of how to write a test in OCP that relies on ValidatingAdmissionPolicy

mrunalp:
@cdoern @yuqi-zhang (for some of the items we need to address)

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+        # it probably shouldn't have create powers either, but we can control that
+        operations:  ["CREATE", "UPDATE"]
+        resources:   ["configmaps"]
```
mrunalp, line 14: Can any CRD be specified here?

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
mrunalp: Shouldn't the last check be  `!=` ?

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+  variables:
+    - name: hasNodeName
```
enj, line 16: I would expect such a policy to fail closed and consider a missing node name to be invalid.

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
enj: This is because we don't have the body of the object right?

Wouldn't it be safer to include logic to fail all delete attempts?

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
deads2k: > Can any CRD be specified here?

Yes

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
deads2k: > This is because we don't have the body of the object right?

Yes



> Wouldn't it be safer to include logic to fail all delete attempts?

I think it would be safer to not grant deletion powers in RBAC.

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
deads2k: > I would expect such a policy to fail closed and consider a missing node name to be invalid.

Regular users and system:masters will not have node names, but should be allowed to update if they have the RBAC permissions.

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
deads2k: > Shouldn't the last check be `!=` ?

When the expression results in true, the request is allowed.  If the metadata.name == the node name of the allowed pod SA, then we want to allow the request.  If it doesn't match, then fail.

I think both sides of this are tested in the e2e test itself.

xpivarc:
@deads2k any plans to make the extra info available on 4.17?

deads2k:
> @deads2k any plans to make the extra info available on 4.17?

I think that promotion will happen by default.

test/extended/authorization/per_node_validatingadmissionpolicy.yaml
deads2k: > … and it is really need to do this in an admission policy.

It's the only spot where we have access to the content of the change.

deads2k:
moved upstream to https://github.com/kubernetes/kubernetes/pull/124711

PR #28675, opened 2024-03-28
OCPBUGS-31484: Use centos7 tag instead of latest for cmd images tests

ardaguclu:
Images' `latest` tag and auto built mechanisms have been permanently disabled in `quay.io/centos7` repository. 
This PR modifies these images to use `centos7` tag instead of latest. So that cmd/images.sh tests will continue passing. 

But there is no indication whether, at some point in the future, these images or this tag will be removed also or not. I try to remove unit tests entirely to overcome this issue https://github.com/openshift/oc/pull/1717 but maybe for e2e tests, we can give one more try to continue with `centos7` tag, until it is removed.

### DIFF
--- a/examples/image-streams/image-streams-centos7.json
+++ b/examples/image-streams/image-streams-centos7.json
@@ -208,7 +208,7 @@
             },
             "from": {
               "kind": "DockerImage",
-              "name": "quay.io/centos7/httpd-24-centos7:latest"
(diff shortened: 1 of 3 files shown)


soltysh (approved):
/lgtm
/approve

ardaguclu:
@bparees could you please take a look at this?. Thanks.

sanchezl:
This should be back-ported.

PR #28680, opened 2024-04-02
OCPBUGS-31030, OCPBUGS-31032: Reduce deploymentconfig test 3GBi memory size to reasonable

ardaguclu:
`tools:latest` image takes 922792256 bytes (i.e. ~1GBi). So that there is no reason to allocate memory size to 3Gi, besides this test fails on the cluster having limited resources. This PR reduces the allocated memory size to 1.4Gi. This value has been chosen to have a safety gap in the cases where tools image's size is increased. 

Moreover, there are numerous deploymentconfig deprecated warnings stemming from deploymentconfig status check function that is executed per 200ms, this PR increases period to 500ms in order to reduce the count of warning messages.


ardaguclu:
@ingvagabund does this change make sense?. It seems to be safe to me but I'd like to ensure that not causing any issues.

ingvagabund:
Worth trying how this goes.
/approve
/lgtm

PR #28682, opened 2024-04-02
WIP: add monitor test for kubelet selinux labels

rphillips:


pkg/monitortests/node/kubeletselinuxlabels/util.go
```diff
@@ -0,0 +1,74 @@
...
+const (
+	busyScript = `
+#!/bin/bash
+
+set -eou pipefail
+
+SUB='kubelet_t'
+
+while true; do
+	kubelet_label=$(ps -Z -C kubelet)
```
kannon92, line 24: This doesn't work. :(

The shells that we were using don't have -C kubelet. Its why I went with the ugly version of awk.

PR #28683, opened 2024-04-03
CFE-853: e2e test case for DNSNameResolver and EgressFirewall integration

bharath-b-rh:
e2e test case covering integration of CoreDNS and EgressFirewall as proposed in the [enhancement](https://github.com/openshift/enhancements/pull/1335)

### DIFF
--- a/test/extended/networking/egress_firewall.go
+++ b/test/extended/networking/egress_firewall.go
@@ -22,9 +22,11 @@ import (
 const (
 	egressFWTestPod      = "egressfirewall"
 	egressFWE2E          = "egress-firewall-e2e"
+	wcEgressFWE2E        = "wildcard-egress-firewall-e2e"
 	noEgressFWE2E        = "no-egress-firewall-e2e"
 	egressFWTestImage    = "registry.k8s.io/e2e-test-images/agnhost:2.47"
 	oVNKManifest         = "ovnk-egressfirewall-test.yaml"
+	oVNKWCManifest       = "ovnk-egressfirewall-wildcard-test.yaml"
 	openShiftSDNManifest = "sdn-egressnetworkpolicy-test.yaml"
 )
 
@@ -37,7 +39,7 @@ var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
 	InOVNKubernetesContext(
 		func() {
 			g.It("should ensure egressfirewall is created", func() {
-				doEgressFwTest(egFwf, egFwoc, oVNKManifest, true)
+				doEgressFwTest(egFwf, egFwoc, oVNKManifest, true, false)
 			})
 		},
 	)
@@ -45,10 +47,11 @@ var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
 	InOpenShiftSDNContext(
 		func() {
 			g.It("should ensure egressnetworkpolicy is created [apigroup:network.openshift.io]", func() {
-				doEgressFwTest(egFwf, egFwoc, openShiftSDNManifest, false)
+				doEgressFwTest(egFwf, egFwoc, openShiftSDNManifest, false, false)
 			})
 		},
 	)
+
 	noegFwoc := exutil.NewCLIWithPodSecurityLevel(noEgressFWE2E, admissionapi.LevelBaseline)
 	noegFwf := noegFwoc.KubeFramework()
 	g.It("egressFirewall should have no impact outside its namespace", func() {
@@ -81,7 +84,25 @@ var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
 	})
 })
 
-func doEgressFwTest(f *e2e.Framework, oc *exutil.CLI, manifest string, nodeSelectorSupport bool) error {
+var _ = g.Describe("[sig-network][OCPFeatureGate:DNSNameResolver][Feature:EgressFirewall]", func() {
+	// When OVNKubernetes subnet and coredns-ocp-dnsnameresolver plugins are enabled.
+	// coredns-ocp-dnsnameresolver plugin is a TechPreview feature.
+	// TODO:
+	// - Merge this section with main section when feature is GA.
+	// - Merge oVNKManifest & oVNKWCManifest contents.
+	// - Update doEgressFwTest and sendEgressFwTraffic functions.
+	wcEgFwOc := exutil.NewCLIWithPodSecurityLevel(wcEgressFWE2E, admissionapi.LevelPrivileged)
+	wcEgFwF := wcEgFwOc.KubeFramework()
+	InOVNKubernetesContext(
+		func() {
+			g.It("should ensure egressfirewall with wildcard dns rules is created", func() {
+				doEgressFwTest(wcEgFwF, wcEgFwOc, oVNKWCManifest, true, true)
+			})
+		},
+	)
+})
+
+func doEgressFwTest(f *e2e.Framework, oc *exutil.CLI, manifest string, nodeSelectorSupport, checkWildcard bool) error {
 	g.By("creating test pod")
 	o.Expect(createTestEgressFw(f, egressFWTestPod)).To(o.Succeed())
 
@@ -98,14 +119,14 @@ func doEgressFwTest(f *e2e.Framework, oc *exutil.CLI, manifest string, nodeSelec
 	err := oc.AsAdmin().Run("create").Args("-f", egFwYaml).Execute()
 	o.Expect(err).NotTo(o.HaveOccurred(), "created egress-firewall object")
 
-	o.Expect(sendEgressFwTraffic(f, oc, egressFWTestPod, nodeSelectorSupport)).To(o.Succeed())
+	o.Expect(sendEgressFwTraffic(f, oc, egressFWTestPod, nodeSelectorSupport, checkWildcard)).To(o.Succeed())
 
 	g.By("deleting test pod")
 	deleteTestEgressFw(f)
 	return err
 }
 
-func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelectorSupport bool) error {
+func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelectorSupport, checkWildcard bool) error {
 	infra, err := oc.AdminConfigClient().ConfigV1().Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred(), "failed to get cluster-wide infrastructure")
 
@@ -128,10 +149,22 @@ func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelec
 	_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://docs.openshift.com").Output()
 	expectNoError(err)
 
-	// Test curl to www.google.com:80 should fail
-	// because we don't have allow dns rule for www.google.com:80
+	if checkWildcard {
+		// Test curl to `www.google.com` and `translate.google.com` should pass
+		// because we have allow dns rule for `*.google.com`.
+		g.By("sending traffic to `www.google.com` that matches allow dns rule for `*.google.com`")
+		_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://www.google.com").Output()
+		expectNoError(err)
+
+		g.By("sending traffic to `translate.google.com` that matches allow dns rule for `*.google.com`")
+		_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://translate.google.com").Output()
+		expectNoError(err)
+	}
+
+	// Test curl to www.redhat.com should fail
+	// because we don't have allow dns rule for www.redhat.com
 	g.By("sending traffic that does not match allow dns rule")
-	_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "http://www.google.com:80").Output()
+	_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "http://www.redhat.com").Output()
 	expectError(err)
 
 	if nodeSelectorSupport {
--- a/test/extended/testdata/egress-firewall/ovnk-egressfirewall-wildcard-test.yaml
+++ b/test/extended/testdata/egress-firewall/ovnk-egressfirewall-wildcard-test.yaml
@@ -1,0 +1,23 @@
+apiVersion: k8s.ovn.org/v1
+kind: EgressFirewall
+metadata:
+  name: default
+spec:
+  egress:
+  - type: Allow
+    to:
+      dnsName: docs.openshift.com
+  - type: Allow
+    to:
+      dnsName: "*.google.com"
+  - type: Allow
+    to:
+      cidrSelector: 8.8.8.8/32
+  - type: Allow
+    to:
+      nodeSelector:
+        matchLabels:
+          node-role.kubernetes.io/control-plane: ''
+  - type: Deny
+    to:
+      cidrSelector: 0.0.0.0/0
--- a/test/extended/util/annotate/generated/zz_generated.annotations.go
+++ b/test/extended/util/annotate/generated/zz_generated.annotations.go
@@ -1437,6 +1437,8 @@ var Annotations = map[string]string{
 
 	"[sig-network][Feature:vlan] should create pingable pods with vlan interface on an in-container master [apigroup:k8s.cni.cncf.io]": " [Suite:openshift/conformance/parallel]",
 
+	"[sig-network][OCPFeatureGate:DNSNameResolver][Feature:EgressFirewall] when using openshift ovn-kubernetes should ensure egressfirewall with wildcard dns rules is created": " [Suite:openshift/conformance/parallel]",
+
 	"[sig-network][OCPFeatureGate:NetworkDiagnosticsConfig][Serial] Should be enabled by default": " [Suite:openshift/conformance/serial]",
 
 	"[sig-network][OCPFeatureGate:NetworkDiagnosticsConfig][Serial] Should function without any target pods": " [Suite:openshift/conformance/serial]",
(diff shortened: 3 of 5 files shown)


test/extended/testdata/egress-firewall/ovnk-egressfirewall-wildcard-test.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: default
+spec:
+  egress:
+  - type: Allow
+    to:
+      dnsName: docs.openshift.com
+  - type: Allow
+    to:
+      dnsName: "*.ci.openshift.org"
```
arkadeepsen, line 12:
```suggestion
  - type: Allow
    to:
      dnsName: "*.google.com"
```
We can utilize the exiting curl to `www.google.com` for the wildcard DNS rule.

test/extended/networking/egress_firewall.go
```diff
@@ -128,6 +192,17 @@ func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelec
...
+	// - Check for coredns-ocp-dnsnameresolver plugin.
+	// - Remove TecHPreview check when feature is GA.
+	if networkPluginName() == OVNKubernetesPluginName && exutil.IsTechPreviewNoUpgrade(oc) {
+		// Test curl to registry.build05.ci.openshift.org should pass
+		// because we have allow dns rule for *.ci.openshift.org
+		g.By("sending traffic that matches allow dns rule")
+		_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", " https://registry.build05.ci.openshift.org").Output()
+		expectNoError(err)
+	}
+
```
arkadeepsen, line 205: We can utilize the existing curl to `www.google.com` rather than adding a new curl here. Also, the current support added for wildcard DNS name is to check for only one level of subdomain. Thus, `*.ci.openshift.org` will not match `registry.build05.ci.openshift.org` as it has 2 levels of subdomain after `ci.openshift.org`.

test/extended/networking/egress_firewall.go
```diff
@@ -49,6 +56,25 @@ var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
 			})
 		},
 	)
+
+	// When OVNKubernetes subnet and coredns-ocp-dnsnameresolver plugins are enabled.
+	// coredns-ocp-dnsnameresolver plugin is a TechPreview in 4.16.
```
arkadeepsen, line 61: Better not to mention the OCP release.
```suggestion
	// coredns-ocp-dnsnameresolver plugin is a TechPreview feature.
```

test/extended/networking/egress_firewall.go
```diff
@@ -100,6 +126,44 @@ func doEgressFwTest(f *e2e.Framework, oc *exutil.CLI, manifest string, nodeSelec
...
+		o.Expect(len(items)).ShouldNot(o.Equal(0))
+
+		for _, item := range items {
+			dnsName := item.Get("spec.name").String()
+			resolvedNames := item.Get("status.resolvedNames").InterSlice()
+			o.Expect(len(resolvedNames)).Should(o.Equal(1))
+			name, ok := resolvedNames[0].(nwv1alpha1.DNSNameResolverResolvedName)
+			o.Expect(ok).Should(o.Equal(true))
+			switch dnsName {
+			case "docs.openshift.com":
```
arkadeepsen, line 154:
```suggestion
			case "docs.openshift.com.":
```
DNSNameResolver objects stores the DNS names with a trailing period.

test/extended/networking/egress_firewall.go
```diff
@@ -100,6 +126,44 @@ func doEgressFwTest(f *e2e.Framework, oc *exutil.CLI, manifest string, nodeSelec
...
+				o.Expect(name.DNSName).Should(o.Equal("docs.openshift.com"))
```
arkadeepsen, line 155:
```suggestion
				o.Expect(name.DNSName).Should(o.Equal("docs.openshift.com."))
```
DNSNameResolver objects stores the DNS names with a trailing period.

test/extended/networking/egress_firewall.go
```diff
@@ -100,6 +126,44 @@ func doEgressFwTest(f *e2e.Framework, oc *exutil.CLI, manifest string, nodeSelec
...
+				o.Expect(name.ResolutionFailures).Should(o.Equal(0))
+			case "*.ci.openshift.org":
+				o.Expect(name.DNSName).Should(o.Equal("registry.build05.ci.openshift.org"))
+				o.Expect(name.ResolutionFailures).Should(o.Equal(0))
```
arkadeepsen, line 159:
```suggestion
			case "*.google.com.":
				o.Expect(name.DNSName).Should(o.Equal("www.google.com."))
				o.Expect(name.ResolutionFailures).Should(o.Equal(0))
```

test/extended/networking/egress_firewall.go
bharath-b-rh: I have missed it. I had understood wildcards could be used only at the beginning and there was no restriction on the depth.

test/extended/testdata/egress-firewall/ovnk-egressfirewall-wildcard-test.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - type: Deny
+    to:
+      dnsName: www.google.com
+    ports:
+    - port: 80
+      protocol: TCP
```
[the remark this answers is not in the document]
bharath-b-rh: Deny is a legacy feature correct, don't we need to test along the new feature?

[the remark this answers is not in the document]
arkadeepsen: Both the features won't be active at the same time. If you enable the feature through TechPreview, then only the new feature will be active and it won't support Deny rules.

test/extended/networking/egress_firewall.go
```diff
@@ -100,6 +125,44 @@ func doEgressFwTest(f *e2e.Framework, oc *exutil.CLI, manifest string, nodeSelec
...
+			case "docs.openshift.com.":
+				o.Expect(name.DNSName).Should(o.Equal("docs.openshift.com."))
+				o.Expect(name.ResolutionFailures).Should(o.Equal(0))
+			case "*.google.com.":
+				o.Expect(name.DNSName).Should(o.Equal("www.google.com."))
+				o.Expect(name.ResolutionFailures).Should(o.Equal(0))
+			default:
+				e2e.Logf("DNSNameResolver not expected to be created for DNSName %s", dnsName)
+				o.Expect(dnsName).NotTo(o.HaveOccurred())
+			}
```
npinaeva, line 162: do we even need this part? considering it's a part of egress firewall test, we only care about connectivity allowed/denied, I would say we should have a separate test suite for dnsNameResolver, where we would check if the status is correct (potentially later)

test/extended/networking/egress_firewall.go
```diff
@@ -128,6 +191,17 @@ func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelec
...
 	expectNoError(err)
 
+	// TODO:
+	// - Check for coredns-ocp-dnsnameresolver plugin.
+	// - Remove TecHPreview check when feature is GA.
+	if networkPluginName() == OVNKubernetesPluginName && exutil.IsTechPreviewNoUpgrade(oc) {
+		// Test curl to registry.build05.ci.openshift.org should pass
+		// because we have allow dns rule for *.ci.openshift.org
+		g.By("sending traffic that matches allow dns rule")
+		_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://www.google.com").Output()
```
npinaeva, line 201: can we add one more check with an extra dns label, like `translate.google.com`?

test/extended/networking/egress_firewall.go
```diff
@@ -128,6 +191,17 @@ func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelec
 	_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://docs.openshift.com").Output()
 	expectNoError(err)
 
+	// TODO:
+	// - Check for coredns-ocp-dnsnameresolver plugin.
+	// - Remove TecHPreview check when feature is GA.
+	if networkPluginName() == OVNKubernetesPluginName && exutil.IsTechPreviewNoUpgrade(oc) {
+		// Test curl to registry.build05.ci.openshift.org should pass
+		// because we have allow dns rule for *.ci.openshift.org
```
npinaeva, line 199: that comment doesn't look right

npinaeva (commented):
some nits

test/extended/networking/egress_firewall.go
bharath-b-rh: Since it was a new feature added, thought to verify it as well. Should I remove it or keep it so that when this block code is moved to dedicated test suite, this would be moved as well?

test/extended/networking/egress_firewall.go
bharath-b-rh: suggestion incorprated.

test/extended/networking/egress_firewall.go
bharath-b-rh: Missed correcting comment when addressing a review comment.
Code comment updated.

test/extended/networking/egress_firewall.go
npinaeva: I would say it makes sense to remove it for now until we have a separate test set for dnsNameResolver
@arkadeepsen do you agree?

arkadeepsen: I agree. We can plan for adding it in a separate test when we plan for GA release of the feature. For now removing it makes sense.

bharath-b-rh: suggestion incorporated.

test/extended/networking/egress_firewall.go
```diff
@@ -235,3 +272,18 @@ func platformSupportICMP(platformType configv1.PlatformType) bool {
...
+		return from.ObjxMapSlice()
+	case from.IsInterSlice():
+		for _, i := range from.InterSlice() {
+			if msi, ok := i.(map[string]interface{}); ok {
+				values = append(values, objx.Map(msi))
+			}
+		}
+	}
+	return values
+}
```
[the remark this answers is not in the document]
bharath-b-rh: Yeah correct, my bad missed to remove it.

test/extended/util/annotate/generated/zz_generated.annotations.go
```diff
@@ -7,98 +7,10 @@ import (
 )
```
bharath-b-rh, line 7: These changes were suggested by verify CI, and I am not sure why.

test/extended/networking/egress_firewall.go
```diff
@@ -128,6 +149,21 @@ func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelec
 	_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://docs.openshift.com").Output()
 	expectNoError(err)
 
+	// TODO:
+	// - Check for coredns-ocp-dnsnameresolver plugin.
+	// - Remove TechPreview check when feature is GA.
+	if networkPluginName() == OVNKubernetesPluginName && exutil.IsTechPreviewNoUpgrade(oc) {
```
npinaeva, line 155: I think this should be just an argument of sendEgressFwTraffic, like `checkWildcard`, and it can be propagated to `doEgressFwTest`. Then we won't need extra plugin + feature gate check

test/extended/networking/egress_firewall.go
```diff
@@ -128,6 +149,21 @@ func sendEgressFwTraffic(f *e2e.Framework, oc *exutil.CLI, pod string, nodeSelec
...
+		// because we have allow dns rule for `*.google.com`.
+		g.By("sending traffic to `www.google.com` that matches allow dns rule for `*.google.com`")
+		_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://www.google.com").Output()
+		expectNoError(err)
+
+		g.By("sending traffic to `translate.google.com` that matches allow dns rule for `*.google.com`")
+		_, err = oc.Run("exec").Args(pod, "--", "curl", "-q", "-s", "-I", "-m3", "https://translate.google.com").Output()
+		expectNoError(err)
+	}
+
```
npinaeva, line 166: I think we need to update endpoint on line 170, as `http://www.google.com:` now matches the wildcard rule, and the test will fail. We should pick another dns name that is not allowed by all dns configs we have, maybe try `redhat.com`

pkg/monitortests/network/legacynetworkmonitortests/networking.go
```diff
@@ -249,7 +249,7 @@ func testPodSandboxCreation(events monitorapi.Intervals, clientConfig *rest.Conf
 	// add our successes
 	ret = append(ret, successes...)
 
-	return append(ret)
+	return ret
```
npinaeva, line 252: is there a reason for this change?

npinaeva:
Also i think you may need to run `make update` since verify job is failing

test/extended/networking/egress_firewall.go
bharath-b-rh: I will make the change to pass it as an arg.
Since this was for TP, I thought not to change the function signature.

pkg/monitortests/network/legacynetworkmonitortests/networking.go
bharath-b-rh: Change was suggested by `make verify`, without fix CI job was failing.

test/extended/networking/egress_firewall.go
bharath-b-rh: suggestion incorporated.

npinaeva: I didn't mean to add a new check, but to change existing one here https://github.com/openshift/origin/blob/master/test/extended/networking/egress_firewall.go#L133-L134, it is now on line 172.

bharath-b-rh: Sorry, misread the comment. Code updated now.

test/extended/util/annotate/generated/zz_generated.annotations.go
```diff
@@ -7,1076 +7,206 @@ import (
...
 
 var Annotations = map[string]string{
-	"[Conformance][sig-api-machinery][Feature:APIServer] local kubeconfig \"lb-ext.kubeconfig\" should be present on all masters and work": " [Suite:openshift/conformance/parallel/minimal]",
```
npinaeva, line 10: this change doesn't look right though, maybe you can try to run it again?

npinaeva: here is the diff I get if I run make update
```
diff --git a/test/extended/util/annotate/generated/zz_generated.annotations.go b/test/extended/util/annotate/generated/zz_generated.annotations.go
index 8222ca4287..4bb6c0490e 100644
--- a/test/extended/util/annotate/generated/zz_generated.annotations.go
+++ b/test/extended/util/annotate/generated/zz_generated.annotations.go
@@ -1333,6 +1333,8 @@ var Annotations = map[string]string{
 
        "[sig-network][Feature:EgressFirewall] when using openshift ovn-kubernetes should ensure egressfirewall is created": " [Suite:openshift/conformance/parallel]",
 
+       "[sig-network][Feature:EgressFirewall] when using openshift ovn-kubernetes should ensure egressfirewall with wildcard dns rules is created": " [Suite:openshift/conformance/parallel]",
+
        "[sig-network][Feature:EgressFirewall] when using openshift-sdn should ensure egressnetworkpolicy is created [apigroup:network.openshift.io]": " [Suite:openshift/conformance/parallel]",
 
        "[sig-network][Feature:EgressIP][apigroup:operator.openshift.io] [external-targets][apigroup:user.openshift.io][apigroup:security.openshift.io] EgressIPs can be assigned automatically [Skipped:Network/OVNKubernetes]": " [Serial] [Suite:openshift/conformance/serial]",

```

test/extended/networking/egress_firewall.go
```diff
@@ -37,18 +39,38 @@ var _ = g.Describe("[sig-network][Feature:EgressFirewall]", func() {
...
 	InOpenShiftSDNContext(
 		func() {
 			g.It("should ensure egressnetworkpolicy is created [apigroup:network.openshift.io]", func() {
-				doEgressFwTest(egFwf, egFwoc, openShiftSDNManifest, false)
+				doEgressFwTest(egFwf, egFwoc, openShiftSDNManifest, false, false)
 			})
 		},
 	)
+
+	// When OVNKubernetes subnet and coredns-ocp-dnsnameresolver plugins are enabled.
```
kyrtapz, line 55: Instead of checking whether the cluster is techpreview you should create featuregate specific tests:
https://github.com/openshift/api?tab=readme-ov-file#defining-featuregate-e2e-tests
Example: https://github.com/openshift/origin/pull/28719

npinaeva: I am thinking we will make a whole new context in the next dev cycle following all instructions ^, but we wanted 1 simple test in the existing context before we even merge tech preview feature. Hopefully that makes sense

npinaeva: not sure if this helps, but some existing tests actually were an example for this one https://github.com/openshift/origin/blob/75f2c029adf1f56413a95bf0118052a00db41c6a/test/extended/builds/volumes.go#L161-L163

kyrtapz: The approach of testing featuregates is fairly new as far as I know. I think it is better than checking for techpreview because it doesn't require any followup work. It should be really easy to test whether it works in the existing `g.It`

bharath-b-rh: suggestion incorporated. New `Describe` block added with `OCPFeatureGate` tag.

npinaeva:
To do the test run here we need to wait for the last approved pr https://github.com/openshift/cluster-network-operator/pull/2131 to be included to the nightly (which may take quite some time, and the FF is tomorrow). So do you think we can merge this test based on my manual test run with the latest CNO changes https://gist.github.com/npinaeva/021cddfeff438dd2506b8728cb93367f? @kyrtapz @trozet

trozet (approved):
/lgtm

PR #28687, opened 2024-04-03
OCPBUGS-20097: Switch to use must-gather image in build multi-stage

soltysh:
During transition period, when we're updating RHEL versions for oc, it's possible that tools and cli images will be on different levels of RHEL. Whereas cli and must-gather will match. To prevent problems in tests use cli & must-gather in multi-stage build tests.

/assign @ardaguclu


soltysh:
This is required to land https://github.com/openshift/oc/pull/1652

ardaguclu:
None of the failures are due to the test this PR updates.
/lgtm
/hold cancel

soltysh:
/hold 
for ci issues

PR #28689, opened 2024-04-04
NO-JIRA: add service accounts to resourcewatcher

deads2k:
so we can debug when particular imagepullsecrets were added.


rphillips:
This will help debugging a problem within a kubelet in regards to pods starting, and not getting pull secrets updated.

/lgtm

PR #28691, opened 2024-04-05
OCPBUGS-31210: add support for multi-arch codeready-builder-for-rhel-9

prb112:
OCPBUGS-31210: add support for multi-arch codeready-builder-for-rhel-9

1. verified that the arches do have supporting codeready-builder's
2. arch reports aarch64, s390x, ppc64le and x86_64 properly
3. arch matches to the proper repo


adambkaplan (commented):
This will address rhel-9, but not rhel-8 or rhel-7.

prb112:
I have another teammate working on this now.

PR #28693, opened 2024-04-05
NO-ISSUE: automated update synthetic test data

DennisPeriquet:
This is https://github.com/openshift/origin/pull/28642 with a change to the unit test to match `azure` instead of `aws` where we can meet the minJobs requirement.


DennisPeriquet:
this PR is about the query_results.json files and a unit test; the unit test passes so the failure I'm overriding contributes nothing to testing.

DennisPeriquet:
the metal job does not contribute to testing of this PR; only the unit test matters.

PR #28699, opened 2024-04-10
WIP: MON-3513: add resource metrics api availability test

slashpai:
openshift/origin e2e tests don't explicitly exercise the metrics.k8s.io API group. We should add tests under the test/extended/prometheus proving that the API works before/after upgrade.

It will help showing off that the prometheus-adpater -> metrics-server migration happens seamlessly when we turn the feature gate on by default.

### DIFF
--- a/test/extended/prometheus/upgrade.go
+++ b/test/extended/prometheus/upgrade.go
@@ -83,6 +83,42 @@ func (t *MetricsAvailableAfterUpgradeTest) Skip(_ upgrades.UpgradeContext) bool 
 	return !isPersistentStorageEnabled(client)
 }
 
+// Check Resource Metrics API is also available after the upgrade.
+type ResourceMetricsAPIAvailableAfterUpgradeTest struct {
+	executionTimestamp time.Time
+}
+
+func (t *ResourceMetricsAPIAvailableAfterUpgradeTest) Name() string {
+	return "resource-metrics-api-available-after-upgrade"
+}
+
+func (t *ResourceMetricsAPIAvailableAfterUpgradeTest) DisplayName() string {
+	return "[sig-instrumentation] Resource Metrics API should be available after an upgrade"
+}
+
+func (t *ResourceMetricsAPIAvailableAfterUpgradeTest) Setup(ctx context.Context, f *e2e.Framework) {
+	oc := exutil.NewCLIWithFramework(f)
+	g.By("getting the nodes.metrics before the upgrade")
+	preUpgradeResponse, err := oc.Run("get").Args("nodes.metrics").Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(preUpgradeResponse).NotTo(o.BeEmpty())
+}
+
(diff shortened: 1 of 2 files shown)


test/extended/prometheus/upgrade.go
```diff
@@ -83,6 +83,42 @@ func (t *MetricsAvailableAfterUpgradeTest) Skip(_ upgrades.UpgradeContext) bool
...
+
+func (t *ResourceMetricsAPIAvailableAfterUpgradeTest) Test(ctx context.Context, f *e2e.Framework, done <-chan struct{}, upgrade upgrades.UpgradeType) {
+	<-done
+
+	oc := exutil.NewCLIWithFramework(f)
+
+	g.By("verifying whether the nodes.metrics command works after the upgrade")
+	postUpgradeResponse, err := oc.Run("get").Args("nodes.metrics").Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(postUpgradeResponse).NotTo(o.BeEmpty())
```
rexagod, line 115: Not sure if there are any cases where non-emptiness could equate to an unexpected output (without any errors), but if so, would it be worth checking the contents of the output as well, just to be safe?

machine424: I think that checking for the correctness should happen on `oc` and/or CMO sides, IIUC the test is only to show that the API continues to work after the p-a->m-server switch.
We can even get rid of the test in future versions.

If not already done, JP could provoke a failure (putting some nonexisting stuff in `Args()`) of the test so we get an idea about the how it'd look like and to test the test :)

I know the release folks are migrating to the new [monitortests](https://github.com/openshift/origin/blob/master/pkg/monitortestframework/types.go#L39) framework for upgrade tests: you can see how we use it here https://github.com/openshift/origin/pull/28361/files. But for a temp test, I'm ok with keeping it like this.

machine424: (If you agree with me that it's a temp test, it'd be great to add that as a comment in the test and to have a ticket that will revert it in the future)

slashpai: The idea to have the test is to make sure resource metrics api is always available. We didn't have test for that but it was getting tested indirect way for HPA. The test will be still useful even full migration (PA->MS) happened as well. But for now ya we needed a test to prove it will always be functional

We do have other tests as well so would that mean we will need to move all tests using new framework?

machine424: If it's not a temp test, I'd go for the monitortests framework (I could help if you have questions.)
But let's see what the others think.
(I'd be interested in a commit that provokes a failure in both cases though)

simonpasquier: Good remark on monitortests, I'm definitely not up to date with the origin testing :sweat_smile: 
It looks like a good option and the statefulsets example is a good starting point from what I can tell.

machine424: As discussed with JP, we'll go with https://github.com/openshift/origin/pull/28737

PR #28703, opened 2024-04-12
NO-JIRA: add to the operator list

deads2k:
Adding this label so that `/manifests` dir is pulled into the ReleaseInfo.  This is necessary to include the number of tests per feature.

/assign @wking 

Trevor: does this have any other side-effects?

### DIFF
--- a/images/tests/Dockerfile.rhel
+++ b/images/tests/Dockerfile.rhel
@@ -20,6 +20,7 @@ RUN PACKAGES="git gzip util-linux" && \
     git config --system user.email test@test.com && \
     chmod g+w /etc/passwd
 LABEL io.k8s.display-name="OpenShift End-to-End Tests" \
+      io.openshift.release.operator=true \
       io.k8s.description="OpenShift is a platform for developing, building, and deploying containerized applications." \
       io.openshift.build.versions="kubernetes-tests=1.29.0" \
       io.openshift.tags="openshift,tests,e2e"


wking (approved):
> does this have any other side-effects?

No, it just transitions you from being an image some already-operator-labeled image is referencing in their `image-references` to one that `oc adm release new ...` slurps in directly, and gets you the manifest-scraping you want to make #28694 useful ([docs][1]).

/lgtm

[1]: https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/operators.md#how-do-i-ensure-the-right-images-get-used-by-my-manifests

lance5890:
> > does this have any other side-effects?
> 
> No, it just transitions you from being an image some already-operator-labeled image is referencing in their `image-references` to one that `oc adm release new ...` slurps in directly, and gets you the manifest-scraping you want to make #28694 useful ([docs](https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/operators.md#how-do-i-ensure-the-right-images-get-used-by-my-manifests)).
> 
> /lgtm

- But I noticed that the `test-reporting.yaml` is added into the release-image file ,such as 
image

- Maybe this is included by `cluster-samples-operator`

https://github.com/openshift/cluster-samples-operator/blob/62b3623c6a2860e7a8cc368dbe76e767508cd261/manifests/08-openshift-imagestreams.yaml#L76-L92

PR #28705, opened 2024-04-12
NO-JIRA: split the per-user summary to include verbs for resources

deads2k:
to ease "what is this operator doing" checks.  Could be uploaded to associate operator and per-resource,per-verb usage.

/assign @neisw 
/cc @yuqi-zhang


neisw:

unrelated install failures

PR #28706, opened 2024-04-15
Networking: add a test to validate Communication Matrix

sabinaaledort:
The commatrix module automatically generates an accurate and up-to-date communication  
flows matrix that can be delivered to customers as part of product documentation for all  
ingress flows of OpenShift (multi-node and single-node deployments).

This test validates that the generated Communication Matrix covers all the ingress traffic open ports on the nodes by reviewing the `matrix-diff-ss` file.

The generated artifacts are:
1. communication-matrix - The generated communication matrix.
2. ss-generated-matrix - The communication matrix that generated by the `ss` command.
3. matrix-diff-ss - Shows the variance between two matrices. Entries present in the communication matrix but absent in the ss matrix are marked with '+', while entries present in the ss matrix but not in the communication matrix are marked with '-'.
4. raw-ss-tcp - The raw `ss` output for TCP.
5. raw-ss-udp - The raw `ss` output for UDP.


test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+const (
+	serviceNodePortMin = 30000
+	serviceNodePortMax = 32767
+)
+
+var _ = g.Describe("[sig-network][Feature:commatrix][Serial]", func() {
+	g.It("should validate the communication matrix ports match the node's listening ports", func() {
+		// TEMP
```
shirmoran, line 38: Should the following line be removed?

aabughosh: she need to rebase and remove the utils.InitializeLog(false) we dont need that anymore

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,185 @@
...
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+)
+
+const (
+	serviceNodePortMin = 30000
+	serviceNodePortMax = 32767
+)
+
+var _ = g.Describe("[sig-network][Feature:commatrix][Serial]", func() {
+	g.It("should validate the communication matrix ports match the node's listening ports", func() {
```
pacevedom, line 37: Is it possible to add `[apigroup:config.openshift.io]` to the name so that it is automatically skipped for MicroShift?

PR #28709, opened 2024-04-15
OCPBUGS-26440: feat: add buffer interval for SNO pathological events

eggfoobar:
during single node deployments knowing the window between graceful shutdown end and readyz start is helpful to determine when connection refused events are expected. This update makes use of a generated buffer interval between the time the graceful shutdown signal is received and some specified buffer of time to account for the API server coming back up.

### DIFF
--- a/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
+++ b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
+	"time"
 
 	v1 "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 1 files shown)


pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -898,14 +899,46 @@ func newSingleNodeConnectionRefusedEventMatcher(finalIntervals monitorapi.Interv
...
+		switch eventInterval.StructuredLocator.Keys[monitorapi.LocatorNamespaceKey] {
+		case ocpAPINamespace, ocpOAuthAPINamespace:
+			identifiedSkipInterval = true
+		case defaultNamespace:
+			identifiedSkipInterval = strings.Contains(eventInterval.StructuredMessage.HumanMessage, "apiserver")
+		}
+
+		return isTargetDownAlert && identifiedSkipInterval
 	})
+	ocpAPISeverTargetDownIntervals = append(ocpAPISeverTargetDownIntervals, bufferInterval...)
```
neisw, line 941: Looks like the order doesn't matter for allowIfWithinIntervals so you are good.  But I did wonder if this could use sorting by when I first saw it.  Not an ask, just observation.

eggfoobar: Good idea, won't hurt and will have a non zero improvement. Since I had to rebase, I went ahead and updated this. Thanks!

PR #28715, opened 2024-04-16
NO-JIRA: Create autodl file for per user resource request audit entries

neisw:
Builds on https://github.com/openshift/origin/pull/28705 to create json output for bq loading

### DIFF
--- a/pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
+++ b/pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
@@ -3,14 +3,19 @@ package auditloganalyzer
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/origin/pkg/dataloader"
+	"github.com/sirupsen/logrus"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 )
+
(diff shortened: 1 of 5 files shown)


deads2k:
code lgtm, but two possible improvements
1. fix the TODO and skip the audit log entries outside the beginning and end (this might matter for the upgrade conformance tests)
2. only upload users in...
    1. openshift-*
    2. kubelets
    3. kcm identity
    4. kas identity
    5. ks identity

pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
```diff
@@ -3,15 +3,20 @@ package auditloganalyzer
...
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 )
 
+var monitoredUsers = []string{"kube-", "openshift-", "system:node"}
```
deads2k, line 18: want to make these all prefixes?  `system:serviceaccount:kube-`, `system:serviceaccount:openshift-`, `system:node`.

pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
```diff
@@ -445,6 +450,42 @@ func URIToParts(uri string) (string, schema.GroupVersionResource, string, string
 	return ns, gvr, name, ""
 }
 
+func isMonitoredUser(user string) bool {
+	for _, partialUser := range monitoredUsers {
+		if strings.Contains(user, partialUser) {
```
deads2k, line 455: Then use `strings.HasPrefix` here.

deads2k:
optional suggestion (could also do a followup)

/lgtm

PR #28717, opened 2024-04-17
STOR-1838: add test for vsphere driver snapshot configuration

RomanBednar:


RomanBednar:
/hold

I think we need to add vsphere techpreview job first: https://github.com/openshift/release/pull/51039

jsafrane:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+				successfulSnapshotsCreated: 3,
+			},
+			{
+				name: "vsan snapshot option",
+				clusterCSIDriverOptions: map[string]int{
+					snapshotOptions["vsan"]["clusterCSIDriver"]: 4,
+				},
+				cloudConfigOptions: map[string]int{
+					snapshotOptions["vsan"]["cloudConf"]: 4,
+				},
```
jsafrane, line 114: Why so complicated?

```
clusterCSIDriverOptions: opv1.VSphereCSIDriverConfigSpec {
    granularMaxSnapshotsPerBlockVolumeInVSAN: ptr.To(uint32(4)),
},

cloudConfigOptions: map[string]string{
    "granular-max-snapshots-per-block-volume-vsan": "4".
}
```

(note change of the field types)

It will be slightly harder to compute the ClusterCSIDriver patch, but you can use `Update()` instead or encode VSphereCSIDriverConfigSpec to json.
On the positive side, it's crystal clear what field is set and what ini file is expected without reading a separate map.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+					o.Expect(err).NotTo(o.HaveOccurred())
+
+					errMsg, err := oc.Run("get").Args("volumesnapshot/test-snapshot-failed", "-o", "jsonpath={.status.error.message}").Output()
+					o.Expect(err).NotTo(o.HaveOccurred())
+
+					e2e.Logf("VolumeSnapshot error message: %s readyToUse %s", errMsg, readyToUse)
+					if !strings.Contains(errMsg, "failed to take snapshot of the volume") && readyToUse != "false" {
+						e2e.Failf("VolumeSnapshot \"test-snapshot-failed\" should have failed and should not be ready to use")
+					}
+				}
```
jsafrane, line 194: This should go to its own function, just like `loadAndCheckCloudConf`.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+				for option, value := range t.clusterCSIDriverOptions {
+					e2e.Logf("updating %s to %d in clustercsidriver", option, value)
+					setClusterCSIDriverSnapshotOptions(oc, option, value)
+				}
+
+				for option, value := range t.cloudConfigOptions {
+					o.Eventually(func() error {
+						return loadAndCheckCloudConf(oc, "Snapshot", option, value)
+					}, time.Minute, time.Second).Should(o.Succeed())
+				}
```
jsafrane, line 154: To check all three ini file keys you `Get` the same object from the API server three times? `loadAndCheckCloudConf` can get a `map[string]string` and check all of them with a single `Get`.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return fmt.Errorf("section %s not found in cloud.conf: %v", sectionName, err)
+	}
+
+	key, err := section.GetKey(keyName)
+	if err != nil {
+		return fmt.Errorf("key %s not found in section %s: %v", keyName, sectionName, err)
+	}
+
+	o.Expect(key.String()).To(o.Equal(fmt.Sprintf("%d", expectedValue)))
```
jsafrane, line 237: You should check that the other keys are *not* set.
I wonder what `section.KeysHash()` returns. Is it the same map as `t.cloudConfigOptions`, if you used `map[string]string` instead of `int` as I suggest? Will `DeepEqual` work then?

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+				}
+
+				for option, value := range t.cloudConfigOptions {
+					o.Eventually(func() error {
+						return loadAndCheckCloudConf(oc, "Snapshot", option, value)
+					}, time.Minute, time.Second).Should(o.Succeed())
+				}
+
+				if t.successfulSnapshotsCreated > 0 {
+					pvc, err := createTestPVC(oc, oc.Namespace(), "test-pvc", "1Gi")
```
jsafrane, line 157: Where is the PVC deleted?

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+					o.Expect(err).NotTo(o.HaveOccurred())
+
+					_, err = createTestPod(oc, pvc.Name, oc.Namespace())
```
jsafrane, line 160: Where is the Pod deleted?

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+							return err
+						}
+						if pvc.Status.Phase != v1.ClaimBound {
+							return fmt.Errorf("PVC not bound")
+						}
+						return nil
+					})
+
+					for i := 0; i < t.successfulSnapshotsCreated; i++ {
+						err := createSnapshot(oc, oc.Namespace(), fmt.Sprintf("test-snapshot-%d", i), "test-pvc")
```
jsafrane, line 176: Where are the snapshots deleted?

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+						o.Expect(err).NotTo(o.HaveOccurred())
+					}
+
+					// Next snapshot creation should be over the set limit and fail.
+					err = createSnapshot(oc, oc.Namespace(), "test-snapshot-failed", "test-pvc")
```
jsafrane, line 181: It's not guaranteed that it's the the last snapshot that is going to fail. The test creates several VolumeSnapshots in a quick succession, the controller + CSI driver may process them in any order. You should wait until the "good snapshots" are ready to use.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+					readyToUse, err := oc.Run("get").Args("volumesnapshot/test-snapshot-failed", "-o", "jsonpath={.status.readyToUse}").Output()
+					o.Expect(err).NotTo(o.HaveOccurred())
+
+					errMsg, err := oc.Run("get").Args("volumesnapshot/test-snapshot-failed", "-o", "jsonpath={.status.error.message}").Output()
+					o.Expect(err).NotTo(o.HaveOccurred())
+
+					e2e.Logf("VolumeSnapshot error message: %s readyToUse %s", errMsg, readyToUse)
+					if !strings.Contains(errMsg, "failed to take snapshot of the volume") && readyToUse != "false" {
+						e2e.Failf("VolumeSnapshot \"test-snapshot-failed\" should have failed and should not be ready to use")
+					}
```
jsafrane, line 193: I miss some loop waiting for the snapshots to get failed.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,326 @@
...
+				}
+			})
+
+		}
+	})
+})
+
+func setClusterCSIDriverSnapshotOptions(oc *exutil.CLI, snapshotOptions string, value int) {
+	patch := []byte(fmt.Sprintf("{\"spec\":{\"driverConfig\":{\"vSphere\":{\"%s\": %d}}}}", snapshotOptions, value))
```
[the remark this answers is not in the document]
RomanBednar: Removed, using Update() instead.

test/extended/storage/driver_configuration.go
RomanBednar: Changed and bumping openshift/api to pull in the new clusterCSIDriver fields.

test/extended/storage/driver_configuration.go
RomanBednar: Moved to separate function.

test/extended/storage/driver_configuration.go
RomanBednar: Yeah, I missed that, changed.

test/extended/storage/driver_configuration.go
RomanBednar: DeepEqual() seems to work well in this case, thanks for the suggestion.

test/extended/storage/driver_configuration.go
RomanBednar: Added removal. The cleanup slows down the test a bit, how bad is it to rely on the project teardown (in oc from util.CLI)?

test/extended/storage/driver_configuration.go
RomanBednar: Same as PVC.

test/extended/storage/driver_configuration.go
RomanBednar: Same as PVC - snapshot removal is taking the longest.

test/extended/storage/driver_configuration.go
RomanBednar: You're right, adding a simple check with o.Eventually() counting ready snapshots.

test/extended/storage/driver_configuration.go
RomanBednar: Snapshots? I have just one for which I validate failure, do we need more?

Also, this could be improved with o.Consistently() to see if the snapshot stays in not ready state with an error for some time. See if the global timeout and polling setting around lines 36-39 is ok for this test.

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,374 @@
...
+	o.Consistently(func() bool {
+		ready, err := isSnapshotReady(oc, failedSnapshotName)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		errMsg, err := oc.Run("get").Args(fmt.Sprintf("volumesnapshot/%s", failedSnapshotName), "-o", "jsonpath={.status.error.message}").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		e2e.Logf("Snapshot %s readyToUse: %t, error message: %s", failedSnapshotName, ready, errMsg)
+
+		return strings.Contains(errMsg, "failed to take snapshot of the volume") && !ready
```
Phaow, line 275:
```suggestion
		return strings.Contains(errMsg, "failed to take snapshot of the volume") && strings.Contains(errMsg, "reaches the configured maximum") && !ready
```
Maybe better to check the failed reason is `reaches the configured maximum`.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,374 @@
...
+	"time"
+)
+
+const (
+	projectName  = "csi-driver-configuration"
+	providerName = "csi.vsphere.vmware.com"
+)
+
+// This is [Serial] because it modifies ClusterCSIDriver.
+var _ = g.Describe("[sig-storage][Feature:VSphereDriverConfiguration][Serial] vSphere CSI Driver Configuration", func() {
```
Phaow, line 30:
```suggestion
var _ = g.Describe("[sig-storage][Feature:VSphereDriverConfiguration][Serial] [apigroup:operator.openshift.io] vSphere CSI Driver Configuration", func() {
```
Better also add `[apigroup:operator.openshift.io]` tag for exclude cases for microshift, exclude it is better than skip.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,374 @@
...
+		return fmt.Errorf("check of %s section in cloud.conf failed, got: %v expected: %v", sectionName, section.KeysHash(), cloudConfigOptions)
+	}
+
+	e2e.Logf("Validation of %s section in cloud.conf succeeded", sectionName)
+
+	return nil
+}
+
+func validateSnapshotCreation(ctx context.Context, oc *exutil.CLI, successfulSnapshotsCreated int) {
+	e2e.Logf("Validating snapshot creation.")
```
Phaow, line 201:
```suggestion
	e2e.Logf("Validating snapshot creation.")
	if successfulSnapshotsCreated == 0 {
	      e2e.Logf("Skip: Validating successfulSnapshotsCreated sets to 0 ")
	      return
	}
```
When `successfulSnapshotsCreated == 0 ` seems we could return directly.

test/extended/storage/driver_configuration.go
RomanBednar: Changed.

test/extended/storage/driver_configuration.go
RomanBednar: Added.

test/extended/storage/driver_configuration.go
RomanBednar: I'm skipping this validation on L144 to make it visible since it's a parameter of the tests table: https://github.com/openshift/origin/pull/28717/files#diff-fcce86821b4d4cf7d50308a15881ca4063a33dd2294a4e975a56b68315e46b22R144

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,399 @@
...
+		}
+
+		e2e.Logf("Restoring original driverConfig of ClusterCSIDriver")
+		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			clusterCSIDriver, err := oc.AdminOperatorClient().OperatorV1().ClusterCSIDrivers().Get(ctx, providerName, metav1.GetOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to get ClusterCSIDriver")
+
+			clusterCSIDriver.Spec.DriverConfig = *originalDriverConfigSpec
+			_, err = oc.AdminOperatorClient().OperatorV1().ClusterCSIDrivers().Update(ctx, clusterCSIDriver, metav1.UpdateOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to update ClusterCSIDriver")
```
jsafrane, line 71: This will mark the whole test as failed on conflict. I think `RetryOnConflict` will do the right thing on simple `return err` here.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,399 @@
...
+
+			return err
+		}); err != nil {
+			e2e.Failf("failed to restore original driverConfig of ClusterCSIDriver: %v", err)
+		}
```
jsafrane, line 76:
```
err := retry.RetryOnConflict( {
... })
o.Expect(err).NotTo(o.HaveOccurred(), "failed to update ClusterCSIDriver")
```

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,399 @@
...
+				},
+				successfulSnapshotsCreated: 0,
+			},
+		}
+
+		for _, t := range tests {
+			t := t
+			g.It(fmt.Sprintf("%s", t.name), func() {
+
+				if t.clusterCSIDriverOptions != nil {
```
jsafrane, line 144: This assumes the ClusterCSIDriver has `Spec.DriverConfig.VSphere == nil`. IMO it would be better to force set `nil` there, overriding any current setting.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,399 @@
...
+				return
+			}
+			snapshotsCreatedChan <- snapshotName
+		}(i)
+	}
+	wg.Wait()
+	close(snapshotsCreatedChan)
+
+	var snapshotsCreated = make([]string, 0, successfulSnapshotsCreated)
+	for snapshotName := range snapshotsCreatedChan {
```
jsafrane, line 247: Isn't the channel closed here ?!

jsafrane: I'm not sure why there is a channel with snapshot names and then it's transformed into array of names. Can't you create array of snapshot names directly?

test/extended/storage/driver_configuration.go
RomanBednar: `t.clusterCSIDriverOptions` is a `VSphereCSIDriverConfigSpec` just for tests to skip setting options if unset/nil (just the first test now). Maybe I misunderstand the suggestion, could you elaborate please?

test/extended/storage/driver_configuration.go
RomanBednar: Indeed, just wanted a quick check with payload to see if concurrent snapshots work fine and didn't get back to it yet. Corrected.

test/extended/storage/driver_configuration.go
RomanBednar: Changed.

test/extended/storage/driver_configuration.go
RomanBednar: I saw it's being used like this here: https://github.com/openshift/origin/blob/98d9c70b162c863f3dd6856ded4d214cf9aca977/test/extended/templates/templateinstance_impersonation.go#L269-L277

And after testing it for a few times it seemed ok, but let's use `return err` to be on the safe side.

RomanBednar:
Tests passed with payload test: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-origin-28717-nightly-4.16-e2e-vsphere-ovn-techpreview-serial/1782353557949779968/artifacts/e2e-vsphere-ovn-techpreview-serial/openshift-e2e-test/build-log.txt

test/extended/storage/driver_configuration.go
Phaow: Got it, thanks for the clarify!

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

test/extended/storage/driver_configuration.go
jsafrane: You don't know what's in the original `ClusterCSIDriver` instance present in the cluster. There could be some snapshot setting that the test does not expect and it may allow more (or less) than expected snapshots.

test/extended/storage/driver_configuration.go
```diff
@@ -0,0 +1,395 @@
...
+		go func(i int) {
+			defer wg.Done()
+			snapshotName := fmt.Sprintf("test-snapshot-%d", i)
+			err := createSnapshot(oc, oc.Namespace(), snapshotName, "test-pvc")
+			if err != nil {
+				e2e.Failf("failed to create snapshot: %v", err)
+				return
+			}
+			snapshotsCreated = append(snapshotsCreated, snapshotName)
+		}(i)
```
jsafrane, line 243: This is racy.

```suggestion
		wg.Add(1)
		snapshotName := fmt.Sprintf("test-snapshot-%d", i)
		snapshotsCreated = append(snapshotsCreated, snapshotName)
		go func(snapshotName string) {
			defer wg.Done()
			err := createSnapshot(oc, oc.Namespace(), snapshotName, "test-pvc")
			if err != nil {
				e2e.Failf("failed to create snapshot: %v", err)
				return
			}
		}(snapshotName)
```

RomanBednar: Applied.

test/extended/storage/driver_configuration.go
RomanBednar: I see now, I'll let the test override the setting then. Also adding `RetryOnConflict`  since the update could fail there too. Thanks!

test/extended/storage/driver_configuration.go
RomanBednar: Actually...it looks cleaner with the skip message, adding it.

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

jsafrane:
/hold
for the test to succeed

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

RomanBednar:
Rebased to include: https://github.com/openshift/origin/pull/28733

jsafrane:
periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial [failed](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-origin-28717-nightly-4.16-e2e-vsphere-ovn-techpreview-serial/1783485632853577728) and the failure looks related:

> event happened 25 times, something is wrong: namespace/openshift-cluster-storage-operator deployment/cluster-storage-operator hmsg/79ca95c7bd - reason/OperatorStatusChanged Status for clusteroperator/storage changed: Progressing changed from False to True ("VSphereCSIDriverOperatorCRProgressing: VMwareVSphereDriverNodeServiceControllerProgressing: Waiting for DaemonSet to deploy node pods") (15:37:07Z) result=reject }


/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

RomanBednar:
Seems like driver operator has issues connecting to vcenter, not sure why yet, but could it be the reason for the events to be repeated so frequently?
```
18h         Warning   vsphere_connection_failed               deployment/vmware-vsphere-csi-driver-operator                  Marking cluster upgrade status unknown because Failed to connect to vSphere: Post "https://vcenter.ibmc.devcluster.openshift.com/sdk": dial tcp: lookup vcenter.ibmc.devcluster.openshift.com on 172.30.0.10:53: read udp 10.130.0.10:4
```

Similar for shared resource operator connecting to api server (`SHARESCSIDriverOperatorCRProgressing` condition):
```
SharedResourcesDriverNodeServiceController reconciliation failed: Get "https://172.30.0.1:443/apis/apps/v1/namespaces/openshift-cluster-csi-drivers/daemonsets/shared-resource-csi-driver-node": dial tcp 172.30.0.1:443: connect: connection refused
```

Tried to run snapshot test isolated again and observed operator status changes - the DaemonSet redeploys pods when cloud conf ConfigMap is changed, but that's expected along with status changing once:
```
onSet to deploy node pods\nVSphereCSIDriverOperatorCRProgressing: VMwareVSphereDriverNodeServiceControllerProgressing: Waiting for DaemonSet to deploy node pods" to "VSphereCSIDriverOperatorCRProgressing: VMwareVSphereDriverNodeServiceControllerProgressing: Waiting for DaemonSet to deploy node pods"
33m         Normal    OperatorStatusChanged                            deployment/cluster-storage-operator                       Status for clusteroperator/storage changed: Progressing changed from False to True ("VSphereCSIDriverOperatorCRProgressing: VMwareVSphereDriverNodeServiceControllerProgressing: Waiting for DaemonSet to act on changes")
33m         Normal    OperatorStatusChanged                            deployment/cluster-storage-operator                       Status for clusteroperator/storage changed: Progressing changed from False to True ("VSphereCSIDriverOperatorCRProgressing: VMwareVSphereDriverNodeServiceControllerProgressing: Waiting for DaemonSet to deploy node pods")
```

I wonder how the `events should not repeat pathologically` test evaluates events -  maybe by adding this test that flips status we've exceeded some threshold?

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

RomanBednar:
cc @dgoodwin @soltysh for approval

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

jsafrane:
https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/230 has merged

/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -445,6 +445,16 @@ func NewUniversalPathologicalEventMatchers(kubeConfig *rest.Config, finalInterva
...
+	registry.AddPathologicalEventMatcherOrDie(&SimplePathologicalEventMatcher{
+		name: "StorageOperatorsFlipsProgressingTooOften",
+		locatorKeyRegexes: map[monitorapi.LocatorKey]*regexp.Regexp{
+			monitorapi.LocatorNamespaceKey: regexp.MustCompile(`^openshift-cluster-storage-operator$`),
+		},
+		messageReasonRegex: regexp.MustCompile(`^OperatorStatusChanged$`),
+		messageHumanRegex:  regexp.MustCompile(`Progressing changed.*`),
+		jira:               "https://issues.redhat.com/browse/OCPBUGS-24061",
+	})
+
```
dobsonj, line 457: Can you drop this commit and run the payload job again now that we have a fix merged?

RomanBednar:
/payload-job periodic-ci-openshift-release-master-nightly-4.16-e2e-vsphere-ovn-techpreview-serial

soltysh (approved):
/approve

PR #28718, opened 2024-04-17
OCPBUGS-32293: Update Jenkins test to use Deployment

apoorvajagtap:
Failing Jenkins Pipeline test blocks multiple PRs from merging. As agreed, skipping these tests temporarily.

### DIFF
--- a/test/extended/builds/pipeline_origin_bld.go
+++ b/test/extended/builds/pipeline_origin_bld.go
@@ -40,7 +40,7 @@ var _ = g.Describe("[sig-builds][Feature:JenkinsRHELImagesOnly][Feature:Jenkins]
 			o.Expect(err).NotTo(o.HaveOccurred())
 
 			g.By("waiting for jenkins deployment")
-			err = exutil.WaitForDeploymentConfig(oc.KubeClient(), oc.AppsClient().AppsV1(), oc.Namespace(), "jenkins", 1, false, oc)
+			err = exutil.WaitForDeploymentReady(oc, "jenkins", oc.Namespace())
 			if err != nil {
 				exutil.DumpApplicationPodLogs("jenkins", oc)
 			}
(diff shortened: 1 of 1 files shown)


test/extended/builds/pipeline_origin_bld.go
```diff
@@ -91,6 +91,8 @@ var _ = g.Describe("[sig-builds][Feature:JenkinsRHELImagesOnly][Feature:Jenkins]
 		g.Describe("jenkins pipeline build config strategy", func() {
 			g.It("using a jenkins instance launched with the ephemeral template [apigroup:build.openshift.io]", func() {
 				defer cleanup()
+
+				g.Skip("the test failure blocks multiple PRs. As this strategy is deprecated, skipping the test to unblock https://issues.redhat.com/browse/OCPBUGS-32293")
```
adambkaplan, line 95:
```suggestion
				g.Skip("OCPBUGS-32293: temporary skip to unblock build and Jenkins suite tests")
```

adambkaplan (approved):
/approve

Minor suggestion on the skip message.

sayan-biswas (approved):

adambkaplan:
:thinking: Now I'm wondering why it took us this long to catch the change...

soltysh (changes requested):
/hold
Where that jenkins-ephemeral is coming from, I checked both https://github.com/openshift/origin/blob/master/examples/jenkins/jenkins-ephemeral-template.json and https://github.com/openshift/origin/blob/master/test/extended/testdata/cmd/test/cmd/testdata/jenkins/jenkins-ephemeral-template.json both of these are creating DCs not deployment? 

Alternatively, if you're changing to a deployment (if the above what I said is wrong), you'll need to update logs command in line 65, it currently requests `logs -f dc/jenkins` which if we're working with a deployment should be `logs -f deploy/jenkins`

soltysh:
Found the culprit in https://github.com/openshift/cluster-samples-operator/pull/540, so yes, aside from that wait, you need to update that run as I mentioned above. 

Also, given we have those two extra places which don't seem to be in-sync with what we have in samples, I'd suggest to sync them as well. To make them all consistent.

soltysh:
/approve cancel
Please ping me directly on slack once all of the above is fixed, I don't want to have a half baked solution.

soltysh:
Also `DumpApplicationPodLogs` works for DCs only, so that needs fixing as well.

soltysh:
Reverting changes to cluster-samples-operator in https://github.com/openshift/cluster-samples-operator/pull/542

adambkaplan:
/close

Replacing with https://github.com/openshift/origin/pull/28732

PR #28719, opened 2024-04-17
SDN-4783: Add e2e tests for NetworkDiagnosticsConfig featuregate

kyrtapz:
This is blocked by https://github.com/openshift/cluster-network-operator/pull/2339


test/extended/networking/network_diagnostics.go
```diff
@@ -0,0 +1,248 @@
...
+
+	g.AfterEach(func(ctx context.Context) {
+		// Reset network diagnostics config
+		netConfigApply := applyconfigv1.Network(clusterConfig).WithSpec(
+			applyconfigv1.NetworkSpec().WithNetworkDiagnostics(nil),
+		)
+		_, err := oc.AdminConfigClient().ConfigV1().Networks().Apply(ctx, netConfigApply,
+			metav1.ApplyOptions{FieldManager: fieldManager, Force: true})
+		o.Expect(err).NotTo(o.HaveOccurred())
+	})
```
ricky-rav, line 39: Would it make sense to also check this state, either in a separate test or before each test begins?

test/extended/networking/network_diagnostics.go
```diff
@@ -0,0 +1,248 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func() bool {
+			// Should not affect the Progressing condition of network.operator
+			oper, err := oc.AdminOperatorClient().OperatorV1().Networks().Get(ctx, clusterConfig, metav1.GetOptions{})
+			if err != nil {
+				framework.Logf("Error getting cluster operator: %v", err)
+				return false
+			}
+			for _, operCondition := range oper.Status.Conditions {
```
ricky-rav, line 232: Should we also check that CNO isn't progressing in the other tests where `meta.IsStatusConditionFalse`?

test/extended/networking/network_diagnostics.go
```diff
@@ -0,0 +1,248 @@
...
+					),
+			),
+		)
+		_, err := oc.AdminConfigClient().ConfigV1().Networks().Apply(ctx, netConfigApply,
+			metav1.ApplyOptions{FieldManager: fieldManager, Force: true})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func() bool {
+			// Should not affect the Progressing condition of network.operator
+			oper, err := oc.AdminOperatorClient().OperatorV1().Networks().Get(ctx, clusterConfig, metav1.GetOptions{})
```
ricky-rav, line 227: would it be redundant to also check for no source pods here?

ricky-rav (commented):
Overall looks good, I left a few questions.

test/extended/networking/network_diagnostics.go
kyrtapz: I've added the test to verify the default behaviour.

test/extended/networking/network_diagnostics.go
kyrtapz: I don't think it is required in other cases as nothing impacts pod deployment.

test/extended/networking/network_diagnostics.go
kyrtapz: The pods will be there but it will be stack in scheduling as it's waiting for a valid node.

ricky-rav:
/lgtm
Thanks, @kyrtapz!

trozet (approved):

kyrtapz:
CI seems to be back
/retest

kyrtapz:
@dgoodwin are you ok with the cliet-go bump? It looks like the bot needs your blessing

kyrtapz:
Made the tests ordered so they do not run in parallel with each other:
```
Ordered is a decorator that allows you to mark a container as ordered.  Specs in the container will always run in the order they appear.
They will never be randomized and they will never run in parallel with one another, though they may run in parallel with other specs.
```

kyrtapz:
> I believe that moves them to the serial suite and thus they will then only run in serial jobs. I would not expect them to be running in e2e-gcp-ovn-techpreview anymore. How long does this set of tests take to run? Just want to ask in terms of time added to the suite.

Each test has a timeout of 3mins, from my testing each test doesn't take more than a minute.

dgoodwin:
> > I believe that moves them to the serial suite and thus they will then only run in serial jobs. I would not expect them to be running in e2e-gcp-ovn-techpreview anymore. How long does this set of tests take to run? Just want to ask in terms of time added to the suite.
> 
> Each test has a timeout of 3mins, from my testing each test doesn't take more than a minute.

That should be fine in the serial suite.

kyrtapz:
Rebased and ran `make update`
/test e2e-gcp-ovn-techpreview-serial

kyrtapz:
Network diagnostics tests passed previously, lets try again
/test e2e-gcp-ovn-techpreview-serial

kyrtapz:
Network diagnostics tests passed, the unrelated failure is already [tracked](https://issues.redhat.com/browse/OCPBUGS-30224)

kyrtapz:
Had to re-vendor after rebasing to master.

PR #28720, opened 2024-04-17
HOSTEDCP-1466: test/ginkgo: allow many external binaries

stevekuznetsov:
Not only do we ultimately need to support external binaries from more than one image in the payload, we also need to support repositories that use `go test` for their end-to-end tests along with the old-school Ginkgo approach.

/assign @soltysh 
/cc @deads2k

### DIFF
--- a/pkg/test/ginkgo/external.go
+++ b/pkg/test/ginkgo/external.go
@@ -6,11 +6,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/test/ginkgo/external.go
```diff
@@ -24,53 +26,135 @@ type serializedTest struct {
...
+}
+
+// extractGoTestRunnerBinariesFromReleaseImage is responsible for resolving the tag from
+// release image and extracting go test binaries, returns path to the binary or error.
+// The expectation is that a tree of directories containing test binaries and their package names
+// exists:
+// /tests/
+// ├── a
+// │ ├── b
+// │ ├── b.package.txt
```
stevekuznetsov, line 100: n.b. the Go module being tested is information that is lost during `go test -c` ... so we need to record it in a side-channel and consume that here.

pkg/test/ginkgo/external.go
```diff
@@ -113,19 +197,184 @@ func extractBinaryFromReleaseImage(tag, binary string) (string, error) {
...
+		if err == io.EOF {
+			break
+		}
+		var event TestEvent
+		err = json.Unmarshal([]byte(line), &event)
+		if err != nil {
+			return TestFailed, []byte(fmt.Errorf("failed unmarshalling test event from output %s: %w", string(testOutputBytes), err).Error())
+		}
+
+		// TODO: lots of events are possible here, and it's possible more than one test runs, etc
```
stevekuznetsov, line 356: Open to suggestions here, the execution model does not accurately map here.

PR #28722, opened 2024-04-18
NO-JIRA: Record Kube event first/last timestamp in the resulting intervals

dgoodwin:


dgoodwin:
Updated with a fix for the unit tests.

pkg/monitortests/testframework/watchevents/event.go
```diff
@@ -172,6 +172,10 @@ func recordAddOrUpdateEvent(
 		}
 		return
 	}
+
+	message = message.WithAnnotation("FirstTimestamp", obj.FirstTimestamp.Format(time.RFC3339))
```
eggfoobar, line 176: Small nit, was looking at the output data seems like the other annotations are in camel case, should we change this to that as well?

```json
                "annotations": {
                    "FirstTimestamp": "2024-04-20T05:35:19Z",
                    "LastTimestamp": "2024-04-20T06:00:28Z",
                    "count": "24",
                    "interesting": "true",
                    "pathological": "true",
                    "reason": "EtcdCertSignerControllerUpdatingStatus"
                }
```

dgoodwin:
Good catch, will update.

dgoodwin:
Updated, just needs an lgtm.

PR #28725, opened 2024-04-19
OCPBUGS-32477: Also rely on oomkilled exit code 137 in build test

ardaguclu:
In 4.15, when pod is killed due to insufficient memory, it raises an `Error` status rather than expected `OOMKilled`. This seems to be a bug in upstream but in order to unblock the presubmit jobs, we need to have a stop-gap solution. This PR also checks the `Error` status with the 137 exit code which corresponds to the OOMKilled.

Additionally, mariadb:10.3 version seems to be problematic in 4.15, so that this PR drops mariadb:10.3

### DIFF
--- a/test/extended/builds/failure_status.go
+++ b/test/extended/builds/failure_status.go
@@ -174,14 +174,34 @@ var _ = g.Describe("[sig-builds][Feature:Builds][Slow] update failure status", f
 				var build *buildv1.Build
 				wait.PollImmediate(200*time.Millisecond, 30*time.Second, func() (bool, error) {
 					build, err = oc.BuildClient().BuildV1().Builds(oc.Namespace()).Get(context.Background(), br.Build.Name, metav1.GetOptions{})
-					if build.Status.Reason != buildv1.StatusReasonOutOfMemoryKilled {
+					// In 4.15, status reason may be filed as Error rather than OOMKilled
+					// (tracked in https://issues.redhat.com/browse/OCPBUGS-32498) and also there is a similar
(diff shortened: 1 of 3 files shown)


test/extended/builds/failure_status.go
```diff
@@ -174,14 +174,32 @@ var _ = g.Describe("[sig-builds][Feature:Builds][Slow] update failure status", f
...
+					for _, c := range pod.Status.ContainerStatuses {
+						if c.State.Terminated == nil {
+							continue
+						}
+						if c.State.Terminated.ExitCode == 137 {
+							oomKilledExitCodeFound = true
+						}
+					}
+					o.Expect(oomKilledExitCodeFound).To(o.BeTrue())
+				}
```
adambkaplan, line 202: Won't we expect exit code 137, regardless of status? If so, then I don't think we need the if/else.

adambkaplan (approved):
/approve

Marking approval so that we unblock CVE fixes for openshift-controller-manager.
I expect this PR to be reverted or modified once the root cause is found.

I find it a bit odd that this test is not failing in OCP 4.16 - I would have expected us to see this unless there was an interim (latent?) bug in cri-o that was fixed in 4.16, but hasn't been backported to 4.15.

soltysh (approved):
/lgtm
/approve

crio-o issue for reference is https://github.com/cri-o/cri-o/issues/7035

adambkaplan:
Bug to update cri-o in 4.15 (it appears to have indeed been fixed for 4.16): https://issues.redhat.com/browse/OCPBUGS-32498

PR #28727, opened 2024-04-19
TRT-1610: Intervals, move over to new schema, finally...

dgoodwin:
This PR completes the months long process we've been working towards to smoothly migrate to a new schema for intervals. 

  * The non-spyglass intervals chart is attempted to port to structured locators/messages. Could be bugs here but my sample comparison came out looking the same.
  * Flat string locators/messages are removed. tempStructuredLocator and tempStructuredMessage are flipped over to those names, locator and message are now structured.
  * tempSource in json is now promoted to source. I have more plans for the concept and the premise has proven sound.

This will cut the size of intervals files and their charts quite substantially as the bulk of the data was duplicated.


DennisPeriquet:
/hold
/lgtm

Please cancel hold when ready to merge.

dgoodwin:
Sorry Dennis this needed another lgtm, merge conflict.

PR #28732, opened 2024-04-22
OCPBUGS-32293: Use In-Tree Template for Jenkins

adambkaplan:
Update the JenkinsPipeline build strategy test to use an in-tree template to deploy Jenkins during the test setup phase. This decouples the test from the Samples Operator, which is officially deprecated in 4.16 and will be phased out in the future. Previously, this test implicitly relied on the Samples Operator to install the Jenkins ephemeral template.

### DIFF
--- a/test/extended/builds/pipeline_origin_bld.go
+++ b/test/extended/builds/pipeline_origin_bld.go
@@ -14,6 +14,8 @@ import (
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 )
 
+// TODO: Remove this test if/when the JenkinsPipeline build strategy is removed.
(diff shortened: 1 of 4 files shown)


adambkaplan:
/assign @soltysh 

/cc @fbm3307 @sayan-biswas 

This will decouple the JenkinsPipeline build test test from the Samples Operator. I'm setting aside the fact that we technically called this thing deprecated in OCP 4.3!

adambkaplan:
/hold

This still has an implict dependency on the samples operator (Jenkins imagestream). Once I see the build test get back to "green," I'll fix that.

adambkaplan:
/hold cancel

PR updated

adambkaplan:
/retest

Build suite tests are passing! 😃

PR #28735, opened 2024-04-23
TRT-1576: Fail if operator has Available=False unless in upgrade window

DennisPeriquet:
For this test: `[bz-%v] clusteroperator/%v should not change condition/Available]`:

* For non-upgrade jobs, fail when operator goes to Available=False
* For upgrade-jobs, fail when operator goes to Available=False unless it's during an upgrade window and the condition lasts for less than 10 minutes.

Once the PR where [storage operator stops reporting Available status](https://github.com/openshift/cluster-storage-operator/pull/456) merges, we can remove the exception for it.

### DIFF
--- a/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
+++ b/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
@@ -45,12 +45,11 @@ func testStableSystemOperatorStateTransitions(events monitorapi.Intervals, clien
 		}
 
 		// For the non-upgrade case, if any operator has Available=False, fail the test.
-		if condition.Type == configv1.OperatorAvailable {
-
-			// We'll add an exception for single node for now.
-			if condition.Status == configv1.ConditionFalse && !isSingleNode {
-				return "", nil
+		if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionFalse {
+			if isSingleNode {
+				return "Operators are allowed to go degraded on single-node for now", nil
 			}
+			return "", nil
 		}
 		return "We are not worried about Degraded=True blips for stable-system tests yet.", nil
 	}
@@ -61,7 +60,7 @@ func testStableSystemOperatorStateTransitions(events monitorapi.Intervals, clien
 func isSingleNodeCheck(clientConfig *rest.Config) (bool, error) {
 	configClient, err := clientconfigv1.NewForConfig(clientConfig)
 	if err != nil {
-		logrus.Warnf("Error creating config client to check for Single Node configuration: %v", err)
+		logrus.WithError(err).Error("Error creating config client to check for Single Node configuration")
 		return false, err
 	}
 	return exutil.IsSingleNode(context.Background(), configClient)
--- a/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
+++ b/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
@@ -10,21 +10,24 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/origin/pkg/monitortests/clusterversionoperator/operatorstateanalyzer"
+	"github.com/sirupsen/logrus"
 
 	configv1 "github.com/openshift/api/config/v1"
+	clientconfigv1 "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 	"github.com/openshift/origin/pkg/monitortestlibrary/platformidentification"
 	platformidentification2 "github.com/openshift/origin/pkg/monitortestlibrary/platformidentification"
 	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+	exutil "github.com/openshift/origin/test/extended/util"
 	"k8s.io/client-go/rest"
 )
 
 // exceptionCallback consumes a suspicious condition and returns an
 // exception string if does not think the condition should be fatal.
-type exceptionCallback func(operator string, condition *configv1.ClusterOperatorStatusCondition, clientConfig *rest.Config) (string, error)
+type exceptionCallback func(operator string, condition *configv1.ClusterOperatorStatusCondition, eventInterval monitorapi.Interval, clientConfig *rest.Config) (string, error)
 
-func testStableSystemOperatorStateTransitions(events monitorapi.Intervals) []*junitapi.JUnitTestCase {
-	except := func(_ string, condition *configv1.ClusterOperatorStatusCondition, _ *rest.Config) (string, error) {
+func testStableSystemOperatorStateTransitions(events monitorapi.Intervals, clientConfig *rest.Config) []*junitapi.JUnitTestCase {
+	except := func(_ string, condition *configv1.ClusterOperatorStatusCondition, _ monitorapi.Interval, clientConfig *rest.Config) (string, error) {
 		if condition.Status == configv1.ConditionTrue {
 			if condition.Type == configv1.OperatorAvailable {
 				return fmt.Sprintf("%s=%s is the happy case", condition.Type, condition.Status), nil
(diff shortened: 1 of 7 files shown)


DennisPeriquet:
/payload-job periodic-ci-openshift-release-master-ci-4.16-e2e-vsphere-ovn-upgrade

This will see if my new exception allows the upgrade job to pass despite the single storage operator replica.

DennisPeriquet:
/payload-job periodic-ci-openshift-release-master-ci-4.16-e2e-vsphere-ovn-upgrade

retry because the last one didn't really run

DennisPeriquet:
/payload-job periodic-ci-openshift-release-master-ci-4.16-e2e-vsphere-ovn-upgrade

DennisPeriquet:
/payload-job periodic-ci-openshift-release-master-ci-4.16-e2e-vsphere-ovn-upgrade

DennisPeriquet:
re: [the last /payload with vsphere](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-origin-28735-ci-4.16-e2e-vsphere-ovn-upgrade/1787257947932332032):

```

: [bz-Image Registry] clusteroperator/image-registry should not change condition/Available | 1h34m34s
-- | --
{  4 unexpected clusteroperator state transitions during e2e test run.  These did not match any known exceptions, so they cause this test-case to fail:
May 06 00:30:25.569 E clusteroperator/image-registry condition/Available reason/NoReplicasAvailable status/False Available: The deployment does not have available replicas\nNodeCADaemonAvailable: The daemon set node-ca has available replicas\nImagePrunerAvailable: Pruner CronJob has been created May 06 00:30:25.569 - 51s   E clusteroperator/image-registry condition/Available reason/NoReplicasAvailable status/False Available: The deployment does not have available replicas\nNodeCADaemonAvailable: The daemon set node-ca has available replicas\nImagePrunerAvailable: Pruner CronJob has been created 
May 06 00:50:27.986 E clusteroperator/image-registry condition/Available reason/NoReplicasAvailable status/False Available: The deployment does not have available replicas\nNodeCADaemonAvailable: The daemon set node-ca has available replicas\nImagePrunerAvailable: Pruner CronJob has been created May 06 00:50:27.986 - 50s   E clusteroperator/image-registry condition/Available reason/NoReplicasAvailable status/False Available: The deployment does not have available replicas\nNodeCADaemonAvailable: The daemon set node-ca has available replicas\nImagePrunerAvailable: Pruner CronJob has been created
```

Those two events happened within the upgrade window (but the logs indicate no replicas, which I'm betting is why the test failed):

```
$ cat e2e-events_20240506-000613.json | jq '.items[] | select(.source == "KubeEvent" and .locator.keys.clusterversion? == "cluster")| "\(.from) \(.to) \(.message.reason)"'
"2024-05-06T00:07:00Z 2024-05-06T00:07:00Z UpgradeStarted"
"2024-05-06T00:58:25Z 2024-05-06T00:58:25Z UpgradeVersion"
"2024-05-06T00:58:25Z 2024-05-06T00:58:25Z UpgradeComplete"
```

wking:
> re: [the last /payload with vsphere](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-origin-28735-ci-4.16-e2e-vsphere-ovn-upgrade/1787257947932332032):

image

I'm not clear on why that run has an `Available=False` `image-registry` while we don't have any exceptions in place around that component today besides [a single-replica carve-out][1].  This wasn't a single-control-plane-node test:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-origin-28735-ci-4.16-e2e-vsphere-ovn-upgrade/1787257947932332032/ar
tifacts/e2e-vsphere-ovn-upgrade/gather-extra/artifacts/nodes.json | jq -r '.items[].metadata.name'
ci-op-6wykcgk2-d2645-7nlts-master-0
ci-op-6wykcgk2-d2645-7nlts-master-1
ci-op-6wykcgk2-d2645-7nlts-master-2
ci-op-6wykcgk2-d2645-7nlts-worker-0-6bdsp
ci-op-6wykcgk2-d2645-7nlts-worker-0-8c5wm
ci-op-6wykcgk2-d2645-7nlts-worker-0-kxfhn
```

And the cluster was configured for highly-available infrastructure (which includes the registry):

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-origin-28735-ci-4.16-e2e-vsphere-ovn-upgrade/1787257947932332032/artifacts/e2e-vsphere-ovn-upgrade/gather-must-gather/artifacts/must-gather.tar | tar -xOz registry-apps-build02-vmc-ci-openshift-org-ci-op-6wykcgk2-stable-sha256-e7b33149e705570ebcdcebe24c57af8336229175099fb5d53100330fd61015f1/cluster-scoped-resources/config.openshift.io/infrastructures/cluster.yaml | yaml2json | jq -r .status.infrastructureTopology
HighlyAvailable
```

And yet:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-origin-28735-ci-4.16-e2e-vsphere-ovn-upgrade/1787257947932332032/artifacts/e2e-vsphere-ovn-upgrade/gather-extra/artifacts/deployments.json | jq -r '.items[] | select(.metadata.name == "image-registry").spec.replicas'
1
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-origin-28735-ci-4.16-e2e-vsphere-ovn-upgrade/1787257947932332032/artifacts/e2e-vsphere-ovn-upgrade/gather-extra/artifacts/deployments.json | jq -c '.items[] | select(.metadata.name == "image-registry").spec.strategy'
{"type":"Recreate"}
```

I don't think the registry operator should be trying to wake the admin from sleep with an `Available=False` ClusterOperator condition, when it is configuring 1 replica and a `Recreate` strategy (which makes continual availability impossible).  Either the registry operator should configure its operand to be more available (in line with `infrastructureTopology: HighlyAvailable`), or it should accept that 1 `Recreate` pod will not be highly available and not alarm anyone on a brief, expected pod-handoff gap to match [the `Available` godocs contract][2].

[1]: https://github.com/openshift/origin/blame/15c35210bc8e0a1b06bd54e6fe5731d3a700b670/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go#L100-L102
[2]: https://github.com/openshift/api/blob/0a2abb2b630b660e5ac03d320b5c8ef2d36daad5/config/v1/types_cluster_operator.go#L160-L163

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -35,14 +35,94 @@ func testStableSystemOperatorStateTransitions(events monitorapi.Intervals) []*ju
...
+// isInUpgradeWindow determines if the given eventInterval falls within an upgrade window.
+// If we don't find any upgrade starting point, we assume the eventInterval is not in an upgrade window.
+// If we don't find any upgrade ending point, we assume the ending point is at the end of the test.
+func isInUpgradeWindow(eventList monitorapi.Intervals, eventInterval monitorapi.Interval) bool {
+	type upgradeWindowHolder struct {
+		startInterval *monitorapi.Interval
+		endInterval   *monitorapi.Interval
+	}
+
+	var upgradeWindows []*upgradeWindowHolder
```
dgoodwin, line 59: I'll leave this for you to decide, but I think an actual interval for the overall upgrade window using the logic here would be generally useful. That would mean a new minimal monitortest that just creates some new intervals with a new Source like UpgradeWindow and sets the from-to appropriately. Then this function could be scanning all intervals for the UpgradeWindows, see if our interval is within one, and we could chart the overall upgrade start/end times.

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -35,14 +35,94 @@ func testStableSystemOperatorStateTransitions(events monitorapi.Intervals) []*ju
...
+					currentWindow.endInterval = &monitorapi.Interval{
+						Condition: monitorapi.Condition{
+							Message: monitorapi.Message{
+								Reason: event.Message.Reason,
+							},
+						},
+						From: event.From,
+						To:   event.To,
+					}
+				}
```
dgoodwin, line 83: I'm not sure if your assumption is correct, did you confirm that UpgradeRollback is not possible after UpgradeComplete? I would expect it to be but I don't know for sure. If so you'd want to handle currentWindow == nil here and open a new one?

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -97,9 +186,17 @@ func testUpgradeOperatorStateTransitions(events monitorapi.Intervals, clientConf
 			if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionFalse && condition.Reason == "ClusterServiceVersionNotSucceeded" {
 				return "https://issues.redhat.com/browse/OCPBUGS-23744", nil
 			}
-		case "image-registry":
-			if replicaCount, _ := checkReplicas("openshift-image-registry", operator, clientConfig); replicaCount == 1 {
-				return "image-registry has only single replica", nil
+		case "image-registry", "storage":
```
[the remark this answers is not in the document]
DennisPeriquet: I removed storage since https://github.com/openshift/cluster-storage-operator/pull/456 is now merged.

dgoodwin:
/hold

Something doesn't look right in those risk analysis failures...

DennisPeriquet:
In the meantime, I'm pushing a change inside `vendor/` that skips the test to see how the SNO serial job does.

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -35,14 +38,118 @@ func testStableSystemOperatorStateTransitions(events monitorapi.Intervals) []*ju
...
+		isSingleNode, err := isSingleNodeCheck(clientConfig)
+		if err != nil {
+			logrus.Warnf("Error checking for Single Node configuration on stable system (unable to make exception): %v", err)
+			isSingleNode = false
+		}
+
+		// For the non-upgrade case, if any operator has Available=False, fail the test.
+		if condition.Type == configv1.OperatorAvailable {
+
+			// We'll add an exception for single node for now.
```
dgoodwin, line 50: Confusing comment I think, says we're adding an exception but we don't actually return the exception, we return empty string and nil which appears to actually mean no exception. I think we then trigger the degraded=True message below as the exception, which isn't even accurate for Available=false. 

I would suggest something like:

```
if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionFalse {

			// We'll add an exception for single node for now.
                        if isSingleNode {
                               return "Operators are allowed to go degraded on single-node for now", nil 
                       }
			return "", nil
		}
```

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -35,14 +38,118 @@ func testStableSystemOperatorStateTransitions(events monitorapi.Intervals) []*ju
...
+		return "We are not worried about Degraded=True blips for stable-system tests yet.", nil
+	}
+
+	return testOperatorStateTransitions(events, []configv1.ClusterStatusConditionType{configv1.OperatorAvailable, configv1.OperatorDegraded}, except, clientConfig)
+}
+
+func isSingleNodeCheck(clientConfig *rest.Config) (bool, error) {
+	configClient, err := clientconfigv1.NewForConfig(clientConfig)
+	if err != nil {
+		logrus.Warnf("Error creating config client to check for Single Node configuration: %v", err)
```
dgoodwin, line 64: log.WithError(err).Error()

DennisPeriquet:
The [pull-ci-openshift-origin-master-e2e-aws-ovn-single-node-upgrade](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-origin-master-e2e-aws-ovn-single-node-upgrade) does not have any passing jobs in the past 2 months.

PR #28737, opened 2024-04-24
MON-3513: Add availability test for Metrics API

machine424:
This should ensure the availability of the Metrics API during e2e tests including upgrades.
Thus it should also help with https://issues.redhat.com/browse/MON-3539.
    
The correctness of the API: whether the right/expected content is returned, should be tested elsewhere (we already have tests for that in CMO, and the HPA tests already make use of that etc.). This tests only check the availability.

- Got inspired by the existing availability tests.
- Maybe this could be generalized for all the aggregation layer?

### DIFF
--- a/pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
+++ b/pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
@@ -1,0 +1,130 @@
+package disruptionmetricsapi
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/openshift/origin/pkg/monitortestframework"
+
+	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/openshift/origin/pkg/monitor/backenddisruption"
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/monitortestlibrary/disruptionlibrary"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+)
+
+const (
+	monitoringNamespace            = "openshift-monitoring"
+	prometheusAdapterDeployentName = "prometheus-adapter"
+	metricsServerDeploymentName    = "metrics-server"
+)
+
+type availability struct {
+	disruptionChecker  *disruptionlibrary.Availability
+	notSupportedReason error
+}
+
+func NewAvailabilityInvariant() monitortestframework.MonitorTest {
+	return &availability{}
+}
+
+func createBackendSampler(clusterConfig *rest.Config, disruptionBackendName, url string, connectionType monitorapi.BackendConnectionType) (*backenddisruption.BackendSampler, error) {
+	backendSampler, err := backenddisruption.NewAPIServerBackend(clusterConfig, disruptionBackendName, url, connectionType)
+	if err != nil {
+		return nil, err
+	}
+	backendSampler = backendSampler.WithUserAgent(fmt.Sprintf("openshift-external-backend-sampler-%s-%s", connectionType, disruptionBackendName))
+	return backendSampler, nil
+}
+
+func (w *availability) StartCollection(ctx context.Context, adminRESTConfig *rest.Config, recorder monitorapi.RecorderWriter) error {
+	var err error
+
+	kubeClient, err := kubernetes.NewForConfig(adminRESTConfig)
+	if err != nil {
+		return err
+	}
+	var deployment *appsv1.Deployment
+	deployment, err = kubeClient.AppsV1().Deployments(monitoringNamespace).Get(ctx, metricsServerDeploymentName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		// TODO: remove this in 4.17
+		deployment, err = kubeClient.AppsV1().Deployments(monitoringNamespace).Get(ctx, prometheusAdapterDeployentName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
+	}
+	// Skip for single replica Deployments.
+	if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas == 1 {
+		w.notSupportedReason = &monitortestframework.NotSupportedError{Reason: fmt.Sprintf("%s only has a single replica", deployment.Name)}
+		return w.notSupportedReason
+	}
+
+	disruptionBackedName := "metrics-api"
+	newConnectionTestName := fmt.Sprintf("[sig-instrumentation] disruption/%s connection/new should be available throughout the test", disruptionBackedName)
+	reusedConnectionTestName := fmt.Sprintf("[sig-instrumentation] disruption/%s connection/reused should be available throughout the test", disruptionBackedName)
+	url := "/apis/metrics.k8s.io/v1beta1"
+
+	newConnections, err := createBackendSampler(adminRESTConfig, disruptionBackedName, url, monitorapi.NewConnectionType)
+	if err != nil {
+		return err
+	}
+	reusedConnections, err := createBackendSampler(adminRESTConfig, disruptionBackedName, url, monitorapi.ReusedConnectionType)
+	if err != nil {
+		return err
+	}
+
+	w.disruptionChecker = disruptionlibrary.NewAvailabilityInvariant(
+		newConnectionTestName, reusedConnectionTestName,
+		newConnections, reusedConnections,
+	)
+
+	if err := w.disruptionChecker.StartCollection(ctx, adminRESTConfig, recorder); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (w *availability) CollectData(ctx context.Context, storageDir string, beginning, end time.Time) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error) {
+	if w.notSupportedReason != nil {
+		return nil, nil, w.notSupportedReason
+	}
+	// we failed and indicated it during setup.
+	if w.disruptionChecker == nil {
+		return nil, nil, nil
+	}
+
+	return w.disruptionChecker.CollectData(ctx)
+}
+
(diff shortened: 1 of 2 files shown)


pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
```diff
@@ -0,0 +1,170 @@
...
+		w.notSupportedReason = &monitortestframework.NotSupportedError{Reason: fmt.Sprintf("%s only has a single replica", deployment.Name)}
+		return w.notSupportedReason
+	}
+
+	disruptionBackedName := "metrics-api"
+
+	newConnectionTestName := "[sig-instrumentation] disruption/metrics-api connection/new should be available throughout the test"
+	reusedConnectionTestName := "[sig-instrumentation] disruption/metrics-api connection/reused should be available throughout the test"
+
+	// TODO: clean up/refactor following.
```
machine424, line 77: TBD before merge

pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
```diff
@@ -0,0 +1,170 @@
...
+	if w.notSupportedReason != nil {
+		return nil, nil, w.notSupportedReason
+	}
+
+	intervals := monitorapi.Intervals{}
+	junits := []*junitapi.JUnitTestCase{}
+	errs := []error{}
+
+	for i := range w.disruptionCheckers {
+		// TODO: check for nil after refactoring in StartCollection
```
machine424, line 128: TBD before merge

pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas == 1 {
+		w.notSupportedReason = &monitortestframework.NotSupportedError{Reason: fmt.Sprintf("%s only has a single replica", deployment.Name)}
+		return w.notSupportedReason
+	}
+
+	disruptionBackedName := "metrics-api"
+
+	newConnectionTestName := "[sig-instrumentation] disruption/metrics-api connection/new should be available throughout the test"
+	reusedConnectionTestName := "[sig-instrumentation] disruption/metrics-api connection/reused should be available throughout the test"
```
dgoodwin, line 75: Just want to share some info on these tests, they're extremely forgiving, P99 over past three weeks + a healthy margin before they'll fail. Our primary system for detecting real regressions uses large sets of data and alerting, but this dashboard provides a strong visualization of that same data: https://grafana-loki.ci.openshift.org/d/ISnBj4LVk/disruption?orgId=1 Very handy if you want to monitor how your component is doing. Once this lands, it'll eventually start appearing in TRT's alerting if something is wrong and we'll get in touch if we detect a sustained change.

pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// For pods metrics, monitoringNamespace is always available and we don't want to ask cluster wide.
+	newConnections, err = createAPIServerBackendSampler(adminRESTConfig, disruptionBackedName, fmt.Sprintf("/apis/metrics.k8s.io/v1beta1/namespaces/%s/pods", monitoringNamespace), monitorapi.NewConnectionType)
+	if err != nil {
+		return err
+	}
+	reusedConnections, err = createAPIServerBackendSampler(adminRESTConfig, disruptionBackedName, fmt.Sprintf("/apis/metrics.k8s.io/v1beta1/namespaces/%s/pods", monitoringNamespace), monitorapi.ReusedConnectionType)
+	if err != nil {
+		return err
+	}
```
dgoodwin, line 102: It looks like you're re-using the disruptionBackedName here, which is probably a bug as both results are probably getting muddled or ignored, but do you actually need a second set of endpoints being monitored here? Typically we'd only have one pair, making sure the APIserver handling the request is up, not that the various APIs within that server are working. Assuming these are handled by the same APIServer, I'd pitch dropping one of these sets and just hit one new+reused pair.  The cpu and network load of these requests every second is not insignificant.

dgoodwin (commented):
I'll need to check back once you've got job results to make sure the data in artifacts looks good, but this is really cool, and one of very few instances where someone has added additional disruption monitoring for their component.

machine424:
> I'll need to check back once you've got job results to make sure the data in artifacts looks good, but this is really cool, and one of very few instances where someone has added additional disruption monitoring for their component.

Thanks, I'll check with the team, and maybe we'll add more for the other components (prometheus, etc.). The monitortest framework really makes it easy to write such tests, good job on that :)

What do you think about adding (in another PR), a generic disruptioncheck for all the aggregation layer APIs?, I assume it's possible to get the available APIs from Kube or we can provide a list of URLs...

I'll continue testing, make the tests fail learn more about the behavior, then I'll ask you for another review, once the CI is greener :)

pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
```diff
@@ -0,0 +1,170 @@
...
+
+	disruptionBackedName := "metrics-api"
+
+	newConnectionTestName := "[sig-instrumentation] disruption/metrics-api connection/new should be available throughout the test"
+	reusedConnectionTestName := "[sig-instrumentation] disruption/metrics-api connection/reused should be available throughout the test"
+
+	// TODO: clean up/refactor following.
+
+	// For nodes metrics
+	newConnections, err := createAPIServerBackendSampler(adminRESTConfig, disruptionBackedName, "/apis/metrics.k8s.io/v1beta1/nodes", monitorapi.NewConnectionType)
```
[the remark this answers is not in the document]
machine424: It'll depend on the nodes count.
I pushed a new version where we only ask for the metrics of the Metrics API backend Pods themselves.

pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
machine424: Really interesting, thanks for the details.
We'll try ti have a look at those dashboards from time to time.

pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Skip for single replica Deployments.
+	kubeClient, err := kubernetes.NewForConfig(adminRESTConfig)
+	if err != nil {
+		return err
+	}
+	var deployment *appsv1.Deployment
+	deployment, err = kubeClient.AppsV1().Deployments(monitoringNamespace).Get(ctx, metricsServerDeploymentName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		// TODO: remove this in 4.17
```
machine424, line 59: TODO: open ticket for this before merge

machine424:
/payload-with-prs pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade openshift/api#1865 openshift/cluster-monitoring-operator#2268

machine424:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.16-e2e-aws-ovn-upgrade openshift/api#1865 openshift/cluster-monitoring-operator#2268

machine424:
/payload-job-with-prs pull-ci-openshift-origin-master-e2e-aws-ovn-upgrade openshift/api#1865 openshift/cluster-monitoring-operator#2268

machine424:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.16-e2e-aws-ovn openshift/api#1865 openshift/cluster-monitoring-operator#2268

dgoodwin:
> > I'll need to check back once you've got job results to make sure the data in artifacts looks good, but this is really cool, and one of very few instances where someone has added additional disruption monitoring for their component.
> 
> Thanks, I'll check with the team, and maybe we'll add more for the other components (prometheus, etc.). The monitortest framework really makes it easy to write such tests, good job on that :)
> 
> What do you think about adding (in another PR), a generic disruptioncheck for all the aggregation layer APIs?, I assume it's possible to get the available APIs from Kube or we can provide a list of URLs...
> 
> I'll continue testing, make the tests fail learn more about the behavior, then I'll ask you for another review, once the CI is greener :)

It's an awesome idea but we've caught this framework noticably impacting CPU/memory/network when the polling increased dramatically. If this involved a bunch of new endpoints, it very likely could start to skew CPU at the very least in ways that show up and send people on wild goose chases. If we knew about it and were expecting it, perhaps we could get away with it though, provided the resource requirements are not significantly different. How many endpoints would we be talking about?

pkg/monitortests/monitoring/disruptionmetricsapi/monitortest.go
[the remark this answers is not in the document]
machine424: Actually, I think that checking "/apis/metrics.k8s.io/v1beta1" is sufficient.
The response looks like
```
oc get --raw "/apis/metrics.k8s.io/v1beta1"  | jq
{
  "kind": "APIResourceList",
  "apiVersion": "v1",
  "groupVersion": "metrics.k8s.io/v1beta1",
  "resources": [
    {
      "name": "nodes",
      "singularName": "node",
      "namespaced": false,
      "kind": "NodeMetrics",
      "verbs": [
        "get",
        "list"
      ]
    },
    {
      "name": "pods",
      "singularName": "pod",
      "namespaced": true,
      "kind": "PodMetrics",
      "verbs": [
        "get",
        "list"
      ]
    }
  ]
}
```
And when the backends are down:
```
Error from server (ServiceUnavailable): the server is currently unable to handle the request
```

machine424:
> It's an awesome idea but we've caught this framework noticably impacting CPU/memory/network when the polling increased dramatically. If this involved a bunch of new endpoints, it very likely could start to skew CPU at the very least in ways that show up and send people on wild goose chases. If we knew about it and were expecting it, perhaps we could get away with it though, provided the resource requirements are not significantly different. How many endpoints would we be talking about?

I see, I just need to discuss that with the team but I can think of 3 or 4 APIs.
Regarding the generic test for the other aggreg APIs, I was suggesting that for the other teams (that may be interested in such tests), we, monitoring, only provide the "Metrics API"

dgoodwin:
Code looks good, I see the data coming out looking good. 

Adding another 3-4 of these we could probably get away with, your generic framework could be very useful.

/lgtm

PR #28739, opened 2024-04-24
OCPBUGS-18534: monitor test fix to wait before connecting to a non-existent dns on PowerVS and IBMCloud platforms

jcho02:
[OCPBUGS-18534](https://issues.redhat.com/browse/OCPBUGS-18534)  
Monitor test changes to fix loadbalancer issue in PowerVS platforms.

Test checks if platform is either a PowerVS or IBMCloud cluster and then runs function `checkHostnameReady` where it debugs the master node of the cluster and verifies the hostname is active before hitting through the load balancer

### DIFF
--- a/pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
+++ b/pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/openshift/origin/pkg/monitortestframework"
@@ -203,6 +204,15 @@ func (w *availability) StartCollection(ctx context.Context, adminRESTConfig *res
 		return fmt.Errorf("error creating PDB: %w", err)
 	}
 
+	// On certain platforms hitting the hostname before it is ready leads to a blackhole, this code checks
+	// the host from the cluster's context
+	if infra.Spec.PlatformSpec.Type == configv1.PowerVSPlatformType || infra.Spec.PlatformSpec.Type == configv1.IBMCloudPlatformType {
+		nodeTgt := "node/" + nodeList.Items[0].ObjectMeta.Name
+		if err := checkHostnameReady(tcpService, nodeTgt); err != nil {
+			return err
+		}
+	}
+
 	// Hit it once before considering ourselves ready
 	fmt.Fprintf(os.Stderr, "hitting pods through the service's LoadBalancer\n")
 	timeout := 10 * time.Minute
@@ -343,4 +353,35 @@ func httpGetNoConnectionPoolTimeout(url string, timeout time.Duration) (*http.Re
 	}
 
 	return client.Get(url)
+}
+
+// Uses the first node in the cluster to verify the LoadBalancer host is active before returning
+func checkHostnameReady(tcpService *corev1.Service, nodeTgt string) error {
+	oc := exutil.NewCLIForMonitorTest(tcpService.GetObjectMeta().GetNamespace())
+
+	var (
+		stdOut string
+		err    error
+	)
+
+	for i := 0; i < 60; i++ {
+		fmt.Printf("Checking load balancer host is active \n")
+		wait.Poll(3*time.Second, 60*time.Second, func() (bool, error) {
+			stdOut, _, err = oc.AsAdmin().WithoutNamespace().RunInMonitorTest("debug").Args(nodeTgt, "--", "dig", "+short", tcpService.Status.LoadBalancer.Ingress[0].Hostname).Outputs()
+			if err != nil {
+				return false, nil
+			}
+			return true, nil
+		})
+
+		output := strings.TrimSpace(stdOut)
+		if output == "" {
+			fmt.Println("waiting for the LB to come active")
+			time.Sleep(1 * time.Minute)
+			continue
+		} else {
+			break
+		}
+	}
+	return nil
 }
--- a/pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
+++ b/pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
@@ -208,7 +208,7 @@ func (w *availability) StartCollection(ctx context.Context, adminRESTConfig *res
 	// the host from the cluster's context
 	if infra.Spec.PlatformSpec.Type == configv1.PowerVSPlatformType || infra.Spec.PlatformSpec.Type == configv1.IBMCloudPlatformType {
 		nodeTgt := "node/" + nodeList.Items[0].ObjectMeta.Name
-		if err := checkHostnameReady(tcpService, nodeTgt); err != nil {
+		if err := checkHostnameReady(ctx, tcpService, nodeTgt); err != nil {
 			return err
 		}
 	}
@@ -356,7 +356,7 @@ func httpGetNoConnectionPoolTimeout(url string, timeout time.Duration) (*http.Re
 }
 
 // Uses the first node in the cluster to verify the LoadBalancer host is active before returning
-func checkHostnameReady(tcpService *corev1.Service, nodeTgt string) error {
+func checkHostnameReady(ctx context.Context, tcpService *corev1.Service, nodeTgt string) error {
 	oc := exutil.NewCLIForMonitorTest(tcpService.GetObjectMeta().GetNamespace())
 
 	var (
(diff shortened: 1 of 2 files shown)


pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
```diff
@@ -344,3 +355,31 @@ func httpGetNoConnectionPoolTimeout(url string, timeout time.Duration) (*http.Re
...
+func checkHostnameReady(tcpService *corev1.Service, nodeTgt string) error {
+	for i := 0; i < 60; i++ {
+		lbTgt := tcpService.Status.LoadBalancer.Ingress[0].Hostname
+		cmd := exec.Command("oc", "debug", nodeTgt, "--", "/bin/bash", "-c", "dig +short "+lbTgt)
+		out := &bytes.Buffer{}
+		errOut := &bytes.Buffer{}
+		cmd.Stdout = out
+		cmd.Stderr = errOut
+
+		if err := cmd.Run(); err != nil {
```
jaypoulz, line 370: I'm not super familiar with the run command, but it looks like the expectation is for this to run synchronously. My suggestion would be to set an explicit timeout to cover a case where scheduling the debug pod gets stuck in limbo for whatever reason. If we have to cancel the command upon said timeout, it should be safe to fail out, since being unable to schedule for whatever reason is something of concern. (But there should probably an error message indicating what happened).

jaypoulz:
Overall I'm happy with this patch. I left a note about a potential concern we may want to address. In order for this to be approved, you'll need someone to approve it from this list: https://github.com/openshift/origin/blob/master/OWNERS#L12-L18.

If he has time, @deads2k has helped provide very helpful feedback for our changes to disruption tests before. Otherwise, consider asking in slack for a review from #forum-ocp-release-oversight. I think a lot of the folks on this list are on the TRT team.

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
prb112: Oh great - point, let me coordinate with Jason

prb112: Addressed in latest commit

prb112:
> If he has time, @deads2k has helped provide very helpful feedback for our changes to disruption tests before. Otherwise, consider asking in slack for a review from #forum-ocp-release-oversight. I think a lot of the folks on this list are on the TRT team.

Done

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
```diff
@@ -344,3 +356,31 @@ func httpGetNoConnectionPoolTimeout(url string, timeout time.Duration) (*http.Re
...
 }
+
+// Uses the first node in the cluster to verify the LoadBalancer host is active before returning
+func checkHostnameReady(tcpService *corev1.Service, nodeTgt string) error {
+	for i := 0; i < 60; i++ {
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+		defer cancel()
+		fmt.Printf("Checking load balancer host is active \n")
+		lbTgt := tcpService.Status.LoadBalancer.Ingress[0].Hostname
+		cmd := exec.CommandContext(ctx, "oc", "debug", nodeTgt, "--", "/bin/bash", "-c", "dig +short "+lbTgt)
```
neisw, line 367: We were discussing something similar in another piece of code earlier today.  Are you familiar with the `oc` wrappers vs exec command?  Could be worth switching just for the sake of future maintainability 

Example: https://github.com/openshift/origin/blob/a1d43cfa1c5d7869f375aa1f8f64a97e5b4ba8cf/test/extended/util/nodes.go#L44

prb112: Yes, we are - interesting thing in this code, we couldn't get it to run in the context of this test.

prb112: We can double check.

neisw: One other thing worth reviewing is the loop / sleep.  [PollUntilContextTimeout](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout) would be another consideration.  Looks like worst case scenario this could block for 2 hours?  Understand that isn't likely and doubtful the exec would take a minute to timeout each time.

prb112: I read that there are cases where LB take over 45 minutes (and in our testing too). So we opted to expand to 60 minutes to give a buffer during activation. The PollUntilContextTimeout could be lowered form our perspective to 10 seconds, we opted to think in the worst of case.

neisw: My thought was if you used PollUntilContextTimeout you could set your max timeout to 45, 60, whatever you needed.  Then you wouldn't need the loop or sleep.  You could still use the 1 minute context for you exec (or oc call if that works) and configure whatever delay you wanted in between calls, currently 1 minute.

Similar to the cleanup call that has an overall timeout for the main call but then calls a helper method.  
 https://github.com/openshift/origin/blob/a1d43cfa1c5d7869f375aa1f8f64a97e5b4ba8cf/pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go#L314

jcho02: Running `oc.AsAdmin()` in the latest commit

jcho02:
Expected error message
`Checking load balancer host is active 
  May  2 15:55:05.788: INFO: Running 'oc debug node/rdr-multi-ca-ocp416-c6kqv-master-0 -- dig +short XYZABC-us-south.lb.appdomain.cloud'
waiting for the LB to come active
Checking load balancer host is active 
  May  2 15:56:11.138: INFO: Running 'oc debug node/rdr-multi-ca-ocp416-c6kqv-master-0 -- dig +short XYZABC-us-south.lb.appdomain.cloud'
  May  2 15:56:17.396: INFO: Error running /Users/paulbastide/.local/bin/oc debug node/rdr-multi-ca-ocp416-c6kqv-master-0 -- dig +short XYZABC-us-south.lb.appdomain.cloud:
  StdOut>

  StdErr>
  Starting pod/rdr-multi-ca-ocp416-c6kqv-master-0-debug-wxmhr ...
  To use host binaries, run `chroot /host`

  Removing debug pod ...
  error: unable to create the debug pod "rdr-multi-ca-ocp416-c6kqv-master-0-debug-wxmhr"`

Expected message on a successful run
 `May  2 15:56:20.138: INFO: Running 'oc debug node/rdr-multi-ca-ocp416-c6kqv-master-0 -- dig +short XYZABC-us-south.lb.appdomain.cloud'
waiting for the LB to come active
Checking load balancer host is active 
  May  2 15:57:25.511: INFO: Running 'oc debug node/rdr-multi-ca-ocp416-c6kqv-master-0 -- dig +short XYZABC-us-south.lb.appdomain.cloud'
waiting for the LB to come active
Checking load balancer host is active 
  May  2 15:58:30.867: INFO: Running 'oc debug node/rdr-multi-ca-ocp416-c6kqv-master-0 -- dig +short XYZABC-us-south.lb.appdomain.cloud'
hitting pods through the service's LoadBalancer
All monitor tests started.
Monitor started, waiting for ctrl+C to stop...`

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
```diff
@@ -344,3 +354,34 @@ func httpGetNoConnectionPoolTimeout(url string, timeout time.Duration) (*http.Re
...
+		if output == "" {
+			fmt.Println("waiting for the LB to come active")
+			time.Sleep(1 * time.Minute)
+			continue
+		} else {
+			break
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
prb112: Sure - we can look at it

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
```diff
@@ -344,3 +354,29 @@ func httpGetNoConnectionPoolTimeout(url string, timeout time.Duration) (*http.Re
...
+// Uses the first node in the cluster to verify the LoadBalancer host is active before returning
+func checkHostnameReady(ctx context.Context, tcpService *corev1.Service, nodeTgt string) error {
+	oc := exutil.NewCLIForMonitorTest(tcpService.GetObjectMeta().GetNamespace())
+
+	var (
+		stdOut string
+		err    error
+	)
+
+	wait.PollUntilContextTimeout(ctx, 15*time.Second, 60*time.Minute, true, func(ctx context.Context) (bool, error) {
```
neisw, line 367: Just double checking, this is a conscious choice not to capture the err returned from PollUntilContextTime ( could be context closed or timeout, etc.) and only return the last err, if any, returned from oc.AsAdmin call?

jcho02: Yes it was intentional to not capture the err

neisw: I have to admit I'm curious.  So this allows for a path where PollUntilContextTimeout returns an error but checkHostNameReady does not and that will allow StartCollection to proceed and presumably error out down the line?

jcho02: Yes it will error out here right after the checkHostNameReady function call since the service LoadBalancer is not verified
```      
fmt.Fprintf(os.Stderr, "hitting pods through the service's LoadBalancer\n")
timeout := 10 * time.Minute
```
The test will time out

neisw: Ahh, thanks for the explanation!

test/extended/util/client.go
```diff
@@ -179,6 +179,32 @@ func NewCLIWithoutNamespace(project string) *CLI {
 	return cli
 }
 
+// NewCLIForMonitorTest initializes the CLI and Kube framework helpers
+// without a namespace. Should be called outside of a Ginkgo .It()
+// function.
+func NewCLIForMonitorTest(project string) *CLI {
```
neisw, line 185: So you needed new non ginko versions of this and Run to support running in the monitor test environment?

Thanks for setting those up.

prb112: Correct

neisw:

e2e-gcp-ovn-builds is known to be failing outside of this pr and is being investigated

PR #28741, opened 2024-04-25
OCPBUGS-32923: Fix allowed firing alerts not triggering

dgoodwin:
This looks to have been partially implemented some time ago and thus has been broken ever since.

### DIFF
--- a/pkg/alerts/allowed_conformance.go
+++ b/pkg/alerts/allowed_conformance.go
@@ -2,47 +2,49 @@ package alerts
 
 import (
 	configv1 "github.com/openshift/api/config/v1"
-	helper "github.com/openshift/origin/test/extended/util/prometheus"
 )
 
 // AllowedAlertsDuringConformance lists all alerts that are allowed to be pending or firing during
 // conformance testing.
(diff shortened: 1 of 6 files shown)


pkg/alerts/allowed_conformance.go
```diff
@@ -2,61 +2,63 @@ package alerts
 
 import (
 	configv1 "github.com/openshift/api/config/v1"
-	helper "github.com/openshift/origin/test/extended/util/prometheus"
```
dgoodwin, line 5: I relocated the MetricCondition types to avoid a circular import when i started adding unit tests for the helper, which needed to access the allowed alerts in this package.

pkg/alerts/types.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Text is the description of why this alert condition matched.
+	Text string
+
+	Matches func(sample *model.Sample) bool
+}
+
+type MetricConditions []MetricCondition
+
+func (c MetricConditions) MatchesInterval(alertInterval monitorapi.Interval) *MetricCondition {
```
dgoodwin, line 21: Key fix was in here, the old version of this function was only partially implemented and thus our alert matching for allowed alerts hasn't been working for likely months.

PR #28738, opened 2024-04-24
TRT-1618: Allow Upgradeable:False with OpenShiftSDN

danwinship:
Unbreak the payload by updating the tests to allow an Upgradeable: False condition for OpenShift SDN clusters.

PR #28742, opened 2024-04-25
trt-1618: Allow Upgradeable:False with OpenShiftSDN

neisw:
Skip Upgradeable:False check for OpenshiftSDN 4.16 and above


sdodson:
/close
#28738 has merged

PR #28748, opened 2024-04-26
Revert "TRT-1625: Revert #28741 "OCPBUGS-32923: Fix allowed firing alerts not triggering""

dgoodwin:
Reverts openshift/origin#28747

False alarm, this wasn't the source of the problem and the failures were likely real. Thus no fix commit layered on top.


dgoodwin:

Broken everywhere.

PR #28755, opened 2024-04-29
OCPBUGS-33125: Allow flexible rhel in tests

joepvd:


ardaguclu:
I think, we don't need to run payload blocking nightlies for this test. Because current passing jobs prove that the patch is working;
/hold cancel

PR #28761, opened 2024-05-01
OCPBUGS-31868: allow for some errors checking namespace delete

jluhrsen:


pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
```diff
@@ -301,8 +301,12 @@ func (w *availability) Cleanup(ctx context.Context) error {
 		log.Info("waiting for namespace deletion to complete")
 		err := wait.PollUntilContextTimeout(ctx, 15*time.Second, 20*time.Minute, true, w.namespaceDeleted)
 		if err != nil {
-			log.WithError(err).Error("error waiting for namespace to delete")
-			return err
+			if errors.Is(err, context.DeadlineExceeded) {
+				log.Errorf("Timed out after 20 minutes waiting for deleted namespace: %s, %s", w.namespaceName, err)
+				return err
+			} else {
+				log.Errorf("Encountered error while waiting for deleted namespace: %s, %s", w.namespaceName, err)
```
neisw, line 308: Could keep a successive error count, clear the value any time we don't get an error, bail when it gets to 10 or something just so we don't spin for 15 minutes if things are really busted..

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
```diff
@@ -281,7 +282,6 @@ func (w *availability) namespaceDeleted(ctx context.Context) (bool, error) {
 	}
 
 	if err != nil {
-		logrus.Errorf("Error checking for deleted namespace: %s, %s", w.namespaceName, err.Error())
```
neisw, line 284: Do you mean to suppress the return false here?  Looks like this just move the logging up but the comment in the jira had me thinking you wanted to keep polling.

neisw: I suppose it would be return false, nil to keep polling but still log the error?

jluhrsen: I think I confused myself that a `return false, err` would not stop the polling. but that's actually what was already happening and how we got to this bug.

but, if we `return false, nil` then I'm not sure we have a way to keep polling. 🤔

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
jluhrsen: I realize this doesn't work after thinking about your other comment. I have a new idea coming.

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
jluhrsen, line 1: @neisw , how about something like this.

bumped the poller to 30s and then added 6 tries in the namespaceDeleted() in case it hits an error. So, if we did hit some error 6 times in a row then we log all of them and pass the last one back out to the poller which will exit and fail the test case.

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
neisw: I had to go look it up earlier myself.  I think by returning `false, nil` it will continue to process.  So you could keep the log entry but return nil for the error.  This doesn't accomplish my other suggestion but that was just a thought.

[PollUntilContextTimeout](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout)
```
PollUntilContextCancel tries a condition func until it returns true, an error, or the context is cancelled or hits a deadline
```

pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go
```diff
@@ -275,16 +275,18 @@ func (w *availability) WriteContentToStorage(ctx context.Context, storageDir, ti
 }
 
 func (w *availability) namespaceDeleted(ctx context.Context) (bool, error) {
-	_, err := w.kubeClient.CoreV1().Namespaces().Get(ctx, w.namespaceName, metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		return true, nil
-	}
+	for retry := 0; retry < 6; retry++ {
+		_, err := w.kubeClient.CoreV1().Namespaces().Get(ctx, w.namespaceName, metav1.GetOptions{})
```
neisw, line 279: I think being in the loop here will just hammer it 6 times in a row without a pause in between and then exit.  It is the PollUntilContextTimeout that waits in between the calls.
https://github.com/openshift/origin/blob/e07b8e3e77ae250299f9bbb6bcd135ef2f6408d9/pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go#L304

neisw: Well, I missed the sleep that you have but still not sure this is the best way.

neisw: Ok, so if the error is nil we return false, nil

If the error is not nil and not `IsNotFound` we try up to 6 times with a 5 second sleep in between.  But we still always return false, nil.  So really when we just try a few more times in the same call with a smaller interval when we see an non `IsNotFound` but ultimately keep going until we return true or timeout..

neisw: Are you wanting to quit polling after 6 consecutive failures?  Maybe preserve the error and return it at the end.
so `var err error` outside the loop and return `false,err` at the end of the function.  Should only return there after the 6 tries and no `IsNotFound` or nil error. 

https://github.com/openshift/origin/blob/e07b8e3e77ae250299f9bbb6bcd135ef2f6408d9/pkg/monitortests/network/disruptionserviceloadbalancer/monitortest.go#L290

jluhrsen: yes, that was my intention originally. I've updated it now. look ok?

jluhrsen:
@neisw , good with you now? [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/28761/pull-ci-openshift-origin-master-e2e-azure-ovn-upgrade/1786140439569502208/artifacts/e2e-azure-ovn-upgrade/openshift-e2e-test/build-log.txt) is the test log file which didn't hit any errors in the
Get() to check the namespace, but at least it validates that the changes are not so bad
that they break the tests :)

PR #28766, opened 2024-05-03
Add images coming in 1.30

dinhxuanvu:


dinhxuanvu:
@bertinatto PTAL. Thanks!

soltysh (changes requested):
It looks like you have a few duplicates, but also missing a few images. I've opened a WIP PR with k8s 1.30 in origin in https://github.com/openshift/origin/pull/28778 compare the images from there and you'll want to add the new ones here. 

Be careful to also add the duplicate images, the reason for that is they changed order (there was windows image removed) which means the number after `e2e-<number>` is smaller by one for those images :disappointed: .

PR #28773, opened 2024-05-06
[wip] use the proper cert to verify the oauth-server

deads2k:
I don't plan to finish out or refactor to make it more reasonable, but hopefully this explains better than my previous words.

### DIFF
--- a/test/extended/util/oauthserver/oauthserver.go
+++ b/test/extended/util/oauthserver/oauthserver.go
@@ -222,6 +222,11 @@ func waitForOAuthServerPodReady(oc *exutil.CLI) error {
 }
 
 func waitForOAuthServerRouteReady(oc *exutil.CLI) error {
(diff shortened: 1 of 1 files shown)


test/extended/util/oauthserver/oauthserver.go
```diff
@@ -222,6 +222,11 @@ func waitForOAuthServerPodReady(oc *exutil.CLI) error {
 }
 
 func waitForOAuthServerRouteReady(oc *exutil.CLI) error {
+	routerSecret, err := oc.AdminKubeClient().CoreV1().Secrets("openshift-config-managed").Get(context.Background(), "router-certs", metav1.GetOptions{})
```
stlaz, line 225: you should use the `default-ingress-cert` CM instead.

stlaz (changes requested):

deads2k:
hopefully this example helped. As in description, no intention of taking it further.

PR #28782, opened 2024-05-08
ETCD-488: Fix feature gate for etcd profiles e2e test

dusk125:
I had the wrong feature gate option so this test wasn't getting scraped properly


soltysh (approved):
/lgtm
/approve

PR #28787, opened 2024-05-09
USHIFT-3143: Skip metrics and adjust termination message policy monitor tests

pacevedom:
MicroShift does not support metrics server yet. It is unclear how it will do when it does, so for now skip this test.


dgoodwin:
/lgtm

We need to get  a presubmit for microshift in here soon.

pacevedom:
/hold
Need to investigate a thing about these tests.

dgoodwin:
I'm not totally sure what's happening here but I fear your presubmits are stuck running code that was reverted last week. Those panics should have been resolved, and I don't see them firing anywhere else. We've seen this before where PRs get "stuck" on old code and have to be reopened as a new PR. Honestly I think that's what I'd recommend at this point.

PR #28799, opened 2024-05-14
USHIFT-3143: Skip metrics and adjust termination message policy monitor tests

pacevedom:
/cc @dgoodwin


pacevedom:
Agreed with @dgoodwin over slack.

PR #28808, opened 2024-05-16
NO-JIRA: quota.sh: 4.16 no longer creates legacy API tokens

petr-muller:
The `quota.sh` testcase is making assumptions on how many secrets are automatically created in namespaces, and this number is lower in OCP 4.16.

Context provided by Luis Sanchez:
> As of OCP 4.11, service account API token secrets (now referred to as “legacy”) were no longer generated automatically for each service account.
> All users were supposed to migrate to using the TokenRequest API.
>  Unfortunately, the integrated image registry was also generating a legacy service account API token for its own use, and some users started to accidentally pick up the token intended for the image registry and missed migrating to the TokenRequest API.
> Starting in v4.15, if the image registry is not enabled, the tokens will not be generated. The image registry is still enabled by default, so again another missed opportunity to catch people not using the TokenRequest API.
> Finally, in v4.16 the image registry no longer generates the legacy service account API token at all.

Based on the above, it seems that the testcase should be improved to expect some specific conditions (and e.g. dont allow high numbers for 4.16 clusters) but @soltysh indicated there's an intent to port this testcase to Go, so any logic improvements will be done at that time and for now we can just allow several smaller counts.

### DIFF
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -34263,11 +34263,13 @@ os::test::junit::declare_suite_start "cmd/quota/clusterquota"
 os::cmd::expect_success 'oc new-project quota-foo --as=deads --as-group=system:authenticated --as-group=system:authenticated:oauth'
 os::cmd::expect_success 'oc label namespace/quota-foo owner=deads'
 # before k8s 1.24 this will return 9, starting from 1.24 it'll return 6
-os::cmd::try_until_text 'oc get secrets -o name -n quota-foo | wc -l' '6|9'
+# OCP 4.16+ no longer creates legacy API tokens so it will only return 3
+os::cmd::try_until_text 'oc get secrets -o name -n quota-foo | wc -l' '3|6|9'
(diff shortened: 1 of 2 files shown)


ardaguclu:
It requires make update to update bindata
/lgtm

ardaguclu:
Lastly, I think this https://github.com/openshift/origin/blob/0e559a6adcc4d4e47c82968b2518eef26282f9ed/test/extended/testdata/cmd/test/cmd/quota.sh#L45C190-L45C224 should be `used=(5|6|7|8|9|1[0-9]|20|21|22|23|24|25|26)` or simpler `used=([5-9]|1[0-9]|20|21|22|23|24|25|26)`?

ardaguclu:
@petr-muller it requires an additional `make update`

ardaguclu:
That sounds good to me. Thank you for spending time on this.

/lgtm

soltysh (approved):
/approve

PR #28815, opened 2024-05-20
STOR-1892: Add option to disable all monitors

jsafrane:
`openshift-tests run --disable-monitor all` disables all monitors, so the output and artifacts are not polluted by monitors, and contain the ginkgo test results only.

This is useful when running the tests by a human and not in CI - the monitors are _very_ talkative on stdout. We use openshift-tests as a certification suite for CSI drivers and  3rd party CSI driver vendors are not interested in megabytes of logs of OCP health, they're interested in the CSI driver test results.

In addition, when all monitors are disabled, do not collect `AdditionalEvents__in_cluster_disruption.json` from nodes - no monitoring means no monitoring DaemonSets running and thus there is nothing to collect from them.


dgoodwin:
This is a reasonable line of thought but I fear it's too big of a hammer. The monitor tests contain an immense amount of testing, without which the CSI tests could be giving false positives where the tests pass but the cluster was outlandishly unwell. They check the state of cluster operators, network connectivity, watch for suspicious kube events, watch for suspicious etcd and kubelet logging, etc. 

Particularly given the issue appears to be with how much is logged, it seems like a better first step would be to analyze what's logging so much and see if we could adjust there.

PR #28817, opened 2024-05-21
ETCD-578: remove Tech Preview check from etcd profiles e2e test

dusk125:
The GA of this feature has merged [1] so this check is no longer needed.

[1] https://github.com/openshift/api/pull/1844


soltysh (approved):
/lgtm
/approve

PR #28824, opened 2024-05-23
OCPBUGS-34385: Add exception for openshift-e2e-loki KubePodNotReady

dgoodwin:
This should not be able to fail jobs.


dgoodwin:
/override ci/prow/e2e-gcp-ovn

Not related and we need this ASAP.

PR #28825, opened 2024-05-23
OCPBUGS-34403: Only look for thanos connections to platform monitoring stack

cblecker:
When observing the number of thanos connections and comparing to the number of platform monitoring stack replicas, the thanos connections to the user-workload-monitoring stack are not accounted for. This will cause metric collection to fail if user-workload-monitoring is enabled.

This fixes the query to only focus on the connections specifically to the platform monitoring stack, as the type of data collection the monitoring tests are doing focuses on platform metrics as opposed to ones from the UWM stack.


cblecker:
/cherrypick release-4.16 release-4.15 release-4.14 release-4.13

PR #28832, opened 2024-05-27
API-1783: Conditionally skip tests to help with k8s 1.30

soltysh:
/assign @jsafrane @dgoodwin

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -620,6 +620,8 @@ func (o *GinkgoRunSuiteOptions) filterOutRebaseTests(restConfig *rest.Config, te
 		`[Driver: azure-disk] [Testpattern: Pre-provisioned PV`,
 		`[sig-api-machinery] ValidatingAdmissionPolicy [Privileged:ClusterAdmin] should type check a CRD`,
 		`[sig-node] [Feature:PodLifecycleSleepAction] when create a pod with lifecycle hook using sleep action valid prestop hook using sleep action`,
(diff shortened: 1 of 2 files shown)


dgoodwin:
/hold

Actually looks like this may not hit them all.

pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -615,6 +615,9 @@ func (o *GinkgoRunSuiteOptions) filterOutRebaseTests(restConfig *rest.Config, te
 		// compare https://github.com/kubernetes/kubernetes/pull/123405
 		// which changed the healthz handler name
 		`[sig-api-machinery] health handlers should contain necessary checks`,
+		`PersistentVolumes GCEPD.*`,
```
dgoodwin, line 618: Down below this looks like a strings.Contains, I don't think that will work with the .* wildcard, just need to remove that.

pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -615,6 +615,9 @@ func (o *GinkgoRunSuiteOptions) filterOutRebaseTests(restConfig *rest.Config, te
...
+		`Multi-AZ Cluster Volumes should schedule pods in the same zones as statically provisioned PVs`,
+		`[Driver: azure-disk] [Testpattern: Pre-provisioned PV.*`,
```
dgoodwin, line 620: Same here.

soltysh:
/override ci/prow/lint 
this will be fixed in https://github.com/openshift/origin/pull/28834 with a timeout bump

PR #28834, opened 2024-05-27
Bump golangci-lint timeout

soltysh:
/assign @p0lyn0mial


soltysh:
Picked into https://github.com/openshift/origin/pull/28832

PR #28797, opened 2024-05-13
SDN-4168: Optimize timing for IPsec tests 

pperiyasamy:
The ipsec test suite takes about 5-6 hours to finish running the tests,  This PR optimizes ipsec tests such that it reboots only worker nodes for cert installation and tests execution time can be greatly reduced to 2-3 hrs.

PR #28840, opened 2024-05-29
OCPBUGS-33486: Removes dependency on samples operator images

apoorvajagtap:
OpenShift-Samples operator still points to stale image versions. With the sample operator deprecated with OCP v4.16, these non-existing image references lead to test failures. This commit eliminates the dependency on the samples operator for importing the ruby IS, and instead creates a custom ruby image stream.

### DIFF
--- a/test/extended/testdata/builds/test-context-build.json
+++ b/test/extended/testdata/builds/test-context-build.json
@@ -42,7 +42,7 @@
           "git": {
             "uri":"https://github.com/sclorg/s2i-ruby-container"
           },
-          "contextDir": "3.0/test/puma-test-app"
+          "contextDir": "3.3/test/puma-test-app"
         },
         "strategy": {
           "type": "Source",
@@ -54,8 +54,8 @@
               }
             ],
             "from": {
-              "kind": "DockerImage",
-              "name": "image-registry.openshift-image-registry.svc:5000/openshift/ruby:3.0-ubi8"
+              "kind": "ImageStreamTag",
+              "name": "ruby:latest"
             }
           }
         },
--- a/test/extended/util/framework.go
+++ b/test/extended/util/framework.go
@@ -238,10 +238,81 @@ func processScanError(log string) error {
 	return fmt.Errorf(log)
 }
 
+// getImageStreamObj returns the updated spec for imageStream object
+func getImageStreamObj(imageStreamName, imageRef string) *imagev1.ImageStream {
+	imageStream := &imagev1.ImageStream{
+		ObjectMeta: metav1.ObjectMeta{Name: imageStreamName},
+		Spec: imagev1.ImageStreamSpec{
+			LookupPolicy:          imagev1.ImageLookupPolicy{Local: true},
+			DockerImageRepository: imageRef,
+			Tags: []imagev1.TagReference{{
+				Name: "latest",
+				ImportPolicy: imagev1.TagImportPolicy{
+					ImportMode: imagev1.ImportModePreserveOriginal,
+				},
+			}},
+		},
+		Status: imagev1.ImageStreamStatus{
+			DockerImageRepository: imageRef,
+			Tags: []imagev1.NamedTagEventList{{
+				Tag: "latest",
+			}},
+		},
+	}
+	return imageStream
+}
+
+// WaitForImageStreamImport creates & waits for custom ruby imageStream to be available in current namespace
+// TODO: To eliminate the dependency on OpenShift Samples Operator in future,
+// WaitForImageStreamImport should be a replacement of WaitForOpenShiftNamespaceImageStreams func
+func WaitForImageStreamImport(oc *CLI) error {
+	ctx := context.Background()
+	var registryHostname string
+
+	// TODO: Reference an image from registry.redhat.io
+	images := map[string]string{
+		"ruby": "registry.access.redhat.com/ubi8/ruby-33",
+	}
+
+	// Check to see if we have ImageRegistry enabled
+	hasImageRegistry, err := IsCapabilityEnabled(oc, configv1.ClusterVersionCapabilityImageRegistry)
+	if err != nil {
+		return err
+	}
+	if hasImageRegistry {
+		registryHostname, err = WaitForInternalRegistryHostname(oc)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Create custom imageStream using `oc import-image`
+	e2e.Logf("waiting for imagestreams to be imported")
+	for imageStreamName, imageRef := range images {
+		err := CustomImageStream(oc, getImageStreamObj(imageStreamName, imageRef))
+		if err != nil {
+			e2e.Logf("failed while creating custom imageStream")
+			return err
+		}
+
+		// Wait for imageRegistry to be ready
+		pollErr := wait.PollUntilContextTimeout(ctx, 10*time.Second, 150*time.Second, false, func(context.Context) (bool, error) {
+			return checkNamespaceImageStreamImported(ctx, oc, imageStreamName, registryHostname, oc.Namespace())
+		})
+		// pollErr will be not nil if there was an immediate error, or we timed out.
+		if pollErr == nil {
+			return nil
+		}
+		DumpImageStream(oc, oc.Namespace(), imageStreamName)
+		return pollErr
+	}
+	return nil
+}
+
 // WaitForOpenShiftNamespaceImageStreams waits for the standard set of imagestreams to be imported
 func WaitForOpenShiftNamespaceImageStreams(oc *CLI) error {
 	ctx := context.Background()
-	images := []string{"ruby", "nodejs", "perl", "php", "python", "mysql", "postgresql", "jenkins"}
+	images := []string{"nodejs", "perl", "php", "python", "mysql", "postgresql", "jenkins"}
 
 	hasSamplesOperator, err := IsCapabilityEnabled(oc, configv1.ClusterVersionCapabilityOpenShiftSamples)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


test/extended/util/framework.go
```diff
@@ -238,10 +238,55 @@ func processScanError(log string) error {
...
+		// Wait for imageRegistry to be ready
+		pollErr := wait.PollUntilContextTimeout(ctx, 10*time.Second, 150*time.Second, false, func(context.Context) (bool, error) {
+			return checkOpenShiftNamespaceImageStreamImported(ctx, oc, image, registryHostname)
+		})
+		// pollErr will be not nil if there was an immediate error, or we timed out.
+		if pollErr != nil {
+			return pollErr
+		}
+		DumpImageStream(oc, "openshift", image)
+	}
```
apoorvajagtap, line 281: @adambkaplan here, I have returned the pollErr immediately w/o calling DumpImageStream(), unlike what we have in [WaitForSamplesImagestream](https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L306-L309). Should this be fine, or there's any reason we want to call DumpImageStream() & then return the pollErr?

test/extended/util/framework.go
```diff
@@ -238,10 +238,56 @@ func processScanError(log string) error {
...
+		registryHostname, err = WaitForInternalRegistryHostname(oc)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Create custom imageStream using `oc import-image`
+	e2e.Logf("waiting for imagestreams to be imported")
+	for image, location := range images {
+		err := oc.Run("import-image").Args(image, fmt.Sprintf("--from=%s", location), "--confirm").Execute()
```
prb112, line 267: please add --import-mode=PreserveOriginal
it'll cause issues on arm64

```

    --import-mode='Legacy':
        Imports the full manifest list of a tag when set to 'PreserveOriginal'. Defaults to
        'Legacy'.
```

test/extended/util/framework.go
```diff
@@ -238,10 +238,56 @@ func processScanError(log string) error {
...
+		registryHostname, err = WaitForInternalRegistryHostname(oc)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Create custom imageStream using `oc import-image`
+	e2e.Logf("waiting for imagestreams to be imported")
+	for image, location := range images {
+		err := oc.Run("import-image").Args(image, fmt.Sprintf("--from=%s", location), "--import-mode=PreserveOriginal", "--confirm").Execute()
```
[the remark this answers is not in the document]
apoorvajagtap: Thanks for highlighting that! :)

test/extended/util/framework.go
prb112: I do think your fix was resolved. You just hit a few flakes on gcp. Most of those are pretty common.

prb112: Any intent to add `--import-mode=PreserveOriginal` back?

test/extended/util/framework.go
```diff
@@ -238,10 +238,56 @@ func processScanError(log string) error {
 	return fmt.Errorf(log)
 }
 
+// WaitForImageStreams creates custom ruby imageStream in OpenShift namespace
+// TODO: To eliminate the dependency on OpenShift Samples Operator in future,
+// WaitForImageStreams should be a replacement of WaitForOpenShiftNamespaceImageStreams func
+func WaitForImageStreams(oc *CLI) error {
```
adambkaplan, line 244: We can do better here and add a utility function that wraps `oc import-image` generically.

```suggestion
// ImportImageStream imports the provided image reference to an imagestream with the given name in the
// oc client's current namespace.
func ImportImageStream(oc *CLI, imageRef string, imageStreamName string) error {
```

test/extended/util/framework.go
adambkaplan, line 267: Won't this create an imagestream in oc's current namespace? If that is the case, the check for the openshift namespace imagestream will fail once we turn off the samples operator.

test/extended/util/framework.go
```diff
@@ -238,10 +238,56 @@ func processScanError(log string) error {
...
+
+		// Wait for imageRegistry to be ready
+		pollErr := wait.PollUntilContextTimeout(ctx, 10*time.Second, 150*time.Second, false, func(context.Context) (bool, error) {
+			return checkOpenShiftNamespaceImageStreamImported(ctx, oc, image, registryHostname)
+		})
+		// pollErr will be not nil if there was an immediate error, or we timed out.
+		if pollErr == nil {
+			return nil
+		}
+		DumpImageStream(oc, "openshift", image)
```
adambkaplan, line 280: Ditto from above - shouldn't this dump the imagestream in oc's current namespace?

test/extended/testdata/builds/test-context-build.json
```diff
@@ -42,7 +42,7 @@
           "git": {
             "uri":"https://github.com/sclorg/s2i-ruby-container"
           },
-          "contextDir": "3.0/test/puma-test-app"
+          "contextDir": "3.3/test/puma-test-app"
```
adambkaplan, line 45: You should also change the "from" image to reference the ImageStream being created in the test. See [doc](https://docs.openshift.com/container-platform/4.15/cicd/builds/build-strategies.html#builds-strategy-s2i-build_build-strategies-docker)

adambkaplan (changes requested):
I don't think this completely removes the samples operator dependency - see my comments.

test/extended/util/framework.go
```diff
@@ -320,6 +367,18 @@ func WaitForSamplesImagestream(ctx context.Context, oc *CLI, imagestream string,
...
 }
 
+// ImportImageStream imports the provided image reference to an imagestream with
+// the given name in the oc client's current namespace.
+func ImportImageStreams(oc *CLI, imageStreamName, imageRef string) error {
+	err := oc.Run("import-image").Args(
+		imageStreamName,
+		fmt.Sprintf("--from=%s", imageRef),
+		"--import-mode=PreserveOriginal",
+		"--confirm").Execute()
```
adambkaplan, line 377: From prior feedback I received - invoking the `oc` binary in this manner is discouraged for running OpenShift tests moving forward. When possible (as is the case here), the preference is to use Kubernetes clients to create/update the respective k8s objects.

Example (pseduo-code)

```go
is, err := oc.ImageClient().ImageV1().Imagestreams(namespace).Create(ctx, isObj, metav1.CreateOptions{})
```

For ImageStreams, you will want to use the openshift Image api clients to create an ImageStream object. See https://github.com/openshift/api/blob/master/image/v1/types.go#L233 for more info.

adambkaplan (changes requested):

test/extended/util/framework.go
```diff
@@ -320,6 +391,12 @@ func WaitForSamplesImagestream(ctx context.Context, oc *CLI, imagestream string,
 	return pollErr
 }
 
+// CustomImageStream uses the provided imageStreamObj reference to create an imagestream with the given name in the given namespace.
+func CustomImageStreams(oc *CLI, imageStream *imagev1.ImageStream) error {
```
adambkaplan, line 395: nit: singular instead of plural

```suggestion
func CustomImageStream(oc *CLI, imageStream *imagev1.ImageStream) error {
```

test/extended/util/framework.go
```diff
@@ -300,7 +371,7 @@ func WaitForSamplesImagestream(ctx context.Context, oc *CLI, imagestream string,
 				return false, nil
 			}
 		}
-		return checkOpenShiftNamespaceImageStreamImported(ctx, oc, imagestream, registryHostname)
+		return checkNamespaceImageStreamImported(ctx, oc, imagestream, registryHostname, "openshift")
```
adambkaplan, line 374: Yes! :+1:

test/extended/util/framework.go
```diff
@@ -238,10 +238,81 @@ func processScanError(log string) error {
...
+	for imageStreamName, imageRef := range images {
+		err := CustomImageStreams(oc, getImageStreamObj(imageStreamName, imageRef))
+		if err != nil {
+			e2e.Logf("failed while creating custom imageStream")
+			return err
+		}
+
+		// Wait for imageRegistry to be ready
+		pollErr := wait.PollUntilContextTimeout(ctx, 10*time.Second, 150*time.Second, false, func(context.Context) (bool, error) {
+			return checkNamespaceImageStreamImported(ctx, oc, imageStreamName, registryHostname, oc.Namespace())
```
adambkaplan, line 300: Yes! :+1:

test/extended/util/framework.go
```diff
@@ -238,10 +238,81 @@ func processScanError(log string) error {
...
+			}},
+		},
+	}
+	return imageStream
+}
+
+// WaitForImageStreams creates & waits for custom ruby imageStream to be available in current namespace
+// TODO: To eliminate the dependency on OpenShift Samples Operator in future,
+// WaitForImageStreams should be a replacement of WaitForOpenShiftNamespaceImageStreams func
+func WaitForImageStreams(oc *CLI) error {
```
adambkaplan, line 268: nit (won't block on this). We should consider adding the imagestream name and image ref as parameter. Don't need to do this now since this is refactoring previous code, but is worth considering as the Samples Operator becomes "off by default."

Ex: 

```suggestion
func WaitForImageStreamImport(oc *CLI, imageStreamName string, imageRef string) error {
```

adambkaplan (approved):
/approve

One final nit on a function name/godoc mismatch, otherwise looks good.

adambkaplan:
Outstanding test failures appear unrelated - build test suite passes.

test/extended/util/framework.go
apoorvajagtap: How about introducing this change while we implement the same logic for rest of the images too? At present [WaitForOpenShiftNamespaceImageStreams](https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L244) is still being used to import other imagestreams (as we have focused just on ruby atm).

apoorvajagtap:
The `e2e-aws-ovn-ipsec-serial` test failures might be fixed with https://github.com/openshift/origin/pull/28797

ardaguclu:
CI results are promising, why we don't merge this PR still?

adambkaplan (approved):
/lgtm

Looks good - noted a few opportunities for the future as we plan for the Samples Operator to be deprecated/off by default.

prb112:
Hey @adambkaplan @apoorvajagtap do you know when the 4.15/4.14 cherry-picks will land? Thanks, Paul

> /cherrypick release-4.16 release-4.15 release-4.14

PR #28844, opened 2024-05-30
WIP: feat: add leases to resource watch

eggfoobar:
adding leases to resource watch help determine if we are releasing leader election lock, this is very useful for SNO deployments where we want to make sure pods are releasing their lock when they are terminating


eggfoobar:
No longer needed

PR #28856, opened 2024-06-04
trt-1689: Filter must-gather and truncate system node users

neisw:


pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
```diff
@@ -453,12 +457,33 @@ func URIToParts(uri string) (string, schema.GroupVersionResource, string, string
 func isMonitoredUser(user string) bool {
 	for _, userPrefix := range monitoredUsers {
 		if strings.HasPrefix(user, userPrefix) {
+			// see if we have any filters defined for this prefix
+			if filters, ok := filteredUsers[userPrefix]; ok {
```
deads2k, line 461: Can this just be an exclude pattern that takes precedence?

pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
```diff
@@ -453,12 +457,33 @@ func URIToParts(uri string) (string, schema.GroupVersionResource, string, string
...
+				}
+			}
+
 			return true
 		}
 	}
 	return false
 }
 
+func cleanupUser(user string) string {
```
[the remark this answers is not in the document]
neisw: Yes, the import is generic no logic to change data coming in.  We could preserve the original value in a new column and just not use it in the query if that works for you.

pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
```diff
@@ -468,8 +490,14 @@ func writeAuditLogDL(artifactDir, timeSuffix string, auditLogSummary *AuditLogSu
 		for rk, rv := range uv.perResourceRequestCount {
 			for vk, vv := range rv.perVerbRequestCount {
 				for sk, sv := range vv.perHTTPStatusRequestCount {
-					rows = append(rows, map[string]string{"User": uv.user, "Resource": rk.Resource, "Verb": vk, "HttpStatus": strconv.FormatInt(int64(sk), 10), "RequestCount": strconv.FormatInt(int64(sv), 10)})
+					user, unmodifiedUser := cleanupUser(uv.user)
+					data := map[string]string{"User": user, "Resource": rk.Resource, "Verb": vk, "HttpStatus": strconv.FormatInt(int64(sk), 10), "RequestCount": strconv.FormatInt(int64(sv), 10)}
+					if len(unmodifiedUser) > 0 {
+						data["unmodified_user"] = unmodifiedUser
```
neisw, line 496: If we have modified the user the unmodified value will be added to the outupt data.  It isn't added to the schema so it will not be loaded in the bigquery table but we can do that if we want.

neisw: Produces
```    
{
            "HttpStatus": "200",
            "RequestCount": "2",
            "Resource": "runtimeclasses",
            "User": "system:node-0",
            "UserUnmodified": "system:node:ci-op-6cqn1jmf-03fd1-5qt7q-worker-a-srqxc",
            "Verb": "list"
 },
```

pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go
```diff
@@ -468,8 +490,14 @@ func writeAuditLogDL(artifactDir, timeSuffix string, auditLogSummary *AuditLogSu
 		for rk, rv := range uv.perResourceRequestCount {
 			for vk, vv := range rv.perVerbRequestCount {
 				for sk, sv := range vv.perHTTPStatusRequestCount {
-					rows = append(rows, map[string]string{"User": uv.user, "Resource": rk.Resource, "Verb": vk, "HttpStatus": strconv.FormatInt(int64(sk), 10), "RequestCount": strconv.FormatInt(int64(sv), 10)})
+					user, unmodifiedUser := cleanupUser(uv.user)
```
deads2k, line 493: does this need to have a single record for all of the same name?

neisw: I [map](https://github.com/openshift/origin/blob/8bced75d5a89bb5836fd804b891a50a61fcc9c94/pkg/monitortests/kubeapiserver/auditloganalyzer/audit_log_summarizer.go#L477) the unique user value to a standarized prefix-count value but maintain the records separately so we don't have to sum them all up and use only a single record.  Could if we needed too but seemed more straightforward to keep them separate.

neisw:
permafailing across multiple prs

neisw:
/override ci/prow/e2e-gcp-ovn
job passed previously on the same commit.

PR #28857, opened 2024-06-05
[release-4.16] ETCD-578: Fix etcd profiles e2e test to check returned…

dusk125:
… status for updated value


dusk125:
need to wait for master to go in first

PR #28866, opened 2024-06-10
ETCD-606: Open up the RequiredInstallerResourcesMissing test to more resources.

dgoodwin:
This was limited to secrets, there are others including a new bundle in
https://github.com/openshift/cluster-etcd-operator/pull/1271. Open the
regex up to match more, and this has it's own test which only adds a
flake threshold, but keeps the 20 fail threshold.

### DIFF
--- a/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
+++ b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
@@ -575,7 +575,6 @@ var AllowBackOffRestartingFailedContainer = &SimplePathologicalEventMatcher{
 var EtcdRequiredResourcesMissing = &SimplePathologicalEventMatcher{
 	name:               "EtcdRequiredResourcesMissing",
 	messageReasonRegex: regexp.MustCompile(`^RequiredInstallerResourcesMissing$`),
-	messageHumanRegex:  regexp.MustCompile(`secrets: etcd-all-certs-[0-9]+`),
 }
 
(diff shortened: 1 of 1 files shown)


pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -575,7 +575,6 @@ var AllowBackOffRestartingFailedContainer = &SimplePathologicalEventMatcher{
 var EtcdRequiredResourcesMissing = &SimplePathologicalEventMatcher{
 	name:               "EtcdRequiredResourcesMissing",
 	messageReasonRegex: regexp.MustCompile(`^RequiredInstallerResourcesMissing$`),
```
tjungblu, line 577: I think we can remove all etcd exceptions now, because I've changed the event sink globally for the whole operator

tjungblu: at least we can try, guess we can run another payload test to verify

dgoodwin: This actually doesn't appear to be an exception that allowed more events, it looks like it was just to isolate them to a specific test so it would be easy to trace down how often it's happening. Once we merge this we can look CI fleet wide to find runs where it would break 20, or where it got close. (the flaked tests are over 10)

tjungblu:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.17-upgrade-from-stable-4.16-e2e-aws-ovn-upgrade 5

PR #28851, opened 2024-06-03
Revert "TRT-1691: Revert #28735 "TRT-1576: Fail if operator has Available=False unless in upgrade window""

DennisPeriquet:
Reverts openshift/origin#28847

For this test: [bz-%v] clusteroperator/%v should not change condition/Available]:

* For non-upgrade jobs, fail when operator goes to Available=False
* For upgrade-jobs, fail when operator goes to Available=False unless it's during an upgrade window and the condition lasts for less than 10 minutes.

Exceptions:

* authentication operator is affected by etcd performance issues
* image-registry is affected by other tests that cause all image-registry replicas to go down

For the above exceptions, their tests still run but they will flake as they did before this PR.

Testing tasks:

- [ ] Test with azure since this is where the authentication operator has problems due to etcd performance issues; the clusteroperator/authentication flakes 8, pass 2
  * see [/payload for azure](https://pr-payload-tests.ci.openshift.org/runs/ci/d8511d30-281a-11ef-9083-d014afac8f40-0)
- [ ] Test with gcp techpreview serial to ensure image-registry operator test flakes as it did before (now that we give it an exception)
  * see [/payload for techpreview jobs](https://pr-payload-tests.ci.openshift.org/runs/ci/0d6e9310-27e1-11ef-8f24-4a8b14294335-0); 5 of 10 jobs finished, those 5 passed on clusteroperator/image-registry test.
- [ ] Run /payload 4.17 nightly blocking
  * see [/payload for 4.17 blocking](https://pr-payload-tests.ci.openshift.org/runs/ci/6c1e6040-2866-11ef-9091-677516e6db09-0)
  * still trying to run the [metal job](https://pr-payload-tests.ci.openshift.org/runs/ci/189a17a0-28b8-11ef-8adf-49cc0ed6dd05-0) that failed twice earlier due to `Could not acquire CI resource: No available resource found`
  * the metal job finally [ran](https://pr-payload-tests.ci.openshift.org/runs/ci/3d5e7420-291f-11ef-9445-888b9e3f685e-0)

PR #28872, opened 2024-06-11
OCPEDGE-1099: feat: add sno consideration for operator availability

eggfoobar:
in single node upgrade runs we expect operatros to transition their Availability to false during the upgrade, adding a check for sno to the exception logic for those situations.


eggfoobar:
/close 

Closing in favor of #28851

PR #28873, opened 2024-06-12
WIP: Use ephemeral cluster docker config.json for test binary extraction

jupierce:
The service account running the e2e tests does not necessarily have pull access to the component images in the release payload. The cluster under test, however, must have a valid pull secret. Utilize the ephemeral cluster's pull secret to extract binaries from the payload under test.

### DIFF
--- a/pkg/test/ginkgo/external.go
+++ b/pkg/test/ginkgo/external.go
@@ -83,7 +83,7 @@ func extractBinaryFromReleaseImage(tag, binary string) (string, error) {
 		return "", fmt.Errorf("cannot determine release image from ClusterVersion resource")
 	}
 
-	if err := runImageExtract(releaseImage, "/release-manifests/image-references", tmpDir); err != nil {
+	if err := runImageExtract(releaseImage, "/release-manifests/image-references", tmpDir, ""); err != nil {
 		return "", fmt.Errorf("failed extracting image-references: %w", err)
 	}
 	jsonFile, err := os.Open(filepath.Join(tmpDir, "image-references"))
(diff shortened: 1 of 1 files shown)


pkg/test/ginkgo/external.go
```diff
@@ -110,10 +110,26 @@ func extractBinaryFromReleaseImage(tag, binary string) (string, error) {
...
 	}
+
+	// The preceding runImageExtract was against a release payload that was created in the local
+	// ci-operator namespace. Our process was free to access it. The release payload, however,
+	// may be referencing images in other registries that we don't have access to (e.g. quay.io,
+	// registry.ci.openshift.org). One location that does have access is in the cluster under test.
+	// It cluster-wide pull-secret must have pull access to these images in order for it to have
+	// installed. Read it's dockerconfigjson value and use it when extract external test binaries
+	// from images referenced by the release payload.
+	clusterPullSecret, err := oc.AdminKubeClient().CoreV1().Secrets("openshift-config").Get(context.Background(), "pull-secret", metav1.GetOptions{})
```
soltysh, line 121: You're swallowing this error, we need to check this bit. I'd propose something like:
```golang
dockerConfigJsonPath := ""
if err == nil {
    clusterDockerConfig := clusterPullSecret.Data[".dockerconfigjson"]
	dockerConfigJsonPath = filepath.Join(tmpDir, ".dockerconfigjson")
	err = os.WriteFile(dockerConfigJsonPath, clusterDockerConfig, 0644)
	if err != nil {
		return "", fmt.Errorf("unable to serialize ephemeral cluster pull secret locally: %v", err)
	}
}
```
This would only try to use that `.dockerconfigjson` only if one existed, and if not we'd try w/o, and fallback to the old path.

soltysh (commented):
I'm still testing the actual test list, but based on the preliminary results I got in https://github.com/openshift/origin/pull/28900 it's looking ok. For now we definitely need that missing error check as pointed out above.

pkg/test/ginkgo/external.go
jupierce: Good catch on not swallowing the error. I've fixed that. When considering a fallback, I'd prefer a hard failure for now (which I've added) that can be easily isolated by reviewing CI test results. There are three reasons it could fail:
1. resource does not exist -- any OpenShift install should have this secret
2. connectivity issues
3. other
In each case, a failure in the approach would be news I'd like to understand without potentially hiding with a fallback. Ultimately, we want external binaries to be the canonical way of testing, so our understanding of the approach needs to be comprehensive -- and I think hard failures will help with that.

soltysh (approved):
/lgtm
/approve

PR #28874, opened 2024-06-12
USHIFT-3455: Skip legacy-etcd monitor tests in microshift

pacevedom:


pkg/monitortests/etcd/legacyetcdmonitortests/monitortest.go
```diff
@@ -24,6 +27,23 @@ func NewLegacyTests() monitortestframework.MonitorTest {
...
+	if isMicroShift {
+		w.notSupportedReason = &monitortestframework.NotSupportedError{
+			Reason: "platform MicroShift not supported",
+		}
+		return w.notSupportedReason
+	}
+
 	jobType, err := platformidentification.GetJobType(ctx, adminRESTConfig)
 	if err != nil {
 		return fmt.Errorf("unable to determine job type: %v", err)
```
dgoodwin, line 49: Looking at how w.jobType is actually used, it's just to temporarily adjust a threshold for vsphere which is struggling. This seems too thin a reason to permanently kill off several tests which help identify that etcd is struggling. 

I'd suggest a safer route would be to modify this to not return an error here, set a default of an empty JobType{} with an empty string Platform.

pacevedom: So that would certainly do, but I wonder if these tests are meaningful in MicroShift. We do not run etcd in a pod, and we have no events because of that (which is what all of these tests check).
Does it make sense to use a fake empty platform (or legit, depends on how you look at it!) only to know that any etcd test afterwards will check on non-existing events?

dgoodwin: Looks like about 5 tests are in here, 3 involve pod logs, so it sounds like those are irrelvant, the other is install related, which I'm guessing is not relevant to you either then, and lastly the clusteroperator state, do you have an etcd cluster operator in microshift? If not, then I would just add a comment along side your disable explaning why and we can merge this as is.

pacevedom: Done

PR #28884, opened 2024-06-14
NO-ISSUE: fix the node reboot test to actually run

deads2k:
/assign @rphillips 
/cc @jupierce

### DIFF
--- a/test/extended/machines/cluster.go
+++ b/test/extended/machines/cluster.go
@@ -1,10 +1,13 @@
 package operators
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"fmt"
+	"sort"
 	"strings"
+	"time"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
@@ -21,7 +24,6 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework/pod"
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	psapi "k8s.io/pod-security-admission/api"
-	"k8s.io/utils/strings/slices"
 )
 
 var _ = g.Describe("[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should", func() {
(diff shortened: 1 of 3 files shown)


test/extended/machines/cluster.go
```diff
@@ -120,83 +122,169 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
-		expected := make(map[string]int)
-		actual := make(map[string]int)
+		bootTimelinesByNode := make(map[string][]bootTimelineEntry)
 		errs := make([]error, 0)
 
+		allNodeLogs := &bytes.Buffer{}
 		// List nodes, set actual and expected number of reboots
 		for _, node := range nodes.Items {
-			// Examine only the nodes which are available at the start of the test
-			if !slices.Contains(staticNodeNames, node.Name) {
```
deads2k, line 130: this check resulted in zero nodes, because `openshift-tests` starts a new process to run an individual test in, so the tests must be self-contained.  The `staticNodeNames` had no values, so no logs were checked, so the test always passed without checking anything

test/extended/machines/cluster.go
```diff
@@ -120,83 +122,169 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
-				return a, fmt.Errorf("errors found: %v", errs)
+		for nodeName, timelineEvents := range bootTimelinesByNode {
+			// every reboot (except maybe the first), should have a rebootRequest before it.
+			// we reversed the sort so we can step backwards in time like this
+			for i, timelineEvent := range timelineEvents {
+				// boots should be the events, reboots should be the odds
+				expectedReboot := (i % 2) == 1
+				expectedBoot := !expectedReboot
+				isActualBoot := timelineEvent.action == "Boot"
+				isActualReboot := timelineEvent.action == "RebootRequest"
```
deads2k, line 153: this is the insight I'm using for the test.  Every other entry should be a rebootrequest.  If it's not, the test should fail.

test/extended/machines/cluster.go
```diff
@@ -120,83 +122,169 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
...
+func getNumberOfBootsForNode(kubeClient kubernetes.Interface, namespaceName, nodeName string) ([]bootTimelineEntry, []bootTimelineEntry, string, error) {
 	ctx := context.Background()
 
+	nodeLogs := &bytes.Buffer{}
 	desiredPod := displayRebootsPod.DeepCopy()
 	desiredPod.Namespace = namespaceName
 	desiredPod.Spec.NodeName = nodeName
 
-	errs := []error{}
-	for i := 0; i < 10; i++ {
```
deads2k, line 161: repeatedly running should be necessary.

test/extended/machines/display_reboots_pod.yaml
```diff
@@ -1,13 +1,17 @@
...
 metadata:
+  annotations:
+    openshift.io/required-scc: privileged
   generateName: list-boots-
 spec:
   restartPolicy: Never
   hostPID: true
   containers:
-    - command:
-        - exec chroot /host-root journalctl --list-boots
```
deads2k, line 10: this kept resulting a "command not found" style error.  I didn't bother trying to work out why. By executing bash instead, we get the desired result.

test/extended/machines/cluster.go
```diff
@@ -120,83 +122,170 @@ var _ = g.Describe("[sig-node] Managed cluster", func() {
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(nodes.Items).NotTo(o.HaveLen(0))
 
-		expected := make(map[string]int)
-		actual := make(map[string]int)
+		bootTimelinesByNode := make(map[string][]bootTimelineEntry)
 		errs := make([]error, 0)
 
+		e2e.Logf("###### HERER!!!!!")
```
rphillips, line 128: debug line. should be removed

rphillips:
`e2e-aws-ovn-cgroupsv2 ` ran the test suite and passed.

test/extended/machines/cluster.go
deads2k: > debug line. should be removed

Fixed.

rphillips:
e2e-gcp-ovn passed

rphillips:
/retitle NO-ISSUE: fix the node reboot test to actually run

rphillips:
This PR fixes a number of issues with the `should verify that nodes have no unexpected reboots [Late]` test. LGTM, but I'm going to hold off tagging until Monday :).

PR #28768, opened 2024-05-05
MON-3775: [WIP] add tests for CMO collection profiles

rexagod:
Refer: openshift/enhancements#1298

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

PR #28889, opened 2024-06-18
MON-3775: add tests for the `MetricsCollectionProfiles` feature-gate

rexagod:
OEP: openshift/enhancements#1298

/cc @simonpasquier


rexagod:
@deads2k Following up on https://github.com/openshift/origin/pull/28768#discussion_r1636819912, I dropped `g.Context` but `/verify` still seems to fail. I believe the root cause is the discrepancy between the targets, `update` and `verify`.

I'm able to include my tests in the generated manifests using `update`, but doing so makes `verify` fail. Similarly, if I leave them out, and don't push the generated manifests that include the tests, `verify` passed but the tests don't run (as there's nothing to run).

rexagod:
@deads2k Following up on https://github.com/openshift/origin/pull/28768#discussion_r1636819912, I dropped `g.Context` but `/verify` still seems to fail. I believe the root cause is the discrepancy between the targets, `update` and `verify`.

I'm able to include my tests in the generated manifests using `update`, but doing so makes `verify` fail. Similarly, if I leave them out, and don't push the generated manifests that include the tests, `verify` passes but the tests don't run (as there's nothing to run).

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,257 @@
+package prometheus
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	pv1 "github.com/prometheus/client_golang/api/prometheus/v1"
```
simonpasquier, line 9:
```suggestion
	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
```

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,257 @@
...
+	"gopkg.in/yaml.v3"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
```
simonpasquier, line 15:
```suggestion
```

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,257 @@
...
+		r.kclient, err = kubernetes.NewForConfig(oc.AdminConfig())
+		if err != nil {
+			g.Fail(fmt.Sprintf("failed to create kubernetes client: %v", err))
+		}
+		r.mclient, err = pov1client.NewForConfig(oc.AdminConfig())
+		if err != nil {
+			g.Fail(fmt.Sprintf("failed to create monitoring client: %v", err))
+		}
+		r.pclient = oc.NewPrometheusClient(tctx)
+		operatorConfiguration, err := r.kclient.CoreV1().ConfigMaps(operatorNamespaceName).Get(tctx, operatorConfigurationName, metav1.GetOptions{})
```
simonpasquier, line 77: it should probably retry to avoid failing on API hiccups.

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,257 @@
...
+				}
+
+				return nil
+			}).Should(o.BeNil())
+
+		}
+	})
+})
+
+func (r *runner) fetchMonitorsFor(selectors ...[2]string) (*pov1api.ServiceMonitorList, error) {
```
simonpasquier, line 189: the function signature isn't really obvious. Also it doesn't seem that the first item varies (e.g. it's always `collectionProfileFeatureLabel`).

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,257 @@
...
+				}
+				if len(queryResponse.Data.Result) == 0 {
+					return fmt.Errorf("no result found for profile %q", profile)
+				}
+
+				return nil
+			}).Should(o.BeNil())
+		}
+	})
+	g.It("should have at-least one implementation for each collection profile", func() {
```
simonpasquier, line 123:
```suggestion
	g.It("should have at least one implementation for each collection profile", func() {
```

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,257 @@
...
+				if len(monitors.Items) == 0 {
+					return fmt.Errorf("no monitors found with collection profile %q", profile)
+				}
+
+				return nil
+			}).Should(o.BeNil())
+		}
+	})
+	g.It("should hide a default-only metric when minimal collection profile is enabled", func() {
+		defaultOnlyMetric := "kube_deployment_status_replicas"
```
simonpasquier, line 142: can we pick a Prometheus metric which we know for sure will (likely) never be present in the minimal profile (for instance `prometheus_engine_query_log_enabled`).

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+
+			return nil
+		}).Should(o.BeNil())
+	})
+	g.It("should expose information about the applied collection profile using meta-metrics", func() {
+		vectorExpression := "profile:cluster_monitoring_operator_collection_profile:max{profile=\"%s\"} == 1"
+
+		for _, profile := range collectionProfilesSupportedList {
+			err := r.makeCollectionProfileConfigurationFor(tctx, profile)
```
simonpasquier, line 107: because reconciling the stack takes time on a collection profile change, should we rather change the profile and test all the expectations for this profile in one go?

rexagod: Some cases depend on ordered profile transitions, IIUC this may interfere with changing the design from the current one:
```
<caseA>
	<profileA/B>
	<profileB/A>
<caseB>
	<profileA>
	<profileB/A>
```
to the proposed one,
```
<profileA>
	<caseA>
	<caseB>
<profileB>
	<caseA>
	<caseB>
```
LMK if I missed something.

rexagod: (cc @simonpasquier for a clarification here)

jan--f: I'm no ginko expert, but wouldn't we best represent the profile transistions as ginko containers? So we end up with a structure similar to this:
```
Describe CP tests
  BeforeAll setup make sure default profile is set
  Describe subjects for default profile
    ...
  Describe transistion to minimal
    BeforeEach check or setup minimal
      ...
```

rexagod: I've made two changes to the latest patch (although I still need to take a look at the CI runs to make sure the benchmark results are as expected). These are:

- Based on Jan's review above, encapsulate all cases that start under the default profile in their own context, and same for the minimal profile. Also, ensure this by using `ginkgo.Context`s.
- Drop the bottleneck approach for the 4th test-case (marking a series for deletion, cleaning the tombstones for it, and then polling for change) in favor of checking if the difference between `scrape_samples_scraped` and `scrape_samples_post_metric_relabeling` equal to the number of relabelings mentioned in the `ServiceMonitor` for a particular profile.

pkg/clioptions/suiteselection/feature_filter.go
```diff
@@ -66,11 +67,14 @@ func (f *featureGateFilter) includeTest(name string) bool {
 		featureGates = append(featureGates, featureGate)
 	}
 
-	if f.disabled.HasAny(featureGates...) {
-		return false
+	shouldIncludeTest := !f.disabled.HasAny(featureGates...) && f.enabled.HasAll(featureGates...)
```
jan--f, line 70: Any functional reason to change this? The logging can be added to the existing version no?

I prefer the existing code for clarity, no need to solve a logic statement in my head. Consider this a nit though.

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,323 @@
...
+					}
+
+					return nil
+				}).Should(o.BeNil())
+			}
+		})
+		g.It("should hide all default-only metrics when minimal collection profile is enabled", func() {
+			appNameSelector := "app.kubernetes.io/name"
+			appName := "kube-state-metrics"
+			for _, profile := range []string{collectionProfileFull, collectionProfileMinimal} {
```
jan--f, line 148: Iiuc this is the test case to test applying the minimal profile (after full was applied). Why does this iterate over all profiles, including full?

pkg/clioptions/suiteselection/feature_filter.go
rexagod: I included this after a discussion around [an issue that still persists](https://github.com/openshift/origin/pull/28889#issuecomment-2175898244) for this patch (but not caused by it): [#forum-api-review](https://redhat-internal.slack.com/archives/CE4L0F143/p1718119344698569?thread_ts=1718116330.342849&cid=CE4L0F143). This is purely a cosmetic change and will most likely be dropped once addressed.

rexagod: ACK, on that note I believe the following should be more readable while preventing `if` clause repetitions. Although, it is likely we will drop this once [the issue](https://redhat-internal.slack.com/archives/CE4L0F143/p1718116330342849) is addressed.


```go
    hasAnyDisabledFeatureGates := f.disabled.HasAny(featureGates...)
	hasAllEnabledFeatureGates := f.enabled.HasAll(featureGates...)
	if hasAnyDisabledFeatureGates || !hasAllEnabledFeatureGates {
		logrus.Info("Excluding test: %q (requires feature-gates: %v)", name, featureGates)
		return false
	}

    logrus.Info("Including test: %q", name, featureGates)
    return true
```

rexagod: Refactored for readability.

test/extended/prometheus/collection_profiles.go
rexagod: ACK, dropped the previous `full` profile safety checks since we've changed the approach.

pkg/clioptions/suiteselection/feature_filter.go
jan--f: I would drop this. It's an opinionated change not related to the PR topic. 

While readability is important, here the downside of this change for a future reader of this code trying to determine why the code is this way weighs heavier imo.

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,312 @@
...
+						return fmt.Errorf("no monitors found with collection profile %q", profile)
+					}
+
+					return nil
+				}).Should(o.BeNil())
+			}
+		})
+	})
+
+	var _ = g.Context("in an environment that has the minimal profile pre-set", func() {
```
jan--f, line 147: I have to again point out I'm not a ginko expert, but shouldn't this context have a `BeforeAll` node to apply the minimal profile? Iiuc the intention is that ginko can reorder context nodes as it sees fit, so I'm not sure this will succeed in all cases. This currently relies on the previous context to apply the minimal profile.

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,312 @@
...
+					}
+
+					return nil
+				}).Should(o.BeNil())
+			}
+		})
+	})
+})
+
+func (r *runner) fetchMonitorsFor(selectors ...[2]string) (*prometheusoperatorv1.ServiceMonitorList, error) {
```
jan--f, line 244: Nit:
```suggestion
func (r runner) fetchMonitorsFor(selectors ...[2]string) (*prometheusoperatorv1.ServiceMonitorList, error) {
```

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	for _, selector := range selectors {
+		managedMonitorsSelectors = append(managedMonitorsSelectors, fmt.Sprintf("%s=%s", selector[0], selector[1]))
+	}
+	return r.mclient.ServiceMonitors(operatorNamespaceName).List(tctx, metav1.ListOptions{
+		LabelSelector: strings.Join(managedMonitorsSelectors, ","),
+	})
+}
+
+func (r *runner) makeCollectionProfileConfigurationFor(ctx context.Context, collectionProfile string) error {
```
jan--f, line 256: Nit:
```suggestion
func (r runner) makeCollectionProfileConfigurationFor(ctx context.Context, collectionProfile string) error {
```

test/extended/prometheus/collection_profiles.go
rexagod: The [documentation](https://pkg.go.dev/github.com/onsi/ginkgo/v2@v2.19.1#Describe) mentions that `Context` ([alias for `Describe`](https://github.com/onsi/ginkgo/blob/224be5bf5590e4159b19e750f013cb77d68f2609/dsl/core/core_dsl.go#L47)) supports decorators (i.e., the `Ordered` parameter being passed here) while nesting more containers internally. The containers execute [as they appear](https://github.com/onsi/ginkgo/blob/v2.19.1/internal/ordering.go#L52) (among other "weights" that work towards the same affect, even [`ID`s](https://github.com/onsi/ginkgo/blob/v2.19.1/internal/node.go#L22), in case of `It`s).

As long as the structure is maintained, things should happen as expected. But LMK if this seems prone to manual oversight and I'll add those in.

pkg/clioptions/suiteselection/feature_filter.go
```diff
@@ -68,16 +67,11 @@ func (f *featureGateFilter) includeTest(name string) bool {
...
-	if shouldIncludeTest {
-		logrus.Infof("Including test: %q", name)
-	} else {
-		logrus.Infof("Excluding test: %q (requires feature-gates: %v)", name, featureGates)
+	if f.disabled.HasAny(featureGates...) {
+		return false
 	}
 
-	return shouldIncludeTest
+	return f.enabled.HasAll(featureGates...)
```
rexagod, line 74: Reverted to pre-patch state since this [isn't needed anymore](https://redhat-internal.slack.com/archives/CE4L0F143/p1718119344698569?thread_ts=1718116330.342849&cid=CE4L0F143) (`/verify` seems to pass now).

test/extended/prometheus/collection_profiles.go
rexagod: The [documentation](https://pkg.go.dev/github.com/onsi/ginkgo/v2@v2.19.1#Describe) mentions that `Context` ([alias for `Describe`](https://github.com/onsi/ginkgo/blob/224be5bf5590e4159b19e750f013cb77d68f2609/dsl/core/core_dsl.go#L47)) supports decorators (i.e., the `Ordered` parameter being passed here) while nesting more containers internally. The containers seem to execute [as they appear](https://github.com/onsi/ginkgo/blob/v2.19.1/internal/ordering.go#L52) (among other "weights" that work towards the same affect, and [`ID`s](https://github.com/onsi/ginkgo/blob/v2.19.1/internal/node.go#L22), in case of `It`s).

I believe that as long as the structure is maintained, things should happen as expected. But LMK if this seems prone to manual oversight and I'll add those in.

pkg/clioptions/suiteselection/feature_filter.go
```diff
@@ -5,9 +5,10 @@ import (
 	"fmt"
 	"regexp"
 
-	clientconfigv1 "github.com/openshift/client-go/config/clientset/versioned"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+
+	clientconfigv1 "github.com/openshift/client-go/config/clientset/versioned"
```
jan--f, line 11: Let's drop this too.

test/extended/prometheus/collection_profiles.go
jan--f: How about this?

test/extended/prometheus/collection_profiles.go
jan--f: Ahh I missed that. Should we maybe have a comment preceding the context so future readers don't miss this like I did?

jan--f (changes requested):

rexagod:
Testcases breaking after rebase, in addition to the Ginkgo's earlier `Interrupted by user` for two of them. I'll take a look.

rexagod:
Two of the testcases throw a `discovery.StaleGroupVersionError`, which seems to be stemming from non-affected area: https://github.com/openshift/origin/blob/e0b5e7d37ae77bbee2a33b43624b60a05f1a70fd/test/extended/util/client.go#L313

test/extended/prometheus/collection_profiles.go
```diff
@@ -144,92 +163,52 @@ var _ = g.Describe("[sig-instrumentation][OCPFeatureGate:MetricsCollectionProfil
...
-						relabelledMetricsCountPre = int(queryResponse.Data.Result[0].Value)
-					}
-					if j == 1 {
-						relabelledMetricsCountPost = int(queryResponse.Data.Result[0].Value)
-					}
+				enabled, err := r.isProfileEnabled(profile)
+				if err != nil {
+					return err
 				}
-				if profile == collectionProfileMinimal && relabelledMetricsCountPost-relabelledMetricsCountPre != regexpStringWholeMetricNamesCount {
```
rexagod: I confirmed that this is flaky and may not be the case (most likely owing to the fact that `len(samples @ ts) != len(metric_families @ ts)` due to cached artifacts), and may not be a safe assumption under unpredictable environments, hence the revert to a stable approach.

test/extended/prometheus/collection_profiles.go
```diff
@@ -92,28 +93,46 @@ var _ = g.Describe("[sig-instrumentation][OCPFeatureGate:MetricsCollectionProfil
...
 
-	var _ = g.Context("in an environment that migrates from default to minimal collection profile", func() {
-		g.It("should apply default collection profile initially", func() {
+	defaultOnlyMetric := "kube_deployment_status_replicas"
+
+	g.Context("initially, in a homogeneous default environment,", func() {
+		g.BeforeAll(func() {
 			o.Eventually(func() error {
-				monitors, err := r.fetchMonitorsFor([2]string{collectionProfileFeatureLabel, collectionProfileDefault})
+				enabled, err := r.isProfileEnabled(collectionProfileDefault)
```
rexagod, line 101: Introduced a `r.isProfileEnabled` to improve resiliency and be order independent (except the `initial` context, as it's reasonable that it executes first (for certain pre-checks), all others can be shuffled).

test/extended/prometheus/collection_profiles.go
```diff
@@ -92,28 +93,46 @@ var _ = g.Describe("[sig-instrumentation][OCPFeatureGate:MetricsCollectionProfil
 		o.Expect(err).To(o.BeNil())
 	})
 
-	var _ = g.Context("in an environment that migrates from default to minimal collection profile", func() {
-		g.It("should apply default collection profile initially", func() {
+	defaultOnlyMetric := "kube_deployment_status_replicas"
+
+	g.Context("initially, in a homogeneous default environment,", func() {
```
rexagod, line 98: Reworded the context texts to be more readable but also concise. Same for their contents as well, which have been stripped of redundant checks, etc. Also added a comment at the top for the execution order of sub-containers being guaranteed to execute in the order in which they encounter.

test/extended/prometheus/collection_profiles.go
```diff
@@ -92,28 +93,46 @@ var _ = g.Describe("[sig-instrumentation][OCPFeatureGate:MetricsCollectionProfil
 		o.Expect(err).To(o.BeNil())
 	})
 
-	var _ = g.Context("in an environment that migrates from default to minimal collection profile", func() {
-		g.It("should apply default collection profile initially", func() {
+	defaultOnlyMetric := "kube_deployment_status_replicas"
```
rexagod, line 96: Following up on https://github.com/openshift/origin/pull/28889#discussion_r1646407926: `prometheus_engine_query_log_enabled` was being emitted long after the profile was switched to `minimal` (PTAL below). I suspect I'd have to mark the series for deletion and clean its respective tombstones, but that wasn't the case for this. Nonetheless, the metric itself is checked for (a) [existence in a default environment](https://github.com/openshift/origin/pull/28889/commits/e1651b581419bea5b5a379060a1d2cc2973f356d#diff-00da964b40cc78eccb31c5bd15423de5364fa3dfa65c08a09e089c651cb28281R112), and (b) [absence in a minimal environment](https://github.com/openshift/origin/pull/28889/commits/e1651b581419bea5b5a379060a1d2cc2973f356d#diff-00da964b40cc78eccb31c5bd15423de5364fa3dfa65c08a09e089c651cb28281R205), so we can catch an incident where it was added to the minimal set of metrics (the tests will fail).

<details>
<summary>Details</summary>

Screenshot 2024-08-04 at 4 29 35 AM

</details>

test/extended/prometheus/collection_profiles.go
```diff
@@ -201,14 +201,78 @@ var _ = g.Describe("[sig-instrumentation][OCPFeatureGate:MetricsCollectionProfil
...
+				if err != nil {
+					return err
+				}
+				if len(queryResponse.Data.Result) == 0 {
+					return fmt.Errorf("no result found for metric %q", postRelabelingMetric)
+				}
+				wantCount := int(queryResponse.Data.Result[0].Value)
+
+				kubeStateMetricsMainMetricsString := strings.Join(kubeStateMetricsMainMetrics, "")
+				kubeStateMetricsMainMetricsCountQuery := fmt.Sprintf("count({__name__=~\"%s\"})", kubeStateMetricsMainMetricsString[:len(kubeStateMetricsMainMetricsString)-1 /* drop the last "|" or ")" */])
```
rexagod, line 264: In order to check if a minimal environment is established, compare the sample count for minimal metrics specified in a components' respective profile, to the number of samples captured in Prometheus for the same component.

jan--f:
I think the test code looks good. Do you want to squash the fixup commits down?

rexagod:
/retest

Tests are flaking.

rexagod:
/test e2e-aws-ovn-ipsec-serial

None of the tests affect `e2e-aws-ovn-ipsec-serial`, which is also a non-required job. I believe the job is flaky and may be skipped.

cc @jan--f for `lgtm`. Posting this in the API review forum.

neisw:
Anyone reviewing the failures in [e2e-gcp-ovn-techpreview-serial](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/28889/pull-ci-openshift-origin-master-e2e-gcp-ovn-techpreview-serial/1818990848365301760/artifacts/e2e-gcp-ovn-techpreview-serial/openshift-e2e-test/build-log.txt)?

```
  • [FAILED] [902.571 seconds]
  [sig-instrumentation][OCPFeatureGate:MetricsCollectionProfiles][Serial][Conformance] The collection profiles feature-set in an environment that has the minimal profile pre-set [It] should hide all default-only metrics when minimal collection profile is enabled [Suite:openshift/conformance/serial/minimal]
  github.com/openshift/origin/test/extended/prometheus/collection_profiles.go:148

    [FAILED] Timed out after 900.001s.
    Expected
        <*errors.errorString | 0xc001fcfc90>: 
        no monitors found with collection profile: "minimal" and "app.kubernetes.io/name"="kube-state-metrics-minimal"
        {
            s: "no monitors found with collection profile: \"minimal\" and \"app.kubernetes.io/name\"=\"kube-state-metrics-minimal\"",
        }
    to be nil
    In [It] at: github.com/openshift/origin/test/extended/prometheus/collection_profiles.go:172 @ 08/01/24 15:27:20.654
```

rexagod:
@neisw The test quoted [here](https://github.com/openshift/origin/pull/28889#issuecomment-2374082008) ran on `Thu Aug  1 13:52:31 UTC 2024`. The patch has seen multiple changes since then, in addition to the test not being serialized anymore, so any `-serial` jobs should exclude them.

Feel free to correct me but looking at the JUnit output, none of the logs in the currently failing jobs seem to be related to any of the tests introduced here.

neisw:
Thanks @rexagod - Just seemed worth double checking since this is targeting techpreview.  Did note that run was from a few commits back.  Likely not related but appreciate the recheck.

test/extended/prometheus/collection_profiles.go
```diff
@@ -0,0 +1,365 @@
...
+var _ = g.Describe("[sig-instrumentation][OCPFeatureGate:MetricsCollectionProfiles] The collection profiles feature-set", g.Ordered, func() {
+	defer g.GinkgoRecover()
+
+	o.SetDefaultEventuallyTimeout(15 * time.Minute)
+	o.SetDefaultEventuallyPollingInterval(5 * time.Second)
+
+	r := &runner{}
+
+	g.BeforeAll(func() {
+		if !exutil.IsTechPreviewNoUpgrade(tctx, oc.AdminConfigClient()) {
```
rexagod, line 67: Signature changed during rebase, so modified this (from `exutil.IsTechPreviewNoUpgrade(oc)`.

rexagod:
/test e2e-aws-ovn-ipsec-serial
/test e2e-gcp-ovn-techpreview-serial

Scheduling failures and flakes, retesting.

rexagod:
/cc @jan--f 

(for the re-`/lgtm`)

FYI https://github.com/openshift/origin/pull/28889#discussion_r1792974600 was the only change made since the last one (in addition to the rebase).

rexagod:
cc @neisw for LGTM :)

rexagod:
🎉

PR #28890, opened 2024-06-18
ETCD-611: update cert rotation tests

tjungblu:
This adds the following fixes:
* manual CA rotation does not need openshift-config signers anymore
* fixing the dynamic cert rotation by selecting a node that's actually alive
* recreating the trust bundle
* move to a dedicated test suite to not impact recovery tests

### DIFF
--- a/test/extended/dr/cert_rotation.go
+++ b/test/extended/dr/cert_rotation.go
@@ -2,19 +2,21 @@ package dr
 
 import (
 	"context"
+	"fmt"
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
+	"github.com/openshift/origin/test/extended/prometheus/client"
 	exutil "github.com/openshift/origin/test/extended/util"
 	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"math/rand"
 	"strings"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


test/extended/dr/cert_rotation.go
```diff
@@ -152,7 +156,17 @@ var _ = g.Describe("[sig-etcd][Feature:CertRotation][Suite:openshift/etcd/recove
 		etcdConfigMapClient := oc.AdminKubeClient().CoreV1().ConfigMaps("openshift-etcd")
 		bundleName := "etcd-ca-bundle"
 
-		err := etcdConfigMapClient.Delete(ctx, bundleName, v1.DeleteOptions{})
+		get, err := etcdConfigMapClient.Get(ctx, bundleName, v1.GetOptions{})
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		// it makes little sense to recreate a bundle that only has a single CA in it, so we skip this test when the
+		// randomization will run this before any signer rotation.
+		// TODO(thomas): we should serialize this with the signer rotation somehow
```
hasbro17, line 164: Not holding up the PR because of this but can you elaborate why we don't want to run this if we haven't rotated the signer first.
I get that rotating the signer will give us 2 CAs in the configmap here and then we delete and recreate which gives us 1.
But even with 1 CA having the bundle configmap be recreated is useful no? Although I wonder why anyone would delete that configmap if they haven't already rotated the signer, so maybe not a legit workflow?

tjungblu: > But even with 1 CA having the bundle configmap be recreated is useful no?

this doesn't test much besides that you can regenerate the same configmap. I'll revise this after I'm done with the bundle pruning.

PR #28892, opened 2024-06-19
kubelet down 4.16

dgoodwin:
- **quota.sh: OCP 4.16 no longer creates legacy API tokens**
- **[ETCD-565](https://issues.redhat.com//browse/ETCD-565): add manual etcd signer cert rotation e2e test**
- **bump(k8s.io)**
- **React to library-go changes**
- **make update**
- **Add a "limit" exception for iptables-alerter**
- **bump 4.16 to latest o/k 4.16 release branch**
- **Add new intervals for kubelet metrics endpoints down**

PR #28896, opened 2024-06-20
TRT-1718: Add new intervals for kubelet metrics endpoints down

dgoodwin:
New monitortest for querying prom at the end of run to get precise times
it found the kubelet /metrics and /metrics/cadvisor endpoints down. This
is to attempt to gain better insights into what's going on in the
cluster at the time this occurs.

Also includes a test that will flake if we see this outside a node
reboot.


dgoodwin:
The problem we're chasing here *might* only be happening in 4.16, I've been testing this change over here: https://github.com/openshift/origin/pull/28892

Last attempt (first time with the new test) did not install, so learned my lesson and launched three. None the less review would be helpful here.

/hold

pkg/monitortests/testframework/metricsendpointdown/monitortest.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/openshift/origin/pkg/monitortestframework"
+	"github.com/openshift/origin/pkg/monitortests/clusterversionoperator/operatorstateanalyzer"
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+	"github.com/openshift/origin/pkg/test/ginkgo/junitapi"
+	"k8s.io/client-go/rest"
+)
+
+const testName = "[sig-node] kubelet metrics endpoints should not be unreachable"
```
neisw, line 18: 'should not be unreachable' -> 'should be reachable' might be a little more readable.

neisw:
/lgtm

/hold

nit on the test naming but release the hold if you prefer to get it merged as is

pkg/monitortests/testframework/metricsendpointdown/monitortest.go
dgoodwin: changed to "should always be reachable"

dgoodwin:
We need to get this back into 4.16 and prior to the test rename, this had passed all required presubmits. As such I think I'm going to override everything so we can get the backport started.

neisw:
The bot really wants you to know that all tests passed...

PR #28901, opened 2024-06-21
TRT-1721: Add new intervals for kubelet metrics endpoints down

dgoodwin:
Backporting #28896 to 4.16 so we can chase the problem we're after.


soltysh (approved):
/label staff-eng-approve
/label backport-risk-assessed
/label cherry-pick-approved

PR #28920, opened 2024-07-02
NO-JIRA: fix: update skip logic for watchdog

eggfoobar:
the logic for skipping watch dog check on sno upgrade needs to be inverted since FromRelease will be populated with a string version when the jobtype is a single node upgrade


neisw:
/lgtm

lines up with https://github.com/openshift/origin/blob/f0a97dfd58cb15cd200af211297281a10aebe58c/pkg/monitortestlibrary/allowedalerts/watchdog.go#L85

PR #28926, opened 2024-07-08
TRT-1702: record risk analysis results in BQ

sosiouxme:


pkg/riskanalysis/cmd.go
```diff
@@ -92,80 +95,216 @@ func (opt *Options) Run() error {
...
 		defer cancelFn()
-		startTime := time.Now()
-		logrus.Infof("Requesting risk analysis (attempt %d/%d) from: %s", i, maxRetries, sippyURL)
+
+		logrus.Infof("Requesting risk analysis (attempt %d/%d) from: %s", i, maxTries, sippyURL)
 		resp, err = client.Do(req.WithContext(ctx))
-		endTime := time.Now()
-		duration := endTime.Sub(startTime)
-		logrus.Infof("Call to sippy finished after: %s", duration)
+		cancelFn() // cancel the context timeout we just used.
```
neisw, line 193: Looks like you call this via defer on line 189, is it intentional to call it again?

neisw: Ok, looks like the same call was previously at line 121.

pkg/riskanalysis/cmd.go
```diff
@@ -92,80 +95,216 @@ func (opt *Options) Run() error {
...
+	for _, test := range analysis.Tests {
+		dataFile.Rows = append(dataFile.Rows, map[string]string{
+			"TestName":  test.Name,
+			"RiskLevel": fmt.Sprintf("%d", test.Risk.Level.Level),
+			"RiskName":  test.Risk.Level.Name,
+		})
+	}
+	err = dataloader.WriteDataFile(filepath.Join(opt.JUnitDir, raTestRiskFileName), dataFile)
+	if err != nil {
+		logrus.WithError(err).Error("Error writing risk analysis test results to autodl file")
```
neisw, line 306: nit, but noticed you used the filename for the requests logs https://github.com/openshift/origin/blob/d30f5407d37771c8c9074049240960701f5b154f/pkg/riskanalysis/cmd.go#L254

pkg/riskanalysis/cmd.go
sosiouxme: it's how it was when i started, just moved slightly. i wasn't sure i understood the context well enough to change. i gather it's to clean up the timeout, and if it hasn't already timed out, keep it from firing in the middle of the rest of the code (but what would happen if it did?). the defer does seem superfluous as there's no way it could make it through an iteration without calling `cancelFn()` but... calling it twice seems unlikely to cause problems?

pkg/riskanalysis/cmd.go
```diff
@@ -92,80 +95,216 @@ func (opt *Options) Run() error {
...
+		TableName: "risk_analysis_test_results",
+		Schema: map[string]dataloader.DataType{
+			"TestName":  dataloader.DataTypeString,
+			"RiskLevel": dataloader.DataTypeInteger,
+			"RiskName":  dataloader.DataTypeString,
+		},
+		Rows: []map[string]string{},
+	}
+	dataFile.Rows = append(dataFile.Rows, map[string]string{
+		"TestName":  "", // no test name means overall risk
```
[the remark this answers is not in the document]
sosiouxme: I ended up separating these into two tables after all, since they had different enriched content.

pkg/riskanalysis/cmd.go
sosiouxme: reproduced this pattern everywhere I think. a bit superfluous since i think the attached error includes the filename anyway, but why not.

sosiouxme:
not sure what the bot is smoking, since blocking tests all passed. looks like it's trying to retest but there are none required? anyway i don't see a need for the hold.

the RA looks scary but there's a lot going on and these can't possibly be related to my change.

/hold cancel

sosiouxme:
hm. yeah. the bots seem to have mixed ideas about what tests are required.
/test e2e-metal-ipi-ovn-ipv6

PR #28933, opened 2024-07-10
SDN-5133: Add udn test framework

maiqueb:
This PR introduces a test framework with the e2e tests for the user defined primary network feature.

### DIFF
--- a/test/extended/networking/network_segmentation.go
+++ b/test/extended/networking/network_segmentation.go
@@ -1,0 +1,861 @@
+package networking
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"math/big"
+	"net"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	nadapi "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	nadclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/typed/k8s.cni.cncf.io/v1"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/rand"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/kubectl/pkg/util/podutils"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2ekubectl "k8s.io/kubernetes/test/e2e/framework/kubectl"
+	frameworkpod "k8s.io/kubernetes/test/e2e/framework/pod"
+	admissionapi "k8s.io/pod-security-admission/api"
+	utilnet "k8s.io/utils/net"
+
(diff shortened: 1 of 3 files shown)


maiqueb:
Addressed the PSA issues in the latest forced push.

Seems now we need to figure out what's broken in the feature. 

Maybe something's not enabled  ?...

test/extended/networking/network_segmentation.go
```diff
@@ -0,0 +1,862 @@
...
+						continue
+					}
+					framework.Logf("Found default route %v, deleting", defRoute)
+					cmd := append(ipCommand, "route", "del")
+					_, err = e2ekubectl.RunKubectl(udnPod.Namespace,
+						append(cmd, strings.Split(defRoute, " ")...)...)
+					Expect(err).NotTo(HaveOccurred())
+
+					// 2. Add a new default route to use default network interface
+					gatewayIP := NextIP(Network(podIP).IP)
```
maiqueb, line 323: to avoid vendoring a LOT of changes I've copied this code directly from `containernetworking/plugins`, which has an apache 2.0 license.

Still, is there any reason for us to require using this function ? Is there another alternative - e.g. using an hard-coded IP, etc ? @npinaeva (tagging you since you've wrote this test upstream).

maiqueb:
I'll need some help from @npinaeva to debug her isolation tests.

Let's sync tomorrow about this. 

The L3 east / west test is also not passing on ipv6 (it passes on ipv4 ...). Will also need some help w/ that.

test/extended/networking/network_segmentation.go
maiqueb: After syncing offline, we reached the conclusion that we **could** ensure one of the default network pods are scheduled in the same node as the UDN pod, parse its annotations, and grab its gateway IP (we would need that for layer3, where each node has a different subnet).

Seems too much work - let's keep this copied code for now.

maiqueb:
aaaaaaaaaaaand we have a [green lane](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/28933/pull-ci-openshift-origin-master-e2e-gcp-ovn-techpreview/1813936709109288960).

/acknowledge-critical-fixes-only 
/cc @qinqon

maiqueb:
@tssurya / @kyrtapz 

how can I clear the [vendor-update](https://github.com/openshift/origin/labels/vendor-update) label ?

tssurya:
no idea lol maybe @knobunc can help?

qinqon (commented):
/lgtm
/approve

tssurya:
@DennisPeriquet : could you please help out here? the vendor-update label is not going away even if now we don't do vendor updates and manually copied over things we needed for the tests.. could you PTAL and give your blessing on this one? Thanks in advance!

DennisPeriquet:
@tssurya according to [tide](https://prow.ci.openshift.org/pr?query=is%3Apr+repo%3Aopenshift%2Forigin+author%3Amaiqueb+head%3Aadd-udn-test-framework), the                     `vendor-update` is not holding back the merge. But the `acknowledge-critical-fixes-only` is.  I believe the TRT team is [about to remove](https://github.com/openshift/release/pull/54634) the `acknowledge-critical-fixes-only` requirement very soon so we can wait for that.

trozet (approved):
/approve

tssurya:
@dgoodwin could you PTAL this

PR #28935, opened 2024-07-11
CNF-8553: add a test for applying the firewall by the commatrix that created by the endpointslice and make sure that will not affect the behavior of the cluster

aabughosh:
add a test for applying the firewall by the commatrix that created by the endpointslice and make sure that will not affect the behavior of the cluster
jira link https://issues.redhat.com/browse/CNF-8553


sosiouxme:
administrative:
- looks like you at least need to run `hack/update-generated.sh` to update generated tests.
- could you link a jira or other justification?
- could changes be consolidated into one or a few commits?

content concerns:
- are all the dependency updates really necessary?
- is there not already a function like `isSNOCluster`?
- this sets up a firewall, should it not also tear it down at some point?
- should this make an exclusion for microshift which lacks the infrastructure api?

aabughosh:
> * could changes be consolidat
are all the dependency updates really necessary? i need the commatrix repo i ran go mod vendor
is there not already a function like isSNOCluster? i didnt check the code but i created mine
this sets up a firewall, should it not also tear it down at some point? thats test adding all the used ports to the nft table and drop the other, first we are seeing all the used ones by the endpointslice resource 
should this make an exclusion for microshift which lacks the infrastructure api? mm im not sure

test/extended/networking/apply_commatrix_firewall.go
```diff
@@ -0,0 +1,136 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-network][Feature:commatrix][Serial]", func() {
+	g.It("should apply firewall by open ports on nodes", func() {
+		artifactsDir := filepath.Join(exutil.ArtifactDirPath(), "commatrix")
```
sabinaaledort, line 29: Not used

test/extended/networking/apply_commatrix_firewall.go
```diff
@@ -0,0 +1,136 @@
...
+		g.By("generating the commatrix")
+		kubeconfig, ok := os.LookupEnv("KUBECONFIG")
+		if !ok {
+			g.Fail("must set the KUBECONFIG environment variable")
+		}
+		// Get the current working directory
+		dir, err := os.Getwd()
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		fmt.Println(" getting current working directory:", dir)
```
sabinaaledort, line 42: Why do you need the working directory?

test/extended/networking/apply_commatrix_firewall.go
```diff
@@ -0,0 +1,136 @@
...
+		deployment := commatrix.MNO
+		isSNO, err := isSNOCluster(cs.ConfigV1Interface)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if isSNO {
+			deployment = commatrix.SNO
+		}
+		format := "nft"
+		mat, err := commatrix.New(kubeconfig, "", "", commatrix.Cloud, deployment)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		err = commatrix.WriteMatrixToFileByType(*mat, "communication-matrix", format, deployment, dir)
```
sabinaaledort, line 56: Why not storing in artifacts so it can be reviewed if needed?

test/extended/networking/apply_commatrix_firewall.go
```diff
@@ -0,0 +1,136 @@
...
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		defer func() {
+			err := debug.DeleteNamespace(cs, ns)
+			o.Expect(err).ToNot(o.HaveOccurred())
+		}()
+		g.By("apply firewall on master nodes")
+
+		localFilePath := dir + "/communication-matrix-master.nft"
+		fmt.Println(" local path:", localFilePath)
```
sabinaaledort, line 70: Why do you print here the file path?

test/extended/networking/apply_commatrix_firewall.go
```diff
@@ -0,0 +1,136 @@
...
+		fileContent, err := os.ReadFile(localFilePath)
```
sabinaaledort, line 71: Can you generate the nftables format and use it instead of writing to file and reading?

test/extended/networking/apply_commatrix_firewall.go
```diff
@@ -0,0 +1,136 @@
...
+})
+
+func applyRulesToNode(cs *clientutil.ClientSet, role, ns string, fileContent []byte) error {
+	nodesList, err := cs.CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list nodes: %w", err)
+	}
+	for _, node := range nodesList.Items {
+		nodeRole := nodesutil.GetRole(&node)
+		if nodeRole != role {
```
sabinaaledort, line 98: Can just check nodesutil.GetRole(&node), nodeRole var is not really used

test/extended/networking/apply_commatrix_firewall.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to create debug pod on node %s: %w", node.Name, err)
+		}
+		defer func() {
+			err := debugPod.Clean()
+			if err != nil {
+				fmt.Printf("failed cleaning debug pod %s: %v", debugPod, err)
+			}
+		}()
+		tempFilePath := "firewall.nft"
```
sabinaaledort, line 111: tempFilePath var is not really used, can just set in the print function

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+		}()
+
+		command := fmt.Sprintf("echo '%s' > %s", string(NFTtable), "firewall.nft")
+		_, err = debugPod.Exec(command)
+		if err != nil {
+			return fmt.Errorf("failed to execute rule %q on node %s: %w", command, node.Name, err)
+		}
+
+		command = "nft -f firewall.nft "
```
sabinaaledort, line 109: Would this be saved after a node reboot? Some of the tests perform reboots

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,151 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		ns := consts.DefaultDebugNamespace + "firewall"
+		err = debug.CreateNamespace(cs, ns)
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		defer func() {
+			err := debug.DeleteNamespace(cs, ns)
+			o.Expect(err).ToNot(o.HaveOccurred())
+		}()
```
sosiouxme, line 66: does cleaning up the namespace also clean up the firewall for the following tests?

sosiouxme (commented):
/approve

i do not have enough context to lgtm so i'll leave that to the other reviewer(s)

test/extended/networking/commatrix.go
aabughosh: no

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,163 @@
...
+	nodesutil "github.com/openshift-kni/commatrix/nodes"
+	"github.com/openshift-kni/commatrix/types"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("[sig-network][Feature:commatrix][Serial]", func() {
+	g.It("should apply firewall by open ports on nodes", func() {
```
sabinaaledort, line 27: I think better description is needed here, it's blocking all ports except the ones OCP is listening on

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,163 @@
...
+		g.By("generating the commatrix")
+		kubeconfig, ok := os.LookupEnv("KUBECONFIG")
+		if !ok {
+			g.Fail("must set the KUBECONFIG environment variable")
+		}
+		// Get the current working directory
+		dir, err := os.Getwd()
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		fmt.Println(" getting current working directory:", dir)
```
sabinaaledort, line 37: Why do you print the working directory?

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,163 @@
...
+
+var _ = g.Describe("[sig-network][Feature:commatrix][Serial]", func() {
+	g.It("should apply firewall by open ports on nodes", func() {
+		g.By("generating the commatrix")
+		kubeconfig, ok := os.LookupEnv("KUBECONFIG")
+		if !ok {
+			g.Fail("must set the KUBECONFIG environment variable")
+		}
+		// Get the current working directory
+		dir, err := os.Getwd()
```
sabinaaledort, line 34: Not used

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,163 @@
...
+			err := debugPod.Clean()
+			if err != nil {
+				fmt.Printf("failed cleaning debug pod %s: %v", debugPod, err)
+			}
+		}()
+
+		command := fmt.Sprintf("echo '%s' > %s", string(NFTtable), "firewall.nft")
+		_, err = debugPod.Exec(command)
+		if err != nil {
+			return fmt.Errorf("failed to execute rule %q on node %s: %w", command, node.Name, err)
```
sabinaaledort, line 107: wrong err msg

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		command = "nft -f firewall.nft "
+		_, err = debugPod.Exec(command)
+		if err != nil {
+			return fmt.Errorf("failed to execute rule %q on node %s: %w", command, node.Name, err)
```
sabinaaledort, line 113: wrong err msg, it's not executing a rule it's applying a set of rules

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		// Check the output of nft list ruleset
+		checkCommand := "nft list ruleset"
+		output, err := debugPod.Exec(checkCommand)
+		if err != nil {
+			return fmt.Errorf("failed to list NFT ruleset on node %s: %w", node.Name, err)
+		}
+
+		if len(output) == 0 {
```
sabinaaledort, line 118: This doesn't actually validating the rules you applied since this output is never empty

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,158 @@
...
+			return fmt.Errorf("no output from 'nft list ruleset' on node %s: ", node.Name)
+		}
+
+		command = fmt.Sprintf("echo '%s' > /host/etc/nftables.conf", strings.ReplaceAll(string(output), "'", "\\'"))
```
sabinaaledort, line 122: Can we print the applied ruleset to artifacts?

aabughosh: yes

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,215 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var _ = g.Describe("[sig-network][Feature:commatrix][Serial]", func() {
+	g.It("should apply firewall by blocking all ports except the ones OCP is listening on", func() {
+		g.By("generating the commatrix")
+
+		cs, err := client.New()
+		if err != nil {
+			panic(fmt.Errorf("failed creating the k8s client: %w", err))
```
sabinaaledort, line 34: It's a test, use o.Expect(err).NotTo(o.HaveOccurred()) instead of panic

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,215 @@
...
+			g.By("apply firewall on worker nodes")
+			err = applyRulesToNode(cs, "worker", workerNFT)
+			o.Expect(err).ToNot(o.HaveOccurred())
+		}
+	})
+})
+
+func applyRulesToNode(cs *client.ClientSet, role string, NFTtable []byte) error {
+	utilsHelpers := utils.New(cs)
+	ns := consts.DefaultDebugNamespace + "firewall"
```
sabinaaledort, line 82: Why do you use  consts.DefaultDebugNamespace + "firewall", can you use the same ns? or just set here the ns you want :)

test/extended/networking/commatrix.go
aabughosh: mistake

test/extended/networking/commatrix.go
aabughosh: hh yah i wanted to use with firewall name so i changed now

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,215 @@
...
+		command := []string{
+			"bash", "-c",
+			fmt.Sprintf("mkdir -p nft && echo '%s' > nft/firewall.nft", string(NFTtable)),
+		}
+		_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+		if err != nil {
+			return fmt.Errorf("failed to save rule set to file on node %s: %w", node.Name, err)
+		}
+
+		installCmd := []string{"bash", "-c", "yum update -y && yum install -y nftables"}
```
sabinaaledort, line 133: Maybe you should use an image that has nftables

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,215 @@
...
+			"/usr/sbin/nft list ruleset > /etc/nftables.conf",
+		}
+
+		_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+		if err != nil {
+			return fmt.Errorf("failed to save NFT ruleset to file on node %s: %w", node.Name, err)
+		}
+
+		command = []string{"bash", "-c",
+			"systemctl enable nftables",
```
sabinaaledort, line 169: Why is this needed?

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,215 @@
...
+			return fmt.Errorf("failed to list NFT ruleset on node %s: %w", node.Name, err)
+		}
+		fmt.Printf("Command output: %s\n", string(output))
+
+		if len(output) == 0 {
+			return fmt.Errorf("no output from 'nft list ruleset' on node %s: ", node.Name)
+		}
+		command = []string{
+			"bash", "-c",
+			"/usr/sbin/nft list ruleset > /etc/nftables.conf",
```
sabinaaledort, line 160: Is there a reason to list the ruleset twice?

aabughosh: one for the output and the other for the save after reboot

aabughosh: i did twice since the output is so long somtimes and i cant echo it to file so better to do like this

test/extended/networking/commatrix.go
aabughosh: to save after reboot

test/extended/networking/commatrix.go
aabughosh: mm i dont know how to find really
we can create ours that have the image and then to install to it the nftables
u have an idea from where i can find?

test/extended/networking/commatrix.go
sabinaaledort: can you add a comment?

PR #28946, opened 2024-07-23
Add communication matrix test under networking

shirmoran:
The commatrix module automatically generates an accurate and up-to-date communication
flows matrix that can be delivered to customers as part of product documentation for all
ingress flows of OpenShift (multi-node and single-node deployments).

This test validates that the [documented network flow Matrix](https://69720--ocpdocs-pr.netlify.app/openshift-enterprise/latest/installing/install_config/configuring-firewall.html#network-flow-matrix_configuring-firewall) covers all the ingress traffic open ports on the nodes, by comparing it to the generated commatrix.
If the documented matrix differ from the generated one, the test will fail, and the documented matrix should be updated (manually) according to the generated matrix.

The generated artifacts are:

1.  doc-commatrix - The documented communication matrix.
2. new-commatrix - The new generated communication matrix (in case of test failure, documented matrix should be updated to new-commatrix) 
3. doc-diff-new - Diff between the documented and new matrices.


test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,218 @@
...
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		g.By("Filter documented commatrix for diff generation")
+		// if cluster is a SNO exclude MNO static entries in diff generation
+		if isSNO {
+			docComDetailsList = excludeGivenStaticEntries(docComDetailsList, &types.ComMatrix{Matrix: types.MNOStaticEntries}, nil)
+		} else {
+			// Exclude specific master entries that are documented as both worker and master,
+			// and are open only on the worker node in MNO clusters.
+			docComDetailsList = excludeGivenStaticEntries(docComDetailsList, &types.ComMatrix{Matrix: MNOExcludedMasterComDetails}, nil)
```
danwinship, line 115: This seems weird. It seems like it would be better to have separate CSV (or whatever) files for the different types, or else to have a single "merged" file with another column/field indicating which types of clusters the row applies to.

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,218 @@
...
+
+		}
+
+		// if cluster is running on BM exclude Cloud static entries in diff generation
+		// else cluster is running on Cloud and exclude BM static entries in diff generation.
```
danwinship, line 120: likewise, but also, I don't think you can really treat "cloud" as a single kind of thing, especially if this is eventually going to have Egress rules as well (since different clouds need access to different APIs)

danwinship (commented):
I am just catching up on this (and the associated enhancement, and I'm going to comment there as well)...

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,161 @@
...
+	clientutil "github.com/openshift-kni/commatrix/pkg/client"
+	commatrixcreator "github.com/openshift-kni/commatrix/pkg/commatrix-creator"
+	"github.com/openshift-kni/commatrix/pkg/endpointslices"
+	"github.com/openshift-kni/commatrix/pkg/types"
+	"github.com/openshift-kni/commatrix/pkg/utils"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const (
+	docCommatrixUrl  = "https://raw.githubusercontent.com/openshift/openshift-docs/main/snippets/network-flow-matrix.csv"
```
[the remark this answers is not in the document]
yuvalk: the idea of the test is to make sure the doc is up to date
and we dont regress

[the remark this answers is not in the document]
yuvalk: agree it should be versioned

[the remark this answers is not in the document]
yuvalk: re CSV, am not sure if docs can consume something else? but if so, I believe this is an easy change

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,252 @@
...
+		docComMatrix := &types.ComMatrix{Matrix: docComDetailsList}
+
+		g.By("generating diff between matrices for testing purposes")
+		diff := matrixdiff.Generate(newComMatrix, docComMatrix)
+		diffStr, err := diff.String()
+		o.Expect(err).ToNot(o.HaveOccurred())
+		err = os.WriteFile(filepath.Join(artifactsDir, "doc-diff-new"), []byte(diffFileComments+diffStr), 0644)
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		g.By("comparing new and documented commatrices")
```
aabughosh, line 160: Typo: commatrix

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,252 @@
...
+			g.Skip(fmt.Sprintf("If the cluster version is lower than the lowest version that "+
+				"has a documented communication matrix (%v), skip test", minimalDocCommatrixVersion))
+		}
+		docCommatrixVersionedUrl := strings.Replace(docCommatrixBaseUrl, "VERSION", clusterXYVersion, 1)
+		resp, err := http.Get(docCommatrixVersionedUrl)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		defer resp.Body.Close()
+		// expect response status code to differ from status not found
+		docURLStatusIsNotFound := resp.StatusCode == http.StatusNotFound
+		o.Expect(docURLStatusIsNotFound).To(o.BeFalse())
```
sabinaaledort, line 106: Expect(resp.StatusCode).To(Equal(http.StatusNotFound)))

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,252 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		floatClusterXYVersion, err := strconv.ParseFloat(clusterXYVersion, 64)
+		o.Expect(err).ToNot(o.HaveOccurred())
+		if floatClusterXYVersion < minimalDocCommatrixVersion {
+			g.Skip(fmt.Sprintf("If the cluster version is lower than the lowest version that "+
+				"has a documented communication matrix (%v), skip test", minimalDocCommatrixVersion))
+		}
+		docCommatrixVersionedUrl := strings.Replace(docCommatrixBaseUrl, "VERSION", clusterXYVersion, 1)
+		resp, err := http.Get(docCommatrixVersionedUrl)
+		o.Expect(err).ToNot(o.HaveOccurred())
```
sabinaaledort, line 102: If the status is not found won't this return an error?

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,252 @@
...
+		docComDetailsList, err := docComMatrixCreator.GetComDetailsListFromFile()
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		g.By("Filter documented commatrix for diff generation")
+		// if cluster is a SNO exclude standard deployment type's static entries in diff generation
+		if isSNO {
+			// Exclude static entries of standard deployment type and all worker nodes entries.
+			docComDetailsList = excludeGivenStaticEntries(docComDetailsList, &types.ComMatrix{Matrix: types.StandardStaticEntries}, &types.ComMatrix{Matrix: docComDetailsList})
+		} else {
+			// Exclude specific master entries that are documented as both worker and master,
```
sabinaaledort, line 134: Maybe the comment next to the StandardExcludedMasterComDetails var is enough and here we can just mention in general we exclude from standard (in case we add more due to different reasons)

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,252 @@
...
+	})
+})
+
+// getClusterXYersion return cluster's version in X.Y stream
+func getClusterXYersion(configClient *configv1client.ConfigV1Client) (string, error) {
+	clusterVersion, err := configClient.ClusterVersions().Get(context.Background(), "version", metav1.GetOptions{})
+	if err != nil {
+		return "", nil
+	}
+	ClusterVersionParts := strings.SplitN(clusterVersion.Status.Desired.Version, ".", 3)
```
sabinaaledort, line 172: Is that var with a capital letter by mistake?

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,252 @@
...
+
+		g.By("generating diff between matrices for testing purposes")
+		diff := matrixdiff.Generate(newComMatrix, docComMatrix)
+		diffStr, err := diff.String()
+		o.Expect(err).ToNot(o.HaveOccurred())
+		err = os.WriteFile(filepath.Join(artifactsDir, "doc-diff-new"), []byte(diffFileComments+diffStr), 0644)
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		g.By("comparing new and documented commatrices")
+		areEqual := comMatricesAreEqual(*newComMatrix, *docComMatrix)
```
sabinaaledort, line 161: We should use the diff object here, if something is missing in the diff package let's add it

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,252 @@
...
+	if len(diff1.Matrix) > 0 || len(diff2.Matrix) > 0 {
+		return false
+	}
+
+	// Diff matrices are empty, ComMatrices are equal
+	return true
+}
+
+// excludeBareMetalEntries excludes and returns only comDetails that are not bare metal static entries
+func excludeGivenStaticEntries(comDetails []types.ComDetails, masterStaticEntriesMatrix, workerStaticEntriesMatrix *types.ComMatrix) []types.ComDetails {
```
sabinaaledort, line 237: I think this is confusing, it gets master and worker entries but in master for example you send StandardStaticEntries which have both master and worker. I think this can be a func that gets a matrix and a list of entries to ignore/remove (checks the role match), can also be added in the matrix repo

test/extended/networking/commatrix.go
shirmoran: commatrices --> commatrix in plural

test/extended/networking/commatrix.go
shirmoran: which status are you referring to?

test/extended/networking/commatrix.go
shirmoran: yep, thanks!

test/extended/networking/commatrix.go
shirmoran: We want to differ `resp.StatusCode` from `http.StatusNotFound`.
Though `Expect(resp.StatusCode).ToNot(Equal(http.StatusNotFound)))` is a indeed better, thanks!

sebrandon1 (approved):

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,208 @@
...
+			Namespace: "openshift-ingress",
+			Pod:       "router-default",
+			Container: "router",
+			Optional:  false,
+		},
+	}
+)
+
+var _ = g.Describe("[sig-network][Feature:commatrix][Serial]", func() {
+	g.It("generated communication matrix should be equal to documented communication matrix", func() {
```
pacevedom, line 65: Is it possible to annotate this test with `[apigroup:config.openshift.io]` so that it is skipped automatically for MicroShift?
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/28946/pull-ci-openshift-origin-master-e2e-aws-ovn-microshift-serial/1833405970730979328

PR #28950, opened 2024-07-24
NO-JIRA: Revert "pathologicalevents: Ignore leaky RecreatingTerminatedPod"

atiratree:
Reverts https://github.com/openshift/origin/pull/28530 and https://github.com/openshift/origin/pull/28538 which introduced a skip for leaky StatefulSet events. 

The original https://issues.redhat.com/browse/OCPBUGS-27262 bug has been fixed so we can revert these 2 commits.


jan--f:
Nice, thank you.
/lgtm

atiratree:
@dgoodwin @jan--f 
btw, should we also open the cherry-picks?

dgoodwin:
I wouldn't consider it critical to tighten up testing by removing this exception in past releases, but I would not object if the team would prefer to do so.

PR #28945, opened 2024-07-23
SDN-4919: NetworkSegmentation: Add tests for EndpointSlices mirror controller

kyrtapz:
Based on https://github.com/openshift/origin/pull/28964.
Depends on https://github.com/openshift/ovn-kubernetes/pull/2241

This PR refactors and adds tests for NetworkSegmentation featuregate which is currently in TechPreview.

PR #28964, opened 2024-07-31
SDN-4919: Skip network segmentation tests

kyrtapz:
Skip the NetworkSegmentation tests temporarily to allow for breaking ovn-kubernetes changes. 
This will be reverted in https://github.com/openshift/origin/pull/28945. 
/cc @tssurya


tssurya (approved):
/lgtm
temporary step needed given apparently tech preview jobs on an aggregated level are payload blocking

PR #28966, opened 2024-08-01
WIP: Allow udn on sno

maiqueb:
This PR adapts the UDN tests to run on single node openshift; instead of skipping when there is only a single node, we schedule all workloads on the only available node.


maiqueb:
/hold

this PR depends on #28945

maiqueb:
Abandoning in lieu of #28966

Issue #4303, opened 2015-08-21
My sti builder can not reference OpenShift services


jorgemoralespou:
I have created a nexus server in OpenShift, and created a service for it:
````
nexus.ci-tools.svc.cluster.local
````
I`m using the https://github.com/openshift/sti-wildfly builder image, and have a settings.xml file to build using my nexus server, referenced via the service. 

The problem I'm seeing is that the assemble does not get to the service:
````
Downloading: http://nexus.ci-tools.svc.cluster.local:8081/content/groups/public/org/jboss/spec/jboss-javaee-6.0/3.0.2.Final-redhat-15/jboss-javaee-6.0-3.0.2.Final-redhat-15.pom
[ERROR] The build could not read 1 project -> [Help 1]
````

If I try to access same url from the registry pod, I can:
````
--2015-08-21 11:59:02--  http://nexus.ci-tools.svc.cluster.local:8081/content/groups/public/org/jboss/spec/jboss-javaee-6.0/3.0.2.Final-redhat-15/jboss-javaee-6.0-3.0.2.Final-redhat-15.pom
Resolving nexus.ci-tools.svc.cluster.local (nexus.ci-tools.svc.cluster.local)... 127.0.53.53, 172.30.22.4
Connecting to nexus.ci-tools.svc.cluster.local (nexus.ci-tools.svc.cluster.local)|127.0.53.53|:8081... failed: Connection refused.
Connecting to nexus.ci-tools.svc.cluster.local (nexus.ci-tools.svc.cluster.local)|172.30.22.4|:8081... connected.
HTTP request sent, awaiting response... 200 OK
Length: 18316 (18K) [application/xml]
Saving to: 'jboss-javaee-6.0-3.0.2.Final-redhat-15.pom.5'
````

Is the origin-sti-builder:v1.0.3 image propagating the information in resolv.conf to the sti-wildfly? If not, I think this is a big problem, since I can not use internal service references for maven builds.

bparees:
I thought docker always propagated the /etc/resolv.conf from the host into the running container.  is your hosts /etc/resolv.conf correct?

jorgemoralespou:
Do you mean that resolv.conf is get from the host in case of a builder,  as it's plain docker daemon the one executing the build? I'll check it on Monday, once I have access to my laptop.

bparees:
[...]

PR #28971, opened 2024-08-02
OCPBUGS-34816: Fix conformance tests for #4303

jparrill:
The validating admission policies included in the mentioned PR are using a new user declared as system:hosted-cluster-config-operator. This makes the test to fail identifying all the operators querying the KAS.

- https://github.com/openshift/hypershift/pull/4303


jparrill:
/close

I've modified the original PR in order to make it compatible with the current conformance tests

PR #28973, opened 2024-08-02
OCPBUGS-37245: Ignore pod sandbox adds expected to fail

dgoodwin:


pkg/monitortests/network/legacynetworkmonitortests/networking.go
```diff
@@ -136,6 +136,12 @@ func testPodSandboxCreation(events monitorapi.Intervals, clientConfig *rest.Conf
 			// These tests are trying to cause pod sandbox failures, so the errors are intended.
 			continue
 		}
+		if strings.Contains(event.Locator.Keys[monitorapi.LocatorNamespaceKey], "e2e-test-ns-global") &&
+			strings.Contains(event.Locator.Keys[monitorapi.LocatorPodKey], "test-ipv6") {
```
xueqzhan, line 140: Should we check if it is a IPV4 cluster and only skip for that?

PR #28975, opened 2024-08-02
IR-471: Skipping chunksizemib tests

deepsm007:
Skipping the chunksizemib as it will fail until both IR and API PR merges.

https://github.com/openshift/api/pull/1948 
https://github.com/openshift/cluster-image-registry-operator/pull/1073 

TODO revert the PR once above 2 merges


neisw:
/lgtm 
per slack discussion

PR #28978, opened 2024-08-05
TRT-1688: Revert "allow etcd log test to flake under vsphere"

sosiouxme:
Removing the vsphere exclusion as it may obscure useful signal. The affected vsphere tests have not failed since this change, not even to flake.


sosiouxme:
unrelated to the change

PR #28979, opened 2024-08-05
OCPEDGE-1182: fix: remove un-needed test

eggfoobar:
This test does not accuratly capture the bug it references, the matcher would match on correct reasons to mark a mc as degraded thus giving false positives. Also the test is not needed because if the condition the test is checking for were ever to arise the install would simply fail before we ever would come to this state. The bug the test references also has tests so we should be fine to remove this check since it's not providing any value.

### DIFF
--- a/test/extended/node_tuning/node_tuning.go
+++ b/test/extended/node_tuning/node_tuning.go
@@ -1,7 +1,6 @@
 package node_tuning
 
 import (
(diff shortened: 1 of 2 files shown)


eggfoobar:
Forgot to re-gen test list. @suleymanakbas91 can you lgtm again?

PR #28986, opened 2024-08-07
NO-JIRA: create a test that repeatedly rolls out the kube-apiserver and uses m…

deads2k:
…onitor tests to watch for unusual activity

This isn't done yet.  I haven't tested locally, but my goal is to run as `/openshift-tests run openshift/kube-apiserver/rollout`.  However, it appears that `--dry-run` without a kubeconfig is broken again.  Joy.


deads2k:
my clusterbot cluster isn't likely to be ready for a while, but this is the approach I'm thinking about.  I don't expect significant change.

PR #28987, opened 2024-08-07
add monitor test to detect unknown status in nodes

kannon92:


pkg/monitortests/node/nodestateanalyzer/node.go
```diff
@@ -61,6 +62,10 @@ func intervalsFromEvents_NodeChanges(events monitorapi.Intervals, _ monitorapi.R
 			if event.Source == monitorapi.SourceNodeMonitor {
 				nodeStateTracker.OpenInterval(nodeLocator, notReadyState, event.From)
 			}
+		case "Unknown":
+			if event.Source == monitorapi.SourceNodeMonitor {
+				nodeStateTracker.OpenInterval(nodeLocator, unknownState, event.From)
+			}
```
deads2k, line 68: at this point in the code, the ready/notready is accurate for our purposes.  I don't think we need the new unknown state here.

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -73,6 +85,15 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
...
 			}
+
+		case wasUnknown && isUnknown:
+			return []monitorapi.Interval{
+				monitorapi.NewInterval(monitorapi.SourceNodeMonitor, monitorapi.Info).
+					Locator(monitorapi.NewLocator().NodeFromName(node.Name)).
+					Message(monitorapi.NewMessage().Reason("Unknown").
+						WithAnnotation(monitorapi.AnnotationRoles, nodeRoles(node)).
+						HumanMessage("node status is unknown")).Build(now, now),
+			}
```
deads2k, line 96: instead of going unknown here, when we transition to NotReady, can we try to determine if it is an expected notready or an unexpected not ready.

if we can't do it here, we should be able to do it in the analysis, but that should only be required there if we need additional resource/interval insights.

deads2k:
if we find an unexpected notready node, we should fail a test.  You can do this inside of `EvaluateTestsFromConstructedIntervals`

PR #28989, opened 2024-08-07
OCPBUGS-38061: add unexpected ready for monitor tests

kannon92:
We want to add an error on detection for unexpected ready.

We want to fail a test if node goes not ready without a MCO config change.

1) add a test if node goes not ready for unexpected reasons.
2) add a test if node has unreachable taint for unexpected reasons.

### DIFF
--- a/pkg/monitortests/node/watchnodes/monitortest.go
+++ b/pkg/monitortests/node/watchnodes/monitortest.go
@@ -2,6 +2,8 @@ package watchnodes
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/openshift/origin/pkg/monitortestframework"
@@ -42,7 +44,11 @@ func (*nodeWatcher) ConstructComputedIntervals(ctx context.Context, startingInte
 }
 
 func (*nodeWatcher) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
-	return nil, nil
+	// Fail tests when monitor test flags this as an error
+	junits := []*junitapi.JUnitTestCase{}
+	junits = append(junits, unexpectedNodeNotReadyJunit(finalIntervals)...)
+	junits = append(junits, unreachableNodeTaint(finalIntervals)...)
+	return junits, nil
 }
 
 func (*nodeWatcher) WriteContentToStorage(ctx context.Context, storageDir, timeSuffix string, finalIntervals monitorapi.Intervals, finalResourceState monitorapi.ResourcesMap) error {
@@ -52,4 +58,58 @@ func (*nodeWatcher) WriteContentToStorage(ctx context.Context, storageDir, timeS
 func (*nodeWatcher) Cleanup(ctx context.Context) error {
 	// TODO wire up the start to a context we can kill here
 	return nil
+}
+
+func unexpectedNodeNotReadyJunit(finalIntervals monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[sig-node] node-lifecycle detects unexpected not ready node"
+	var failures []string
+	for _, event := range finalIntervals {
+		if event.Message.Reason == monitorapi.NodeUnexpectedReadyReason {
+			failures = append(failures, fmt.Sprintf("%v - %v at from: %v - to: %v", event.Locator.OldLocator(), event.Message.OldMessage(), event.From, event.To))
+		}
+	}
+
+	// failures during a run always fail the test suite
+	var tests []*junitapi.JUnitTestCase
+	if len(failures) > 0 {
+		tests = append(tests, &junitapi.JUnitTestCase{
+			Name:      testName,
+			SystemOut: strings.Join(failures, "\n"),
+			FailureOutput: &junitapi.FailureOutput{
+				Output: fmt.Sprintf("node-lifecycle reports %d unexpected notReady events. The node went NotReady in an unpredicted way.\n\n%v", len(failures), strings.Join(failures, "\n")),
+			},
+		})
+	}
+
+	if len(tests) == 0 {
+		tests = append(tests, &junitapi.JUnitTestCase{Name: testName})
+	}
+	return tests
+}
+
+func unreachableNodeTaint(finalIntervals monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[sig-node] node-lifecycle detects unreachable state on node"
+	var failures []string
+	for _, event := range finalIntervals {
+		if event.Message.Reason == monitorapi.NodeUnreachableReason {
+			failures = append(failures, fmt.Sprintf("%v - %v from %v to %v", event.Locator.OldLocator(), event.Message.OldMessage(), event.From, event.To))
+		}
+	}
+
+	// failures during a run always fail the test suite
+	var tests []*junitapi.JUnitTestCase
+	if len(failures) > 0 {
+		tests = append(tests, &junitapi.JUnitTestCase{
+			Name:      testName,
+			SystemOut: strings.Join(failures, "\n"),
+			FailureOutput: &junitapi.FailureOutput{
+				Output: fmt.Sprintf("node-lifecycle reports %d unexpected node unreachable events. The node went unreachable in an unpredicted way.\n\n%v", len(failures), strings.Join(failures, "\n")),
+			},
+		})
+	}
+
+	if len(tests) == 0 {
+		tests = append(tests, &junitapi.JUnitTestCase{Name: testName})
+	}
+	return tests
 }
--- a/pkg/monitortests/node/watchnodes/node.go
+++ b/pkg/monitortests/node/watchnodes/node.go
@@ -23,19 +23,11 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
 			isCreate = true
 		}
 
-		isReady := false
-		if c := findNodeCondition(node.Status.Conditions, corev1.NodeReady, 0); c != nil {
-			isReady = c.Status == corev1.ConditionTrue
-		}
-
+		isReady := isNodeReady(node)
 		wasReady := false
 		if !isCreate {
-			if c := findNodeCondition(oldNode.Status.Conditions, corev1.NodeReady, 0); c != nil {
-				wasReady = c.Status == corev1.ConditionTrue
-			}
-
+			wasReady = isNodeReady(oldNode)
 		}
-
 		now := time.Now()
 		switch {
 		case isCreate && !isReady:
@@ -74,7 +66,6 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
 						HumanMessage("node is ready")).Build(now, now),
 			}
 		}
-
 		return nil
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/monitortests/node/nodestateanalyzer/monitortest.go
```diff
@@ -34,6 +34,14 @@ func (*nodeStateAnalyzer) ConstructComputedIntervals(ctx context.Context, starti
 }
 
 func (*nodeStateAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
+	// Fail tests when monitor test flags this as an error
+	for _, val := range finalIntervals {
+		if val.Level.String() == "Error" {
```
deads2k, line 39: I don't think this will do what you expect.  `finalIntervals` is all the intervals from every monitortest.  Many of them will choose `Error` for their level.  I think you also need to check for the locator being a node and the reason being  `NotReadyUnexpected`.

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -73,8 +80,16 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
 						WithAnnotation(monitorapi.AnnotationRoles, nodeRoles(node)).
 						HumanMessage("node is ready")).Build(now, now),
 			}
-		}
 
+		case isNodeNotReady && !didConfigChange:
```
deads2k, line 84: This case will never be reached because every time `isNodeNotReady` is true, one of the branches above will be taken.

kannon92: Ah. that is what I get for trying to be cute.

I'm just going to add a new function for the unexpected section. I think that may be more clear and less error prone. I wanted to try and reuse the ready check but I think I am overthinking/over optimizing this.

pkg/monitortests/node/nodestateanalyzer/monitortest.go
kannon92: Thanks for your help on that. I'll work on that.

kannon92: pushed an update. PTAL.

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -73,8 +80,16 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
...
+			return []monitorapi.Interval{
+				monitorapi.NewInterval(monitorapi.SourceNodeMonitor, monitorapi.Error).
+					Locator(monitorapi.NewLocator().NodeFromName(node.Name)).
+					Message(monitorapi.NewMessage().Reason("NotReadyUnexpected").
```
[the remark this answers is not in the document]
kannon92: Updated.

pkg/monitortests/node/watchnodes/node.go
kannon92: Updated. PTAL.

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -237,3 +274,27 @@ func findNodeCondition(status []corev1.NodeCondition, name corev1.NodeConditionT
...
+		isReady = c.Status == corev1.ConditionTrue
+	}
+	return isReady
+}
+
+func doesNodeHaveUnreachableTaints(node *corev1.Node) bool {
+	if node == nil {
+		return false
+	}
+	annotations := node.Annotations
```
deads2k, line 293: why not `node.spec.taints`?

kannon92: you are right. Pushed up a fix for that.

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -157,6 +148,52 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
...
+			var intervals []monitorapi.Interval
+
+			oldConfig := oldNode.Annotations["machineconfiguration.openshift.io/currentConfig"]
+			newConfig := node.Annotations["machineconfiguration.openshift.io/currentConfig"]
+			isOldNodeReady := isNodeReady(oldNode)
+			isNewNodeReady := isNodeReady(node)
+			isConfigTheSame := oldConfig == newConfig
+
+			now := time.Now()
+			if (!isOldNodeReady || !isNewNodeReady) && isConfigTheSame {
```
deads2k, line 167: this doesn't seem quite right.  We only want a single interval created when we transition from Ready to NotReady when the configuration is the same.  In this case, every node update (say we change an image on the node status), will result in a new interval created, right?

perhaps `if isOldNodeReady && !isNewNodeReady && isConfigTheSame` ?

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -157,6 +148,52 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
...
+			var intervals []monitorapi.Interval
+
+			oldConfig := oldNode.Annotations["machineconfiguration.openshift.io/currentConfig"]
+			newConfig := node.Annotations["machineconfiguration.openshift.io/currentConfig"]
+			isOldNodeUnReachable := doesNodeHaveUnreachableTaints(oldNode)
+			isNewNodeUnReachable := doesNodeHaveUnreachableTaints(node)
+			isConfigTheSame := oldConfig == newConfig
+
+			now := time.Now()
+			if (isOldNodeUnReachable || isNewNodeUnReachable) && isConfigTheSame {
```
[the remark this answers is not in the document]
kannon92: So are you saying you want the tolerations to flag a interval but you don't want that interval to be flagged as an error?

We only want to flag an error on NodeReady -> NodeNotReady on an unexpected change?

pkg/monitortests/node/watchnodes/node.go
kannon92: I think we would want an interval in that case?

I thought that the goal would be to create an interval if a node goes NotReady outside of a MCO config? So I think that you would want this.

kannon92: I think your suggestion is good though. We only want an inteval is the newNode is NotReady? If NotReady -> Ready I guess we don't want an interval.

pkg/monitortests/node/watchnodes/monitortest.go
```diff
@@ -42,6 +42,19 @@ func (*nodeWatcher) ConstructComputedIntervals(ctx context.Context, startingInte
...
+			return []*junitapi.JUnitTestCase{{
+				Name: "UnexpectedNodeNotReady", SystemOut: val.Message.HumanMessage,
+			}}, nil
+		}
+		if val.Source == monitorapi.SourceNodeMonitor && val.Message.Reason == monitorapi.NodeUnreachableReason {
+			return []*junitapi.JUnitTestCase{{
+				Name: "Unreachable", SystemOut: val.Message.HumanMessage,
+			}}, nil
+		}
+	}
```
deads2k, line 57: a few things here.

1. to fail a test, the failureoutput in the junit test case must be set
2. a test should only fail once per run
3. a test must either fail or succeed, it cannot be missing entirely.  This means the test needs to be added in success cases.
4. tests should follow a format like `[Jira:"Node / Kubelet"] nodes should not go unready unexpectedly`  The Jira bit automatically categorizes ownership and the test is descriptive

I noticed when checking https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/28989/pull-ci-openshift-origin-master-e2e-metal-ipi-ovn-kube-apiserver-rollout/1823032010419998720

kannon92: Ah thanks for catching that. I will add this.

kannon92: Pushed a fix.

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -157,6 +148,48 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
...
+			oldConfig := oldNode.Annotations["machineconfiguration.openshift.io/currentConfig"]
+			newConfig := node.Annotations["machineconfiguration.openshift.io/currentConfig"]
+			isOldNodeReady := isNodeReady(oldNode)
+			isNewNodeReady := isNodeReady(node)
+			isConfigTheSame := oldConfig == newConfig
+
+			now := time.Now()
+			if isOldNodeReady && !isNewNodeReady && isConfigTheSame {
+				intervals = append(intervals,
+					monitorapi.NewInterval(monitorapi.SourceNodeMonitor, monitorapi.Error).
```
dgoodwin, line 169: Suggest adding a new Source per reason essentially, and make sure to call .Display() before the Build() call. 

The new sippy UI should automatically pick that up as an interval category, you should then see it in the row of boxes at the top of the chart, and can enable it quickly from there.

Source per type of interval will avoid potential overlap where if the node locator is identical, and two things overlap, one could be hidden as the other renders over it. We hack around this in NodeState with the "row" you see in the locator on the right. But if you do the above Source per type of interval you are generating (looks like two here), this should not be an issue.

So essentially just change source here to something new, monitorapi.SourceNodeUnexpectedNotReady, and monitorapi.SourceNodeUnreachable, those both should then automatically appear as checkboxes in the sippy intervals UI. (remembering to set Display() on both)

pkg/monitortests/node/watchnodes/monitortest.go
```diff
@@ -53,3 +59,57 @@ func (*nodeWatcher) Cleanup(ctx context.Context) error {
...
+	return tests
+}
+
+func unreachableNodeTaint(finalIntervals monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[sig-node] node-lifecycle detects unreachable state on node"
+	var failures []string
+	for _, event := range finalIntervals {
+		if strings.Contains(event.Message.HumanMessage, "unexpected node unreachable") {
+			failures = append(failures, fmt.Sprintf("%v - %v", event.Locator.OldLocator(), event.Message.OldMessage()))
```
dgoodwin, line 95: I would output the time the problem occurred in the test output as well, the info is here on event.

pkg/monitortests/node/watchnodes/node.go
kannon92: Updated.

pkg/monitortests/node/watchnodes/monitortest.go
kannon92: So this is the error so the test fails. The interval should display the time.

In this case, I want an error if I detect any intervals that match this reason. So it would be multiple times in this message which may be confusing.

dgoodwin: Ok that's fine but I would include event.From and .To, test output should be clear about when the problem was detected.

dgoodwin: Probably a better idea to check event.Reason == NodeUnreachableReason as well, that is meant for programatic comparison and uses a constant.

kannon92: I'll address this after the CI reports these changes. Just want to see if this works right now.

kannon92: > Probably a better idea to check event.Reason == NodeUnreachableReason as well, that is meant for programatic comparison and uses a constant.

Good call. I have that stashed once I see some jobs run.

kannon92:
Found one test failure in scaling machineSets.

pkg/monitortests/node/watchnodes/monitortest.go
kannon92: pushed. PTAL.

pkg/monitortests/node/watchnodes/node.go
```diff
@@ -157,6 +148,53 @@ func startNodeMonitoring(ctx context.Context, m monitorapi.RecorderWriter, clien
...
+			isOldNodeUnReachable := doesNodeHaveUnreachableTaints(oldNode)
+			isNewNodeUnReachable := doesNodeHaveUnreachableTaints(node)
+			isNodeUnscheduable := node.Spec.Unschedulable
+
+			now := time.Now()
+			if !isOldNodeUnReachable && isNewNodeUnReachable && !isNodeUnscheduable {
+				intervals = append(intervals,
+					monitorapi.NewInterval(monitorapi.SourceUnreachable, monitorapi.Warning).
+						Locator(monitorapi.NewLocator().NodeFromName(node.Name)).
+						Message(monitorapi.NewMessage().Reason(monitorapi.NodeUnreachableReason).
```
deads2k, line 191: UnexpectedUnreachable.

deads2k:
minor comment, looks good otherwise.  I'll take an IOU and a PR next week if you like

/approve
/lgtm

pkg/monitortests/node/watchnodes/node.go
kannon92: NodeUnexpectedUnreachableReason?

Trying to follow similar naming as the other reasons.

kannon92:
```
	NodeUnexpectedReadyReason IntervalReason = "UnexpectedNotReady"
	NodeUnexpectedUnreachableReason     IntervalReason = "UnexpectedUnreachable"

```

deads2k: Yeah, and probably also have a regular unreachable we don't error on

PR #28698, opened 2024-04-10
OCPBUGS-31738: Bump timeout for the pod-network-service endpoints check

mandre:
In certain environments, it can take 3 minutes to download the agnhost image from quay.io. Bump the timeout to 5 minutes to accommodate such slow environments.

PR #28993, opened 2024-08-09
OCPBUGS-38472: adds logging and synchronization to wait-on pod creation before starting the poller

KeerthanaAP:


djslavens (approved):
/lgtm

pkg/monitortests/network/disruptionpodnetwork/monitortest.go
```diff
@@ -202,11 +202,48 @@ func (pna *podNetworkAvalibility) StartCollection(ctx context.Context, adminREST
...
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
 }
 
+// Checks if the Pods are ready for the given deployment
+func (pna *podNetworkAvalibility) podsAreReady(ctx context.Context, deployment *appsv1.Deployment) (bool, error) {
```
deads2k, line 220: to check if a deployment is ready, use the .status.AvailableReplicas and compare against .spec.Replicas.

mandre:
If I understand correctly, this patch prevents the `[sig-network] can collect pod-to-service poller pod logs` and `[sig-network] can collect host-to-service poller pod logs` tests from failing when `[Jira:"Network / ovn-kubernetes"] monitor test pod-network-avalibility setup` fails, but won't fix the underlying error which is that `monitor test pod-network-avalibility setup` is a frequent flake. I traced it to a slow image download from quay.io and proposed https://github.com/openshift/origin/pull/28698 - see my analysis in https://issues.redhat.com//browse/OCPBUGS-31738 - but couldn't figure out exactly _why_ it took longer for this particular image. This seems to affect not only openstack, but other platforms as well.

prb112:
Hey @mandre that was my analysis as well. We've been tracking this one for a while. We assumed it was a throttling along the way - we didn't know it was related to slow image pull. Glad to see we're not the only ones who have investigated.

Perhaps we should merge yours? 

Thanks,

Paul

PR #29003, opened 2024-08-13
OCPBUGS-38859: add api-unreachable-from-client monitor test 

tkashem:
This adds a new timeline `api-uneachable`, grouped by source: `{internal-lb|service-network|external-lb|localhost}`. It scrapes the `rest_client_requests_total` metrics
```
sum(rate(rest_client_requests_total{code="<error>"}[1m])) by(host)
```
The number of timelines in the UI is bound by `source`, so it is a limited number. 

The following shows the disruption intervals where  clients observed api errors via `api-int` 

image
From https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29003/pull-ci-openshift-origin-master-e2e-aws-ovn-kube-apiserver-rollout/1824519484648460288

BTW, looks like the source `webhook.openshift-console-operator.svc:9443`  shows permanent error, looks to be a bad conversion webhook?
```
2024-08-16T21:24:15.202127907Z W0816 21:24:15.202025      14 reflector.go:547] storage/cacher.go:/console.openshift.io/consoleplugins: failed to list console.openshift.io/v1alpha1, Kind=ConsolePlugin: conversion webhook for console.openshift.io/v1, Kind=ConsolePlugin failed: Post "https://webhook.openshift-console-operator.svc:9443/crdconvert?timeout=30s": service "webhook" not found
```

image

Looks like they removed some references of the bad conversion webhook from their crd schema [1]. but I still see reference here [2].

[1]: https://github.com/openshift/api/commit/99bfe922c18172af063871c6ac0c57a86b20b5d2
[2]: https://github.com/openshift/nmstate-console-plugin/blob/7023cea6a5308d9c3470ce77581d9847697b026f/scripts/yaml/crds/console.openshift.io_consoleplugins.yaml#L421


tkashem:
Follow up from: https://github.com/openshift/origin/pull/27976, https://github.com/openshift/origin/pull/27986

/cc @dgoodwin @p0lyn0mial @sanchezl @vrutkovs

pkg/monitortests/client/api_unreachable.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	prometheustypes "github.com/prometheus/common/model"
+)
+
+const (
+	MonitorName = "api-unreachable-from-client"
+)
+
+func NewAPIUnReachableFromClientMonitorTest() monitortestframework.MonitorTest {
```
dgoodwin, line 28: Godoc would be nice here. Initial questions I had looking over this, what client, where the metrics come from, how we interpret them to generate these intervals, and what those intervals mean.

e2echart/e2e-chart-template.html
```diff
@@ -476,6 +484,9 @@ <h5 class="modal-title">Resource</h5>
         timelineGroups.push({group: "apiserver-shutdown", data: []})
         createTimelineData(apiserverShutdownValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isGracefulShutdownActivity, regex)
 
+        timelineGroups.push({group: "api-unreachable", data: []})
+        createTimelineData(isAPIUnreachableFromClientValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isAPIUnreachableFromClientActivity, regex)
+
```
dgoodwin, line 489: This is all good, just for some additional info, we're slowly trying to move to using Sippy for the intervals UI (linked in Debug tools on prow job spyglass), and the UI there will now dynamically adapt to new interval "Sources" and let you display them without code changes. It's URL can also be used to link folks directly to the view / filters you're using.

pkg/defaultmonitortests/types.go
```diff
@@ -122,6 +123,7 @@ func newDefaultMonitorTests(info monitortestframework.MonitorTestInitializationI
 
 	monitorTestRegistry.AddMonitorTestOrDie("monitoring-statefulsets-recreation", "Monitoring", statefulsetsrecreation.NewStatefulsetsChecker())
 	monitorTestRegistry.AddMonitorTestOrDie("metrics-api-availability", "Monitoring", disruptionmetricsapi.NewAvailabilityInvariant())
+	monitorTestRegistry.AddMonitorTestOrDie(clienttest.MonitorName, "API Unreachable", clienttest.NewAPIUnReachableFromClientMonitorTest())
```
dgoodwin, line 126: Second string param should be a component name. kube-apiserver looks to be used above so this might be best here.

pkg/monitor/monitorapi/types.go
```diff
@@ -141,6 +143,8 @@ const (
 	LocatorRowKey                   LocatorKey = "row"
 	LocatorServerKey                LocatorKey = "server"
 	LocatorMetricKey                LocatorKey = "metric"
+
+	LocatorAPIUnreachableSourceKey LocatorKey = "source"
```
dgoodwin, line 147: A little overloaded with interval.Source. Could we come up with a better name? Is "host" ok given that's what it's sourced from on the metric. (even if it's mutated)

pkg/monitor/monitorapi/types.go
```diff
@@ -212,6 +216,8 @@ const (
 	UpgradeCompleteReason IntervalReason = "UpgradeComplete"
 
 	NodeInstallerReason IntervalReason = "NodeInstaller"
+
+	APIUnreachableIntervalReason IntervalReason = "APIUnreachableInterval"
```
dgoodwin, line 220: FromClient would be good to include here as well, and Reason should not mention "Interval". Our Reason's often seem to be equal to Sources when it will only contain one Reason. If anything, Source should be the more generic one if we plan for other "Reason's" to appear there someday.

pkg/monitortests/metrics/rate_analyzer.go
```diff
@@ -0,0 +1,55 @@
...
+// RateSeriesAnalyzer analyzes a rate bases timeseries
+type RateSeriesAnalyzer struct{}
+
+func (RateSeriesAnalyzer) Analyze(ctx context.Context, query QueryRunner, start, end time.Time, callback Callback) error {
+	result, err := query.RunQuery(ctx, start, end)
+	if err != nil {
+		return fmt.Errorf("query returned error, monitor: %s, err: %w", callback.Name(), err)
+	}
+	if result.Type() != prometheustypes.ValMatrix {
+		return fmt.Errorf("expected a prometheus Matrix type, monitor: %s", callback.Name())
```
p0lyn0mial, line 20: we could add `result.Type()` to the err msg for easier debugging.

pkg/monitortests/metrics/rate_analyzer.go
```diff
@@ -0,0 +1,55 @@
...
+						from = nil
+					}
+				default:
+					if from == nil {
+						from = &current
+					}
+				}
+				to = &current
+			}
+			// is the entire range is a disruption?
```
p0lyn0mial, line 48: typo, change to `is the entire range a disruption?`

pkg/monitortests/metrics/rate_analyzer.go
```diff
@@ -0,0 +1,55 @@
...
+
+	zero := prometheustypes.SampleValue(0)
+	matrix := result.(prometheustypes.Matrix)
+	for _, series := range matrix {
+		func() {
+			callback.StartSeries(series.Metric)
+			defer callback.EndSeries()
+
+			var from, to *prometheustypes.SamplePair
+			for _, current := range series.Values {
```
p0lyn0mial, line 31: does the `range` creates the `current` var once and then reuses it ?
if yes, then we cannot simply store `&current` as it will point to the same variable.

pkg/monitortests/metrics/rate_analyzer.go
```diff
@@ -0,0 +1,55 @@
...
+	result, err := query.RunQuery(ctx, start, end)
+	if err != nil {
+		return fmt.Errorf("query returned error, monitor: %s, err: %w", callback.Name(), err)
+	}
+	if result.Type() != prometheustypes.ValMatrix {
+		return fmt.Errorf("expected a prometheus Matrix type, monitor: %s", callback.Name())
+	}
+
+	zero := prometheustypes.SampleValue(0)
+	matrix := result.(prometheustypes.Matrix)
```
p0lyn0mial, line 24: should we check for `nil`/`empty` responses ?
why not to do a type assertion ? (not sure if line `19` is enough)

pkg/monitortests/client/api_unreachable.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (b *intervalBuilder) Name() string { return MonitorName }
+func (b *intervalBuilder) StartSeries(metric prometheustypes.Metric) {
+	source := b.source(metric)
+	b.locator = monitorapi.NewLocator().WithAPIUnreachableFromClient(source)
+}
+func (b *intervalBuilder) EndSeries() { b.locator = monitorapi.Locator{} }
+
+func (b *intervalBuilder) NewDisruptionInterval(metric prometheustypes.Metric, start, end *prometheustypes.SamplePair) {
```
p0lyn0mial, line 119: should we check if `start` and `end` are not `nil` ?
same for `metric` which is a `map` I think ?

pkg/monitortests/client/api_unreachable.go
```diff
@@ -0,0 +1,164 @@
...
+	kubeSvc, err := kubeClient.CoreV1().Services(metav1.NamespaceDefault).Get(ctx, "kubernetes", metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cluster IP from kubernetes.default.svc - %v", err)
+	}
+
+	mt.monitor = &monitor{
+		query: &metrics.PrometheusQueryRunner{
+			Client:      client,
+			QueryString: `sum(rate(rest_client_requests_total{code="<error>"}[1m])) by(host)`,
+			Step:        time.Minute,
```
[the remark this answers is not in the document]
tkashem: I tried to explain this in the godoc:
```
The intervals are scraped from metrics, so they don't have the same granularity, as other intervals, since:
a) in OpenShift, metrics are scraped every 30s
b) for rate to be calculated, we need at lease two samples
```

If an api unreachable interval overlaps with an apiserver shutdown window, it is typically indicative of network issues at the load balancer layer. Since the intervals are grouped by host, we can also narrow it down to a particular host, for example, we have seen cases where connections over internal load balancer to be faulty at times while the service network operated just fine.
Let me know your thoughts.

pkg/monitortests/client/api_unreachable.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+		analyzer: metrics.RateSeriesAnalyzer{},
+		builder: &intervalBuilder{
+			serviceNetworkIP: kubeSvc.Spec.ClusterIP,
```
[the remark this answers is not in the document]
tkashem: not clear what you mean, I believe we are using the cluster IP of `kubernetes.default.svc`

pkg/monitortests/client/api_unreachable.go
tkashem: I don't think it's necessary, the analyzer invokes `NewInterval` with a non nil `start`, and `end`. But if you want i can add nil check

pkg/monitortests/client/api_unreachable.go
```diff
@@ -0,0 +1,164 @@
...
+	b.locator = monitorapi.NewLocator().WithAPIUnreachableFromClient(source)
+}
+func (b *intervalBuilder) EndSeries() { b.locator = monitorapi.Locator{} }
+
+func (b *intervalBuilder) NewDisruptionInterval(metric prometheustypes.Metric, start, end *prometheustypes.SamplePair) {
+	startTime := start.Timestamp.Time()
+	endTime := end.Timestamp.Time()
+	if start == end {
+		// a disruption window with one sample
+		endTime = end.Timestamp.Time().Add(time.Minute)
```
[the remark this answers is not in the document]
tkashem: since we scrape every 30s, a `1s` granularity is not very meaningful here, i think. I am approximating the interval to [t-30s ... t+30s] for now. Thoughts?

pkg/monitortests/client/api_unreachable.go
```diff
@@ -0,0 +1,164 @@
...
+		Condition: monitorapi.Condition{
+			Level:   monitorapi.Error,
+			Locator: b.locator,
+			Message: message,
+		},
+		From:    startTime,
+		To:      endTime,
+		Display: true,
+	}
+	b.intervals = append(b.intervals, interval)
```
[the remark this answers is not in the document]
tkashem: thanks for pointing it out, using the builder now

pkg/monitortests/client/api_unreachable.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// the endpoint used to reach the apiserver
+func (b *intervalBuilder) source(metric prometheustypes.Metric) string {
```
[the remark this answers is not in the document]
tkashem: yes, moved to construction.go

tkashem:
@dgoodwin a new screenshot of the intervals in the UI
image

We can see client talking to the kube-apiserver using the internal load balancer is experiencing error and that coincides with apiserver shutdown interval

pkg/monitortests/client/api_unreachable.go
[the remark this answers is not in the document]
dgoodwin: Ah that makes sense, thanks. Fine as is then.

dgoodwin:
This all looks ok to me. You'll need a jira in title, I'll approve in case there's more you want to sort out with @p0lyn0mial .

/approve

tkashem:
it needs an `acknowledge-critical-fixes-only` label, cc @dgoodwin

tkashem:

(it's informing only, does not fail any test, it will help with the load balancer issues we are seeing in CI)

PR #29008, opened 2024-08-14
[POC][WIP] a new test for detecting ungraceful kube-apiserver termination by kubelet

p0lyn0mial:
Initially I wanted to compare shutdown related events but after [inspecting](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-ci-4.18-e2e-gcp-ovn-upgrade/1822964770265894912) I have realised that event's are sometimes not even persisted in the storage.

The new test added by this PR simply looks for the `Previous pod .* did not terminate gracefully string in the -termination.log` files.

This PR could be made simpler and smaller by copying the existing code instead of reusing it.

### DIFF
--- a/test/extended/apiserver/graceful_termination.go
+++ b/test/extended/apiserver/graceful_termination.go
@@ -279,3 +279,47 @@ func extractAPIServerNameFromAuditFile(auditFileName string) string {
 	}
 	return auditFileName[0:pos]
 }
+
+func gatherMustGatherFor() {
+	// download the audit logs from the cluster
+	o.Expect(oc.Run("adm", "must-gather").Args(args...).Execute()).To(o.Succeed())
+
+	// act
+	expectedDirectoriesToExpectedCount := []string{path.Join(clihelpers.GetPluginOutputDir(tempDir), "audit_logs", "kube-apiserver")}
+	for _, auditDirectory := range expectedDirectoriesToExpectedCount {
+		err := filepath.Walk(auditDirectory, func(path string, info os.FileInfo, err error) error {
+			g.By(path)
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			if info.IsDir() {
+				return nil
+			}
+			fileName := filepath.Base(path)
+			if !clihelpers.IsAuditFile(fileName) {
+				return nil
+			}
+
+			file, err := os.Open(path)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			defer file.Close()
+			fi, err := file.Stat()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if fi.Size() == 0 {
+				return nil
+			}
+
+			gzipReader, err := gzip.NewReader(file)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			err = gzipReader.Close()
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			apiServerName := extractAPIServerNameFromAuditFile(fileName)
+			o.Expect(apiServerName).ToNot(o.BeEmpty())
+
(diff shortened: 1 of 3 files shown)


test/extended/apiserver/graceful_termination.go
```diff
@@ -61,6 +62,45 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Late]", func() {
...
+			g.Skip("oc must-gather doesn't currently support external controlPlaneTopology")
+		}
+
+		tempDir, err := ioutil.TempDir("", "test.oc-adm-must-gather.")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		defer os.RemoveAll(tempDir)
+
+		var finalMessageBuilder strings.Builder
+		terminationRegexp := regexp.MustCompile(`Previous pod .* did not terminate gracefully`)
+		terminationLogsPerServer := gatherMustGatherFor(oc, tempDir, clihelpers.IsTerminationLog, extractAPIServerNameFromTerminationFile)
```
deads2k, line 79: Don't use must-gather for this.  If you just need a hostmount pod, use that.  Removing layers will help future us.

test/extended/apiserver/graceful_termination.go
```diff
@@ -61,6 +62,45 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Late]", func() {
...
+		if *controlPlaneTopology == configv1.ExternalTopologyMode {
+			g.Skip("oc must-gather doesn't currently support external controlPlaneTopology")
+		}
+
+		tempDir, err := ioutil.TempDir("", "test.oc-adm-must-gather.")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		defer os.RemoveAll(tempDir)
+
+		var finalMessageBuilder strings.Builder
+		terminationRegexp := regexp.MustCompile(`Previous pod .* did not terminate gracefully`)
```
deads2k, line 78: Why doesn't the  kube-apiserver itself produce an event when this is detected?

test/extended/apiserver/graceful_termination.go
p0lyn0mial: What is your concern here? 

Please note that this test doesn't download the entire must-gather tarball. It uses must-gather to only collect the audit logs. While it is true that the tarball contains more files than this test needs, I should have given the function a more specific name.

An alternative approach would be to copy only the files we need. Starting new pods on each master, checking their state, and then copying the output doesn't seem simpler.

Another option could be to use `oc adm node-logs` to copy specific files. However, we would still need to identify all the nodes on which this command should be executed.

test/extended/apiserver/graceful_termination.go
p0lyn0mial: It [does](https://github.com/openshift/kubernetes/blob/6892b57d65d25fb0588693bce3d338d8b8b1d2b4/cmd/watch-termination/main.go#L149), we even have a test for [it](https://github.com/openshift/origin/blob/master/test/extended/apiserver/graceful_termination.go#L53). The issue is that sometimes the events are not reported (probably deserves more investigation). For example, the test passed in [the faulty run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-ci-4.18-e2e-gcp-ovn-upgrade/1822964770265894912) even though the servers were not gracefully terminated.

test/extended/apiserver/graceful_termination.go
p0lyn0mial: I have opened a PR that doesn't use the must-gather tool for downloading audit logs - https://github.com/openshift/origin/pull/29015

p0lyn0mial:
/close 

in favour of https://github.com/openshift/origin/pull/29015

PR #29012, opened 2024-08-16
Ensure correct deletion order for pods in e2e

tssurya:
cc @JacobTanenbaum @trozet


deads2k:
/hold

Can we get a deeper explanation for this on Monday?  This change suggests to me that standard namespace deletion will not function as expected once this feature is enabled.

tssurya:
@deads2k : we don't plan to merge this PR; this was a planB for another thing we were testing out (to fix a bug around deletion for a downstream merge => this was part of pre-merge testing as well). In the end this fix didn't help as well because we had a bug even with force delete of the pod, so closing this PR because we got our planA done.

PR #29024, opened 2024-08-20
[WIP] add LateConnection event to the apiserver shutdown intervals

tkashem:


pkg/monitortests/kubeapiserver/apiservergracefulrestart/monitortest.go
```diff
@@ -145,7 +161,12 @@ func podFrom(locator string) monitorapi.PodReference {
 }
 
 func (*apiserverGracefulShutdownAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
-	return nil, nil
+	testName := "[sig-arch] connection(s) to a kube-apiserver instance should not be created very late in the graceful termination process (more than 80% has passed)"
```
kannon92, line 164: would it make more sense to make this sig-apimachinery?

In k8s, I find sig-arch to encombassing a wide range of the project and not usually related to a single team. I'd name it to correspond to the team that would triage this if the test failed.

pkg/monitortests/kubeapiserver/apiservergracefulrestart/monitortest.go
```diff
@@ -161,7 +182,7 @@ func interesting(interval monitorapi.Interval) (monitorapi.IntervalReason, bool)
 	reason := interval.Message.Reason
 	switch reason {
 	// openshift-apiserver still is using the old event name TerminationStart
-	case "ShutdownInitiated", "TerminationStart", "TerminationGracefulTerminationFinished":
+	case "ShutdownInitiated", "TerminationStart", "TerminationGracefulTerminationFinished", "LateConnections":
```
p0lyn0mial, line 185: we already have a test that checks for "LateConnections" event, here: https://github.com/openshift/origin/blob/254f1e216526315abb05d9cdea025609f97f8bbd/test/extended/apiserver/graceful_termination.go#L132

tkashem: yes, thanks for pointing out, i will remove the junit test failed bits, just keep the part where the apiserver shutdown interval is marked red with an error message.

pkg/monitortests/kubeapiserver/apiservergracefulrestart/monitortest.go
tkashem: yeah, i will remove the just test bits, as @p0lyn0mial pointed out, we already have a test for it and it is marked as `sig-apimachinery`

tkashem:
/periodic-ci-openshift-release-master-nightly-4.17-e2e-metal-ipi-upgrade-ovn-ipv6

tkashem:
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-metal-ipi-upgrade-ovn-ipv6 10

PR #29026, opened 2024-08-21
NO-JIRA: monitor/recorder: snapshot makes copy before returning events

p0lyn0mial:
In go, passing slices between functions doesn't copy the internal array. 
Instead, a pointer to the same underlying array is shared among all receivers.


deads2k:
is this better than doing a copy?

/lgtm
/approve

p0lyn0mial:
/hold cancel

Either I was unlucky, or the intervals actually changed. I'm trying one more time

PR #29032, opened 2024-08-22
NO-JIRA: add basic graceful lease checking

deads2k:
It'll be fine....

Adding to identify when leases for operators are non-gracefully released by watching for acquisition time which should be milliseconds long.  When it's longer than 10s, mark failures.  These failures nearly always mean that for some reason the kube-apiserver was inaccessible **for minutes**.

### DIFF
--- a/pkg/defaultmonitortests/types.go
+++ b/pkg/defaultmonitortests/types.go
@@ -44,6 +44,7 @@ import (
 	"github.com/openshift/origin/pkg/monitortests/testframework/knownimagechecker"
 	"github.com/openshift/origin/pkg/monitortests/testframework/legacytestframeworkmonitortests"
 	"github.com/openshift/origin/pkg/monitortests/testframework/metricsendpointdown"
+	"github.com/openshift/origin/pkg/monitortests/testframework/operatorloganalyzer"
 	"github.com/openshift/origin/pkg/monitortests/testframework/pathologicaleventanalyzer"
 	"github.com/openshift/origin/pkg/monitortests/testframework/timelineserializer"
 	"github.com/openshift/origin/pkg/monitortests/testframework/trackedresourcesserializer"
@@ -180,6 +181,8 @@ func newUniversalMonitorTests(info monitortestframework.MonitorTestInitializatio
 	monitorTestRegistry.AddMonitorTestOrDie("e2e-test-analyzer", "Test Framework", e2etestanalyzer.NewAnalyzer())
 	monitorTestRegistry.AddMonitorTestOrDie("event-collector", "Test Framework", watchevents.NewEventWatcher())
 	monitorTestRegistry.AddMonitorTestOrDie("clusteroperator-collector", "Test Framework", watchclusteroperators.NewOperatorWatcher())
+	monitorTestRegistry.AddMonitorTestOrDie("initial-and-final-operator-log-scraper", "Test Framework", operatorloganalyzer.InitialAndFinalOperatorLogScraper())
+	monitorTestRegistry.AddMonitorTestOrDie("lease-checker", "Test Framework", operatorloganalyzer.OperatorLeaseCheck())
 
 	monitorTestRegistry.AddMonitorTestOrDie("azure-metrics-collector", "Test Framework", azuremetrics.NewAzureMetricsCollector())
 	monitorTestRegistry.AddMonitorTestOrDie("watch-request-counts-collector", "Test Framework", watchrequestcountscollector.NewWatchRequestCountSerializer())
--- a/pkg/monitor/monitor.go
+++ b/pkg/monitor/monitor.go
@@ -114,7 +114,8 @@ func (m *Monitor) Stop(ctx context.Context) (ResultState, error) {
 	m.junits = append(m.junits, computedJunit...)
 
 	fmt.Fprintf(os.Stderr, "Evaluating tests.\n")
-	finalEvents := m.recorder.Intervals(m.startTime, m.stopTime)
+	// compute intervals based on *all* the intervals.  Individual monitortests can choose how to restrict themselves.
+	finalEvents := m.recorder.Intervals(time.Time{}, time.Time{})
 	filename := fmt.Sprintf("events_used_for_junits_%s.json", m.startTime.UTC().Format("20060102-150405"))
 	if err := monitorserialization.EventsToFile(filepath.Join(m.storageDir, filename), finalEvents); err != nil {
 		fmt.Fprintf(os.Stderr, "error: Failed to junit event info: %v\n", err)
--- a/pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
+++ b/pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
@@ -1,0 +1,143 @@
+package operatorloganalyzer
+
+import (
+	"context"
+	"errors"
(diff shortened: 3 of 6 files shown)


pkg/monitor/monitor.go
```diff
@@ -114,7 +114,8 @@ func (m *Monitor) Stop(ctx context.Context) (ResultState, error) {
 	m.junits = append(m.junits, computedJunit...)
 
 	fmt.Fprintf(os.Stderr, "Evaluating tests.\n")
-	finalEvents := m.recorder.Intervals(m.startTime, m.stopTime)
+	// compute intervals based on *all* the intervals.  Individual monitortests can choose how to restrict themselves.
+	finalEvents := m.recorder.Intervals(time.Time{}, time.Time{})
```
deads2k, line 118: this is needed to allow a monitortest to fail for something that is broken during install.

pkg/defaultmonitortests/types.go
```diff
@@ -180,6 +181,8 @@ func newUniversalMonitorTests(info monitortestframework.MonitorTestInitializatio
 	monitorTestRegistry.AddMonitorTestOrDie("e2e-test-analyzer", "Test Framework", e2etestanalyzer.NewAnalyzer())
 	monitorTestRegistry.AddMonitorTestOrDie("event-collector", "Test Framework", watchevents.NewEventWatcher())
 	monitorTestRegistry.AddMonitorTestOrDie("clusteroperator-collector", "Test Framework", watchclusteroperators.NewOperatorWatcher())
+	monitorTestRegistry.AddMonitorTestOrDie("initial-and-final-operator-log-scraper", "Test Framework", operatorloganalyzer.InitialAndFinalOperatorLogScraper())
+	monitorTestRegistry.AddMonitorTestOrDie("lease-checker", "Test Framework", operatorloganalyzer.OperatorLeaseCheck())
```
p0lyn0mial, line 185: I'm wondering why this doesn't show up as an availability issue? After all, not being able to renew the lease is an availability issue

deads2k: > I'm wondering why this doesn't show up as an availability issue? After all, not being able to renew the lease is an availability issue

Availability is checked at a particular time. During install, openshift-tests is not running.  This check works by pulling the operator logs, so it can inspect failures from before openshift-tests started running.

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		return false
+	})
+
+	testNameToFailures := map[string][]string{}
+	for _, interval := range leaseIntervals {
+		ns := monitorapi.NamespaceFromLocator(interval.Locator)
+		testName := fmt.Sprintf("[sig-arch] all leases in ns/%s must gracefully release", ns)
+
+		intervalDuration := interval.To.Sub(interval.To)
```
p0lyn0mial, line 95: is this computation correct ? shouldn't it be ?:
```
intervalDuration := interval.To.Sub(interval.From)
```

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
```diff
@@ -0,0 +1,143 @@
...
+			case monitorapi.LeaseAcquired:
+				if lastAcquiringFrom == nil {
+					errs = append(errs, fmt.Errorf("missing acquiring stage for %v", podLocator))
+				} else {
+					ret = append(ret, monitorapi.NewInterval(monitorapi.SourcePodLog, monitorapi.Info).
+						Locator(interval.Locator).
+						Message(monitorapi.NewMessage().
+							Reason(monitorapi.LeaseAcquiring).
+							Constructed(monitorapi.ConstructionOwnerLeaseChecker).
+							HumanMessage("Waiting for lease."),
```
p0lyn0mial, line 67: This msg is misleading. The leas was acquired, no?

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
```diff
@@ -0,0 +1,143 @@
...
+		return false
+	})
+
+	testNameToFailures := map[string][]string{}
+	for _, interval := range leaseIntervals {
+		ns := monitorapi.NamespaceFromLocator(interval.Locator)
+		testName := fmt.Sprintf("[sig-arch] all leases in ns/%s must gracefully release", ns)
+
+		intervalDuration := interval.To.Sub(interval.To)
+		if intervalDuration < 10*time.Second {
```
p0lyn0mial, line 96: successful lease acquisition will always be < 10*time.Second, no?

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
```diff
@@ -0,0 +1,143 @@
...
+			switch interval.Message.Reason {
+			case monitorapi.LeaseAcquiringStarted:
+				// only overwrite if there isn't one already
+				if lastAcquiringFrom == nil {
+					lastAcquiringFrom = &interval.From
+				}
+
+			case monitorapi.LeaseAcquired:
+				if lastAcquiringFrom == nil {
+					errs = append(errs, fmt.Errorf("missing acquiring stage for %v", podLocator))
```
p0lyn0mial, line 60: are those errors going to fail the test?

pkg/monitortests/testframework/operatorloganalyzer/operator_log_scraper.go
```diff
@@ -0,0 +1,145 @@
...
+
+	errs := []error{}
+	for _, pod := range pods.Items {
+		if !strings.HasPrefix(pod.Namespace, "openshift-") {
+			continue
+		}
+		if !strings.Contains(pod.Name, "operator") {
+			continue
+		}
+		if pod.Status.Phase == corev1.PodPending || pod.Status.Phase == corev1.PodUnknown {
```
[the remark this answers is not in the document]
deads2k: this check is to ensure that the logs are present.  The pod doesn't have to be ready to have viable logs.

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
deads2k: > This msg is misleading. The leas was acquired, no?

This has a from and to that are the time that was spent waiting for the lease.

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
deads2k: heh, yeah, need to fix this.

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
deads2k: > are those errors going to fail the test?

These will, yes. But they also mean that somehow the beginning of the log was missing.

pkg/monitortests/testframework/operatorloganalyzer/operator_lease_check.go
p0lyn0mial: what about the `intervalDuration < 10*time.Second ` condition ?

deads2k:
no conflict with last.  merging.

PR #29037, opened 2024-08-27
OCPBUGS-39000: add node-exporter to the list of default SCCs

liouk:
The required-scc monitor test checks the list of default SCCs in order to determine whether the assigned SCC is custom or not; that list is missing the `node-exporter` SCC which means that the test will not suggest this to be pinned, as it thinks it is a custom one.


liouk:
We must stick to the kube-apiserver-operator's standard SCCs and avoid using node-exporter. This SCC comes from the monitoring operator and some clusters don't have monitoring, therefore it will be missing from those.

Closing.

/close

PR #29048, opened 2024-08-29
OCPBUGS-39232: Change Operator Installed for Smoke Test

dtfranz:
The cluster-logging operator is (currently) not available in the 4.17 catalog, so we need to use a different candidate for our smoke test in 4.17+ (at least for now).

I'm open to other suggestions for packages to use for this test.


dinhxuanvu (commented):
/lgtm

dinhxuanvu:
Branching is completed so make sure backporting this to 4.17.

PR #29056, opened 2024-08-30
NO-JIRA: add exclusions to restart tests

kannon92:
Exclusions were found by starting https://search.dptools.openshift.org/?search=container%2Fnetworking-console-plugin+restarted+.*+times+at%3A&maxAge=168h&context=1&type=junit&name=4.18&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job for 4.18.

We filed bugs to hopefully catch these exclusions.

The goal of this PR is to start making this test useful again so we catch bad restarts early. This job was flaking for a long time and many people ignored it.

By making the job fail, we hope to catch actual bugs in our product.

### DIFF
--- a/pkg/monitortests/node/legacynodemonitortests/exclusions.go
+++ b/pkg/monitortests/node/legacynodemonitortests/exclusions.go
@@ -1,0 +1,26 @@
+package legacynodemonitortests
+
+import "regexp"
+
+func isThisContainerRestartExcluded(locator string) bool {
+	// This is a list of known container restart failures
+	// Our goal is to conquer these restarts.
+	// So we are sadly putting these as exceptions.
+	// If you discover a container restarting more than 3 times, it is a bug and you should investigate it.
+	exceptions := []string{
+		"container/metal3-static-ip-set",      // https://issues.redhat.com/browse/OCPBUGS-39314
+		"container/ingress-operator",          // https://issues.redhat.com/browse/OCPBUGS-39315
+		"container/networking-console-plugin", // https://issues.redhat.com/browse/OCPBUGS-39316
+	}
+
+	for _, val := range exceptions {
+		matched, err := regexp.MatchString(val, locator)
+		if err != nil {
+			return false
+		}
+		if matched {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/monitortests/node/legacynodemonitortests/exclusions.go
+++ b/pkg/monitortests/node/legacynodemonitortests/exclusions.go
@@ -1,8 +1,18 @@
 package legacynodemonitortests
 
-import "regexp"
+import (
+	"regexp"
+	"strings"
 
-func isThisContainerRestartExcluded(locator string) bool {
+	"github.com/openshift/origin/pkg/monitortestlibrary/platformidentification"
+)
+
+type Exclusion struct {
+	upgradeJob  bool
+	clusterData platformidentification.ClusterData
+}
+
+func isThisContainerRestartExcluded(locator string, exclusion Exclusion) bool {
 	// This is a list of known container restart failures
 	// Our goal is to conquer these restarts.
 	// So we are sadly putting these as exceptions.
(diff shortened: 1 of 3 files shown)


pkg/monitortests/node/legacynodemonitortests/exclusions.go
```diff
@@ -0,0 +1,26 @@
+package legacynodemonitortests
+
+import "regexp"
+
+func isThisContainerRestartExcluded(locator string) bool {
```
deads2k, line 5: we have different dimensions to create the list of exclusions.

1. topology: single node, HA, hypershift
2. architecture: amd64, arm64, multiarch
3. lowest installed version: don't enforce on levels that have installed 4.17 or below

kannon92: I'm not sure we expose that information at the event level.

deads2k: https://github.com/openshift/origin/blob/5c601963b6e077c89d4d44c2a2e4bed38f2ddd77/pkg/monitortestlibrary/platformidentification/types.go#L59

might be a place to get thsi info.

pkg/monitortests/node/legacynodemonitortests/exclusions.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func isThisContainerRestartExcluded(locator string, exclusion Exclusion) bool {
+	// This is a list of known container restart failures
+	// Our goal is to conquer these restarts.
+	// So we are sadly putting these as exceptions.
+	// If you discover a container restarting more than 3 times, it is a bug and you should investigate it.
+	exceptions := []string{
+		"container/metal3-static-ip-set",      // https://issues.redhat.com/browse/OCPBUGS-39314
+		"container/ingress-operator",          // https://issues.redhat.com/browse/OCPBUGS-39315
```
deads2k, line 22: this only happens on single node clusters, so we should be sure that this exclusion is only valid for single node clsuters, not for HA clusters.

pkg/monitortests/node/legacynodemonitortests/exclusions.go
```diff
@@ -0,0 +1,47 @@
...
+	clusterData platformidentification.ClusterData
+}
+
+func isThisContainerRestartExcluded(locator string, exclusion Exclusion) bool {
+	// This is a list of known container restart failures
+	// Our goal is to conquer these restarts.
+	// So we are sadly putting these as exceptions.
+	// If you discover a container restarting more than 3 times, it is a bug and you should investigate it.
+	exceptions := []string{
+		"container/metal3-static-ip-set",      // https://issues.redhat.com/browse/OCPBUGS-39314
```
deads2k, line 21: this only happens on metal jobs.  If it happened vsphere jobs it woudl be a regression, we should restrict to metal.

pkg/monitortests/node/legacynodemonitortests/exclusions.go
```diff
@@ -0,0 +1,47 @@
...
+	// Upgrades seem to have a lot of failures.
+	// Let's exclude these for now generally.
+	if exclusion.upgradeJob {
+		return true
+	}
+	for _, val := range exclusion.clusterData.ClusterVersionHistory {
+		if strings.Contains(val, "4.17") {
+			return true
+		}
+	}
```
deads2k, line 35: this over of the upgrade job check.  this check is good because we are unlikely to backport fixes to 4.17, but we're working in 4.18 so any micro fix we should get improvement from.

pkg/monitortests/node/legacynodemonitortests/exclusions.go
```diff
@@ -0,0 +1,47 @@
...
+		"container/metal3-static-ip-set",      // https://issues.redhat.com/browse/OCPBUGS-39314
+		"container/ingress-operator",          // https://issues.redhat.com/browse/OCPBUGS-39315
+		"container/networking-console-plugin", // https://issues.redhat.com/browse/OCPBUGS-39316
+	}
+
+	// Upgrades seem to have a lot of failures.
+	// Let's exclude these for now generally.
+	if exclusion.upgradeJob {
+		return true
+	}
```
deads2k, line 30: this can probably be removed for the lower check.

pkg/monitortests/node/legacynodemonitortests/kubelet.go
```diff
@@ -245,11 +248,17 @@ func testContainerFailures(events monitorapi.Intervals) []*junitapi.JUnitTestCas
...
+	clusterDataPlatform, _ := platformidentification.BuildClusterData(context.Background(), adminRestConfig)
+
+	exclusions := Exclusion{upgradeJob: isUpgrade, clusterData: clusterDataPlatform}
 	for locator, messages := range containerExits {
 		if len(messages) > 0 {
 			messageSet := sets.NewString(messages...)
-			// Blanket fail for restarts over 3
-			if len(messages) > maxRestartCount {
+			// Blanket fail for restarts over maxRestartCount
+			if !isThisContainerRestartExcluded(locator, exclusions) && len(messages) > maxRestartCount {
```
deads2k, line 261: How about we start failing (rather than flaking) for maxRestartCount=3.  And flake a *different* test name for maxRestartCount=2.

PR #29041, opened 2024-08-27
SDN-5202:Porting testing for netConfig and UDN and CIDR overlapping tests from upstream

JacobTanenbaum:
This PR ports using a correct primary network configuration for all user defined networking tests, using both netConfig and UDN crd for all currently ported tests, and overlapping CIDR tests from upstream ovn-kubernetes

PR #29060, opened 2024-09-03
[wip] Udn upstream testing 

JacobTanenbaum:


tssurya:
@JacobTanenbaum what about the test in https://github.com/ovn-org/ovn-kubernetes/pull/4589
Is that being brought down OR do you want to wait for https://github.com/ovn-org/ovn-kubernetes/pull/4625 to also be done so that those overlapping tests are also being run with CRDs?

JacobTanenbaum:
added the testing from https://github.com/ovn-org/ovn-kubernetes/pull/4589 to https://github.com/openshift/origin/pull/29041. I am trying to have this PR only be the changes I have to upstream that I have in https://github.com/ovn-org/ovn-kubernetes/pull/4625 I think that would make it so there is less review and because 4625 is taking some time to get merged

PR #29066, opened 2024-09-05
NO-JIRA: add termination-log-analyzer

sanchezl:


pkg/monitortests/kubeapiserver/terminationlog/logs.go
```diff
@@ -0,0 +1,141 @@
...
+
+type ClusterTerminationLogs struct {
+	errs []error
+}
+
+func NewClusterTerminationLogs(ctx context.Context, client kubernetes.Interface) *ClusterTerminationLogs {
+	return &ClusterTerminationLogs{}
+}
+
+func (p *ClusterTerminationLogs) Entries(ctx context.Context, client kubernetes.Interface) <-chan *LogEntry {
```
p0lyn0mial, line 26: I'm wondering if we could slightly simplify this function to something like:

```
func (p *ClusterTerminationLogs) Process(logHandler LogHandler) {
   masterNodes := getMasterNodes()
   for _, masterNode := range masterNodes {
     nodeTerminationLog := NewNodeTerminationLogs(masterNode)
     rawFile := nodeTerminationLog.Read()
     reader := bytes.NewReader(data) 
     scanner := bufio.NewScanner(reader) 
     for scanner.Scan() { 
        logHandler(scanner.Text()) 
     }
   }
}
```

WDYT ?

p0lyn0mial: We could even preprocess the line before calling the logHandler to extract some data like timestamp, for example:
```
for scanner.Scan() {
  line := scanner.Text()
  timestamp := extractTimestampFrom(line)
  structuredLogLine := {timestamp: timestamp, locator: nodeName/podName (?)
  logHandler(structuredLogLine)
}
```

pkg/monitortests/kubeapiserver/terminationlog/monitor.go
```diff
@@ -0,0 +1,97 @@
...
+	return nil
+}
+
+func (c *clusterMonitor) CollectData(ctx context.Context, storageDir string, beginning, end time.Time) (monitorapi.Intervals, []*junitapi.JUnitTestCase, error) {
+	client, err := kubernetes.NewForConfig(c.config)
+	if err != nil {
+		return nil, nil, err
+	}
+	var intvs []monitorapi.Interval
+	terminationLogs := NewClusterTerminationLogs(ctx, client)
```
p0lyn0mial, line 39: and here we would just do:
```
ClusterTerminationLogs.Process(func(line structuredLogLine {

// process the line

})) 
```

pkg/monitortests/kubeapiserver/terminationlog/monitor.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (c *clusterMonitor) ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals, recordedResources monitorapi.ResourcesMap, beginning, end time.Time) (monitorapi.Intervals, error) {
+	//TODO implement me
+	return nil, nil
+}
+
+func (c *clusterMonitor) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
+	//TODO implement me
+	//TODO are there any useful tests we could run?
```
deads2k, line 71: if you have any late arriving requests, fail the test.

pkg/monitortests/kubeapiserver/terminationlog/monitor.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		//  ip-10-0-117-206.ec2.internal-termination.log:
+		// W0830 19:35:46.150968      14 patch_genericapiserver.go:201]
+		// Request to "/api/v1/pods" (source IP 10.0.118.127:64460, user agent "ip-10-0-61-3/ovnkube@90d145b1785d (linux/amd64)
+		//    kubernetes/v0.30.2") through a connection created very late in the graceful termination process (more than 80% has
+		//    passed), possibly a sign for a broken load balancer setup.
+		if strings.Contains(logEntry.Msg, "possibly a sign for a broken load balancer setup") {
+			intv := monitorapi.NewInterval(monitorapi.SourceTerminationLog, monitorapi.Error).
+				// Locator().
+				Message(monitorapi.NewMessage().Cause(logEntry.Msg)).
```
deads2k, line 52: you want a reason, the reason should be something like late arriving request via LB

deads2k, line 52: you want an annotation for the user this request is from and another for the resource being accessed.

deads2k, line 52: human message containing the user an resource

deads2k: annotation for audit ID if you ahve one.

pkg/monitortests/kubeapiserver/terminationlog/logs.go
p0lyn0mial: This method could actually accept an array of `logHandler LogHandler` so that we could register multiple logHandlers. Each handler could produce an interval. WDYT?

PR #29077, opened 2024-09-09
OCPBUGS-41550: fix: add check for image registry capability

eggfoobar:
When the image registry capability is disabled, service account pull secret and role bindings are not going to be populated. We need to skip waiting for those events when creating namespaces.

### DIFF
--- a/pkg/monitortests/network/disruptionpodnetwork/monitortest.go
+++ b/pkg/monitortests/network/disruptionpodnetwork/monitortest.go
@@ -27,6 +27,8 @@ import (
 	"k8s.io/klog/v2"
 	k8simage "k8s.io/kubernetes/test/utils/image"
 
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/origin/pkg/monitor/monitorapi"
 	monitorserialization "github.com/openshift/origin/pkg/monitor/serialization"
 	"github.com/openshift/origin/pkg/monitortestframework"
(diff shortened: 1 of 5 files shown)


test/extended/util/client.go
```diff
@@ -356,13 +356,25 @@ func (c *CLI) setupProject() string {
...
+			defaultRoleBindings = append(defaultRoleBindings, "system:deployers")
+		}
+	} else {
+		DefaultServiceAccounts = []string{}
+		defaultRoleBindings = []string{}
 	}
+
 	for _, sa := range DefaultServiceAccounts {
 		framework.Logf("Waiting for ServiceAccount %q to be provisioned...", sa)
 		err = WaitForServiceAccountWithSecret(c.KubeClient().CoreV1().ServiceAccounts(newNamespace), sa)
```
deads2k, line 380: seems better to update `WaitForServiceAccountWithSecret` to handle the case when no image pull secrets will be created.

pkg/monitortests/network/disruptionpodnetwork/monitortest.go
```diff
@@ -126,12 +128,20 @@ func (pna *podNetworkAvalibility) StartCollection(ctx context.Context, adminREST
 		return err
 	}
 
+	irEnabled, err := imageRegistryEnabled(adminRESTConfig)
```
deads2k, line 131: `imageRegistryEnabled`.  don't be stingy.

deads2k, line 131: This determination is actually made by checking for endpoints I thik.  Check to see how the openshift-controller-manager-operator does it.  I don't think it's the capability.

eggfoobar: lol I was half in half out on the ʕ◔ϖ◔ʔ mentality, updating

test/extended/util/client.go
eggfoobar: That's a great point, updating it. However, I think we still gotta do the check here since there a role binding check after this that will also not work when image registry isn't present. At least from what I could find, that should only be populated when the image-registry operator is enabled via this [controller code](https://github.com/openshift/openshift-controller-manager/blob/master/pkg/cmd/controller/authorization.go#L41) that looks like it's [conditional based on this](https://github.com/openshift/openshift-controller-manager/blob/master/pkg/cmd/openshift-controller-manager/controller_manager.go#L141)

pkg/monitortests/network/disruptionpodnetwork/monitortest.go
eggfoobar: It does seem like we do use the capabilities, at least in this [internal image registry controller](https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/internalimageregistry/management_state.go#L12) for openshift-controller-manager-oprator

PR #29081, opened 2024-09-10
AUTH-539: Expose image to unauthenticated test

ShazaAldawamneh:
Create a test that exposes an internal image in OpenShift to unauthenticated users.


test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1243,67 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
 		})
 	})
 })
+
+var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("bootstrap-policy")
+
+	g.Context("", func() {
+		g.Describe("ROSA-OSD_CCS-ARO-Author:jitli-LEVEL0-Critical-48959-Should be able to get public images connect to the server and have basic auth credentials", func() {
```
ibihim, line 1252: > ROSA-OSD_CCS-ARO-Author:jitli-LEVEL0-Critical-48959

What does that mean? Can we drop it?

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1243,67 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
 		})
 	})
 })
+
+var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", func() {
```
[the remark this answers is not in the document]
ibihim: I think that the tags are correct like `[sig-auth]` and `[Feature:OpenShiftAuthorization]`, but I don't think that plain `authorization` is accurate. We are not testing `authorization` in itself, we are testing that a particular feature works, based on the current authorization.

I would assume that something like `[sig-auth][Feature:OpenShiftAuthorization] ImageRegistry access` would describe it better.

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1243,67 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+	regRoute, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("route", route, "-n", ns, "-o=jsonpath={.spec.host}").Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return regRoute
+}
+
+func waitRouteReady(route string) {
+	curlCmd := "curl -k https://" + route
+	var output []byte
+	var curlErr error
+	pollErr := wait.Poll(5*time.Second, 1*time.Minute, func() (bool, error) {
```
ibihim, line 1297: `wait.Poll` seems to be deprecated, could you take an appropriate wait function that is not deprecated?

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1243,67 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+	oc := exutil.NewCLI("bootstrap-policy")
+
+	g.Context("", func() {
+		g.Describe("PublicImageAccessWithBasicAuthShouldSucceed", func() {
+			g.It("should succeed [apigroup:image.openshift.io]", func() {
+				// Skip Hypershift external OIDC clusters
+				isExternalOIDCCluster, err := exutil.IsExternalOIDCCluster(oc)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				if isExternalOIDCCluster {
+					g.Skip("Skipping the test as we are running against a Hypershift external OIDC cluster")
```
ibihim, line 1258: Do we need this skip?

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1243,67 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+
+func exposeRouteFromSVC(oc *exutil.CLI, rType, ns, route, service string) string {
+	err := oc.AsAdmin().WithoutNamespace().Run("create").Args("route", rType, route, "--service="+service, "-n", ns).Execute()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	regRoute, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("route", route, "-n", ns, "-o=jsonpath={.spec.host}").Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return regRoute
+}
+
+func waitRouteReady(route string) {
```
ibihim, line 1293: Did you check if there are similar functions out there, that we could reuse?

ibihim (commented):
Great that you are adding a test for this!

test/extended/authorization/authorization.go
ShazaAldawamneh: In my test, I'm allowing unauthenticated users to access public images, but external OIDC clusters could interfere with this by enforcing authentication, which may block unauthenticated access and cause the test to fail for reasons unrelated to the functionality being tested. The g.Skip check is necessary to prevent running the test in such environments, ensuring it only runs where the authentication model aligns with the test's requirements. Without this skip, the test could become unreliable in external OIDC clusters, where authentication rules differ.

this is my understanding of the external OIDC clusters, if this is not the case please let me know.

test/extended/authorization/authorization.go
ShazaAldawamneh: I did, couldn't find something similar.

vrutkovs: Lets not call external binaries here:
```
	tr := utilnet.SetTransportDefaults(&http.Transport{
		TLSClientConfig:   &tls.Config{InsecureSkipVerify: true},
	})
	client := &http.Client{
		Transport: tr,
		Timeout: time.Minute,
	}
	resp, err := client.Get(fmt.Sprintf("https://%s",route))
	o.Expect(resp.StatusCode).Should(o.Equal(http.StatusOK))
```

vrutkovs: Oh, we had this already? Okay, lets leave it then

test/extended/authorization/authorization.go
ibihim: How does a cluster with an external OIDC IDP interfere with us testing the access for unauthenticated users?

ibihim: I mean it is not important to test this on an external OIDC cluster, if it is only HyperShift. But in a world, where most of our customers would use it,... it would be quite nice to test it there too. :)

neisw:
Noticed failures in [single-node-e2e-test/build-log.txt](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/29081/pull-ci-openshift-origin-master-e2e-aws-ovn-single-node-upgrade/1845753933364989952/artifacts/e2e-aws-ovn-single-node-upgrade/single-node-e2e-test/build-log.txt)

```
  Summarizing 1 Failure:
    [FAIL] [sig-auth][Feature:OpenShiftAuthorization] ImageRegistry access  PublicImageAccessWithBasicAuthShouldSucceed [It] should succeed [apigroup:image.openshift.io] [Suite:openshift/conformance/parallel]
    github.com/openshift/origin/test/extended/authorization/authorization.go:1270

```

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1242,93 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+func isImageRegistryAvailable(oc *exutil.CLI) bool {
+    output, err := oc.AsAdmin().Run("get").Args(
+        "co", "image-registry",
+        "-o=jsonpath={.status.conditions[?(@.type=='Available')].status}",
+    ).Output()
+    o.Expect(err).NotTo(o.HaveOccurred())
+    return strings.TrimSpace(output) == "True"
+}
+
+func exposeRouteFromSVC(oc *exutil.CLI, rType, ns, route, service string) string {
```
flavianmissi, line 1306: Looks like this function is no longer needed?

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1242,93 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+        })
+    })
+})
+
+func isImageRegistryAvailable(oc *exutil.CLI) bool {
+    output, err := oc.AsAdmin().Run("get").Args(
+        "co", "image-registry",
+        "-o=jsonpath={.status.conditions[?(@.type=='Available')].status}",
+    ).Output()
+    o.Expect(err).NotTo(o.HaveOccurred())
```
flavianmissi, line 1302: TL;DR: have a look at this function: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/internalimageregistry/management_state.go#L12 - I think you might want to copy/paste it here or at least get some inspiration for how to do this check.

---

You might need to adapt this to work in the following situations:

1. when running in a cluster where the image registry operator is installed, but `.spec.managementState` is set to `Removed`: here I'm uncertain if the Available condition will be set with status true... even if the Available condition would be set to true, it would be misleading for this function to return true, since the image registry would be Removed.
2. when running in a cluster with the `ImageRegistry` capability disabled: in this case, this assertion will definitely end up in an error - I suspect the image registry Cluster Operator isn't installed at all in a cluster where the ImageRegistry capability is disabled.

`.spec.managementState` will be set to `Removed` automatically by the image registry operator when its running in a cluster without object storage available, like metal or vsphere installations for example. If the image registry operator is Removed then you'll want to skip your tests. You can detect this by looking for a Removed condition.

In both these cases you probably want your function to return `false` without it being treated as an error.
You probably want to treat both these cases as correct behavior (instead of an error), and return `false` since the registry will not be available.

[This function](https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/internalimageregistry/management_state.go#L12) implements this check, you might want to use it as inspiration or even copy/paste here and use it instead.

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1242,93 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+                g.By("Enable default route for the image registry")
+                err := oc.AsAdmin().WithoutNamespace().Run("patch").Args(
+                    "configs.imageregistry.operator.openshift.io/cluster",
+                    "--type=merge",
+                    "-patch", `{"spec": {"defaultRoute": true}}`,
+                ).Execute()
+                o.Expect(err).NotTo(o.HaveOccurred())
+
+                g.By("Wait for the image registry to become available")
+                available := isImageRegistryAvailable(oc)
```
flavianmissi, line 1262: You probably want to run this check before attempting to enable the default route, i.e as the first thing you do in your tests, then skip the tests if it's not _present_, and perhaps adding a wait/retry if it's progressing.

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1242,93 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+        g.Describe("PublicImageAccessWithBasicAuthShouldSucceed", func() {
+            g.It("should succeed [apigroup:image.openshift.io]", func() {
+
+                g.By("Enable default route for the image registry")
+                err := oc.AsAdmin().WithoutNamespace().Run("patch").Args(
+                    "configs.imageregistry.operator.openshift.io/cluster",
+                    "--type=merge",
+                    "-patch", `{"spec": {"defaultRoute": true}}`,
+                ).Execute()
+                o.Expect(err).NotTo(o.HaveOccurred())
```
flavianmissi, line 1259: Consider using `waitRouteReady` here to make sure you only continue your tests once the route is fully provisioned. Sometimes provisioning a route can take a second or two, and you might end up with some flake runs of this if you don't wait.

The default route's name is `default-route`.

test/extended/authorization/authorization.go
```diff
@@ -1241,3 +1242,93 @@ var _ = g.Describe("[sig-auth][Feature:OpenShiftAuthorization] authorization", f
...
+
+func exposeRouteFromSVC(oc *exutil.CLI, rType, ns, route, service string) string {
+	err := oc.AsAdmin().WithoutNamespace().Run("create").Args("route", rType, route, "--service="+service, "-n", ns).Execute()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	regRoute, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("route", route, "-n", ns, "-o=jsonpath={.spec.host}").Output()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	return regRoute
+}
+
+func waitRouteReady(route string) {
```
flavianmissi, line 1314: Consider bringing [this code](https://github.com/openshift/cluster-image-registry-operator/blob/master/test/framework/route.go?plain=1#L17-L44) from the registry operator's own tests into this function - you'd need to adapt it slightly but it does what you want by watching the route object itself, which is perhaps slightly more elegant 😅

flavianmissi (commented):
Great work so far! This will work well for classic OCP installs. Left a few comments that should cover most special cases.

Please let me know if you have comments or questions. OpenShift is very flexible and that adds a lot of complexity, no one is expected to keep that all in their brain at all times 😅

flavianmissi (commented):
Thanks for implementing the suggested changes - LGTM if the tests pass 🙌🏼

test/extended/util/clusterversion.go
```diff
@@ -4,13 +4,18 @@ import (
...
+	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 
+	g "github.com/onsi/ginkgo/v2"
```
vrutkovs, line 17:
```suggestion
	e2e "k8s.io/kubernetes/test/e2e/framework"

	g "github.com/onsi/ginkgo/v2"
```
`e2e.Logf` is used below

PR #29083, opened 2024-09-10
OCPBUGS-41270: Update the NotFound case for CNI plugin to reflect changes

deads2k:
https://github.com/openshift/multus-cni/commit/a95c1f46#diff-2a0a19881ba102c36dc17295504ba6fb9f4a0fd42f54410323014e6fdeb7d1b1L535 changed the string we need to search for to differently categorize this problem


neisw:
/retitle NO-JIRA: Update the NotFound case for CNI plugin to reflect changes

PR #29087, opened 2024-09-11
NO-JIRA: improve excessive restart tests

kannon92:
1) Add better exclusions to catch more restarts.
2) add a flake test for 1 or more restarts.
3) add ns/%v to the test output to quickly identity the team responsible.

### DIFF
--- a/pkg/monitortests/node/legacynodemonitortests/exclusions.go
+++ b/pkg/monitortests/node/legacynodemonitortests/exclusions.go
@@ -8,7 +8,6 @@ import (
 )
 
 type Exclusion struct {
-	upgradeJob  bool
 	clusterData platformidentification.ClusterData
 }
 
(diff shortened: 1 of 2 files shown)


pkg/monitortests/node/legacynodemonitortests/exclusions.go
```diff
@@ -17,30 +16,54 @@ func isThisContainerRestartExcluded(locator string, exclusion Exclusion) bool {
...
 			return false
 		}
 		if matched {
-			return true
+			// if container matches but platform is different, this is a regression.
+			if val.platformsToExclude != "" && val.platformsToExclude != exclusion.clusterData.Platform {
+				return true
+			}
+			// if container matches but topology is different, this is a regression.
+			if val.topology != "" && val.topology != exclusion.clusterData.Topology {
```
deads2k, line 60: This check looks backwards.

how about
```go

switch{
case val.platformsToExclude != "" && val.platformsToExclude == exclusion.clusterData.Platform:
  return false
case val.topology != "" && val.topology != exclusion.clusterData.Topology:
  return false
default:
  return true
}

```

pkg/monitortests/node/legacynodemonitortests/kubelet.go
```diff
@@ -238,64 +242,105 @@ func testContainerFailures(adminRestConfig *rest.Config, events monitorapi.Inter
...
+				SystemOut: strings.Join(excessiveExits, "\n"),
+				FailureOutput: &junitapi.FailureOutput{
+					Output: fmt.Sprintf("%d containers with multiple restarts\n\n%s", len(excessiveExits), strings.Join(excessiveExits, "\n\n")),
+				},
+			})
+		}
 	}
+	for _, namespace := range sets.List(openshiftNamespaces) { // this ensures we create test case for every namespace, even in success cases
+		excessiveExits := excessiveExitsByNamespaceForFlakeTests[namespace]
+		excessiveRestartTestName := fmt.Sprintf("[sig-architecture] platform pods in ns/%s should not exit more than %d with a non-zero exit code", namespace, maxRestartCountForFlakes)
```
deads2k, line 331: need a different name for this test.

pkg/monitortests/node/legacynodemonitortests/exclusions.go
```diff
@@ -17,30 +16,52 @@ func isThisContainerRestartExcluded(locator string, exclusion Exclusion) bool {
...
+			// We did not find failures for vsphere where this is also run
+			// So if we start seeing failures on vsphere this would be a regression.
+			containerName:      "container/metal3-static-ip-set", // https://issues.redhat.com/browse/OCPBUGS-39314
+			platformsToExclude: "metal",
+		},
+		{
+			// ingress operator seems to only fail on the single topology.
+			// platform did not matter.
+			containerName:     "container/ingress-operator", // https://issues.redhat.com/browse/OCPBUGS-39315
+			topologyToExclude: "single",
```
deads2k, line 36: This one is a topology to *include* isn't it?

kannon92: I was thinking no. Based on what you suggested, we are seeing falures for ingress-operator single so I want to exclude single. 

I figured having one field where it is exclude and other than is include would cause me to cry in the long run so I decided against that.

pkg/monitortests/node/legacynodemonitortests/kubelet.go
kannon92: updated

pkg/monitortests/node/legacynodemonitortests/exclusions.go
kannon92: updated.

kannon92:
@deads2k can you give this a lgtm when you get a chance?

PR #29089, opened 2024-09-11
MON-4010: metrics API disruption check: do not fall back on prometheus-adapter as metrics-server is GA since 4.16

machine424:


simonpasquier (commented):
/lgtm

machine424:
the failures don't seem to be related, but
/retest

PR #29090, opened 2024-09-11
API-1789: tls artifacts: list ondisk locations alongside secret/configmaps

vrutkovs:
TLS artifact reports prefer in-cluster locations, so that they can be assigned metadata via annotations. This change prints other ondisk locations so that secret/configmap could be linked to the file on disk


sanchezl (approved):
/lgtm

tls/ownership/ownership.md
```diff
@@ -1409,15 +1945,33 @@
...
 
       **Description:** Service CA configmap contains the data for the PEM-encoded CA signing bundle which will be injected to resources annotated with 'service.beta.openshift.io/inject-cabundle=true'
       
 
+      Other locations:
+
+      * file /etc/docker/certs.d/image-registry.openshift-image-registry.svc.cluster.local:5000/ca.crt
+      * file /etc/docker/certs.d/image-registry.openshift-image-registry.svc:5000/ca.crt
+      
+
```
deads2k, line 1964: I bet we'll end up finding disputes of "who actually owns this ca bundle" because while the content matches, the actual owner seems more likely to be the image registry.  This is likely a common problem and suggests that even though the content is the same, the thing we're actually providing ownership for is the file itself.

vrutkovs:
/hold

Alongside "this matches the content" we should also match by purpose - i.e. /etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/trusted-ca-bundle/ca-bundle.crt belongs to kube-apiserver and so on

tls/ownership/ownership.md
```diff
@@ -98,6 +100,35 @@
...
+      Other locations:
+
+      * file /etc/docker/certs.d/virthost.ostest.test.metalkube.org:5000/ca.crt
+      
+
+
+
+## Image Registry (2)
+### Certificate Authority Bundles (2)
+1. file /etc/docker/certs.d/image-registry.openshift-image-registry.svc.cluster.local:5000/ca.crt
```
vrutkovs, line 112: These files both have the same owner and the same content, so they are printed twice here. Not sure if its worth fixing tbh

dinhxuanvu (commented):
/lgtm

tls/ownership/ownership.md
deads2k: > These files both have the same owner and the same content, so they are printed twice here. Not sure if its worth fixing tbh

Listing every file, including those with duplicate content, is the proper thing to do.  Think about when descriptions are added and those descriptions indicate which binaries and which flags for those binaries the file is used for.  We need those descriptions to be unique.  Even if this specific case doesn't need it, the general case does.

vrutkovs:
/hold cancel

gcp-builds is now passing

PR #29082, opened 2024-09-10
OCPBUGS-39373: Use a supported Perl version in the test template.

metlos:
This should fix the CI tests for the Perl image stream.

PR #29096, opened 2024-09-12
OCPBUGS-39373: Redo the PR #29082

metlos:
Turns out the fix in PR #29082 was not correct and wouldn't have been sufficient even if it were.

This PR fixes the Perl test and updates it to work with the new `perl:5.32-ubi8` image that has different defaults and settings available (notice the disappearance of `PERL_APACHE2_RELOAD` env var setting in https://github.com/sclorg/s2i-perl-container/tree/master/5.32 vs https://github.com/sclorg/s2i-perl-container/tree/master/5.30).

### DIFF
--- a/test/extended/image_ecosystem/s2i_perl.go
+++ b/test/extended/image_ecosystem/s2i_perl.go
@@ -3,7 +3,7 @@ package image_ecosystem
 import (
 	"context"
 	"fmt"
-	"strings"
+	"os"
 	"time"
 
 	g "github.com/onsi/ginkgo/v2"
@@ -11,7 +11,6 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/util/wait"
 	e2e "k8s.io/kubernetes/test/e2e/framework"
 
 	exutil "github.com/openshift/origin/test/extended/util"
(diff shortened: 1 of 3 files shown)


metlos:
Note that e2e-gcp-ovn-image-ecosystem actually should still fail, but it should not fail because of the perl tests, which is what we are fixing here.

adambkaplan (changes requested):
Please squash your commits, and ensure the commit has a well formatted title and description. Include the reasoning for deleting the hotfix portion of the test.

Is there someone from the RHEL team who can verify that this test is a sufficient "smoke test" for the perl s2i image?

metlos:
@phracek, would you be able to comment on whether the smoke test that we perform is sufficient? Or if not, would you be able to recommend someone that could?

Previously, our test built a custom image from the `perl:5.30-el7` image and a custom app that opens some http endpoint and we checked that it returns the expected data in the http response. Then we also tested that hot-deploy is off by default by trying modifying the filesystem directly in the container and expecting the http response doesn't change.

Afterwards we restarted the image with `PERL_APACHE2_RELOAD=true` and repeated the above procedure, now ensuring that hot deploy is switched on and directly modifying the filesystem of
the container causes Perl to hot-reload the app and return a new http response.

Because `perl:5.30-el7` has been removed, we need to upgrade to the next "oldest" image available that I believe is `perl:5.32-ubi8`.

By looking at https://github.com/sclorg/s2i-perl-container/tree/master/5.32 I found out that the env var to switch hot-deploy on has disappeared from the image and that hot deploy is on
by default.

Therefore I merely removed the part of the test that tests hot-reload switched off.

In your opinion, is testing the perl image using a custom webapp and checking that it a) returns expected data as an http response and b) that hot-deploy works in the image, enough as a smoke test?

metlos:
I'm closing this in favor of https://github.com/openshift/origin/pull/29116.

PR #29097, opened 2024-09-12
OCPBUGS-39363: Reenable Rails tests and update the sample repos test to support deployments

metlos:
After a change of decision about the support for Rails, we need to reenable it in the Samples Operator and therefore also in the integration tests here.

In addition to re-enabling the dedicated ruby test, we also need to support this in the samples_repo tests. Because the rails template now uses Deployments instead of DeploymentConfigs, it was also needed to add support for that in the tests.


metlos:
Note that the `e2e-gcp-ovn-image-ecosystem` test should fail because of a problem in perl template (addressed https://github.com/openshift/origin/pull/29096) but the tests for Rails should both pass.

metlos:
I'm closing this in favor of https://github.com/openshift/origin/pull/29116.

PR #29104, opened 2024-09-13
NO-JIRA: images.sh: Do not rely on the tag order

ardaguclu:
ImageStreams object's tag order seems to be changed and this test fully relies on the first object of the tag list and it is perma-failing https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/1877/pull-ci-openshift-oc-master-e2e-agnostic-ovn-cmd/1834521049262198784. This PR modifies the test by searching for the perl 5.32 image in the list instead of checking the first item.


ardaguclu:
This revert https://github.com/openshift/origin/pull/29105 should fix it;
/close

PR #29110, opened 2024-09-17
NO-JIRA: reduce node lease back off flames to 2

kannon92:


pkg/monitortests/node/kubeletlogcollector/monitortest.go
```diff
@@ -71,8 +71,7 @@ func nodeFailedLeaseErrors(startedAt time.Time, finalIntervals monitorapi.Interv
 	const testName = "[sig-node] kubelet-log-collector detects node failed to lease events"
 	var failures []string
 
-	intervalsToFailOn := findLeaseIntervalsImportant(finalIntervals)
```
deads2k, line 74: need to filter this down to the type you're looking for, right?

pkg/monitortests/node/kubeletlogcollector/node.go
```diff
@@ -526,49 +526,6 @@ func leaseFailBackOff(nodeLocator monitorapi.Locator, logLine string) monitorapi
...
-		// We will assume that intervals that have the same node and occur at the same time
-		// are duplicated events.
-		previousInterval := intervalsByNode[nodeName][len(intervalsByNode[nodeName])-1]
-		if val.From != previousInterval.From && val.To != previousInterval.To {
-			intervalsByNode[nodeName] = append(intervalsByNode[nodeName], val)
-		}
-	}
-	importantIntervals := monitorapi.Intervals(nil)
-	for _, val := range intervalsByNode {
-		if len(val) < 3 {
```
deads2k, line 557: what if we reduce this to 2 so we flake if we get two instances and see how many hits we get?

kannon92: updated.

kannon92:
/retitle NO-JIRA: reduce node lease back off flames to 2

pkg/monitortests/node/kubeletlogcollector/node.go
```diff
@@ -554,7 +554,7 @@ func findLeaseIntervalsImportant(intervals monitorapi.Intervals) monitorapi.Inte
 	}
 	importantIntervals := monitorapi.Intervals(nil)
 	for _, val := range intervalsByNode {
-		if len(val) < 3 {
+		if len(val) < 2 {
 			continue
 		}
 		for i, interval := range val {
```
deads2k, line 560: in this for loop the `2`s need to become 1s

PR #29114, opened 2024-09-18
Omit etcds resource for HyperShift clusters

Evan-Reilly:
Related: https://github.com/openshift/hypershift/pull/3807#issuecomment-2341551844

### DIFF
--- a/test/extended/cli/explain.go
+++ b/test/extended/cli/explain.go
@@ -498,11 +498,36 @@ var (
 	}
 )
 
+func remove(slice []schema.GroupVersionResource, s int) []schema.GroupVersionResource {
+	if s > 0 || s > len(slice)-1 {
+		return slice
+	}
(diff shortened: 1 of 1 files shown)


test/extended/cli/explain.go
```diff
@@ -498,12 +498,37 @@ var (
...
+			if baseCRDTypes[i].Version != "v1" {
+				continue
+			}
+			if baseCRDTypes[i].Resource != "etcds" {
+				continue
+			}
+			etcdsResourceIndex = i
+			break
+		}
+		remove(baseCRDTypes, etcdsResourceIndex)
```
rtheis, line 530: Based on how the rest of the code is structured, it seems like etcd should be removed from the baseCRDTypes and to its own type that is added for non-HyperShift.  Although this is a fairly significant refactor and I'd like Red Hat's opinion before asking you to make this change.

PR #29115, opened 2024-09-18
NO-JIRA: add test that ensures our informer cache matches the live results

deads2k:
Trying to figure out why pods that were deleted in job runs like https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.18-upgrade-from-stable-4.17-e2e-metal-ipi-ovn-upgrade-cgroupsv1/1832612917804011520 (we very rarely catch the condition because pod IPs have to collide) were still present in our caches.

Reminder to self that some analysis is in https://gist.github.com/deads2k/5f9b725894bdb7ba03011889c06b2127

In this case
1. pods/pod-network-to-service-disruption-poller-565f9f69c4-nqjgl was deleted 05:18:52 (see audit log for delete and a get that 404'd)
2. pod-network-to-host-network-disruption-poller-794dd9846-9sc9m was created at 05:18:52
3. at 2024-09-08T05:18:59, the lister in the pod_ip_controller returned a list that contained *both* 565f9f69c4-nqjgl and 794dd9846-9sc9m with the same IP

even if the watch was latent, the delete (lower resourceversion) should have happened before the create and pod IP assignment.

@trozet ptal

### DIFF
--- a/pkg/monitortests/node/watchpods/cache_check.go
+++ b/pkg/monitortests/node/watchpods/cache_check.go
@@ -1,0 +1,99 @@
+package watchpods
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	coreinformers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func checkCacheState(ctx context.Context, kubeClient kubernetes.Interface, podInformer coreinformers.PodInformer) ([]string, error) {
+	// We are running into problems where the cached list of pods contains pods that don't exist in the kube-apiserver.
+	// Let's try to check exactly.
+	cachedPods, err := podInformer.Lister().List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error listing cached pods: %w", err)
+	}
+
+	return checkCacheStateFromList(ctx, kubeClient, cachedPods)
+}
+
+func checkCacheStateFromList(ctx context.Context, kubeClient kubernetes.Interface, cachedPods []*corev1.Pod) ([]string, error) {
+	if len(cachedPods) == 0 {
+		return nil, fmt.Errorf("somehow the lister is missing pods")
+	}
+
+	biggestResourceVersion := -1
+	for _, cachedPod := range cachedPods {
+		if len(cachedPod.ResourceVersion) == 0 {
+			return nil, fmt.Errorf("encounted cached pod/%s -n %s without a resourceversion", cachedPod.Namespace, cachedPod.Name)
+		}
+		currRV, err := strconv.Atoi(cachedPod.ResourceVersion)
+		if err != nil {
+			return nil, fmt.Errorf("cached pod/%s -n %s has a non-integer RV %q: %w", cachedPod.Namespace, cachedPod.Name, cachedPod.ResourceVersion, err)
+		}
+		if currRV > biggestResourceVersion {
+			biggestResourceVersion = currRV
+		}
+	}
+
+	livePodsAtSameRV, err := kubeClient.CoreV1().Pods("").List(ctx, metav1.ListOptions{
+		ResourceVersion:      strconv.Itoa(biggestResourceVersion),
+		ResourceVersionMatch: metav1.ResourceVersionMatchExact,
+	})
+	if err != nil && strings.Contains(err.Error(), "The resourceVersion for the provided list is too old") {
+		// if the resourceversion is too old (this happened a bunch), just skip the test.
+		// Again, we'll write a better test with a reflector that isn't vulnerable to this, but to gain some insight early we'll skip this case.
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error listing pods resourceVersion=%v: %w", biggestResourceVersion, err)
+	}
+
+	failures := []string{}
+	for _, livePod := range livePodsAtSameRV.Items {
+		found := false
+		for _, cachedPod := range cachedPods {
+			if cachedPod.Namespace == livePod.Namespace && cachedPod.Name == livePod.Name {
+				found = true
+				if cachedPod.ResourceVersion != livePod.ResourceVersion {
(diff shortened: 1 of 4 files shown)


pkg/monitortests/node/watchpods/cache_check.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, cachedPod := range cachedPods {
+		if len(cachedPod.ResourceVersion) == 0 {
+			return nil, fmt.Errorf("encounted cached pod/%s -n %s without a resourceversion", cachedPod.Namespace, cachedPod.Name)
+		}
+		currRV, err := strconv.Atoi(cachedPod.ResourceVersion)
+		if err != nil {
+			return nil, fmt.Errorf("cached pod/%s -n %s has a non-integer RV %q: %w", cachedPod.Namespace, cachedPod.Name, cachedPod.ResourceVersion, err)
+		}
+		if currRV > biggestResourceVersion {
+			biggestResourceVersion = currRV
```
trozet, line 41: so it looks like you are trying to get the highest RV of all pods, and then you list in the kubeclient with that RV. Is this list RV always equivalent to the highest RV of all pods?

pkg/monitortests/node/watchpods/cache_check.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	for _, cachedPod := range cachedPods {
+		found := false
+		for _, livePod := range livePodsAtSameRV.Items {
+			if cachedPod.Namespace == livePod.Namespace && cachedPod.Name == livePod.Name {
+				found = true
+			}
+		}
+		if !found {
+			failures = append(failures, "cached pod/%s -n %s RV=%v is not present in the live results", cachedPod.Name, cachedPod.Namespace, cachedPod.ResourceVersion)
```
trozet, line 76: couldn't the pod have been deleted or its resource version change by the time you issued the kubeClient cmd? I thought that k8s wont hold past versions of resources and only hold latest, and if you attempt to get a resource version that is obsolete you will get nothing back.

pkg/monitortests/node/watchpods/cache_check.go
pacevedom: Looks like this is not always the case in MicroShift. Some pods may be at resourceVersion 0 (which is invalid for an exact version match) and others may be lower than the biggest one.
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29115/pull-ci-openshift-origin-master-e2e-aws-ovn-microshift-serial/1836509395513511936 (0 example)
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29115/pull-ci-openshift-origin-master-e2e-aws-ovn-microshift-serial/1836559313703800832 (lower resource version example)

p0lyn0mial: I agree with the comments above. To correct this code, we would need to obtain the lastSyncedResourceVersion from the informer, stop it from receiving new updates, and then make a live list request. Otherwise, we are comparing different snapshots. Note that the list's Resource Version is not the same as the RV of individual pods/objects.

pkg/monitortests/node/watchpods/cache_check.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, fmt.Errorf("error listing pods resourceVersion=%v: %w", biggestResourceVersion, err)
+	}
+
+	failures := []string{}
+	for _, livePod := range livePodsAtSameRV.Items {
+		found := false
+		for _, cachedPod := range cachedPods {
+			if cachedPod.Namespace == livePod.Namespace && cachedPod.Name == livePod.Name {
+				found = true
+				if cachedPod.ResourceVersion != livePod.ResourceVersion {
```
p0lyn0mial, line 59: Also, is this event thread-safe? If I recall correctly, the lister returns pointers to items that might change at any time.

pkg/monitortests/node/watchpods/cache_check.go
```diff
@@ -0,0 +1,81 @@
...
+					failures = append(failures, "matching pod/%s -n %s has different resourceVersions: live=%v cached=%v", livePod.Name, livePod.Namespace, livePod.ResourceVersion, cachedPod.ResourceVersion)
+				}
+			}
+		}
+		if !found {
+			failures = append(failures, "live pod/%s -n %s RV=%v is not present in the cached results", livePod.Name, livePod.Namespace, livePod.ResourceVersion)
```
p0lyn0mial, line 65: in this [run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29115/pull-ci-openshift-origin-master-e2e-aws-csi/1836559296096112640) the test failed because it didn't find a pod which must have been deleted from the cache and this wasn't reflected in the code that calculates the max RV

pkg/monitortests/node/watchpods/cache_check.go
benluddy: The addressed objects shouldn't change. Wouldn't that make it impossible to read anything but the length of the slice without a data race?

pkg/monitortests/node/watchpods/cache_check.go
deads2k: > Some pods may be at resourceVersion 0

Pods can never be at resourceversion zero.

pkg/monitortests/node/watchpods/cache_check.go
deads2k: >I thought that k8s wont hold past versions of resources and only hold latest

That's the default behavior, but an list can reflect an *exact* resource version.

However, @p0lyn0mial's point of not knowing the resourceVersion of deleted items is true.  I think to handle that I'm going to limit the test failures to only cases where the cache has an item that no longer exists in the list (the case we think we're seeing in the pod IP reuse test.  Then try to get that merged quickly (probably also excluding RV too old).  Then build a separate and specialized reflector handler that can do the "pause" described in this thread to do the inverse under a different test.

deads2k: > if you attempt to get a resource version that is obsolete you will get nothing back.

Added a check for "too old resourceversion" so we don't fail on it, we just skip this check.

deads2k:
none of the failures were of this new test

/retest

benluddy:
/lgtm

OK merging this with the current caveats in hope of collecting some interesting data.

PR #29127, opened 2024-09-24
WIP kubeapiserver auditloganalyzer: spot handler panics in audit log

vrutkovs:
Don't let any useragent cause too many panics in apiserver

### DIFF
--- a/pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
+++ b/pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
@@ -31,6 +31,7 @@ type auditLogAnalyzer struct {
 	invalidRequestsChecker        *invalidRequests
 	requestsDuringShutdownChecker *lateRequestTracking
 	violationChecker              *auditViolations
+	apiserverPanicChecker         *apiserverPaniced
 
 	countsForInstall *CountsForRun
 }
@@ -42,6 +43,7 @@ func NewAuditLogAnalyzer() monitortestframework.MonitorTest {
 		invalidRequestsChecker:        CheckForInvalidMutations(),
 		requestsDuringShutdownChecker: CheckForRequestsDuringShutdown(),
 		violationChecker:              CheckForViolations(),
+		apiserverPanicChecker:         CheckForApiserverPaniced(),
 	}
 }
 
@@ -85,6 +87,7 @@ func (w *auditLogAnalyzer) CollectData(ctx context.Context, storageDir string, b
 		w.invalidRequestsChecker,
 		w.requestsDuringShutdownChecker,
 		w.violationChecker,
+		w.apiserverPanicChecker,
 	}
 	if w.requestCountTracking != nil {
 		auditLogHandlers = append(auditLogHandlers, w.requestCountTracking)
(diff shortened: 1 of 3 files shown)


pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
```diff
@@ -122,6 +125,35 @@ func (w *auditLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Con
...
+			failures := append(failures, fmt.Sprintf("failed auditIDs: %s", strings.Join(ids, ",")))
+			ret = append(ret,
+				&junitapi.JUnitTestCase{
+					Name: testName,
+					FailureOutput: &junitapi.FailureOutput{
+						Message: strings.Join(failures, "\n"),
+						Output:  "details in audit log",
+					},
+				},
+			)
```
deads2k, line 146: Add a success here so that we flake instead of fail to start.  We can still search for flakes.

pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
```diff
@@ -122,6 +125,35 @@ func (w *auditLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Con
...
+	for userAgent, panics := range w.apiserverPaniced.panicEventsPerUserAgent {
+		testName := fmt.Sprintf("user %s must not produce too many apiserver handler panics", userAgent)
+
+		failures := []string{}
+		if len(panics) > 5 {
+			ids := []string{}
+			for _, id := range panics.UnsortedList() {
+				ids = append(ids, string(id))
+			}
+			failures := append(failures, fmt.Sprintf("failed auditIDs: %s", strings.Join(ids, ",")))
```
deads2k, line 137: we need times for these to track them more effectively.

vrutkovs: we store audit-id, so later more info - like starting time - can be extracted. Do we want time to do more complicated alerting like "no more that n panics in x minutes"?

deads2k: > Do we want time to do more complicated alerting like "no more that n panics in x minutes"?

I don't think these are panics.  They may look like panics, but that's because the timeout handler prints stacks.

>we store audit-id, so later more info - like starting time - can be extracted.
you store the entire auditEvent to pull the times out.  It's just a lot more useful to see it on the spyglass page and compare it to "what's  happening then" than to download an audit log and parse backwards.

pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
```diff
@@ -367,6 +370,37 @@ func (w *auditLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Con
 
 	ret = append(ret, w.violationChecker.CreateJunits()...)
 
+	for userAgent, userAgentPanics := range w.apiserverPanicChecker.panicEventsPerUserAgent.panicEvents {
+		testName := fmt.Sprintf("user %s must not produce too many apiserver handler panics", userAgent)
```
dgoodwin, line 374: This is generating random test names, infinitely increasing the tests in sippy even as a presubmit. 150 tests so far. We'll clean those up but this needs to be a megatest, can't do a test per uncontrolled user inputs.

dgoodwin: ex name | user ip-10-0-21-189/ovnkube@038aea1608e0 (linux/amd64) kubernetes/v0.31.1 must not produce too many apiserver handler panics

vrutkovs: Right, this may not have an authenticated user. 
Should we use "too many apiserver handler panics produced" as testname and move useragent to the error message?
Perhaps, instead we could add username if available and use "unknown" otherwise? The usernames may be infinite too though

dgoodwin: It does seem dangerous even if just authed users, presumably tests could be making some with random names? Typically we would write a whitelist for known users you're particularly interested in broken into their own tests, plus a fallback catch all single test.

PR #29129, opened 2024-09-24
NO-JIRA: add per-second request count tracking from audit log

deads2k:
1. counts all audit events and produces per-second summaries
2. produces CSV listing every second and number of concurrent requests, number of requests received, number of requests received at this second that later end in 500
3. produces a separate CSV for the time during install
4. produces png files rending line charts for the information in 2 and 3

Examples here from a clusterbot cluster of 4.18

**install**
request-counts-for-install_

**install+idle cluster**
request-counts-by-second_

Those regular spikes are **huge**.  Also notice that the 500s are a small portion of traffic.

/assign @vrutkovs @dgrisonnet
/cc  @bertinatto we may want to do some smarter stuff here for apply failures.


deads2k:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.17-e2e-aws-upgrade-ovn-single-node 10

let's see what audit log tracking looks like on these.  I bet single-node is different than HA.

deads2k:
I haven't installed 4.17 yet, but 4.15 does not have the huge spikes.

p0lyn0mial:
Drive by: I think Prometheus is quite good at counting and displaying time series data. Would it make sense to install Prometheus in a pod and then export the desired data from the audit logs to it ? Then we could download the metrics and issue all sorts of queries to display the desired graphs.

vrutkovs:
Alternative idea: shove the logs the into (local) Loki, which can parse logs and build metrics from it and display them from Grafana. This makes the solution excellent for investigation, later on we can run interesting queries in `auditloganalyzer.go` (see existing excessive applies test)

deads2k:
Happy to see any and all improvements after the fact.

deads2k:
I'm glad to see energy in the area.  I'm going to get the current pictures in their current state, but I'm definitely in favor of refining collection to get the data in whatever format is required.

PR #29132, opened 2024-09-25
API-1835: add test to ensure we dont' accidentally drop a condition during the …

deads2k:
…apply refactor

/assign @bertinatto

PR #29133, opened 2024-09-25
Prove condition

deads2k:
prove https://github.com/openshift/origin/pull/29132 does something

/hold

should never merge, but needed for proof.


deads2k:
worked.

PR #29135, opened 2024-09-26
OCPBUGS-42435: Enforce the required-scc monitor test and validate usage of non-standard OCP SCCs

kramaranya:
Enforced required-scc monitor test for fixed namespaces. Extended the required-scc monitor test to validate non-standard SCCs and ensure they are only used in allowed namespaces.

### DIFF
--- a/pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
+++ b/pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
@@ -31,6 +31,11 @@ var defaultSCCs = sets.NewString(
 	"restricted-v2",
 )
 
+var nonStandardSCCNamespaces = map[string]sets.Set[string]{
+	"node-exporter":                   sets.New("openshift-monitoring"),
+	"machine-api-termination-handler": sets.New("openshift-machine-api"),
+}
+
 type requiredSCCAnnotationChecker struct {
 	kubeClient kubernetes.Interface
 }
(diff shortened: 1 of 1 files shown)


pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -81,6 +86,17 @@ func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageD
 
 			suggestedSCC := suggestSCC(&pod)
 			owners := ownerReferences(&pod)
+
+			// check if the SCC is non-standard and whether it's allowed in this namespace
+			validatedSCC := pod.Annotations[securityv1.ValidatedSCCAnnotation]
+			if _, isNonStandard := nonStandardSCCNamespaces[validatedSCC]; isNonStandard {
+				allowedNamespaces := nonStandardSCCNamespaces[validatedSCC]
```
liouk, line 93: You can assign `allowedNamespaces` directly at the if statement, where `isNonStandard` is assigned as well.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -135,7 +151,7 @@ func suggestSCC(pod *v1.Pod) string {
 		return fmt.Sprintf("suggested required-scc: '%s'", pod.Annotations[securityv1.ValidatedSCCAnnotation])
 	}
 
-	return "cannot suggest required-scc, validated SCC is custom"
+	return "cannot suggest required-scc, validated SCC is a custom/non-standard SCC"
```
liouk, line 154: Since you can now distinguish between a custom vs non-standard SCC, use separate messages.

liouk (changes requested):
What if we have already pinned a non-standard SCC to a namespace that's not allowed?

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -75,12 +80,23 @@ func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageD
...
+				if allowedNamespaces, isNonStandard := nonStandardSCCNamespaces[pod.Annotations[securityv1.ValidatedSCCAnnotation]]; isNonStandard && !allowedNamespaces.Has(ns.Name) {
+					failures = append(failures, fmt.Sprintf("pod '%s' has a non-standard SCC '%s' that is not allowed in namespace '%s'", pod.Name, validatedSCC, ns.Name))
+				}
 				continue
 			}
 
 			suggestedSCC := suggestSCC(&pod)
 			owners := ownerReferences(&pod)
+
+			if allowedNamespaces, isNonStandard := nonStandardSCCNamespaces[validatedSCC]; isNonStandard && !allowedNamespaces.Has(ns.Name) {
```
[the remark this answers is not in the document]
kramaranya: Makes sense! I've updated the code, thanks

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -75,13 +80,50 @@ func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageD
 
 		failures := make([]string, 0)
 		for _, pod := range pods.Items {
+			validatedSCC := pod.Annotations[securityv1.ValidatedSCCAnnotation]
+			allowedNamespaces, isNonStandard := nonStandardSCCNamespaces[validatedSCC]
+
 			if _, exists := pod.Annotations[securityv1.RequiredSCCAnnotation]; exists {
+				if isNonStandard && !allowedNamespaces.Has(ns.Name) {
+					failures = append(failures, fmt.Sprintf(
+						"pod '%s' has a non-standard SCC '%s' not allowed in namespace '%s'",
```
liouk, line 89: Allowed namespaces would be useful here as well.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -75,13 +80,50 @@ func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageD
...
+					failures = append(failures, fmt.Sprintf(
+						"annotation missing from pod '%s'%s; pod is using non-standard SCC '%s' not allowed in namespace '%s'; allowed namespaces are: %s",
+						pod.Name, owners, validatedSCC, ns.Name, strings.Join(allowedNamespaces.UnsortedList(), ", ")))
+				}
+				continue
+			}
+
+			failures = append(failures, fmt.Sprintf(
+				"annotation missing from pod '%s'%s; cannot suggest required-scc, validated SCC '%s' is a custom SCC",
+				pod.Name, owners, validatedSCC))
```
liouk, line 126: Turn this block into a switch with the custom SCC case as the default, no need to `continue` for every case.

liouk (changes requested):

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -73,6 +86,20 @@ func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageD
...
+			if runLevel == "0" || runLevel == "1" {
+				isFlakingNamespace = true
+			}
+		}
+
+		// check if the namespace contains workloads without SCC pinning
+		if namespacesWithPendingSCCPinning.Has(ns.Name) {
+			isFlakingNamespace = true
+		}
+
```
liouk, line 102: This should be clear enough:
```suggestion
isFlakingNamespace := ns.Labels["openshift.io/run-level"] == "0" ||
  ns.Labels["openshift.io/run-level"] == "1" ||
  namespacesWithPendingSCCPinning.Has(ns.Name)
```

`len(ns.Labels["openshift.io/run-level"])` could also be enough, but it's better to be explicit.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -130,18 +157,29 @@ func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageD
...
+		} else {
+			// enforce as a failure
+			junits = append(junits,
+				&junitapi.JUnitTestCase{
+					Name:          testName,
+					SystemOut:     failureMsg,
+					FailureOutput: &junitapi.FailureOutput{Output: failureMsg},
+				},
+			)
+		}
```
liouk, line 182: In both cases we need one failing `JUnitTestCase`, so append it always. If we need a flake as well, append one.

pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go
```diff
@@ -73,6 +86,20 @@ func (w *requiredSCCAnnotationChecker) CollectData(ctx context.Context, storageD
 			continue
 		}
 
+		isFlakingNamespace := false
```
liouk, line 89: Name suggestion: `flakeWhenFailed`

liouk (changes requested):

neisw:
/payload-job periodic-ci-openshift-hypershift-release-4.19-periodics-e2e-aws-ovn-conformance

PR #29137, opened 2024-09-26
API-1835: add test to prevent invalid apply statements

deads2k:
To catch things like

> 2024-09-26T09:00:24.282667248Z E0926 09:00:24.282588       1 base_controller.go:271] OAuthAPIServerControllerWorkloadController reconciliation failed: Authentication.operator.openshift.io "cluster" is invalid: [conditions[55].type: Invalid value: "": conditions[55].type in body should match '^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$', conditions[55].status: Unsupported value: "": supported values: "True", "False", "Unknown", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]

in our operator logs.  It would be better to track client metic invalid request rates, but I don't know how at the moment.

### DIFF
--- a/pkg/monitor/monitorapi/types.go
+++ b/pkg/monitor/monitorapi/types.go
@@ -235,6 +235,8 @@ const (
 	LeaseAcquiring        IntervalReason = "Acquiring"
 	LeaseAcquiringStarted IntervalReason = "StartedAcquiring"
 	LeaseAcquired         IntervalReason = "Acquired"
+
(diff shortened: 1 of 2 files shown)


bertinatto (commented):
/lgtm

PR #29139, opened 2024-09-26
extended: add user namespace test

haircommander:
Specifically, add a test that verifies the kubelet denies a pod in a user namespace if it doesn't support doing so.
This is to protect against cases where FeatureGate rollout is in progress, and the KAS believes UserNamespaces
are supported, when they're not yet.


test/extended/kubelet/userns.go
```diff
@@ -0,0 +1,361 @@
...
+				}
+			}()
+			originalMCName := getMCName(oc, workerMCPName)
+			g.By(fmt.Sprintf("Starting with MC name %s", originalMCName))
+
+			// create mcp
+			cleanupFuncs = append(cleanupFuncs, createMCP(oc, mcpName))
+
+			// create mc and wait for it to roll out
+			mcName, deleteMCFunc := createKubeletDropinMC(oc, originalMCName, mcpName)
```
deads2k, line 86: need to wait for this to roll out?

test/extended/kubelet/userns.go
```diff
@@ -0,0 +1,361 @@
...
+			pod := podframework.MakePod(oc.Namespace(), map[string]string{mcpNameToRole(mcpName): ""}, nil, admissionapi.LevelBaseline, "")
+			falseVar := false
+			pod.Spec.HostUsers = &falseVar
+			g.By("Creating user namespaced pod")
+			pod, err := oc.KubeClient().CoreV1().Pods(oc.Namespace()).Create(context.Background(), pod, metav1.CreateOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			// Pod should stay in pending
+			// Events would be a better way to tell this, as we could actually read the event from the kubelet saying the exact issue,
+			// but events are best-effort and can't be relied on.
```
deads2k, line 102: as a practical matter, they tend to be present in our CI runs.  Perhaps if the event is present end the test early?

test/extended/kubelet/userns.go
```diff
@@ -0,0 +1,361 @@
...
+		// - If the kubelet doesn't support user namespaces because of version skew
+		// - If the kubelet doesn't support user namespaces because of a FeatureGate rollout/upgrade
+		// In either of these cases, the root cause that needs to be verified is the kubelet doesn't have the feature gate
+		// on, but the KAS does.
+		// Since we can't easily test this through the openshift API, we (hackily) use a MachineConfig to create a drop-in
+		// kubelet service configuration. This configuration will add the `--feature-gate="UserNamespacesSupport=false"` flag
+		// to the kubelet's CLI, thus disabling the feature gate.
+		// Then, we check that the kubelet fails to create a pod that requests a user namespace.
+		// A lot of the code is adapted from MCO e2e tests, as well as an e2e-node test case that verifies the same.
+		g.It("Kubelet should fail to create a user namespaced pod if featuregate is off [Serial]", func() {
```
deads2k, line 70: since it's changing the machineconfig, this should probalby be `[Serial]`

test/extended/kubelet/userns.go
```diff
@@ -0,0 +1,361 @@
...
+
+			// create mcp
+			cleanupFuncs = append(cleanupFuncs, createMCP(oc, mcpName))
+
+			// create mc and wait for it to roll out
+			mcName, deleteMCFunc := createKubeletDropinMC(oc, originalMCName, mcpName)
+			cleanupFuncs = append(cleanupFuncs, deleteMCFunc)
+
+			// label random node in worker pool
+			cleanupFuncs = append(cleanupFuncs, labelRandomNodeFromPool(oc, mcpName, workerMCPName, mcName, originalMCName))
```
deads2k, line 90: ah ha!  this one does the wait.

PR #29144, opened 2024-09-27
OCPEDGE-1276: fix: adding in a SNO time override for AWS Metal

jeff-roche:
Adding an upgrade time override for AWS Metal instances which have a longer reboot cycle due to not being virtualized. This longer reboot cycle is causing upgrade failures on AWS Metal based SNO lanes.


neisw:
/approve
again

PR #29145, opened 2024-09-30
NO-JIRA: update recovery tests with rev bumping

tjungblu:


test/extended/dr/common.go
```diff
@@ -569,9 +567,7 @@ func runDeleteAndRestoreScript(oc *exutil.CLI, restoreNode *corev1.Node, backupN
          mv \$SNAPSHOT /home/core/backup/snapshot.db
          sudo rm -rf /var/lib/etcd
          
-         sudo ETCD_ETCDCTL_RESTORE_ENABLE_BUMP=true /usr/local/bin/cluster-restore.sh /home/core/backup
-         # this will cause the pod to disappear effectively, must be the last statement
-         sudo systemctl restart kubelet.service
```
tjungblu, line 574: this is likely something we need to keep

PR #29147, opened 2024-10-01
API-1835: don't allow platform users to make invalid requests

deads2k:
We want to catch users making invalid requests to prevent mistakes.


deads2k:
Sigh. Looks like some openshift controllers sometimes create invalid resources. We'll have to exclude those

PR #29148, opened 2024-10-02
NO-JIRA: test for checking the healthz endpoints for the openshift-apiserver

p0lyn0mial:
The test could be defined in the operand repo, but since the operand repo doesn't have any CI job defined, I decided to add the test directly to the origin.

This might cause minor issues during rebases when the output changes, but this shouldn't happen frequently.

Also we should add test cases for the remaining aggregated APIs.

Required for https://github.com/openshift/openshift-apiserver/pull/452

### DIFF
--- a/test/extended/apiserver/health_endpoints.go
+++ b/test/extended/apiserver/health_endpoints.go
@@ -1,0 +1,114 @@
+package apiserver
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/onsi/ginkgo/v2"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/kubernetes/test/e2e/framework"
+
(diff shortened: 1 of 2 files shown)


test/extended/apiserver/health_endpoints.go
```diff
@@ -0,0 +1,108 @@
...
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = ginkgo.Describe("[sig-api-machinery] API health endpoints", func() {
+	f := framework.NewDefaultFramework("api-health-endpoints")
+	f.SkipNamespaceCreation = true
+
+	oc := exutil.NewCLIWithoutNamespace("api-health-endpoints").AsAdmin()
+
+	ginkgo.It("should contain the required checks for the openshift-apiserver APIs", func() {
```
pacevedom, line 26: Is it possible to skip this one in MicroShift? We do not have the openshift-apiserver components.

p0lyn0mial: sure, thanks for letting me know.

p0lyn0mial: done.

dinhxuanvu (approved):
/lgtm

PR #29149, opened 2024-10-02
OPNET-588: On-Prem test for apiserver reachability from the LB

mkowalski:


mkowalski:
/retitle Initial attempt for e2e haproxy test

It's not WIP actually

pkg/monitortests/network/onpremhaproxy/monitortest.go
```diff
@@ -0,0 +1,291 @@
...
+		}
+
+		//deletionTime := time.Now()
+		//deletedIntervals := monitorapi.Intervals(allChangesForTheNode).Filter(func(eventInterval monitorapi.Interval) bool {
+		//	return eventInterval.Message.Reason == monitorapi.APIServiceDeletedInAPI
+		//})
+		//if len(deletedIntervals) > 0 {
+		//	deletionTime = deletedIntervals[0].To
+		//}
+		//if len(lastAvailableStatus) > 0 {
```
deads2k, line 149: Something like this check is needed so that if the job run ends with the `lastAvailableStatus == "DOWN"`, there is an interval on the timeline.

pkg/monitortests/network/onpremhaproxy/monitortest.go
```diff
@@ -0,0 +1,291 @@
...
+
+func scanAllOperatorPods(ctx context.Context, kubeClient kubernetes.Interface, logHandlers ...podaccess.LogHandler) error {
+	pods, err := kubeClient.CoreV1().Pods("").List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("couldn't list pods: %w", err)
+	}
+
+	errs := []error{}
+	for _, pod := range pods.Items {
+		if !(strings.HasPrefix(pod.Namespace, "openshift-kni-infra") || strings.HasPrefix(pod.Namespace, "openshift-openstack-infra") || strings.HasPrefix(pod.Namespace, "openshift-vsphere-infra")) {
```
vrutkovs, line 54: Lets list pods from these namespaces instead of listing all pods?

pkg/monitortests/network/onpremhaproxy/monitortest.go
```diff
@@ -0,0 +1,291 @@
...
+			continue
+		}
+		// Pods in cloud provider platforms do not have nice names, e.g. "haproxy-i7rd6ssm-ac890-q5tzv-master-0" in OpenStack
+		if !strings.Contains(pod.Name, "haproxy") || !strings.Contains(pod.Name, "master") {
```
vrutkovs, line 58: Limit to a labelselector?

pkg/monitortests/network/onpremhaproxy/monitortest.go
```diff
@@ -0,0 +1,291 @@
...
+		//			Display().
+		//			Build(previousChangeTime, deletionTime),
+		//	)
+		//}
+	}
+
+	return constructedIntervals, nil
+}
+
+func (*operatorLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
```
vrutkovs, line 174: This should create a failed test if collection failed for some reason

vrutkovs (commented):
This needs an interval template updated, so that we could correlate haproxy events with apiserver disruption events, see https://github.com/openshift/origin/commit/08cb5ebe156aec7945651288ff4124ad5742f946 for example

prb112:
@Shilpa-Gokul heads up on this PR / Monitor Test.

pkg/monitortests/network/onpremhaproxy/monitortest.go
mkowalski: We should let it live for a while before we decide to mark the test as Failed, opposed to current flakiness. At this moment I am not sure there is a common agreement on how widespread the outage should be for us to mark it as a Failure.

When only 1 haproxy instance sees only 1 apiserver as down, do you already call it an "outage"? From client perspective you still reach the API. And so on, and so on

mkowalski:
/retitle OPNET-588: On-Prem test for apiserver reachability from the LB

pkg/monitortests/network/onpremhaproxy/monitortest.go
vrutkovs: Ah, I agree, it should flake but my point is that failure here means collection failed.

>When only 1 haproxy instance sees only 1 apiserver as down, do you already call it an "outage"? From client perspective you still reach the API. And so on, and so on

No, we have separate outage monitoring, which uses both load-balancer and service network paths to reach API server. The intervals created by this test would be useful for us during investigation - but imo this shouldn't outright fail the test which a single apiserver instance is marked as down in the LB (this is expected to happen during revision rollouts)

pkg/monitortests/network/onpremhaproxy/monitortest.go
mkowalski: Done

pkg/monitortests/network/onpremhaproxy/monitortest.go
mkowalski: Done

pkg/monitortests/network/onpremhaproxy/monitortest.go
```diff
@@ -0,0 +1,293 @@
...
+	})
+
+	testName := fmt.Sprint("[Jira: Networking / On-Prem Host Networking] Haproxy must be able to reach kubeapi server")
+	success := &junitapi.JUnitTestCase{Name: testName}
+	somethingFailed := false
+
+	testNameToFailures := map[string][]string{}
+	for _, interval := range leaseIntervals {
+		if interval.Message.Reason == monitorapi.OnPremHaproxyDetectsDown {
+			//CHOCOBOMB(mko) Status change itself is NOT an indication of failure. It's here only for development
```
cybertron, line 191: Maybe we could mark the test as failed if all three apiservers are down at the same time? We might have to filter out the initial state since I believe that will always be the case when haproxy first starts, but if it happens a second time it could be considered a failure.

mkowalski: Can we first merge a working version and afterwards work on tweaks and make it shiny? I already get a whole long wishlist of what it would be cool if that was doing, but as the first pass it is important that this test keeps running all the time

pkg/monitortests/network/onpremhaproxy/monitortest.go
mkowalski: Ok, I may need your help then. Because what we have here is the following

* there are 3 namespaces of our interest
* max 1 of them contains our workload that has interesting logs

but at the same time this tests when running on AWS will not collect a single line of log. So I cannot easily say "if len(logs) == 0, fail".

Do we have some elegant programmatic way of doing "if platform.isOnPrem()" in this repo?

neisw:
/payload-job periodic-ci-openshift-hypershift-release-4.18-periodics-e2e-aws-ovn-conformance

pkg/monitortests/network/onpremhaproxy/monitortest.go
cybertron: Yes, I'm good with that. The biggest win here is having the HAProxy status in the timeline, so let's get that in.

PR #29154, opened 2024-10-03
TRT-1843: Revert "add test to ensure we dont' accidentally drop a condition dur…

deads2k:
…ing the apply refactor"

This reverts commit a9668eaebedadd7176a546406cd4507f429b71f0.

reverts https://github.com/openshift/origin/pull/29132


dgoodwin:
/retitle TRT-1843: Revert "add test to ensure we dont' accidentally drop a condition dur…

PR #29160, opened 2024-10-04
OCPEDGE-1266: fix: adding oslat updates and outputting test results to the artifacts

jeff-roche:
- Pins the oslat main thread to the first thread under test to accommodate workload partitioning
- Writes the outputs of oslat, cyclictest and rteval to the artifacts dir

### DIFF
--- a/test/extended/testdata/kernel/rt-tests-environment.yaml
+++ b/test/extended/testdata/kernel/rt-tests-environment.yaml
@@ -9,7 +9,6 @@ metadata:
     pod-security.kubernetes.io/warn: privileged
   annotations:
     workload.openshift.io/allowed: management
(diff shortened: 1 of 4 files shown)


test/extended/testdata/kernel/rt-tests-environment.yaml
```diff
@@ -9,7 +9,6 @@ metadata:
     pod-security.kubernetes.io/warn: privileged
   annotations:
     workload.openshift.io/allowed: management
-    openshift.io/node-selector: "node-role.kubernetes.io/worker="
```
jeff-roche, line 12: This was needed when running on GCP but no longer desired as these tests are going to be targeted for SNO on 4.17+

PR #29161, opened 2024-10-04
NO-JIRA: test/extended/apiserver: test for checking the healthz endpoints for oauth-apiserver

dinhxuanvu:


test/extended/apiserver/health_endpoints.go
```diff
@@ -115,3 +164,15 @@ func testPath(transport http.RoundTripper, path string, requiredChecks sets.Set[
 	}
 	return nil
 }
+
+func checkFlatform(ctx context.Context, name string, oc *exutil.CLI) error {
```
p0lyn0mial, line 168: I would change the signature of this function to `isSupportedPlatform(ctx, name, oc) bool`.
For handling errors inside this function I would use `framework.ExpectNoError(err)`.

test/extended/apiserver/health_endpoints.go
```diff
@@ -115,3 +164,15 @@ func testPath(transport http.RoundTripper, path string, requiredChecks sets.Set[
...
+	isMicroShift, err := exutil.IsMicroShiftCluster(oc.AdminKubeClient())
+	framework.ExpectNoError(err)
+	if isMicroShift {
+		return fmt.Errorf("MicroShift does not have this component %s", name)
```
p0lyn0mial, line 172: and here i would return `false, "MicroShift does not have this component %s", name"

dinhxuanvu:
@p0lyn0mial Updated. PTAL. Thanks.

PR #29163, opened 2024-10-07
NO-JIRA: prove json patch

p0lyn0mial:


test/extended/apiserver/patch.go
```diff
@@ -0,0 +1,91 @@
...
+	"k8s.io/client-go/dynamic"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	"github.com/openshift/origin/test/extended/apiserver/patch"
+	"github.com/openshift/origin/test/extended/testdata"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-api-machinery] JSON Patch", func() {
```
pacevedom, line 22: Can you tag this with `operator.openshift.io` so that MicroShift will skip it?

pacevedom (commented):
Wondering about the usage of the apiserver resource, would it be feasible to use a k8s object instead so we wouldnt need to skip?

test/extended/apiserver/patch.go
deads2k: > Can you tag this with `operator.openshift.io` so that MicroShift will skip it?

still outstanding.

test/extended/apiserver/patch.go
```diff
@@ -0,0 +1,176 @@
...
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	"github.com/openshift/origin/test/extended/apiserver/jsonpatch"
+	"github.com/openshift/origin/test/extended/testdata"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-api-machinery] JSON Patch", func() {
```
deads2k, line 25: I don't see the test case that checks what happens when the precondition in the patch fails.

test/extended/apiserver/patch.go
```diff
@@ -0,0 +1,192 @@
...
+
+	g.BeforeEach(func() {
+		dynamicClient, err := dynamic.NewForConfig(oc.AdminConfig())
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		gvr := operatorv1.GroupVersion.WithResource("kubeapiservers")
+		gvk = operatorv1.GroupVersion.WithKind("KubeAPIServer")
+		resourceClient = dynamicClient.Resource(gvr)
+
+		err = resourceClient.Delete(ctx, "test-apiserver", metav1.DeleteOptions{})
```
deads2k, line 42: What deletes the last one of these tests that runs?

test/extended/apiserver/patch.go
```diff
@@ -0,0 +1,192 @@
...
+
+var _ = g.Describe("[sig-api-machinery] JSON Patch [apigroup:operator.openshift.io]", func() {
+	var (
+		resourceClient dynamic.ResourceInterface
+		gvk            schema.GroupVersionKind
+	)
+	ctx := context.TODO()
+	oc := exutil.NewCLIWithoutNamespace("json-patch")
+
+	g.BeforeEach(func() {
```
deads2k, line 34: How do you ensure that none of the `It`s run in parallel in a parallel e2e test?

deads2k: > How do you ensure that none of the `It`s run in parallel in a parallel e2e test?

If you don't know, perhaps what you want is to create in each test with a generateName (not a name) and then cleanup locally.

PR #29169, opened 2024-10-08
ETCD-657: add simple quorum restore test

tjungblu:
this adds a simple quorum restore test using the newly introduced quorum-restore.sh


test/extended/dr/recovery.go
```diff
@@ -165,3 +160,62 @@ var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Suite:openshift/etcd/re
 		assertPostBackupResourcesAreNotFound(oc)
 	})
 })
+
+var _ = g.Describe("[sig-etcd][Feature:DisasterRecovery][Suite:openshift/etcd/recovery][Timeout:1h]", func() {
```
tjungblu, line 164: TODO we have to add an assertion here that the restore static pods are properly cleaned up, if we go with renaming them to another pod/yaml file

PR #29175, opened 2024-10-09
TRT-1575: Add more exceptions for operator degraded cases

xueqzhan:


pacevedom:
`e2e-aws-ovn-microshift-serial` are failing with this error, which is not present in the parallel version?

PR #29180, opened 2024-10-10
NE-1815: Adapt router metrics tests for DCM support

alebedev87:
This PR follows up on a Dynamic Configuration Manager (DCM) gap identified during [the smoke test](https://github.com/openshift/cluster-ingress-operator/pull/1113#issuecomment-2315530761).

When DCM is enabled by default, server management behavior changes. Not all servers will be in the UP state, and this must be accounted for in the router metrics tests. This PR updates the tests to handle scenarios where some servers may be DOWN. The scenarios when DCM is not enabled should still be working.

### DIFF
--- a/test/extended/router/metrics.go
+++ b/test/extended/router/metrics.go
@@ -159,9 +159,12 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 				metrics, err = p.TextToMetricFamilies(bytes.NewBufferString(results))
 				o.Expect(err).NotTo(o.HaveOccurred())
 
-				if len(findGaugesWithLabels(metrics["haproxy_server_up"], serverLabels)) == 2 {
-					if findGaugesWithLabels(metrics["haproxy_backend_connections_total"], routeLabels)[0] >= float64(times) {
-						return true, nil
+				if len(findNonZeroGaugesWithLabels(metrics["haproxy_server_up"], serverLabels)) == 2 {
+					if g := findGaugesWithLabels(metrics["haproxy_backend_connections_total"], routeLabels); len(g) > 0 {
+						// stop retrying if the route got expected number of connections.
+						if g[0] >= float64(times) {
+							return true, nil
+						}
 					}
 					// send a burst of traffic to the router
 					g.By("sending traffic to a weighted route")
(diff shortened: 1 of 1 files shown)


test/extended/router/metrics.go
```diff
@@ -159,9 +159,12 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 				metrics, err = p.TextToMetricFamilies(bytes.NewBufferString(results))
 				o.Expect(err).NotTo(o.HaveOccurred())
 
-				if len(findGaugesWithLabels(metrics["haproxy_server_up"], serverLabels)) == 2 {
-					if findGaugesWithLabels(metrics["haproxy_backend_connections_total"], routeLabels)[0] >= float64(times) {
-						return true, nil
+				if len(findNonZeroGaugesWithLabels(metrics["haproxy_server_up"], serverLabels)) == 2 {
+					for _, g := range findNonZeroGaugesWithLabels(metrics["haproxy_backend_connections_total"], routeLabels) {
+						// stop retrying if any route got expected number of connections.
```
gcs278, line 164: Aren't we looking for a specific route namespace/name label combination? Is it safe to assume that will only return 1 unique route?

test/extended/router/metrics.go
```diff
@@ -366,13 +374,34 @@ func findCountersWithLabels(f *dto.MetricFamily, promLabels map[string]string) [
...
 }
 
+// findNonZeroGaugesWithLabels skips gauges with 0 value to avoid false negatives
+// from the server slots which are DOWN.
+// The Dynamic Configuration Manager (DCM) adds dynamic server slots and
+// uses both these slots and the servers added during HAProxy config rendering.
+// When DCM is enabled, not all server slots are always in the UP state,
+// unlike the behavior without DCM where all slots are UP by default.
+// As a result, we may want to filter out servers where the gauge is set to 0 (== DOWN).
+func findNonZeroGaugesWithLabels(f *dto.MetricFamily, promLabels map[string]string) []float64 {
```
gcs278, line 397: Would it be more precise to filter out dynamic servers based on their name? I know we have functions to include labels, so you would need a function to exclude labels. The names of the servers are `_dynamic-pod-#` (2 right now). Could be a regex, or a `contains` type of thing.

Do you think that filtering out zero value gauges here is "weakening" the precision of the test, and opening possibilities for metrics bugs to slip through this E2E test? What happens if haproxy is writing a bunch of bogus metrics for servers, but they have zero value? Would that slip through?

gcs278 (commented):
Nice PR, just have some concern that we might be losing test precision, and suggested an idea. Feel free to let me know if you disagree.

test/extended/router/metrics.go
alebedev87: Upd: added a `TODO` comment.

test/extended/router/metrics.go
```diff
@@ -159,9 +159,12 @@ var _ = g.Describe("[sig-network][Feature:Router]", func() {
 				metrics, err = p.TextToMetricFamilies(bytes.NewBufferString(results))
 				o.Expect(err).NotTo(o.HaveOccurred())
 
-				if len(findGaugesWithLabels(metrics["haproxy_server_up"], serverLabels)) == 2 {
-					if findGaugesWithLabels(metrics["haproxy_backend_connections_total"], routeLabels)[0] >= float64(times) {
-						return true, nil
+				if len(findNonZeroGaugesWithLabels(metrics["haproxy_server_up"], serverLabels)) == 2 {
+					for _, g := range findNonZeroGaugesWithLabels(metrics["haproxy_backend_connections_total"], routeLabels) {
```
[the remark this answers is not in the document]
alebedev87: `findNonZeroGaugesWithLabels` is generic enough to filter out any zero value gauges. But I agree, let's keep it for the server metrics only as its description states.

test/extended/router/metrics.go
gcs278: Sounds like a reasonable future approach. :+1:

PR #29189, opened 2024-10-14
OPRUN-3576: Add OLMv1 origin tests

tmshort:
This is the initial test. Only works on tech-preview. Checks for presence of OLMv1 CRDs.

### DIFF
--- a/test/extended/olm/olmv1.go
+++ b/test/extended/olm/olmv1.go
@@ -1,0 +1,60 @@
+package operators
+
+import (
+	"fmt"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const (
+	olmv1GroupName = "olm.operatorframework.io"
+)
+
+var _ = g.Describe("[sig-olmv1] OLMv1 CRDs", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLIWithoutNamespace("default")
+
+	g.It("should be installed", func(ctx g.SpecContext) {
+		// Check for tech preview, if this is not tech preview, bail
+		if !exutil.IsTechPreviewNoUpgrade(ctx, oc.AdminConfigClient()) {
+			g.Skip("Test only runs in tech-preview")
+		}
+
+		// supports multiple versions during transision
+		providedAPIs := []struct {
+			group   string
+			version []string
+			plural  string
+		}{
+			{
+				group:   olmv1GroupName,
+				version: []string{"v1alpha1", "v1"},
+				plural:  "clusterextensions",
+			},
+			{
+				group:   olmv1GroupName,
+				version: []string{"v1alpha1", "v1"},
+				plural:  "clustercatalogs",
+			},
+		}
+
+		for _, api := range providedAPIs {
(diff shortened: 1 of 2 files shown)


tmshort:
Since you were lucky enough to be randomly chosen by github... @deads2k @mike-nguyen, my tests only do something under tech-preview (until OLMv1 is out of tech preview), but a number of other tests seem to be failing? Is this normal? How do I get this PR to progress?

LalatenduMohanty:
@wking @petr-muller are my go to origin test experts. Can you guys take a look?

dgoodwin:
/override ci/prow/e2e-aws-ovn-ipsec-serial

Known broken.

test/extended/olmv1/olmv1.go
```diff
@@ -0,0 +1,60 @@
...
+			},
+		}
+
+		for _, api := range providedAPIs {
+			g.By(fmt.Sprintf("checking %s at version %s [apigroup:%s]", api.plural, api.version, api.group))
+			// Ensure expected version exists in spec.versions and is both served and stored
+			var err error
+			var raw string
+			for _, ver := range api.version {
+				raw, err = oc.AsAdmin().Run("get").Args("crds", fmt.Sprintf("%s.%s", api.plural, api.group), fmt.Sprintf("-o=jsonpath={.spec.versions[?(@.name==\"%s\")]}", ver)).Output()
```
joelanford, line 50: nit: to avoid having to use escaped quotation marks

```suggestion
				raw, err = oc.AsAdmin().Run("get").Args("crds", fmt.Sprintf("%s.%s", api.plural, api.group), fmt.Sprintf(`-o=jsonpath={.spec.versions[?(@.name==%q)]}`, ver)).Output()
```

joelanford (commented):
Seems like a reasonable foray into OLMv1 origin tests, but I do wonder if this will really stand the test of time.

I would somewhat assume that this would be quickly obviated by origin tests that actually exercise the APIs that we expect to exist.

test/extended/olmv1/olmv1.go
tmshort: This is a first pass at getting *something* in. The plan is to actually do a happy-path, but I had been waiting on other work to complete before I could get happy-path working.

tmshort:
> Seems like a reasonable foray into OLMv1 origin tests, but I do wonder if this will really stand the test of time.
> 
> I would somewhat assume that this would be quickly obviated by origin tests that actually exercise the APIs that we expect to exist.

This is a first pass at getting something in. The plan is to actually do a happy-path, but I had been waiting on other work to complete before I could get happy-path working.

tmshort:
@deads2k @mike-nguyen looking for approval

dtfranz:
@tmshort Any reason in particular we're creating a new `test/extended` subfolder? IMO we could also just stick this in the existing `olm` folder, and then we wouldn't need to manage another `OWNERS` file...

tmshort:
> @tmshort Any reason in particular we're creating a new `test/extended` subfolder? IMO we could also just stick this in the existing `olm` folder, and then we wouldn't need to manage another `OWNERS` file...

To keep them separate. Makes it easier for later... but I get your point...

dtfranz (approved):
/lgtm

LalatenduMohanty (approved):
/lgtm

PR #29191, opened 2024-10-15
NO-JIRA: a new test for the jsonpatch library 

p0lyn0mial:
for https://github.com/openshift/library-go/pull/1835


p0lyn0mial:
/hold

need to remove the fake bump

PR #29210, opened 2024-10-20
OCPBUGS-43512: New ManagedBootImages test appears to be failing too often

djoshy:
This test extends the timeout for the following test to 2 minutes, as 30 seconds seems to be too short.
```
[sig-mco][OCPFeatureGate:ManagedBootImages][Serial] Should degrade on a MachineSet with an OwnerReference [apigroup:machineconfiguration.openshift.io] [Suite:openshift/conformance/serial]
```


yuqi-zhang (approved):
/lgtm

Seems sane to me

PR #29214, opened 2024-10-21
NO-JIRA: update-tls-artifacts: set owner to unknown if unset

vrutkovs:
Avoid empty string in some markdown reports for TLS registry, instead sets the label as "Unknown"


tls/autoregenerate-after-expiry/autoregenerate-after-expiry.md
```diff
@@ -3,7 +3,7 @@
 ## Table of Contents
   - [How to meet the requirement](#How-to-meet-the-requirement)
   - [Items Do NOT Meet the Requirement (258)](#Items-Do-NOT-Meet-the-Requirement-258)
-    - [ (4)](#-4)
+    - [Unknown (4)](#Unknown-4)
```
p0lyn0mial, line 6: nit: how about `Unknown Owner` ? (at least for me that would read better)

vrutkovs: Good idea, updated

PR #29219, opened 2024-10-22
MCO-1370: Add boot image configmap test to openshift/origin

djoshy:
This adds one more boot image update test to the MCO's test suite. This test ensures that the boot image configmap "stamp" action by the operator is always done, as long as at least one master node is up to date.

### DIFF
--- a/test/extended/machine_config/boot_image_update_agnostic.go
+++ b/test/extended/machine_config/boot_image_update_agnostic.go
@@ -129,3 +129,34 @@ func DegradeOnOwnerRefTest(oc *exutil.CLI, fixture string) {
 	}, 30*time.Second, 2*time.Second).Should(o.BeTrue())
 	framework.Logf("Succesfully verified that the cluster operator is no longer degraded")
 }
+
+func EnsureConfigMapStampTest(oc *exutil.CLI, fixture string) {
+	// Update boot image configmap stamps with a "fake" value, wait for it to be updated back by the operator.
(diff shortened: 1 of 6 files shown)


test/extended/machine_config/boot_image_update_agnostic.go
```diff
@@ -129,3 +129,34 @@ func DegradeOnOwnerRefTest(oc *exutil.CLI, fixture string) {
...
+func EnsureConfigMapStampTest(oc *exutil.CLI, fixture string) {
+	// Update boot image configmap stamps with a "fake" value, wait for it to be updated back by the operator.
+	err := oc.Run("patch").Args("configmap", cmName, "-p", `{"data": {"MCOVersionHash": "fake-value"}}`, "-n", mcoNamespace).Execute()
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	err = oc.Run("patch").Args("configmap", cmName, "-p", `{"data": {"MCOReleaseImageVersion": "fake-value"}}`, "-n", mcoNamespace).Execute()
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	// Ensure atleast one master node is ready
+	WaitForOneMasterNodeToBeReady(oc)
```
yuqi-zhang, line 142: Curious, does the readiness of masters affect this at all?

djoshy: So although the test is not actively doing anything to affect the masters, this is our gate before the operator would stamp the configmap with current version/hash: https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/sync.go#L1825-L1826

The MSBIC will wait until this stamp is present before updating the boot image - this is our check to ensure that the boot images are not updated too early

PR #29220, opened 2024-10-22
SDN-4930: network_segmentation: allow multiple UDN status conditions

npinaeva:
Fix UDN status check to assert required condition is present in the status instead of checking that status has only 1 condition.
required to make https://github.com/openshift/ovn-kubernetes/pull/2314 pass


trozet:
/retitle NOJIRA: network_segmentation: allow multiple UDN status conditions

tssurya:
/retitle SDN-4930: network_segmentation: allow multiple UDN status conditions

PR #29232, opened 2024-10-24
SDN-4168: Add IPsec resilience tests

pperiyasamy:
This PR adds required resilience e2e tests which ensures IPsec deployment and pod traffic are working in all such scenarios.

### DIFF
--- a/pkg/testsuites/standard_suites.go
+++ b/pkg/testsuites/standard_suites.go
@@ -55,6 +55,7 @@ var staticSuites = []ginkgo.TestSuite{
 		},
 		Parallelism:          30,
 		MaximumAllowedFlakes: 15,
+		TestTimeout:          20 * time.Minute,
 	},
 	{
 		Name: "openshift/conformance/serial",
--- a/test/extended/networking/ipsec.go
+++ b/test/extended/networking/ipsec.go
@@ -12,15 +12,18 @@ import (
 	v1 "github.com/openshift/api/operator/v1"
 	exutil "github.com/openshift/origin/test/extended/util"
 	"golang.org/x/sync/errgroup"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
 	e2eoutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	"k8s.io/kubernetes/test/e2e/framework/statefulset"
 	admissionapi "k8s.io/pod-security-admission/api"
 
@@ -92,6 +95,8 @@ var (
 	rightServerCertName = "right_server"
 	// Expiration date for certificates.
 	certExpirationDate = time.Date(2034, time.April, 10, 0, 0, 0, 0, time.UTC)
+	// http endpoint port for the pod traffic test
+	port uint16 = 8080
 )
 
 type trafficType string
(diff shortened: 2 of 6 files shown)


pperiyasamy:
/assign @jcaamano @trozet 

PTAL. Let's make use of this PR to add all missing tests for sufficient IPsec coverage.

pperiyasamy:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-ovn-ipsec-step-registry

pkg/testsuites/standard_suites.go
```diff
@@ -55,6 +55,7 @@ var staticSuites = []ginkgo.TestSuite{
 		},
 		Parallelism:          30,
 		MaximumAllowedFlakes: 15,
+		TestTimeout:          20 * time.Minute,
```
martinkennelly, line 58: I think you need to have a discussion with the maintainers regarding this - did it occur on another comms channel?

test/extended/networking/ipsec.go
```diff
@@ -632,6 +653,117 @@ var _ = g.Describe("[sig-network][Feature:IPsec]", g.Ordered, func() {
...
+	var ipsecMode v1.IPsecMode
+
+	InOVNKubernetesContext(func() {
+		g.BeforeAll(func() {
+			var err error
+			ipsecMode, err = getIPsecMode(oc)
+			o.Expect(err).NotTo(o.HaveOccurred())
+		})
+
+		g.It("check pod traffic are working across nodes", func() {
```
martinkennelly, line 668: "are" isnt right - should be is

test/extended/networking/ipsec.go
```diff
@@ -632,6 +653,117 @@ var _ = g.Describe("[sig-network][Feature:IPsec]", g.Ordered, func() {
...
+				e2eskipper.Skipf("cluster is configured with IPsec %s mode, so skipping the test", ipsecMode)
+			}
+			g.By("creating test pods")
+			pods := createWebServerPods(oc, f.Namespace.Name)
+			g.By("checking crossing connectivity over the pods")
+			checkPodCrossConnectivity(pods)
+		})
+
+		g.It("check pod traffic are working across nodes after ipsec daemonset restart", func() {
+			if ipsecMode != v1.IPsecModeFull {
```
martinkennelly, line 679: why not do this for other modes?

test/extended/networking/ipsec.go
```diff
@@ -632,6 +653,117 @@ var _ = g.Describe("[sig-network][Feature:IPsec]", g.Ordered, func() {
...
+				o.Expect(err).NotTo(o.HaveOccurred())
+				g.By("checking crossing connectivity over the pods")
+				checkPodCrossConnectivity(pods)
+			}
+		})
+	})
+})
+
+func createWebServerPods(oc *exutil.CLI, namespace string) []corev1.Pod {
+	g.GinkgoHelper()
```
martinkennelly, line 699: what does this do?

test/extended/networking/ipsec.go
```diff
@@ -632,6 +653,117 @@ var _ = g.Describe("[sig-network][Feature:IPsec]", g.Ordered, func() {
...
+}
+
+func checkPodCrossConnectivity(pods []corev1.Pod) {
+	g.GinkgoHelper()
+	var testFns []func() error
+	for _, sourcePod := range pods {
+		for _, targetPod := range pods {
+			if sourcePod.Name == targetPod.Name {
+				continue
+			}
```
martinkennelly, line 756: should you make sure also to only try to ping pods that are on different nodes?

test/extended/networking/ipsec.go
```diff
@@ -632,6 +653,117 @@ var _ = g.Describe("[sig-network][Feature:IPsec]", g.Ordered, func() {
...
+			g.By("creating test pods")
+			pods := createWebServerPods(oc, f.Namespace.Name)
+			g.By("checking crossing connectivity over the pods")
+			checkPodCrossConnectivity(pods)
+			// Restart IPsec daemonset few times and check pod traffic is not impacted.
+			for i := 1; i <= 5; i++ {
+				g.By(fmt.Sprintf("attempt#%d restarting IPsec pods", i))
+				err := restartIPsecDaemonSet(oc)
+				o.Expect(err).NotTo(o.HaveOccurred())
```
martinkennelly, line 690: nit; i like to see the reasons why something failed with this expects otherwise you have to go look at the line of code that it failed on.

pkg/testsuites/standard_suites.go
pperiyasamy: moved this test to ipsec test suite, so timeout change is not needed here.

test/extended/networking/ipsec.go
```diff
@@ -224,6 +229,22 @@ func ensureIPsecDisabled(oc *exutil.CLI) error {
 		})
 }
 
+func restartIPsecDaemonSet(oc *exutil.CLI) error {
+	ds, err := getDaemonSet(oc, ovnNamespace, ovnIPsecDsName)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
pperiyasamy: making it as ginkgo helper function now, so this is not needed now.

test/extended/networking/ipsec.go
pperiyasamy: the previous tested is valid for all ipsec modes, but this one is valid for full mode, added a comment in the code, hope that helps.

test/extended/networking/ipsec.go
pperiyasamy: oh! yes martin, good catch. added `restartIPsecDaemonSet` as a test helper function, that will solve the problem that you described.

test/extended/networking/ipsec.go
pperiyasamy:
```
/*
GinkgoHelper marks the function it's called in as a test helper.  When a failure occurs inside a helper function, Ginkgo will skip the helper when analyzing the stack trace to identify where the failure occurred.

This is an alternative, simpler, mechanism to passing in a skip offset when calling Fail or using Gomega.
*/
```

test/extended/networking/ipsec.go
pperiyasamy: yes, the `sourcePod.Name == targetPod.Name`  check avoids the self ping.

pperiyasamy (commented):
As you suggested, move this change into another PR https://github.com/openshift/origin/pull/29563, let's continue our review there.
so i'm going to close this PR,

PR #29238, opened 2024-10-28
NO-JIRA: UserDefinedPrimaryNetworks: curl the KAPI IP through eth0

kyrtapz:
Instead of relying on routes use a specific interface for accessing KAPI.

cc: @trozet


trozet:
/retitle NO-JIRA: UserDefinedPrimaryNetworks: curl the KAPI IP through eth0

tssurya:
/retitle SDN-4930: UserDefinedPrimaryNetworks: curl the KAPI IP through eth0
for UDN PRs please use the EPIC number, thanks in advance!

PR #29240, opened 2024-10-28
SDN-4930: Set startup probe for UDN tests

trozet:
Hit downstream:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2314/pull-ci-openshift-ovn-kubernetes-master-e2e-gcp-ovn-techpreview/1849904878739001344

Taken from upstream:
https://github.com/ovn-org/ovn-kubernetes/pull/4811


tssurya:
/retitle SDN-4930: Set startup probe for UDN tests

4930 is the epic for UDN, we can use that so that PR's are tracked instead of NO-JIRA

PR #29264, opened 2024-11-04
CNF-15476: Network tests: add advanced routing capabilities tests

fedepaol:
Validating that CNO can deploy frr-k8s and that frr-k8s works properly:

We test that:

- pods are deployed and undeployed
- crds are deployed but not undeployed
- the configuration is reflected in the status CRD


fedepaol:
cc @jcaamano @trozet

PR #29276, opened 2024-11-07
MCO-1419: OCL E2E Testing

RishabhSaini:
ocl test:
Tests whether one can turn on and opt in a pool for OCL. This involves verifying if the custom image specified in the mosc is built and successfully rolled out the specified nodes within 20 mins

go.mod changes:
adds github.com/machine-config-operator 2024 version
which updates github.com/docker/docker from v20.10.27 -> v24.0.7
which breaks github.com/fsouza/go-dockerclient v1.7.1 since it relies on docker v20.10.27

Hence, I manually update both github.com/docker/docker to v27.3.1 and github.com/fsouza/go-dockerclient to v1.12.0 which is their latest version

This keeps go dependencies happy :)


sergiordlr:
Hello! In which platforms will the test be executed?

In the test we are using the internal registry to store the osImage. It means that the test may fail in the following installations:
- Clusters with the ImageRegistry capacity disabled (no internal registry)
- Clusters with the ImageRegistry capacity enabled, but not using a persistent volume for the internal registry

PR #29279, opened 2024-11-11
API-1802: chart template: show cert rotation events

vrutkovs:
Followup for https://github.com/openshift/origin/pull/28639:
* Add "Certificate Rotation" section to e2e-chart-everything.html
* Show cert rotation events from certrotationcontroller, csrapprover and ingress controller

### DIFF
--- a/e2echart/non-spyglass-e2e-chart-template.html
+++ b/e2echart/non-spyglass-e2e-chart-template.html
@@ -254,6 +254,7 @@
                         <option value="e2e_test_failed">e2e Test Failed</option>
                         <option value="e2e_test_flaked">e2e Test Flaked</option>
                         <option value="interesting_events">Interesting Events</option>
+                        <option value="certificate_rotation">Certificate Rotation</option>
                     </select>
 `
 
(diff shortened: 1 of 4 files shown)


dgoodwin:
This looks highly useful.  However this UI code is destined to be killed off, the new UI is at https://github.com/openshift/sippy/blob/main/sippy-ng/src/prow_job_runs/IntervalsChart.js and available under Debug Tools today. 

It is largely based on Source, and thus for what you're attempting here it would be best to have an explicit Source for your CertRotation. I would suggest adding a monitortest that scans the intervals already at the end of a run using the logic you have here, and just copies them to a new Source.  This keeps the UI logic cleaner as it no longer has to interpret the intervals as we see here.

If you'd like to also remove them from this chart in the meantime that's ok too, lets just do it based on an explicit Source instead of the UI interpreting.

vrutkovs:
>It is largely based on Source, and thus for what you're attempting here it would be best to have an explicit Source for your CertRotation

Multiple different controllers issue these events - do we need a separate monitortest for this if we already have all necessary intervals scanned?

dgoodwin:
I think a separate monitortest would be easiest, it does duplicate a handful of intervals but in the 30-60k range I'm not too worried about that. The separate monitortest scanning the intervals from those other controllers and assembling them into one source category would be easier than trying to update all the controllers. Honestly I've used Cursor for things like this and if described well enough, it could probably do it in a few minutes for you.

dgoodwin:
As you've discovered, new ui can also have handcrafted sections which I thought I had removed entirely. This is ok for now as written, though the use of a custom Source is cleaner and would avoid the need for dual impementations, as Sources can be used immediately with the new UI. I'll leave it up to you, but I don't consider the monitortest to copy to a new Source required at this point.

/lgtm
/hold

Cancel if you'd like to proceed with this as is.

PR #29265, opened 2024-11-05
MON-4072: Run MetricsCollectionProfiles tests on all platforms

rexagod:
Needed for https://github.com/openshift/api/pull/2084.

PR #29284, opened 2024-11-12
NO-JIRA: address `MetricsCollectionProfiles` failures on `metal` platform

rexagod:
Address `MetricsCollectionProfiles` failures on `metal` platform due to the absence of a CMO configuration.

***

Branched off of https://github.com/openshift/origin/pull/29265 since we do not want to run these tests on all platforms yet (that'll happen once this makes it into the default set of features).


rexagod:
/test e2e-gcp-ovn-techpreview

Mentioning here that it's not possible to spin up a 4.18 `metal` cluster through cluster-bot currently, or do a `/test e2e-metal-ipi-ovn-techpreview` (metal tech-preview jobs are not supported), so I believe the best way to see green in Sippy is to merge this.

PR #29286, opened 2024-11-12
TRT-1761: kubelet disk pressure

dgoodwin:
- **Add new intervals when a node starts/stops reporting disk pressure**
- **Flake a test if we see kubelet report DiskPressure**


pkg/monitortests/node/watchnodes/monitortest.go
```diff
@@ -119,3 +120,31 @@ func intervalStartDuring(needle monitorapi.Interval, haystack monitorapi.Interva
 	}
 	return false
 }
+
+func nodeDiskPressure(finalIntervals monitorapi.Intervals) []*junitapi.JUnitTestCase {
+	const testName = "[[Jira:\"Test Framework\"]] kubelet should not report DiskPressure"
```
stbenjam, line 125:
```suggestion
	const testName = "[Jira:\"Test Framework\"] kubelet should not report DiskPressure"
```

What's the double bracketing for?
image

PR #29141, opened 2024-09-27
WRKLDS-1492: Loosen output check in policy scc-subject-review, scc-review test

ardaguclu:
By this https://github.com/openshift/api/pull/2017 (and the corresponding changes in upstream), all resources have metadata field. When we bump the dependencies in oc, these changes are also coming in (https://github.com/openshift/oc/pull/1877). On the other hand, openshift-apiserver is not ready yet to understand the metadata field in resources and so it raises a warning about `Warning: unknown field \"metadata\"`. Although this is just a warning, this causes failures in the test `[sig-cli] policy scc-subject-review, scc-review` (e.g. [failed test run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-oc-1877-nightly-4.18-e2e-metal-ipi-ovn-ipv6/1836336155059032064)).

Ideally we should wait the merge of https://github.com/openshift/openshift-apiserver/pull/444 (which waits for https://github.com/openshift/openshift-apiserver/pull/441) and expect that problem disappears automatically. 

But these changes may not be merged in soon and this PR alternatively loosens the test check to unblock the oc bump 1.31. Once the changes in openshift-apiserver is merged, we can revert back to equality check.

PR #29289, opened 2024-11-13
OCPBUGS-42532: Revert loosened policy test

ardaguclu:
Since https://github.com/openshift/openshift-apiserver/pull/458 has merged, we can revert https://github.com/openshift/origin/pull/29141


ardaguclu:
/cc @stbenjam 
as the approver of the previous PR

atiratree:
/lgtm
@ardaguclu btw, it is better to use `git revert` for posterity

ardaguclu:
> /lgtm @ardaguclu btw, it is better to use `git revert` for posterity

You are right, that would be better. Next time, I'll use it :) and thanks for tagging.

ardaguclu:
No failure of the enabled tests, but anyway it is better to retest one more time;
/retest

PR #29302, opened 2024-11-18
disable e2e olmv1 incompatible operators

gallettilance:


test/extended/olm/olmv1.go
```diff
@@ -136,6 +136,8 @@ var _ = g.Describe("[sig-olmv1][OCPFeatureGate:NewOLM][Serial] OLMv1 operator in
 			})
 		o.Expect(err).NotTo(o.HaveOccurred())
 
+		g.Skip("Testing for incompatible operators is disabled")
```
joelanford, line 139: Maybe just remove this and the code that follows.

How do we even know that "no olm.maxopenshiftversion is specified" for the bundle that gets installed? It seems like it would be possible for quay-operator-3.13 to get that property in the future or for us to change that package name/version without realizing we needed to check its properties to see if they still do not specify this property.

tmshort: The operators are deleted between tests.

gallettilance: bundles are immutable and this specific version of quay doesn't have it

joelanford: > bundles are immutable and this specific version of quay doesn't have it

In theory. But you're right. I was more worried about the latter case of our future selves changing the package name or version and not knowing to read 30 lines down to see that we're making assumptions about the chosen bundle.

PR #29327, opened 2024-11-26
OCPBUGS-57049: TLS registry: refactor testcase annotations

vrutkovs:
Instead of specifying a testcase in every annotation this change requires a separate annotation for test case name. The reason being is that this test case will be repeated in annotations multiple times, as we need to verify that certkey pair is refreshed, regenerated after expiry and more. As a result, autoregenerate annotation now shorted to just "true" or "false".

This also adds a new "refresh period" annotation so that we could check which certificates are being rotated during normal cluster lifecycle


bertinatto (commented):
/approve

PR #29329, opened 2024-11-26
OCPBUGS-44970: Excluding loki prod-bearer-token container from excessive restarts test in SNO

jaypoulz:
This container is deployed by https://github.com/openshift/release/tree/master/ci-operator/step-registry/ipi/install/hosted-loki and is sensitive to resolving sso.redhat.com. During an SNO upgrade, there is a brief period of time where dns is unavailable because the only instance of the pod has been killed and the new pod is starting up. The goal of this change is to exclude this container from the excessive restart checks since, depending on the exact timing, it can restart enough times to trigger a failed test.

### DIFF
--- a/pkg/monitortests/node/legacynodemonitortests/exclusions.go
+++ b/pkg/monitortests/node/legacynodemonitortests/exclusions.go
@@ -35,6 +35,14 @@ func isThisContainerRestartExcluded(locator string, exclusion Exclusion) bool {
 			topologyToExclude: "single",
 		},
 		{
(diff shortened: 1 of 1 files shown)


jaypoulz:
/hold

I really don't like this fix. But it's here in case this test creates too much noise in the blocking SNO upgrade lane.
See https://redhat-internal.slack.com/archives/C01CQA76KMX/p1732646630858199?thread_ts=1732543112.314749&cid=C01CQA76KMX for follow up.

PR #29330, opened 2024-11-27
TRT-1898: Revert "NO-JIRA: address malformed CMO configmap post-test"

xueqzhan:
Reverts openshift/origin#29309; tracked by https://issues.redhat.com/browse/OCPBUGS-45153

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are reverting this breaking change to get CI and/or nightly payloads flowing again.

This PR has causd the nightly failures for many techpreview jobs. Here is one example: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview/1861677512929579008

Test is stuck with this error:

  I1127 09:11:20.327512 62130 client.go:421] Project "e2e-test-monitoring-collection-profiles-mqvzf" has been fully provisioned.
    STEP: restoring the original configuration for the operator @ 11/27/24 09:11:21.232
    [INTERRUPTED] in [AfterAll] - github.com/openshift/origin/test/extended/prometheus/collection_profiles.go:99 @ 11/27/24 09:26:06.355
    ------------------------------
    Interrupted by User
    First interrupt received; Ginkgo will run any cleanup and reporting nodes but will skip all remaining specs.  Interrupt again to skip cleanup.
    Here's a current progress report:
      [sig-instrumentation][OCPFeatureGate:MetricsCollectionProfiles] The collection profiles feature-set initially, in a homogeneous default environment, should expose default metrics [Suite:openshift/conformance/parallel] (Spec Runtime: 14m49.846s)
        github.com/openshift/origin/test/extended/prometheus/collection_profiles.go:140
        In [AfterAll] (Node Runtime: 14m45.187s)
          github.com/openshift/origin/test/extended/prometheus/collection_profiles.go:99
          At [By Step] restoring the original configuration for the operator (Step Runtime: 14m45.123s)
            github.com/openshift/origin/test/extended/prometheus/collection_profiles.go:104

          Spec Goroutine
          goroutine 236 [select]
            github.com/onsi/gomega/internal.(*AsyncAssertion).match(0xc000fa2070, {0x7e574e8, 0xc23fb80}, 0x1, {0x0, 0x0, 0x0})
              github.com/onsi/gomega@v1.34.2/internal/async_assertion.go:538
[...]

PR #29331, opened 2024-11-28
Update extended prometheus tests

theobarberbany:
Use komega + `Eventually` to avoid 409s in the extended prometheus tests. This way if we fail with a 409, we will retry until the default `Eventually` timeout.

We've been seeing [consistent](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/238/pull-ci-openshift-cluster-capi-operator-main-e2e-gcp-ovn-techpreview/1861783509195034624) [failures](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/238/pull-ci-openshift-cluster-capi-operator-main-e2e-aws-ovn-techpreview/1861666923993894912) of our [techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/238/pull-ci-openshift-cluster-capi-operator-main-e2e-azure-ovn-techpreview/1861667167863312384) e2es, which has been blocking out ability to merge. This looks like a particularly bad flake, as with enough retests eventually we go green. 

cc/ @rexagod as I think you've touched this last :)


rexagod:
Thank you for looking into this, @theobarberbany. I did a bit of digging myself and ended up putting together https://github.com/openshift/origin/pull/29333.

To provide some context, the issue, at its core, seems to be a race between handling CMO's configmap between different specs. Since, by default, specs are run in parallel, the tests are `o.Eventually` able to get a hold of the configmap (the long green streaks signify when they start trying to get a hold of it, whereas the actual test is only a minor % of it that happens when they do).

### Parallel specs

Screenshot 2024-11-30 at 2 26 54 AM

### Serial specs

Screenshot 2024-11-30 at 2 25 49 AM

Owing to this, parallelized specs are more prone to flake/fail if a certain operation is not waited for "eventually" (`Update`, in case of this patch). Serializing would ensure we don't need to rely on waiting (as evident from the screenshots above), while cutting down the run-time significantly.

All that being said, as time is of the essence, I'm happy to see this go in to fix the tests as it is right now, so folks are unblocked, and I can follow-up on this with these changes once things are more stable again.

Thanks again for looking into this! 🍀

rexagod:
Ah, looking at the failing workflows linked in the description, I believe the revert PR https://github.com/openshift/origin/pull/29330 may have already fixed them for you. Can you retest?

theobarberbany:
Ah fab, looking at job history it seems that the revert has helped! :)

Thanks!

JoelSpeed:
Is this PR still required?

theobarberbany:
Going to go with no. Closing.

PR #29339, opened 2024-12-03
OCPBUGS-43777: apiserver: add tests to verify kube-apiserver is accessible via different endpoints

vrutkovs:
Add tests which check that kube-apiserver is reachable via service network, external and internal api endpoints.

In standarf e2e test flow this is being verified by monitor test, but cert rotation tests are disruptive and can't observe the cluster when time is being skewed to emulate age. As a result, these tests need to be oneshot verification tests

### DIFF
--- a/test/extended/apiserver/access.go
+++ b/test/extended/apiserver/access.go
@@ -1,0 +1,89 @@
+package apiserver
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"regexp"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/image"
+)
+
+// These tests are duplicating check-endpoints and monitor test functionality. They are required
+// for cert rotation suites where running monitor alongside test is impossible as we need to
+// skew time to emulate the age of the cluster which is disruptive.
+var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] kube-apiserver should be accessible via", func() {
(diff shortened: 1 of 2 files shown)


test/extended/apiserver/access.go
```diff
@@ -0,0 +1,79 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			if apiPath == "api-int" || apiPath == "api-ext" {
+				externalAPIUrl, err := url.Parse(infra.Status.APIServerURL)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				if apiPath == "api-ext" {
+					apiPath = externalAPIUrl.Host
+				} else {
+					apiPath = strings.Replace(externalAPIUrl.Host, "api.", "api-int.", 1)
+				}
```
everettraven, line 52: Is `api-int.*` always going to resolve to the internal API server endpoint? 

Would pulling the internal API server URL using [`infra.Status.APIServerInternalURL` ](https://github.com/openshift/api/blob/c1fdeb0788c16659238d93ec45ce2e798c14eb88/config/v1/types_infrastructure.go#L89-L93) be more reflective of the flow that a user/client may attempt to take if they were to look at the `Infrastructure` API?

vrutkovs: Yes, good point. iiuc Hypershift may use a different URL for internal URL

vrutkovs: Fixed in e1fd7a84533

everettraven (approved):
Changes look reasonable to me.

/lgtm

everettraven (approved):
/lgtm

test/extended/apiserver/access.go
```diff
@@ -0,0 +1,86 @@
...
+	"k8s.io/client-go/util/retry"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/image"
+)
+
+var _ = g.Describe("[Conformance][sig-api-machinery][Feature:APIServer] kube-apiserver should be accessible via", func() {
```
deads2k, line 24: Can you add a comment describing the need for a specific one-off test to verify that the certificates function properly after rotation as you indicated in slack?  It's not obvious. After that I think it's good to approve.

vrutkovs: Done

PR #29348, opened 2024-12-05
SDN-5568: only provision workloads when network creation has started

maiqueb:


test/extended/networking/livemigration.go
```diff
@@ -188,13 +199,14 @@ var _ = Describe("[sig-network][OCPFeatureGate:PersistentIPsForVirtualization][F
...
-					Expect(err).NotTo((HaveOccurred()))
+					Expect(err).NotTo(HaveOccurred())
 				}),
-				Entry("UserDefinedNetwork", func(c networkAttachmentConfigParams) {
+				Entry("UserDefinedNetwork", func(c networkAttachmentConfigParams) networkAttachmentConfig {
 					udnManifest := generateUserDefinedNetworkManifest(&c)
 					By(fmt.Sprintf("Creating UserDefinedNetwork %s/%s", c.namespace, c.name))
 					Expect(applyManifest(c.namespace, udnManifest)).To(Succeed())
 					Expect(waitForUserDefinedNetworkReady(c.namespace, c.name, udnCrReadyTimeout)).To(Succeed())
+					return networkAttachmentConfig{networkAttachmentConfigParams{networkName: c.namespace + "." + c.name}}
```
maiqueb, line 209: @ormergi how can I have access to the name of the network ? 

Right now, the only thing I think I can do is to read the corresponding NAD, unmarshal, and return that ... 

Is there something simpler ?

maiqueb: @npinaeva how about you ?

maiqueb: Fetched it from the NAD generated for the UDN.

I think we can assume the name is the same (i.e. NAD.metadata.Name == UDN.metadata.Name) 

If that changes, we'll need to be able to derive it somehow. Not thinking much about that, because the change must be backwards compatible. 

@npinaeva

deads2k:
/hold

fix the product, not the test.

test/extended/networking/livemigration.go
ormergi: I would fetch the NAD that created by the UDN controller and parse the network-name it got.
Doing so ensure tests wont break even if network-name will be changed, no need to go back to tests and do some adjustments.

test/extended/networking/livemigration.go
```diff
@@ -462,3 +477,80 @@ func checkEastWestTraffic(virtClient *kubevirt.Client, vmiName string, podIPsByN
...
+		match := re.FindSubmatch([]byte(outReader.Text()))
+		if len(match) != 2 {
+			continue
+		}
+		podName = string(match[1])
+		break
+	}
+	if podName == "" {
+		return false, fmt.Errorf("could not find a valid ovnkube-node pod on node '%s'", nodeName)
+	}
```
ormergi, line 507: It seems this code looks for ovnkube-node pod name, cant it be done using the programmatic go-client (similar to line 508)?

maiqueb: It can, but I just copied this code from somewhere else - I still need to make this coalesce.

And still, I guess staff engineering opposes this PR: https://github.com/openshift/origin/pull/29348#issuecomment-2532208445

test/extended/networking/livemigration.go
maiqueb: Right, that would make sense. I have the owner reference for that, right ? @ormergi

ormergi: Yes, you can also fetch the NAD with the meta.name as the UDN object as well

deads2k:
> And still, I guess staff engineering opposes this PR: https://github.com/openshift/origin/pull/29348#issuecomment-2532208445

How about creating a copy of this test that is showing the UDN problem and making it minimum reproducer. Then gate that test on UDN. This ensures that the failure mode is obvious and we don't lose the signal.  With the signal preserved, it is then ok to adjust other tests to compensate for a product bug because the signal won't be lost and will block the promotion of UDN.

maiqueb:
The UDN "smoke signal" test was executed and passed in https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29348/pull-ci-openshift-origin-master-e2e-aws-ovn-virt-techpreview/1868669031075024896

test/extended/networking/livemigration.go
maiqueb: Refactored the code. At least now it's clear I'm re-using existent code that can be improved on a follow-up.

PR #29353, opened 2024-12-06
extended: add minimum kubelet version test

haircommander:


test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,273 @@
...
+	kubeletVersion string
+	testFunc       func()
+}
+
+var _ = g.Describe("[sig-node][OCPFeatureGate:MinimumKubeletVersion] [Serial]", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		nodeImpersonatingClient clientset.Interface
+		nodeName                = "fakenode"
```
kannon92, line 56: It may be worth having a random value for fakenode so we don't get conflicts.

You do clean it up so we may not get any issues.

haircommander: yeah honestly I feel like a conflict would be better because that means we'd catch if the test didn't clean up after itself or is suddenly not running serially

kannon92:
Code looks good to me!

But the jobs seem unhappy.

haircommander:
/retest
I don't believe any of these should be related to the new tests

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,273 @@
...
+		waitForAPIServerRollout(newKasStatus.Status.LatestAvailableRevision, operatorClient)
+	}
+}
+
+func updateMinimumKubeletVersion(oc *exutil.CLI, version string, expectedErr bool) func() {
+	nodesConfigOrig, err := oc.AdminConfigClient().ConfigV1().Nodes().Get(context.Background(), "cluster", metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	if nodesConfigOrig.Spec.MinimumKubeletVersion == version {
+		return nil
```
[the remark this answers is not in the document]
haircommander: not short circuting here means we still call Update() and do a bunch of needless waiting, so we need to branch in some way here to save time

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,273 @@
...
+					_, err := nodeImpersonatingClient.AuthorizationV1().SubjectAccessReviews().Create(context.Background(), sar, metav1.CreateOptions{})
+					o.Expect(err).NotTo(o.HaveOccurred())
+				},
+			},
+			{
+				testName:       "should block node from listing pods if too old",
+				kubeletVersion: "v1.29.0",
+				testFunc: func() {
+					_, err := nodeImpersonatingClient.CoreV1().Pods(f.Namespace.Name).List(context.Background(), metav1.ListOptions{})
+					o.Expect(err).To(o.HaveOccurred())
```
p0lyn0mial, line 123: worth checking if we received the expected error.

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,273 @@
...
+		oc                      = exutil.NewCLIWithoutNamespace("minimum-kubelet-version")
+	)
+
+	g.BeforeEach(func() {
+		g.DeferCleanup(updateMinimumKubeletVersionAndWait(oc, minimumVersion))
+
+		ginkgo.By("Creating a kubernetes client that impersonates a node")
+		config, err := framework.LoadConfig()
+		framework.ExpectNoError(err, "failed to load kubernetes client config")
+		config.Impersonate = restclient.ImpersonationConfig{
```
p0lyn0mial, line 69: good idea!

p0lyn0mial (commented):
Are you planning to add tests for the admission bits as well?

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,273 @@
...
+	var (
+		nodeImpersonatingClient clientset.Interface
+		nodeName                = "fakenode"
+		asUser                  = nodeNamePrefix + nodeName
+		minimumVersion          = "1.30.0"
+		f                       = framework.NewDefaultFramework("minimum-kubelet-version")
+		oc                      = exutil.NewCLIWithoutNamespace("minimum-kubelet-version")
+	)
+
+	g.BeforeEach(func() {
```
p0lyn0mial, line 63: I think this is per `g.It`

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,273 @@
...
+			{
+				testName:       "should be able to get node",
+				kubeletVersion: "v1.29.0",
+				testFunc: func() {
+					_, err := nodeImpersonatingClient.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+					o.Expect(err).NotTo(o.HaveOccurred())
+				},
+			},
+			{
+				testName:       "should be able to perform subjectaccessreviews",
```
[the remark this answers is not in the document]
haircommander: I like having the cases covered personally and it barely adds any time to the test per

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,273 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		nodesConfigCurrent.Spec = *nodesConfigOrig.Spec.DeepCopy()
+
+		_, err = oc.AdminConfigClient().ConfigV1().Nodes().Update(context.Background(), nodesConfigCurrent, metav1.UpdateOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+	}
+}
+
+func waitForAPIServerRollout(previousLatestRevision int32, operatorClient operatorv1client.Interface) {
```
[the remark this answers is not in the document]
haircommander: this function segfaults for some reason and also doesn't seem to wait for the inital api server to rollout so we may hit races with the rollout. I'd prefer this way I think

test/extended/node/minimum_kubelet_version.go
haircommander: good idea!

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,276 @@
...
+	desiredTestDuration = 25 * time.Minute
+)
+
+var _ = g.Describe("[sig-node][OCPFeatureGate:MinimumKubeletVersion] ", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLIWithoutNamespace("minimum-kubelet-version")
+
+	g.DescribeTable("admission", func(version string, expectedErr bool) {
+		defer updateMinimumKubeletVersion(oc, version, expectedErr)
```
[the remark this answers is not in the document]
haircommander: for these admission tests we don't need the rollout as we're just checking whether we can change the value or not. Below we pair the update with a wait to make sure we deny the node access if appropriate

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,276 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+const (
+	nodesGroup          = "system:nodes"
+	nodeNamePrefix      = "system:node:"
+	desiredTestDuration = 25 * time.Minute
+)
+
+var _ = g.Describe("[sig-node][OCPFeatureGate:MinimumKubeletVersion] ", func() {
```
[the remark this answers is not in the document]
haircommander: I haven't seem them step on each other in a way that implies that but I'm open to it if you think it's better.

test/extended/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,272 @@
...
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLIWithoutNamespace("minimum-kubelet-version")
+
+	g.DescribeTable("admission", func(version string, handleErrFunc func(error)) {
+		defer updateMinimumKubeletVersion(oc, version, handleErrFunc)
+	},
+		g.Entry("should allow an empty minimum kubelet version", "", expectNoError),
+		g.Entry("should allow an old minimum kubelet version", "1.30.0", expectNoError),
```
p0lyn0mial, line 40: Maybe I don't understand something, but I think this case will set a new minimum version, which will cause a new revision. 

In that case, I think we should ensure that a parallel test (if this test runs in parallel with others) won't observe a "partial state". I think that we should make sure the servers are stable.

test/extended/node/minimum_kubelet_version.go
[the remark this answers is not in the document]
p0lyn0mial: see https://github.com/openshift/origin/pull/29353/files#r1988587538

test/extended/node/minimum_kubelet_version.go
haircommander: good point. I have replaced those two with one other failing test. both 1.30.0 and "" will be set in the authorization test anyway, so I think this is better. now the failure cases will deny and can be run in parallel

p0lyn0mial:
@haircommander are we merging this tests ?

kannon92:
@haircommander is this still relevant?

PR #29354, opened 2024-12-09
TRT-1575: Use catchall card for operator degraded exceptions for major operators

xueqzhan:


neisw:
By removing the reason checks doesn't that mean that anytime the operator goes degraded it will now get an exception and flake the [test](https://github.com/openshift/origin/blob/678dab7de111723253c5f67a05ded69317cd29c7/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go#L575) for all reasons?

neisw:
> By removing the reason checks doesn't that mean that anytime the operator goes degraded it will now get an exception and flake the [test](https://github.com/openshift/origin/blob/678dab7de111723253c5f67a05ded69317cd29c7/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go#L575) for all reasons?

Ken pointed out that the default is currently to [flake](https://github.com/openshift/origin/blob/200ae23697754f638e56bb662fd76aebdb5015db/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go#L438) so this is just changing to use operator specific bugs for all reasons instead of select ones and falling back to the TRT one.

PR #29365, opened 2024-12-11
STOR-2136: Remove CSI snapshot webhook conditions checks

gnufied:
We don't need these conditions, because we are no longer deploying snapshot webhook stuff in OCP.

See - https://github.com/openshift/cluster-csi-snapshot-controller-operator/pull/219


jsafrane:
/lgtm

/approve
(just in case I can)

PR #29367, opened 2024-12-11
TRT-1803: Wait for operators to become stable before test

xueqzhan:


pkg/clioptions/clusterinfo/wait_for_stable_cluster.go
```diff
@@ -0,0 +1,133 @@
...
+		}, err
+	}
+
+	interval := 10 * time.Second
+	timeout := 10 * time.Minute
+	minimumStablePeriod := 3 * time.Minute
+	recoveredOperators := map[string]string{}
+	unstableOperators := map[string]string{}
+	var stabilityStarted *time.Time
+	waitErr := wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(waitCtx context.Context) (bool, error) {
```
pacevedom, line 66: You may want to skip this one for MicroShift, as it does not have config apigroup.

xueqzhan: Good catch! Thanks

PR #29370, opened 2024-12-12
print the error result from UDN test failure

JacobTanenbaum:
currently we do not print out the result if the connection does not isolate traffic for UDN tests. This does not show us what the pod is connecting to and as such don't know if there are artifacts of the old pods around or if the traffic is not isolated.

Print out the result of the query so that we have more insight as to how the test failed.

### DIFF
--- a/test/extended/networking/network_segmentation.go
+++ b/test/extended/networking/network_segmentation.go
@@ -504,7 +504,7 @@ var _ = Describe("[sig-network][OCPFeatureGate:NetworkSegmentation][Feature:User
 								//don't try with your own IP
 								continue
 							}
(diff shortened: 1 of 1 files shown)


JacobTanenbaum:
/payload-job periodic-ci-openshift-release-master-ci-4.18-e2e-gcp-ovn-techpreview
/payload-job periodic-ci-openshift-release-master-ci-4.18-e2e-gcp-ovn-techpreview
/payload-job periodic-ci-openshift-release-master-ci-4.18-e2e-gcp-ovn-techpreview
/payload-job periodic-ci-openshift-release-master-ci-4.18-e2e-gcp-ovn-techpreview

PR #29374, opened 2024-12-13
SDN-5566: virt, udn: primary UDN requires the net seg FG

maiqueb:
Mark the primary UDN tests as requiring the network segmentation FG.


trozet (approved):
/lgtm

PR #29392, opened 2024-12-30
OCPBUGS-47700: images: Fix layers test to work with manifestlisted imagestreams

Prashanth684:
Look at the layers in the submanifest rather than the top level manifest if the imagestream is manifestlisted. Refer to the bug for details.

This fixes techpreview jobs for clusters installed with the multi-payload. Example of a failure: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-multiarch-master-nightly-4.19-ocp-e2e-aws-ovn-techpreview-multi-a-a/1873687894279852032

### DIFF
--- a/test/extended/images/layers.go
+++ b/test/extended/images/layers.go
@@ -24,6 +24,26 @@ import (
 	"github.com/openshift/origin/test/extended/util/image"
 )
 
+func verifyLayerData(layers *imagev1.ImageStreamLayers, name string) imagev1.ImageBlobReferences {
+	l, ok := layers.Images[name]
+	o.Expect(ok).To(o.BeTrue())
+	if l.ImageMissing {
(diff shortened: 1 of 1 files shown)


flavianmissi:
Test failures seem unrelated.
/retest
/lgtm

Prashanth684:
@flavianmissi added a change to the logic - we noticed that for clusters of non-x86 architectures running the multi payload, comparing the x86 manifest is not the correct approach - so now each of the submanifest layers are matched to see if one of them match to ensure that it works on any architecture cluster (not just x86).

Prashanth684:
unit failure seems unrelated - happening on other PRs as well

djslavens (approved):
/lgtm

PR #29397, opened 2025-01-06
OCPBUGS-48126: kubeapiserver auditloganalyzer: show resources updated too often

vrutkovs:
Along with users issuing too many updates its useful to list resources being updated often. Several users may update a single resource, so they may not show up in existing tests.

This PR adds another set of flaking tests `resource <group>.<resource>/<name> [-n <namespace>] has been updated too often` with error message similar to `resource <group>.<resource>/<name> [-n <namespace>] had <n> applies, check the audit log and operator log to figure out why  details in audit log`

### DIFF
--- a/pkg/monitortests/kubeapiserver/auditloganalyzer/handle_excessive_applies.go
+++ b/pkg/monitortests/kubeapiserver/auditloganalyzer/handle_excessive_applies.go
@@ -90,7 +90,7 @@ func (s *excessiveApplies) HandleAuditLogEvent(auditEvent *auditv1.Event, beginn
 	if !ok {
 		users = map[string]int{}
 	}
-	users[auditEvent.User.Username] = users[auditEvent.User.Username] + 1
+	users[auditEvent.User.Username] += 1
 	s.namespacesToUserToNumberOfApplies[nsName] = users
 
 	obj := auditEvent.ObjectRef
--- a/pkg/monitortests/kubeapiserver/auditloganalyzer/handle_excessive_applies.go
+++ b/pkg/monitortests/kubeapiserver/auditloganalyzer/handle_excessive_applies.go
@@ -1,21 +1,70 @@
 package auditloganalyzer
 
 import (
+	"fmt"
+	"sort"
+	"strings"
+	"sync"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
 	"k8s.io/apiserver/pkg/authentication/serviceaccount"
-	"strings"
-	"sync"
 )
+
+type userApplyCount struct {
+	User    string
+	Applies int
+}
+
+type numberOfAppliesWithExamples struct {
+	numberOfApplies   int
+	users             map[string]int
+	firstAuditEventID string
+	lastAuditEventID  string
+	firstOccurredAt   metav1.Time
+	lastOccurredAt    metav1.Time
+}
+
+func (n numberOfAppliesWithExamples) toErrorString() string {
+	appliesPerUser := []userApplyCount{}
+	for k, v := range n.users {
+		appliesPerUser = append(appliesPerUser, userApplyCount{k, v})
+	}
+	sort.Slice(appliesPerUser, func(i, j int) bool {
+		return appliesPerUser[i].Applies > appliesPerUser[j].Applies
+	})
+	topUsernames := []string{}
+	maxUsernames := 5
+	if len(appliesPerUser) < maxUsernames {
+		maxUsernames = len(appliesPerUser)
+	}
+	for i := 0; i < maxUsernames; i++ {
+		topUsernames = append(topUsernames, appliesPerUser[i].User)
+	}
+
+	return fmt.Sprintf(`
+Time: started %s, finished at %s
+Top 5 usernames: %s
+Audit Log IDs: %s ... %s
+`,
+		n.firstOccurredAt,
+		n.lastOccurredAt,
+		strings.Join(topUsernames, ", "),
+		n.firstAuditEventID,
+		n.lastAuditEventID,
+	)
+}
 
 type excessiveApplies struct {
 	lock                              sync.Mutex
 	namespacesToUserToNumberOfApplies map[string]map[string]int
+	resourcesToNumberOfApplies        map[string]numberOfAppliesWithExamples
 }
 
 func CheckForExcessiveApplies() *excessiveApplies {
 	return &excessiveApplies{
 		namespacesToUserToNumberOfApplies: map[string]map[string]int{},
+		resourcesToNumberOfApplies:        map[string]numberOfAppliesWithExamples{},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
```diff
@@ -255,6 +256,28 @@ func (w *auditLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Con
 
 	}
 
+	for resource, numberOfApplies := range w.excessiveApplyChecker.resourcesToNumberOfApplies {
+		if numberOfApplies < 200 {
+			continue
```
dgoodwin, line 261: A test must always produce a result if it ran, so if you're doing a test per resource, it must return a success junit, otherwise all the stats in sippy and component readiness will go sideways and think it always flakes/fails.

With that in mind, how many tests are we talking here when that is corrected? It would be a massive number wouldn't it? If so, this may have to become a more generic single test.

pkg/monitortests/kubeapiserver/auditloganalyzer/handle_excessive_applies.go
```diff
@@ -43,4 +47,22 @@ func (s *excessiveApplies) HandleAuditLogEvent(auditEvent *auditv1.Event, beginn
...
+	}
+	if obj.APIGroup != "" {
+		resource = fmt.Sprintf("%s.%s", obj.APIGroup, resource)
+	}
+	objApplies, ok := s.resourcesToNumberOfApplies[resource]
+	if !ok {
+		objApplies = 0
+	}
+	objApplies++
+	s.resourcesToNumberOfApplies[resource] = objApplies
```
dgoodwin, line 67: If you were to map resource to a slice of the actual messages for the audit events, you could output those in the test output and save the user a big debugging step, while still maintaining the ability to check the count via len().

dgoodwin (changes requested):

pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
vrutkovs: Yes, it would add hundred of tests if we create a test for each resource type. Since resource is already part of the error message, I'll create one test to rule them all

pkg/monitortests/kubeapiserver/auditloganalyzer/handle_excessive_applies.go
vrutkovs: If the controller is hotlooping it may produce thousands of audit events. Lets store up to 10 audit event IDs so that the user could find those easier

dgoodwin: 10 doesn't really cut it if we're failing at 200, it may not be sufficient to show the problem actor. I would suggest maybe 200-1000 as the upper limit per resource that fails. That extra text in a junit failure output doesn't sound particularly worrisome.

In this location, I would not filter at all, just add them all, and move the limiting to the location where we generate junit output. This eliminates the need to define a struct, your count is the length of the slice.

I would also not just put the audit IDs in, I would suggest putting more info from the audit event with a fmt.Sprintf. This will likely mean the person who hits the failure can immediately see what the problem was, right in the junit output. Serializing the full audit event looks a little too big, but interesting fields like User and UserAgent, maybe some others are worth including, whatever would save the user the trouble of hunting out the audit log given we've already got the info here.

pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
```diff
@@ -255,6 +256,40 @@ func (w *auditLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Con
 
 	}
 
+	testName := "resources has been updated not too often"
```
dgoodwin, line 259: suggest: [Jira:"kube-apiserver"] API resources are not updated excessively

pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
```diff
@@ -255,6 +256,40 @@ func (w *auditLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Con
...
+	flakes := []string{}
+	for resource, applies := range w.excessiveApplyChecker.resourcesToNumberOfApplies {
+		if applies.numberOfApplies < 200 {
+			continue
+		}
+		errorMessage := fmt.Sprintf("resource %s had %d applies, audit log ids: %s", resource, applies.numberOfApplies, strings.Join(applies.auditEventIDs, ", "))
```
dgoodwin, line 265: If the above is uncontroversial, you should probably improve this output a little more to show the resource, then all the audit event data we captured for it one per line with some indentation, and then move on to the next resource.

vrutkovs: I don't think its needed - alongside with printing all found events. Audit logs are verbose, so printing all 100+ events in junit would be excessive. There are better tools to analyze it - https://github.com/vrutkovs/audit-span for instance. 
imo printing 10 sample audit event IDs would be sufficient to find the rest of them - and have the details viewed in a better tool

dgoodwin: This may be true for an engineer on the control plane who knows of such tools, I don't think it's true for many others though who might see this in a presubmit once it can fail. Failing if you cross 200, then providing ten audit IDs which may not even originate from whatever is out of control, is not sufficiently helpful. I really would like the tests to give a less familiar user a chance to understand and not just ignore it and retest, or have to reach out to control plane team for help.

When this fires, am I correct that we know the resource already via the test output, we're interested in who issues and when. If I can see those three bits of data, it's going to be a pretty quick spot to see what happened no? 

Alternatively, could you breakdown the count by callers and include the top say 5 who issued the requests?

Either of these approaches means most users could hit this failure and immediately see what's going on.

vrutkovs: >When this fires, am I correct that we know the resource already via the test output, we're interested in who issues and when

We'll see the resource, but we won't see who and when. Its not trivial to display who updated it too often - it might have been continuous and at a steady rate (some operator is hotlooping) - or it could have been a spike (some action caused operator to update it too often). 
Also, in most cases its several users updating the operator - i.e. several controllers update different conditions in status.  

> could you breakdown the count by callers and include the top say 5 who issued the requests?

We have a test for that - `users in ns/%s must not produce too many applies`.

I think top 5 users is implementable, not sure how to display when updates happened though - maybe print first and last occurrence?

dgoodwin: If it could print out the top offenders during the whole CI run, and that is likely going to point to the problem, I could live without timestamps.

vrutkovs: Now looks like this:
```
{resource operator.openshift.io.openshiftcontrollermanagers/cluster had 968 applies, 
Time: started 2025-01-17 14:55:31.685374 +0000 UTC, finished at 2025-01-17 15:19:29.636384 +0000 UTC
Top 5 usernames: system:serviceaccount:openshift-controller-manager-operator:openshift-controller-manager-operator
Audit Log IDs: b936a56b-3e6f-4f76-afda-3ce1809fb830 ... 0f6db36b-930c-498e-8297-4f4745336a8c

resource operator.openshift.io.olms/cluster had 1374 applies, 
Time: started 2025-01-17 15:02:50.873214 +0000 UTC, finished at 2025-01-17 15:18:53.063365 +0000 UTC
Top 5 usernames: system:serviceaccount:openshift-cluster-olm-operator:cluster-olm-operator
Audit Log IDs: 63ac043b-2aae-4ada-ac09-f4962f76b9a3 ... 4941d6a6-f44e-457b-906c-e9e09e687e1b
  details in audit log}
```

dgoodwin:
Looks great, thanks.

/lgtm

PR #29398, opened 2025-01-06
OCPBUGS-47796: fix year for string compare

kannon92:
For the machine monior tests we compare directly against the error message. The year is included there so we will inject 2024 for those tests.

This way the tests won't fail in 2025.

For the lease tests we don't actually use the year so we can use the current year without any issue.

Note:

I am not sure if this is also failing for 4.17..


xueqzhan:
/lgtm

Thanks for fixing this.

kannon92:
@xueqzhan 

should we override some of these jobs to get the required unit tests passing?

PR #29404, opened 2025-01-07
NO-JIRA: tls: update instructions on how to update TLS registry

vrutkovs:
Add step-by-step instructions on how to update TLS registry

### DIFF
--- a/tls/README.md
+++ b/tls/README.md
@@ -184,6 +184,15 @@ Apart from reports the registry is checked for requirement violations.
   that don't have ownership annotation set. This file is meant to be "remove-only", meaning adding 
   new entries is prohibited. This is enforced by using a separate `OWNERS` file for this directory and an e2e test (see below).
 
+## Updating TLS registry
+
+In order to include unregistered TLS artifact or update certificate metadata fresh raw TLS info 
+needs to be placed in origin's `tls/raw-data`. Raw TLS info can be obtained from test artifacts, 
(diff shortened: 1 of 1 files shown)


tls/README.md
```diff
@@ -184,6 +184,15 @@ Apart from reports the registry is checked for requirement violations.
   that don't have ownership annotation set. This file is meant to be "remove-only", meaning adding 
   new entries is prohibited. This is enforced by using a separate `OWNERS` file for this directory and an e2e test (see below).
 
+## Updating TLS registry
+
+In order to include unregistered TLS artifact or update certificate metadata fresh raw TLS info 
+needs to be placed in origin's `tls/raw-data`. Raw TLS info can be obtained from test artifacts,
```
tmshort, line 190: There are currently files in `tls/raw-data`, does this process include updating those? i.e. should we commit overwrite the appropriate file with the `rarwTLSInfo` or add `rawTLSInfo` as a new file? Do we commit files from `tls/raw-data`?

vrutkovs: >should we overwrite the appropriate file with the rarwTLSInfo or add rawTLSInfo as a new file?

Ideally you'd overwrite matching file if this adds a new certificate. 

However, if you're updating metadata for existing file it needs to be consistent across other raw TLS files. This means you'd need to run all tests for all platforms and their techpreview variants. 

> Do we commit files from tls/raw-data? Or are they temporary?

Yes, these files need to be committed (so that `make verify` would check that reports / violations are generated correctly)

PR #29414, opened 2025-01-09
NO-JIRA: Add test case for checking EgressFirewall DNS names in caps

arkadeepsen:
This PR adds test case for checking EgressFirewall DNS names in caps.


arkadeepsen:
/retitle NO-JIRA: Add test case for checking EgressFirewall DNS names in caps

npinaeva (approved):

PR #29433, opened 2025-01-13
start tracking number of seconds without an etcd leader

deads2k:
Basic tracking for audit annotations in https://github.com/openshift/kubernetes/pull/2170.

Could clearly add
1. intervals showing when the no leader message happens
2. number of seconds without a leader for upload and later display in grafana


deads2k:
Well this PR definitely didn't break install

/retest

PR #29434, opened 2025-01-14
NO-JIRA: fix: increase watch count for monitoring operator

eggfoobar:
Increasing watch count for monitoring operator 110% above p95 for SNO 4.19 runs

(P95 * 1.1)/2 =~ 101

```
SELECT
  APPROX_QUANTILES(WatchRequestCount, 100)[OFFSET(50)] AS Median,
  APPROX_QUANTILES(WatchRequestCount, 100)[OFFSET(95)] AS Percentile_95,
  APPROX_QUANTILES(WatchRequestCount, 100)[OFFSET(99)] AS Percentile_99
FROM `openshift-ci-data-analysis.ci_data_autodl.operator_watch_requests` AS Watches
  INNER JOIN openshift-gce-devel.ci_analysis_us.jobs AS JobRuns
    ON JobRuns.prowjob_build_id = Watches.JobRunName
  INNER JOIN openshift-ci-data-analysis.ci_data.JobsWithVariants AS Jobs
    ON Jobs.JobName = JobRuns.prowjob_job_name
WHERE
JobRuns.prowjob_job_name LIKE "%4.19%"
AND Watches.Operator = "cluster-monitoring-operator"
AND Watches.ControlPlaneTopology = "SingleReplica"
AND Watches.PlatformType = "AWS"
```

### DIFF
--- a/test/extended/apiserver/api_requests.go
+++ b/test/extended/apiserver/api_requests.go
@@ -340,7 +340,7 @@ var _ = g.Describe("[sig-arch][Late]", func() {
 				"cluster-autoscaler-operator":            44,
 				"cluster-baremetal-operator":             125.0,
 				"cluster-image-registry-operator":        119,
-				"cluster-monitoring-operator":            88,
+				"cluster-monitoring-operator":            101,
 				"cluster-node-tuning-operator":           97,
 				"cluster-samples-operator":               27,
 				"cluster-storage-operator":               202,


eggfoobar:
/hold

Hey @sosiouxme, even though the increase is small, it's odd since HA runs seems to have stayed the same. Could not identify any major changes in the monitoring operator but this all started with the new kubelet version. Will quickly look, but feel free to unblock if you're okay with this change.

deads2k:
What monitoring operator code change led to this versus the known etcd quorum stability problems?

eggfoobar:
> What monitoring operator code change led to this versus the known etcd quorum stability problems?

At this point, I don't think it has anything to do with the monitoring operator, at least I couldn't identify any change that would cause an added watch count. I wasn't aware of a quorum issue, this PR is just a quick ready to go fix since I found it odd that only the monitoring operator was failing here.

PR #29435, opened 2025-01-14
OCPBUGS-48340: skip OperatorHubSourceError metric checking when disableAllDefaultSources is true

jianzhangbjz:
When disableAllDefaultSources is true, all default catalogSource are disabled, so there is no metric.
```console
    [
      {
        "metric": {
          "__name__": "ALERTS",
          "alertname": "OperatorHubSourceError",
          "alertstate": "firing",
          "container": "catalog-operator",
          "endpoint": "https-metrics",
          "exported_namespace": "openshift-marketplace",
          "instance": "[fd01:0:0:1::1a]:8443",
          "job": "catalog-operator-metrics",
          "name": "community-operators",
          "namespace": "openshift-operator-lifecycle-manager",
          "pod": "catalog-operator-6c446dcbbb-sxvjz",
          "prometheus": "openshift-monitoring/k8s",
          "service": "catalog-operator-metrics",
          "severity": "warning"
        },
        "value": [
          1736751753.045,
          "1"
        ]
      }
    ]
```

### DIFF
--- a/test/extended/prometheus/prometheus.go
+++ b/test/extended/prometheus/prometheus.go
@@ -747,6 +747,10 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 			if isManagedService {
 				allowedAlertNames = append(allowedAlertNames, "KubeDaemonSetMisScheduled")
 			}
+			// https://issues.redhat.com/browse/OCPBUGS-48340
+			if SkipOperatorHubMetricsCheck(oc) {
+				allowedAlertNames = append(allowedAlertNames, "OperatorHubSourceError")
+			}
 
 			tests := map[string]bool{
 				// openshift-e2e-loki alerts should never fail this test, we've seen this happen on daemon set rollout stuck when CI loki was down.
@@ -1013,4 +1017,12 @@ func hasTelemeterClient(client clientset.Interface) bool {
 		e2e.Failf("could not list pods: %v", err)
 	}
 	return true
+}
+
+func SkipOperatorHubMetricsCheck(oc *exutil.CLI) bool {
+	stdout, stderr, err := oc.AsAdmin().Run("get").Args("operatorhub", "cluster", "-o=jsonpath={.spec.disableAllDefaultSources}").Outputs()
+	if err != nil {
+		fmt.Printf("command failed: %v\nstderr: %s\nstdout:%s", err, stderr, stdout)
+	}
+	return stdout == "true"
 }


jianzhangbjz:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.18-e2e-metal-ipi-upgrade-ovn-ipv6

simonpasquier:
this isn't the correct approach, you should rather add the alert name to `AllowedAlertNames` in pkg/monitortestlibrary/allowedalerts/types.go (with a link to the bug) to unblock the situation. Then work to fix the alert expression so it doesn't fire in the first place when OperatorHub is disabled.

jianzhangbjz:
Hi @simonpasquier , thanks for the help! Could you help review it again? Thanks!

pkg/monitortestlibrary/allowedalerts/types.go
```diff
@@ -47,4 +47,5 @@ var AllowedAlertNames = []string{
 	"CDIDefaultStorageClassDegraded", // Installing openshift virt with RWX storage fire an alarm, that is not relevant for most of the tests.
 	"VirtHandlerRESTErrorsHigh",      // https://issues.redhat.com/browse/CNV-50418
 	"VirtControllerRESTErrorsHigh",   // https://issues.redhat.com/browse/CNV-50418
+	"OperatorHubSourceError",         // https://issues.redhat.com/browse/OCPBUGS-48340, should skip this metric checking when disableAllDefaultSources is true
```
simonpasquier, line 50:
```suggestion
	"OperatorHubSourceError",         // https://issues.redhat.com/browse/OCPBUGS-48340
```

test/extended/prometheus/prometheus.go
```diff
@@ -747,6 +747,10 @@ var _ = g.Describe("[sig-instrumentation] Prometheus [apigroup:image.openshift.i
 			if isManagedService {
 				allowedAlertNames = append(allowedAlertNames, "KubeDaemonSetMisScheduled")
 			}
+			// https://issues.redhat.com/browse/OCPBUGS-48340
+			if SkipOperatorHubMetricsCheck(oc) {
+				allowedAlertNames = removeElement(allowedAlertNames, "OperatorHubSourceError")
+			}
```
simonpasquier, line 753: no need for this

test/extended/prometheus/prometheus.go
```diff
@@ -1014,3 +1018,21 @@ func hasTelemeterClient(client clientset.Interface) bool {
...
+
+func removeElement(slice []string, element string) []string {
+	var result []string
+	for _, v := range slice {
+		if v != element {
+			result = append(result, v)
+		}
+	}
+	return result
+}
```
simonpasquier, line 1038: same here

jianzhangbjz:
Hi @simonpasquier , could you help review it again? Thanks! The logic is that add `OperatorHubSourceError` to `allowedAlertNames` when SkipOperatorHubMetricsCheck is true.

joelanford:
@jianzhangbjz Can you explain the rationale here more? This alert is only firing 7% of the time right now. Where are the default sources being disabled in CI clusters?

jianzhangbjz:
Hi @joelanford , I believe they disable it in the [baremetalds-devscripts-setup](https://github.com/openshift/release/blob/14145d08a2d8b367133ab9883bdde96f8443a2f0/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh) step, see the https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.18-e2e-metal-ipi-upgrade-ovn-ipv6/1877545344619778048/artifacts/e2e-metal-ipi-upgrade-ovn-ipv6/baremetalds-devscripts-setup/artifacts/root/dev-scripts/logs/06_create_cluster-2025-01-09-205310.log 
```console
2025-01-09 21:45:15 +(utils.sh:754): add_local_certificate_as_trusted(): oc patch image.config.openshift.io/cluster --patch '{"spec":{"additionalTrustedCA":{"name":"registry-config"}}}' --type=merge
2025-01-09 21:45:15 image.config.openshift.io/cluster patched
2025-01-09 21:45:15 +(./06_create_cluster.sh:81): [[ -n true ]]
2025-01-09 21:45:15 +(./06_create_cluster.sh:81): [[ true != \f\a\l\s\e ]]
2025-01-09 21:45:15 +(./06_create_cluster.sh:82): oc patch OperatorHub cluster --type json -p '[{"op": "add", "path": "/spec/disableAllDefaultSources", "value": true}]'
2025-01-09 21:45:16 operatorhub.config.openshift.io/cluster patched
2025-01-09 21:45:16 +(./06_create_cluster.sh:86): [[ -n '' ]]
```

jianzhangbjz:
Besides, the operatorhub is disabled in many scenarios, such as disconnected and proxy envs, for example, this https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/openstack/pre/disconnected/ipi-openstack-pre-disconnected-chain.yaml and https://github.com/openshift/release/blob/14145d08a2d8b367133ab9883bdde96f8443a2f0/ci-operator/step-registry/telco5g/cnf/tests/telco5g-cnf-tests-commands.sh#L451

jianzhangbjz:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.18-e2e-metal-ipi-upgrade-ovn-ipv6

joelanford:
So as soon as a catalog is disabled, the catalog operator will remove the `catalogsource_ready` time series from its metrics endpoint.

The rule for this alert is defined here: https://github.com/operator-framework/operator-marketplace/blob/5776ce8e796910c2dfc98a42062f97ed98e81b2c/manifests/12_prometheus_rule.yaml#L23

I'm not a prometheus expert, but I think the expression there will just _not_ have results for the time ranges that the default catalogs are disabled. If there are no results, I would think that would mean the expression would not evaluate to true. But I haven't tested this theory.

One possibility to consider. If the catalog source is unhealthy before the catalog source is disabled, the the most recent sample for a given `catalogsource_ready` time series will be `0`. In these clusters where catalog sources are not expected to work (and are thus disabled) it seems likely that they will be unhealthy prior to being disabled, right?

jianzhangbjz:
>  If there are no results, I would think that would mean the expression would not evaluate to true.

Yes, I think so.

> In these clusters where catalog sources are not expected to work (and are thus disabled) it seems likely that they will be unhealthy prior to being disabled, right?

Make sense, I will try to check the log.

jan--f:
/lgtm
Imho this here will work to unblock the readiness check. The alerting rule should and/or metric export can be then be fixed.

jianzhangbjz:
Hi @jan--f , could you help add the `acknowledge-critical-fixes-only` label? Thanks!

simonpasquier:
> The logic is that add `OperatorHubSourceError` to `allowedAlertNames` when SkipOperatorHubMetricsCheck is true.

Late to review but i think that it's simpler to exclude `OperatorHubSourceError` in all tests: the command checking whether OperatorHub is installed could fail or return unexpected results. The root cause for the alert firing despite OperatorHub not installed should be quickly addressed and the exclusion only a temporary counter-measure.

jianzhangbjz:
Hi @soltysh , could you help look at this [e2e-aws-ovn-serial](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29435/pull-ci-openshift-origin-master-e2e-aws-ovn-serial/1882108800668798976) job failure? Thanks! It fails for weeks, see [job history](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-origin-master-e2e-aws-ovn-serial?buildId=1882103965064105984)
```console

[sig-imageregistry][Serial] Image signature workflow can push a signed image to openshift registry and verify it [apigroup:user.openshift.io][apigroup:image.openshift.io] [Skipped:Disconnected] [Suite:openshift/conformance/serial] expand_less | 5m11s
-- | --
{  fail [github.com/openshift/origin/test/extended/images/signatures.go:40]: Unexpected error:     <*errors.errorString \| 0xc002302060>:      timed out while waiting of an image stream tag e2e-test-registry-signing-dgrb8/signer:latest     {         s: "timed out while waiting of an image stream tag e2e-test-registry-signing-dgrb8/signer:latest",     } occurred Ginkgo exit error 1: exit with code 1}

```

PR #29440, opened 2025-01-14
create file to upload missing leader times

deads2k:
builds on https://github.com/openshift/origin/pull/29433 to provide an upload file that needs review from @neisw

### DIFF
--- a/pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
+++ b/pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
@@ -153,39 +153,139 @@ func (w *auditLogAnalyzer) CollectData(ctx context.Context, storageDir string, b
 	return retIntervals, nil, err
 }
 
-func (*auditLogAnalyzer) ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals, recordedResources monitorapi.ResourcesMap, beginning, end time.Time) (monitorapi.Intervals, error) {
-	return nil, nil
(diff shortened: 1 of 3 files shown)


pkg/monitortests/kubeapiserver/auditloganalyzer/monitortest.go
```diff
@@ -380,6 +507,8 @@ func (w *auditLogAnalyzer) WriteContentToStorage(ctx context.Context, storageDir
 		if err != nil {
 			return err
 		}
+
```
neisw, line 510: Jobs like [e2e-gcp-ovn-upgrade](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29440/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1879241172476497920) have `request-counts-by-second` artifacts but not the `audit-no-leader-seconds` artifact.

Wondering if there is an error returned from the [png save](https://github.com/openshift/origin/blob/master/pkg/monitortests/kubeapiserver/auditloganalyzer/handle_count_tracking.go#L249), looks like we don't [log anything](https://github.com/openshift/origin/blob/79ea7f544426c261c29df4f231ba151bb8d4a573/pkg/monitortestframework/panic.go#L77) for the errors.

neisw: I see more entries in the table now including the most recent [e2e-gcp-ovn-upgrade](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/29440/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1879302858550022144/artifacts/e2e-gcp-ovn-upgrade/openshift-e2e-test/artifacts/junit/audit-no-leader-seconds_20250115-002518-autodl.json)

PR #29462, opened 2025-01-22
no-jira: add a monitor test to detect concurrent installer pod or static pod

tkashem:
The monitor test works as follows:
- a) parse the kubelet log for SyncLoop lines (both PLEG and probe lines)
- b) generate events/interval for each of these lines from `a`
- c) add a monitor test that inspects the events from `b` and construct/compute new intervals:
     - etcd installer pod duration (derived from PLEG container start, and exit )
     - static pod (etcd for now) unready interval (derived from SyncLoop probe events)

This is an example, blue is installer pod duration, the red is etcd static pod unready window:
image

from: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29462/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1882095583997464576 

The monitor test also flakes if:
- it finds two concurrent installer pods running on separate nodes
- it finds two concurrent unready window for the etcd static pod on separate nodes, which could potentially lead to etcd quorum loss

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -142,6 +142,10 @@
         return (eventInterval.source === "APIUnreachableFromClient")
     }
 
+    function isStaticPodInstallMonitorActivity(eventInterval) {
+        return (eventInterval.source === "StaticPodInstallMonitor")
+    }
+
     function isEndpointConnectivity(eventInterval) {
         if (eventInterval.message.reason !== "DisruptionBegan" && eventInterval.message.reason !== "DisruptionSamplerOutageBegan") {
             return false
@@ -270,6 +274,10 @@
 
     function isAPIUnreachableFromClientValue(item) {
         return [buildLocatorDisplayString(item.locator), "", "APIUnreachableFromClientMetrics"]
+    }
+
+    function isStaticPodInstallMonitorValue(item) {
+        return [buildLocatorDisplayString(item.locator), "", item.message.reason]
     }
 
     function disruptionValue(item) {
@@ -486,6 +494,9 @@
 
         timelineGroups.push({group: "api-unreachable", data: []})
         createTimelineData(isAPIUnreachableFromClientValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isAPIUnreachableFromClientActivity, regex)
+
+        timelineGroups.push({group: "staticpod-install", data: []})
+        createTimelineData(isStaticPodInstallMonitorValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isStaticPodInstallMonitorActivity, regex)
 
         timelineGroups.push({ group: "etcd-leaders", data: [] })
         createTimelineData(etcdLeadershipLogsValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isEtcdLeadershipAndNotEmpty, regex)
--- a/pkg/monitortests/kubeapiserver/staticpodinstall/kubeletlogparser/parser.go
+++ b/pkg/monitortests/kubeapiserver/staticpodinstall/kubeletlogparser/parser.go
@@ -1,0 +1,51 @@
+package kubeletlogparser
+
+import (
+	"strings"
+
+	"github.com/openshift/origin/pkg/monitor/monitorapi"
+)
+
+// NewEtcdStaticPodEventsFromKubelet returns a parser that will parse kubelet
+// log for SyncLoop PLEG and probe events.
+// For now, we are interested in the etcd installer and the etcd static pod
+func NewEtcdStaticPodEventsFromKubelet() func(node, line string) monitorapi.Intervals {
+	filter := func(node, ns, podName string) bool {
+		return ns == "openshift-etcd" && (strings.HasPrefix(podName, "installer-") || podName == "etcd-"+node)
+	}
+	p := parsers{
+		// we want to observe the unready window for ectd static pods
+		&SyncLoopProbeParser{
+			source: monitorapi.SourceKubeletLog,
+			want:   filter,
+		},
+		// we want to observe the container start and exit pleg events for the etcd installer pods
+		&SyncLoopPLEGParser{
+			source: monitorapi.SourceKubeletLog,
+			filter: filter,
+		},
+	}
+	return p.parse
+}
+
+// interval is created only if the given filter func returns true
+type filterFunc func(node, ns, podName string) bool
+
+type parser interface {
+	Parse(node, line string) (monitorapi.Intervals, bool)
+}
+
+type parsers []parser
+
+// this complies with the signature the default node log parser in origin requires
+func (p parsers) parse(node, line string) monitorapi.Intervals {
+	accummulated := monitorapi.Intervals{}
+	for _, parser := range p {
+		intervals, handled := parser.Parse(node, line)
+		accummulated = append(accummulated, intervals...)
+		if handled {
+			return accummulated
+		}
+	}
+	return accummulated
+}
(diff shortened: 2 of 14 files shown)


pkg/monitortests/kubeapiserver/staticpodinstall/monitortest.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (mt *monitorTest) ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals, recordedResources monitorapi.ResourcesMap, beginning, end time.Time) (monitorapi.Intervals, error) {
+	mt.computed = mt.construct(startingIntervals)
+
+	return mt.computed, nil
+}
+
+func (mt *monitorTest) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
+	junitTest := &junitTest{
+		name:     "[sig-apimachinery] installer Pods should not run concurrently on two or more node",
```
dgoodwin, line 52: little cleaner to lower case pods, and pluralize node.

pkg/monitortests/kubeapiserver/staticpodinstall/monitortest.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// flake
+	failed := &junitapi.JUnitTestCase{
+		Name:          jut.name,
+		SystemOut:     fmt.Sprintf("installer pods running concurrently on two or more nodes"),
+		FailureOutput: &junitapi.FailureOutput{},
+	}
+	for _, interval := range concurrent {
+		failed.FailureOutput.Output = fmt.Sprintf("%s\n%s", failed.FailureOutput.Output, interval.String())
+	}
```
dgoodwin, line 140: Link to help find flakes, though it will need adjusting if you change the test name: https://sippy.dptools.openshift.org/sippy-ng/tests/Presubmits/analysis?test=%5Bsig-apimachinery%5D%20installer%20Pods%20should%20not%20run%20concurrently%20on%20two%20or%20more%20node&filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22%5Bsig-apimachinery%5D%20installer%20Pods%20should%20not%20run%20concurrently%20on%20two%20or%20more%20node%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22never-stable%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22aggregated%22%7D%5D%2C%22linkOperator%22%3A%22and%22%7D

Very high flake rate it looks like.

e2echart/e2e-chart-template.html
```diff
@@ -487,6 +495,9 @@ <h5 class="modal-title">Resource</h5>
         timelineGroups.push({group: "api-unreachable", data: []})
         createTimelineData(isAPIUnreachableFromClientValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isAPIUnreachableFromClientActivity, regex)
 
+        timelineGroups.push({group: "staticpod-install", data: []})
+        createTimelineData(isStaticPodInstallMonitorValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isStaticPodInstallMonitorActivity, regex)
+
```
dgoodwin, line 500: Are the solid blue bars that look to be overlapping with a darker shade of blue what you would expect here: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29462/pull-ci-openshift-origin-master-e2e-gcp-ovn-rt-upgrade/1881869312688394240

Almost looks like locators are overlapping, but also is it supposed to show overlap for all that time?

tkashem: no, this is from a run with an earlier version of this PR, I have fixed the issue, any run from today should not have the large blue bar.

pkg/monitortests/kubeapiserver/staticpodinstall/monitortest.go
tkashem: almost all of these flakes are from earlier runs with the buggy version of this PR, the newer runs from today do not have these flakes.

dgoodwin: Ah that makes sense, I followed the link to the last failure.

pkg/monitortests/kubeapiserver/staticpodinstall/kubeletlogparser/parser.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type parsers []parser
+
+// this complies with the signature the default node log parser in origin requires
+func (p parsers) parse(node, line string) monitorapi.Intervals {
+	accummulated := monitorapi.Intervals{}
+	for _, parser := range p {
+		intervals, handled := parser.Parse(node, line)
+		accummulated = append(accummulated, intervals...)
```
ingvagabund, line 46: What if (in the future) either of Parse implementations accidentally returns non-nil intervals and `handled=false`? Will accummulated be still a valid list of intervals?

dgoodwin:
/approve

Feel free to get someone on your team to lgtm, the core things we look for are all good in here.

pkg/monitortests/kubeapiserver/staticpodinstall/kubeletlogparser/parser.go
tkashem: we can tweak it in the future, forcefully run all parsers, we also want to streamline all the existing  parsers - https://github.com/openshift/origin/blob/5b828df7f1c6648b6384170f0e5dd0614aee3e59/pkg/monitortests/node/kubeletlogcollector/node.go#L96-L106

I had to move the two parsers in this PR n their own packages so I could write some tests. to your question, right now, if that happens the existing tests will fail, so we have some protection :)

ingvagabund:
This was a good learning opportunity. This is a great and very helpful start. Let's have this sit around for a while to observe a new signal in the CI jobs. Thank you Abu.

/lgtm

ingvagabund:
CI improvements

pkg/monitortests/kubeapiserver/staticpodinstall/monitortest.go
```diff
@@ -0,0 +1,155 @@
...
+
+	framework.Logf("monitor[%s]: found %d intervals of interest", MonitorName, len(junitTest.computed))
+
+	// the following constraints define pass/fail for this test:
+	// a) if we don't find any constructed/computed interval, then
+	// this test is a noop, so we mark the test as skipped
+	// b) we find constructed/computed intervals, but no occurrences of
+	// concurrent situation, this test is a pass
+	// c) otherwise, there is at least one incident of a
+	// concurrent situation, this test is a flake/fail
```
benluddy, line 64: How hard would it be to make the test fail when a logging change causes either parser to observe nothing? If we don't see a single PLEG or probe log for any container (not limited to installer pods), would that be a reliable signal that the logs we're looking for have changed somehow?

ingvagabund:
/hold
In case others wanna review too.

pkg/monitortests/kubeapiserver/staticpodinstall/monitortest.go
tkashem: we want to streamline this for other parsers as well, we will do a follow up PR for this.

benluddy: (I know this is common to all the "log grepping" tests. I'd be happy to see an issue that describes the problem instead of delaying this test, which is useful immediately.)

tkashem: https://issues.redhat.com/browse/OCPBUGS-49301

PR #29466, opened 2025-01-22
NO-JIRA: use quay.io/openshift/origin-cli:latest directly

ricardomaraschini:
we have the following e2e test failing constantly:

```
Image signature workflow can push a signed image to openshift registry and verify it
```

this e2e is failing when attempting to build an image. the reported build failure is as follow:

```
error: build error: building at STEP "RUN source /etc/os-release && rhel_major=${VERSION_ID%!(BADPREC)%!}(MISSING) && yum config-manager --add-repo "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi${rhel_major}/${rhel_major}/\$basearch/baseos/os/"     --add-repo "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi${rhel_major}/${rhel_major}/\$basearch/appstream/os/"":
while running runtime: exit status 6
```

this error seems to be related to the yum wrapper that has been added to the image.

this pr makes the e2e to use the image directly from the upstream repository (quay), that image does not contain the yum wrapper.

### DIFF
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -51638,11 +51638,6 @@ items:
       dockerfile: |
         FROM quay.io/openshift/origin-cli:latest
         WORKDIR /var/lib/origin
-        RUN source /etc/os-release \
-            && rhel_major=${VERSION_ID%.*} \
-            && yum config-manager \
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
None of the failed tests is the one we addressed with this PR. Triggering the faulty ones once more.

/retest

PR #29479, opened 2025-01-25
SDN-4930: UDN tests wait for SCC annotation on namespace

trozet:
Since we create our own namespaces for UDN, we need to wait for the cluster manager to annotate the namespace before we can try to create pods.


trozet:
/retitle SDN-4930: UDN tests wait for SCC annotation on namespace

trozet:
@knobunc @jluhrsen this is good to go if you can lgtm please

PR #29482, opened 2025-01-27
SDN-4930: Fix live migration test detecting dualstack

trozet:
Live migration test was launching a pod to determine if cluster is dualstack. This would happen before UDN is created, so it would fail on a namespace that was enabled for UDN (network did not exist yet).

The check is really not needed, since a few lines above we already determine if the cluster is dualstack or not via a service CIDR check. Leverage that.

PR #29483, opened 2025-01-27
SDN-4930: Get cluster IP family after setting up the network

kyrtapz:
`getIPFamilyForCluster` creates a pod to determine the cluster's IP family. 
When namespaces are labeled for the primary UDN, this must happen after the network is created or the pod will fail to start.

cc: @trozet


kyrtapz:
Fixed here:
https://github.com/openshift/origin/pull/29482

PR #29489, opened 2025-01-28
Udn strict validate crd

maiqueb:


maiqueb:
This is wrong, 4.19 is OK. We're missing a backport to 4.18 - either #29488 or #29491

maiqueb:
This is wrong because 4.19 is OK. 

We're missing a backport to 4.18 - either #29488 or #29491.

PR #29490, opened 2025-01-28
OCPBUGS-49600: Display etcd bootstrap event on timeline

vrutkovs:
Update `operator_log_scraper.go` to record start/finish of etcd bootstrap member removal. This also updates other events tracked there to have a duration of 1 sec otherwise these are not being displayed on spyglass interval chart

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -101,6 +101,11 @@
 
     }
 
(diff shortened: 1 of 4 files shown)


dgoodwin:
Could you please also add the display flag with Display() if they're intended for chart display, the new UI will automatically show those.

PR #29491, opened 2025-01-28
Backwards compatible udn ipam api

maiqueb:
Supersedes #29488


maiqueb:
Abandoning it; another route was chosen.

PR #29499, opened 2025-01-29
OAPE-91: Add E2E tests for Route ExternalCertificate feature

chiragkyal:
Implements end-to-end tests for routes configured with ExternalCertificate feature. These tests cover the following scenarios:

- Validation and pre-requisite of the feature
  - Router serviceaccount should have permission to read the secret.
  - The route and the referenced secret must exist in the same namespace.
  - The secret should be of type `kubernetes.io/tls`.
  - Route with `Passthrough` termination is not supported.
  - Both external certificates and inline certificates are not allowed.

- With a valid setup the router should support external certificate.
- Multiple routes can refer a common secret.
- Secret delete, re-create and update scenarios. 
   - With RBAC permissions validation.
- Route update scenarios (with validation checks)
  - To use a new external certificate.
  - To use the same external certificate.
  - To remove and again re-add the same external certificate.

Part of: https://issues.redhat.com//browse/OAPE-91

### DIFF
--- a/test/extended/router/external_certificate.go
+++ b/test/extended/router/external_certificate.go
@@ -1,0 +1,775 @@
+package router
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	routev1 "github.com/openshift/api/route/v1"
+	"github.com/openshift/origin/test/extended/router/certgen"
+	exutil "github.com/openshift/origin/test/extended/util"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/wait"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/framework/pod"
+)
+
+const (
+	// secretReaderRole is the name of the Role allowing access to the secret.
+	secretReaderRole = "secret-reader-role"
+	// secretReaderRoleBinding is the name of the RoleBinding associating the Role with the router service account.
+	secretReaderRoleBinding = "secret-reader-role-binding"
+	// helloOpenShiftResponse is the HTTP response from hello-openshift example pod.
+	helloOpenShiftResponse = "Hello OpenShift"
+	// defaultCertificateCN is the CommonName of router default certificate.
+	defaultCertificateCN = "ingress-operator"
+)
+
+var _ = g.Describe("[sig-network][OCPFeatureGate:RouteExternalCertificate][Feature:Router][apigroup:route.openshift.io]", func() {
+	defer g.GinkgoRecover()
+	var (
+		oc            = exutil.NewCLIWithPodSecurityLevel("router-external-certificate", admissionapi.LevelBaseline)
+		helloPodPath  = exutil.FixturePath("..", "..", "examples", "hello-openshift", "hello-pod.json")
+		helloPodName  = "hello-openshift"
+		helloPodSvc   = "hello-openshift"
+		defaultDomain string
+		err           error
+	)
+
+	g.BeforeEach(func() {
+		defaultDomain, err = getDefaultIngressClusterDomainName(oc, time.Minute)
+		o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+
+		g.By("creating pod")
+		err = oc.Run("create").Args("-f", helloPodPath, "-n", oc.Namespace()).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("waiting for the pod to be running")
+		err = pod.WaitForPodNameRunningInNamespace(context.TODO(), oc.KubeClient(), helloPodName, oc.Namespace())
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("creating service")
+		err = oc.Run("expose").Args("pod", helloPodName, "-n", oc.Namespace()).Execute()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("waiting for the service to become available")
+		err = exutil.WaitForEndpoint(oc.KubeClient(), oc.Namespace(), helloPodSvc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+	})
+
+	g.Context("with invalid setup", func() {
+		var host string
+		g.BeforeEach(func() {
+			host = oc.Namespace() + "." + defaultDomain
+		})
+
+		g.Describe("the router", func() {
+			g.It("should not support external certificate without proper permissions", func() {
+				g.By("Creating a TLS certificate secret")
+				secret, _, err := generateTLSCertSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, host)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route")
+				route := generateRouteWithExternalCertificate(oc.Namespace(), "route", secret.Name, helloPodSvc, host, routev1.TLSTerminationEdge)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.And(
+					o.ContainSubstring("Forbidden: router serviceaccount does not have permission to get this secret"),
+					o.ContainSubstring("Forbidden: router serviceaccount does not have permission to watch this secret"),
+					o.ContainSubstring("Forbidden: router serviceaccount does not have permission to list this secret")),
+				)
+			})
+
+			g.It("should not support external certificate if the secret is in a different namespace", func() {
+				g.By("Creating a new namespace")
+				differentNamespace := fmt.Sprintf("%s-%s", "router-external-certificate", rand.String(5))
+				err := createNamespace(oc, differentNamespace)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a TLS certificate secret in another namespace")
+				secret, _, err := generateTLSCertSecret(differentNamespace, "my-tls-secret", corev1.SecretTypeTLS, host)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.AdminKubeClient().CoreV1().Secrets(differentNamespace).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route in other namespace")
+				route := generateRouteWithExternalCertificate(oc.Namespace(), "route", secret.Name, helloPodSvc, host, routev1.TLSTerminationEdge)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(`Not found: "secrets \"my-tls-secret\" not found"`))
+			})
+
+			g.It("should not support external certificate if the secret is not of type kubernetes.io/tls", func() {
+				g.By("Creating a secret with the WRONG type (Opaque)")
+				secret, _, err := generateTLSCertSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeOpaque, host) // Incorrect type
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route")
+				route := generateRouteWithExternalCertificate(oc.Namespace(), "route", secret.Name, helloPodSvc, host, routev1.TLSTerminationEdge)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(`Invalid value: "my-tls-secret": secret of type "kubernetes.io/tls" required`))
+			})
+
+			g.It("should not support external certificate if the route termination type is Passthrough", func() {
+				g.By("Creating a TLS certificate secret")
+				secret, _, err := generateTLSCertSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, host)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route with Passthrough termination")
+				passthroughRoute := generateRouteWithExternalCertificate(oc.Namespace(), "passthrough-route", secret.Name, helloPodSvc, host, routev1.TLSTerminationPassthrough)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), passthroughRoute, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(`Invalid value: "my-tls-secret": passthrough termination does not support certificates`))
+			})
+
+			g.It("should not support external certificate if inline certificate is also present", func() {
+				g.By("Creating a TLS certificate secret")
+				secret, _, err := generateTLSCertSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, host)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route")
+				route := generateRouteWithExternalCertificate(oc.Namespace(), "route", secret.Name, helloPodSvc, host, routev1.TLSTerminationEdge)
+				// Add inline certificate
+				route.Spec.TLS.Certificate = "my-crt"
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(`Invalid value: "my-tls-secret": cannot specify both tls.certificate and tls.externalCertificate`))
+			})
+		})
+	})
+
+	g.Context("with valid setup", func() {
+		var (
+			secret       *corev1.Secret
+			routes       []*routev1.Route
+			hosts        []string
+			rootDerBytes []byte
+		)
+
+		g.BeforeEach(func() {
+			// The number of routes here is deliberately set to be greater than 5
+			// to test the OpenShift Router's contention tracker behaviour. (see: https://github.com/openshift/router/blob/b41f9d05467fb7b3f6c2dafa6ac4b5e25164c0b6/pkg/router/controller/contention.go#L86).
+			// This tracker limits the frequency of route status updates.
+			// https://github.com/openshift/router/pull/614 introduced to ignore contention (route status updates) done by this feature (ExternalCertificate).
+			// To ensure proper handling of the contention tracker, we need to test scenarios where a single secret is shared by more than 5 routes.
+			// These routes' statuses should be able to update frequently without being blocked by the contention tracker.
+			const numRoutes = 6
+			var routeNames []string
+
+			for i := 0; i < numRoutes; i++ {
+				hosts = append(hosts, fmt.Sprintf("host-%d-%s.%s", i, oc.Namespace(), defaultDomain))
+				routeNames = append(routeNames, fmt.Sprintf("route-%d", i))
+			}
+
+			g.By("Creating a TLS certificate secret")
+			secret, rootDerBytes, err = generateTLSCertSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, hosts...)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("Providing router service account permissions to get,list,watch the secret")
+			_, err = oc.KubeClient().RbacV1().Roles(oc.Namespace()).Create(context.Background(),
+				generateSecretReaderRole(oc.Namespace(), "my-tls-secret"), metav1.CreateOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			_, err = oc.KubeClient().RbacV1().RoleBindings(oc.Namespace()).Create(context.Background(),
+				generateRouterRoleBinding(oc.Namespace()), metav1.CreateOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			g.By("Creating multiple routes referencing same external certificate")
+			for i := 0; i < numRoutes; i++ {
+				route := generateRouteWithExternalCertificate(oc.Namespace(), routeNames[i], secret.Name, helloPodSvc, hosts[i], routev1.TLSTerminationEdge)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				routes = append(routes, route)
+			}
+		})
+
+		g.Describe("the router should support external certificate", func() {
+			g.It("and routes are reachable", func() {
+				g.By("Sending https request")
+				for _, route := range routes {
+					hostName, err := getHostnameForRoute(oc, route.Name)
+					o.Expect(err).NotTo(o.HaveOccurred())
+					resp, err := httpsGetCall(hostName, rootDerBytes)
+					o.Expect(err).NotTo(o.HaveOccurred())
+					o.Expect(resp).Should(o.ContainSubstring(helloOpenShiftResponse))
+				}
+			})
+
+			g.Context("and the secret is deleted", func() {
+				g.BeforeEach(func() {
+					g.By("Deleting the secret")
+					err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Delete(context.Background(), secret.Name, metav1.DeleteOptions{})
+					o.Expect(err).NotTo(o.HaveOccurred())
+				})
+
+				g.It("then routes are not reachable", func() {
+					g.By("Checking the route status")
+					for _, route := range routes {
+						checkRouteStatus(oc, route.Name, corev1.ConditionFalse, "ExternalCertificateValidationFailed")
+					}
+				})
+
+				g.Context("and re-created again", func() {
+					g.BeforeEach(func() {
+						g.By("Re-creating the deleted secret")
+						_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+					})
+
+					g.It("then routes are reachable", func() {
+						g.By("Sending https request")
+						for _, route := range routes {
+							hostName, err := getHostnameForRoute(oc, route.Name)
+							o.Expect(err).NotTo(o.HaveOccurred())
+							resp, err := httpsGetCall(hostName, rootDerBytes)
+							o.Expect(err).NotTo(o.HaveOccurred())
+							o.Expect(resp).Should(o.ContainSubstring(helloOpenShiftResponse))
+						}
+					})
+				})
+
+				g.Context("and re-created again but RBAC permissions are dropped", func() {
+					g.BeforeEach(func() {
+						g.By("Deleting RBAC permissions")
+						err = oc.KubeClient().RbacV1().RoleBindings(oc.Namespace()).Delete(context.Background(), secretReaderRoleBinding, metav1.DeleteOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+
+						g.By("Re-creating the deleted secret")
+						_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+					})
+
+					g.It("then routes are not reachable", func() {
+						g.By("Checking the route status")
+						for _, route := range routes {
+							checkRouteStatus(oc, route.Name, corev1.ConditionFalse, "ExternalCertificateValidationFailed")
+						}
+					})
+				})
+			})
+
+			g.Context("and the secret is updated", func() {
+				g.BeforeEach(func() {
+					g.By("Updating the existing secret")
+					// build a new secret
+					secret, rootDerBytes, err = generateTLSCertSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, hosts...)
+					o.Expect(err).NotTo(o.HaveOccurred())
+					// update the existing secret with the new secret
+					_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Update(context.Background(), secret, metav1.UpdateOptions{})
+					o.Expect(err).NotTo(o.HaveOccurred())
+				})
+
+				g.It("then also routes are reachable", func() {
+					g.By("Sending https request")
+					for _, route := range routes {
+						hostName, err := getHostnameForRoute(oc, route.Name)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						resp, err := httpsGetCall(hostName, rootDerBytes)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						o.Expect(resp).Should(o.ContainSubstring(helloOpenShiftResponse))
+					}
+				})
+			})
+
+			g.Context("and the secret is updated but RBAC permissions are dropped", func() {
+				g.BeforeEach(func() {
+					g.By("Deleting RBAC permissions")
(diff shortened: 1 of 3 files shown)


test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				)
+			})
+
+			g.It("should not support external certificate if the secret is in a different namespace", func() {
+				g.By("Creating a new namespace")
+				differentNamespace := fmt.Sprintf("%s-%s", "router-external-certificate", rand.String(5))
+				err := createNamespace(oc, differentNamespace)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				secret, _, err := createSecret(differentNamespace, "my-tls-secret", corev1.SecretTypeTLS, host)
```
swghosh, line 101:
```suggestion
				g.By("Creating a TLS certificate secret in another namespace")
				secret, _, err := createSecret(differentNamespace, "my-tls-secret", corev1.SecretTypeTLS, host)
```

swghosh: Add `By(...)` before

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+
+				g.By("Creating a route")
+				route := createRouteWithEC(oc.Namespace(), "route", secret.Name, helloPodSvc, host, routev1.TLSTerminationEdge)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(`Invalid value: "my-tls-secret": secret of type "kubernetes.io/tls" required`))
+			})
+
+			g.It("should not support external certificate if the route termination type is Passthrough", func() {
+				g.By("Creating a secret with crt/key")
```
swghosh, line 128: ditto

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+	g.Context("with invalid setup", func() {
+		var host string
+		g.BeforeEach(func() {
+			host = oc.Namespace() + "." + defaultDomain
+		})
+
+		g.Describe("the router", func() {
+			g.It("should not support external certificate without proper permissions", func() {
+				g.By("Creating a secret with crt/key")
+				secret, _, err := createSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, host)
```
swghosh, line 79: where the cleanup for `${oc.Namespace()}/my-tls-secret` Secret?
I see the same object is being created a few times.

swghosh: briefly recall some tooling/wrappers exist which auto-creates a new namespace when a portion of the test starts, so each time there is a new namespace for `oc.Namespace()`.
is my understanding correct?

And, that'd suggest that the `createNamespace` is only for creating the extra `differentNamespace`.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route with Passthrough termination")
+				passthroughRoute := createRouteWithEC(oc.Namespace(), "passthrough-route", secret.Name, helloPodSvc, host, routev1.TLSTerminationPassthrough)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), passthroughRoute, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(`Invalid value: "my-tls-secret": passthrough termination does not support`))
```
swghosh, line 138: **QQ(implementation)**: what is the expected outcome if user creates a Route with ExternalCertificate and tls termination type Edge, and later changes the termination type to Passthrough?

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				g.By("Creating a secret with crt/key")
+				secret, _, err := createSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, host)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route")
+				route := createRouteWithEC(oc.Namespace(), "route", secret.Name, helloPodSvc, host, routev1.TLSTerminationEdge)
+				// Add inline certificate
+				route.Spec.TLS.Certificate = "my-crt"
```
swghosh, line 151: can a real certificate string be added here?

because, this might error out differently with say a validation error that "invalid certificate" and bypass the other stuff (inline + external cert being present together) we're willing to check.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				})
+
+				g.Context("and re-created again", func() {
+					g.BeforeEach(func() {
+						g.By("Re-creating the deleted secret")
+						_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+					})
+
+					g.It("then routes are reachable", func() {
```
swghosh, line 232: Is there any wait period or slight delay involved?

as the backing secret was deleted, to what it seems here this may cause the test to fail if we do not wait (as there are no retries) / ends up taking for the routes to be propagated back and ready.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+						for i := 0; i < len(routes); i++ {
+							hostName, err := getHostnameForRoute(oc, routes[i].Name)
+							o.Expect(err).NotTo(o.HaveOccurred())
+							resp, err := httpsGetCall(hostName, rootDerBytes)
+							o.Expect(err).NotTo(o.HaveOccurred())
+							o.Expect(resp).Should(o.ContainSubstring("Hello OpenShift"))
+						}
+					})
+				})
+				g.Context("and re-created again but RBAC permissions are dropped", func() {
```
swghosh, line 243: dint get it, why "dropping the RBAC" is under  "**valid setup**"?

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+			})
+
+			g.Context("and the route is updated", func() {
+				var (
+					routeToTest   *routev1.Route
+					newSecretName = "new-ext-crt"
+				)
+
+				g.BeforeEach(func() {
+					// These tests donot *explicity* need verification on multiple routes.
```
swghosh, line 312: Okay, why the above ones need to be tested on multiple routes,
is only because we wanted to re-use the same secret being used by multiple Routes, right?

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+						o.Expect(err).NotTo(o.HaveOccurred())
+						_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+
+						g.By("Updating the exisiting role to add RBAC permissions for the new secret")
+						newRule := fmt.Sprintf(`{"apiGroups": [""],"resources": ["secrets"],"resourceNames":["%s"],"verbs": ["get", "list", "watch"]}`, newSecretName)
+						rolePatch := fmt.Sprintf(`{"rules": [%s]}`, newRule)
+						_, err = oc.KubeClient().RbacV1().Roles(oc.Namespace()).Patch(
+							context.Background(), SECRET_READER_ROLE, types.MergePatchType, []byte(rolePatch), metav1.PatchOptions{},
+						)
```
swghosh, line 330: prefer de-cluttering this by adding a separate func where the patches are made.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+						o.Expect(err).NotTo(o.HaveOccurred())
+
+						g.By("Updating the route to use new external certificate")
+						routePatch := fmt.Sprintf(`{"spec":{"tls":{"externalCertificate":{"name":"%s"}}}}`, newSecretName)
```
swghosh, line 334: similar, we can use a convenience func for this.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+					g.BeforeEach(func() {
+						g.By("Creating a new secret")
+						secret, _, err = createSecret(oc.Namespace(), newSecretName, corev1.SecretTypeTLS, hosts...)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+					})
+					g.It("route update is rejcted", func() {
+						g.By("Updating the route to use new external certificate")
+						routePatch := fmt.Sprintf(`{"spec":{"tls":{"externalCertificate":{"name":"%s"}}}}`, newSecretName)
```
swghosh, line 361: similar.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+						)
+						o.Expect(err).To(o.HaveOccurred())
+						o.Expect(err.Error()).To(o.And(
+							o.ContainSubstring("Forbidden: router serviceaccount does not have permission to get this secret"),
+							o.ContainSubstring("Forbidden: router serviceaccount does not have permission to watch this secret"),
+							o.ContainSubstring("Forbidden: router serviceaccount does not have permission to list this secret")),
+						)
+					})
+				})
+				g.Context("to use new external certificate, but secret is not of type kubernetes.io/tls", func() {
```
swghosh, line 373: nit: new line before.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+						hostName, err := getHostnameForRoute(oc, routeToTest.Name)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						resp, err := httpsGetCall(hostName, rootDerBytes)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						o.Expect(resp).Should(o.ContainSubstring("Hello OpenShift"))
+					})
+
+				})
+				g.Context("to use same external certificate, but RBAC permissions are dropped", func() {
+					g.BeforeEach(func() {
```
swghosh, line 425: **QQ(implementation)**:
does retrospectively (i.e. after Route and Secret was added, in course of an active route endpoint) dropping the RBAC cause router to be aware that the secret watch was lost?

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+						_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Patch(
+							context.Background(), routeToTest.Name, types.MergePatchType, []byte(routePatch), metav1.PatchOptions{},
+						)
+						o.Expect(err).To(o.HaveOccurred())
+						o.Expect(err.Error()).To(o.ContainSubstring(fmt.Sprintf(`Not found: "secrets \"%s\" not found"`, newSecretName)))
+					})
+				})
+				g.Context("to use same external certificate", func() {
+					g.BeforeEach(func() {
+						g.By("Adding some label to trigger route update")
```
swghosh, line 406: QQ(implementation): Certificate Renewal related

haven't been following up to work my way up all the discussions you've had in this regard,
I know from past discussions that an update event is necessary to trigger any update - but does this the Certificate secret contents being updated when renewal changes the contents (I think the singleitemlistwatch should take care)?

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+							o.ContainSubstring("Forbidden: router serviceaccount does not have permission to get this secret"),
+							o.ContainSubstring("Forbidden: router serviceaccount does not have permission to watch this secret"),
+							o.ContainSubstring("Forbidden: router serviceaccount does not have permission to list this secret")),
+						)
+					})
+				})
+				g.Context("to remove and again re-add the same external certificate", func() {
+					g.BeforeEach(func() {
+						g.By("Updating the route to remove the external certificate reference")
+						routePatch := `{"spec": {"tls": null}}`
```
swghosh, line 448: not sure, how easy it is to test out:
unsetting this field `spec.tls`, after this we should still do a HTTPS call with the same route
use InsecureVerifyTLS and verify that the certificate is NOT the same (i.e. a different CA / something on those lines).

swghosh (commented):
din't quite get the structuring about the "with valid setup" and "with invalid setup" segregated blocks as towards of near end of those test cases they seemed to complement each other

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+
+				g.Context("to use new external certificate", func() {
+					g.BeforeEach(func() {
+						g.By("Creating a new secret")
+						secret, rootDerBytes, err = createSecret(oc.Namespace(), newSecretName, corev1.SecretTypeTLS, hosts...)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+
+						g.By("Updating the exisiting role to add RBAC permissions for the new secret")
```
arkadeepsen, line 325: nit
```suggestion
						g.By("Updating the existing role to add RBAC permissions for the new secret")
```

test/extended/router/externalCertificate.go
arkadeepsen: I also have the same question. If we want to test on more than one route makes sense to use 2. If we are using 5, then it would make sense to use different configurations for them. I don't think we are expecting different behaviours for the different routes.

arkadeepsen, line 312:
```suggestion
					// These tests do not *explicitly* need verification on multiple routes.
```

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+
+				g.It("then routes are not reachable", func() {
+					g.By("Checking the route status")
+					for i := 0; i < len(routes); i++ {
+						checkRouteStatus(oc, routes[i].Name, corev1.ConditionFalse, "ExternalCertificateValidationFailed")
+					}
+				})
+
+				g.Context("and re-created again", func() {
+					g.BeforeEach(func() {
```
arkadeepsen, line 226: In this `Context` only one `It` section is added. This `BeforeEach` section will only be invoked once for the single `It` section. Should this be included inside the `It` section itself?

Similar, trend is there throughout the remaining tests. Not adding multiple comments for them.

arkadeepsen (commented):
I concur with the comments added by @swghosh. I have added a few more comments.

zz_generated.manifests/test-reporting.yaml
```diff
@@ -434,6 +434,68 @@ spec:
         networks and persistent ips configured created using [OCPFeatureGate:NetworkSegmentation]
         UserDefinedNetwork [Suite:openshift/network/virtualization] should keep ip
         when the VMI attached to a secondary UDN is migrated between nodes'
+  - featureGate: RouteExternalCertificate
+    tests:
+    - testName: '[sig-network][OCPFeatureGate:RouteExternalCertificate][Feature:Router][apigroup:route.openshift.io]
```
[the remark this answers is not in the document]
chiragkyal: _All_ the tests present under `router` directory have the `[Feature:Router]` in their names. So, I think there should be some filtering logic which makes use of these terms while building the suite. 

And I believe if any test has `OCPFeatureGate` in the test name, then it will be skipped in non-TP job.
https://github.com/openshift/origin/blob/main/pkg/clioptions/suiteselection/feature_filter.go#L58-L82 has something on those lines.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+			if len(ingress.Conditions) == 0 {
+				e2e.Logf("ingress condition is empty, retrying...")
+				return false, nil
+			}
+			for _, condition := range ingress.Conditions {
+				if condition.Reason != ingressConditionReason && condition.Status != ingressConditionStatus {
+					e2e.Logf("unexpected ingres condition, expected: [%s,%v] but got: [%s,%v], retrying...", ingressConditionReason, ingressConditionStatus, condition.Reason, condition.Status)
+					return false, nil
+				} else {
+					e2e.Logf("got the expected ingres condition, reason: %s, status: %v", condition.Reason, condition.Status)
```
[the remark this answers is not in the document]
chiragkyal: >if you want this to be de-cluttered,
> else { log.. } can be be removed

Doesn't look too cluttered to me honestly. 
> as PollUntilTimeout deems to retry until successful. 

True. But I think we can have some log lines to print the success events as well. However I'm fine removing this if it really makes a difference.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				}
+			}
+		}
+		return true, nil
+	})
+}
+
+// createSecret generates a TLS secret containing a certificate and key.
+// The certificate is valid for the provided hostnames.
+func createSecret(namespace, secretName string, secretType corev1.SecretType, hosts ...string) (*corev1.Secret, []byte, error) {
```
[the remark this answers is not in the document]
chiragkyal: Not always, because there are negative cases where we're passing non `kubernetes.io/tls` type e.g `Opaque` and expecting the Validation to fail.

test/extended/router/externalCertificate.go
chiragkyal: > because, this might error out differently with say a validation error that "invalid certificate" and bypass the other stuff (inline + external cert being present together) we're willing to check.

I think the certificate validation will be checked by the router, after the openshift-apiserver validation succeeds. So, technically (as per the validation logic) this should work for any string.

chiragkyal: > and bypass the other stuff (inline + external cert being present together) we're willing to check.

We have checks to catch this 
```go
o.Expect(err.Error()).To(o.ContainSubstring(`Invalid value: "my-tls-secret": cannot specify both tls.certificate and tls.externalCertificate`))
```

test/extended/router/externalCertificate.go
chiragkyal: > Is there any wait period or slight delay involved?

Yes, there is `PollUntilContextTimeout` inside `httpsGetCall` function, which will do the retires.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				ExternalCertificate: &routev1.LocalObjectReference{
+					Name: secretName,
+				},
+			},
+		},
+	}
+}
+
+// createRole creates a role that grants permissions to get, list, and watch the specified secret.
+func createRole(namespace, secretName string) *rbacv1.Role {
```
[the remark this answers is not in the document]
chiragkyal: Renamed to `generateSecretReaderRole`

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				routes = append(routes, route)
+			}
+		})
+
+		g.Describe("the router should support external certificate", func() {
+			g.It("and routes are reachable", func() {
+				g.By("Sending https request")
+				for i := 0; i < len(routes); i++ {
```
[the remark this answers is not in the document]
chiragkyal: Not as of now. I wanted to do something different earlier, which is irrelevant now. 

Updated to `for _, route := range routes`

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+		Spec: routev1.RouteSpec{
+			Host: host,
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: serviceName,
+			},
+			TLS: &routev1.TLSConfig{
+				Termination: termination,
+				ExternalCertificate: &routev1.LocalObjectReference{
+					Name: secretName,
```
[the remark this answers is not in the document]
chiragkyal: I think the term "secret in a different namespace" is slightly misleading here. The validation will check the existence of the secret in the same namespace as the route. 

xRef: https://github.com/openshift/library-go/blob/bad2d1bf3a3765a83bdbdf4f32d14399053524a2/pkg/route/validation/validation.go#L350-L357

Hence the namespace can be derived from the route's metadata and doesn't require an additional field here.

test/extended/router/externalCertificate.go
```diff
@@ -0,0 +1,682 @@
...
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), route, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.And(
+					o.ContainSubstring("Forbidden: router serviceaccount does not have permission to get this secret"),
+					o.ContainSubstring("Forbidden: router serviceaccount does not have permission to watch this secret"),
+					o.ContainSubstring("Forbidden: router serviceaccount does not have permission to list this secret")),
+				)
+			})
+
+			g.It("should not support external certificate if the secret is in a different namespace", func() {
```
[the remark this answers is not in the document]
chiragkyal: https://github.com/openshift/origin/pull/29499#discussion_r1990762891 should help clarify the UX perspective. Users are expected to create the secret in the same namespace as the route.

test/extended/router/externalCertificate.go
chiragkyal: Right, there is some wrapper which does the cleanup for all the resources created and it will delete the `oc.Namespace()` as well. Each test executes in a self-contained manner is what I've understood.

test/extended/router/externalCertificate.go
chiragkyal: [ValidateRouteUpdate](https://github.com/openshift/library-go/blob/bad2d1bf3a3765a83bdbdf4f32d14399053524a2/pkg/route/validation/validation.go#L218-L224) shares the same validation function as the `ValidateRoute`. So, changing the termination type to Passthrough will also get rejected.

test/extended/router/externalCertificate.go
chiragkyal: No, that is not there in the goals of the implementation, since there is no one watching whether the RBAC _always_ exists or not. After the router approves a route, the validation needs some triggers (route creation, route update, secret update) to re-evaluate everything; until that time, the route will be served.

test/extended/router/externalCertificate.go
chiragkyal: Right, Certificate Renewal will update the contents of the secret which will send an update event to the secret monitor (singleitemlistwatch). 

The following test is added to test the scenario

```
"[sig-network][OCPFeatureGate:RouteExternalCertificate][Feature:Router][apigroup:route.openshift.io] with valid setup the router should support external certificate and the secret is updated then also routes are reachable": " [Suite:openshift/conformance/parallel]",
```

chiragkyal: The above test (against which you have commented) is added to test scenarios where route will be updated but `ExternalCertificate` field will remain unchanged -- in that case also, we want to do the validation of the feature, so that dropping RBAC permissions will get checked. 

The following test is added for this:
```
"[sig-network][OCPFeatureGate:RouteExternalCertificate][Feature:Router][apigroup:route.openshift.io] with valid setup the router should support external certificate and the route is updated to use same external certificate, but RBAC permissions are dropped route update is rejected": " [Suite:openshift/conformance/parallel]",

```

test/extended/router/externalCertificate.go
chiragkyal: I think that would help make the tests more structured. 

I wanted to make the `It` blocks more focused (and less cluttered) by adding in necessary checks only (making httpCalls, checking route status etc.) and moving the configuration bits to the `BeforeEach` blocks. 

> In this Context only one It section is added. 

As of now this is true, but when the tests will have to grow in future, I think it would help others IMHO.

test/extended/router/externalCertificate.go
chiragkyal: Moved to a new function `patchRoleWithSecretAccess`

test/extended/router/externalCertificate.go
chiragkyal: Moved to new a helper func `patchRouteWithExternalCertificate`

test/extended/router/externalCertificate.go
chiragkyal: Moved to `patchRouteWithExternalCertificate`

test/extended/router/externalCertificate.go
chiragkyal: The **"valid setup"**  establishes the preconditions required for a route with externalCertificate to function correctly. The subsequent tests, including those involving dropping RBAC, build upon this valid setup to test how the route behaves under different modification scenarios (secret updates, deletions, re-creation or route changes). Without the initial valid configuration, these specific tests wouldn't be meaningful.

Dropping RBAC in this context is specifically designed to re-test the validation after the initial valid setup. It ensures the route behaves as expected (becomes unavailable) when permissions are dropped in between.

chiragkyal:
Cluster install failed 

/test e2e-gcp-ovn-techpreview

swghosh (commented):
/lgtm

test/extended/router/external_certificate.go
```diff
@@ -0,0 +1,769 @@
...
+			}
+		})
+
+		g.Describe("the router should support external certificate", func() {
+			g.It("and routes are reachable", func() {
+				g.By("Sending https request")
+				for _, route := range routes {
+					hostName, err := getHostnameForRoute(oc, route.Name)
+					o.Expect(err).NotTo(o.HaveOccurred())
+					resp, err := httpsGetCall(hostName, rootDerBytes)
```
alebedev87, line 211: We can generate the `caPool` once (from `rootDerBytes`) and pass it to `httpsGetCall()` instead of doing it for each route.

Upd: I think I see why it was done this way, the "update secret" test would change the certificate, so we will have to regenerate caPool later.

alebedev87: Btw would 2 routes be sufficient to test the shared secret case?

test/extended/router/external_certificate.go
```diff
@@ -0,0 +1,769 @@
...
+						checkRouteStatus(oc, route.Name, corev1.ConditionFalse, "ExternalCertificateValidationFailed")
+					}
+				})
+
+				g.Context("and re-created again", func() {
+					g.BeforeEach(func() {
+						g.By("Re-creating the deleted secret")
+						_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+						o.Expect(err).NotTo(o.HaveOccurred())
+					})
```
alebedev87, line 236: Does it make sense to make this `BeforeEach`? It will be run only before `then routes are reachable`, won't it?

test/extended/router/external_certificate.go
```diff
@@ -0,0 +1,769 @@
...
+							o.Expect(err).NotTo(o.HaveOccurred())
+							resp, err := httpsGetCall(hostName, rootDerBytes)
+							o.Expect(err).NotTo(o.HaveOccurred())
+							o.Expect(resp).Should(o.ContainSubstring(helloOpenShiftResponse))
+						}
+					})
+				})
+
+				g.Context("and re-created again but RBAC permissions are dropped", func() {
+					g.BeforeEach(func() {
```
alebedev87, line 251: Same question about this `BeforeEach`. And actually about all `BeforeEach`es which are used for a single `It` clause.

test/extended/router/external_certificate.go
```diff
@@ -0,0 +1,769 @@
...
+						g.By("Sending https request")
+						hostName, err := getHostnameForRoute(oc, routeToTest.Name)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						resp, err := httpsGetCall(hostName, rootDerBytes)
+						o.Expect(err).NotTo(o.HaveOccurred())
+						o.Expect(resp).Should(o.ContainSubstring(helloOpenShiftResponse))
+					})
+				})
+
+				g.Context("to use new external certificate, but RBAC permissions are not added", func() {
```
alebedev87, line 355: What is the point of doing the same tests? That this time a single route is being modified? If so, does this really test a new code path?

alebedev87 (commented):
Thanks a lot for the PR, Chirag! The feature is very well covered.
Just some questions and nit remarks.

test/extended/router/external_certificate.go
```diff
@@ -0,0 +1,769 @@
...
+		helloPodPath  = exutil.FixturePath("..", "..", "examples", "hello-openshift", "hello-pod.json")
+		helloPodName  = "hello-openshift"
+		helloPodSvc   = "hello-openshift"
+		defaultDomain string
+		err           error
+	)
+
+	g.BeforeEach(func() {
+		defaultDomain, err = getDefaultIngressClusterDomainName(oc, time.Minute)
+		o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
```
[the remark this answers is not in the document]
chiragkyal: `getDefaultIngressClusterDomainName` needs `oc` client as an input, so we cannot put it before generating `oc`.  Now, writing it inside the `BeforeEach` or outside will not change its behaviour -- in both cases, it will run for all tests even if we're expecting the same output.

test/extended/router/external_certificate.go
```diff
@@ -0,0 +1,769 @@
...
+				secret, _, err := generateTLSCertSecret(oc.Namespace(), "my-tls-secret", corev1.SecretTypeTLS, host)
+				o.Expect(err).NotTo(o.HaveOccurred())
+				_, err = oc.KubeClient().CoreV1().Secrets(oc.Namespace()).Create(context.Background(), secret, metav1.CreateOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				g.By("Creating a route with Passthrough termination")
+				passthroughRoute := generateRouteWithExternalCertificate(oc.Namespace(), "passthrough-route", secret.Name, helloPodSvc, host, routev1.TLSTerminationPassthrough)
+				_, err = oc.RouteClient().RouteV1().Routes(oc.Namespace()).Create(context.Background(), passthroughRoute, metav1.CreateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(`Invalid value: "my-tls-secret": passthrough termination does not support`))
```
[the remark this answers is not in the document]
chiragkyal: Thanks for the catch. The error message should be `passthrough termination does not support certificates`. 

xRef: https://github.com/openshift/library-go/blob/78a24021cf4073a6ed2bd9c085ba9b6ccfd5122c/pkg/route/validation/validation.go#L271-L275

test/extended/router/external_certificate.go
chiragkyal: > Btw would 2 routes be sufficient to test the shared secret case?

We actually have 5 routes to test the shared secret case. `const numRoutes = 5`. 

xRef: https://github.com/openshift/origin/pull/29499/files#diff-ab54e60f94a61345960f23ba755e95aadcb3a00287fde5d154c08c65009abf0dR174

Let me increase it to `6` since [contention tracker](https://github.com/openshift/router/blob/b41f9d05467fb7b3f6c2dafa6ac4b5e25164c0b6/pkg/router/controller/contention.go#L86) is also at `5`

test/extended/router/external_certificate.go
chiragkyal: True, but I don't want to _clutter_  `It` blocks, and hence moved all the configuration bits into `BeforeEach` blocks. It would help keep the tests more structured and future additions will be comparatively easier. 

We discussed a bit on those lines here: https://github.com/openshift/origin/pull/29499#discussion_r1990915749

test/extended/router/external_certificate.go
chiragkyal: Same as: https://github.com/openshift/origin/pull/29499#discussion_r2016891372

test/extended/router/external_certificate.go
alebedev87: I can understand the point of view. However, I don't see how combining these two would clutter the `It` block. The combination of "re-creation of a secret" "check of the route which uses it" seems to be a complete, self contained test case (== `It` block). You did follow this logic in the `Context` below, for instance `should not support external certificate without proper permissions`  1) generates a secret, 2) creates it and 3) verifies the route creation fails, all in one `It` block.
Also, since Arkadeep found this approach questionable as well, chances are that the future readers will have the same question.

test/extended/router/external_certificate.go
alebedev87: As discussed, let's add a comment explaining the number of the routes.

test/extended/router/external_certificate.go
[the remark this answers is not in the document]
alebedev87: We should follow up on this one. Can you file a bug please?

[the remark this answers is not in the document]
chiragkyal: Sorry, but I could not understand why we need a bug here. 

I've updated this line to `passthrough termination does not support certificates` to match the expected error message from [library-go](https://github.com/openshift/library-go/blob/78a24021cf4073a6ed2bd9c085ba9b6ccfd5122c/pkg/route/validation/validation.go#L271-L275).

Do you think the library-go's error message needs some modification?

test/extended/router/external_certificate.go
chiragkyal: Sure, added a detailed comment explaining the reasoning behind it.

test/extended/router/external_certificate.go
chiragkyal: I've merged the leaf `BeforeEach` and `It` blocks together. Please let me know if it's fine now. Thanks!

test/extended/router/external_certificate.go
chiragkyal: I've merged the all leaf `BeforeEach` and `It` blocks together. Please let me know if it's fine now. Thanks!

chiragkyal:
/payload-job periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview-serial periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview

chiragkyal:
/payload-job periodic-ci-openshift-hypershift-release-4.19-periodics-e2e-aws-ovn

chiragkyal:
```
could not run steps: step hello-openshift-openshift.origin failed: inspect of image "image-registry.openshift-image-registry.svc:5000/ci-op-p6frzzsn/pipeline:hello-openshift-openshift.origin-amd64" failed with error: failed to authorize: failed to fetch oauth token: Post "https://image-registry.openshift-image-registry.svc:5000/openshift/token": EOF 
INFO[2025-04-01T06:23:04Z] Reporting job state 'failed' with reason 'executing_graph:step_failed:building_project_image' 
```

/payload-job periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview-serial

test/extended/router/external_certificate.go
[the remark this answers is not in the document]
alebedev87: Got it, the original substring was missing `certificates` at the end. Sorry, for the confusion.

> Do you think the library-go's error message needs some modification?

No, I think it's fine.

neisw:
/override e2e-metal-ipi-ovn-ipv6

last [e2e-metal-ipi-ovn-ipv6](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29499/pull-ci-openshift-origin-main-e2e-metal-ipi-ovn-ipv6/1907805018447155200) actually passed but timed out at the very end.  History shows passes for the current sha.

neisw:
/skip

last [e2e-metal-ipi-ovn-ipv6](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29499/pull-ci-openshift-origin-main-e2e-metal-ipi-ovn-ipv6/1907805018447155200) actually passed but timed out at the very end.

PR #29501, opened 2025-01-30
SDN-4930: increase probe initialDelaySeconds from 1 to 5

jluhrsen:
a flake was seen where a test pod was restarted because the startup probe failed 3 times in 3 seconds (connection refused) with the first probe fired 1s after the pod was marked as Started. Digging through the logs, the pod did seem to come up in time and nothing egregious in the node or networking pod logs.

the current theory and hope is that the test pod containers app hadn't fully been ready and listening for the probes and increasing this startup delay will be a little nicer in this case.


trozet (approved):
/lgtm

knobunc:
None of the failures are germane because this just bumps two really short timeouts

PR #29507, opened 2025-01-30
virt, udn: skip UDN tests until those are GAed

maiqueb:
This commit will be reverted once the `NetworkSegmentation` FG is lifted.


maiqueb:
Abandoned - no longer needed.

maiqueb:
Abandoned - no longer needed.

PR #29520, opened 2025-02-03
NO-JIRA: network, virt: Retry virtctl console cmd on err

qinqon:
There is a flaky on ovn-kubernetes kubevirt live migration conformance test where there is the error "expecter: No process running", this means that the command "virtctl console -n [namespace]" has die before sending login command so the golang expecter fails.

This change do a retry on that error and use a debug option to print the check status.

### DIFF
--- a/test/extended/networking/kubevirt/console.go
+++ b/test/extended/networking/kubevirt/console.go
@@ -24,11 +24,14 @@ package kubevirt
 import (
 	"bufio"
 	"fmt"
+	"log"
 	"regexp"
 	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
+
+	"k8s.io/client-go/util/retry"
 
 	"google.golang.org/grpc/codes"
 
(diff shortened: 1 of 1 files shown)


test/extended/networking/kubevirt/console.go
```diff
@@ -189,69 +192,83 @@ func LoginToFedora(virtctlPath, vmiNamespace, vmiName, user, password string) er
...
-	expecter, _, err := newExpecter(virtctlPath, vmiNamespace, vmiName, consoleConnectionTimeout, expect.Verbose(true), expect.VerboseWriter(GinkgoWriter))
-	if err != nil {
-		return err
-	}
-	defer expecter.Close()
+	err := retry.OnError(retry.DefaultBackoff,
+		func(err error) bool {
+			return strings.Contains(err.Error(), "expect: Process not running")
+		},
+		func() error {
```
maiqueb, line 199: nit:
```suggestion
	err := retry.OnError(
	        retry.DefaultBackoff,
		func(err error) bool {
			return strings.Contains(err.Error(), "expect: Process not running")
		},
		func() error {
```

test/extended/networking/kubevirt/console.go
```diff
@@ -189,69 +192,83 @@ func LoginToFedora(virtctlPath, vmiNamespace, vmiName, user, password string) er
...
+			expecter, _, err := newExpecter(virtctlPath, vmiNamespace, vmiName, consoleConnectionTimeout,
+				expect.Verbose(true),
+				expect.VerboseWriter(GinkgoWriter),
+				expect.DebugCheck(log.New(GinkgoWriter, "expect: virtctl console check", 0 /* no flags */)),
```
maiqueb, line 203: what does this do @qinqon ? Just log that we have to re-try establishing the connection ?

maiqueb (approved):
Nice ! 

I think re-connecting the console when the process is not available, then using it will address the flakes we see now and then about this feature. 

And the code looks good. 

Huge thanks.

test/extended/networking/kubevirt/console.go
qinqon: It runs the check that of "virtctl command" running and prints it and return back the result, so if it fails we will se the stdout of the virtcl command itself.

PR #29524, opened 2025-02-03
MULTIARCH-5318: Add tests for ImageStreamImportMode feature gate

Prashanth684:
Add tests to exercise these scenarios under a featuregate:
- verify that the imagestream tag's import mode corresponds to the desired.Architecture field in the status reported by CV. multi => PreserveOriginal, else Legacy.
- verify that setting the ImageStreamImportMode field in the image.config.io spec to Legacy, will be the import mode for any newly created imagestream tag henceforth
- verify that setting the ImageStreamImportMode field in the image.config.io spec to PreserveOriginal, will be the import mode for any newly created imagestream tag henceforth


test/extended/images/tagimportmode.go
```diff
@@ -0,0 +1,191 @@
...
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/operator"
+)
+
+var _ = g.Describe("[sig-imageregistry][OCPFeatureGate:ImageStreamImportMode][Serial] ImageStream API", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("imagestream-api")
+
+	g.It("import mode should be PreserveOriginal or Legacy depending on desired.architecture field in the CV [apigroup:image.openshift.io]", func() {
+		TestImageStreamImportMode(g.GinkgoT(), oc)
```
flavianmissi, line 28: Given that the two test cases below cover both `Legacy` and `PreserveOriginal`, do we really need this test case? I can't really see what value it's adding, though I may be missing something.

flavianmissi:
left a comment but lgtm either way
/lgtm

test/extended/images/tagimportmode.go
Prashanth684: this test case is actually the only "necessary" one in my opinion because it changes the default import mode based on the CV's architecture. The other two test cases are specifically for testing the knob to change the import mode which I don't see being used frequently

Prashanth684: the main idea here is that we have these tests which are exercised behind the featuregate and once we let them run for a release (or maybe a few sprint cycles), we use that signal to promote the featuregate to default and then  I was thinking we remove the two tests which change the import mode knob.

flavianmissi: ack, thanks for expanding!

PR #29386, opened 2024-12-19
CORENET-5329: revert primary udn test workaround from virt tests

maiqueb:
/hold until [OCPBUGS-42609](https://issues.redhat.com/browse/OCPBUGS-42609) is available downstream ...

PR #29526, opened 2025-02-04
SDN-5570: virt, udn: revert assert network is ready before starting workloads

maiqueb:


maiqueb:
Didn't notice I had already pushed #29386 ...

PR #29529, opened 2025-02-06
UPSTREAM: <carry>: disable etcd readiness checks by default

ingvagabund:
https://github.com/openshift/kubernetes/pull/2174 for more context.


vendor/k8s.io/kubernetes/test/e2e/apimachinery/health_handlers.go
```diff
@@ -82,7 +82,6 @@ var (
 	requiredReadyzChecks = sets.NewString(
 		"[+]ping ok",
 		"[+]log ok",
-		"[+]etcd ok",
```
bertinatto, line 85: We need to implement this change in o/k, then bump o/k in this repo

PR #29533, opened 2025-02-07
CNTRLPLANE-195: add a test that flakes if etcd cluster is leaderless more than 5 seconds

Elbehery:
resolves https://issues.redhat.com/browse/CNTRLPLANE-195

cc @tkashem @vrutkovs @dgrisonnet   @benluddy

### DIFF
--- a/pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
+++ b/pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
@@ -3,11 +3,14 @@ package etcdloganalyzer
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/kubernetes/test/e2e/framework"
 
 	coreinformers "k8s.io/client-go/informers/core/v1"
 
@@ -23,6 +26,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
+
+const leaderlessTimeout = 5 * time.Second
 
 type etcdLogAnalyzer struct {
 	adminRESTConfig *rest.Config
@@ -77,7 +82,7 @@ func (w *etcdLogAnalyzer) CollectData(ctx context.Context, storageDir string, be
 	return nil, nil, nil
 }
 
-func (*etcdLogAnalyzer) ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals, recordedResources monitorapi.ResourcesMap, beginning, end time.Time) (monitorapi.Intervals, error) {
+func (w *etcdLogAnalyzer) ConstructComputedIntervals(ctx context.Context, startingIntervals monitorapi.Intervals, recordedResources monitorapi.ResourcesMap, beginning, end time.Time) (monitorapi.Intervals, error) {
 	ret := monitorapi.Intervals{}
 
 	leader := ""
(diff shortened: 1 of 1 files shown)


pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
```diff
@@ -138,11 +143,25 @@ func (*etcdLogAnalyzer) ConstructComputedIntervals(ctx context.Context, starting
...
 	}
 
-	return ret, nil
+	w.constructed = ret
+	return w.constructed, nil
 }
 
-func (*etcdLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
-	return nil, nil
+func (w *etcdLogAnalyzer) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
```
benluddy, line 150: What is the difference between the value in the new field `constructed` and the value passed here as the parameter `finalIntervals`? Is the new field necessary?

vrutkovs: Agree with Ben - these are the same intervals:

https://github.com/Elbehery/origin/blob/909a424355bc8039e1af981c5c1f67c695107018/pkg/monitor/monitor.go#L103

https://github.com/Elbehery/origin/blob/909a424355bc8039e1af981c5c1f67c695107018/pkg/monitor/monitor.go#L113

https://github.com/Elbehery/origin/blob/909a424355bc8039e1af981c5c1f67c695107018/pkg/monitor/monitor.go#L123-L126

pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
```diff
@@ -331,3 +350,62 @@ func searchForKey(msg, key string) string {
...
+
+	type leaderless struct {
+		duration   time.Duration
+		prev, next *monitorapi.Interval
+	}
+
+	exceeded := make([]leaderless, 0)
+	for i := 1; i < len(jut.computed); i++ {
+		prev, next := &jut.computed[i-1], &jut.computed[i]
+		if duration := next.From.Sub(prev.To); duration > 10*time.Second {
```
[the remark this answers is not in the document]
vrutkovs: Also in my tests `prev.To` happens a second after `next.From` (possibly due to some rounding?). We should warn if `next.From.Sub(prev.To)` is negative

vrutkovs (changes requested):

pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
vrutkovs: I've updated the test to avoid using `constructed` and instead filter `finalIntervals` PTAL - https://github.com/openshift/origin/commit/ec983ec8992a4874d621a7d53425781253a26b6e

The way to test this is to run the monitor and stop two master machines (one of them have to be etcd leader) so that it could not elect a new one for some time

Elbehery: actually will add this manually, the commit has two parents, cherrypicking will be mess :D

vrutkovs: Feel free to merge it with your commit afterwards

Elbehery: updated 👍🏽

pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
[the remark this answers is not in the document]
Elbehery: updated 👍🏽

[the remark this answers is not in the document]
Elbehery: @vrutkovs which error msg ? 

u mean the duration during the error ?

pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
```diff
@@ -141,8 +146,45 @@ func (*etcdLogAnalyzer) ConstructComputedIntervals(ctx context.Context, starting
...
+		}
+		if value, ok := interval.Message.Annotations[monitorapi.AnnotationEtcdLeader]; ok && len(value) == 0 {
+			continue
+		}
+		etcdIntervals = append(etcdIntervals, interval)
+	}
+	sort.Sort(etcdIntervals)
+
+	junitTest := &junitTest{
+		name:     "[sig-etcd] cluster should not be without a leader for more than 10s",
```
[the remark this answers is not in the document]
Elbehery: done 👍🏽

vrutkovs:
/lgtm

Tested this manually by stopping two masters while the monitor was running

pkg/monitortests/etcd/etcdloganalyzer/monitortest.go
```diff
@@ -141,8 +146,45 @@ func (*etcdLogAnalyzer) ConstructComputedIntervals(ctx context.Context, starting
...
+		}
+		if value, ok := interval.Message.Annotations[monitorapi.AnnotationEtcdLeader]; ok && len(value) == 0 {
+			continue
+		}
+		etcdIntervals = append(etcdIntervals, interval)
+	}
+	sort.Sort(etcdIntervals)
+
+	junitTest := &junitTest{
+		name:     fmt.Sprintf("[sig-etcd] cluster should not be without a leader for more than %s seconds", leaderlessTimeout),
```
sosiouxme, line 175: ITYM
```suggestion
		name:     fmt.Sprintf("[sig-etcd] cluster should not be without a leader for more than %s", leaderlessTimeout),
```

but actually in the event that you change the threshold I don't think you want it in the name at all (you would need to remap the test to be considered "the same" or mark the old name as deliberately removed).

```suggestion
		name:     "[sig-etcd] cluster should not be without a leader for too long",
```

Elbehery: done 👍🏽

vrutkovs (approved):
/lgtm

PR #29536, opened 2025-02-11
 OCPBUGS-49647: Fixing build s2i ruby test data inline with latest ruby version(>=3.0)

Moebasim:
The test data for s2i image creation with source type as binary has issues post base image upgrade to ruby 3.0 (test/extended/testdata/builds/test-env-build.json) This is because webrick server is removed from default ruby libraries from 3.0 version,hence a webserver should be installed externally.The test data is missing this causing permanant failure post ruby 3.0 upgrade.

This fix is updating the test data with required dependencies to create a stable app to fulfill the testcase criteria.

signed-off by : moebasim(mbasim@redhat.com)

### DIFF
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -18595,7 +18595,7 @@ var _testExtendedTestdataBuildsS2iEnvironmentBuildAppGemfile = []byte(`source "h
 
 gem "rack"
 gem "rackup"
-`)
+gem "webrick"`)
 
 func testExtendedTestdataBuildsS2iEnvironmentBuildAppGemfileBytes() ([]byte, error) {
 	return _testExtendedTestdataBuildsS2iEnvironmentBuildAppGemfile, nil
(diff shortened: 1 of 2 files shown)


test/extended/util/annotate/rules.go
```diff
@@ -53,6 +53,9 @@ var (
 			// https://issues.redhat.com/browse/OCPBUGS-3339
 			`\[sig-devex\]\[Feature:ImageEcosystem\]\[mysql\]\[Slow\] openshift mysql image Creating from a template should instantiate the template`,
 			`\[sig-devex\]\[Feature:ImageEcosystem\]\[mariadb\]\[Slow\] openshift mariadb image Creating from a template should instantiate the template`,
+
+			// https://issues.redhat.com/browse/OCPBUGS-37799
```
xueqzhan, line 57: Do I understand it correctly that the goal is to disable this test only for 4.15. Releases after that still has this test enabled?

PR #29541, opened 2025-02-13
OCPBUGS-50628: [test: networking] Skip isolation test when cgroupv1 nodes are present

npinaeva:
cgroupv1 is only supported till 4.18, so I didn't want to put a lot of extra logic for this case.
I have tried listing all nodes and scheduling the test pods on the cgroupv2 nodes, but those are only control-plane nodes (in the rhel8 job that I am trying to fix at least), and it needs more taints to allow scheduling.
So I figured instead I can just skip the isolation test when cgroupv1 nodes are present

See successful run https://github.com/openshift/origin/pull/29541#issuecomment-2717398993

### DIFF
--- a/test/extended/networking/network_segmentation.go
+++ b/test/extended/networking/network_segmentation.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 
 	nadapi "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	nadclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/typed/k8s.cni.cncf.io/v1"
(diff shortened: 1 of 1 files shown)


npinaeva:
/payload-job periodic-ci-openshift-release-master-nightly-4.18-e2e-aws-ovn-workers-rhel8

npinaeva:
/payload-job periodic-ci-openshift-release-master-nightly-4.18-e2e-aws-ovn-workers-rhel8

npinaeva:
/payload-job periodic-ci-openshift-release-master-nightly-4.18-e2e-aws-ovn-workers-rhel8

test/extended/networking/network_segmentation.go
```diff
@@ -231,6 +232,19 @@ var _ = Describe("[sig-network][OCPFeatureGate:NetworkSegmentation][Feature:User
 
 						udnPodConfig.namespace = f.Namespace.Name
 
+						// make sure we don't have cgroupv1 nodes in the cluster
+						events, err := cs.CoreV1().Events("default").List(context.Background(), metav1.ListOptions{})
+						Expect(err).NotTo(HaveOccurred())
+						cgroupv1Nodes := map[string]bool{}
+						for _, event := range events.Items {
+							if event.Reason == "UDNKubeletProbesNotSupported" {
```
tssurya, line 240: the logic is relying on events of the nodes? so I assume in a cluster always there will be at least 1 event emitted at startup of that node if UDNIsolation feature gate is ON? such that by the time the test I executed we would have an event? i think there is a time limit on how many events are stored right? IIRC 1hr in default namespace? will that be enough for a 3hr test? OR is this event emitted every time someone tries to create UDNs so there will be this event happening continuously?

tssurya (commented):
I'm fine with this approach just one inline question
 https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-origin-29541-nightly-4.18-e2e-aws-ovn-workers-rhel8/1899769694722723840 - do you have the snippet of this test's logs from this run?

test/extended/networking/network_segmentation.go
npinaeva: Events are only generated on ovnkube startup, so there should be enough time before the test begins. As for the event lifetime, it seems to be set to 3h https://access.redhat.com/solutions/6968044 so borderline enough for the test

npinaeva:
> I'm fine with this approach just one inline question https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-origin-29541-nightly-4.18-e2e-aws-ovn-workers-rhel8/1899769694722723840 - do you have the snippet of this test's logs from this run?

Yep! the log is
`skip [github.com/openshift/origin/test/extended/networking/network_segmentation.go:245]: cgroupv1 nodes are found: map[ip-10-0-59-79.ec2.internal:true ip-10-0-69-63.ec2.internal:true ip-10-0-98-117.ec2.internal:true], kubelet probes won't work` from https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/openshift-origin-29541-nightly-4.18-e2e-aws-ovn-workers-rhel8/1899769694722723840/artifacts/e2e-aws-ovn-workers-rhel8/openshift-e2e-test/build-log.txt

jluhrsen:
@npinaeva what do we need to get this in?

jluhrsen:
@trozet @danwinship , can you PTAL here?

PR #29552, opened 2025-02-18
OCPBUGS-45205: flake "kubelet terminates kube-apiserver gracefully extended"

atiratree:


atiratree:
flaked 1/2 in the https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29552/pull-ci-openshift-origin-master-e2e-metal-ipi-ovn-ipv6/1891876644201697280 job

test/extended/apiserver/graceful_termination.go
```diff
@@ -96,7 +96,7 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Late]", func() {
 			}
 		}
 		if len(finalMessageBuilder.String()) > 0 {
-			g.GinkgoT().Errorf("The following API Servers weren't gracefully terminated: %v", finalMessageBuilder.String())
+			result.Flakef("The following API Servers weren't gracefully terminated. Probably kubelet or CRI-O is not giving the time to cleanly shut down. This can lead to connection refused and network I/O timeout errors in other components.\n\n%s", finalMessageBuilder.String())
```
p0lyn0mial, line 99: are we going to let the node team know that they should take a look and propose a fix ? or is this an issue with the test ?

atiratree: I think the test fine, just that the kubelet cannot ensure the gracefully termination during the node restart. I will ask the node team for a confirmation. Btw, I have copy pasted the new error message from the test above.

p0lyn0mial: I don't mind tagging this test as flaky as long as the node team is aware of the issue and is willing to fix it in the future.

haircommander: my understanding here is we can't ensure an ordering without graceful node shutdown, which has other issues. So I agree this is something we're looking towards fixing long term, but it's not something we imminently are working on. how does that sound @p0lyn0mial ? cc @rphillips

PR #29558, opened 2025-02-20
OCPBUGS-50587: apiserver/patch: update patch tests to use server side apply

everettraven:
and not set the currentRevision on nodeStatus creation as we should not be setting currentRevision on the creation of a nodeStatus.


everettraven:
closing in favor of #29560

PR #29560, opened 2025-02-24
OCPBUGS-50587: Rely on unstructured for update status instead of type conversion

ardaguclu:
Currently, we are in a process of fixing a bug that is caused by that node-controller and installer-controller attempt to update the node statuses of static operators simultaneously. Consequently, that race results in a hot loop that eventually causes the operator fall into a degraded state in a period of time. 

This PR https://github.com/openshift/api/pull/2208 (which is inherited from https://github.com/openshift/api/pull/2203) adds new validations to only allow;
* node status can only be created (or removed) with only node name (no other fields should be set) -- node controller
* node status can be updated with other fields in subsequent update requests -- installer controller

However, `JSONPatch` tests in this repository does indeed allow to create new node statuses with some dummy but non-nil values (e.g. `currentRevision=0`) and this test fails in openshift/api PRs ^^. 

This PR skips the unstructured - object conversion which eventually adds `currentRevision=0` while updating statuses.

After the patch in openshift/api merges, I'll add new tests specifically covering these new rules in a followup PR (because this PR has to be backported to 4.18).

### DIFF
--- a/test/extended/apiserver/patch.go
+++ b/test/extended/apiserver/patch.go
@@ -112,12 +112,13 @@ func createWellKnownKubeAPIServerOperatorResource(ctx context.Context, resourceC
 	unstructuredKasOperator, err := resourceClient.Create(ctx, unstructuredKasOperatorManifest, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
-	kasOperator := unstructuredToKubeAPIServerOperator(unstructuredKasOperator.Object)
-	kasOperatorFromManifest := unstructuredToKubeAPIServerOperator(unstructuredKasOperatorManifest.Object)
-	kasOperator.Status = kasOperatorFromManifest.Status
-	unstructuredKasOperator, err = resourceClient.UpdateStatus(ctx, kubeAPIServerOperatorToUnstructured(kasOperator), metav1.UpdateOptions{})
+	manifestStatus, _, err := unstructured.NestedMap(unstructuredKasOperatorManifest.Object, "status")
 	o.Expect(err).NotTo(o.HaveOccurred())
-	kasOperator = unstructuredToKubeAPIServerOperator(unstructuredKasOperator.Object)
+	err = unstructured.SetNestedMap(unstructuredKasOperator.Object, manifestStatus, "status")
+	o.Expect(err).NotTo(o.HaveOccurred())
+	unstructuredKasOperator, err = resourceClient.UpdateStatus(ctx, unstructuredKasOperator, metav1.UpdateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	kasOperator := unstructuredToKubeAPIServerOperator(unstructuredKasOperator.Object)
 	o.Expect(kasOperator.Status.NodeStatuses).To(o.Equal([]operatorv1.NodeStatus{
 		{NodeName: "master-1"},
 		{NodeName: "master-2"},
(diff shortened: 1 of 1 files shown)


test/extended/apiserver/patch.go
```diff
@@ -106,16 +106,16 @@ var _ = g.Describe("[sig-api-machinery] JSON Patch [apigroup:operator.openshift.
...
 	unstructuredKasOperatorManifest := resourceread.ReadUnstructuredOrDie(kasOperatorYaml)
 	unstructuredKasOperator, err := resourceClient.Create(ctx, unstructuredKasOperatorManifest, metav1.CreateOptions{})
 	o.Expect(err).NotTo(o.HaveOccurred())
 
 	kasOperator := unstructuredToKubeAPIServerOperator(unstructuredKasOperator.Object)
 	kasOperatorFromManifest := unstructuredToKubeAPIServerOperator(unstructuredKasOperatorManifest.Object)
-	kasOperator.Status = kasOperatorFromManifest.Status
-	unstructuredKasOperator, err = resourceClient.UpdateStatus(ctx, kubeAPIServerOperatorToUnstructured(kasOperator), metav1.UpdateOptions{})
+	statusApplyConfiguration := toStaticPodOperatorStatusApplyConfiguration(kasOperatorFromManifest)
+	unstructuredKasOperator, err = resourceClient.ApplyStatus(ctx, kasOperator.Name, toUnstructuredKubeAPIServerOperatorStatusApplyConfiguration(gvk, kasOperator, statusApplyConfiguration), metav1.ApplyOptions{Force: true, FieldManager: fieldManager})
```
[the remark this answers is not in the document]
ardaguclu: @p0lyn0mial thank you for review. What do you think about the latest version of this PR (it is almost identical with https://github.com/openshift/origin/pull/29560 but still uses `updatestatus`).

test/extended/apiserver/patch.go
```diff
@@ -115,7 +114,23 @@ func createWellKnownKubeAPIServerOperatorResource(ctx context.Context, resourceC
 	kasOperator := unstructuredToKubeAPIServerOperator(unstructuredKasOperator.Object)
 	kasOperatorFromManifest := unstructuredToKubeAPIServerOperator(unstructuredKasOperatorManifest.Object)
 	kasOperator.Status = kasOperatorFromManifest.Status
-	unstructuredKasOperator, err = resourceClient.UpdateStatus(ctx, kubeAPIServerOperatorToUnstructured(kasOperator), metav1.UpdateOptions{})
+	obj := kubeAPIServerOperatorToUnstructured(kasOperator)
+	// We are not going to allow setting NodeStatus.CurrentRevision on creation of a NodeStatus.
```
p0lyn0mial, line 118: The following code performs an update.

I think that the CurrentRevision must be set to `0`. In that case we could simply update the manifest [at](https://github.com/openshift/origin/blob/12142d1cf918d8a5678639ba5555af8b503ebeab/test/extended/testdata/apiserver/operator-kube-apiserver-cr.yaml#L8) - wdyt ?

ardaguclu: Actually with this PR https://github.com/openshift/api/pull/2208 (and https://github.com/openshift/api/pull/2203), currentRevision must not exist (so currentRevision=0 is invalid as well during creation).

ardaguclu:
Results shown in https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-api-2208-openshift-origin-29560-e2e-aws-ovn/1894065137111797760 which was triggered by https://github.com/openshift/api/pull/2208#issuecomment-2679039823 that this test works also when we tighten the nodestatus validation in https://github.com/openshift/api/pull/2208

PR #29562, opened 2025-02-25
OCPBUGS-45214: Add new group snapshot images to mirror

jsafrane:
This is preparation for https://github.com/openshift/kubernetes/pull/2155 that reshuffles some images again. These images need to be available in `openshift-tests images` output before the o/k PR can merge.


jsafrane:
/close
this is not required since https://github.com/openshift/origin/pull/29450

PR #29575, opened 2025-03-04
NO-JIRA: Fix bearer token exposure in exit condition as well

Shilpa-Gokul:


tvardema:
/retitle NO-JIRA: Fix bearer token exposure in exit condition as well

PR #29597, opened 2025-03-11
NE-1968: add e2e tests for FeatureGate GatewayAPI

lihongan:
Implements e2e tests for GatewayAPI featuregate. These tests cover the following scenarios:

- Verify Gateway API CRDs and esnure required CRDs should already be installed
- Verfiy Gateway API CRDs and ensure existing CRDs can not be deleted
- Verify Gateway API CRDs and ensure existing CRDs can not be updated
- Verify Gateway API CRDs and ensure CRD of standard group can not be created
- Verify Gateway API CRDs and ensure CRD of experimental group is not installed
- Verify Gateway API CRDs and ensure CRD of experimental group can not be created

requires https://github.com/openshift/cluster-ingress-operator/pull/1192

### DIFF
--- a/test/extended/router/gatewayapi.go
+++ b/test/extended/router/gatewayapi.go
@@ -1,0 +1,143 @@
+package router
+
+import (
+	"context"
+	"strings"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+var _ = g.Describe("[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io]", func() {
+	defer g.GinkgoRecover()
+	var (
+		oc       = exutil.NewCLIWithPodSecurityLevel("gateway-api", admissionapi.LevelBaseline)
+		crdNames = []string{
+			"gatewayclasses.gateway.networking.k8s.io",
+			"gateways.gateway.networking.k8s.io",
+			"httproutes.gateway.networking.k8s.io",
+			"grpcroutes.gateway.networking.k8s.io",
+			"referencegrants.gateway.networking.k8s.io",
+		}
+		errorMessage = "ValidatingAdmissionPolicy 'openshift-ingress-operator-gatewayapi-crd-admission' with binding 'openshift-ingress-operator-gatewayapi-crd-admission' denied request: Gateway API Custom Resource Definitions are managed by the Ingress Operator and may not be modified"
+	)
+
+	g.Describe("Verify Gateway API CRDs", func() {
+		g.It("and ensure required CRDs should already be installed", func() {
+			g.By("Get and check the installed CRDs")
+			for i := range crdNames {
+				crd, err := oc.AdminApiextensionsClient().ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), crdNames[i], metav1.GetOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				e2e.Logf("Found the Gateway API CRD named: %v", crd.Name)
+			}
+		})
+
+		g.It("and ensure existing CRDs can not be deleted", func() {
+			g.By("Try to delete the CRDs and fail")
+			for i := range crdNames {
+				err := oc.AdminApiextensionsClient().ApiextensionsV1().CustomResourceDefinitions().Delete(context.Background(), crdNames[i], metav1.DeleteOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(errorMessage))
+			}
+		})
+
+		g.It("and ensure existing CRDs can not be updated", func() {
+			g.By("Get the CRDs firstly, add spec.names.shortNames then update CRD")
+			for i := range crdNames {
+				crd, err := oc.AdminApiextensionsClient().ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), crdNames[i], metav1.GetOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				// some CRDs have a shortName but some not, just trying to add one for all
+				crd.Spec.Names.ShortNames = append(crd.Spec.Names.ShortNames, "fakename")
+				_, err = oc.AdminApiextensionsClient().ApiextensionsV1().CustomResourceDefinitions().Update(context.Background(), crd, metav1.UpdateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				o.Expect(err.Error()).To(o.ContainSubstring(errorMessage))
+			}
+		})
+
+		g.It("and ensure CRD of standard group can not be created", func() {
+			fakeCRDName := "fakeroutes.gateway.networking.k8s.io"
+			g.By("Try to create CRD of standard group and fail")
+			fakeCRD := buildGWAPICRDFromName(fakeCRDName)
+			_, err := oc.AdminApiextensionsClient().ApiextensionsV1().CustomResourceDefinitions().Create(context.Background(), fakeCRD, metav1.CreateOptions{})
+			o.Expect(err).To(o.HaveOccurred())
+			o.Expect(err.Error()).To(o.ContainSubstring(errorMessage))
+		})
+
+		g.It("and ensure CRD of experimental group is not installed", func() {
+			g.By("Ensure no CRD of experimental group is installed")
+			crdList, err := oc.AdminApiextensionsClient().ApiextensionsV1().CustomResourceDefinitions().List(context.Background(), metav1.ListOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			for _, crd := range crdList.Items {
+				if crd.Spec.Group == "gateway.networking.x-k8s.io" {
+					e2e.Failf("Found unexpected CRD named: %v", crd.Name)
+				}
+			}
+		})
+
+		g.It("and ensure CRD of experimental group can not be created", func() {
+			expCRDName := "xlistenersets.gateway.networking.x-k8s.io"
+			g.By("Try to create CRD of experimental group and fail")
+			expCRD := buildGWAPICRDFromName(expCRDName)
+			_, err := oc.AdminApiextensionsClient().ApiextensionsV1().CustomResourceDefinitions().Create(context.Background(), expCRD, metav1.CreateOptions{})
+			o.Expect(err).To(o.HaveOccurred())
+			o.Expect(err.Error()).To(o.ContainSubstring(errorMessage))
+		})
+	})
+})
+
+// buildGWAPICRDFromName initializes the fake GatewayAPI CRD deducing most of its required fields from the given name.
+func buildGWAPICRDFromName(name string) *apiextensionsv1.CustomResourceDefinition {
+	var (
+		plural   = strings.Split(name, ".")[0]
+		group, _ = strings.CutPrefix(name, plural+".")
+		// removing trailing "s"
+		singular = plural[0 : len(plural)-1]
+		kind     = strings.Title(singular)
+	)
+
+	return &apiextensionsv1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: plural + "." + group,
+			Annotations: map[string]string{
+				"api-approved.kubernetes.io": "https://github.com/kubernetes-sigs/gateway-api/pull/2466",
+			},
+		},
+		Spec: apiextensionsv1.CustomResourceDefinitionSpec{
+			Group: group,
+			Names: apiextensionsv1.CustomResourceDefinitionNames{
+				Singular: singular,
+				Plural:   plural,
+				Kind:     kind,
+			},
+			Scope: apiextensionsv1.ClusterScoped,
+			Versions: []apiextensionsv1.CustomResourceDefinitionVersion{
+				{
+					Name:    "v1",
+					Storage: true,
+					Served:  true,
+					Schema: &apiextensionsv1.CustomResourceValidation{
+						OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+							Type: "object",
+						},
+					},
+				},
+				{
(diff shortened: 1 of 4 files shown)


lihongan:
https://github.com/openshift/cluster-ingress-operator/pull/1192 got merged

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,168 @@
...
+			for i := range crdNames {
+				crd, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), crdNames[i], metav1.GetOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				e2e.Logf("Found the Gateway API CRD named: %v", crd.Name)
+			}
+		})
+
+		g.It("should not be deleted", func() {
+			ctx := context.Background()
+			crdClient := apiextensionsclientset.NewForConfigOrDie(oc.AdminConfig())
```
alebedev87, line 46: Why creating the crdClient for each `It` clause? Would a single `crdClient` in `var` section above be enough?

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,168 @@
...
+			g.By("Get and check the installed CRDs")
+			for i := range crdNames {
+				crd, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), crdNames[i], metav1.GetOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				e2e.Logf("Found the Gateway API CRD named: %v", crd.Name)
+			}
+		})
+
+		g.It("should not be deleted", func() {
+			ctx := context.Background()
```
alebedev87, line 45: nit: I thin using `context.Background()` directly (like in previous `It`) is fine. Or you can create this variable in `var` section above.

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,168 @@
...
+			crdClient := apiextensionsclientset.NewForConfigOrDie(oc.AdminConfig())
+			g.By("Trying to delete the CRDs")
+			for i := range crdNames {
+				err := crdClient.ApiextensionsV1().CustomResourceDefinitions().Delete(ctx, crdNames[i], metav1.DeleteOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				e2e.Logf("Got error: %v", err.Error())
+				o.Expect(err.Error()).To(o.ContainSubstring(errorMessage))
```
alebedev87, line 52: I suppose that `Expect()` will log the error if it does not match the expected one. Therefore we can remove the log before, to be less verbose.

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		})
+
+		g.It("should not be updated", func() {
+			crdClient := apiextensionsclientset.NewForConfigOrDie(oc.AdminConfig())
```
alebedev87, line 57: Same remark about the crd client.

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,168 @@
...
+			g.By("Get the CRDs firstly, add spec.names.shortNames then update CRD")
+			for i := range crdNames {
+				crd, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), crdNames[i], metav1.GetOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				// some CRDs have a shortName but some not, just trying to add one for all
+				crd.Spec.Names.ShortNames = append(crd.Spec.Names.ShortNames, "fakename")
+				_, err = crdClient.ApiextensionsV1().CustomResourceDefinitions().Update(context.Background(), crd, metav1.UpdateOptions{})
+				o.Expect(err).To(o.HaveOccurred())
+				e2e.Logf("Got error: %v", err.Error())
```
alebedev87, line 66: Same remark about the redundant logging.

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,168 @@
...
+			_, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().Create(context.Background(), fakeCRD, metav1.CreateOptions{})
+			o.Expect(err).To(o.HaveOccurred())
+			e2e.Logf("Got error: %v", err.Error())
+			o.Expect(err.Error()).To(o.ContainSubstring(errorMessage))
+		})
+	})
+
+	g.Describe("The ValidatingAdmissionPolicy for Gateway API", func() {
+		// Adding [Serial] since it could break Gateway API CRDs creation/deletion/update tests
+		g.It("should be recreated after it is deleted [Serial]", func() {
```
alebedev87, line 86: Is adding `[Serial]` to the test name is enough to make it serial?

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,168 @@
...
+				timeout = 5 * time.Minute
+			)
+
+			g.By("The VAP should exist in cluster")
+			_, err := oc.AdminKubeClient().AdmissionregistrationV1().ValidatingAdmissionPolicies().Get(ctx, policyName, metav1.GetOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			g.By("Delete the VAP")
+			err = oc.AdminKubeClient().AdmissionregistrationV1().ValidatingAdmissionPolicies().Delete(ctx, policyName, metav1.DeleteOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			g.By("The VAP should be recreated in allowed time")
```
alebedev87, line 101: I'm not fully convinced that we have to test this part. 1. It's CVO logic, not ingress operator's, 2. It increases the runtime of the origin tests by ~5 minutes (they take a lot of time already).

zz_generated.manifests/test-reporting.yaml
```diff
@@ -65,6 +65,19 @@ spec:
     tests:
     - testName: '[sig-arch][OCPFeatureGate:Example] should only run FeatureGated test
         when enabled'
+  - featureGate: GatewayAPI
```
alebedev87, line 68: Nice! Does it mean that those tests will be run only when the given feature gate is enabled?

alebedev87 (commented):
Overall, LGTM. Just some small remarks:

zz_generated.manifests/test-reporting.yaml
lihongan: Yes, I think so. But need to check the jobs to confirm.

test/extended/router/gatewayapi.go
lihongan: Yes, good question. I also hesitate if we need to add this test. In the original test plan the case is "Deleting the VAP fails", but indeed the VAP is not protected and can be deleted.
Do you think if we have another test for CRDs other than this? maybe @Miciah has some ideas ?

alebedev87: > Do you think if we have another test for CRDs other than this? 

The thing I tested it in the pre-submits in CIO repository was the ability to manipulate CRDs when VAP is not there. For this, I scaled down CVO, removed VAP and tried to remove CRDs which should be possible (without VAP). Then I had to scale the CVO back up, btw this recreates the VAP almost immediately.

lihongan: Yes, seems removing VAP/CRDs with scaling down/up CVO should be quicker than this. Will rewrite this. Thanks

test/extended/router/gatewayapi.go
lihongan: I tried the way you mentioned but always get error when executing `make update`
```
  goroutine 1 [running]:
  runtime/debug.Stack()
  	/home/hongli/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.4.linux-amd64/src/runtime/debug/stack.go:26 +0x5e
  github.com/openshift/origin/test/extended/util.FatalErr({0x6710f60, 0xc001a845d0})
  	/home/hongli/go/src/github.com/openshift/origin/test/extended/util/client.go:1018 +0x25
  github.com/openshift/origin/test/extended/util.(*CLI).AdminConfig(0x6e0b520?)
  	/home/hongli/go/src/github.com/openshift/origin/test/extended/util/client.go:819 +0x3b
  github.com/openshift/origin/test/extended/router.init.func3()
  	/home/hongli/go/src/github.com/openshift/origin/test/extended/router/gatewayapi.go:26 +0x89
  github.com/onsi/ginkgo/v2/internal.NewNode.func2({0xc001b65a40?, 0x0?})
  	/home/hongli/go/src/github.com/openshift/origin/vendor/github.com/onsi/ginkgo/v2/internal/node.go:324 +0x13
  github.com/onsi/ginkgo/v2/internal.(*Suite).PushNode.func1(0xc00078db08?)
  	/home/hongli/go/src/github.com/openshift/origin/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:209 +0x5b
  github.com/onsi/ginkgo/v2/internal.(*Suite).PushNode(_, {0x20d, 0x2, {0x6fe7a54, 0x5c}, 0xc001545f70, {{0x8c1ad5d, 0x52}, 0x16, {0x0, ...}, ...}, ...})
  	/home/hongli/go/src/github.com/openshift/origin/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:211 +0x71e
  github.com/onsi/ginkgo/v2/internal.(*Suite).BuildTree(0xc0010b4008)
  	/home/hongli/go/src/github.com/openshift/origin/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:101 +0xb8
  k8s.io/kubernetes/openshift-hack/e2e/annotate.Run(0xa644818?, 0x71be1e8)
  	/home/hongli/go/src/github.com/openshift/origin/vendor/k8s.io/kubernetes/openshift-hack/e2e/annotate/annotate.go:31 +0xd5
  main.main()
  	/home/hongli/go/src/github.com/openshift/origin/test/extended/util/annotate/annotate.go:14 +0x25

```
But no error if creating the crdClient for each `It` clause, am I missing something ?

test/extended/router/gatewayapi.go
lihongan: Agree, I removed the log from the three cases. Thank you!

test/extended/router/gatewayapi.go
lihongan: Yes, using `context.Background()` here to be consistent.

test/extended/router/gatewayapi.go
alebedev87: Interesting. It's not a big deal (we can move on without this) but just out of curiosity, was there an error message? Because the call stack just shows that there was an error while getting `AdminConfig`.

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,209 @@
...
+			g.By("Creating new CRD in the group")
+			fakeCRD := buildGWAPICRDFromName(fakeCRDName)
+			_, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().Create(context.Background(), fakeCRD, metav1.CreateOptions{})
+			o.Expect(err).To(o.HaveOccurred())
+			o.Expect(err.Error()).To(o.ContainSubstring(errorMessage))
+		})
+	})
+})
+
+// Adding [Serial] since it could break Gateway API CRDs creation/deletion/update tests
```
alebedev87, line 83: You can create a fake CRD from the gateway API group and try to update and remove it. This would minimize any conflicts with existing CRDs.

alebedev87: As a matter of fact, you already do this in `The Gateway API CRDs should not be created`. You can use the same fake crd for all the tests, why trying to delete real CRDs even if you expect them to fail due to the VAP protection?

alebedev87: Wait, I see now why you used the real CRDs - to check that CIO recreates them. But isn't this already tested in `The Gateway API CRDs should already be installed` (CIO manage to create them when it first started)?

test/extended/router/gatewayapi.go
lihongan: There was no clear error message, maybe something wrong in init, not very sure.

test/extended/router/gatewayapi.go
lihongan: Adding [Serial] since the last test case needs scale down CVO and removes the VAP, I cannot run in parallel with other 4 cases. The case is to test the "unsupported" config path that you mentioned in the slack.

lihongan: I updated the comment and hope that explains it more clearly.

lihongan:
The 5 tests passed in the two techpreview jobs: [metal-ipi-ovn-dualstack-bgp-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29597/pull-ci-openshift-origin-main-e2e-metal-ipi-ovn-dualstack-bgp-techpreview/1904363656779403264) and [gcp-ovn-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29597/pull-ci-openshift-origin-main-e2e-gcp-ovn-techpreview/1904364316753137664)

test/extended/router/gatewayapi.go
lihongan: I removed this case and added new one of verifying experimental group CRDs. All five tests could be run in parallel now. Thanks.

test/extended/router/gatewayapi.go
```diff
@@ -0,0 +1,147 @@
...
+			expCRDName := "xlistenersets.gateway.networking.x-k8s.io"
+
+			g.By("Ensure no CRD of experimental group is installed")
+			crdList, err := crdClient.ApiextensionsV1().CustomResourceDefinitions().List(context.Background(), metav1.ListOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			for _, crd := range crdList.Items {
+				if crd.Spec.Group == "gateway.networking.x-k8s.io" {
+					e2e.Failf("Found unexpected CRD named: %v", crd.Name)
+				}
+			}
```
alebedev87, line 87: Should this be a separate `It` clause similar to `g.It("and ensure required CRDs should already be installed"`?
This way we won't need 2 `By` clauses.

lihongan: Ah yes, make sense, that way looks neat and focused.

alebedev87:
/lgtm
/approve

Can I approve?

lihongan:
The 6 tests passed in all techpreview jobs

lihongan:
This is e2e tests for feature gated code

lihongan:
@sosiouxme could you help approve ?

test/extended/router/gatewayapi.go
lihongan: @alebedev87 I added a wrapper in `test/extended/util/client.go` so this issue is solved now. The `crdClient` is not required and we can use `oc.AdminApiextensionsClient()` instead.

PR #29601, opened 2025-03-14
CNTRLPLANE-5: Decode CBOR data in etcd test

bertinatto:
CBOR serving and storage in apiserver should be TechPreview in 4.19, so we need to adjust our test to accept that type of storage in etcd.

### DIFF
--- a/test/extended/etcd/etcd_storage_path.go
+++ b/test/extended/etcd/etcd_storage_path.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	cbor "k8s.io/apimachinery/pkg/runtime/serializer/cbor/direct"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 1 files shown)


bertinatto:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview openshift/api#2235

test/extended/etcd/etcd_storage_path.go
```diff
@@ -712,6 +713,10 @@ func getFromEtcd(kv etcdv3.KV, path string) (*metaObject, error) {
 		if err := json.Unmarshal(value, metaObj); err != nil {
 			return nil, err
 		}
+	case isCBOR(value):
```
benluddy, line 716: Any CBOR in storage should have the prefix 0xd9d9f7. It's not sufficient to check if it can be understood by a CBOR decoder because there can be byte sequences that are valid in more than one protocol.

bertinatto:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview openshift/api#2235

test/extended/etcd/etcd_storage_path.go
```diff
@@ -678,7 +679,10 @@ func JSONToUnstructured(stub, namespace string, mapping *meta.RESTMapping, dynam
 	return dynamicClient.Resource(mapping.Resource).Namespace(namespace), &unstructured.Unstructured{Object: typeMetaAdder}, nil
 }
 
-var protoEncodingPrefix = []byte{0x6b, 0x38, 0x73, 0x00}
+var (
+	protoEncodingPrefix = []byte{0x6b, 0x38, 0x73, 0x00}
+	cborPrefix          = []byte{0xd9, 0xd9, 0xf7, 0xa4}
```
benluddy, line 684:
```suggestion
	cborPrefix          = []byte{0xd9, 0xd9, 0xf7}
```

The prefix is three bytes: `0xd9d9f7` (it's the encoding of the "self-described CBOR" tag, https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.6).

bertinatto: Thanks. I have no idea where I got that extra byte from. Must have been some copy-paste mistake.

benluddy: `0xa4` would be the first byte of a length-4 map. Likely you were looking at an example of a real encoded object and picked it up there.

PR #29604, opened 2025-03-17
OCPBUGS-52358: Fix egress firewall tests by updating the URL from docs.openshift.com to docs.redhat.com after docs migration

abhat:
docs.openshift.com got moved to docs.redhat.com and our egress firewall/egressnetworkpolicy tests were using that fqdn to curl and test for egress firewall functionality. With the DNS name no longer resolving, these tests need to be updated to the docs.redhat.com server instead.


tssurya:
/retitle OCPBUGS-52358: Fix egress firewall tests by updating the URL from docs.openshift.com to docs.redhat.com after docs migration

PR #29606, opened 2025-03-18
TRT-1854: add synthetic test to check for presence of "Skipped" annotations in OTE tests

smg247:
Temporarily check for the presence of the `[Skipped:xyz]` annotation in the test names, once this synthetic test begins to pass we can remove the annotation logic

### DIFF
--- a/pkg/test/ginkgo/cmd_runsuite.go
+++ b/pkg/test/ginkgo/cmd_runsuite.go
@@ -201,6 +201,27 @@ func (o *GinkgoRunSuiteOptions) Run(suite *TestSuite, junitSuiteName string, mon
 		logrus.Infof("Using built-in tests only due to OPENSHIFT_SKIP_EXTERNAL_TESTS being set")
 	}
 
+	// Temporarily check for the presence of the [Skipped:xyz] annotation in the test names, once this synthetic test
+	// begins to pass we can remove the annotation logic
+	var annotatedSkipped []string
+	for _, t := range externalTestCases {
+		if strings.Contains(t.name, "[Skipped") {
+			annotatedSkipped = append(annotatedSkipped, t.name)
+		}
+	}
+	var skippedAnnotationSyntheticTestResults []*junitapi.JUnitTestCase
+	skippedAnnotationSyntheticTestResult := junitapi.JUnitTestCase{
+		Name: "[sig-trt] Skipped annotations present",
+	}
+	if len(annotatedSkipped) > 0 {
+		skippedAnnotationSyntheticTestResult.FailureOutput = &junitapi.FailureOutput{
+			Message: fmt.Sprintf("Skipped Annotations present in tests: %s", strings.Join(annotatedSkipped, ", ")),
+		}
+	}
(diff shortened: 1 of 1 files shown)


pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -201,6 +201,27 @@ func (o *GinkgoRunSuiteOptions) Run(suite *TestSuite, junitSuiteName string, mon
...
+	// begins to pass we can remove the annotation logic
+	var annotatedSkipped []string
+	for _, t := range externalTestCases {
+		if strings.Contains(t.name, "[Skipped") {
+			annotatedSkipped = append(annotatedSkipped, t.name)
+		}
+	}
+	var skippedAnnotationSyntheticTestResults []*junitapi.JUnitTestCase
+	skippedAnnotationSyntheticTestResult := junitapi.JUnitTestCase{
+		Name: "Skipped annotations present",
```
neisw, line 214: Should we consider [sig-trt] or some annotation to associate this with team or component?

pkg/test/ginkgo/cmd_runsuite.go
```diff
@@ -201,6 +201,27 @@ func (o *GinkgoRunSuiteOptions) Run(suite *TestSuite, junitSuiteName string, mon
...
+	}
+	if len(annotatedSkipped) > 0 {
+		skippedAnnotationSyntheticTestResult.FailureOutput = &junitapi.FailureOutput{
+			Message: fmt.Sprintf("Skipped Annotations present in tests: %s", strings.Join(annotatedSkipped, ", ")),
+		}
+	}
+	skippedAnnotationSyntheticTestResults = append(skippedAnnotationSyntheticTestResults, &skippedAnnotationSyntheticTestResult)
+	// Make it flake
+	skippedAnnotationSyntheticTestResults = append(skippedAnnotationSyntheticTestResults, &junitapi.JUnitTestCase{Name: skippedAnnotationSyntheticTestResult.Name})
```
neisw, line 223: Should we only flake when we have a failure?

smg247: I guess it would only flake if there was a failure, right? It would just be 2 successful results otherwise. I could update the comment to make that more clear.

neisw: Yeah, I'm not sure if the flake logic looks for more than one run with at least one pass or more than one run with one failure.  We could ask, it just seemed odd to me to add the second pass in the case it wasn't needed.

smg247: Just to keep the code simple, but I get what you are saying. I wouldn't imagine that 2 successful runs would result in a flake, but maybe my assumption is wrong.

Either way, this is intended to be pretty temporary, so I don't think we need to debate the topic too much ;) If you want, I will wrap it in an if statement.

neisw:
/lgtm

For my own curiosity https://console.cloud.google.com/functions/details/us-east1/ci-to-bigquery?env=gen1&invt=AbtmKg&project=openshift-gce-devel&tab=source line 572 sets flake_count = 0 for success so double success should be flake = 0.

PR #29609, opened 2025-03-19
watch install times and provide and uplod file

deads2k:
This will have a survivor bias, but it gives a clear starting point and a historical record.

@dgoodwin and @neisw

### DIFF
--- a/pkg/monitortests/clusterversionoperator/installertimes/monitortest.go
+++ b/pkg/monitortests/clusterversionoperator/installertimes/monitortest.go
@@ -1,0 +1,150 @@
+package installertimes
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"github.com/openshift/origin/pkg/dataloader"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	"k8s.io/client-go/kubernetes"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+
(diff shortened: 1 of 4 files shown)


pkg/monitortests/clusterversionoperator/installertimes/monitortest.go
```diff
@@ -0,0 +1,150 @@
...
+	approximateInstallStartTime metav1.Time
+	approximateBootstrapEndTime metav1.Time
+	approximateInstallEndTime   metav1.Time
+}
+
+func NewInstallTimeAnalyzer() monitortestframework.MonitorTest {
+	return &installerTimeAnalyzer{}
+}
+
+func (w *installerTimeAnalyzer) StartCollection(ctx context.Context, adminRESTConfig *rest.Config, recorder monitorapi.RecorderWriter) error {
```
dgoodwin, line 37: Let's see what it produces but doing this in StartCollection may not work. This will be lauched with the resource watch observer, before install, in which case StartCollection I think will give you no results as none of this exists yet.

It would run again when e2e launches, but for the jobs you were looking at failing bootstrap we're never making it to e2e, so again no results. 

I think you might need to do this perhaps in CollectData, in the hopes the e2e observer catches it. 

The resource watch observer artifact will probably be bundled up in the tarball and missed by the uploader, unless you add explicit code to copy it outside. We do that for the intervals so they're more accessible. See https://github.com/openshift/release/blob/bf63bb26e14a5a2033aeb16928ce08b80ddfa47f/ci-operator/step-registry/observers/resource-watch/observers-resource-watch-commands.sh#L18

But then you likely also need to account for a situation where the regular e2e monitortest also spits out an artifact, to make sure we don't double count the result.

PR #29611, opened 2025-03-19
[WIP] TLS 1.3 / Modern profile tests

jacobsee:
This PR:

* Removes a skip on an existing TLS test - it appears to have been a setup issue
* Adds a TLS modern profile test to APIServer (see https://github.com/openshift/kubernetes/pull/2135) (tested)
* Adds (WIP) TLS 1.3 test to etcd. (untested, trouble getting a dev cluster at the moment)

### DIFF
--- a/test/extended/apiserver/tls.go
+++ b/test/extended/apiserver/tls.go
@@ -1,13 +1,89 @@
 package apiserver
 
 import (
+	"context"
 	"crypto/tls"
+	"fmt"
+	"strings"
+	"time"
 
 	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
+	operatorv1client "github.com/openshift/client-go/operator/clientset/versioned"
 	"github.com/openshift/library-go/pkg/crypto"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
+
+const tlsTestDuration = 10 * time.Minute
+const tlsWaitForCleanupDuration = 10 * time.Minute
+
+var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Serial]", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLI("apiserver")
+
+	g.It("TestTLSModernProfile", func() {
+		ctx, ctxCancelFn := context.WithTimeout(context.Background(), tlsTestDuration)
+		defer ctxCancelFn()
+
+		t := g.GinkgoT()
+
+		configClient := oc.AdminConfigClient()
+		
+		operatorClient := oc.AdminOperatorClient()
+
+		defer func() {
+			g.By("Cleanup - removing TLS profile")
+
+			cleanupCtx, cancelCleanup := context.WithTimeout(context.Background(), tlsWaitForCleanupDuration)
+			defer cancelCleanup()
+	
+			kasStatus, err := operatorClient.OperatorV1().KubeAPIServers().Get(cleanupCtx, "cluster", metav1.GetOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+	
+			currentRevision := kasStatus.Status.LatestAvailableRevision
+	
+			removeModernProfilePatch := `[{"op":"remove","path":"/spec/tlsSecurityProfile"}]`
+	
+			_, err = configClient.ConfigV1().APIServers().Patch(cleanupCtx, "cluster", types.JSONPatchType, []byte(removeModernProfilePatch), metav1.PatchOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+	
+			_, err = waitForKASIncrementedRevision(cleanupCtx, operatorClient, "cluster", currentRevision)
+			o.Expect(err).NotTo(o.HaveOccurred())
+		}()
+
+		g.By("Applying a JSON Patch to use the modern TLS profile")
+
+		kasStatus, err := operatorClient.OperatorV1().KubeAPIServers().Get(ctx, "cluster", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		currentRevision := kasStatus.Status.LatestAvailableRevision
+
+		addModernProfilePatch := `[{"op":"add","path":"/spec/tlsSecurityProfile","value":{"type":"Modern","modern":{}}}]`
+
+		_, err = configClient.ConfigV1().APIServers().Patch(ctx, "cluster", types.JSONPatchType, []byte(addModernProfilePatch), metav1.PatchOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Dialing the API with a minimum TLS version of 1.3")
+
+		_, err = waitForKASIncrementedRevision(ctx, operatorClient, "cluster", currentRevision)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		config := &tls.Config{MinVersion: tls.VersionTLS13, MaxVersion: tls.VersionTLS13, InsecureSkipVerify: true}
+
+		host := strings.TrimPrefix(oc.AdminConfig().Host, "https://")
+
+		conn, err := tls.Dial("tcp4", host, config)
+		if err != nil {
+			t.Fatalf("Expected success with TLS 1.3, got %v", err)
+		}
+
+		conn.Close()
+	})
+})
 
 var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 	defer g.GinkgoRecover()
@@ -15,8 +91,6 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 	oc := exutil.NewCLI("apiserver")
 
 	g.It("TestTLSDefaults", func() {
-		g.Skip("skipping because it was broken in master")
-
 		t := g.GinkgoT()
 		// Verify we fail with TLS versions less than the default, and work with TLS versions >= the default
 		for _, tlsVersionName := range crypto.ValidTLSVersions() {
@@ -24,8 +98,10 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 			expectSuccess := tlsVersion >= crypto.DefaultTLSVersion()
 			config := &tls.Config{MinVersion: tlsVersion, MaxVersion: tlsVersion, InsecureSkipVerify: true}
 
+			host := strings.TrimPrefix(oc.AdminConfig().Host, "https://")
+
 			{
-				conn, err := tls.Dial("tcp4", oc.AdminConfig().Host, config)
+				conn, err := tls.Dial("tcp4", host, config)
 				if err == nil {
 					conn.Close()
 				}
@@ -62,3 +138,20 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 
 	})
 })
+
+func waitForKASIncrementedRevision(ctx context.Context, operatorClient operatorv1client.Interface, name string, currentRevision int32) (int32, error) {
+	for {
+		kasStatus, _ := operatorClient.OperatorV1().KubeAPIServers().Get(context.Background(), name, metav1.GetOptions{})
+		// Intentionally don't return if this errors, as the API server is most likely still coming online, which is what we're waiting for.
+		
+		if ctx.Err() != nil {
+			return 0, fmt.Errorf("timed out waiting for KubeAPIServer to increment revision")
+		}
+		
+		if kasStatus.Status.LatestAvailableRevision > currentRevision {
+			return kasStatus.Status.LatestAvailableRevision, nil
+		}
+
+		time.Sleep(10 * time.Second)
+	}
+}
--- a/test/extended/etcd/tls.go
+++ b/test/extended/etcd/tls.go
@@ -1,0 +1,59 @@
+package etcd
+
+import (
+	"context"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+const tlsTestDuration = 10 * time.Minute
+const tlsWaitForCleanupDuration = 10 * time.Minute
+
+var _ = g.Describe("[sig-etcd][Serial] etcd", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLIWithoutNamespace("etcd-tls").AsAdmin()
+
+	g.It("Test TLS 1.3 Configuration", func() {
+		ctx, ctxCancelFn := context.WithTimeout(context.Background(), tlsTestDuration)
+		defer ctxCancelFn()
+
+		operatorClient := oc.AdminOperatorClient()
+
+		originalEtcd, err := operatorClient.OperatorV1().Etcds().Get(ctx, "cluster", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		defer func() {
+			g.By("Cleanup - removing TLS configuration")
+
+			cleanupCtx, cancelCleanup := context.WithTimeout(context.Background(), tlsWaitForCleanupDuration)
+			defer cancelCleanup()
+
+			_, err := operatorClient.OperatorV1().Etcds().Update(cleanupCtx, originalEtcd, metav1.UpdateOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			err = waitForEtcdToStabilizeOnTheSameRevision(g.GinkgoT(), oc)
+			err = errors.Wrap(err, "timed out waiting for etcd pods to stabilize after cleaning up TLS 1.3")
+			o.Expect(err).ToNot(o.HaveOccurred())
+		}()
+
+		g.By("Applying a JSON Patch to use TLS 1.3")
+
+		addTLS13Patch := `[{"op":"replace","path":"/spec/observedConfig/servingInfo","value":{"cipherSuites":["TLS_AES_128_GCM_SHA256","TLS_AES_256_GCM_SHA384","TLS_CHACHA20_POLY1305_SHA256"],"minTLSVersion":"VersionTLS13"}}]`
+
+		_, err = operatorClient.OperatorV1().Etcds().Patch(ctx, "cluster", types.JSONPatchType, []byte(addTLS13Patch), metav1.PatchOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Expecting etcd to stabilize without errors")
+
+		err = waitForEtcdToStabilizeOnTheSameRevision(g.GinkgoT(), oc)
+		err = errors.Wrap(err, "timed out waiting for etcd pods to stabilize using TLS 1.3")
+		o.Expect(err).ToNot(o.HaveOccurred())
+	})
+})


jacobsee:
Supports OCPBUGS-37706, OCPSTRAT-1364

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,114 @@
...
-		g.Skip("skipping because it was broken in master")
-
 		t := g.GinkgoT()
 		// Verify we fail with TLS versions less than the default, and work with TLS versions >= the default
 		for _, tlsVersionName := range crypto.ValidTLSVersions() {
 			tlsVersion := crypto.TLSVersionOrDie(tlsVersionName)
 			expectSuccess := tlsVersion >= crypto.DefaultTLSVersion()
 			config := &tls.Config{MinVersion: tlsVersion, MaxVersion: tlsVersion, InsecureSkipVerify: true}
 
+			host := strings.TrimPrefix(oc.AdminConfig().Host, "https://")
```
dusk125, line 108: nit: would you mind adding a comment why this is necessary so we don't forget a year from now :)

jacobsee: Absolutely!

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,118 @@
...
+		podClient := oc.AdminKubeClient().CoreV1().Pods("openshift-etcd")
+		err = library.WaitForPodsToStabilizeOnTheSameRevision(t, podClient, "app=etcd", 5, 24*time.Second, 5*time.Second, 30*time.Minute)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Waiting for the API server to stabilize")
+
+		_, err = waitForKASIncrementedRevision(ctx, operatorClient, "cluster", currentKASRevision)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Dialing the API with a minimum TLS version of 1.3")
```
p0lyn0mial, line 82: shouldn't we check if dialing with a version < 1.3 fails ?

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,118 @@
...
+			removeModernProfilePatch := `[{"op":"remove","path":"/spec/tlsSecurityProfile"}]`
+
+			_, err = configClient.ConfigV1().APIServers().Patch(cleanupCtx, "cluster", types.JSONPatchType, []byte(removeModernProfilePatch), metav1.PatchOptions{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			_, err = waitForKASIncrementedRevision(cleanupCtx, operatorClient, "cluster", currentRevision)
+			o.Expect(err).NotTo(o.HaveOccurred())
+		}()
+
+		g.By("Applying a JSON Patch to use the modern TLS profile")
```
p0lyn0mial, line 59: and maybe before applying the patch we could test that dialing with version < 1.3 actually works.

p0lyn0mial:
which jobs run the new test ? i have checked `metal-ipi-serial`, `aws-ovn-microshift-serial` and  `aws-ovn-serial` but didn't find the new test on the list.

test/extended/apiserver/tls.go
jacobsee: agreed - updated

p0lyn0mial:
> which jobs run the new test ? i have checked metal-ipi-serial, aws-ovn-microshift-serial and aws-ovn-serial but didn't find the new test on the list.

@jacobsee I think you need to run `make update` to actually "add" the test to the list of tests to be executed. For example: https://github.com/openshift/origin/pull/29191/commits

jacobsee:
@p0lyn0mial Thanks! I hadn't seen that. It's in there now, and I've realized that it should probably be marked [slow] as well. To your earlier comment, do we now still need to pick jobs to run this? This is my first brush with the origin tests, so I'm not sure what all is needed to get a new one plugged in correctly.

p0lyn0mial:
> To your earlier comment, do we now still need to pick jobs to run this? 

it looks like you could use `/payload-job-with-prs` to test your new test with the o/k PR, for example:

```
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn openshift/kubernetes#2135
```

xref: https://docs.ci.openshift.org/docs/release-oversight/pull-request-testing/#payload-job

we just need to find a periodic job that will run your test. maybe some serial test [from](https://amd64.ocp.releases.ci.openshift.org/releasestream/4.19.0-0.nightly/release/4.19.0-0.nightly-2025-03-25-075140) ?


so, overall:

1. we add a new test to o/o. (this PR)
2. we use the`/payload-job-with-prs` command to run the test with the o/k PR
3. we make sure the test is green
3. we merge the o/k PR
4. we merge this PR

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,146 @@
...
+	"github.com/openshift/library-go/test/library"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
+	operatorv1client "github.com/openshift/client-go/operator/clientset/versioned"
 	"github.com/openshift/library-go/pkg/crypto"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
+const tlsTestDuration = 45 * time.Minute
```
p0lyn0mial, line 21: setting a timeout in the test is unusual, most test use `TestSuite.TestTimeout`

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,146 @@
...
+
+		g.By("Checking if TLS 1.2 is usable before the modern TLS profile is applied")
+
+		// We're going to be dialing TCP directly, not connecting over HTTP as usual, so we don't want the protocol on the host.
+		tlsHost := strings.TrimPrefix(oc.AdminConfig().Host, "https://")
+
+		config := &tls.Config{MinVersion: tls.VersionTLS12, MaxVersion: tls.VersionTLS12, InsecureSkipVerify: true}
+
+		conn, err := tls.Dial("tcp4", tlsHost, config)
+		if err != nil {
```
p0lyn0mial, line 67: how about using `o.Expect` ?

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,146 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		currentKASRevision := kasStatus.Status.LatestAvailableRevision
+
+		addModernProfilePatch := `[{"op":"add","path":"/spec/tlsSecurityProfile","value":{"type":"Modern","modern":{}}}]`
+
+		_, err = configClient.ConfigV1().APIServers().Patch(ctx, "cluster", types.JSONPatchType, []byte(addModernProfilePatch), metav1.PatchOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("Waiting for etcd to stabilize")
```
p0lyn0mial, line 87: I think we should only wait for `WaitForAPIServerToStabilizeOnTheSameRevision`, as it takes the longest to roll out the servers. Besides, we're only testing the servers, even though updating the profiles affects other components as well.

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,146 @@
...
+const tlsTestDuration = 45 * time.Minute
+const tlsWaitForCleanupDuration = 10 * time.Minute
+
+var _ = g.Describe("[sig-api-machinery][Feature:APIServer][Serial][Slow]", func() {
+	defer g.GinkgoRecover()
+
+	oc := exutil.NewCLIWithoutNamespace("apiserver")
+
+	g.It("TestTLSModernProfile", func() {
+		ctx, ctxCancelFn := context.WithTimeout(context.Background(), tlsTestDuration)
```
p0lyn0mial, line 30: I would just use `ctx := context.TODO()`

test/extended/apiserver/tls.go
```diff
@@ -62,3 +177,20 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 
 	})
 })
+
+func waitForKASIncrementedRevision(ctx context.Context, operatorClient operatorv1client.Interface, name string, currentRevision int32) (int32, error) {
```
p0lyn0mial, line 181: I would remove this function and just use `WaitForAPIServerToStabilizeOnTheSameRevision`

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,146 @@
...
+		ctx, ctxCancelFn := context.WithTimeout(context.Background(), tlsTestDuration)
+		defer ctxCancelFn()
+
+		t := g.GinkgoT()
+
+		configClient := oc.AdminConfigClient()
+
+		operatorClient := oc.AdminOperatorClient()
+
+		defer func() {
```
p0lyn0mial, line 39: maybe we should use `DeferCleanup` - i think that it will be executed even if the test panics ...

jacobsee:
@p0lyn0mial Addressed several things - however the wait for etcd to stabilize was a deliberate decision after a chat with @dusk125 however, so maybe we should bring that context to this thread here. The thinking was that it is also desirable to have a test for etcd's TLS 1.3 stability, and the test would have 90% the same code as this one (just with a different check in the middle), so we might we well ensure that both stabilize here to avoid duplicating an already-slow test.

jacobsee:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn openshift/kubernetes#2135

jacobsee:
@p0lyn0mial I might not be looking in the right place/way but I'm not seeing jobs that run a `[serial][slow]` test - could it be that it wouldn't get picked up by any? https://sippy.dptools.openshift.org/sippy-ng/tests/4.19?filters=%257B%2522items%2522%253A%255B%257B%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522%255Bserial%255D%2522%257D%252C%257B%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522%255Bslow%255D%2522%257D%255D%252C%2522linkOperator%2522%253A%2522and%2522%257D&pageSize=100&sort=asc&sortField=current_working_percentage

There are a few tests in this repository in a similar boat (like `TestImportRepositoryFromBlockedRegistry`) that I don't see runs for.

This new test works when I run it manually but do you have any tips for getting it integrated into existing jobs?

p0lyn0mial:
> @p0lyn0mial I might not be looking in the right place/way but I'm not seeing jobs that run a [serial][slow] test - could it be that it wouldn't get picked up by any? https://sippy.dptools.openshift.org/sippy-ng/tests/4.19?filters=%257B%2522items%2522%253A%255B%257B%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522%255Bserial%255D%2522%257D%252C%257B%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522%255Bslow%255D%2522%257D%255D%252C%2522linkOperator%2522%253A%2522and%2522%257D&pageSize=100&sort=asc&sortField=current_working_percentage

> There are a few tests in this repository in a similar boat (like TestImportRepositoryFromBlockedRegistry) that I don't see runs for.

> This new test works when I run it manually but do you have any tips for getting it integrated into existing jobs?


@wangke19 @bertinatto could you help with ^ ?

p0lyn0mial:
> @p0lyn0mial Addressed several things - however the wait for etcd to stabilize was a deliberate decision after a chat with @dusk125 however, so maybe we should bring that context to this thread here. The thinking was that it is also desirable to have a test for etcd's TLS 1.3 stability

WDYM ? I think that jobs in CI run monitors that ensure the overall stability of the cluster. Otherwise, each individual test would have to implement this functionality on its own.

p0lyn0mial:
@jacobsee on second thought, maybe we should actually test more control plane components. 

[Here is a list](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/security_and_compliance/tls-security-profiles#tls-profiles-kubernetes-configuring_tls-security-profiles) of components that will be affected after changing the profile. 

For instance, the aggregated APIs (OpenShift API server and OAuth API server) are behind the kube-apiserver (KAS). So, it's not enough to just check the connection to KAS. KAS has a built-in proxy that forwards requests to the aggregated APIs. To properly test the aggregated APIs and the proxy client in KAS, we would need to send an HTTPS request through KAS. 

I believe it's worth the effort — do you agree?  Should we also test the other components?

jacobsee:
@p0lyn0mial I do agree that testing other components makes sense if we anticipate an impact to them, unsure if it makes more sense to try to wrap it all into one PR upfront or just get start with something like this... Regarding that impact to other components, I tried to solicit some additional opinions as I haven't spent much time working with them and found out that what we're doing here may actually fall under the coverage of an existing test. I will track that down in the morning.

p0lyn0mial:
@jacobsee  after the profile change, only the minimum acceptable TLS version will be allowed. I think we should test the affected components and verify that the platform remains stable.

To test the components, it will probably be enough to send a TLS/HTTP2 request. To make sure we haven’t accidentally broken anything else, we’ll likely need to check if all operators—or at least the affected ones—aren’t reporting any errors. I’m not sure if we can rely on the existing monitors for this, or if we should write some custom code to check the status of the operators.

It seems to me that before we merge the code into o/k, we need to make sure the platform is functioning properly. Do you agree?

p0lyn0mial:
To make it more concrete, I think we could:

1. Identify the affected components after the profile change.
2. Check whether version 1.3 is disallowed for each component.
3. Update the profile.
4. Wait for the KAS rollout — we could take a shortcut here and only wait for the KAS, as it takes the longest to roll out.
5. Verify that version 1.3 is now allowed for each component. For most components, we can use the TLS client. The aggregated APIs might be a bit special — for the aggregated APIs, we need to construct the proper URL path so that the KAS proxy forwards traffic to the appropriate server responsible for that group.
6. Read the operator resources of the affected components and ensure the operators are not degraded and are not progressing.

Initially, I thought we could rely on the monitors for step 6, but I'm not sure what exactly they check or whether all jobs actually run them. @dgoodwin maybe you can chime in on whether relying on the monitors would be enough?

jacobsee:
@p0lyn0mial A couple of thoughts:

The Kube API server (and the OAuth server, which I have now also added a check for) are easy to check because there is a route to those services from outside the cluster, where the origin tests will be running. For other components, do you have any tests or otherwise examples you can point me towards in our infrastructure where we need to access the cluster network to make requests directly against services that are not exposed outside the cluster, for testing the TLS dialing? Is the accepted answer there to set up a port forward to the test for each service?

I've also added a request that requires the aggregated OCP API to be working, but it is a regular request through a client, as I don't believe that `tls.Dial` with a path component is valid... I think we would need to hit the OCP API server from the test directly, not through the Kube API's proxy.

I've also added progressing/degraded checks for cluster operators that appeared to be affected.

p0lyn0mial:
@jacobsee oh, you're right, the tests are run from outside the cluster. 

Mhm, I don't have any examples. I think we could either use port-forward or schedule a pod with a custom binary that performs the check. Not sure ,maybe using `exec.CommandContext(ctx, "oc", "port-forward", ...)` would actually be easier. 

Also, I believe other components expose an HTTPS endpoint for metrics and health/liveness probes. For probes, we could potentially rely on the kubelet and the `StaticPodsDegraded` condition (at least for the static pods). For metrics, I doubt the monitoring operator would go degraded just because it can’t scrape metrics. Maybe @dgrisonnet could chime in here? For monitoring, maybe we could simply try scraping the metrics ourselves. 

But the more I think about how to test this change, the more I lean toward simply creating a periodic job that updates the profile, ensures the change has been observed by the affected components and then simply runs some (conformance) tests. 

I’ve also started an internal thread about it. We might also need the test you created, at the very least, we should verify that the affected components actually picked up the new TLS profile.

dgrisonnet:
I don't think cluster-monitoring-operator would report degraded if a target is unavailable, but you could rely on the [TargetDown](https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/TargetDown.md) alert.

I don't remember if we have a catch all for pods liveness, but you could always rely on kubelet's `prober_probe_total` metric.

jacobsee:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn https://github.com/openshift/kubernetes/pull/2135

jacobsee:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn openshift/kubernetes#2135

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,360 @@
...
 	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/origin/test/extended/scheme"
 	exutil "github.com/openshift/origin/test/extended/util"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
 )
 
+var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
```
p0lyn0mial, line 29: I haven’t checked in detail, but it looks like what we need – thanks! :) 

If you don’t mind, please add a comment explaining that this test only checks whether the components pick up the TLS profile, and that there is a periodic test which validates the platform more holistically by setting a TLS profile and running the e2e tests.

dusk125 (commented):
/lgtm

dusk125:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-tls-13 openshift/origin#29611

test/extended/apiserver/tls.go
```diff
@@ -376,7 +409,7 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 			host := strings.TrimPrefix(oc.AdminConfig().Host, "https://")
 
 			{
-				conn, err := tls.Dial("tcp", host, config)
+				conn, err := tls.DialWithDialer(dialer, "tcp", host, config)
```
dusk125, line 412: Dialer is out of scope here

test/extended/apiserver/tls.go
```diff
@@ -401,7 +434,7 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 			config := &tls.Config{CipherSuites: []uint16{cipher}, InsecureSkipVerify: true}
 
 			{
-				conn, err := tls.Dial("tcp", oc.AdminConfig().Host, config)
+				conn, err := tls.DialWithDialer(dialer, "tcp", oc.AdminConfig().Host, config)
```
dusk125, line 437: Dialer is out of scope here

dusk125 (changes requested):

test/extended/apiserver/tls.go
jacobsee: 😶‍🌫️

sjenning:
Worth noting that this test will not work on Hypershift because many of the components to which connections are made are not accessible directly (KCM, OCM, etc)

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,458 @@
...
+	g.It("TestTLSMinimumVersions", func() {
+		ctx := context.TODO()
+
+		coreClient, err := e2e.LoadClientset(true)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		isMicroShift, err := exutil.IsMicroShiftCluster(coreClient)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		if isMicroShift {
```
sjenning, line 50: we need to add hypershift to the skip list here

jacobsee: A portion of this that was failing on hypershift is skipped later on down, on line 141 where the test returns early. We need to skip the entire test on hypershift instead?

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,458 @@
...
+		// Tests below this are not relevant in external control plane topology mode
+		if *controlPlaneTopology == ocpv1.ExternalTopologyMode {
+			return
+		}
+
+		//////
+
+		g.By("Checking the OAuth server")
+
+		oauthRoute, err := oc.AdminRouteClient().RouteV1().Routes("openshift-authentication").Get(ctx, "oauth-openshift", metav1.GetOptions{})
```
p0lyn0mial, line 149: since all components we wish to test are exposed as services, I’m wondering if it wouldn’t be easier to simply use:
```
exec.CommandContext(ctx, "oc", "port-forward", "service/name", "local:remote", "-n", "namespace")
```

The benefit is that this command would automatically select a pod for us, and might even include some built-in resiliency. 

To improve the stability of the test, we could consider calling this method in a loop, say, up to 3 attempts, each time we could even select a different local port (base + index). WDYT?

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,458 @@
...
+
+				// For any other network types, use as is
+				return (&net.Dialer{}).DialContext(ctx, network, address)
+			},
+		}
+
+		dialer := &net.Dialer{
+			Timeout:       30 * time.Second,
+			KeepAlive:     30 * time.Second,
+			DualStack:     true,
```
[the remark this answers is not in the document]
jacobsee: Is there something that can be fetched to look up whether this is intended to be a dual stack or IPv6-only test run? The fact that the dial was first attempting to do a DNS lookup using IPv4 in an IPv6-only job is the reason for the custom dialer/resolver above - to force the resolver to try IPv6 first.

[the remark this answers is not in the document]
p0lyn0mial: We could get the network config and check the `ClusterNetwork`. However, I'm not sure if finding both IP families is sufficient, support from the underlying network plugin might also be required.

Perhaps the most universal approach would be to schedule a pod and check how many IPs were assigned (https://github.com/openshift/origin/blob/06e240d30d3a6be132b901cdaa2184e70f034fed/test/extended/networking/util.go#L607).

test/extended/apiserver/tls.go
p0lyn0mial: ping about ^

jacobsee: I've consolidated all tests to the port-forward based logic instead of having the exposed systems going through the route instead, and also added retry + backoff.

I'm not sure I see a benefit in switch to calling the `oc` binary rather than directly using `k8s.io/client-go/tools/portforward` as this is currently doing... Unless I'm missing something, it looks to me like the CLI also selects a pod client-side and doesn't appear to add any particular resiliency or retries. If you think there is value in attempting multiple pods and randomizing local ports, I can do some further rework soon, though time is a tad limited.

p0lyn0mial: The benefit of using a service is that only healthy pods will be selected. Currently, we select the first pod without checking its state. 

The advantage of calling `oc` directly is that there's less code for us to maintain. I think it comes down to this:
```
cmd := exec.CommandContext(ctx, "oc", "port-forward", "service/name", "local:remote", "-n", "namespace")

stdout, err := cmd.StdoutPipe()

o.Expect(cmd.Start()).NotTo(o.HaveOccurred())

scanner := bufio.NewScanner(stdout)
o.Expect(scanner.Scan()).To(o.BeTrue())
o.Expect(scanner.Err()).NotTo(o.HaveOccurred())

e2e.Logf("command output: %s", scanner.Text())
```


Most of the code is there to capture the output for debugging purposes - maybe there's a way to optimize it. 

In my opinion, this is easier than dealing with channels, goroutines, spdy dialers, etc.

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,523 @@
...
+		config.NegotiatedSerializer = scheme.Codecs
+	}
+	if len(config.UserAgent) == 0 {
+		config.UserAgent = rest.DefaultKubernetesUserAgent()
+	}
+	config.APIPath = "/api"
+	return &config
+}
+
+func getIPFamilyForCluster(client exutil.CLI, namespace string) IPFamily {
```
p0lyn0mial: Is the code related to getting the current IP family exactly the same as the one used in the existing network tests? 

If so, maybe we could move it to a common place and reuse it?

test/extended/apiserver/tls.go
jacobsee: switched to exec'ing the `oc` cli, letting it pick the pod, and randomizing the local port

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+			g.Skip("tls configuration for the apiserver resource is not applicable to microshift or hypershift clusters - skipping")
+		}
+
+		ipFamily := getIPFamilyForCluster(*oc, oc.Namespace())
+
+		if ipFamily != IPv4 {
+			g.Skip("tls configuration is only tested on IPv4 clusters, skipping")
+		}
+
+		insecure := true
```
p0lyn0mial: unused

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+		configFlags := &genericclioptions.ConfigFlags{}
```
p0lyn0mial: unused, please remove.

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+				_, err = tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldNotWork)
+				o.Expect(err).To(o.HaveOccurred())
+			},
+		)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+	})
+})
+
+func ForwardPortsAndExecute(serviceName string, namespace string, ports []string, maxConnectRetries int, initialBackoff time.Duration, toExecute func(int)) error {
```
p0lyn0mial: I think that this function could be simplified a bit:

```
func forwardPortAndExecute(serviceName string, namespace string, remotePort string, toExecute func(localPort int)) error {
	var err error
	for i := 0; i < 3; i++ {
		if err = func() error {
			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
			defer cancel()
			localPort := rand.Intn(65534-1025) + 1025
			args := []string{
				"port-forward",
				fmt.Sprintf("svc/%s", serviceName),
				fmt.Sprintf("%d:%s", localPort, remotePort),
				"-n", namespace,
			}

			cmd := exec.CommandContext(ctx, "oc", args...)

			stdout, stderr, err := e2e.StartCmdAndStreamOutput(cmd)
			if err != nil {
				return err
			}
			defer stdout.Close()
			defer stderr.Close()
			defer e2e.TryKill(cmd)

			e2e.Logf("oc port-forward output: %s", readPartialFrom(stdout, 1024))

			toExecute(localPort)

			return nil
		}(); err != nil {
			err = fmt.Errorf("failed to start oc port-forward command: %w", err)
			e2e.Logf(err.Error())
			time.Sleep(2 * time.Second)
		}
	}

	return err
}

func readPartialFrom(r io.Reader, maxBytes int) string {
	buf := make([]byte, maxBytes)
	n, err := r.Read(buf)
	if err != nil && err != io.EOF {
		return fmt.Sprintf("error reading: %v", err)
	}
	return string(buf[:n])
}
```

WDYT?

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+
+		err = ForwardPortsAndExecute(
+			"apiserver",
+			"openshift-kube-apiserver",
+			[]string{"443"},
+			3,
+			200*time.Millisecond,
+			func(port int) {
+				conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
+				o.Expect(err).NotTo(o.HaveOccurred())
```
p0lyn0mial: We could change the signature of the `toExecute` function to return an err so that we could retry on errors.

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+		}
+
+		//////
+
+		g.By("Checking the Kube API server")
+
+		err = ForwardPortsAndExecute(
+			"apiserver",
+			"openshift-kube-apiserver",
+			[]string{"443"},
```
p0lyn0mial: why [] ?

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+			case configv1.TLSProfileModernType:
+				tlsShouldWork = &tls.Config{MinVersion: tls.VersionTLS13, MaxVersion: tls.VersionTLS13, InsecureSkipVerify: true}
+				tlsShouldNotWork = &tls.Config{MinVersion: tls.VersionTLS12, MaxVersion: tls.VersionTLS12, InsecureSkipVerify: true}
+
+			default:
+				g.Skip("only intermediate or modern profiles are tested")
+			}
+		}
+
+		//////
```
p0lyn0mial: please remove

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+
+		g.By("Checking the OAuth server")
+
+		err = ForwardPortsAndExecute(
+			"oauth-openshift",
+			"openshift-authentication",
+			[]string{"443"},
+			3,
+			200*time.Millisecond,
+			func(port int) {
```
p0lyn0mial: most of the time this function is repeated, extract to a common function and reuse.

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+	})
+	return podIPs, err
+}
+
 var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 	defer g.GinkgoRecover()
 
 	oc := exutil.NewCLI("apiserver")
 
 	g.It("TestTLSDefaults", func() {
```
p0lyn0mial: how does this test differs from the one that has been added ?

test/extended/apiserver/tls.go
p0lyn0mial: ping ^ (we could open a new PR to refactor)

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,467 @@
...
+			3,
+			200*time.Millisecond,
+			func(port int) {
+				conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
+				o.Expect(err).NotTo(o.HaveOccurred())
+
+				conn.Close()
+
+				_, err = tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldNotWork)
+				o.Expect(err).To(o.HaveOccurred())
```
p0lyn0mial: what type/kind of err do we expect here? how are we going to distinguish it from a network err ?

test/extended/apiserver/tls.go
jacobsee: added, but modified to let the check return early on success instead of needing to do the port forward & function call 3 times on every invocation

test/extended/apiserver/tls.go
jacobsee: The older test goes deeper and tests acceptable cipher suites, but it only checks the API server, and only when it's running under the default configuration.

test/extended/apiserver/tls.go
jacobsee: It's mostly the same, but it adds a security context to let the check pod run in restricted contexts

jacobsee: That and the port forwarder are probably both reusable. Where should they live?

test/extended/apiserver/tls.go
```diff
@@ -62,3 +401,24 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
 
 	})
 })
+
+// CheckTLSConnection tries to connect to localhost:port with the provided TLS configs.
+func CheckTLSConnection(port int, tlsShouldWork, tlsShouldNotWork *tls.Config) error {
```
p0lyn0mial: make it private.

test/extended/apiserver/tls.go
```diff
@@ -62,3 +401,24 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
...
+	conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
+	if err != nil {
+		return fmt.Errorf("should work: %w", err)
+	}
+	conn.Close()
```
p0lyn0mial: check the err, otherwise a linter might complain.

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,370 @@
...
+			if err != nil {
+				return err
+			}
+			defer stdout.Close()
+			defer stderr.Close()
+			defer e2e.TryKill(cmd)
+
+			e2e.Logf("oc port-forward output: %s", readPartialFrom(stdout, 1024))
+
+			if err := toExecute(localPort); err != nil {
```
p0lyn0mial: nit: could be simplified to: `return toExecute(localPort)`

test/extended/apiserver/tls.go
```diff
@@ -62,3 +401,24 @@ var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
...
+func CheckTLSConnection(port int, tlsShouldWork, tlsShouldNotWork *tls.Config) error {
+	conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
+	if err != nil {
+		return fmt.Errorf("should work: %w", err)
+	}
+	conn.Close()
+
+	_, err = tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldNotWork)
+	if err == nil {
+		return fmt.Errorf("should not work: connection unexpectedly succeeded")
```
p0lyn0mial: should we close the connection since it was established successfully?

p0lyn0mial: for example: `return fmt.Errorf("should not work: connection unexpectedly succeeded, closing conn status: %v", conn.Close())`

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,370 @@
...
+			"openshift-etcd",
+			"2379",
+			func(port int) error {
+				// We aren't actually going through mTLS authentication with etcd to communicate
+				// with it - just checking TLS protocol versions. So, if it throws a "bad certificate"
+				// error, we're past the version check and consider it a success for this test.
+
+				conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
+				if err != nil {
+					if !strings.Contains(err.Error(), "remote error: tls: bad certificate") {
```
p0lyn0mial: does this err indicates that the client cert was rejected ?

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,370 @@
...
+						return fmt.Errorf("should work: %w", err)
+					}
+				} else {
+					conn.Close()
+				}
+				_, err = tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldNotWork)
+				if err == nil {
+					return fmt.Errorf("should not work: connection unexpectedly succeeded")
```
p0lyn0mial: we should close the connection if no err, for example: `return fmt.Errorf("should not work: connection unexpectedly succeeded, closing conn status: %v", conn.Close())`

test/extended/apiserver/tls.go
```diff
@@ -1,31 +1,370 @@
...
+				// with it - just checking TLS protocol versions. So, if it throws a "bad certificate"
+				// error, we're past the version check and consider it a success for this test.
+
+				conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
+				if err != nil {
+					if !strings.Contains(err.Error(), "remote error: tls: bad certificate") {
+						return fmt.Errorf("should work: %w", err)
+					}
+				} else {
+					conn.Close()
```
p0lyn0mial: check the err, otherwise a linter might complain.

test/extended/apiserver/tls.go
p0lyn0mial: Try to expose https://github.com/openshift/origin/blob/e2e54ef00b13b6b8146c7d54e33c3c3678e84d57/test/extended/networking/util.go#L607 directly. 

(just change the signature)

test/extended/apiserver/tls.go
jacobsee: Yes, that is the exact error that we see when failing to mTLS with etcd (because we aren't attempting to successfully auth to it). But it is only thrown when the connection negotiates as TLS <1.3. TLS 1.3 delays mTLS errors until post-handshake, so we actually hit the err==nil path here if that is the version negotiated.

test/extended/apiserver/tls.go
p0lyn0mial: ping

p0lyn0mial:
@jacobsee  This PR looks good to me. Thanks for all your work. Before merging, let's try creating a new PR for https://github.com/openshift/origin/pull/29611#discussion_r2236224031, so we can reuse some code instead of copying it. WDYT?

test/extended/apiserver/tls.go
```diff
@@ -1,33 +1,305 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	e2e "k8s.io/kubernetes/test/e2e/framework"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/origin/test/extended/networking"
 	exutil "github.com/openshift/origin/test/extended/util"
 )
 
+type IPFamily string
```
p0lyn0mial: it this var now required/used?

test/extended/apiserver/tls.go
```diff
@@ -1,33 +1,305 @@
...
+
+const (
+	namespace = "apiserver-tls-test"
+
+	// How often to poll pods and nodes.
+	poll = 5 * time.Second
```
p0lyn0mial: is this var used ?

p0lyn0mial (commented):
lgtm

p0lyn0mial:
> [sig-api-machinery][Feature:APIServer] TestTLSDefaults [Suite:openshift/conformance/parallel]

failed on `e2e-aws-ovn-microshift` :(

p0lyn0mial:
@jacobsee actually, there are more failures:

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29611/pull-ci-openshift-origin-main-e2e-vsphere-ovn/1973225414637654016

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29611/pull-ci-openshift-origin-main-e2e-metal-ipi-ovn/1973225405426962432

p0lyn0mial:
/lgtm cancel

there are failing jobs

wangke19:
@jacobsee I debugged this pr in local, here is code change: https://privatebin.corp.redhat.com/?aa8939acdfa5271d#2rWYNDacmuSKeNRmkDjtc7S1RGSAj5938DgQS6gjNZkV, you can refer to it.
Both tests have been passed with different TLS profiles.

jacobsee:
@wangke19 Apologies, I was away from home for a couple of weeks there and not checking GitHub notifications - it appears that paste has expired, can you possibly send another? Thanks

wangke19:
@jacobsee Never mind, I've notices your status. I will share you the coe in comments.
```
package apiserver

import (
	"context"
	"crypto/tls"
	"fmt"
	"io"
	"math/rand"
	"os/exec"
	"strings"
	"time"

	g "github.com/onsi/ginkgo/v2"
	o "github.com/onsi/gomega"

	"github.com/openshift/library-go/pkg/crypto"
	"github.com/openshift/origin/test/extended/networking"
)

const (
	namespace = "apiserver-tls-test"
)

// This test only checks whether components are serving the proper TLS version based
// on the expected version set in the TLS profile config. It is a part of the
// openshift/conformance/parallel test suite, and it is expected that there are jobs
// which run that entire conformance suite against clusters running any TLS profiles
// that there is a desire to test.
var _ = g.Describe("[sig-api-machinery][Feature:APIServer]", func() {
	defer g.GinkgoRecover()

	var oc = exutil.NewCLI("apiserver")
	var ctx = context.Background()

	g.BeforeEach(func() {
		isMicroShift, err := exutil.IsMicroShiftCluster(oc.AdminKubeClient())

		isHyperShift, err := exutil.IsHypershift(ctx, oc.AdminConfigClient())

		if isMicroShift || isHyperShift {
			g.Skip("TLS configuration for the apiserver resource is not applicable to MicroShift or HyperShift clusters - skipping")
		}

		hasIPv4, _, err := networking.GetIPAddressFamily(oc)
		if !hasIPv4 {
			g.Skip("TLS configuration is only tested on IPv4 clusters, skipping")
		}
	})

	// --------------------------------------------------------------------------------
	// TestTLSMinimumVersions
	// --------------------------------------------------------------------------------
	g.It("TestTLSMinimumVersions", func() {

		g.By("Getting the APIServer configuration")
		config, err := oc.AdminConfigClient().ConfigV1().APIServers().Get(ctx, "cluster", metav1.GetOptions{})

		g.By("Determining expected TLS behavior based on the cluster's TLS profile")
		var tlsShouldWork, tlsShouldNotWork *tls.Config
		switch {
		case config.Spec.TLSSecurityProfile == nil,
			config.Spec.TLSSecurityProfile.Type == configv1.TLSProfileIntermediateType:
			tlsShouldWork = &tls.Config{MinVersion: tls.VersionTLS12, MaxVersion: tls.VersionTLS13, InsecureSkipVerify: true}
			tlsShouldNotWork = &tls.Config{MinVersion: tls.VersionTLS11, MaxVersion: tls.VersionTLS11, InsecureSkipVerify: true}
			g.By("Using intermediate TLS profile: connections with TLS ≥1.2 should work, <1.2 should fail")
		case config.Spec.TLSSecurityProfile.Type == configv1.TLSProfileModernType:
			tlsShouldWork = &tls.Config{MinVersion: tls.VersionTLS13, MaxVersion: tls.VersionTLS13, InsecureSkipVerify: true}
			tlsShouldNotWork = &tls.Config{MinVersion: tls.VersionTLS12, MaxVersion: tls.VersionTLS12, InsecureSkipVerify: true}
			g.By("Using modern TLS profile: only TLS 1.3 connections should succeed")
		default:
			g.Skip("Only intermediate or modern profiles are tested")
		}

		targets := []struct {
			name, namespace, port string
		}{
			{"apiserver", "openshift-kube-apiserver", "443"},
			{"oauth-openshift", "openshift-authentication", "443"},
			{"kube-controller-manager", "openshift-kube-controller-manager", "443"},
			{"scheduler", "openshift-kube-scheduler", "443"},
			{"api", "openshift-apiserver", "443"},
			{"api", "openshift-oauth-apiserver", "443"},
			{"machine-config-controller", "openshift-machine-config-operator", "9001"},
		}

		g.By("Verifying TLS behavior for core control plane components")
		for _, target := range targets {
			g.By(fmt.Sprintf("Checking %s/%s on port %s", target.namespace, target.name, target.port))
			err = forwardPortAndExecute(target.name, target.namespace, target.port,
				func(port int) error { return checkTLSConnection(port, tlsShouldWork, tlsShouldNotWork) })
		}

		g.By("Checking etcd's TLS behavior")
		err = forwardPortAndExecute("etcd", "openshift-etcd", "2379", func(port int) error {
			conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
			if err != nil && !strings.Contains(err.Error(), "remote error: tls: bad certificate") {
				return fmt.Errorf("should work: %w", err)
			}
			if conn != nil {
				_ = conn.Close()
			}
			conn, err = tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldNotWork)
			if err == nil {
				conn.Close()
				return fmt.Errorf("should not work: connection unexpectedly succeeded")
			}
			return nil
		})
	})

	// --------------------------------------------------------------------------------
	// TestTLSDefaults
	// --------------------------------------------------------------------------------
	g.It("TestTLSDefaults", func() {
		t := g.GinkgoT()

		_, err := e2e.LoadClientset(true)

		g.By("Getting the APIServer config")
		config, err := oc.AdminConfigClient().ConfigV1().APIServers().Get(ctx, "cluster", metav1.GetOptions{})

		if config.Spec.TLSSecurityProfile != nil &&
			config.Spec.TLSSecurityProfile.Type != configv1.TLSProfileIntermediateType {
			g.Skip("Cluster TLS profile is not default (intermediate), skipping cipher defaults check")
		}

		g.By("Verifying TLS version behavior")
		for _, tlsVersionName := range crypto.ValidTLSVersions() {
			tlsVersion := crypto.TLSVersionOrDie(tlsVersionName)
			expectSuccess := tlsVersion >= crypto.DefaultTLSVersion()
			cfg := &tls.Config{MinVersion: tlsVersion, MaxVersion: tlsVersion, InsecureSkipVerify: true}
			host := strings.TrimPrefix(oc.AdminConfig().Host, "https://")

			conn, err := tls.Dial("tcp", host, cfg)
			if err == nil {
				_ = conn.Close()
			}
			if success := err == nil; success != expectSuccess {
				t.Errorf("Expected success %v, got %v with TLS version %s dialing master", expectSuccess, success, tlsVersionName)
			}
		}

		g.By("Verifying cipher suites")
		defaultCiphers := map[uint16]bool{}
		for _, c := range crypto.DefaultCiphers() {
			defaultCiphers[c] = true
		}

		for _, cipherName := range crypto.ValidCipherSuites() {
			cipher, err := crypto.CipherSuite(cipherName)
			if err != nil {
				t.Fatal(err)
			}
			expectFailure := !defaultCiphers[cipher]
			cfg := &tls.Config{CipherSuites: []uint16{cipher}, InsecureSkipVerify: true}

			conn, err := tls.Dial("tcp", oc.AdminConfig().Host, cfg)
			if err == nil {
				if expectFailure {
					t.Errorf("Expected failure on cipher %s, got success dialing master. Closing conn: %v", cipherName, conn.Close())
				}
			}
		}
	})
})

func forwardPortAndExecute(serviceName, namespace, remotePort string, toExecute func(localPort int) error) error {
	var err error
	for i := 0; i < 3; i++ {
		if err = func() error {
			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
			defer cancel()
			localPort := rand.Intn(65534-1025) + 1025
			args := []string{"port-forward", fmt.Sprintf("svc/%s", serviceName), fmt.Sprintf("%d:%s", localPort, remotePort), "-n", namespace}

			cmd := exec.CommandContext(ctx, "oc", args...)
			stdout, stderr, err := e2e.StartCmdAndStreamOutput(cmd)
			if err != nil {
				return err
			}
			defer stdout.Close()
			defer stderr.Close()
			defer e2e.TryKill(cmd)

			e2e.Logf("oc port-forward output: %s", readPartialFrom(stdout, 1024))
			return toExecute(localPort)
		}(); err == nil {
			return nil
		} else {
			e2e.Logf("failed to start oc port-forward command or test: %v", err)
			time.Sleep(2 * time.Second)
		}
	}
	return err
}

func readPartialFrom(r io.Reader, maxBytes int) string {
	buf := make([]byte, maxBytes)
	n, err := r.Read(buf)
	if err != nil && err != io.EOF {
		return fmt.Sprintf("error reading: %v", err)
	}
	return string(buf[:n])
}

func checkTLSConnection(port int, tlsShouldWork, tlsShouldNotWork *tls.Config) error {
	conn, err := tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldWork)
	if err != nil {
		return fmt.Errorf("should work: %w", err)
	}
	_ = conn.Close()

	conn, err = tls.Dial("tcp", fmt.Sprintf("localhost:%d", port), tlsShouldNotWork)
	if err == nil {
		return fmt.Errorf("should not work: connection unexpectedly succeeded, closing conn status: %v", conn.Close())
	}
	if !strings.Contains(err.Error(), "protocol version") &&
		!strings.Contains(err.Error(), "no supported versions satisfy") &&
		!strings.Contains(err.Error(), "handshake failure") {
		return fmt.Errorf("should not work: got error, but not a TLS version mismatch: %w", err)
	}
	return nil
}
```

wangke19:
/retit OCPBUGS-56834: TLS 1.3 / Modern profile tests

wangke19:
/retitle OCPBUGS-56834: TLS 1.3 / Modern profile tests

wangke19:
/retitle OCPBUGS-64799: TLS 1.3 / Modern profile tests

wangke19:
@p0lyn0mial Please review the updated tests.

wangke19:
/cc @sanchezl 
I remembered that you implemented this feature.

PR #29614, opened 2025-03-20
CNTRLPLANE-5: Fix decoding of CBOR data in etcd test

bertinatto:
The Unmarshal function uses strict decoding by default. This means that the target object must define all fields present in the CBOR data. Otherwise, it returns an unknown field error. This patch fixes that by using a intermediary map[string]any.

/assign @benluddy


bertinatto:
This is low risk because it only changes the CBOR case, which shouldn't match with the current feature gates we have.

PR #29616, opened 2025-03-21
Fix broken intervals charts

dgoodwin:
The timelines-chart seems to have changed at this unpkg location, if you
load that in a browser and search for the TimelinesChart object we use,
it's no longer there, though it does load some js.

The github repo for the actual project
https://github.com/vasturiano/timelines-chart
recommends this new url and I've confirmed this works by downloading an
html intervals file and editing it.


dgoodwin:
Sorry forgot to update bindata, needs another lgtm.

PR #29618, opened 2025-03-21
OCPBUGS-52358: Fix egress firewall tests by updating the URL from docs.redhat.com to redhat.com

arkadeepsen:
This PR updates the egressfirewall tests to use redhat.com instead of docs.redhat.com as there are a lot of failure encountered for this URL. The issue with docs.redhat.com is that the TTL is very low (20s) because of which the ovnkube-node and the test pod are resolving to different IP addresses as shown in the following analysis:
- https://github.com/openshift/release/pull/62968#issuecomment-2740905648
- https://github.com/openshift/release/pull/62968#issuecomment-2740933499

The TTL of redhat.com is very large (2695s) which will ensure that the failure of the test is very less.


arkadeepsen:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.19-upgrade-from-stable-4.18-e2e-azure-ovn-upgrade 10

arkadeepsen:
Adding the label as the test continues to be regressed

PR #29621, opened 2025-03-24
OCPBUGS-37799: Disable:Broken for [sig-builds][Feature:Builds][Slow] can use private repositories as build input build using an HTTP token should be able to clone source code via an HTTP token

kunalmemane:
Disable:Broken for [sig-builds][Feature:Builds][Slow] can use private repositories as build input build using an HTTP token should be able to clone source code via an HTTP token [apigroup:build.openshift.io]


sosiouxme:
per [slack request](https://redhat-internal.slack.com/archives/C01CQA76KMX/p1745474014730919)
/lgtm

sosiouxme:
/approve
(thought /lgtm automatically did that)

PR #29625, opened 2025-03-26
NO-ISSUE: Wait for MachineConfiguration status to be upto date

djoshy:
Before running through the remainder for test, we need to ensure that the fixture applied has been propagated through the status. This adds a short wait check to accounts for this.


djoshy:

This shouldn't add any risk and is adding a short wait at the start of each boot image update test.

isabella-janssen (commented):
/lgtm
/approve

PR #29626, opened 2025-03-27
NO-ISSUE: Clean up MachineConfiguration fixture apply 

djoshy:
Follow-up to https://github.com/openshift/origin/pull/29625 

This encapsulate the apply and wait into a single function, and also added it to a couple of spots that I missed in the original PR.


isabella-janssen (commented):
/lgtm

PR #29643, opened 2025-04-02
MCO-1631: Remove "[sig-arch][Early] CRDs for openshift.io should have a status in the CRD schema"

RishabhSaini:
"[sig-arch][Early] CRDs for openshift.io should have a status in the CRD schema" is no longer relevant.

We only want to be checking for:
If a CRD has a "status" field in the schema but no subresource.status defined. (which causes several problems). Its covered by an existing test "[sig-arch][Early] CRDs for openshift.io should have subresource.status". Hence leaving no need for the prior test to exist.

This work blocks https://github.com/openshift/api/pull/2257, which is necessary to ship PIS as a feature for 4.19

### DIFF
--- a/test/extended/operators/check_crd_status.go
+++ b/test/extended/operators/check_crd_status.go
@@ -73,73 +73,6 @@ func checkSubresourceStatus(crdItemList []apiextensionsv1.CustomResourceDefiniti
 	return failures
 }
 
-// checkStatusInSchema returns a list of names of CRDs that don't have a "status" in the CRD schema.
-// For now, it ignores the ones that are currently failing.
-func checkStatusInSchema(crdItemList []apiextensionsv1.CustomResourceDefinition) []string {
-
-	// These CRDs, at the time this test was written, do not have a "status" in the CRD schema
-	// and subresource.status.
-	// These can be skipped for now but we don't want the number to increase.
-	// These CRDs should be tidied up over time.
-	//
-	exceptionsList := sets.NewString(
-		"builds.config.openshift.io",
-		"clusternetworks.network.openshift.io",
-		"consoleclidownloads.console.openshift.io",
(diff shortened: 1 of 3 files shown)


test/extended/operators/check_crd_status.go
```diff
@@ -98,6 +98,7 @@ func checkStatusInSchema(crdItemList []apiextensionsv1.CustomResourceDefinition)
 		"imagecontentpolicies.config.openshift.io",
 		"imagecontentsourcepolicies.operator.openshift.io",
 		"machineconfigs.machineconfiguration.openshift.io",
+		"pinnedimagesets.machineconfiguration.openshift.io",
```
deads2k, line 101: /hold

don't do this.  You could add checking to ensure no status exists, but this opens up to severe errors around permissioning and semantics we cannot fix.

deads2k: the original goal of the testing was

> they aren't required to have status. Some objects don't. But if they have status in the schema, they must have a status subresource

https://github.com/openshift/origin/pull/26954#discussion_r860937576

if the implementation got confused, that can be resolved.  But if a status subelement exists, it *must* have a subresource for it.

deads2k:
Discussed in slack and going into past history found that this test was an unneeded side quest: https://github.com/openshift/origin/pull/26954#issuecomment-1122377912 .  Rather than exception, remove the entire test requiring a status stanza in the schema.

We must keep the test that ensures every CRD with a status stanza also has a status subresource and no new exceptions should appear there.

/hold cancel

everettraven (approved):
LGTM based on the discussion with @deads2k in slack: https://redhat-internal.slack.com/archives/CE4L0F143/p1743606075623449

/lgtm

PR #29644, opened 2025-04-02
NE-1970: e2e origin tests for gatewayApiController

rhamini3:
PR for Network Ingress and DNS team for the 5 e2e tests to graduate featureGate to GA

JIRA: https://issues.redhat.com/browse/NE-1970

Test Scenarios:

1. Confirm that OLM and OSSM resources are created 
2. Confirm that default gatewayClass is accepted
3. Confirm that a custom-gatewayclass can be created, deleted and not affect the state of the OSSM resources
4. A Custom gateway Object can be created
5. A HttpRoute can be created with no errors using a custom-gateway


test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,197 @@
...
+		e2e.Logf("GatewayClass %s successfully installed and accepted!", gwc.Name)
+
+	})
+
+	g.Describe("Verify Gateway API controller resources are created", func() {
+		g.It("and ensure OSSM related resources are created", func() {
+			coreClient := clientset.NewForConfigOrDie(oc.AdminConfig())
+			//check the subscription
+			g.By("Check OSSM subscription and CSV")
+			csvName, err := oc.AsAdmin().Run("get").Args("-n", expectedSubscriptionNamespace, "subscription", expectedSubscriptionName, "-o=jsonpath={.status.installedCSV}").Output()
```
candita, line 59: You may need to poll on getting the subscription, it can take some time.  Also we have a bug fix pending for issues with subscriptions: https://issues.redhat.com/browse/OCPBUGS-53424

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+		}
+		e2e.Logf("Found gatewayclass %s but it is not accepted, retrying...", name)
+		return false, nil
+	})
+
+	if waitErr != nil {
+		return fmt.Errorf("Gatewayclass %s is not accepted", name)
+	}
+	e2e.Logf("Gateway Class %s is created and accpeted", name)
```
candita, line 185: nit: 
```suggestion
	e2e.Logf("Gateway Class %s is created and accepted", name)
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,197 @@
...
+
+		gatewayClass := buildGatewayClass(gatewayClassName, gatewayClassControllerName)
+		gwc, err := gwapiClient.GatewayV1().GatewayClasses().Create(context.TODO(), gatewayClass, metav1.CreateOptions{})
+		if err != nil {
+			e2e.Logf("Gateway Class %s already exists, or has failed to be created, checking its status", gwc.Name)
+		}
+
+		errCheck := checkGatewayClass(oc, gatewayClassName)
+		o.Expect(errCheck).NotTo(o.HaveOccurred())
+		e2e.Logf("GatewayClass %s successfully installed and accepted!", gwc.Name)
```
candita, line 50: I mentioned an issue in the team-network-edge slack channel.  I think there's something external to this that is making the istiod and servicemesh pods fail to install, which would cause the GatewayClass to never be accepted.

test/extended/router/gatewayapicontroller.go
lihongan: The poll loop is added in `g.BeforeEach` `checkGatewayClass` func, if the gatewayclass is accepted in BeforeEach, then no need to poll on getting the subscription in the test itself.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,197 @@
...
+
+	})
+
+	g.Describe("Verify Gateway API controller resources are created", func() {
```
lihongan, line 54: I think `Verify Gateway API controller` is OK for g.Describe.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,197 @@
...
+		g.It("and ensure OSSM related resources are created", func() {
+			coreClient := clientset.NewForConfigOrDie(oc.AdminConfig())
+			//check the subscription
+			g.By("Check OSSM subscription and CSV")
+			csvName, err := oc.AsAdmin().Run("get").Args("-n", expectedSubscriptionNamespace, "subscription", expectedSubscriptionName, "-o=jsonpath={.status.installedCSV}").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			e2e.Logf("The subscription is installed and the CSV is: %v", csvName)
+
+			waitCSV := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 2*time.Minute, false, func(context context.Context) (bool, error) {
```
lihongan, line 63: Same for CSV, deployOSSM and deployIstio, we might not need to poll on getting their status in the test, if they are not ready the gatewayclass cannot reach `Accepted`.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		errCheck := checkGatewayClass(oc, gatewayClassName)
+		o.Expect(errCheck).NotTo(o.HaveOccurred())
+		e2e.Logf("GatewayClass %s successfully installed and accepted!", gwc.Name)
+
+	})
+
+	g.Describe("Verify Gateway API controller resources are created", func() {
+		g.It("and ensure OSSM related resources are created", func() {
```
lihongan, line 55: `and ensure OSSM related resources are created when default gatewayclass is accepted` may be better ?

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,197 @@
...
+		// The expected OSSM subscription name.
+		expectedSubscriptionName = "servicemeshoperator3"
+		// The expected OSSM operator namespace.
+		expectedSubscriptionNamespace = "openshift-operators"
+		// The gatewayclass name used to create ossm and other gateway api resources.
+		gatewayClassName = "openshift-default"
+		// gatewayClassControllerName is the name that must be used to create a supported gatewayClass.
+		gatewayClassControllerName = "openshift.io/gateway-controller/v1"
+	)
+	g.BeforeEach(func() {
```
lihongan, line 38: We need to check platformType and skip the tests if not on `AWSPlatformType, AzurePlatformType, GCPPlatformType`.

test/extended/router/gatewayapicontroller.go
rhamini3: @lihongan I had to poll the CSV because it was not showing the succeeded status instead it was showing `progressing` for subscription since we are extracting the CSV name, that is ok to not be polled.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,533 @@
...
+
+			g.By("Confirm that ISTIO CR is created and in healthy state")
+			resource := types.NamespacedName{Namespace: "openshift-ingress", Name: "openshift-gateway"}
+
+			istioStatus, errIstio := oc.AsAdmin().Run("get").Args("-n", resource.Namespace, "istio", resource.Name, "-o=jsonpath={.status.state}").Output()
+			o.Expect(errIstio).NotTo(o.HaveOccurred())
+			o.Expect(istioStatus).To(o.Equal(`Healthy`))
+		})
+
+		g.It("and ensure defualt gateway objects is created", func() {
```
candita, line 213: nit: 
```suggestion
		g.It("and ensure default gateway objects is created", func() {
```

test/extended/util/annotate/generated/zz_generated.annotations.go
```diff
@@ -1587,6 +1587,16 @@ var Annotations = map[string]string{
...
 
+	"[sig-network][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API controller and ensure HTTPRoute object is created": " [Suite:openshift/conformance/parallel]",
+
+	"[sig-network][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API controller and ensure OSSM and OLM related resources are created after creating GatewayClass": " [Suite:openshift/conformance/parallel]",
+
+	"[sig-network][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API controller and ensure custom gatewayclass can be accepted": " [Suite:openshift/conformance/parallel]",
+
+	"[sig-network][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API controller and ensure default gatewayclass is accepted": " [Suite:openshift/conformance/parallel]",
+
+	"[sig-network][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API controller and ensure defualt gateway objects is created": " [Suite:openshift/conformance/parallel]",
```
candita, line 1598:
```suggestion
	"[sig-network][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API controller and ensure default gateway objects is created": " [Suite:openshift/conformance/parallel]",
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,533 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayapiclientset "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
+)
+
+var _ = g.Describe("[sig-network][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io]", g.Ordered, func() {
```
candita, line 27: It helps alot if you specify these are network edge tests.
```suggestion
var _ = g.Describe("[sig-network-edge][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io]", g.Ordered, func() {
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,533 @@
...
+		// The gateway name used to create gateway resource.
+		gatewayName = "standard-gateway"
+		// The gateway name used to create custom gateway resource.
+		customGatewayName = "custom-gateway"
+	)
+	g.BeforeEach(func() {
+		// create the default gatewayClass
+		gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
+
+		gatewayClass := buildGatewayClass(gatewayClassName, gatewayClassControllerName)
```
candita, line 52: nit: comment is misplaced
```suggestion
		gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
		// create the default gatewayClass
		gatewayClass := buildGatewayClass(gatewayClassName, gatewayClassControllerName)
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,533 @@
...
+					return false, nil
+				}
+				e2e.Logf("The subscription is installed and the CSV is: %v", csvName)
+				return true, nil
+			})
+			if waitVersion != nil {
+				e2e.Failf("Subscription does not have an installed CSV")
+			}
+
+			csvName, err := oc.AsAdmin().Run("get").Args("-n", expectedSubscriptionNamespace, "subscription", expectedSubscriptionName, "-o=jsonpath={.status.installedCSV}").Output()
```
candita, line 93: This can now be removed, as it is already polled in a loop above.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,533 @@
...
+		gatewayClass := buildGatewayClass(gatewayClassName, gatewayClassControllerName)
+		gwc, err := gwapiClient.GatewayV1().GatewayClasses().Create(context.TODO(), gatewayClass, metav1.CreateOptions{})
+		if err != nil {
+			e2e.Logf("Gateway Class %s already exists, or has failed to be created", gwc.Name)
+		}
+	})
+
+	g.Describe("Verify Gateway API controller", func() {
+		g.It("and ensure OSSM and OLM related resources are created after creating GatewayClass", func() {
+			coreClient := clientset.NewForConfigOrDie(oc.AdminConfig())
```
candita, line 61: Do you have to create a new `client` every time in Gingko?

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,533 @@
...
+		})
+
+		g.It("and ensure HTTPRoute object is created", func() {
+			g.By("Getting the default domain for creating custom gateway")
+			defaultIngressDomain, err := getDefaultIngressClusterDomainName(oc, time.Minute)
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+			defaultDomain := strings.Trim(defaultIngressDomain, "apps.")
+
+			g.By("Create the custom API Gateway")
+			createGateway(oc, customGatewayName, gatewayClassName, defaultDomain)
```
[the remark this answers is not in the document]
rhamini3: this is @melvinjoseph86 test, i think he did this to remove dependency between the tests

test/extended/router/gatewayapicontroller.go
rhamini3: put it as a wrapper function in `client.go`

test/extended/router/gatewayapicontroller.go
rhamini3: I am using this variable `csvName` in the polling loop below, is there a way I can get the name from the loop above?

test/extended/util/client.go
```diff
@@ -674,6 +674,10 @@ func (c *CLI) BuildClient() buildv1client.Interface {
 	return buildv1client.NewForConfigOrDie(c.UserConfig())
 }
 
+func (c *CLI) CoreClient() kubernetes.Interface {
+	return kubernetes.NewForConfigOrDie(c.UserConfig())
+}
+
```
lihongan, line 680: See https://github.com/openshift/origin/blob/main/test/extended/util/client.go#L760-L763 there is already a wrapper for kubernetes, we needn't to create `CoreClient` here and just please use `oc.KubeClient` (regular user) in your tests.
And for Admin user please check https://github.com/openshift/origin/blob/main/test/extended/util/client.go#L769-L772

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+				e2e.Logf("Cluster Service Version %s has succeeded!", csvName)
+				return true, nil
+			})
+			if waitCSV != nil {
+				e2e.Failf("Cluster Service Version never got ready")
+			}
+
+			// get OLM Deployment Pod
+			waitErr := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 2*time.Minute, false, func(context context.Context) (bool, error) {
+				deployOSSM, err := oc.AsAdmin().CoreClient().AppsV1().Deployments(expectedSubscriptionNamespace).Get(context, "servicemesh-operator3", metav1.GetOptions{})
```
lihongan, line 113: Please check my comment in `test/extended/util/client.go`, the `CoreClient()` is not required and please use `oc.AdminKubeClient()` to replace all `oc.AsAdmin().CoreClient()`, Thanks

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
lihongan: The original idea is we could run these tests in parallel in any order, at the same time to improve the coverage we could use different gateway `listeners.allowedRoutes.namespaces` settings in different cases, e.g. one case use `from: All` (Shared gateway) and another one use `from: Same` (Dedicated gateway).
We could create follow up PR to enhance this later, and I'd like to merge this as early as possible if it is not a critical issue.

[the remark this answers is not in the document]
lihongan: I see you added `g.Ordered` for all tests, so yes we could use the old one (default gateway created in previous test). 
cc @melvinjoseph86

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+
+			// check OSSM deployment
+			waitIstio := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 2*time.Minute, false, func(context context.Context) (bool, error) {
+				deployIstio, err := oc.AsAdmin().CoreClient().AppsV1().Deployments("openshift-ingress").Get(context, "istiod-openshift-gateway", metav1.GetOptions{})
+				o.Expect(err).NotTo(o.HaveOccurred())
+				if deployIstio.Status.ReadyReplicas != 1 {
+					e2e.Logf("Deployment Pod %s is not ready, retrying...", "istiod-openshift-gateway")
+					return false, nil
+				}
+				e2e.Logf("Deployment Pod %s ready", "istiod-openshift-gateway")
```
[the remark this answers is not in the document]
alebedev87: As a matter of fact I'm wondering whether this is redundant too. Because `Healthy` status on `Istio` CR should mean that `istiod` is running and ready:
```go
    // IstioReasonHealthy indicates that the control plane is fully reconciled and that all components are ready.
    IstioReasonHealthy IstioConditionReason = "Healthy"
```
[Source link](https://github.com/openshift/cluster-ingress-operator/blob/1410fe02a76e33c00208fb3d9a9f00fa38cb527e/vendor/github.com/istio-ecosystem/sail-operator/api/v1/istio_types.go#L237-L240).

alebedev87 (commented):
Thanks for the PR, @rhamini3!

It's my first Look. I would like to clarify some questions first and then I'll have another look.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+				return true, nil
+			})
+			if waitIstio != nil {
+				e2e.Failf("Istio Deployment Pod never got ready")
+			}
+
+		})
+		g.It("and ensure default gatewayclass is accepted", func() {
+
+			g.By("Check if default GatewayClass is accepted after OLM resources are successful")
```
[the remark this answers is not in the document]
rhamini3: I believe so since we are using `g.Ordered` will have to check the logs to confirm

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+
+			g.By("Create Custom GatewayClass")
+			gatewayClass := buildGatewayClass("custom-gatewayclass", gatewayClassControllerName)
+			gwc, err := gwapiClient.GatewayV1().GatewayClasses().Create(context.TODO(), gatewayClass, metav1.CreateOptions{})
+			if err != nil {
+				e2e.Logf("Gateway Class %s already exists, or has failed to be created, checking its status", "custom-gatewayclass")
+			}
+			errCheck := checkGatewayClass(oc, "custom-gatewayclass")
+			o.Expect(errCheck).NotTo(o.HaveOccurred())
+			e2e.Logf("GatewayClass %s successfully installed and accepted!", gwc.Name)
```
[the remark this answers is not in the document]
rhamini3: The check is in place to see whether or not the gatewayClass was created or not. If it is then it checks if it is accepted or not. Since the CI runs the test twice, if the GC is created and not accepted on the first run, it can maybe be accepted on the second run

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+			if waitIstio != nil {
+				e2e.Failf("Istio Deployment Pod never got ready")
+			}
+
+			g.By("Confirm that ISTIO CR is created and in healthy state")
+			resource := types.NamespacedName{Namespace: "openshift-ingress", Name: "openshift-gateway"}
+
+			istioStatus, errIstio := oc.AsAdmin().Run("get").Args("-n", resource.Namespace, "istio", resource.Name, "-o=jsonpath={.status.state}").Output()
+			o.Expect(errIstio).NotTo(o.HaveOccurred())
+			o.Expect(istioStatus).To(o.Equal(`Healthy`))
```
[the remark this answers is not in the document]
rhamini3: yes, removed this check based on above comment

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+		})
+
+		g.It("and ensure default gateway objects is created", func() {
```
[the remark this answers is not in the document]
rhamini3: Based off the criteria for `g.Ordered` it should be

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+
+func checkGatewayClass(oc *exutil.CLI, name string) error {
+	gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
+
+	waitErr := wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 2*time.Minute, false, func(context context.Context) (bool, error) {
+		gwc, err := gwapiClient.GatewayV1().GatewayClasses().Get(context, name, metav1.GetOptions{})
+		if err != nil {
+			e2e.Logf("failed to get gatewayclass %s, retrying...", name)
+			return false, nil
+		}
```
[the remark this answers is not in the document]
rhamini3: this is using the dynamic client, for the OLM resources we are checking the specific YAML using `oc.Run` do you suggest that the if condition should follow the `if err != nil` for all the OLM checks too?

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
alebedev87: Ah, good point, I didn't see this setting in the top `Describe`. Nice!

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
alebedev87: Yes. Even if we use 2 different clients we do the similar thing - getting an object. Is there a reason the `oc.Run` should fail an iteration of a polling while dynamic client should not?

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,543 @@
...
+		// The gateway name used to create custom gateway resource.
+		customGatewayName = "custom-gateway"
+	)
+	g.BeforeEach(func() {
+		gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
+		// create the default gatewayClass
+		gatewayClass := buildGatewayClass(gatewayClassName, gatewayClassControllerName)
+		gwc, err := gwapiClient.GatewayV1().GatewayClasses().Create(context.TODO(), gatewayClass, metav1.CreateOptions{})
+		if err != nil {
+			e2e.Logf("Gateway Class %s already exists, or has failed to be created", gwc.Name)
```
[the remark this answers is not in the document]
alebedev87: Yes, `BeforeAll` seems to be the right choice[ for an ordered container](https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall).

test/extended/router/gatewayapicontroller.go
alebedev87: Yes, you just need a variable which is defined outside of the loop so that it can outlive the scope of the polling loop.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,502 @@
...
+			resource := types.NamespacedName{Namespace: "openshift-ingress", Name: "openshift-gateway"}
+
+			waitCR := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 2*time.Minute, false, func(context context.Context) (bool, error) {
+				istioStatus, errIstio := oc.AsAdmin().Run("get").Args("-n", resource.Namespace, "istio", resource.Name, "-o=jsonpath={.status.state}").Output()
+				o.Expect(errIstio).NotTo(o.HaveOccurred())
+				if istioStatus != "Healthy" {
+					e2e.Logf("Istio CR %s is not healthy, retrying...", resource.Name)
+					return false, nil
+				}
+				e2e.Logf("Istio CR %s is in Healthy state!", resource.Name)
```
alebedev87, line 126:
```suggestion
				e2e.Logf("Istio CR %s is healthy!", resource.Name)
```
Just to be consistent with the analogous `CatalogSource` message.

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
alebedev87: I think that we should do another assertion on the error returned by `Create()` because it doesn't make sense to continue if gateway class was not created.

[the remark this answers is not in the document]
rhamini3: what do you suggest? `gwapiClient.GatewayV1().GatewayClasses().Get(context, name, metav1.GetOptions{})` to just check whether or not it is there or not good enough?

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
rhamini3: no, I agree that both scenarios has the same result. I am just wondering if the oc.run command will result in a non-nil output if the status is not reported in the YAML?

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,502 @@
...
+		})
+
+		g.It("and ensure default gateway objects is created", func() {
+			g.By("Getting the default domain")
+			defaultIngressDomain, err := getDefaultIngressClusterDomainName(oc, time.Minute)
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+			defaultDomain := strings.Trim(defaultIngressDomain, "apps.")
+
+			g.By("Create the default API Gateway")
+			createGateway(oc, gatewayName, gatewayClassName, defaultDomain)
```
alebedev87, line 183:
```suggestion
			createAndCheckGateway(oc, gatewayName, gatewayClassName, defaultDomain)
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,502 @@
...
+func createGateway(oc *exutil.CLI, gwname, gwclassname, domain string) *gatewayapiv1.Gateway {
+	gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
+	ingressNameSpace := "openshift-ingress"
+	gateway := &gatewayapiv1.Gateway{}
+
+	// Get getway class details to create gateway
+	gatewayClass, errGwClass := gwapiClient.GatewayV1().GatewayClasses().Get(context.TODO(), gwclassname, metav1.GetOptions{})
+	if errGwClass != nil {
+		e2e.Failf("Expected gateway class object but not found, the error is %v", errGwClass)
+	}
```
alebedev87, line 272: Can we just pass the gatewayclass name to the function to avoid getting it from OpenShift again?

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,502 @@
...
+
+	// Build the gateway object
+	gatewaybuild := buildGateway(gwname, ingressNameSpace, gatewayClass.Name, "All", domain)
+
+	// Create the gateway object
+	_, errGwObj := gwapiClient.GatewayV1().Gateways(ingressNameSpace).Create(context.TODO(), gatewaybuild, metav1.CreateOptions{})
+	if errGwObj != nil {
+		e2e.Logf("Gateway already exists, or has failed to be created, checking its status")
+	}
```
alebedev87, line 281: Why not returning the error to fail the test? Because if the gateway is not created it doesn't make sense to continue.

alebedev87 (commented):
Second look. I will have to iterate again plus finish the httproute test.

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
lihongan: If we check the pre-submit jobs we would see below for flaky tests 
```
Run #0: Failed   39s
Run #1: Passed   39s
```
And would see below for failed tests
```
Run #0: Failed  	2m5s
Run #1: Failed  	2m5s
```
So if Run #0 is failed then it will try again, but not mean all the tests are run twice.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,504 @@
...
+			istioStatus, errIstio := oc.AsAdmin().Run("get").Args("-n", resource.Namespace, "istio", resource.Name, "-o=jsonpath={.status.state}").Output()
+			o.Expect(errIstio).NotTo(o.HaveOccurred())
+			o.Expect(istioStatus).To(o.Equal(`Healthy`))
+		})
+
+		g.It("and ensure default gateway objects is created", func() {
+			g.By("Getting the default domain")
+			defaultIngressDomain, err := getDefaultIngressClusterDomainName(oc, time.Minute)
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+			defaultDomain := strings.Trim(defaultIngressDomain, "apps.")
```
ShudiLi, line 179: The code `defaultDomain := strings.Trim(defaultIngressDomain, "apps.")`  will remove all character like "s" in defaultIngressDomain, suggest to update it to: `defaultDomain := strings.Split(defaultIngressDomain, "apps.")[1]`

melvinjoseph86: thanks

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,502 @@
...
+
+			g.By("Verify the gateway's LoadBalancer service and DNSRecords")
+			// get the dnsrecord name
+			dnsRecordName, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("-n", "openshift-ingress", "dnsrecord", "-l", "gateway.networking.k8s.io/gateway-name="+gatewayName, "-o=jsonpath={.items[*].metadata.name}").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			e2e.Logf("The gateway API dnsrecord name is: %v", dnsRecordName)
+			// check whether status of dns reccord is True
+			dnsRecordstatus, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("-n", "openshift-ingress", "dnsrecord", dnsRecordName, `-o=jsonpath={.status.zones[0].conditions[0].status}`).Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			o.Expect(dnsRecordstatus).To(o.Equal("True"))
```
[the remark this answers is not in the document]
melvinjoseph86: this  is to confirm the status condition of dns record of the gateway's type publish is true

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,502 @@
...
+
+			// check LB service
+			lbExternalIP, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("-n", "openshift-ingress", "dnsrecord", dnsRecordName, `-o=jsonpath={.spec.targets[0]}`).Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			e2e.Logf("The load balancer external IP is: %v", lbExternalIP)
+			gwwAddress, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("-n", "openshift-ingress", "gateway", gatewayName, "-o=jsonpath={.status.addresses[0].value}").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			e2e.Logf("The gateway Aaddress is: %v", gwwAddress)
+			o.Expect(lbExternalIP).To(o.Equal(gwwAddress))
```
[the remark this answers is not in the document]
melvinjoseph86: so what you suggest, i trying to cross check the LB address of the openshift ingress namespace is same as the address in the gateway's status. i am modifying the same like below 
```

			lbExternalIP, lberr := coreClient.CoreV1().Services("openshift-ingress").Get(context.Background(), gatewayName+"-openshift-default", metav1.GetOptions{})
			o.Expect(lberr).NotTo(o.HaveOccurred())
			e2e.Logf("The load balancer external IP is: %v", lbExternalIP.Status.LoadBalancer.Ingress[0].Hostname)

			gwwAddress, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("-n", "openshift-ingress", "gateway", gatewayName, "-o=jsonpath={.status.addresses[0].value}").Output()
			e2e.Logf("The gateway Aaddress is: %v", gwwAddress)
			o.Expect(lbExternalIP.Status.LoadBalancer.Ingress[0].Hostname).To(o.Equal(gwwAddress))
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,502 @@
...
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+			defaultDomain := strings.Trim(defaultIngressDomain, "apps.")
+
+			g.By("Create the default API Gateway")
+			createGateway(oc, gatewayName, gatewayClassName, defaultDomain)
+
+			g.By("Verify the gateway's LoadBalancer service and DNSRecords")
+			// get the dnsrecord name
+			dnsRecordName, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("-n", "openshift-ingress", "dnsrecord", "-l", "gateway.networking.k8s.io/gateway-name="+gatewayName, "-o=jsonpath={.items[*].metadata.name}").Output()
+			o.Expect(err).NotTo(o.HaveOccurred())
```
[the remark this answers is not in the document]
melvinjoseph86: done

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,504 @@
...
+		gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
+		// create the default gatewayClass
+		gatewayClass := buildGatewayClass(gatewayClassName, gatewayClassControllerName)
+		gwc, err := gwapiClient.GatewayV1().GatewayClasses().Create(context.TODO(), gatewayClass, metav1.CreateOptions{})
+		if err != nil {
+			e2e.Logf("Gateway Class %s already exists, or has failed to be created, checking if it exists", gwc.Name)
+		}
+		gwc, err = gwapiClient.GatewayV1().GatewayClasses().Get(context.Background(), gatewayClassName, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred(), "GatewayClass %q does not exist", gatewayClassName)
+
```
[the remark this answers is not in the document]
rhamini3: we are checking for it's accpeted status after all OLM/OSSM resources in this test: https://github.com/openshift/origin/blob/19eea93b54de4444c59206b3b01c275c50215d00/test/extended/router/gatewayapicontroller.go#L135

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,499 @@
...
+	})
+
+	if waitErr != nil {
+		e2e.Failf("The gateway is still not up and running and here is error %v", waitErr)
+		return nil, waitErr
+	} else {
+		e2e.Logf("Gateway %s successfully installed and accepted!", gateway.Name)
+		return gateway, nil
+	}
+
```
alebedev87, line 311:
```suggestion
	if waitErr != nil {
		return nil, fmt.Errorf("timed out waiting for gateway %q to become programmed: %w", gateway.Name, waitErr)
	}
	e2e.Logf("Gateway %q successfully programmed!", gateway.Name)
	return gateway, nil
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,499 @@
...
+	})
+	g.It("Ensure default gatewayclass is accepted", func() {
+
+		g.By("Check if default GatewayClass is accepted after OLM resources are successful")
+		errCheck := checkGatewayClass(oc, gatewayClassName)
+		o.Expect(errCheck).NotTo(o.HaveOccurred())
+		e2e.Logf("GatewayClass %s successfully installed and accepted!", gatewayClassName)
+
+	})
+	g.It("Ensure custom gatewayclass can be accepted", func() {
```
candita, line 142: This is not really a different test from creating the first gatewayclass.  And we don't need another gatewayclass.  It's better to just fatally exit if the default gatewayclass is not accepted.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,504 @@
...
+	g.BeforeAll(func() {
+		gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
+		// create the default gatewayClass
+		gatewayClass := buildGatewayClass(gatewayClassName, gatewayClassControllerName)
+		gwc, err := gwapiClient.GatewayV1().GatewayClasses().Create(context.TODO(), gatewayClass, metav1.CreateOptions{})
+		if err != nil {
+			e2e.Logf("Gateway Class %s already exists, or has failed to be created, checking if it exists", gwc.Name)
+		}
+		gwc, err = gwapiClient.GatewayV1().GatewayClasses().Get(context.Background(), gatewayClassName, metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred(), "GatewayClass %q does not exist", gatewayClassName)
```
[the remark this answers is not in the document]
rhamini3: @alebedev87 if the gatewayClass was previously created on the cluster, wouldn't this cause the test to fail?

test/extended/router/gatewayapicontroller.go
rhamini3: the main goal of this test is to confirm that two gatewayClasses can exist together and if one is removed, the rest of the OSSM resources don't get affected

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
rhamini3: ok makes sense, I guess it is not an issue when it is running on the OpenShift test suite, if we are to locally run it, we'd have to make sure the GatewayClass is deleted

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,491 @@
...
+		// The gatewayclass name used to create ossm and other gateway api resources.
+		gatewayClassName = "openshift-default"
+		// gatewayClassControllerName is the name that must be used to create a supported gatewayClass.
+		gatewayClassControllerName = "openshift.io/gateway-controller/v1"
+		//OSSM Deployment Pod Name
+		deploymentOSSMName = "servicemesh-operator3"
+		// The gateway name used to create gateway resource.
+		gatewayName = "standard-gateway"
+		// The gateway name used to create custom gateway resource.
+		customGatewayName = "custom-gateway"
```
candita, line 50: This variable is not used.  Did you mean to create a variable for the custom gateway class?

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,491 @@
...
+
+		g.By("Confirm that ISTIO CR is created and in healthy state")
+		resource := types.NamespacedName{Namespace: "openshift-ingress", Name: "openshift-gateway"}
+
+		istioStatus, errIstio := oc.AsAdmin().Run("get").Args("-n", resource.Namespace, "istio", resource.Name, "-o=jsonpath={.status.state}").Output()
+		o.Expect(errIstio).NotTo(o.HaveOccurred())
+		o.Expect(istioStatus).To(o.Equal(`Healthy`))
+	})
+
+	g.It("Ensure default gateway objects is created", func() {
```
candita, line 170:
```suggestion
	g.It("Ensure LB and service are created for a Gateway object", func() {
```

test/extended/router/gatewayapicontroller.go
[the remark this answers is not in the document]
candita: Part of the advantage of waiting in the BeforeAll step is that the rest of the tests will work properly.  For example, in this one it waits:  https://github.com/openshift/origin/blob/d2eafccc23b5b1b5def63412952e4650a8384ed3/test/extended/networking/ipsec.go#L475
and in this one it uses the o.Eventually construct to wait:  https://github.com/openshift/origin/blob/d2eafccc23b5b1b5def63412952e4650a8384ed3/test/extended/prometheus/collection_profiles.go#L132-L142

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,491 @@
...
+
+		g.By("Verify the gateway's LoadBalancer service and DNSRecords")
+		// check LB service
+		lbExternalIP, lberr := oc.AdminKubeClient().CoreV1().Services("openshift-ingress").Get(context.Background(), gatewayName+"-openshift-default", metav1.GetOptions{})
+		o.Expect(lberr).NotTo(o.HaveOccurred())
+		e2e.Logf("The load balancer external IP is: %v", lbExternalIP.Status.LoadBalancer.Ingress[0].Hostname)
+
+		gwwAddress, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("-n", "openshift-ingress", "gateway", gatewayName, "-o=jsonpath={.status.addresses[0].value}").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		e2e.Logf("The gateway Aaddress is: %v", gwwAddress)
```
candita, line 188:
```suggestion
		e2e.Logf("The gateway address is: %v", gwwAddress)
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,491 @@
...
+		})
+		o.Expect(waitCR).NotTo(o.HaveOccurred(), "Istio CR %q did not reach healthy state in time", resource.Name)
+
+	})
+	g.It("Ensure default gatewayclass is accepted", func() {
+
+		g.By("Check if default GatewayClass is accepted after OLM resources are successful")
+		errCheck := checkGatewayClass(oc, gatewayClassName)
+		o.Expect(errCheck).NotTo(o.HaveOccurred())
+		e2e.Logf("GatewayClass %s successfully installed and accepted!", gatewayClassName)
```
candita, line 135:
```suggestion
		o.Expect(errCheck).NotTo(o.HaveOccurred(), "GatewayClass %s was not installed and accepted", gatewayClassName)
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,491 @@
...
+			}
+		}
+		e2e.Logf("Found gatewayclass %s but it is not accepted, retrying...", name)
+		return false, nil
+	})
+
+	if waitErr != nil {
+		return fmt.Errorf("Gatewayclass %s is not accepted", name)
+	}
+	e2e.Logf("Gateway Class %s is created and accepted", name)
```
candita, line 243: Also, why not use the `o.Expect(err).NotTo(o.HaveOccurred()` in these functions?
```suggestion
	o.Expect(waitErr).NotTo(o.HaveOccurred, "Gatewayclass %s is not accepted", name)	
```

lihongan:
/payload-job periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview

test/extended/router/gatewayapicontroller.go
melvinjoseph86: o.Expect(waitErr).NotTo(o.HaveOccurred(), "Gatewayclass %s is not accepted", name)

test/extended/router/gatewayapicontroller.go
rhamini3: removing this

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,491 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	//clientset "k8s.io/client-go/kubernetes"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayapiclientset "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
+)
+
+var _ = g.Describe("[sig-network-edge][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io]", g.Ordered, func() {
```
Miciah, line 27: Please try adding `g.Serial`:
```suggestion
var _ = g.Describe("[sig-network-edge][OCPFeatureGate:GatewayAPIController][Feature:Router][apigroup:gateway.networking.k8s.io]", g.Ordered, g.Serial, func() {
```

rhamini3:
/close
PR is not updating

PR #29647, opened 2025-04-03
CNTRLPLANE-371: Use endpointslices in idling tests instead of deprecated endpoints

ardaguclu:
`endpoints` has been deprecated in 1.33 officially and we'll get deprecation warnings that causes failures in our test suite. This PR switches to `endpointslices` to fix the issue permanently in idling tests.


ardaguclu:
This PR is complementary with https://github.com/openshift/oc/pull/2001 but not directly depend on each other.

ardaguclu:
/cc @bertinatto 
could you PTAL?

danwinship:
> we'll get deprecation warnings that causes failures in our test suite

Ooh, I didn't realize we were that strict. Sorry for surprising you with these failures.

danwinship (approved):
This lgtm; you don't need to worry that there might be multiple EndpointSlices because it's a single-stack service with fewer then 100 endpoints.

PR #29658, opened 2025-04-08
NE-1970: all five e2e tests for gatewayApiController to origin

melvinjoseph86:
PR for the 5 e2e tests to graduate featureGate to GA

JIRA: https://issues.redhat.com/browse/NE-1970

Test Scenarios:

   1. Confirm that OLM and OSSM resources are created
   2. Confirm that default gatewayClass is accepted
   3. Confirm that a custom-gatewayclass can be created, deleted and not affect the state of the OSSM resources
   4. A Custom gateway Object can be created
   5. A HttpRoute can be created with no errors using a custom-gateway


melvinjoseph86:
/close
in favor of https://github.com/openshift/origin/pull/29703

PR #29671, opened 2025-04-10
WIP: add e2e tests for GatewayAPIController featuregate

lihongan:
Copied from https://github.com/openshift/origin/pull/29644 and made some changes to allow 5 tests run in parallel and fixed some failures/flakes.


test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,539 @@
...
+
+		g.By("Verify the gateway's LoadBalancer service and DNSRecords")
+		// check gateway LB service, note that External-IP might be hostname (AWS) or IP (Azure/GCP)
+		lbService, err := oc.AdminKubeClient().CoreV1().Services("openshift-ingress").Get(context.Background(), gatewayName+"-openshift-default", metav1.GetOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		if lbService.Status.LoadBalancer.Ingress[0].Hostname != "" {
+			lbAddress = lbService.Status.LoadBalancer.Ingress[0].Hostname
+		} else {
+			lbAddress = lbService.Status.LoadBalancer.Ingress[0].IP
+		}
```
lihongan, line 203: The old logic is failing on GCP/Azure because it always gets empty string of `.Hostname`.

test/extended/router/gatewayapicontroller.go
```diff
@@ -0,0 +1,539 @@
...
+	})
+
+	g.It("Ensure custom gatewayclass can be accepted", func() {
+		gwapiClient := gatewayapiclientset.NewForConfigOrDie(oc.AdminConfig())
+		customGatewayClassName := "gatewayclass-custom"
+
+		g.By("Ensure default GatewayClass is accepted and Istio is healthy")
+		errCheck := checkGatewayClass(oc, gatewayClassName)
+		o.Expect(errCheck).NotTo(o.HaveOccurred(), "GatewayClass %q was not installed and accepted", gatewayClassName)
+		waitForIstioHealthy(oc)
```
lihongan, line 153: If the test is the first one to be executed among the 5 tests, then we need to wait for Istio to be `Healthy` as well, otherwise it might be flaky in last checkpoint. Seems Istio status is not updated in time even gatewayclass status is accepted.

PR #29681, opened 2025-04-14
MCO-1634: Improve MCN test stability

isabella-janssen:
This skips the MCN tests that run on SNO on connection errors. This should decrease the failures we see in SNO that are due to issues connecting to the cluster.


isabella-janssen:
/close

Based on how the tests have been performing, it does not appear that this is necessary. Thus, I'm closing this PR.

PR #29684, opened 2025-04-15
 MCO-1615: Mco node degraded mcn condition

pablintino:
The MCN API has a condition to signal a failure durint the update of a MachineConfigNode. The condition was already in the API but the MCO was not handling it.
With the MCO now handling the condition our testing now needs to check it is properly set/unset when needed.


isabella-janssen:
LGTM too! Thanks @pablintino

PR #29694, opened 2025-04-20
OCPEDGE-1678: Fix failed etcd test on arbiter ci run

Neilhamza:


eggfoobar:
/lgtm

@dusk125 I'm not sure if you think we should rename the test from `master nodes` to `control plane nodes` to better capture this test, we'll just need to update the test name mapping to make sure we don't loose historical data if we go that route, what do you think?

dusk125:
@eggfoobar 
> I'm not sure if you think we should rename the test from master nodes to control plane nodes to better capture this test, we'll just need to update the test name mapping to make sure we don't loose historical data if we go that route, what do you think?

Should be fine, could do it in another PR though I think.

eggfoobar:
@dusk125 Sounds good, thanks! Then I think we're good with this PR if you want to approve it

PR #29706, opened 2025-04-23
(WIP) update mco-1599 happy scope test to change observed version

isabella-janssen:


isabella-janssen:
/close

These tests were fixed by properly implementing cleanup functionality in #29695 and have been stable since. Thus, this PR is no longer needed.

PR #29710, opened 2025-04-24
OCPBUGS-55238: spyglass: hide disruption events for localhost

vrutkovs:
Don't display localhost-related disruptions on spyglass. These are still displayed on non-spyglass reports in case unexpected localhost disruption happens

### DIFF
--- a/e2echart/non-spyglass-e2e-chart-template.html
+++ b/e2echart/non-spyglass-e2e-chart-template.html
@@ -417,7 +417,7 @@
         if (eventInterval.message.reason !== "DisruptionBegan" && eventInterval.message.reason !== "DisruptionSamplerOutageBegan") {
             return false
         }
-        if (eventInterval.source === "Disruption") {
+        if (eventInterval.source === "Disruption" || eventInterval.source === "DisruptionLocalhost") {
             return true
         }
         if (eventInterval.locator.keys["namespace"] === "e2e-k8s-service-lb-available") {
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -57940,7 +57940,7 @@ var _e2echartNonSpyglassE2eChartTemplateHtml = []byte(`<html lang="en">
         if (eventInterval.message.reason !== "DisruptionBegan" && eventInterval.message.reason !== "DisruptionSamplerOutageBegan") {
             return false
         }
-        if (eventInterval.source === "Disruption") {
+        if (eventInterval.source === "Disruption" || eventInterval.source === "DisruptionLocalhost") {
             return true
         }
         if (eventInterval.locator.keys["namespace"] === "e2e-k8s-service-lb-available") {


dgoodwin:
The problem with leaving expected disruption in and hiding it in the UI is the larger system used to monitor disruption data, all of which needs the same accommodations otherwise it flags localhost disruption as disruption and starts monitoring for changes. This would include the grafana dashboard, the alerts in dpcr cluster, and the metrics published by sippy for those alerts, as well as scheduled queries in bigquery used for the reporting.

Do you intend to have this monitored for changes in disruption and pursue fixes for those issues? 

If so then maybe we leave it in. (but we wouldn't to hide it on interval charts)

If not, these intervals really should be classified with a different source. That would immediately remove them from the analysis framework, and they would not appear in this chart. 

Also remember the new intervals UI under debug tools is at https://github.com/openshift/sippy/blob/main/sippy-ng/src/prow_job_runs/IntervalsChart.js and it is largely based on categorizing by Source.

vrutkovs:
>Do you intend to have this monitored for changes in disruption and pursue fixes for those issues?

Localhost disruptions are expected when pod restarts (on rollout), but may be misleading - in most cases they are expected to happen.

>If so then maybe we leave it in. (but we wouldn't to hide it on interval charts)

We're hiding them on the main chart, but leaving on non-spyglass charts for completeness.

>If not, these intervals really should be classified with a different source. That would immediately remove them from the analysis framework, and they would not appear in this chart.

I don't think these are being sent for analysis anyway

dgoodwin:
They have been spamming #trt-alerts for weeks now, up to and including today, they are definitely going into the analysis system. 

Can you skip generating the intervals when it's expected?

vrutkovs:
I think it's easier to move them to a different source

dgoodwin:
This looks great, thank you, just waiting to see the resulting files.

dgoodwin:
https://sippy.dptools.openshift.org/sippy-ng/job_runs/1947614946313375744/pull-ci-openshift-origin-main-e2e-gcp-ovn-upgrade/openshift_origin/29710/intervals?end=2025-07-22T13%3A43%3A07Z&filterText=&intervalFile=e2e-timelines_spyglass_20250722-123441.json&overrideDisplayFlag=0&selectedSources=OperatorAvailable&selectedSources=OperatorProgressing&selectedSources=OperatorDegraded&selectedSources=KubeletLog&selectedSources=EtcdLog&selectedSources=EtcdLeadership&selectedSources=Alert&selectedSources=Disruption&selectedSources=E2EFailed&selectedSources=APIServerGracefulShutdown&selectedSources=KubeEvent&selectedSources=NodeState&selectedSources=DisruptionLocalhost&start=2025-07-22T11%3A56%3A40Z

Looks good to me.

/lgtm
/hold

Release when you're happy with the results.

vrutkovs:
/hold cancel

Yup, looks good

PR #29718, opened 2025-04-25
(olmv1 tests) Increase timeout threshold for `verifyAPIEndpoints`

anik120:
30 seconds timeout is causing the tests to fail in the first run, but pass in the second run. Increasing the timeout to 2 minutes.


anik120:
Closing this in favor of #29724

PR #29724, opened 2025-04-25
(olmv1 tests) Wait for default catalogs to start serving

anik120:
Tests that use default catalogs are flaking because catalogs might not be ready before tests start. This PR introduces `waitForDefaultCatalogs` that is then used by all the tests to make sure catalogs are available and serving before performing any test.

### DIFF
--- a/test/extended/olm/olmv1.go
+++ b/test/extended/olm/olmv1.go
@@ -27,6 +27,7 @@ const (
 
 	typeInstalled   = "Installed"
 	typeProgressing = "Progressing"
+	typeServing     = "Serving"
 
 	reasonRetrying = "Retrying"
 )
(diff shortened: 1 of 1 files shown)


anik120:
Looking at the test `e2e-aws-ovn-fips`, good news is that only one test shows up as flaky with `context deadline reached` as the failed reason the first time (same issue this PR is trying to solve), instead of all tests:

```
[sig-olmv1][OCPFeatureGate:NewOLM][Skipped:Disconnected] OLMv1 openshift-community-operators Catalog should serve FBC via the /v1/api/all endpoint [Suite:openshift/conformance/parallel] 
```

which means the timeout deadline of 2 minutes looks like the sweet spot. If the test doesn't see default catalogs within 2 mins it's probably better to release the resources and try again when the catalogs are more likely to be ready. 

/test e2e-aws-ovn-fips

anik120:
All required tests passed, but the PR needs `acknowledge-critical-fixes-only` label to go in now. 

The flakiness isn't actually causing outright failures right now, so leaving this PR here for now till we're free to merge without that label again.

test/extended/olm/olmv1.go
```diff
@@ -714,3 +729,38 @@ spec:
...
+func waitForDefaultCatalogs(ctx context.Context, oc *exutil.CLI) {
+	var defaultCatalogs = []string{
+		"openshift-certified-operators",
+		"openshift-community-operators",
+		"openshift-redhat-marketplace",
+		"openshift-redhat-operators",
+	}
+	for _, cat := range defaultCatalogs {
+		g.By(fmt.Sprintf("checking that %q is exists and is serving content", cat))
+		err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 2*time.Minute, true, func(ctx context.Context) (bool, error) {
```
tmshort, line 742: It would be good to know how many iterations occur here, so that we can see if we're waiting on catalogs at all for these tests.

test/extended/olm/olmv1.go
```diff
@@ -714,3 +729,38 @@ spec:
...
+			output, err := oc.AsAdmin().WithoutNamespace().Run("get").Args("clustercatalogs.olm.operatorframework.io", cat, "-o=jsonpath={.status.conditions}").Output()
+			if err != nil {
+				g.GinkgoLogr.Info(fmt.Sprintf("error getting catalog: %v, retrying..", err))
```
tmshort, line 745: Are we sure we see these messages?

anik120:
Closing in favor of #29756

PR #29726, opened 2025-04-27
TRT-2091: Revert #29714 "OPRUN-3782: OLMv1: Add support for preflight permissions checks"

sosiouxme:
Reverts #29714 ; tracked by [TRT-2091](https://issues.redhat.com//browse/TRT-2091)

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are reverting this breaking change to get CI and/or nightly payloads flowing again.

Broke aws-ovn-techpreview beginning with 4.19.0-0.nightly-2025-04-26-092016

To unrevert this, revert this PR, and layer an additional separate commit on top that addresses the problem. Before merging the unrevert, please run these jobs on the PR and check the result of these jobs to confirm the fix has corrected the problem:

```
run nightly job e.g. /payload-job periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-techpreview
```

CC: @tmshort

<div align="right">
PR created by <a href="https://github.com/stbenjam/revertomatic">Revertomatic<sup>:tm:</sup></a>
</div>

PR #29730, opened 2025-04-28
OPRUN-3783: OLMv1: Add support for preflight permissions checks (fixed)

tmshort:
Adds several data files, of which -0 is the basic one that allows for installation, and -1 through -6 have minor removals to cause a failure. They are each combined with -base to attempt to install an operator.

This reapplies cf207152efc99fb5474650649bddd43cea89c71b with fixes due to upstream-to-downstream merge.

### DIFF
--- a/test/extended/olm/olmv1-preflight.go
+++ b/test/extended/olm/olmv1-preflight.go
@@ -1,0 +1,136 @@
+package operators
+
+import (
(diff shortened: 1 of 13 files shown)


tmshort:
This is a revert of #29726 with fixes.

PR #29731, opened 2025-04-28
TRT-2092: Fix default cert issuer name for Managed Service cluster (ROSA classic) in `RouteExternalCertificate` test case

chiragkyal:
This PR will decide the default cert issuer name for Managed Service clusters (ROSA classic). 

If the tests are running on Managed Service cluster, the default cert issuer name would be `R11`

See [Slack Thread](https://redhat-internal.slack.com/archives/C01CQA76KMX/p1745844051668559)

PR #29735, opened 2025-04-29
TRT-2092: Fix default cert issuer name in RouteExternalCertificate test case

chiragkyal:
The current logic for verifying the default certificate relies on checking the issuer CommonName against predefined values for different cluster profiles (Self-Managed, HyperShift, Managed Service). However, for managed service clusters (like ROSA), the default certificate issuer CN is dynamic (e.g. R10, R11, etc.), making it difficult to know the exact value beforehand.

This PR modifies the verification logic to handle the dynamic issuer CN. Instead of checking for a specific CN, the updated logic checks if the certificate is not issued by the external certificate issuer. This ensures that the route will not use the external certificate once removed.

Follow up: https://github.com/openshift/origin/pull/29731

### DIFF
--- a/test/extended/router/certgen/certgen.go
+++ b/test/extended/router/certgen/certgen.go
@@ -47,7 +47,7 @@ func MarshalCertToPEMString(derBytes []byte) (string, error) {
 // GenerateKeyPair creates root CA, certificate and key with optional
 // hosts. Certificate is valid from notBefore and expires after
 // notAfter.
-func GenerateKeyPair(notBefore, notAfter time.Time, hosts ...string) ([]byte, []byte, *ecdsa.PrivateKey, error) {
+func GenerateKeyPair(rootCertCN string, notBefore, notAfter time.Time, hosts ...string) ([]byte, []byte, *ecdsa.PrivateKey, error) {
 	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
 	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


test/extended/router/certgen/certgen.go
```diff
@@ -47,7 +47,7 @@ func MarshalCertToPEMString(derBytes []byte) (string, error) {
 // GenerateKeyPair creates root CA, certificate and key with optional
 // hosts. Certificate is valid from notBefore and expires after
 // notAfter.
-func GenerateKeyPair(notBefore, notAfter time.Time, hosts ...string) ([]byte, []byte, *ecdsa.PrivateKey, error) {
+func GenerateKeyPair(rootCertCN string, notBefore, notAfter time.Time, hosts ...string) ([]byte, []byte, *ecdsa.PrivateKey, error) {
```
chiragkyal, line 50: Taking the root certificate issuer CN as an argument

test/extended/router/external_certificate.go
```diff
@@ -764,7 +740,7 @@ func generateTLSCertSecret(namespace, secretName string, secretType corev1.Secre
 	notAfter := time.Now().Add(24 * time.Hour)
 
 	// Generate crt/key for secret
-	rootDerBytes, tlsCrtData, tlsPrivateKey, err := certgen.GenerateKeyPair(notBefore, notAfter, hosts...)
+	rootDerBytes, tlsCrtData, tlsPrivateKey, err := certgen.GenerateKeyPair(rootCertIssuerCN, notBefore, notAfter, hosts...)
```
chiragkyal, line 743: Creating certificate with defined Root CA Issuer CN

chiragkyal:
Note: Payload job does not work for ROSA jobs. So we cannot test `periodic-ci-openshift-release-master-nightly-4.19-e2e-rosa-sts-ovn` against this PR before the merge.

PR #29742, opened 2025-04-30
OKD-243: Skip GatewayAPIController tests for OKD

Prashanth684:
The OSSM operator is not available as a community operator for OKD.
Since OKD does not include the RedHat catalog, this feature cannot be
used.

### DIFF
--- a/test/extended/router/gatewayapicontroller.go
+++ b/test/extended/router/gatewayapicontroller.go
@@ -50,6 +50,14 @@ var _ = g.Describe("[sig-network-edge][OCPFeatureGate:GatewayAPIController][Feat
 		deploymentOSSMName = "servicemesh-operator3"
 	)
 	g.BeforeAll(func() {
+		isokd, err := isOKD(oc)
+		if err != nil {
+			e2e.Failf("Failed to get clusterversion to determine if release is OKD: %v", err)
+		}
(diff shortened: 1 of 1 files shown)


test/extended/router/gatewayapicontroller.go
```diff
@@ -539,3 +542,15 @@ func assertHttpRouteSuccessful(oc *exutil.CLI, gwName, name string) (*gatewayapi
 	e2e.Logf("httpRoute %s/%s successful", namespace, name)
 	return checkHttpRoute, nil
 }
+
+// Check for the existence of the okd-scos string in the version name to determine if it is OKD
+func isOKD(oc *exutil.CLI) bool {
+	current, err := exutil.GetCurrentVersion(context.TODO(), oc.AdminConfig())
+	if err != nil {
+		return false
+	}
```
Miciah, line 551: Would it be better to return the error and fail the test in case of an error?  I think it would be appropriate at least to log the error.

Prashanth684: done. fails in case fetching the clusterversion errors out.

Prashanth684:
@Miciah could you please approve - thanks.

Miciah:
Thanks!
/lgtm

PR #29748, opened 2025-05-01
NO-JIRA: add metal serial techpreview skips

neisw:


neisw:
/hold
may skip the skips and allow the failures

neisw:
/close
adding as candidate job for 4.19 instead

PR #29757, opened 2025-05-05
OCPBUGS-49441: Use different certificates for gRPC interoperability test routes

alebedev87:
Before this change the gRPC routes were using the same certificate which prevented the usage of ALPN to avoid the connection coalescing. This commit introduces different certificates for edge and reencrypt routes.


test/extended/router/grpc-interop.go
```diff
@@ -187,6 +187,15 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 			pemCrt, err := certgen.MarshalCertToPEMString(tlsCrtData)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
+			_, tlsCrt2Data, tlsPrivateKey2, err := certgen.GenerateKeyPair("Root CA", notBefore, notAfter)
```
Thealisyed, line 190: Why is a new Root CA being generated here? Is `tlsCrt2Data` intended to be a different root CA or a certificate signed by the first root CA?

test/extended/router/grpc-interop.go
```diff
@@ -253,8 +262,8 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 						TLS: &routev1.TLSConfig{
 							Termination:                   routev1.TLSTerminationReencrypt,
 							InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
-							Key:                           derKey,
-							Certificate:                   pemCrt,
+							Key:                           derKey2,
+							Certificate:                   pemCrt2,
```
Thealisyed, line 266: Why use those formats (`pem` and `der`) for the cert and key?

test/extended/router/grpc-interop.go
alebedev87: Why is a new Root CA being generated here?

That's how [`GenerateKeyPair`](https://github.com/openshift/origin/blob/6f06a24aa96bb85b3b0e979ed9b98880b42b5774/test/extended/router/certgen/certgen.go#L50) works. It generates [a self signed root CA certificate](https://github.com/openshift/origin/blob/6f06a24aa96bb85b3b0e979ed9b98880b42b5774/test/extended/router/certgen/certgen.go#L62-L79) and then uses it for [the test certificate](https://github.com/openshift/origin/blob/6f06a24aa96bb85b3b0e979ed9b98880b42b5774/test/extended/router/certgen/certgen.go#L113-L116) (which we get as second return value).

> Is tlsCrt2Data intended to be a different root CA or a certificate signed by the first root CA?

It's intended to be a different certificate. I think that it may still be possible to sign it with the same root CA though. But I stick to what `GenerateKeyPair` does.

Having different certificates for secure routes [instructs router](https://github.com/openshift/router/commit/0bc6f5ecf4c52169ca2a802cede164ac0a46f138) to enable `alpn` option in [`crt-list` command](https://docs.haproxy.org/2.8/configuration.html#5.1-crt-list) on HAProxy frontends (e.g. [fe_sni](https://github.com/openshift/router/blob/b41f9d05467fb7b3f6c2dafa6ac4b5e25164c0b6/images/router/haproxy/conf/haproxy-config.template#L334)). This results into the usage of ALPN TLS extension for the protocol negotiation. Without ALPN, `grpc` client will refuse to establish a connection to OpenShift router.

test/extended/router/grpc-interop.go
alebedev87: Those are the formats returned by `crypto/x509`'s [CreateCertificate function](https://pkg.go.dev/crypto/x509#CreateCertificate). Both can be used to encode certificates or keys. DER is binary format while PEM is base64. I didn't see a lot the usage of DER for route's keys but it seems to be accepted by the API and router validation. I prefer to be consistent across the router's tests and keep using DER (http2 test does use DER format too).

alebedev87 (commented):
@Thealisyed : Thanks for the review! Here are the answers, feel free to follow up in the comments.

test/extended/router/grpc-interop.go
alebedev87: [I followed up](https://github.com/openshift/origin/pull/29855) on this one. It turned out that `MarshalCertToDERFormat` was a misnomer.

Thealisyed:
Thanks for the answers and for the follow up PR for `MarshalCertToDERFormat` 

/lgtm

alebedev87:
Rebased from `main` to start the testing campaign from scratch.

PR #29760, opened 2025-05-06
NO-JIRA: support skipping [OCP]FeatureGate tests

dfajmon:
Enable skipping of tests named `OCPFeatureGate` and `FeatureGate`


jsafrane:
surprising amount of job runs failed to install the cluster, which should be unrelated to this PR (or is it?)
/retest

PR #29761, opened 2025-05-06
NO-JIRA: increase api client timeout to match those allowed by kube-apiserver

xueqzhan:


pkg/monitor/backenddisruption/disruption_backend_sampler.go
```diff
@@ -140,6 +140,8 @@ func NewAPIServerBackend(clientConfig *rest.Config, disruptionBackendName, path
 		return nil, err
 	}
 
+	// Kube-apiserver allows 34s timeout for api requests. We want to match it here.
+	timeout := 34 * time.Second
```
dgoodwin, line 144: Old code wanted some percentage of 20s (old total timeout) with a comment indicating it was to catch diff failure modes? Should we do the same percentage of 34s total here?

dgoodwin: Ah excellent, thanks, that explains why I wasn't seeing that comment where the change was made. :)

PR #29776, opened 2025-05-09
MCO-1652: Add MCO disruptive suite

yuqi-zhang:
Many MCO tests require node disruption, so it was determined that these would best live as a separate suite.

Create the MCO suite and move existing MCO tests in origin to it. The next goal is to add On Cluster Layering tests to this as well, potentially via OTE.

### DIFF
--- a/test/extended/machine_config/machine_config_node.go
+++ b/test/extended/machine_config/machine_config_node.go
@@ -29,7 +29,7 @@ const (
 	arbiter = "arbiter"
 )
 
-var _ = g.Describe("[sig-mco][OCPFeatureGate:MachineConfigNodes]", func() {
+var _ = g.Describe("[Suite:openshift/machine-config-operator/disruptive][sig-mco][OCPFeatureGate:MachineConfigNodes]", func() {
 	defer g.GinkgoRecover()
 	var (
 		MCOMachineConfigPoolBaseDir    = exutil.FixturePath("testdata", "machine_config", "machineconfigpool")
--- a/test/extended/machine_config/machine_config_node.go
+++ b/test/extended/machine_config/machine_config_node.go
@@ -52,7 +52,7 @@ var _ = g.Describe("[Suite:openshift/machine-config-operator/disruptive][sig-mco
 		}
 	})
 
-	g.It("Should have MCN properties matching associated node properties for nodes in default MCPs [apigroup:machineconfiguration.openshift.io]", func() {
+	g.It("[Suite:openshift/conformance/parallel]Should have MCN properties matching associated node properties for nodes in default MCPs [apigroup:machineconfiguration.openshift.io]", func() {
 		if IsSingleNode(oc) || IsTwoNode(oc) { //handle SNO & two-node clusters
 			// In SNO and standard two-node openshift clusters, the nodes have both worker and master roles, but are a part
 			// of the master MCP. Thus, the tests for these clusters will be limited to checking master MCP association.
@@ -67,13 +67,13 @@ var _ = g.Describe("[Suite:openshift/machine-config-operator/disruptive][sig-mco
 		}
 	})
 
-	g.It("[Serial]Should have MCN properties matching associated node properties for nodes in custom MCPs [apigroup:machineconfiguration.openshift.io]", func() {
+	g.It("[Suite:openshift/conformance/serial][Serial]Should have MCN properties matching associated node properties for nodes in custom MCPs [apigroup:machineconfiguration.openshift.io]", func() {
 		skipOnSingleNodeTopology(oc) //skip this test for SNO
 		skipOnTwoNodeTopology(oc)    //skip this test for two-node openshift
 		ValidateMCNPropertiesCustomMCP(oc, infraMCPFixture)
 	})
 
-	g.It("[Serial]Should properly transition through MCN conditions on rebootless node update [apigroup:machineconfiguration.openshift.io]", func() {
+	g.It("[Suite:openshift/conformance/serial][Serial]Should properly transition through MCN conditions on rebootless node update [apigroup:machineconfiguration.openshift.io]", func() {
 		if IsSingleNode(oc) {
 			ValidateMCNConditionTransitionsOnRebootlessUpdateSNO(oc, nodeDisruptionFixture, nodeDisruptionEmptyFixture, masterMCFixture)
 		} else {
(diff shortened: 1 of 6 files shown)


yuqi-zhang:
/hold

Not 100% decided on whether we'd like to move existing tests here yet, or we should just set this up for layering

test/extended/machine_config/machine_config_node.go
```diff
@@ -29,7 +29,7 @@ const (
 	arbiter = "arbiter"
 )
 
-var _ = g.Describe("[sig-mco][OCPFeatureGate:MachineConfigNodes]", func() {
+var _ = g.Describe("[Suite:openshift/machine-config-operator/disruptive][sig-mco][OCPFeatureGate:MachineConfigNodes]", func() {
```
yuqi-zhang, line 32: As far as we can tell, this would cause the test to only run in the new MCO suite, and not for serial runs anymore.

We're depending on this for signal, so while we're transitioning between the tests, would it be possible to run the test on both Serial and the new MCO suite? Would we have to define this test twice with different labels?

test/extended/machine_config/machine_config_node.go
```diff
@@ -67,29 +67,29 @@ var _ = g.Describe("[sig-mco][OCPFeatureGate:MachineConfigNodes]", func() {
...
+	g.It("[Suite:openshift/conformance/serial][Serial][Slow]Should properly report MCN conditions on node degrade [apigroup:machineconfiguration.openshift.io]", func() {
 		if IsSingleNode(oc) { //handle SNO clusters
 			ValidateMCNConditionOnNodeDegrade(oc, invalidMasterMCFixture, true)
 		} else { //handle standard, non-SNO, clusters
 			ValidateMCNConditionOnNodeDegrade(oc, invalidWorkerMCFixture, false)
 		}
 	})
 
-	g.It("[Serial][Slow]Should properly create and remove MCN on node creation and deletion [apigroup:machineconfiguration.openshift.io]", func() {
+	g.It("[Suite:openshift/conformance/serial][Serial][Slow]Should properly create and remove MCN on node creation and deletion [apigroup:machineconfiguration.openshift.io]", func() {
```
isabella-janssen, line 92: These two tests don't currently run in the default serial suite (or any default suite) since they are labeled with slow, as can be seen [here](https://github.com/isabella-janssen/openshift-origin/blob/main/test/extended/util/annotate/generated/zz_generated.annotations.go#L1382-L1384). So if it's possible to run some in the new suite and not others, these would be good candidates since they are not currently being used for our component readiness signal.

isabella-janssen: > Should I remove the serial suite label entirely then? Theoretically it shouldn't make a difference?

I don't think the serial suite label would do anything here, but I'm not 100% sure. We can maybe do a rehearsal payload job and see if this test shows up.

isabella-janssen:
/payload-job periodic-ci-openshift-release-master-ci-4.20-e2e-aws-ovn-techpreview-serial

test/extended/machine_config/machine_config_node.go
neisw: Have you tried building locally and using --dry-run to see what tests are included with which suites?

I feel like if you don't expect it to be included in `serial` suite due to the [slow] annotation (which I believe is correct) you shouldn't add [Suite:openshift/conformance/serial].  Adding extra annotation now will just make it harder to rename later if you go back and cleanup and want to matchup the old tests with the new names.

Also I'm wondering if you need both [Suite:openshift/conformance/serial] and [Serial] or [Serial] alone will do it.

It's an open question since you are adding `[Suite:openshift/machine-config-operator/disruptive]` but you can test locally with something like 

`./openshift-tests run "openshift/conformance/serial" -o "${ARTIFACT_DIR}/e2e.log" --junit-dir "${ARTIFACT_DIR/junit}" --dry-run`

and 
`./openshift-tests run "openshift/machine-config-operator/disruptive" -o "${ARTIFACT_DIR}/e2e.log" --junit-dir "${ARTIFACT_DIR/junit}" --dry-run`

to see what tests would run.   You can skip the --dry-run as well and actually run the tests if you like.  Unfortunately you do need a cluster even for the --dry-run

yuqi-zhang: >Also I'm wondering if you need both [Suite:openshift/conformance/serial] and [Serial] or [Serial] alone will do it.

I've tried with the existing tests via e2e-aws-ovn-serial-1of2/e2e-aws-ovn-serial-2of2 below, and [Serial] alone does not make the tests show up, whereas adding the full suite does (this is with the bootimage tests we have already, all the other tests are techpreview only so far)

And yes, let me remove the serial suite tag from the slow test.

yuqi-zhang:
/hold cancel

Let's try to add these tests for the MCO suite, it should be fine since they will still run via serial/parallel so we shouldn't lose any signal.

isabella-janssen:
/payload-job periodic-ci-openshift-release-master-ci-4.20-e2e-aws-ovn-techpreview-serial periodic-ci-openshift-release-master-ci-4.20-e2e-aws-ovn-techpreview 

Running these payloads to confirm that the same MCO tests continue running in the same test suites as before.

yuqi-zhang:
Looks like the list of tests is correct although the run didn't pass

isabella-janssen:
> Looks like the list of tests is correct although the run didn't pass

I am only seeing one of the MCN parallel tests in the payload rehearsal (only `Should have MCN properties matching associated node properties for nodes in default MCPs`). Can you try adding the `[Suite:openshift/conformance/parallel]` tag to the [three other MCN parallel tests](https://github.com/yuqi-zhang/origin/blob/c2bcd59717ecafb680506e83c20b8463ec7a92ed/test/extended/machine_config/machine_config_node.go#L97-L107), please @yuqi-zhang?

Otherwise, the test failure does not look related to this work, so hopefully that was just a bad run.

yuqi-zhang:
Ack, sorry, missed those for some reason, should be fixed now, thanks!

isabella-janssen:
/payload-job periodic-ci-openshift-release-master-ci-4.20-e2e-aws-ovn-techpreview-serial periodic-ci-openshift-release-master-ci-4.20-e2e-aws-ovn-techpreview

Running a rehearsal with the MCN parallel tests tagged.

isabella-janssen:
/lgtm

Looks good to me from the MCO side of things! All tests seem to be running in the same suites as previously.

PR #29793, opened 2025-05-15
OCPBUGS-55376: add wait for network ready before launching pods

jluhrsen:
if the network is not fully ready before pods are launched they may fail to come up on the first try due to missing annotations, etc. This could lead to excessive retries and if it happens just enough in CI this could trigger an alert and fail the job


jluhrsen:
/close
no longer needed

PR #29800, opened 2025-05-15
MCO-1652: Add new MCO disruptive suite

yuqi-zhang:
Add an empty MCO suite, for testing purposes. Future tests will be moved over from serial and added via MCO OTE.

Alternative to https://github.com/openshift/origin/pull/29776


isabella-janssen (commented):
lgtm too!

PR #29801, opened 2025-05-16
tagimportmode: Do not skip tests if the import mode is the same as the import mode of the cluster

Prashanth684:
Rather than skipping the test, just create an imagestream and verify the functionality so the tests are not skipped and provide signal against the feature gate.


Prashanth684:
@flavianmissi the gcp-ovn-techpreview-serial test results seem to indicate the tests passed successfully. Can you approve please?

PR #29807, opened 2025-05-16
OCPBUGS-56381: feat: update test to account for two node fencing

eggfoobar:
updated baremetal test to account for the difference in two node fencing baremetal deployment


jaypoulz:
Awesome! one less permafail on the board!

/lgtm

PR #29808, opened 2025-05-16
MULTIARCH-5391: Add new featuregate test to check if desired.Architecture is set as intended

Prashanth684:
A new featuregated field was introduced a while ago under clusterversion status - `desired.Architecture` which denotes the architecture of the release image being reconciled to. As we plan to promote the feature gate to default - this adds a basic test to check if the `desired.Architecture` is set based on the `release.openshift.io/architecture: multi` being present as part of the release metadata.

### DIFF
--- a/test/extended/clusterversion/OWNERS
+++ b/test/extended/clusterversion/OWNERS
@@ -1,0 +1,5 @@
+reviewers:
+  - cluster-version-operator-test-case-reviewers
+  - Prashanth684
+approvers:
+  - cluster-version-operator-test-case-approvers
--- a/test/extended/clusterversion/clusterversion.go
+++ b/test/extended/clusterversion/clusterversion.go
@@ -1,0 +1,43 @@
+package clusterversion
+
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+// The desired architecture field was introduced as part of a feature that provides a toggle for an imagestream's importMode: https://github.com/openshift/api/pull/2024
+var _ = g.Describe("[sig-cluster-lifecycle][OCPFeatureGate:ImageStreamImportMode] ClusterVersion API", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLIWithoutNamespace("")
+
+	g.It("desired architecture should be valid when architecture is set in release payload metadata [apigroup:config.openshift.io]", func() {
+		TestClusterVersionDesiredArchitecture(g.GinkgoT(), oc)
+	})
+
+})
+
+func TestClusterVersionDesiredArchitecture(t g.GinkgoTInterface, oc *exutil.CLI) {
+	ctx := context.Background()
+
+	archMetadata, _, err := oc.AsAdmin().Run("adm", "release", "info", `-ojsonpath={.metadata.metadata.release\.openshift\.io\/architecture}`).Outputs()
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	// Check desired.Architecture in the CV
+	configClient, err := configclient.NewForConfig(oc.AdminConfig())
+	o.Expect(err).NotTo(o.HaveOccurred())
+	clusterVersion, err := configClient.ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	if archMetadata == "multi" {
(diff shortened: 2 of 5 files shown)


test/extended/clusterversion/OWNERS
```diff
@@ -0,0 +1,5 @@
+reviewers:
+  - cluster-version-operator-test-case-reviewers
```
wking, line 2: Our existing consumer for this alias is [in `test/extended/util/openshift/clusterversionoperator`][1].  I'm not all that clear on how `test/extended` is organized, but it seems like we should keep these CVO-functionality tests collected in the same directory?

[1]: https://github.com/openshift/origin/tree/94a4d2a6f202f64f89a1b747b44484c358f8299d/test/extended/util/openshift/clusterversionoperator

test/extended/clusterversion/clusterversion.go
```diff
@@ -0,0 +1,46 @@
...
+	g.It("desired.architecture field in the CV should be multi or empty depending on the 'release.openshift.io/architecture' field being set in the release payload's metadata [apigroup:config.openshift.io]", func() {
+		TestClusterVersionDesiredArchitecture(g.GinkgoT(), oc)
+	})
+
+})
+
+func TestClusterVersionDesiredArchitecture(t g.GinkgoTInterface, oc *exutil.CLI) {
+	ctx := context.Background()
+	if !exutil.IsTechPreviewNoUpgrade(ctx, oc.AdminConfigClient()) {
+		g.Skip("skipping, this feature is only supported on TechPreviewNoUpgrade clusters")
```
wking, line 29: I don't think we want this guard.  The `OCPFeatureGate:ImageStreamImportMode` naming should already gate this test-case to clusters where `ImageStreamImportMode` is enabled.  That's _currently_ in tech-preview clusters, but the feature-gate will eventually GA, and we want the origin suite to gracefully understand that transition and begin running the test-case against those GA clusters, without us having to come back in and remove this skip then.

Prashanth684: makes sense. changed.

test/extended/clusterversion/OWNERS
Prashanth684: that file seems to be the utils file whereas [this](https://github.com/openshift/origin/blob/94a4d2a6f202f64f89a1b747b44484c358f8299d/test/extended/adminack/adminack.go#L25) is where the test is actually defined, listing cvo folks as [owners](https://github.com/openshift/origin/blob/94a4d2a6f202f64f89a1b747b44484c358f8299d/test/extended/adminack/OWNERS#L4). This is similar.

test/extended/clusterversion/clusterversion.go
```diff
@@ -0,0 +1,42 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
+var _ = g.Describe("[sig-cluster-lifecycle][OCPFeatureGate:ImageStreamImportMode] ClusterVersion API", func() {
```
dgoodwin, line 16: ImageStreamImportMode doesn't sound right for this, is it correct?

test/extended/clusterversion/clusterversion.go
```diff
@@ -0,0 +1,42 @@
...
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLIWithoutNamespace("")
+
+	g.It("desired.architecture field in the CV should be multi or empty depending on the 'release.openshift.io/architecture' field being set in the release payload's metadata [apigroup:config.openshift.io]", func() {
```
dgoodwin: CV is redundant with ClusterVersion API in the Describe above, and this is a very long test name, and the extra quotes probably not really necessary. You also don't want to get too specific as if anything changes such as a new value beyond multi/empty, a test rename requires a separate PR in the future, so best to plan for a test name that can be eternal.

How about ClusterVersion API desired architecture should be valid when architecture is set in release payload metadata [apigroup:config.openshift.io]

dgoodwin: or "should match release payload metadata"

test/extended/clusterversion/clusterversion.go
wking: [It's correct][1], although probably worth a commit-message comment to make the connection more clear?

[1]: https://github.com/openshift/api/blob/c8c30d5b3acc2ec3be9c2a6debc7d3109ce90076/config/v1/types_cluster_version.go#L739

Prashanth684: added a comment in the code as well as in the commit message.

test/extended/clusterversion/clusterversion.go
Prashanth684: makes sense. Changed to `desired architecture should be valid when architecture is set in release payload metadata`

test/extended/clusterversion/clusterversion.go
dgoodwin: Thanks!

PR #29819, opened 2025-05-19
tls/violations: add sjenning to OWNERS

sjenning:
cc @deads2k


vrutkovs:
/retitle NO-JIRA: tls/violations: add sjenning to OWNERS

PR #29816, opened 2025-05-19
NO-JIRA: ovnk, virt: Filter out ipv6 LLA

qinqon:

PR #29824, opened 2025-05-20
ovn, virt: Support proxy when downloading virtctl

qinqon:
The http transport used was not configuring the proxy mechanism so http proxy was deactivated, this change copy the transport from the DefaultTransport and disable TLS, so it gets the default environment env proxy mechanism.


kyrtapz:
/lgtm

@qinqon you need a valid jira here.

qinqon:
virt tests are propertly download virtctl since we see `virtctl console` output at logs, it's failing on status with ipv6 LLA address fixed by https://github.com/openshift/origin/pull/29816

PR #29833, opened 2025-05-21
OCPEDGE-1481: Initial tests for Two Nodes OCP with Fencing (TNF) cluster

clobrano:
Add initial topology tests
* Ensure correct number of ControlPlanes, and Arbiters
* Ensure correct number of static etcd pod containers
* Ensure correct number of podman etcd containers

Add initial behavior tests
* Ensure the cluster can handle a graceful node shutdown

Closes: [OCPEDGE-1481](https://issues.redhat.com//browse/OCPEDGE-1481), [OCPEDGE-1482](https://issues.redhat.com//browse/OCPEDGE-1482)

---

As a starting point for test integration within this new cluster
environment, this change enables only a minimal set of monitors. These
monitors are known to be reliable in general, but are currently
exhibiting unexpected behavior in this specific cluster.

This approach allows us to establish a foundational test base. Further
investigation into the reasons for their misbehavior in this new cluster
will be conducted once this initial test setup is merged.

### DIFF
--- a/test/extended/two_node/common.go
+++ b/test/extended/two_node/common.go
@@ -1,0 +1,26 @@
+package two_node
+
+import (
+	"fmt"
+
+	v1 "github.com/openshift/api/config/v1"
+	exutil "github.com/openshift/origin/test/extended/util"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+)
+
+const (
+	labelNodeRoleMaster       = "node-role.kubernetes.io/master"
+	labelNodeRoleControlPlane = "node-role.kubernetes.io/control-plane"
+	labelNodeRoleWorker       = "node-role.kubernetes.io/worker"
+	labelNodeRoleArbiter      = "node-role.kubernetes.io/arbiter"
+)
+
+func skipIfNotTopology(oc *exutil.CLI, wanted v1.TopologyMode) {
+	current, err := exutil.GetControlPlaneTopology(oc)
+	if err != nil {
+		e2eskipper.Skip(fmt.Sprintf("Could not get current topology, skipping test: error %v", err))
+	}
+	if *current != wanted {
+		e2eskipper.Skip(fmt.Sprintf("Cluster is not in %v topology, skipping test", wanted))
+	}
+}
--- a/test/extended/two_node/tnf_recovery.go
+++ b/test/extended/two_node/tnf_recovery.go
@@ -28,7 +28,7 @@ const (
 	pollInterval                 = 5 * time.Second
 )
 
-var _ = g.Describe("[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:DualReplica][Suite:openshift/two-node] Two Node with Fencing etcd recovery", func() {
+var _ = g.Describe("[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:DualReplica][Suite:openshift/two-node][Disruptive] Two Node with Fencing etcd recovery", func() {
 	defer g.GinkgoRecover()
 
 	var (
--- a/test/extended/two_node/tnf_recovery.go
+++ b/test/extended/two_node/tnf_recovery.go
@@ -1,0 +1,198 @@
+package two_node
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"net"
+	"slices"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	v1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/origin/test/extended/etcd/helpers"
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/pkg/errors"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:DualReplica] Two Node with Fencing etcd recovery", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc                = exutil.NewCLIWithoutNamespace("").AsAdmin()
+		etcdClientFactory *helpers.EtcdClientFactoryImpl
+		nodeA, nodeB      corev1.Node
+	)
+
+	g.BeforeEach(func() {
+		skipIfNotTopology(oc, v1.DualReplicaTopologyMode)
+
+		nodes, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		o.Expect(err).ShouldNot(o.HaveOccurred(), "Expected to retrieve nodes without error")
+		o.Expect(len(nodes.Items)).To(o.BeNumerically("==", 2), "Expected to find 2 Nodes only")
+
+		// Select the first index randomly
+		randomIndex := rand.Intn(len(nodes.Items))
+		nodeA = nodes.Items[randomIndex]
+		// Select the remaining index
+		nodeB = nodes.Items[(randomIndex+1)%len(nodes.Items)]
+		g.GinkgoT().Printf("Randomly selected %s (%s) to be gracefully shut down and %s (%s) to take the lead\n", nodeB.Name, nodeB.Status.Addresses[0].Address, nodeA.Name, nodeA.Status.Addresses[0].Address)
+
+		kubeClient := oc.KubeClient()
+		etcdClientFactory = helpers.NewEtcdClientFactory(kubeClient)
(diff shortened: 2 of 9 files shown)


clobrano:
Temporarily converting it to draft to investigate a crash

clobrano:
Ready again for review

test/extended/tnf/recovery.go
```diff
@@ -0,0 +1,199 @@
...
+	v1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/origin/test/extended/etcd/helpers"
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/pkg/errors"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("[sig-node][apigroup:config.openshift.io] Two Nodes OCP with fencing recovery", func() {
```
eggfoobar, line 21: Can you please add the annotation `[OCPFeatureGate:DualReplica]` to the test names to allow the feature gate to be captured

clobrano: Sure :+1:

eggfoobar (commented):
Looking good :), just had some small suggestion around the helpers.

test/extended/two_node/common.go
```diff
@@ -0,0 +1,41 @@
...
+	infraStatus, err := getInfraStatus(oc)
+	if err != nil {
+		e2eskipper.Skip(fmt.Sprintf("Could not get Topology mode, skipping test: error %v", err))
+	}
+	if infraStatus.ControlPlaneTopology != topology {
+		e2eskipper.Skip(fmt.Sprintf("Cluster is not in %v topology, skipping test", topology))
+	}
+}
+
+func getInfraStatus(oc *exutil.CLI) (*v1.InfrastructureStatus, error) {
```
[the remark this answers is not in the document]
clobrano: Good point, I'll replace this and the one below

test/extended/two_node/common.go
```diff
@@ -0,0 +1,26 @@
...
+	labelNodeRoleMaster       = "node-role.kubernetes.io/master"
+	labelNodeRoleControlPlane = "node-role.kubernetes.io/control-plane"
+	labelNodeRoleWorker       = "node-role.kubernetes.io/worker"
+	labelNodeRoleArbiter      = "node-role.kubernetes.io/arbiter"
+)
+
+func skipIfNotTopology(oc *exutil.CLI, wanted v1.TopologyMode) {
+	current, err := exutil.GetControlPlaneTopology(oc)
+	if err != nil {
+		e2eskipper.Skip(fmt.Sprintf("Could not get current topology, skipping test: error %v", err))
```
jaypoulz, line 21: This may be a little strange, but I think we should default to running the test when we don't know the topology. The reason is - this will likely lead to the tests running and failing, which will help us identify misconfigured clusters.

We want to avoid failing silently.

test/extended/two_node/tnf_recovery.go
```diff
@@ -0,0 +1,196 @@
...
+	v1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/origin/test/extended/etcd/helpers"
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/pkg/errors"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = g.Describe("[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:DualReplica] Two Node with Fencing etcd recovery", func() {
```
jaypoulz: I think there is an existing convention for naming "disruptive" or tests that should be run in serial. I'm not sure graceful recovery qualifies as that latter, but it definitely qualifies as the former since API requests may be routed to the dead node depend on how the load-balancer handles the rebooting node.

test/extended/two_node/tnf_recovery.go
```diff
@@ -0,0 +1,196 @@
...
+		etcdClientFactory *helpers.EtcdClientFactoryImpl
+		nodeA, nodeB      corev1.Node
+	)
+
+	g.BeforeEach(func() {
+		skipIfNotTopology(oc, v1.DualReplicaTopologyMode)
+
+		nodes, err := oc.AdminKubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		o.Expect(err).ShouldNot(o.HaveOccurred(), "Expected to retrieve nodes without error")
+		o.Expect(len(nodes.Items)).To(o.BeNumerically("==", 2), "Expected to find 2 Nodes only")
```
jaypoulz, line 45: Should we add a filter for control-plane nodes? I'm imagining a future where we're asked to support/test compute nodes.

jaypoulz: It might also be good to verify that there are 0 arbiter nodes.

test/extended/two_node/tnf_recovery.go
```diff
@@ -0,0 +1,196 @@
...
+				return fmt.Errorf("Not enough members")
+			}
+
+			if started, learner, err := getMemberState(&nodeA, members); err != nil {
+				return err
+			} else if !started || learner {
+				return fmt.Errorf("Expected node: %s to be a started and voting member. Membership: %+v", nodeA.Name, members)
+			}
+
+			// Ensure GNS node is unstarted and a learner member (i.e. !learner)
```
jaypoulz: I am confused by this comment.
Wouldn't `!learner` mean that it's not a learner?

test/extended/two_node/tnf_recovery.go
```diff
@@ -0,0 +1,196 @@
...
+			if started, learner, err := getMemberState(&nodeB, members); err != nil {
+				return err
+			} else if started || !learner {
+				return fmt.Errorf("Expected node: %s to be a unstarted and learning member. Membership: %+v", nodeB.Name, members)
+			}
+
+			g.GinkgoT().Logf("membership: %+v", members)
+			return nil
+		}, 2*time.Minute, 15*time.Second).ShouldNot(o.HaveOccurred())
```
jaypoulz: We may want to pull these timing values out to a shared place in the file to keep them consistent with our non-graceful shutdown test - or even just to quickly be able to adjust timeouts and check frequency across the test suite.

test/extended/two_node/tnf_recovery.go
jaypoulz: Looking below, we have another test for that. :) So I think just verifying 2 control-plane nodes should be sufficient.

test/extended/two_node/tnf_topology.go
```diff
@@ -0,0 +1,117 @@
...
+	defer g.GinkgoRecover()
+	var (
+		oc = exutil.NewCLIWithoutNamespace("")
+	)
+
+	g.BeforeEach(func() {
+		skipIfNotTopology(oc, v1.DualReplicaTopologyMode)
+	})
+
+	g.It("Should validate the number of control-planes, workers and arbiters as configured", func() {
```
jaypoulz: Only concern here is the potential for compute nodes introduced down the line. I would keep the check to 2 control-plane nodes, and omit the general node-count check.

jaypoulz (commented):
I really love how this is coming together. My biggest concern is how we need to properly label the tests that have the potential to affect other tests.

In other words - the node reboot/restart tests need to have some kind of label to indicate that they should be run serially and/or are disruptive.
The existing logic for that is https://github.com/openshift/origin/blob/9508b94995c110e3a6c988dbf35a8920bc36c3ed/pkg/testsuites/standard_suites.go#L59-L92.

I'm not familiar enough with how the final test list is composed to know if it's sane to tag the graceful shutdown test as part of the serial suite.

test/extended/two_node/common.go
clobrano: The point is that under the same suite we have both Arbiter and TNF, which are mutually exclusive, so a failure won't tell us much about misconfigured cluster, isn't it?

test/extended/two_node/tnf_recovery.go
clobrano: You are right, I was confused by the following `if` branch we test for "!learner", but to catch the error :)

test/extended/two_node/common.go
jaypoulz: Originally, I was thinking we would run these tests as part of the e2e suite. But since we've organized them as a separate suite that we only run when we've explicitly set the topology, I agree. :)

jaypoulz (approved):
/lgtm

jaypoulz:
/retitle [OCPEDGE-1481](https://issues.redhat.com//browse/OCPEDGE-1481): Initial tests for Two Nodes OCP with Fencing (TNF) cluster

PR #29842, opened 2025-05-23
[WIP] DRA e2e test to validate Nvidia gpu

tkashem:


e2e-monitor-tests__20250915-155137.xml
```diff
@@ -0,0 +1,10 @@
+<testsuite name="openshift-tests" tests="7" skipped="0" failures="0" time="0">
```
kannon92: This looks like it should not be committed.

events_used_for_junits_20250915-155437.json
```diff
@@ -0,0 +1,332 @@
+{
```
kannon92: same here. https://github.com/openshift/origin/pull/29842/files#r2349993368

test-failures-summary_monitor_20250915-155137.json
```diff
@@ -0,0 +1,23 @@
+{
```
kannon92: Is this needed?

tkashem: mistakenly added, i fixed it.

events_used_for_junits_20250915-155437.json
tkashem: removed

e2e-monitor-tests__20250915-155137.xml
tkashem: yes, removed

test/extended/dra/common_spec.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"testing"
+
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	helper "github.com/openshift/origin/test/extended/dra/helper"
+
+	corev1 "k8s.io/api/core/v1"
+	resourceapi "k8s.io/api/resource/v1beta1"
```
kannon92: https://github.com/kubernetes-sigs/kueue/pull/5873#discussion_r2354833594

I still think we should use v1beta2 rather than v1beta1.

I was told that DRA community broke the API in v1beta and they are roundtripping v1beta2 to v1.

harche: /hold

tkashem: the test suite (client) can use v1beta1, or v1beta2 as long as the api version is being served. also, if we want this suite to run against 4.19 cluster v1beta1 is the only option

kannon92: /hold cancel

PR #29844, opened 2025-05-26
NO-JIRA: OVNK BGP: set VRF-Lite timeout

jcaamano:
The VRF-Lite test is composed of many serial checks which makes it long. But it is also taking longer due to OCPBUGS-56488.


jcaamano:
/retitle NO-JIRA: OVNK BGP: set VRF-Lite timeout

PR #29853, opened 2025-05-27
NO-JIRA: OVNK BGP: workaround OCPBUGS-56488

jcaamano:
Due to a bug in the liveness probe of kubernetes-nmstate reported in OCPBUGS-56488, we need, as a workaround, to scale down the operator and disable the probe.

### DIFF
--- a/test/extended/networking/route_advertisements.go
+++ b/test/extended/networking/route_advertisements.go
@@ -23,6 +23,7 @@ import (
 	meta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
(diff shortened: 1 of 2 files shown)


test/extended/networking/route_advertisements.go
```diff
@@ -488,6 +489,10 @@ var _ = g.Describe("[sig-network][OCPFeatureGate:RouteAdvertisements][Feature:Ro
 					g.By("deploy nmstate handler")
 					err = deployNmstateHandler(oc)
 					o.Expect(err).NotTo(o.HaveOccurred())
+
+					// This is a workaround for OCPBUGS-56488: scale down
+					// nmstate operator and disable liveness probe
+					o.Eventually(workaroundOCPBUGS56488).WithArguments(oc).WithTimeout(3 * time.Minute).WithPolling(5 * time.Second).Should(o.BeTrue())
```
qinqon, line 495: Can this end up running in paralel ? Or we have just one test at this BeforeEach ?

test/extended/networking/route_advertisements.go
```diff
@@ -1505,3 +1510,17 @@ func gatherDebugInfo(oc *exutil.CLI, snifferDaemonset *v1.DaemonSet, targetNames
 		}
 	}
 }
+
+func workaroundOCPBUGS56488(oc *exutil.CLI) (bool, error) {
+	opPatch := []byte(`{"spec":{"replicas": 0}}`)
+	_, err := oc.AdminKubeClient().AppsV1().Deployments(nmstateNamespace).Patch(context.Background(), "nmstate-operator", types.MergePatchType, opPatch, metav1.PatchOptions{})
+	if err != nil {
+		return false, err
+	}
```
qinqon, line 1519: maybe we need to wait nmstate-operator pods to not be there after scale them down ?

qinqon (changes requested):

test/extended/networking/route_advertisements.go
jcaamano: its only one test, but it wouldn't be an issue if it ran in parallel

PR #29870, opened 2025-05-29
MCO-1722: Handle check for Disconnnected Clusters for PIS Testing using curl

RishabhSaini:
helpers: Handle Disconnnected Clusters using curl
pinnedimages: In metal disconnected case, check for existence of images prior to MCN status

### DIFF
--- a/test/extended/machine_config/helpers.go
+++ b/test/extended/machine_config/helpers.go
@@ -128,8 +128,8 @@ func IsTwoNodeArbiter(oc *exutil.CLI) bool {
 
 // `IsDisconnected` returns true if the cluster is a Disconnected cluster and false otherwise
 func IsDisconnected(oc *exutil.CLI, nodeName string) bool {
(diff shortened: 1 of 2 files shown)


test/extended/machine_config/helpers.go
```diff
@@ -128,8 +128,8 @@ func IsTwoNodeArbiter(oc *exutil.CLI) bool {
 
 // `IsDisconnected` returns true if the cluster is a Disconnected cluster and false otherwise
 func IsDisconnected(oc *exutil.CLI, nodeName string) bool {
-	networkStatus, _ := exutil.DebugNodeRetryWithOptionsAndChroot(oc, nodeName, "openshift-machine-config-operator", "systemctl", "is-active", "NetworkManager-wait-online.service")
-	if networkStatus == "active" {
+	networkStatus, _ := exutil.DebugNodeRetryWithOptionsAndChroot(oc, nodeName, "openshift-machine-config-operator", "sh", "-c", "curl -s --connect-timeout 5 https://www.google.com &>/dev/null && echo \"Connected\" || echo \"Disconnected\"")
```
pablintino, line 131: I'm not a big fan of using such type of URL, specially given that we are RH, maybe `http://fedoraproject.org/static/hotspot.txt` is a better fit (this is the URL Fedora uses for its internal check)

PR #29885, opened 2025-06-03
OCPBUGS-54927: Skip test causing issues on single node

dusk125:


dusk125:
Test in question, e2e-aws-ovn-single-node-techpreview-serial, looks happy

test/extended/images/tagimportmode.go
```diff
@@ -24,6 +24,14 @@ var _ = g.Describe("[sig-imageregistry][OCPFeatureGate:ImageStreamImportMode][Se
 	defer g.GinkgoRecover()
 	oc := exutil.NewCLI("imagestream-api")
 
+	g.BeforeEach(func() {
+		isSingleNode, err := exutil.IsSingleNode(context.Background(), oc.AdminConfigClient())
+		o.Expect(err).ToNot(o.HaveOccurred())
+		if isSingleNode {
+			g.Skip("the feature being tested is not relevant for single node")
```
flavianmissi, line 31: After reading the bug comments I find this message slightly misleading. Isn't the reason we're skipping this test is that a different test was added which then causes a rollout of the openshift-apiserver, leaving the o-apiserver degraded on single-node cluster, which then causes this test to fail ([comment](https://issues.redhat.com/browse/OCPBUGS-54927?focusedId=27060856&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27060856))?
It's okay to go with this approach, as it does solve the problem in the short term. I also think it's worth exploring a different way to solve this, as I'm not sure the [original test](https://github.com/openshift/origin/pull/29524/files) was written with the knowledge that it would cause the o-apiserver to become degraded for extended periods of time.
CCing @Prashanth684 as he wrote the original test.

Prashanth684:
The `TestImageConfigImageStreamImportModeLegacy` and `TestImageConfigImageStreamImportModePreserveOriginal` already skip these tests and should not run on SNO already. Not sure why this PR has been raised here. The other test doesn't disrupt apiserver and can be enabled

Prashanth684:
https://github.com/openshift/origin/pull/29715 already made it so the tests don't run on SNO

dusk125:
Closing in favor of https://github.com/openshift/origin/pull/29715

test/extended/images/tagimportmode.go
Prashanth684: There are three tests in this suite. Two of those toggle the importmode in the image config, which would trigger a rollout of the openshiftapiserver as the apiserver's observed config would change.

PR #29640, opened 2025-04-02
OCPNODE-2849: Add testcase to verify removal of cgroupv1 in OCP 4.19

asahay19:
**PR Description:** 
This PR adds a test to verify that modifying the CgroupMode field of the nodes.config object to v1 fails as expected since OpenShift 4.19 does not support cgroupv1.
The test follows the structure of existing  node tests and utilizes a debug pod to verify the failure scenario.

Tested locally , test case has executed successfully. 

PTAL @sairameshv

PR #29886, opened 2025-06-03
Revert "OCPNODE-2849: Add testcase to verify removal of cgroupv1 in OCP 4.19"

bertinatto:
Reverts openshift/origin#29640.

This broke `make update`, and when https://github.com/openshift/origin/pull/29883 attempted to fix it, it revealed that the test introduced is also broken. I suggest we revert this change and reintroduce it later with a proper fix.


sairameshv (approved):
/lgtm

PR #29887, opened 2025-06-04
TRT-1854: set network for MicroShift to OVNKubernetes

smg247:
MicroShift has no `--network` flag set when running OTE tests which is causing it to miss a skip that it should have.

There is some [special configuration](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/microshift/e2e/origin-conformance/openshift-microshift-e2e-origin-conformance-commands.sh#L46-L71) that sets it to `OVNKubernetes` for tests. I confirmed that this is a safe assumption to make in [this thread](https://redhat-internal.slack.com/archives/C03DP9PABNC/p1748958779190589). If, in the future, they start testing it with other network plugins we will have to be smarter about differentiating them.


smg247:
It passed the first time it ran here

PR #29895, opened 2025-06-05
WRKLDS-1716: test/extended/cli: Test oc create token

tchap:
Originally I was planning to remove tests for `oc serviceaccounts` as I wanted to work on removing that group of deprecated subcommands, but I ended up postponing that. I still think we can merge the test change that extends the current tests to also check `oc create token` and remove the associated `TODO` notes. `oc serviceaccounts` calls can be removed later and the test added here kept in place.

### DIFF
--- a/test/extended/util/client.go
+++ b/test/extended/util/client.go
@@ -309,6 +309,32 @@ func (c CLI) WithToken(token string) *CLI {
 	return &c
 }
 
+// WithKubeConfigCopy copies the current kubeconfig into a temporary file and sets the copy as the current kubeconfig
+// for the duration of the invocation of fnc. The temporary file is removed once fnc returns.
+func (c CLI) WithKubeConfigCopy(fnc func(*CLI)) {
+	out, err := os.CreateTemp("", "kubeconfig")
+	if err != nil {
+		FatalErr(err)
+	}
+	defer os.Remove(out.Name())
+
+	in, err := os.Open(c.configPath)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


ardaguclu:
> oc create token is now supposed to be used, which is just forwarding to
kubectl create token, so there is nothing to test on our side.

Conceptually you are right. But there is no test that exercises this upstream functionality against OCP cluster. Yes, we know that this works on vanilla Kubernetes. But we also have to know that oc create token works on OCP. So it is better to update our test to at least simply run create token.

test/extended/util/client.go
```diff
@@ -309,6 +309,32 @@ func (c CLI) WithToken(token string) *CLI {
...
+	}
+
+	_, err = io.Copy(out, in)
+	in.Close()
+	out.Close()
+	if err != nil {
+		FatalErr(err)
+	}
+
+	c.configPath = out.Name()
```
ardaguclu, line 334: Whenever `configPath` is modified, what will happen to the other tests that are running in parallel relying on `configPath`?. If that creates a flaky behavior, that would be a head ache. 

But I understand the intention in here, can we overcome this situation without updating the `configPath`?

tchap: The config path is only updated for the copy since the function has `(c CLI)` receiver. Other function are not affected. This is how many other utility functions in the file work.

ardaguclu:
/lgtm
from workloads team

PR #29540, opened 2025-02-13
TRT-2006: Revert "NO-JIRA: Add test case for checking EgressFirewall DNS names in caps"

bertinatto:
Reverts openshift/origin#29414

Just checking if this is causing issues on Azure.

PR #29898, opened 2025-06-06
OCPBUGS-50701: Add test case for checking EgressFirewall DNS names in caps

arkadeepsen:
This PR adds test case for checking EgressFirewall DNS names in caps.

This is a follow-up of https://github.com/openshift/origin/pull/29540.

The issue faced here is similar to https://issues.redhat.com/browse/OCPBUGS-52358 which is fixed by https://github.com/openshift/origin/pull/29618.

In this PR `amazon.com` is used which has a large TTL value (~700s).


arkadeepsen:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.20-upgrade-from-stable-4.19-e2e-azure-ovn-upgrade 10

PR #29901, opened 2025-06-09
OVNK BGP: use a config map to allocate EIPs and UDN subnets

jcaamano:
to avoid collisions on parallel running tests


jcaamano:
Everything looks good. As a side note, EgressIP tests are perma-failing in `ci/prow/e2e-aws-ovn-serial-publicnet-1of2`, not an issue with this PR.
/cc @pliurh PTAL

jcaamano:
/retitle NO-JIRA: OVNK BGP: use a config map to allocate EIPs and UDN subnets

test/extended/networking/route_advertisements.go
```diff
@@ -447,11 +466,11 @@ var _ = g.Describe("[sig-network][OCPFeatureGate:RouteAdvertisements][Feature:Ro
 
 			afterEach := func() {
 				g.GinkgoHelper()
-
-				runOcWithRetry(oc.AsAdmin(), "delete", "deploy", deployName)
-				runOcWithRetry(oc.AsAdmin(), "delete", "pod", "--all")
-				runOcWithRetry(oc.AsAdmin().WithoutNamespace(), "delete", "ra", testCUDNName)
-				runOcWithRetry(oc.AsAdmin().WithoutNamespace(), "delete", "clusteruserdefinednetwork", testCUDNName)
+				o.Expect(runOcWithRetryIgnoreOutput(oc.AsAdmin(), "delete", "deploy", deployName)).To(o.Succeed())
```
pliurh: These commands are only for clean-up. We shall ensure that all of them are executed. And I don't think we should fail a test when a clean-up command returns an error.

test/extended/networking/util.go
```diff
@@ -997,10 +995,176 @@ func hasNetworkConfigWriteAccess(oc *exutil.CLI) (bool, error) {
...
+			if cm.Data[key] != "" {
+				err := json.Unmarshal([]byte(cm.Data[key]), &existing)
+				if err != nil {
+					return err
+				}
+			}
+			existingSet := sets.New(existing...)
+			ipsv4, err = allocate(t4, n4, min, max, existingSet)
+			if err != nil {
+				return fmt.Errorf("could not allocate %d IPv4 from pool %s with existin allocations %v: %w", n4, key, existing, err)
```
pliurh: existing allocations

test/extended/networking/util.go
```diff
@@ -997,10 +995,176 @@ func hasNetworkConfigWriteAccess(oc *exutil.CLI) (bool, error) {
...
+			}
+			ipsv6, err = allocate(t6, n6, min, max, existingSet)
+			if err != nil {
+				return fmt.Errorf("could not allocate %d IPv6 from pool %s with existin allocations %v: %w", n6, key, existing, err)
```
pliurh: dito

test/extended/networking/util.go
```diff
@@ -997,10 +995,176 @@ func hasNetworkConfigWriteAccess(oc *exutil.CLI) (bool, error) {
...
+func deallocateIPs(oc *exutil.CLI, namespace, name, key string, ips ...string) error {
+	return retry.RetryOnConflict(
+		wait.Backoff{Steps: 30, Duration: time.Second},
+		func() error {
+			cm, err := oc.AdminKubeClient().CoreV1().ConfigMaps(allocationConfigMapNamespace).Get(context.Background(), allocationConfigMapName, metav1.GetOptions{})
+			if err != nil {
+				return fmt.Errorf("failed to get config map %q: %w", allocationConfigMapName, err)
+			}
+			existing := []string{}
+			if cm.Data[key] != "" {
```
pliurh: If the `key` doesn't exist, we don't need to update the CM. We can return from here.

test/extended/networking/util.go
```diff
@@ -997,10 +995,176 @@ func hasNetworkConfigWriteAccess(oc *exutil.CLI) (bool, error) {
...
+		// If an error was found, either return the error, or retry if a timeout error was found.
+		if err != nil {
+			if strings.Contains(strings.ToLower(err.Error()), "i/o timeout") {
+				// Retry on "i/o timeout" errors
+				e2e.Logf("Warning: oc command encountered i/o timeout.\nerr=%v\n)", err)
+				continue
+			}
+			return stdout, err
+		}
+
```
pliurh, line 1167: Shall we add a backoff before the next retry?

test/extended/networking/util.go
```diff
@@ -997,10 +995,176 @@ func hasNetworkConfigWriteAccess(oc *exutil.CLI) (bool, error) {
...
+		// callers, ignore it
+		stdout, _, err = oc.Run(cmd).Args(args...).Outputs()
+		// If an error was found, either return the error, or retry if a timeout error was found.
+		if err != nil {
+			if strings.Contains(strings.ToLower(err.Error()), "i/o timeout") {
+				// Retry on "i/o timeout" errors
+				e2e.Logf("Warning: oc command encountered i/o timeout.\nerr=%v\n)", err)
+				continue
+			}
+			return stdout, err
```
pliurh, line 1165: Can we also log the error message for debugging?

test/extended/networking/route_advertisements.go
jcaamano: I think we should definitely fail the test case if cleanup fails, otherwise we will never know.

I didn't think that ensuring all of them are executed was important enough, as none of them should fail in the first place or if they do, then we should have something to fix. But I will try to execute them all, even if one of them fails.

test/extended/networking/util.go
jcaamano: Maybe, but it should be a different PR. I only moved this method verbatim to a better place.

test/extended/networking/util.go
jcaamano: I don't think its appropriate. The caller gets the error so the caller should be responsible to log. Otherwise it will all be prone to double logging.

jcaamano:
@pliurh PTAL

sdodson:
/override ci/prow/4.12-upgrade-from-stable-4.11-e2e-aws-ovn-upgrade-rollback

PR #29914, opened 2025-06-12
OCPBUGS-57397: properties should wrap around property

pruan-rht:
`openshift-tests` is generating invalid unit xml like 
```
<testsuite name="openshift-tests" tests="14" skipped="0" failures="1" time="240">
<property name="TestVersion" value="4.19.0-202506020913.p0.g33fe530.assembly.stream.el9-33fe530"/>
```


pruan-rht:
addresses JIRA https://issues.redhat.com/browse/OCPBUGS-57397

pruan-rht:
@stbenjam I think the failures are actual test issues and not related to the fix, can we bypass the failures and merge this PR?

PR #29917, opened 2025-06-12
CNTRLPLANE-945: Add tests for ExternalOIDC and ExternalOIDCWithUIDAndExtraClaimMappings features

everettraven:
This PR adds the tests necessary to promote the ExternalOIDC and ExternalOIDCWithUIDAndExtraClaimMappings feature-gates on OpenShift.

It adds a new test suite specific to tests that test the external OIDC provider authentication mode.
Some tests are intentionally marked as skipped as the functionality does not yet exist to test, but I wanted to still provide the skeleton for these tests so that we can easily make some updates when that functionality _is_ implemented.

### DIFF
--- a/pkg/testsuites/standard_suites.go
+++ b/pkg/testsuites/standard_suites.go
@@ -430,6 +430,16 @@ var staticSuites = []ginkgo.TestSuite{
 		},
 		TestTimeout: 60 * time.Minute,
 	},
+	{
+		Name: "openshift/auth/external-oidc",
+		Description: templates.LongDesc(`
+		This test suite runs tests to validate cluster behavior when cluster authentication is configured to use an external OIDC provider.
+		`),
+		Qualifiers: []string{
+			`name.contains("[Suite:openshift/auth/external-oidc") && !name.contains("[Skipped]")`,
+		},
+		TestTimeout: 120 * time.Minute,
+	},
 }
 
 func withExcludedTestsFilter(baseExpr string) string {
--- a/test/extended/authentication/keycloak_client.go
+++ b/test/extended/authentication/keycloak_client.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"crypto/tls"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -43,11 +42,15 @@ func keycloakClientFor(keycloakURL string) (*keycloakClient, error) {
 	}, nil
 }
 
+type group struct {
+	Name string `json:"name"`
+}
+
 func (kc *keycloakClient) CreateGroup(name string) error {
 	groupURL := kc.adminURL.JoinPath("groups")
 
-	group := map[string]interface{}{
-		"name": name,
+	group := group{
+		Name: name,
 	}
 
 	groupBytes, err := json.Marshal(group)
@@ -69,20 +72,41 @@ func (kc *keycloakClient) CreateGroup(name string) error {
 	return nil
 }
 
+type user struct {
+	Username      string       `json:"username"`
+	Email         string       `json:"email"`
+	Enabled       bool         `json:"enabled"`
+	EmailVerified bool         `json:"emailVerified"`
+	Groups        []string     `json:"groups"`
+	Credentials   []credential `json:"credentials"`
+}
+
+type credential struct {
+	Temporary bool           `json:"temporary"`
+	Type      credentialType `json:"type"`
+	Value     string         `json:"value"`
+}
+
+type credentialType string
+
+const (
+	credentialTypePassword credentialType = "password"
+)
+
 func (kc *keycloakClient) CreateUser(username, password string, groups ...string) error {
 	userURL := kc.adminURL.JoinPath("users")
 
-	user := map[string]interface{}{
-		"username":      username,
-		"email":         fmt.Sprintf("%s@payload.openshift.io", username),
-		"enabled":       true,
-		"emailVerified": true,
-		"groups":        groups,
-		"credentials": []map[string]interface{}{
+	user := user{
+		Username:      username,
+		Email:         fmt.Sprintf("%s@payload.openshift.io", username),
+		Enabled:       true,
+		EmailVerified: true,
+		Groups:        groups,
+		Credentials: []credential{
 			{
-				"temporary": false,
-				"type":      "password",
-				"value":     password,
+				Temporary: true,
+				Type:      credentialTypePassword,
+				Value:     password,
 			},
 		},
 	}
@@ -106,14 +130,18 @@ func (kc *keycloakClient) CreateUser(username, password string, groups ...string
 	return nil
 }
 
+type authenticationResponse struct {
+	AccessToken string `json:"access_token"`
+	IDToken     string `json:"id_token"`
+}
+
 func (kc *keycloakClient) Authenticate(clientID, username, password string) error {
-	data := url.Values{
-		"username":   []string{username},
-		"password":   []string{password},
-		"grant_type": []string{"password"},
-		"client_id":  []string{clientID},
-		"scope":      []string{"openid"},
-	}
+	data := url.Values{}
+	data.Set("username", username)
+	data.Set("password", password)
+	data.Set("grant_type", "password")
+	data.Set("client_id", clientID)
+	data.Set("scope", "openid")
 
 	tokenURL := *kc.adminURL
 	tokenURL.Path = fmt.Sprintf("/realms/%s/protocol/openid-connect/token", kc.realm)
@@ -124,38 +152,15 @@ func (kc *keycloakClient) Authenticate(clientID, username, password string) erro
 	}
 	defer resp.Body.Close()
 
-	respBody := map[string]interface{}{}
-	respBodyData, err := io.ReadAll(resp.Body)
+	respBody := &authenticationResponse{}
+
+	err = json.NewDecoder(resp.Body).Decode(respBody)
 	if err != nil {
-		return fmt.Errorf("reading response data: %w", err)
+		return fmt.Errorf("unmarshalling response data: %w", err)
 	}
 
-	err = json.Unmarshal(respBodyData, &respBody)
-	if err != nil {
-		return fmt.Errorf("unmarshalling response body %s: %w", respBodyData, err)
-	}
-
-	accessTokenData, ok := respBody["access_token"]
-	if !ok {
-		return errors.New("unable to extract access token from the response body: access_token field is missing")
-	}
-
-	accessToken, ok := accessTokenData.(string)
-	if !ok {
-		return fmt.Errorf("expected accessToken to be of type string but was %T", accessTokenData)
-	}
-	kc.accessToken = accessToken
-
-	idTokenData, ok := respBody["id_token"]
-	if !ok {
-		return errors.New("unable to extract id token from the response body: id_token field is missing")
-	}
-
-	idToken, ok := idTokenData.(string)
-	if !ok {
-		return fmt.Errorf("expected idToken to be of type string but was %T", idTokenData)
-	}
-	kc.idToken = idToken
+	kc.accessToken = respBody.AccessToken
+	kc.idToken = respBody.IDToken
 
 	return nil
 }
@@ -191,41 +196,66 @@ func (kc *keycloakClient) ConfigureClient(clientId string) error {
 		return fmt.Errorf("getting client %q: %w", clientId, err)
 	}
 
-	id, ok := client["id"]
-	if !ok {
-		return fmt.Errorf("client %q doesn't have 'id'", clientId)
-	}
-
-	idStr, ok := id.(string)
-	if !ok {
-		return fmt.Errorf("client %q 'id' is not of type string: %T", clientId, id)
-	}
-
-	if err := kc.CreateClientGroupMapper(idStr, "test-groups-mapper", "groups"); err != nil {
+	if err := kc.CreateClientGroupMapper(client.ID, "test-groups-mapper", "groups"); err != nil {
 		return fmt.Errorf("creating group mapper for client %q: %w", clientId, err)
 	}
 
-	if err := kc.CreateClientAudienceMapper(idStr, "test-aud-mapper"); err != nil {
+	if err := kc.CreateClientAudienceMapper(client.ID, "test-aud-mapper"); err != nil {
 		return fmt.Errorf("creating audience mapper for client %q: %w", clientId, err)
 	}
 
 	return nil
 }
 
+type groupMapper struct {
+	Name           string            `json:"name"`
+	Protocol       protocol          `json:"protocol"`
+	ProtocolMapper protocolMapper    `json:"protocolMapper"`
+	Config         groupMapperConfig `json:"config"`
+}
+
+type protocol string
+
+const (
+	protocolOpenIDConnect protocol = "openid-connect"
+)
+
+type protocolMapper string
+
+const (
+	protocolMapperOpenIDConnectGroupMembership protocolMapper = "oidc-group-membership-mapper"
+	protocolMapperOpenIDConnectAudience        protocolMapper = "oidc-audience-mapper"
+)
+
+type groupMapperConfig struct {
+	FullPath           booleanString `json:"full.path"`
+	IDTokenClaim       booleanString `json:"id.token.claim"`
+	AccessTokenClaim   booleanString `json:"access.token.claim"`
+	UserInfoTokenClaim booleanString `json:"userinfo.token.claim"`
+	ClaimName          string        `json:"claim.name"`
+}
+
+type booleanString string
+
+const (
+	booleanStringTrue  booleanString = "true"
+	booleanStringFalse booleanString = "false"
+)
+
 func (kc *keycloakClient) CreateClientGroupMapper(clientId, name, claim string) error {
 	mappersURL := *kc.adminURL
 	mappersURL.Path += fmt.Sprintf("/clients/%s/protocol-mappers/models", clientId)
 
-	mapper := map[string]interface{}{
-		"name":           name,
-		"protocol":       "openid-connect",
-		"protocolMapper": "oidc-group-membership-mapper", // protocol-mapper type provided by Keycloak
-		"config": map[string]string{
-			"full.path":            "false",
-			"id.token.claim":       "true",
-			"access.token.claim":   "true",
-			"userinfo.token.claim": "true",
-			"claim.name":           claim,
+	mapper := &groupMapper{
+		Name:           name,
+		Protocol:       protocolOpenIDConnect,
+		ProtocolMapper: protocolMapperOpenIDConnectGroupMembership,
+		Config: groupMapperConfig{
+			FullPath:           booleanStringFalse,
+			IDTokenClaim:       booleanStringTrue,
+			AccessTokenClaim:   booleanStringTrue,
+			UserInfoTokenClaim: booleanStringTrue,
+			ClaimName:          claim,
 		},
 	}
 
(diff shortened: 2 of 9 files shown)


test/extended/util/client.go
```diff
@@ -853,6 +853,17 @@ func (c *CLI) NewPrometheusClient(ctx context.Context) prometheusv1.API {
 }
 
 func (c *CLI) UserConfig() *rest.Config {
+	if c.token != "" {
```
everettraven, line 863: Note for reviewers - this change was added because whenever a token was set here, getting the user config would attempt to pull the config from an empty config path. 

This resulted in a consistent error so I updated this to handle the case where a token is explicitly set.

pkg/testsuites/standard_suites.go
```diff
@@ -440,6 +440,16 @@ var staticSuites = []ginkgo.TestSuite{
 		},
 		TestTimeout: 60 * time.Minute,
 	},
+	{
+		Name: "openshift/auth/external-oidc",
+		Description: templates.LongDesc(`
+		This test suite runs tests to validate cluster behavior when cluster authentication is configured to use an external OIDC provider.
+		`),
+		Qualifiers: []string{
+			`name.contains("[Suite:openshift/auth/external-oidc") && !name.contains("[Skipped]")`,
```
liouk, line 439:
```suggestion
			`name.contains("[Suite:openshift/auth/external-oidc]") && !name.contains("[Skipped]")`,
```

test/extended/authentication/oidc.go
```diff
@@ -0,0 +1,580 @@
...
+
+				observedConfig := map[string]interface{}{}
+				err = json.Unmarshal(kas.Spec.ObservedConfig.Raw, &observedConfig)
+				o.Expect(err).NotTo(o.HaveOccurred(), "should not encounter an error unmarshalling the KAS observed configuration")
+
+				apiServerArgs := observedConfig["apiServerArguments"].(map[string]interface{})
+
+				o.Expect(apiServerArgs["authentication-token-webhook-config-file"]).To(o.BeNil(), "authentication-token-webhook-config-file argument should not be specified when OIDC authentication is configured")
+				o.Expect(apiServerArgs["authentication-token-webhook-version"]).To(o.BeNil(), "authentication-token-webhook-version argument should not be specified when OIDC authentication is configured")
+				o.Expect(apiServerArgs["authConfig"]).To(o.BeNil(), "authConfig argument should not be specified when OIDC authentication is configured")
```
liouk: `authConfig` is not under `apiServerArguments`, it's on the same level.

test/extended/authentication/oidc.go
```diff
@@ -0,0 +1,580 @@
...
+				// should always be able to create an SSAR for yourself
+				o.Eventually(func(gomega o.Gomega) {
+					// always re-authenticate to get a new token
+					err := keycloakCli.Authenticate("admin-cli", username, password)
+					gomega.Expect(err).NotTo(o.HaveOccurred(), "should not encounter an error authenticating as keycloak user")
+
+					copiedOC := *oc
+					tokenOC := copiedOC.WithToken(keycloakCli.AccessToken())
+
+					_, err = tokenOC.KubeClient().AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), &authzv1.SelfSubjectAccessReview{
```
liouk: nit: Why not kill two birds with one stone with the next SSR test? Proves both token acceptance and cluster identity mapping. We could still split it into two tests that validate these two things.

Also, why not also use an SSR to prove the previous oauth token test as well?

test/extended/authentication/oidc.go
```diff
@@ -0,0 +1,580 @@
...
+				// Wait until we can authenticate using the configured external IdP
+				o.Eventually(func(gomega o.Gomega) {
+					// always re-authenticate to get a new token
+					err := keycloakCli.Authenticate("admin-cli", username, password)
+					gomega.Expect(err).NotTo(o.HaveOccurred(), "should not encounter an error authenticating as keycloak user")
+
+					copiedOC := *oc
+					tokenOC := copiedOC.WithToken(keycloakCli.AccessToken())
+
+					_, err = tokenOC.KubeClient().AuthorizationV1().SelfSubjectAccessReviews().Create(context.TODO(), &authzv1.SelfSubjectAccessReview{
```
liouk: nit: In the spirit of previous comments, maybe an SSR is sufficient?

test/extended/authentication/oidc.go
```diff
@@ -0,0 +1,580 @@
...
+
+				observedConfig := map[string]interface{}{}
+				err = json.Unmarshal(kas.Spec.ObservedConfig.Raw, &observedConfig)
+				o.Expect(err).NotTo(o.HaveOccurred(), "should not encounter an error unmarshalling the KAS observed configuration")
+
+				apiServerArgs := observedConfig["apiServerArguments"].(map[string]interface{})
+
+				o.Expect(apiServerArgs["authentication-token-webhook-config-file"]).NotTo(o.BeNil(), "authentication-token-webhook-config-file argument should be specified when OIDC authentication is not configured")
+				o.Expect(apiServerArgs["authentication-token-webhook-version"]).NotTo(o.BeNil(), "authentication-token-webhook-version argument should be specified when OIDC authentication is not configured")
+				o.Expect(apiServerArgs["authConfig"]).NotTo(o.BeNil(), "authConfig argument should be specified when OIDC authentication is not configured")
```
liouk: Same as before, `authConfig` isn't part of `apiServerArguments`.

liouk (changes requested):
We could move the `keycloak_*.go` files into their own package; it'll make it clearer to reuse functionality in other tests if ever needed.

Looks great in general, just a few comments. I won't block the PR on the nits :)

pkg/testsuites/standard_suites.go
everettraven: I had intentionally left off the end bracket because it seems to be the pattern other test suites follow. 

Presumably this is so you can run sub-suites of this as part of this suite (i.e if I did something like `[Suite:openshift/auth/external-oidc/some-sub-thing]` the test with this "tag" would still run as part of the `openshift/auth/external-oidc` test suite.

If we think this is unnecessary and that we should deviate from existing convention, I can add the closing bracket.

test/extended/authentication/oidc.go
everettraven: Good catch. I think what I meant here is `authentication-config` - will update

test/extended/authentication/oidc.go
everettraven: > nit: Why not kill two birds with one stone with the next SSR test? Proves both token acceptance and cluster identity mapping. We could still split it into two tests that validate these two things. 

Fair question. We can certainly perform a shared SSR.

> Also, why not also use an SSR to prove the previous oauth token test as well?

Certainly can. I think I just went with the GET on Pods because it was a quick and easy way to say "we are now unauthorized to get something we were previously authorized to get".

I think an SSR is much more reliable, so I'll switch it to use that.

everettraven:
>We could move the keycloak_*.go files into their own package; it'll make it clearer to reuse functionality in other tests if ever needed.

Is it likely that we use this beyond testing authentication? IMO we can extract this whenever we have a use case to re-use it. For now, I imagine we only care to use it for authentication related tests.

everettraven:
/hold

botched rebase

pkg/testsuites/standard_suites.go
liouk: Right, that's a good point -- I just thought this was a typo, but this makes sense :+1:

test/extended/authentication/oidc.go
liouk: I think `authConfig` is still a valid oauth related arg to validate missing -- it basically contains the path for the oauth metadata file. It's just in the same level as `apiServerArguments` instead of part of it.

liouk:
Thanks for the changes @everettraven :+1: 

/lgtm

test/extended/util/client.go
xueqzhan: Do we have a job that exercise this path? This seems to the only change that might impact common path. So I want to double check.

xueqzhan (commented):
So the intention is to run those FG tests in separate jobs that exercise the new suite? Just want to confirm that non of the FG tests here will be exercised by existing conformance jobs and therefore you might need to create different variants of jobs to have enough coverage to graduate the feature.

everettraven:
> So the intention is to run those FG tests in separate jobs that exercise the new suite? Just want to confirm that non of the FG tests here will be exercised by existing conformance jobs and therefore you might need to create different variants of jobs to have enough coverage to graduate the feature.

@xueqzhan Correct. I have https://github.com/openshift/release/pull/66980 up to add a new periodic job to run the new suite. It doesn't yet contain all the variants to graduate the feature but I intend to expand that once this merges and I've determined the proper run time for this suite (it needs to be a long running suite due to each test needing to rollout a new revision of the KAS).

test/extended/util/client.go
everettraven: It looks like there are a few references in tests that I would expect to run in various jobs: https://github.com/search?q=repo%3Aopenshift/origin%20UserConfig&type=code (planning to search for an explicit job and update here)

If it feels too risky to make this change here, I could try to come up with some other approach.

everettraven:
Looks like there is a build error I need to look into https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29917/pull-ci-openshift-origin-main-e2e-gcp-fips-serial-2of2/1947670304780193792

/hold

everettraven:
Build issues were due to a need to rebase and properly pickup changes from an o/api bump I missed. Should be fixed now.

/hold cancel

test/extended/util/client.go
everettraven: Looks like some oauthserver and etcd tests use the `UserConfig()` method.

Specifically, https://github.com/openshift/origin/blob/e1ae86a0eda7f9052fe91d8f8fae85a65dc1ee45/test/extended/etcd/etcd_test_runner.go#L29-L57 looks like it is run as part of the standard OCP serial conformance suite - https://sippy.dptools.openshift.org/sippy-ng/tests/4.20/analysis?test=%5Bsig-api-machinery%5D%20API%20data%20in%20etcd%20should%20be%20stored%20at%20the%20correct%20location%20and%20version%20for%20all%20resources%20%5BSerial%5D%20%5BSuite%3Aopenshift%2Fconformance%2Fserial%5D&filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22%5Bsig-api-machinery%5D%20API%20data%20in%20etcd%20should%20be%20stored%20at%20the%20correct%20location%20and%20version%20for%20all%20resources%20%5BSerial%5D%20%5BSuite%3Aopenshift%2Fconformance%2Fserial%5D%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22never-stable%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22aggregated%22%7D%5D%2C%22linkOperator%22%3A%22and%22%7D

Because it calls https://github.com/openshift/origin/blob/e1ae86a0eda7f9052fe91d8f8fae85a65dc1ee45/test/extended/etcd/etcd_test_runner.go#L54 it should end up eventually triggering this code path because that call ends up using the same underlying call that I use in my test - `GetClientConfigForUser()`: https://github.com/openshift/origin/blob/e1ae86a0eda7f9052fe91d8f8fae85a65dc1ee45/test/extended/util/client.go#L268

test/extended/authentication/keycloak_client.go
```diff
@@ -0,0 +1,328 @@
...
+		adminURL: baseURL.JoinPath("admin", "realms", "master"),
+	}, nil
+}
+
+func (kc *keycloakClient) CreateGroup(name string) error {
+	groupURL := kc.adminURL.JoinPath("groups")
+
+	group := map[string]interface{}{
+		"name": name,
+	}
```
ibihim: Nit:

Alternatives:

```suggestion
	group := struct{
		Name string `json:"name"`
	}{
		Name: name,
	}
```

```suggestion
	group := map[string]any{
		"name": name,
	}
```

Applies to similar code in this file.

test/extended/authentication/keycloak_client.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+func (kc *keycloakClient) Authenticate(clientID, username, password string) error {
+	data := url.Values{
+		"username":   []string{username},
+		"password":   []string{password},
+		"grant_type": []string{"password"},
+		"client_id":  []string{clientID},
+		"scope":      []string{"openid"},
+	}
```
ibihim: Nit:

```suggestion
	data := url.Values{}
	data.Set("username", username)
	data.Set("password", password)
	data.Set("grant_type", "password")
	data.Set("client_id", clientID)
	data.Set("scope", "openid")
```

test/extended/authentication/keycloak_client.go
```diff
@@ -0,0 +1,328 @@
...
+	respBody := map[string]interface{}{}
+	respBodyData, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("reading response data: %w", err)
+	}
+
+	err = json.Unmarshal(respBodyData, &respBody)
+	if err != nil {
+		return fmt.Errorf("unmarshalling response body %s: %w", respBodyData, err)
+	}
```
ibihim: Streams without buffering entire reponse:

```suggestion
	if err := json.NewDecoder(resp.Body).Decode(&respBody); err != nil {
		return fmt.Errorf("unmarshalling response data: %w", err)
	}
```

Applies to other occurrences as well. Only thing that I feel strongly about. But I would not block the PR as this is not production code.

test/extended/authentication/keycloak_client.go
```diff
@@ -0,0 +1,328 @@
...
+	tokenURL := *kc.adminURL
+	tokenURL.Path = fmt.Sprintf("/realms/%s/protocol/openid-connect/token", kc.realm)
+
+	resp, err := kc.DoRequest(http.MethodPost, tokenURL.String(), "application/x-www-form-urlencoded", false, bytes.NewBuffer([]byte(data.Encode())))
+	if err != nil {
+		return fmt.Errorf("authenticating as user %q: %w", username, err)
+	}
+	defer resp.Body.Close()
+
+	respBody := map[string]interface{}{}
```
ibihim: nit: I am a strong believer in anonymous structs for type-safety :smile:

```go
  var tokenResponse struct {
      AccessToken string `json:"access_token"`
      IDToken     string `json:"id_token"`
      TokenType   string `json:"token_type,omitempty"`
      ExpiresIn   int    `json:"expires_in,omitempty"`
      Scope       string `json:"scope,omitempty"`
  }
```

test/extended/authentication/keycloak_client.go
```diff
@@ -0,0 +1,328 @@
...
+		"emailVerified": true,
+		"groups":        groups,
+		"credentials": []map[string]interface{}{
+			{
+				"temporary": false,
+				"type":      "password",
+				"value":     password,
+			},
+		},
+	}
```
ibihim: Nit:

```go
  user := struct {
      Username      string `json:"username"`
      Email         string `json:"email"`
      Enabled       bool   `json:"enabled"`
      EmailVerified bool   `json:"emailVerified"`
      Groups        []string `json:"groups"`
      Credentials   []struct {
          Temporary bool   `json:"temporary"`
          Type      string `json:"type"`
          Value     string `json:"value"`
      } `json:"credentials"`
  }{
      Username:      username,
      Email:         fmt.Sprintf("%s@payload.openshift.io", username),
      Enabled:       true,
      EmailVerified: true,
      Groups:        groups,
      Credentials: []struct {
          Temporary bool   `json:"temporary"`
          Type      string `json:"type"`
          Value     string `json:"value"`
      }{
          {
              Temporary: false,
              Type:      "password",
              Value:     password,
          },
      },
  }
```

test/extended/authentication/keycloak_client.go
```diff
@@ -0,0 +1,328 @@
...
+		"password":   []string{password},
+		"grant_type": []string{"password"},
+		"client_id":  []string{clientID},
+		"scope":      []string{"openid"},
+	}
+
+	tokenURL := *kc.adminURL
+	tokenURL.Path = fmt.Sprintf("/realms/%s/protocol/openid-connect/token", kc.realm)
+
+	resp, err := kc.DoRequest(http.MethodPost, tokenURL.String(), "application/x-www-form-urlencoded", false, bytes.NewBuffer([]byte(data.Encode())))
```
ibihim, line 149: Nit:

Maybe something like [this](https://pkg.go.dev/net/http@go1.24.5#Client.PostForm), would be cool:

```go
    kc.client.PostForm(tokenURL.String(), data)
```

test/extended/authentication/keycloak_helpers.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	cleanups = append(cleanups, cleanup)
+
+	cleanup, err = createKeycloakCAConfigMap(ctx, corev1Client)
+	if err != nil {
+		return cleanups, fmt.Errorf("creating CA configmap for keycloak: %w", err)
+	}
+	cleanups = append(cleanups, cleanup)
+
+	return cleanups, waitForKeycloakAvailable(ctx, client, namespace)
```
ibihim, line 81: hehe, let me talk you through the wonderful world of functional programming, the next time we drink a coffee together... :smile: 

Nit: a builder pattern would be cool (like we use it frequently in library-go) and a dedicated package (looks pretty interesting to import for kube-rbac-proxy and OIDC testing :stuck_out_tongue:)

test/extended/authentication/oidc.go
```diff
@@ -0,0 +1,548 @@
...
+		cleanups, err = deployKeycloak(ctx, oc, keycloakNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred(), "should not encounter an error deploying keycloak")
+
+		kcURL, err := admittedURLForRoute(ctx, oc, keycloakResourceName, keycloakNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred(), "should not encounter an error getting keycloak route URL")
+
+		keycloakCli, err = keycloakClientFor(kcURL)
+		o.Expect(err).NotTo(o.HaveOccurred(), "should not encounter an error creating a keycloak client")
+
+		// First authenticate as the admin keyloak user so we can add new groups and users
```
ibihim: Nit:

```suggestion
		// First authenticate as the admin keycloak user so we can add new groups and users
```

test/extended/authentication/oidc.go
```diff
@@ -0,0 +1,548 @@
...
+		if removal == nil {
+			continue
+		}
+		err := removal(ctx)
+		if err != nil && !apierrors.IsNotFound(err) {
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.Join(errs...)
```
ibihim: Nit:

[`errors.NewAggregate`](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go) is very popular and widely used around our codebases. Otherwise, there seem to be some nice filtering capabilities as well:

```go
errors.FilterOut(errors.NewAggregate(errs), apierrors.IsNotFound)
```

test/extended/authentication/oidc.go
```diff
@@ -0,0 +1,548 @@
...
+		return true, nil
+	})
+
+	return err
+}
+
+func waitForRollout(ctx context.Context, client *exutil.CLI) {
+	kasCli := client.AdminOperatorClient().OperatorV1().KubeAPIServers()
+
+	// First wait for KAS to flip to progessing
```
ibihim: Nit:

```suggestion
	// First wait for KAS to flip to progressing
```

test/extended/authentication/keycloak_client.go
everettraven: We only post form data for this one type of request. If we find ourselves repeating this logic quite frequently I could see value in a common method.

test/extended/authentication/keycloak_helpers.go
everettraven: A builder pattern feels a bit overkill for now. If we have use cases for a common library here, lets take that as a follow-up action to create that and use it across all the places necessary?

PR #29927, opened 2025-06-19
OCPBUGS-57385: Add time for each test operation

iurygregory:


test/extended/baremetal/high_availability.go
```diff
@@ -51,10 +51,14 @@ var _ = g.Describe("[sig-installer][Feature:baremetal][Serial] Baremetal platfor
 		c, err := e2e.LoadClientset()
 		o.Expect(err).ToNot(o.HaveOccurred())
 
+		start := time.Now()
+
 		g.By("delete cluster baremetal operator")
 		err = c.AppsV1().Deployments(baremetalNamespace).Delete(context.Background(), clusterBaremetalOperator, v1.DeleteOptions{})
 		o.Expect(err).NotTo(o.HaveOccurred())
+		e2e.Logf("Phase: delete cluster baremetal operator took %s\n", time.Since(start))
```
dgoodwin: i would go simpler, ditch the e2e.Logf which doesn't timestamp, import logrus, log.Infof("Deleting operator"), log each step you get too, this gives us real timestamps we can use to correlate with other things going on in the cluster.

no real need to track time deltas then we can see it from the timestamps, but its also ok to do so if it helps show the timing more easily.

iurygregory: @dgoodwin tks! i went to `e2e.Logf` since it was in the file 😅 , will update things

rhdmalone:
hi @iurygregory 
I have a PR that also has a failing mandatory test e2e_gcp_ovn_upgrade
The test e2e_gcp_ovn_upgrade is failing for the same reason as the one on this PR - the Dockerfile could not be built.
I also see in the test log file: 
No match for argument: python3-cinderclient
Error: Unable to find a match: python3-cinderclient

iurygregory:
@rhdmalone I'm not sure who owns this job and why it cant find the package. sorry

iurygregory:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.20-e2e-metal-ipi-serial-ovn-ipv6-runc

PR #29928, opened 2025-06-19
OCPBUGS-56691: Pass volume-percentage to 85 in must-gather tests

ardaguclu:
volume-percentage flag's default value 50 causes failures in some CI jobs that are executed in limited environments. Therefore, we need to set higher values in must-gather tests to not cause any failures, when we enable this PR https://github.com/openshift/oc/pull/2028 again (which was reverted https://github.com/openshift/oc/pull/2042)


ardaguclu:
After discussing with @ingvagabund offline, I need to make changes more carefully.
/close
closing for now

PR #29932, opened 2025-06-20
TRT-2172: Generate autodl artifact to study e2e tests correlated with high CPU

dgoodwin:
Look for e2e test intervals that overlap with high CPU alerts, be they
pending or firing. Output a datafile that will be automatically ingested
into bigquery where we can then study what tests are failing due to high
CPU, and perhaps what tests are causing it.

### DIFF
--- a/pkg/monitortests/testframework/highcputestanalyzer/monitortest_test.go
+++ b/pkg/monitortests/testframework/highcputestanalyzer/monitortest_test.go
@@ -443,6 +443,79 @@ func TestFindE2EIntervalsOverlappingHighCPU(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "same test runs twice and overlaps high cpu alert each time",
+			intervals: monitorapi.Intervals{
+				{
+					Source: monitorapi.SourceAlert,
+					Condition: monitorapi.Condition{
+						Locator: monitorapi.Locator{
+							Keys: map[monitorapi.LocatorKey]string{
+								"alert": "ExtremelyHighIndividualControlPlaneCPU",
+							},
+						},
+					},
+					From: now,
+					To:   now.Add(15 * time.Minute),
+				},
+				{
+					Source: monitorapi.SourceAlert,
+					Condition: monitorapi.Condition{
+						Locator: monitorapi.Locator{
+							Keys: map[monitorapi.LocatorKey]string{
+								"alert": "ExtremelyHighIndividualControlPlaneCPU",
+							},
+						},
+					},
+					From: now.Add(30 * time.Minute),
+					To:   now.Add(45 * time.Minute),
+				},
+				{
+					Source: monitorapi.SourceE2ETest,
+					Condition: monitorapi.Condition{
+						Locator: monitorapi.Locator{
+							Keys: map[monitorapi.LocatorKey]string{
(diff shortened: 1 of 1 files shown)


pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+
+	logrus.Infof("High CPU correlated tests: %d successful, %d failed", highCPUSuccessfulTests, highCPUFailedTests)
+
+	return rows
+}
+
+// overlaps checks if two intervals overlap in time
+func overlaps(interval1, interval2 monitorapi.Interval) bool {
```
xueqzhan: I feel this can be put somewhere common so that other use cases can call the same. We have similar checks in multiple places like https://github.com/openshift/origin/blob/25259415f83d2cac48232a5944ae65ffbc1d418c/pkg/monitortests/clusterversionoperator/operatorstateanalyzer/e2etest.go#L26.

pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return nil
+}
+
+func (*highCPUTestAnalyzer) Cleanup(ctx context.Context) error {
+	return nil
+}
+
+// findE2EIntervalsOverlappingHighCPU finds E2E test intervals that overlap with high CPU alert intervals
+func findE2EIntervalsOverlappingHighCPU(intervals monitorapi.Intervals) []map[string]string {
```
xueqzhan, line 82: Does kube event buffering (events might not be logged immediately) affect this?

pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
dgoodwin: Yeah I liked this implementation best, but the others didn't look like a clean switchover. I will see if I can merge them into one.

pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
dgoodwin: Shouldn't be, e2e-test intervals are created directly by the test runner, and alert intervals are scraped from prom at the end of the job run. Kube events should be out of the picture afaik.

pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
xueqzhan: I don't know if you need to merge them for now. But maybe put this in a common lib area for future use?

pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
xueqzhan: Yeah, I wasn't sure about the alert ones.

pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
dgoodwin: I've got them merged, testing now.

dgoodwin:
needs more work. Test can repeat, for example upgrade should succeed. Also seeing a lot of hits.

PR #29946, opened 2025-06-30
trt-2163: return error

neisw:
We should return an error when we fail to determine the openshift-tests image

cc: @vrutkovs

### DIFF
--- a/pkg/monitortests/kubeapiserver/disruptioninclusterapiserver/monitortest.go
+++ b/pkg/monitortests/kubeapiserver/disruptioninclusterapiserver/monitortest.go
@@ -328,8 +328,7 @@ func (i *InvariantInClusterDisruption) StartCollection(ctx context.Context, admi
 	cmd.Stdout = out
 	cmd.Stderr = errOut
 	if err := cmd.Run(); err != nil {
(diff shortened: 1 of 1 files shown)


vrutkovs:
/lgtm cancel

It seems to break `e2e-metal-ipi-ovn-ipv6`:
```
unable to determine openshift-tests image: exit status 1: error: unable to read image virthost.ostest.test.metalkube.org:5000/localimages/local-release-image@sha256:bfbcb49f5d0444487e97c892fdedaac3c1c9b73d9ceb5e1b6b1d772c06c2d378: Get "https://virthost.ostest.test.metalkube.org:5000/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

neisw:
Ok so then for ipv6 we don't expect this to work at all then I guess.  Question is can we detect that and set the notsupportedreason prior to this check or not.

vrutkovs:
We can disable it for metal:
```
jobType, err := platformidentification2.GetJobType(context.TODO(), clientConfig)
if err == nil && jobType.Platform == "metal" {
	i.notSupportedReason = fmt.Sprintf("unable to determine openshift-tests image: %v: %v", err, errOut.String())
	return nil
}
```

PR #29947, opened 2025-06-30
OCPBUGS-56685: Skip SigstoreImageVerification skip test on ipv6 platform

QiWang19:
It seems images should follow this https://github.com/openshift/origin/tree/main/test/extended/util/image#to-add-a-new-image so the openshift-test binary can generate a file that is used by `oc image mirror` for a disconnected environment.
But the example file does not include the signature `.sig` tag entry for mirroring, and `oc image mirror` does not copy the signature when only specifying the image.

### DIFF
--- a/test/extended/imagepolicy/imagepolicy.go
+++ b/test/extended/imagepolicy/imagepolicy.go
@@ -3,6 +3,7 @@ package imagepolicy
 import (
 	"context"
 	"fmt"
+	"net"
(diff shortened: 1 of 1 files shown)


QiWang19:
/payload-job periodic-ci-openshift-release-master-nightly-4.20-e2e-metal-ipi-serial-ovn-ipv6-techpreview

PR #29953, opened 2025-07-04
NO-JIRA: OWNERS_ALIASES: Update OTA members / team lead

petr-muller:
- @PratikMahajan replaced Lala as a team lead
- Both @DavidHurta and @hongkailiu are more than qualified to review and approve OTA-related tests


wking (approved):
/lgtm

PR #29955, opened 2025-07-04
Improve performance of run-resourcewatch

mdbooth:
This makes a number of substantial changes to run-resourcewatch, but does not
change the output format, or the behaviour of any existing commands.

- **run-resourcewatch: re-implement watch loop**

This re-implements how we observe resources and the write loop. We no longer
use informers, but instead use explicit list and watch. We also de-duplicate
resources, and reconstruct missed deletes.

Each resource now has its own goroutine, but all resource observations are
pushed to a single channel. This means that writes to the git repository are
now serialised. This is a natural fit, as git does not support concurrent
writers in any case.

We buffer between resource observations and writing to git by making the
resource channel very large. This means that any write backlog will be buffered
in memory.

- **run-resourcewatch: Add --to-json and --from-json**

These are additional, mutually exclusive command line options.

  --to-json causes run-resourcewatch to write resource observations as a stream
  of json objects to a single file

  --from-json causes run-resourcewatch to read resource observations from a
  a file instead of from a cluster.

- **run-resourcewatch: Add a gitstorage benchmark**

This adds a simple git writing benchmark test and some captured data from an
AWS cluster installation.

Note that this adds a 14M gzip compressed json file to the repo as test data.

- **run-resourcewatch: Remove goroutines from gitstorage**
- **run-resourcewatch: Fix delete of non-observed resource**
- **run-resourcewatch: Don't run git commands with bash**
- **run-resourcewatch: Remove command polling**

Together the above changes yielded more than a 10x performance improvement when
writing to git as measured by the benchmark test executed for 30 seconds, at
least on my workstation. The majority of the performance gains come from the
first commit which completely serialises gitstorage. This yielded a 7x
performance improvement on its own.

An AWS cluster installation takes approximately 30-40 minutes, and on my
workstation at least run-resourcewatch can write the full dataset from json to
git in 7 minutes. I believe it should be able to keep up without requiring a
post-processing step.

- **run-resourcewatch: Use observation timestamp as commit date**

This change uses the time at which the observation was made as the commit date
rather than the time of the commit. This means that commit date are still
potentially meaningful even if we had some buffering due to slow writes, or
even if we wrote to json and post-processed the output to a git repo.

- **run-resourcewatch: Disable git automatic gc**
- **run-resourcewatch: Re-add a retry loop for git commands**

Together these resolve the issue of git commands failing due to 'index.lock'.
It seems that this was caused when a git command triggered automatic garbage
collection, which leaves a background process running holding a lock, causing
subsequent commands to fail until the gc completes. We now disable auto gc
during collection, and instead run a single gc before exiting.

I re-added a retry loop only because I thought I saw an error even after
disabling auto gc. However I have not been able to reproduce this after many
attempts, so this may have been a hallucination.

In general, we never expect local git commands to fail.


mdbooth:
I can't reproduce those lint errors locally, and they're not obviously related to this PR. Related to the Go bump, perhaps? Does anybody know what's going on here?

mdbooth:
Bumping go to 1.24 separately: https://github.com/openshift/origin/pull/29965

mdbooth:
The Go 1.24 bump looks like a can of worms. I've updated the benchmark test to use only 1.23 features.

mdbooth:
Results of running the gitstore benchmark with `-benchtime=30s` on my workstation:

Before optimisation:
```
     358         223620723 ns/op
```

After optimisation:
```
    3026          15785690 ns/op
```

neisw:
Looks like [verify-deps](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/29955/pull-ci-openshift-origin-main-verify-deps/1943021133062213632) failures `go.mod` content is incorrect - did you run `go mod tidy`?`

pkg/resourcewatch/git/git_store.go
```diff
@@ -0,0 +1,325 @@
...
+	gitOpModified
+	gitOpDeleted
+	gitOpError
+)
+
+// NewGitStorage returns the resource event handler capable of storing changes observed on resource
+// into a Git repository. Each change is stored as separate commit which means a full history of the
+// resource lifecycle is preserved.
+func NewGitStorage(path string) (*GitStorage, error) {
+	// If the repo does not exists, do git init
```
neisw: I wasn't gonna comment on the nit but since there is a verify-deps issue exists -> exist

mdbooth: Looks like this typo has been around since the code was added in 2020, but I'll fix it now we've seen it 👍

neisw: Thanks, just popped out due to getting moved around I guess.  No issue if you just want to hit the `go mod tidy`.

PR #29960, opened 2025-07-07
OCPBUGS-57032: upgrade.go: wait some time after node upgrade

dgrisonnet:
For the "Cluster should remain functional during upgrade" test, TRT noticed flakes from the step that verifies that deamonsets are running on all expected nodes after an upgrade.  This flake was caused by the verification of the deamonset happening too quickly after the upgrade. As soon as the last upgraded node becomes ready the check happens, but it doesn't always leave enough time for the deamonset to restart, thus causing the test to fail.

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -188,6 +188,9 @@ var _ = g.Describe("[sig-arch][Feature:ClusterUpgrade]", func() {
 						clusterUpgrade(f, client, dynamicClient, config, upgCtx.Versions[i]),
 						fmt.Sprintf("during upgrade to %s", upgCtx.Versions[i].NodeImage))
 				}
+				// Sleep to give some time to the workloads on the last upgraded
+				// node to restart.
+				time.Sleep(30 * time.Second)
 			},
 		)
 	})


test/e2e/upgrade/upgrade.go
```diff
@@ -188,6 +188,9 @@ var _ = g.Describe("[sig-arch][Feature:ClusterUpgrade]", func() {
 						clusterUpgrade(f, client, dynamicClient, config, upgCtx.Versions[i]),
 						fmt.Sprintf("during upgrade to %s", upgCtx.Versions[i].NodeImage))
 				}
+				// Sleep to give some time to the workloads on the last upgraded
+				// node to restart.
+				time.Sleep(30 * time.Second)
```
dgoodwin: Could you take the poll approach so if we only need a few seconds, we don't use the full 30? With 5000+ tests we need to minimize the sleeps whenever possible. One of the wait.Poll functions.

sdodson: Used Cursor to refactor this with a polling loop in https://github.com/openshift/origin/pull/30062 tried to give credit in the PR description for where the real work happened.

dgrisonnet:
superseded by https://github.com/openshift/origin/pull/30062

sdodson:
Looking at this more closely I don't think #30062 actually achieved anything, it's probably not harmful but it's outside the scope of where this is failing. 

In all of the cases I've looked at the time between when the daemonset check and the pod being started the time is generally no more than two seconds. If we're not happy with 30s then I'd propose we go with 5s as a start. Theonly other solution i could think of is to add in a polling function to look at all pods on the cluster after the MCP is updated and wait up to 30 seconds for any pods that are initializing to finish, like just before https://github.com/openshift/origin/blob/main/test/e2e/upgrade/upgrade.go#L688

@dgoodwin do you mind reconsidering the simple sleep after each upgrade?

/reopen

test/e2e/upgrade/upgrade.go
```diff
@@ -188,6 +188,13 @@ var _ = g.Describe("[sig-arch][Feature:ClusterUpgrade]", func() {
 						clusterUpgrade(f, client, dynamicClient, config, upgCtx.Versions[i]),
 						fmt.Sprintf("during upgrade to %s", upgCtx.Versions[i].NodeImage))
 				}
+				err = wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
+					if err := checkUpgradeability(client); err != nil {
```
dgrisonnet: I don't think polling `checkUpgradeability` will improve the situation as it doesn't check for any statuses that would report the deamonset as not fully operating. I approved the changes with this check in the other PR because I inadvertently thought that it was this test, but this one doesn't check for cluster upgradeability.

I initially went with a time.Sleep because I couldn't find a status that would be easy to poll, but maybe we could find something.

dgrisonnet:
>  If we're not happy with 30s then I'd propose we go with 5s as a start

I think that would be reasonable as it is difficult to find something to poll. @dinhxuanvu is looking into it right now but if we can't find anything, then maybe having a 5s sleep is reasonnable and will be enough to improve this test passing rate.

test/e2e/upgrade/upgrade.go
dinhxuanvu: Updated the sleeptime to 5s. I think that is reasonable enough.

sdodson:
/approve
LGTM but wanted to give Damien a chance to weigh in. I'll go ahead with the merge tomorrow if we don't hear anything negative.

dinhxuanvu:
> /approve LGTM but wanted to give Damien a chance to weigh in. I'll go ahead with the merge tomorrow if we don't hear anything negative.

Damien and I discussed this offline and he agreed with my change. I will ask him to lgtm on the PR in a bit. Thanks
@dgrisonnet PTAL

dinhxuanvu:
@sdodson Ready for approval and merge :)

PR #29965, opened 2025-07-07
NO-JIRA: Bump Go to 1.24

mdbooth:
Go 1.24 requires format strings to be string constants.


mdbooth:
The lint errors are wrong. In the past I've found these problems normally go away when I bump golanci-lint. However, the newest golangci-lint image we have is built against Go 1.23, so we can't use it with Go 1.24:

https://github.com/openshift/release/blob/8f06238188359fa692b5e270e3d16aafaefc7034/core-services/image-mirroring/_config.yaml#L836-L837

PR #29976, opened 2025-07-09
WIP:test/extended/testdata/cmd/test/cmd/images: Collect all ImageStreams

wking:
For debugging when they don't import successfully.


wking:
I never got this debug gathering working, but openshift/openshift-apiserver#533 is up to revert the schema-1 removal that was behind the CI failures, so I don't need more debugging in this test-case at the moment.  Closing this pull.

PR #29978, opened 2025-07-09
OCPBUGS-59159: Adjust DNS success rate to 98.0% for SNO upgrade test.

jaypoulz:
The old solution for this was to wait longer for the probe to gather additional DNS data. Neither that solution nor this one addresses the root cause of the issue - that SNO has more recovery disruption than other topologies.

This solution aims to allow us to refine the success threshold for SNO to be something that should always pass, while still detecting statistically significant regressions.

### DIFF
--- a/test/e2e/upgrade/dns/dns.go
+++ b/test/e2e/upgrade/dns/dns.go
@@ -26,6 +26,11 @@ import (
 
 type UpgradeTest struct{}
 
+const (
+	successRateThresholdDefault    = 99.0
+	acceptableSingleNodeDisruption = 1.0
+)
+
 var appName string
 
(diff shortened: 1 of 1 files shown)


eggfoobar:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.20-e2e-aws-upgrade-ovn-single-node 10

jaypoulz:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.20-e2e-aws-upgrade-ovn-single-node 10

jaypoulz:
that's 20/20 for DNS successes. Fixing the prow verify, but then I think we're ready to move forward and monitor how it does live

PR #29985, opened 2025-07-14
OCPBUGS-59176: skip specific tests if DNSManaged is false

lihongan:
changes:

- if no dns zones in dns.config, then skip the specific tests that need create/publish dnsrecords (e.g. for shard ingresscontroller or gateway)
- update shard ingressconroller to use different domain to avoid overlapping with default one `*.apps.<baseDomain>`

update: 
The `gcp-user-provisioned-dns` cluster set `LoadBalancer.DNSManagementPolicy ` as `UnmanagedLoadBalancerDNS`, that is why DNSManaged is false. No dns zones is just one of conditions to set DNSManaged as false and cannot be used in this case.

### DIFF
--- a/test/extended/router/gatewayapicontroller.go
+++ b/test/extended/router/gatewayapicontroller.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"reflect"
 	"strings"
 	"time"
 
@@ -218,6 +219,14 @@ var _ = g.Describe("[sig-network-edge][OCPFeatureGate:GatewayAPIController][Feat
 	})
 
 	g.It("Ensure HTTPRoute object is created", func() {
+		isDNSManaged, err := isDNSManaged(oc, time.Minute)
+		if err != nil {
+			e2e.Failf("Failed to get default ingresscontroller DNSManaged status: %v", err)
+		}
+		if !isDNSManaged {
+			g.Skip("Skipping on this cluster since DNSManaged is false")
+		}
+
 		g.By("Ensure default GatewayClass is accepted")
 		errCheck := checkGatewayClass(oc, gatewayClassName)
 		o.Expect(errCheck).NotTo(o.HaveOccurred(), "GatewayClass %q was not installed and accepted", gatewayClassName)
@@ -425,14 +434,32 @@ func assertDNSRecordStatus(oc *exutil.CLI, gatewayName string) {
 		for _, record := range gatewayDNSRecords.Items {
 			if record.Labels["gateway.networking.k8s.io/gateway-name"] == gatewayName {
 				gatewayDNSRecord = &record
+				e2e.Logf("Found the desired dnsrecord and spec is: %v", gatewayDNSRecord.Spec)
 				break
 			}
+		}
+
+		// skip status check if privateZone/publicZone in dns.config is nil or empty
+		emptyDNSZone := &configv1.DNSZone{}
+		dns, err := oc.AdminConfigClient().ConfigV1().DNSes().Get(context, "cluster", metav1.GetOptions{})
+		if err != nil {
+			e2e.Logf("Get dnses.config/cluster failed: %v, retrying...", err)
+			return false, nil
+		}
+		if dns.Spec.PublicZone == nil && dns.Spec.PrivateZone == nil {
+			e2e.Logf("All zones in dns.config are nil, needn't to check status")
+			return true, nil
+		}
+		if reflect.DeepEqual(dns.Spec.PublicZone, emptyDNSZone) || reflect.DeepEqual(dns.Spec.PrivateZone, emptyDNSZone) {
+			e2e.Logf("Zone in dns.config is empty, needn't to check status")
+			return true, nil
 		}
 
 		// checking the gateway DNS record status
 		for _, zone := range gatewayDNSRecord.Status.Zones {
 			for _, condition := range zone.Conditions {
 				if condition.Type == "Published" && condition.Status == "True" {
+					e2e.Logf("The published status is true for zone %v", zone.DNSZone)
 					return true, nil
 				}
 			}
--- a/test/extended/router/grpc-interop.go
+++ b/test/extended/router/grpc-interop.go
@@ -53,8 +53,16 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 				g.Skip("Skip on platforms where the default router is not exposed by a load balancer service.")
 			}
 
-			defaultDomain, err := getDefaultIngressClusterDomainName(oc, time.Minute)
-			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+			isDNSManaged, err := isDNSManaged(oc, time.Minute)
+			if err != nil {
+				e2e.Failf("Failed to get default ingresscontroller DNSManaged status: %v", err)
+			}
+			if !isDNSManaged {
+				g.Skip("Skipping on this cluster since DNSManaged is false")
+			}
+
+			baseDomain, err := getClusterBaseDomainName(oc, time.Minute)
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find base domain name")
 
 			g.By("Locating the canary image reference")
 			image, err := getCanaryImage(oc)
@@ -196,7 +204,7 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 			pemCrt2, err := certgen.MarshalCertToPEMString(tlsCrt2Data)
 			o.Expect(err).NotTo(o.HaveOccurred())
 
-			shardFQDN := oc.Namespace() + "." + defaultDomain
+			shardFQDN := oc.Namespace() + "." + baseDomain
 
 			g.By("Creating routes to test for gRPC interoperability")
 			routeType := oc.Namespace()
--- a/test/extended/router/http2.go
+++ b/test/extended/router/http2.go
@@ -24,6 +24,7 @@ import (
 	utilpointer "k8s.io/utils/pointer"
 
 	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	routeclientset "github.com/openshift/client-go/route/clientset/versioned"
 
@@ -109,8 +110,16 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
 				g.Skip("Skip on platforms where the default router is not exposed by a load balancer service.")
 			}
 
-			defaultDomain, err := getDefaultIngressClusterDomainName(oc, time.Minute)
-			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+			isDNSManaged, err := isDNSManaged(oc, time.Minute)
+			if err != nil {
+				e2e.Failf("Failed to get default ingresscontroller DNSManaged status: %v", err)
+			}
+			if !isDNSManaged {
+				g.Skip("Skipping on this cluster since DNSManaged is false")
+			}
+
+			baseDomain, err := getClusterBaseDomainName(oc, time.Minute)
+			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find base domain name")
 
 			g.By("Locating the canary image reference")
 			image, err := getCanaryImage(oc)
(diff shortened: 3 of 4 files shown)


test/extended/router/http2.go
```diff
@@ -598,3 +623,22 @@ func platformHasHTTP2LoadBalancerService(platformType configv1.PlatformType) boo
...
+	noDNSZones := false
+
+	if err := wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		dns, err := oc.AdminConfigClient().ConfigV1().DNSes().Get(context.TODO(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			e2e.Logf("Get dnses.config/cluster failed: %v, retrying...", err)
+			return false, nil
+		}
+		if dns.Spec.PublicZone == nil && dns.Spec.PrivateZone == nil {
+			noDNSZones = true
```
lihongan: The condition `dns.Spec.PublicZone == nil && dns.Spec.PrivateZone == nil` is not available to check the `GCP user provisioned DNS` cluster, instead the cluster set `LoadBalancer.DNSManagementPolicy` as below
```
			if platformStatus.GCP != nil && platformStatus.GCP.CloudLoadBalancerConfig != nil &&
				platformStatus.GCP.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
				effectiveStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS
			}
```

see https://github.com/openshift/cluster-ingress-operator/blob/cbc0b217b655f1f0ce0becc9145c2a6042beabea/pkg/operator/controller/ingress/controller.go#L478-L492

We might need to update the func to check default ingresscontroller OperatorCondition `Type:   "DNSManaged"` is `ConditionFalse`

test/extended/router/http2.go
```diff
@@ -598,3 +624,28 @@ func platformHasHTTP2LoadBalancerService(platformType configv1.PlatformType) boo
 		return false
 	}
 }
+
+func isDNSManaged(oc *exutil.CLI, timeout time.Duration) (bool, error) {
+	var status string
```
alebedev87: To avoid 2 conversions down the code:
```suggestion
	var status operatorv1.ConditionStatus
```

test/extended/router/http2.go
```diff
@@ -598,3 +624,28 @@ func platformHasHTTP2LoadBalancerService(platformType configv1.PlatformType) boo
...
+			}
+		}
+		return false, nil
+	}); err != nil {
+		return false, err
+	}
+	if status == string(operatorv1.ConditionFalse) {
+		return false, nil
+	}
+	return true, nil
```
alebedev87:
```suggestion
	return status !=operatorv1.ConditionFalse , nil
```

test/extended/router/gatewayapicontroller.go
```diff
@@ -195,6 +195,14 @@ var _ = g.Describe("[sig-network-edge][OCPFeatureGate:GatewayAPIController][Feat
 	})
 
 	g.It("Ensure LB, service, and dnsRecord are created for a Gateway object", func() {
+		isDNSManaged, err := isDNSManaged(oc, time.Minute)
+		if err != nil {
+			e2e.Failf("Failed to get default ingresscontroller DNSManaged status: %v", err)
+		}
+		if !isDNSManaged {
+			g.Skip("Skipping on this cluster since DNSManaged is false")
+		}
```
alebedev87: Is `DNSRecord` still created if there are no zones in `DNSConfig`? If so, I think that we should not skip but do a dedicated `assertDNSRecord` (instead of to the current `assertDNSRecordStatus`). WDYT?

test/extended/router/grpc-interop.go
```diff
@@ -53,8 +53,16 @@ var _ = g.Describe("[sig-network-edge][Conformance][Area:Networking][Feature:Rou
...
 
-			defaultDomain, err := getDefaultIngressClusterDomainName(oc, time.Minute)
-			o.Expect(err).NotTo(o.HaveOccurred(), "failed to find default domain name")
+			isDNSManaged, err := isDNSManaged(oc, time.Minute)
+			if err != nil {
+				e2e.Failf("Failed to get default ingresscontroller DNSManaged status: %v", err)
+			}
+			if !isDNSManaged {
+				g.Skip("Skipping on this cluster since DNSManaged is false")
+			}
```
alebedev87, line 62: I'm wondering why only GRPC interoperability test should be skipped. We have a lot more tests in `extended/router` suite, do they all work when there are no DNS zones for the cluster?

lihongan: Firstly, there are three conditions for `DNSManaged` status and if any of them are unsatisfied the status will be set as `False`,  DNS zones in dns.config is just one of condition. see https://github.com/openshift/api/blob/fed56f2794b13fb3691cee40d0b429c2111d0e7f/operator/v1/types_ingress.go#L2060-L2065

Secondly, in the job "e2e-gcp-user-provisioned-dns" we just see the 4 failing tests, and the cluster set `LoadBalancer.DNSManagementPolicy = UnmanagedLoadBalancerDNS` but not rely on DNS zones. see https://github.com/openshift/cluster-ingress-operator/blob/cbc0b217b655f1f0ce0becc9145c2a6042beabea/pkg/operator/controller/ingress/controller.go#L478-L492

Thirdly, I believe most tests in `extended/router` suite are just tested with default ingresscontroller and the request go through default router pod, but GRPC/http2 tests create another shard ingresscontroller to test the feature and it rely on cloud DNS to resolve the shard ingresscontroller domain.

test/extended/router/gatewayapicontroller.go
lihongan: Looks `DNSRecord` is always created but the `.status` might be nil (dns.Spec.PublicZone == nil && dns.Spec.PrivateZone == nil) or shows `Published` type as `False` (PublicZone == {} or PrivateZone == {}).

Do you mean we should check `dnses.config.openshift.io/v1` firstly then make decision to `assertDNSRecord` or `assertDNSRecordStatus` ?

test/extended/router/http2.go
lihongan: Updated as suggested code. Thanks.

test/extended/router/gatewayapicontroller.go
lihongan: I removed the skip and added new logic in `assertDNSRecordStatus` to not check status if zone of dns.config is nil or empty.

test/extended/router/grpc-interop.go
sadasu: @alebedev87 , @lihongan just checking to see what the resolution of this is. We need this fix so that the "e2e-gcp-user-provisioned-dns" periodic jobs are not in the Red due to these tests and force this feature to be reverted.

test/extended/router/gatewayapicontroller.go
alebedev87: > I removed the skip and added new logic in assertDNSRecordStatus to not check status if zone of dns.config is nil or empty.

What I meant is that `assertDNSRecordStatus` should check that `DNSRecord`'s Published == `False` and `Status` is `nil` if there are no DNS zones in the DNS config. Just like your mentioned in the comment above the latest:

> Looks DNSRecord is always created but the .status might be nil (dns.Spec.PublicZone == nil && dns.Spec.PrivateZone == nil) or shows Published type as False (PublicZone == {} or PrivateZone == {}).

alebedev87:
As discussed, even if the DNS is not managed on cluster, an IngressController or Gateway can still have internet facing load balancers (provisioned for their publishing services). We can amend the logic of sending of HTTP requests to target the load balancer's DNS name in case the DNS is not managed. This would also need the Host header to be set to the HTTPRoute or Route's hostname ([goclient example of how to do this](https://github.com/openshift/cluster-ingress-operator/blob/c2c37cad52b2375216dca78cd1e322d2c35454de/test/e2e/operator_test.go#L3132-L3139)).

sadasu:
> As discussed, even if the DNS is not managed on cluster, an IngressController or Gateway can still have internet facing load balancers (provisioned for their publishing services). We can amend the logic of sending of HTTP requests to target the load balancer's DNS name in case the DNS is not managed. This would also need the Host header to be set to the HTTPRoute or Route's hostname ([goclient example of how to do this](https://github.com/openshift/cluster-ingress-operator/blob/c2c37cad52b2375216dca78cd1e322d2c35454de/test/e2e/operator_test.go#L3132-L3139)).

@alebedev87 would like to see these changes done as part of this fix or as a follow up to skipping the tests temporarily.

alebedev87:
> would like to see these changes done as part of this fix or as a follow up to skipping the tests temporarily.

@sadasu: We discussed this with Hongan, we thought that we could give it a chance to be implemented. The possibility to do a follow-up PR was discussed too but we were not sure about the priority. Is it needed before 4.20 branch cut?

candita:
This is more likely due to the new ClusterHostedDNS feature being tested on a nightly.

/hold

test/extended/router/grpc-interop.go
candita: Are we planning to make this a permanent skip?  I don't agree with that.

test/extended/router/gatewayapicontroller.go
```diff
@@ -425,14 +434,32 @@ func assertDNSRecordStatus(oc *exutil.CLI, gatewayName string) {
...
 			if record.Labels["gateway.networking.k8s.io/gateway-name"] == gatewayName {
 				gatewayDNSRecord = &record
+				e2e.Logf("Found the desired dnsrecord and spec is: %v", gatewayDNSRecord.Spec)
 				break
 			}
 		}
 
+		// skip status check if privateZone/publicZone in dns.config is nil or empty
+		emptyDNSZone := &configv1.DNSZone{}
+		dns, err := oc.AdminConfigClient().ConfigV1().DNSes().Get(context, "cluster", metav1.GetOptions{})
```
candita, line 444: Can this be checked outside the 10 minute loop, or do we need to check it every time?

test/extended/router/grpc-interop.go
lihongan: As discussed with Andrey, I plan to amend the logic of sending of HTTP requests to target the Load Balancer's IP address in case the DNS is not managed, with this trick then we don't need to skip the tests.

alebedev87:
> This is more likely due to the new ClusterHostedDNS feature being tested on a nightly.

Ack. This explains some things. From [the customer configured DNS EP](https://github.com/openshift/enhancements/pull/1468/files#diff-230cf161d0e8b0dc075e6760516b1171448adae3e6ee937ac9a97a0ce44862ddR106-R120):

> OpenShift will start static CoreDNS pods to provide DNS resolution for API, Internal API and Ingress services that are essential for cluster creation. In order for a master or worker node to come online, they need the machine config server via the api-int domain. So, it is essential for the api-int domain to be resolvable on the bootstrap node. The static CoreDNS pod started on the bootstrap node needs entries for just api-int resolution and the static CoreDNS pod started on the control plane nodes needs to be able to resolve api, api-int and *.apps domains. Although not identical, this approach leverages learnings from OpenShift's approach to providing in-cluster DNS for on-prem platforms.
> 
> After cluster deployment is completed, the customer will update their external DNS solution with the same assigned LB IP addresses used for the configuration of the internal CoreDNS instance. OpenShift will not delete the CoreDNS pod even after cluster installation completes.
> 
> If the user successfully configures their external DNS service with api, api-int and *.apps services, then they could optionally delete the in-cluster CoreDNS pod and the cluster is expected to function fully as expected. This is a completely optional step with this design. If the customer does configure their custom DNS solution and leave the in-cluster CoreDNS pod running, all in-cluster entities will end up using the CoreDNS pod's DNS services and all out of cluster requests will be handled by the external DNS.

The http2 and grpc tests use dedicated ingresscontrollers, so the default wildcard created by the new static CoreDNS won't work for those tests.

@lihongan:

1) Can you please update the description of the bug to mention that the failure was found during the testing of featuregates for the promotion to GA? Currently the bug does not name featuregates and doesn't mention that they are from `TechPreviewNoUpgrade` featureset.
2) Can you please update the PR description and the commit message to explain in details why we change the domain and logic only for GRPC/HTTP2 test cases (explanation above)?

Also, it seems like the GatewayAPI use case was overlooked. We have to follow up on the final decision, so far I see 2 PRs for this:
- https://github.com/openshift/cluster-ingress-operator/pull/1269
- https://github.com/openshift/cluster-ingress-operator/pull/1270

lihongan:
Thanks @alebedev87 , I updated the bug description and posted new PR to try to fix the tests instead of skipping them.

lihongan:
/close
superseded by https://github.com/openshift/origin/pull/30131

PR #29986, opened 2025-07-15
OCPEDGE-1755: Add TNF E2E tests for ungraceful node shutdown 

clobrano:
This change extends the etcd recovery tests in the two_node suite with a new ungraceful shutdown test

Key changes include:
- refactor `test/e2e/upgrade/monitor.go` to allow reusing the utility to trigger graceful/ungraceful node reboots.
- New TNF Ungraceful node shutdown test.

Other minor changes in TNF E2E tests:
- Refactor the Etcd validation logic into a dedicated function, to improve readability and reusability.
- Improve logging and error messages to provide more context during test failures.

### DIFF
--- a/test/extended/two_node/tnf_recovery.go
+++ b/test/extended/two_node/tnf_recovery.go
@@ -12,7 +12,7 @@ import (
 	o "github.com/onsi/gomega"
 	v1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/origin/test/extended/etcd/helpers"
-	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util"
 	"github.com/pkg/errors"
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	corev1 "k8s.io/api/core/v1"
@@ -21,10 +21,11 @@ import (
 
 const (
 	nodeIsHealthyTimeout         = time.Minute
+	etcdOperatorIsHealthyTimeout = time.Minute
 	memberHasLeftTimeout         = 5 * time.Minute
-	memberIsLeaderTimeout        = 2 * time.Minute
+	memberIsLeaderTimeout        = 10 * time.Minute
 	memberRejoinedLearnerTimeout = 10 * time.Minute
-	memberPromotedVotingTimeout  = 10 * time.Minute
+	memberPromotedVotingTimeout  = 15 * time.Minute
 	pollInterval                 = 5 * time.Second
 )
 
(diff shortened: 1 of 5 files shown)


test/extended/two_node/tnf_recovery.go
```diff
@@ -22,146 +22,115 @@ import (
...
-				return err
-			} else if !started || learner {
-				return fmt.Errorf("Expected node: %s to be a started and voting member. Membership: %+v", nodeB.Name, members)
-			}
-
-			g.GinkgoT().Logf("membership: %+v", members)
-			return nil
-		}, memberPromotedVotingTimeout, pollInterval).ShouldNot(o.HaveOccurred())
+			return helpers.EnsureHealthyMember(g.GinkgoT(), etcdClientFactory, targetNode.Name)
+		}, memberHasLeftTimeout, pollInterval).Should(o.HaveOccurred(), fmt.Sprintf("Time out waiting for the node %s to leave the etcd cluster", targetNode.Name))
```
slintes: naming things is a bit confusing here:

- log statement above is about "unreachable"
- test method is about "healthy" member
- timeout var name and comment in case test fails is about "leaving"

test/extended/two_node/tnf_recovery.go
```diff
@@ -22,146 +22,115 @@ import (
...
+
+		//g.By(fmt.Sprintf("Ensuring leader %s is a healthy voting member", survivedNode.Name))
```
slintes: I would assume this should not be commented out?

clobrano: oops, it seems I left some debug code

test/extended/two_node/tnf_recovery.go
clobrano: Indeed, I need to refactor this a bit

clobrano: I think I don't need this block. It is a convoluted way to "wait" for the target node to disappear from the membership, but I can just wait for the survived one to restore the cluster

test/extended/two_node/tnf_recovery.go
```diff
@@ -21,147 +21,108 @@ import (
...
+	g.It("Should recover from ungraceful node shutdown with etcd member re-addition", func() {
+		g.By(fmt.Sprintf("Shutting down %s ungracefully in 1 minute", targetNode.Name))
+		err := util.TriggerNodeRebootUngraceful(oc.KubeClient(), targetNode.Name)
+		o.Expect(err).To(o.BeNil(), "Expected to ungracefully shutdown the node without errors", targetNode.Name, err)
+		time.Sleep(1 * time.Minute)
+
+		g.By(fmt.Sprintf("Ensuring that %s added %s back as learner", survivedNode.Name, targetNode.Name))
+		validateEtcdRecoveryState(etcdClientFactory,
+			&survivedNode, true, false, // survivedNode expected started == true, learner == false
+			&targetNode, true, true, // targetNode expected started == false, learner == true
```
slintes: Arguments (true / true) don't match comment (false / true)? Not sure which one should be correct tbh :D If it's the comment, the only difference to the graceful test would be that the target doesn't leave the member list?

clobrano: Comment is right here :) 

> the only difference to the graceful test would be that the target doesn't leave the member list?

Yes, and the Etcd membership is disrupted until the survived node recovers the cluster, so we cannot "ask" about the status of the membership until then, and at that point the most likely picture is the survived node started=true and voter (learner=false), and the targed not-started and learner.

The test works because this membership is still a valid one, in fact we test it in the very next `validateEtcdRecoveryState`. I'll fix it

test/extended/util/nodes.go
```diff
@@ -87,3 +91,70 @@ func DebugSelectedNodesRetryWithOptionsAndChroot(oc *CLI, selector string, debug
...
+	return triggerNodeReboot(kubeClient, nodeName, 0, command)
+}
+
+// TriggerNodeRebootUngraceful initiates an ungraceful node reboot which does not allow the system to terminate processes cleanly before rebooting.
+func TriggerNodeRebootUngraceful(kubeClient kubernetes.Interface, nodeName string) error {
+	command := "echo 'reboot in 1 minute'; exec chroot /host sudo systemd-run sh -c 'sleep 60 && reboot --force --force'"
+	return triggerNodeReboot(kubeClient, nodeName, 0, command)
+}
+
+func triggerNodeReboot(kubeClient kubernetes.Interface, nodeName string, attempt int, command string) error {
```
xueqzhan, line 107: I assume this function is a straight copy from the original.

clobrano: yes

clobrano: well, except it takes the `command` as input, instead than having it hardcoded inside

clobrano:
> Please have your team review review the tnf portion and LGTM

There is one review already done from @slintes

PR #29991, opened 2025-07-15
OCPBUGS-59372: revert: remove openshift/two-node suite from tna tests

eggfoobar:
We need to revert the addition of the two-node suite label to TNA test as it removed the tests from our regular runs.

### DIFF
--- a/test/extended/two_node/arbiter_topology.go
+++ b/test/extended/two_node/arbiter_topology.go
@@ -29,7 +29,7 @@ var (
 	}
 )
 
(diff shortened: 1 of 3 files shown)


jaypoulz:
The intent was to make sure that these tests were also included in the two-node suite. The unintended site effect was that we didn't mean to remove them from the pool of tests running in parallel e2e. We'll have to investigate in a follow-up if it's possible to run tests in multiple suites.

neisw:
/override ci/prow/e2e-aws-ovn-microshift
/override ci/prow/e2e-aws-ovn-microshift-serial
/skip

passed previously failures unrelated

PR #30003, opened 2025-07-20
CNF-18661: add a test that should validate the communication matrix ports match the node's listening ports

aabughosh:
a test that we already have on our commatrix repo that we want to have it on origin test. that validate the communication matrix ports match the node's listening ports


shirmoran:
Looks good to me.

shirmoran:
Looks good to me.

dgoodwin:
/hold

This test should not go into origin for the reasons outlined in https://issues.redhat.com/browse/OCPBUGS-61996 which also includes a proposal for where the test should go.

aabughosh:
> /hold
> 
> This test should not go into origin for the reasons outlined in https://issues.redhat.com/browse/OCPBUGS-61996 which also includes a proposal for where the test should go.

It’s not the same case here. If this test fails, it means the owner of the newly added port needs to update their EndpointSlice. Unlike the documentation comparison, this test directly compares the ports exposed by ss with the actual EndpointSlice in Kubernetes to ensure every newly exposed port has a corresponding EndpointSlice.

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,237 @@
...
+	listeningsockets "github.com/openshift-kni/commatrix/pkg/listening-sockets"
+	matrixdiff "github.com/openshift-kni/commatrix/pkg/matrix-diff"
+	"github.com/openshift-kni/commatrix/pkg/utils"
+)
+
+const (
+	docTypeHolder           = "TYPE"
+	docVersionHolder        = "VERSION"
+	docReleaseTemplate      = "release-" + docVersionHolder
+	docCommatrixTemplateURL = "https://raw.githubusercontent.com/openshift-kni/commatrix/refs/heads/%s/docs/stable/raw/%s.csv"
```
dgoodwin: Is this unused? It looks like the same thing the other PR did that was causing problems, but it seems it may not be used at all?

test/extended/networking/commatrix.go
```diff
@@ -0,0 +1,237 @@
...
+	"strings"
+
+	"os"
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift-kni/commatrix/pkg/types"
```
dgoodwin, line 18: Commatrix vendoring, like the other pr we had to revert?

Is it possible that something goes wrong, and requires a fix in commatrix, then vendoring, to restore passing tests?

test/extended/networking/commatrix.go
aabughosh: I will remove my mistake 
Its from the last pr its not used

test/extended/networking/commatrix.go
aabughosh: Its not like the other tests?
If they want to update functionality then they need to update the vendor?

dgoodwin: It would be very rare for a failing test to require a fix elsewhere, then vendoring in origin.

neisw:
/override ci/prow/okd-scos-e2e-aws-ovn

unrelated failures occurring outside of this pr

neisw:
/override ci/prow/okd-scos-e2e-aws-ovn
/override ci/prow/e2e-aws-ovn-serial-2of2
/skip

serial has passed previously and okd is not required / failing outside of this pr

PR #30009, opened 2025-07-22
CORENET-6203: Enable test for LGW L2

arkadeepsen:
This PR enables the test for LGW L2 which was previously disabled.

### DIFF
--- a/test/extended/networking/route_advertisements.go
+++ b/test/extended/networking/route_advertisements.go
@@ -385,13 +385,6 @@ var _ = g.Describe("[sig-network][OCPFeatureGate:RouteAdvertisements][Feature:Ro
 
 				var err error
 				var snifferPodsNodes []string
-				// Check if the cluster is in local gateway mode
-				network, err := oc.AdminOperatorClient().OperatorV1().Networks().Get(context.TODO(), "cluster", metav1.GetOptions{})
-				o.Expect(err).NotTo(o.HaveOccurred())
-				if network.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig != nil && network.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig.RoutingViaHost && testCUDNTopology == "layer2" {
-					// TODO: unskip once CORENET-5881 is done.
-					skipper.Skipf("Skipping Layer2 UDN advertisements test for local gateway mode")
-				}
 				if testCUDNTopology == "layer2" {
 					// Running the packet sniffer on all nodes in the cluster for Layer2 UDN
 					nodes, err := clientset.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
@@ -469,7 +462,6 @@ var _ = g.Describe("[sig-network][OCPFeatureGate:RouteAdvertisements][Feature:Ro
 				err := runOcWithRetryIgnoreOutput(oc.AsAdmin(), "delete", "deploy", deployName)
 				errors.Join(err, runOcWithRetryIgnoreOutput(oc.AsAdmin(), "delete", "pod", "--all"))
 				errors.Join(err, runOcWithRetryIgnoreOutput(oc.AsAdmin().WithoutNamespace(), "delete", "ra", testCUDNName))
-				errors.Join(err, runOcWithRetryIgnoreOutput(oc.AsAdmin().WithoutNamespace(), "delete", "ra", testCUDNName))
 				errors.Join(err, runOcWithRetryIgnoreOutput(oc.AsAdmin().WithoutNamespace(), "delete", "clusteruserdefinednetwork", testCUDNName))
 				errors.Join(err, deallocateSubnets(oc, userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet))
 				cleanup()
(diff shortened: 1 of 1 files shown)


jcaamano:
> This PR enables the test for LGW L2 which was previously disabled. Increased the timeout from 30m to 45m to account for the newly enabled test.

@arkadeepsen before merging this, we need to submit a change in some other repo to keep history after the test name change, as the timeout is part of the test name.

arkadeepsen:
> > This PR enables the test for LGW L2 which was previously disabled. Increased the timeout from 30m to 45m to account for the newly enabled test.
> 
> @arkadeepsen before merging this, we need to submit a change in some other repo to keep history after the test name change, as the timeout is part of the test name.

@jcaamano I've already created it here https://github.com/openshift-eng/ci-test-mapping/pull/370

jcaamano:
@arkadeepsen also we need this merged for the tests here to be able to pass
https://github.com/openshift/ovn-kubernetes/pull/2676

arkadeepsen:
> @arkadeepsen also we need this merged for the tests here to be able to pass [openshift/ovn-kubernetes#2676](https://github.com/openshift/ovn-kubernetes/pull/2676)

@jcaamano Ack. However, I think there is some issue with the sniffer daemonsets. The test is trying to deploy it twice without removing it. Either we should remove it and then deploy it again, OR, we should deploy it only once.

jcaamano:
/lgtm
/approve

/hold
for that downstream merge to happen, bgp test pass here and for the test rename to merge

/cc @kyrtapz 
can help with unforeseen L2 LGW BGP issues

jcaamano:
@arkadeepsen when [the time](https://github.com/openshift/origin/pull/30009#issuecomment-3107417402) comes you can unhold and ask for approval in #forum-ocp-release-oversight

jcaamano:
> > @arkadeepsen also we need this merged for the tests here to be able to pass [openshift/ovn-kubernetes#2676](https://github.com/openshift/ovn-kubernetes/pull/2676)
> 
> @jcaamano Ack. However, I think there is some issue with the sniffer daemonsets. The test is trying to deploy it twice without removing it. Either we should remove it and then deploy it again, OR, we should deploy it only once.

The VRF-Lite test case you mean?

jcaamano:
> > > @arkadeepsen also we need this merged for the tests here to be able to pass [openshift/ovn-kubernetes#2676](https://github.com/openshift/ovn-kubernetes/pull/2676)
> > 
> > 
> > @jcaamano Ack. However, I think there is some issue with the sniffer daemonsets. The test is trying to deploy it twice without removing it. Either we should remove it and then deploy it again, OR, we should deploy it only once.
> 
> The VRF-Lite test case you mean?

Yeah, for the VRF-Lite test case I think you are right. I thought I did a better job there :facepalm: 

So as we call `beforeEach` we should also call `afterEach`. 
Additionally, we would need to delete `snifferNamespace`.
Potentially, also delete either things within `targetNamespace`, or delete that namespace itself and replace it with another one tricking the e2e framework.

jcaamano:
/lgtm cancel

due to https://github.com/openshift/origin/pull/30009#issuecomment-3107852989

test/extended/networking/route_advertisements.go
```diff
@@ -472,6 +465,7 @@ var _ = g.Describe("[sig-network][OCPFeatureGate:RouteAdvertisements][Feature:Ro
 				errors.Join(err, runOcWithRetryIgnoreOutput(oc.AsAdmin().WithoutNamespace(), "delete", "ra", testCUDNName))
 				errors.Join(err, runOcWithRetryIgnoreOutput(oc.AsAdmin().WithoutNamespace(), "delete", "clusteruserdefinednetwork", testCUDNName))
 				errors.Join(err, deallocateSubnets(oc, userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet))
+				errors.Join(err, runOcWithRetryIgnoreOutput(oc.AsAdmin(), "delete", "ds", packetSnifferDaemonSet.Name, "-n", snifferNamespace))
```
jcaamano: I don't think you need to to this generally. packetSnifferDaemonSet is created with `SetupProject` which adds it to the framework so it will be deleted after each test case ends. 
We just need to do this in an afterEach specific to the VRF-Lite test case.

arkadeepsen:
> > > > @arkadeepsen also we need this merged for the tests here to be able to pass [openshift/ovn-kubernetes#2676](https://github.com/openshift/ovn-kubernetes/pull/2676)
> > > 
> > > 
> > > @jcaamano Ack. However, I think there is some issue with the sniffer daemonsets. The test is trying to deploy it twice without removing it. Either we should remove it and then deploy it again, OR, we should deploy it only once.
> > 
> > 
> > The VRF-Lite test case you mean?
> 
> Yeah, for the VRF-Lite test case I think you are right. I thought I did a better job there 🤦
> 
> So as we call `beforeEach` we should also call `afterEach`. Additionally, we would need to delete `snifferNamespace`. Potentially, also delete either things within `targetNamespace`, or delete that namespace itself and replace it with another one tricking the e2e framework.

Added a call to `afterEach` at the end of `test` function. Removed the packet sniffer daemonset in the `afterEach` function. This has fixed the issue being faced earlier. Will wait for https://github.com/openshift/ovn-kubernetes/pull/2676 to merge to get a green signal on the bgp test.

arkadeepsen:
https://github.com/openshift/ovn-kubernetes/pull/2676 is now merged.

/test e2e-metal-ipi-ovn-dualstack-bgp-local-gw

test/extended/networking/route_advertisements.go
arkadeepsen: Pushed the suggested changes.

PR #30040, opened 2025-07-31
WIP: oidc: improve testing

everettraven:


test/extended/authentication/keycloak_client.go
```diff
@@ -104,7 +104,7 @@ func (kc *keycloakClient) CreateUser(username, password string, groups ...string
 		Groups:        groups,
 		Credentials: []credential{
 			{
-				Temporary: true,
+				Temporary: false,
```
everettraven, line 107: Testing locally, this looked like it was causing issues where the users we created were considered to be "incomplete" because they needed to set a real password.

test/extended/authentication/keycloak_client.go
```diff
@@ -133,6 +133,8 @@ func (kc *keycloakClient) CreateUser(username, password string, groups ...string
 type authenticationResponse struct {
 	AccessToken string `json:"access_token"`
 	IDToken     string `json:"id_token"`
+	Error       string `json:"error,omitempty"`
+	ErrorDescription string `json:"error_description,omitempty"`
```
everettraven: This will allow us to surface errors from Keycloak as we encounter them during authentication

test/extended/authentication/oidc.go
```diff
@@ -495,7 +496,9 @@ func resetAuthentication(ctx context.Context, client *exutil.CLI, original *conf
 
 		_, err = cli.Update(ctx, current, metav1.UpdateOptions{})
 		if err != nil {
-			return false, err
+			// Only log the error so we continue to retry until the context has timed out
+			g.GinkgoLogr.Error(err, "updating authentication resource")
+			return false, nil
```
everettraven: I had forgetten in the original PR that returning an error here immediately stops the polling so we weren't actually retrying this operation on a 409. 

This should retry the request until the timeout of 5 minutes every 10 seconds and log the error

everettraven:
I tested these changes as best as I could against a clusterbot cluster, but I'd like to find a way to test these changes in a CI environment using the job configurations in https://github.com/openshift/release/pull/66980

I'm hoping I can merge those, even when failing, to be able to use the `/test` command here. If not, I'll have to merge this and wait for payload inclusion to use `/pj-rehearse` in the o/release PR.

Holding until I've got a better grasp on the direction I need to take.
/hold

test/extended/authentication/keycloak_helpers.go
```diff
@@ -351,9 +351,9 @@ func createKeycloakRoute(ctx context.Context, service *corev1.Service, client ty
 }
 
 func waitForKeycloakAvailable(ctx context.Context, client *exutil.CLI, namespace string) error {
-	timeoutCtx, cancel := context.WithDeadline(ctx, time.Now().Add(5*time.Minute))
+	timeoutCtx, cancel := context.WithDeadline(ctx, time.Now().Add(30*time.Minute))
```
sjenning: were we observing timeouts? how long does this normally take? we want timeouts to be as tight as possible while avoiding flakes.

everettraven: Good question - IIRC there were some timeouts related to deploying keycloak when doing the `/pj-rehearse` commands in https://github.com/openshift/release/pull/66980 (ref: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/66980/rehearse-66980-periodic-ci-openshift-cluster-authentication-operator-release-4.20-periodics-e2e-gcp-external-oidc/1950583529473576960) but looking at the most recent run in CI we aren't even coming close to this timeout.

I can tighten this to 10 minutes to have a bit more buffer because we clearly can end up going beyond 5 minutes, but I'd be surprised if we go beyond 10 minutes.

everettraven:
Required failures seem to be infra/known regression related. Removing hold.

/hold cancel

PR #30048, opened 2025-08-01
OCPBUGS-60066, OCPBUGS-60067: Tolerate various expected Upgradeable=False conditions

sdodson:
- RHEL8 worker support was removed in 4.19, therefore 4.18 clsuters have machine-config operator Upgradeable=False with reason RHELNodes
- Cgroup v1 support was removed in 4.19, therefore 4.18 clusters have machine-config operator upgradeable=false with reason ClusterOnCgroupV1

Applying these changes directly to release-4.18 branch as future branches do not have jobs that install these configurations and attempt to upgrade.


sdodson:
On top of #30047 but once I figure out if we can merge directly into release-4.18 or need to backport will decide if I do one PR for two bugs or not

PR #30054, opened 2025-08-04
MON-4290: add test for must-gather gather_metrics

machine424:
needs https://github.com/openshift/must-gather/pull/498

### DIFF
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -338,7 +338,7 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 		})
 	})
 
-	g.It("runs successfully for metrics gathering", func() {
+	g.It("runs successfully for metrics gathering [apigroup:config.openshift.io]", func() {
 		tempDir, err := os.MkdirTemp("", "test.oc-adm-must-gather.")
 		o.Expect(err).NotTo(o.HaveOccurred())
 		defer os.RemoveAll(tempDir)
--- a/test/extended/cli/mustgather.go
+++ b/test/extended/cli/mustgather.go
@@ -337,6 +337,50 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
 			})
 		})
 	})
+
+	g.It("runs successfully for metrics gathering", func() {
+		tempDir, err := os.MkdirTemp("", "test.oc-adm-must-gather.")
+		o.Expect(err).NotTo(o.HaveOccurred())
+		defer os.RemoveAll(tempDir)
+
+		args := []string{
+			"--dest-dir", tempDir,
+			"--",
+			"/usr/bin/gather_metrics",
+			fmt.Sprintf("--min-time=%d", time.Now().Add(-5*time.Minute).UnixMilli()),
+			fmt.Sprintf("--max-time=%d", time.Now().UnixMilli()),
+			"--match=prometheus_ready",
+			"--match=prometheus_build_info",
+		}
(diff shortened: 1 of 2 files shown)


machine424:
/payload-with-prs 4.20 nightly blocking openshift/must-gather#498

test/extended/cli/mustgather.go
```diff
@@ -337,6 +337,50 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+			"/usr/bin/gather_metrics",
+			fmt.Sprintf("--min-time=%d", time.Now().Add(-5*time.Minute).UnixMilli()),
+			fmt.Sprintf("--max-time=%d", time.Now().UnixMilli()),
+			"--match=prometheus_ready",
+			"--match=prometheus_build_info",
+		}
+		o.Expect(oc.Run("adm", "must-gather").Args(args...).Execute()).To(o.Succeed())
+
+		// wait for the contents to show up in the plugin output directory
+		time.Sleep(5 * time.Second)
```
simonpasquier, line 358: This might not be very robust but I see that the same pattern is used in the other tests.

test/extended/cli/mustgather.go
```diff
@@ -337,6 +337,50 @@ var _ = g.Describe("[sig-cli] oc adm must-gather", func() {
...
+		o.Expect(errorContent).To(o.BeEmpty())
+
+		// The metrics file should contain some series with a given format
+		o.Expect(metricsFile).To(o.BeAnExistingFile())
+		metrics, err := os.ReadFile(metricsFile)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		lines := strings.Split(strings.TrimSpace(string(metrics)), "\n")
+		count := len(lines)
+		o.Expect(count).To(o.BeNumerically(">=", 5))
```
simonpasquier, line 377: Can we add a comment on why we expect at least 5 lines?

machine424: It just a sanity check, a number off the top of my head, to be sure it's not empty.
given prom scrape interval, for a range of 5mins, we should have at least 5 series in there.
but probably it's only dumping one metric of the two, it's not a correctness check per se...

test/extended/cli/mustgather.go
machine424: we could add a wait check based on "I finished" must-gather signal (I think it writes that into a file somewhere).
But yes, this should be enough, also we can see it as a check that the sump doesn't take more than 5 seconds...

test/extended/cli/mustgather.go
simonpasquier: should we be checking that we have at least 3 lines (1 for prometheus_ready, 1 for prometheus_build_info and 1 EOF) and that both prometheus_ready and prometheus_build_info are found?

machine424: I added two matchers just to be sure we can always set multiple ones.
We can always improve checks regarding correctness, probably in follow-up PRs.
i can make some changes in this one, if I have a cluster around.

simonpasquier: if the short-term goal is to ensure that the output file contains something and we improve the verification in a follow-up PR, I'm fine with just checking that the file exists and isn't empty.

machine424:
maybe I should wait a little bit more until the fix is taken into account, let's see.

machine424:
Had to make a change

    require [apigroup:config.openshift.io] so the test doesn't run on
    microshift as "oc adm must-gather" isn't supported there.
    
    this was done for other tests in https://github.com/openshift/origin/pull/28015

PR #30056, opened 2025-08-05
OLMv1: Skip check to verify maxOpenshiftVersion block upgrades due missing valid scenario

camilamacedo86:


camilamacedo86:
/hold

We have this PR done as latest resource

camilamacedo86:
we will not need this one

PR #30058, opened 2025-08-05
NO-JIRA: [TNF] add Two Node Fencing exception to accept less than two etcd endpoints

clobrano:
In a Two Node Fencing cluster is it acceptable to have less than two etcd endpoints.

### DIFF
--- a/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
+++ b/pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
@@ -499,6 +499,9 @@ func NewUniversalPathologicalEventMatchers(kubeConfig *rest.Config, finalInterva
 	newCrioReloadedTooOftenEventMatcher := newCrioReloadedTooOftenEventMatcher(finalIntervals)
 	registry.AddPathologicalEventMatcherOrDie(newCrioReloadedTooOftenEventMatcher)
 
+	twoNodeEtcdEndpointsMatcher := newTwoNodeEtcdEndpointsConfigMissingEventMatcher(finalIntervals)
+	registry.AddPathologicalEventMatcherOrDie(twoNodeEtcdEndpointsMatcher)
+
 	return registry
 }
 
@@ -713,6 +716,20 @@ var KubeAPIServerAvoids500s = &SimplePathologicalEventMatcher{
 var CertificateRotation = &SimplePathologicalEventMatcher{
 	name:               "CertificateRotation",
 	messageReasonRegex: regexp.MustCompile(`^(CABundleUpdateRequired|SignerUpdateRequired|TargetUpdateRequired|CertificateUpdated|CertificateRemoved|CertificateUpdateFailed|CSRCreated|CSRApproved|CertificateRotationStarted|ClientCertificateCreated|NoValidCertificateFound)$`),
+}
+
+func newTwoNodeEtcdEndpointsConfigMissingEventMatcher(finalIntervals monitorapi.Intervals) EventMatcher {
+	// Use topology-based detection for DualReplica (two-node fencing) clusters
+	dualReplicaTopology := v1.DualReplicaTopologyMode
+	return &SimplePathologicalEventMatcher{
+		name: "EtcdEndpointsConfigMissingDuringTwoNodeTests",
+		locatorKeyRegexes: map[monitorapi.LocatorKey]*regexp.Regexp{
+			monitorapi.LocatorNamespaceKey: regexp.MustCompile(`^openshift-kube-apiserver-operator$`),
+		},
+		messageReasonRegex: regexp.MustCompile(`^ConfigMissing$`),
+		messageHumanRegex:  regexp.MustCompile(`apiServerArguments\.etcd-servers has less than two live etcd endpoints`),
+		topology:           &dualReplicaTopology,
+	}
 }
 
 // IsEventAfterInstallation returns true if the monitorEvent represents an event that happened after installation.


pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -156,6 +168,15 @@ func isSingleNodeCheck(clientConfig *rest.Config) (bool, error) {
 	return exutil.IsSingleNode(context.Background(), configClient)
 }
 
+func isTwoNodeFencingCheck(clientConfig *rest.Config) (bool, error) {
+	configClient, err := clientconfigv1.NewForConfig(clientConfig)
+	if err != nil {
+		logrus.WithError(err).Error("Error creating config client to check for Single Node configuration")
```
jaypoulz: this error message is misleading :D

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -52,6 +52,12 @@ func testStableSystemOperatorStateTransitions(events monitorapi.Intervals, clien
 		isSingleNode = false
 	}
 
+	isTwoNodeFencing, err := isTwoNodeFencingCheck(clientConfig)
+	if err != nil {
+		logrus.Warnf("Error checking for TwoNodeFencing Node configuration on upgrade (unable to make exception): %v", err)
```
jaypoulz: I would probably rephrase this:
"Error checking for DualReplica controlPlaneTopology (i.e. Two Node OpenShift with Fencing)"

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
```diff
@@ -156,6 +168,15 @@ func isSingleNodeCheck(clientConfig *rest.Config) (bool, error) {
 	return exutil.IsSingleNode(context.Background(), configClient)
 }
 
+func isTwoNodeFencingCheck(clientConfig *rest.Config) (bool, error) {
```
eggfoobar: Since we're already printing the error here and not doing anything with it when it's bubbled up, I would just return a bool here and just print out the error, then simplify the call site.

clobrano: I agree. In this case, however, I intentionally mirrored the signature of `isSingleNodeCheck`. Would you like to change that function as well, for consistency?

https://github.com/openshift/origin/blob/b55b947e8d271ccc94f4589b42f6d3b39e6f1822/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go#L162-L169

clobrano:
I am not sure this change is actually working

/hold

pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go
eggfoobar: Oh yeah, that would be great, thank you!

clobrano:
The original change wasn't effective. I was looking at condition Message, but the pattern wasn't there. It seemed to work, because the test failure is not 100% reproducible.

The change was moved to the library looking for duplicated pathological events, and activated only if topology is DualReplica

Sorry for the initial wasted reviews, but now the PR is ready again :bow: 

/unhold

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -715,6 +718,20 @@ var CertificateRotation = &SimplePathologicalEventMatcher{
 	messageReasonRegex: regexp.MustCompile(`^(CABundleUpdateRequired|SignerUpdateRequired|TargetUpdateRequired|CertificateUpdated|CertificateRemoved|CertificateUpdateFailed|CSRCreated|CSRApproved|CertificateRotationStarted|ClientCertificateCreated|NoValidCertificateFound)$`),
 }
 
+func newTwoNodeEtcdEndpointsConfigMissingEventMatcher(finalIntervals monitorapi.Intervals) EventMatcher {
+	// Use topology-based detection for DualReplica (two-node fencing) clusters
+	dualReplicaTopology := v1.DualReplicaTopologyMode
+	return &SimplePathologicalEventMatcher{
+		name: "EtcdEndpointsConfigMissingDuringTwoNodeTests",
+		locatorKeyRegexes: map[monitorapi.LocatorKey]*regexp.Regexp{
+			monitorapi.LocatorNamespaceKey: regexp.MustCompile(`^openshift-kube-apiserver-operator$`),
```
xueqzhan, line 727: Does this apply to openshift-apiserver and oauth-apiserver too?

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
```diff
@@ -715,6 +718,20 @@ var CertificateRotation = &SimplePathologicalEventMatcher{
...
+		},
+		messageReasonRegex: regexp.MustCompile(`^ConfigMissing$`),
+		messageHumanRegex:  regexp.MustCompile(`apiServerArguments\.etcd-servers has less than two live etcd endpoints`),
+		topology:           &dualReplicaTopology,
```
xueqzhan, line 731: Is single node filtered out too?

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
clobrano: Good question. I only observed events from `openshift-kube-apiserver-operator`, do you think `oauth-apiserver` might incurr in the same error?

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
clobrano: I think this comes from APIServer knowing there is more than 1 endpoint during normal lifecycle, and 1 is a degraded state, but not failure. So I'd say no, only DualReplica.

pkg/monitortestlibrary/pathologicaleventlibrary/duplicated_event_patterns.go
xueqzhan: No, I am not sure. Just wondered if you have seen the same message with those since openshift and oauth apiserver interact with etcd similarly to kube-apiserver. But if you haven't seen it, it probably doesn't.

PR #30060, opened 2025-08-05
NO-JIRA: Update image mirroring instructions to avoid using external binary

bertinatto:
/assign @sjenning


Phaow:
/retitle NO-JIRA: Update image mirroring instructions to avoid using external binary

PR #30062, opened 2025-08-05
OCPBUGS-57032: Add wait.Poll retry logic to checkUpgradeability with 30s timeout

sdodson:
Based on Damien's analysis the test is failing because there wasn't enough time between when the upgrade completed and the DaemonSet pod to become ready. In the cases he reviewed there were single digit seconds between when the test checked and when pods became ready.

In https://github.com/openshift/origin/pull/29960 Damien added 30s sleep but Devan suggested a polling loop instead to limit downtime.

### DIFF
--- a/test/e2e/upgrade/upgrade.go
+++ b/test/e2e/upgrade/upgrade.go
@@ -195,7 +195,18 @@ var _ = g.Describe("[sig-arch][Feature:ClusterUpgrade]", func() {
 		config, err := framework.LoadConfig()
 		framework.ExpectNoError(err)
 		client := configv1client.NewForConfigOrDie(config)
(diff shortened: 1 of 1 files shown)


sdodson:
/uncc @deads2k @sjenning 
/cc @dgrisonnet @dgoodwin 
/retitle OCPBUGS-57032: Add wait.Poll retry logic to checkUpgradeability with 30s timeout

dgrisonnet (approved):
Thank you for following up on that Scott, I sadly didn't get time to look into my original PR again.

I believe this change will improve the situation. 

/lgtm
/approve

sdodson:
This job passed previously but fails now and has been failing for most PRs today.

PR #30063, opened 2025-08-05
OCPBUGS-60154: Update olmv1.go cluster-logging to 6.3.0

bentito:
cluster-logging 6.2.2 set `maxOCPVersion` <= 4.20 so hopefully this bump to 6.3.0 will help

### DIFF
--- a/test/extended/olm/olmv1.go
+++ b/test/extended/olm/olmv1.go
@@ -109,7 +109,7 @@ var _ = g.Describe("[sig-olmv1][OCPFeatureGate:NewOLM][Skipped:Disconnected] OLM
 
 		const (
 			packageName = "cluster-logging"
-			version     = "6.2.2"
+			version     = "6.3.0"
 		)
 
 		cleanup, unique := applyResourceFile(oc, packageName, version, "", ceFile)


test/extended/olm/olmv1.go
```diff
@@ -109,7 +109,7 @@ var _ = g.Describe("[sig-olmv1][OCPFeatureGate:NewOLM][Skipped:Disconnected] OLM
 
 		const (
 			packageName = "cluster-logging"
-			version     = "6.2.2"
+			version     = "6.3.0"
```
camilamacedo86, line 112: Up to 3 hours ago, we did not have this content for index 4.20.
Do we have it now?

camilamacedo86, line 112: Cluster Logging and its version do not exist in 4.20
We need a scenario that works for 4.20 and also for 4.19 :-( 
So that we can move forward and bump 4.20 catalogs

bentito:
This is not the droid you're looking for.

More precisely, the actual problem is not some version of cluster-logging operator that has a better value for maxOCPVersion, it's to find an operator that has maxOCPVersion and with a good value setting maxOCPVersion <= 4.20. 

Closing this since the approach in this PR, I now understand, can't work.

camilamacedo86:
Hi @bentito 

lets keep open if they add we can merge this one instead
What do you think about adding the OCBUG as well in the title, since we know it is related to that one?

bentito:
opened back up

PR #30092, opened 2025-08-11
Remove Perl 5.32 on UBI8

aroyoredhat:
We don't need to test this image as we are removing it here https://github.com/openshift/cluster-samples-operator/pull/646


jcpowermac:
not related to the pr, more the failure with registry s3 bucket

/test e2e-vsphere-ovn

Issue #2074, opened 2015-05-05
[RFE] pod status of "pending" not very useful


thoraxe:
"Pending" can have lots of different meanings. One, for example, is that the underlying host is doing a docker pull of the image. But "osc get pods" doesn't tell you that.

Is it possible to add "more" status to the Pending line? For example: Pending:Pulling 

I'm not sure what other pending states there are.

ncdc:
Definitely a good request, although I assume this will need to go to Kubernetes upstream

thoraxe:
Darnit, sorry.

thoraxe:
https://github.com/GoogleCloudPlatform/kubernetes/issues/7784

PR #30093, opened 2025-08-12
OCPNODE-3557: oc debug managed by label is checked by e2e test

ngopalak-redhat:
**Test for new oc node debug label**
This PR adds an e2e test to verify the functionality of the new oc node debug label. This test has been developed to address a bug fix. The test validates two scenarios:

- **With the keep-labels argument:** Ensures that the oc node debug command correctly adds the new label while preserving all existing labels on the node.

- **Without the keep-labels argument:** Confirms that the command adds labels as expected when the argument is not used.

This test is part of a larger effort to ensure the identification of the pods created by oc node debug command, as documented in the bug fix PR: https://github.com/openshift/oc/pull/2074.

Note: A commented line in the test will be uncommented once the fix in PR #2074 has been merged. This will enable the full verification of the corrected behavior.


ngopalak-redhat:
cc @BhargaviGudi

test/extended/cli/debug.go
```diff
@@ -26,6 +27,7 @@ var _ = g.Describe("[sig-cli] oc debug", func() {
 	defer g.GinkgoRecover()
 
 	oc := exutil.NewCLIWithPodSecurityLevel("oc-debug", admissionapi.LevelBaseline)
+	ocNodeClient := exutil.NewCLIWithPodSecurityLevel("cli", admissionapi.LevelBaseline)
```
ardaguclu: Why not `admissionapi.LevelPrivileged`?

test/extended/cli/debug.go
```diff
@@ -278,4 +280,83 @@ spec:
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(out).To(o.ContainSubstring("image: quay.io/wildfly/wildfly-centos7"))
 	})
+
+	g.It("ensure that the label is set for node debug", func() {
+		var err error
+
+		ns := ocNodeClient.Namespace()
+
+		err = ocNodeClient.AsAdmin().Run("label").Args("namespace", ns, "pod-security.kubernetes.io/enforce=privileged", "pod-security.kubernetes.io/audit=privileged", "--overwrite").Execute()
```
ardaguclu: Labeling node may cause some unexpected issues for some tests. Why do we need this for our test case?

test/extended/cli/debug.go
```diff
@@ -278,4 +280,83 @@ spec:
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		err = ocNodeClient.AsAdmin().Run("auth").Args("can-i", "get", "nodes").Execute()
```
ardaguclu: I think, this execution exercises different behavior. We don't need this in here (we'll get access failures anyway, if we don't have sufficient privilege).

test/extended/cli/debug.go
```diff
@@ -278,4 +280,83 @@ spec:
...
+		o.Expect(err).NotTo(o.HaveOccurred(), "User should be able to get nodes")
+
+		nodesOutput, err := ocNodeClient.AsAdmin().Run("get").Args("nodes", "-l", "node-role.kubernetes.io/worker=", "--no-headers", "-o", "custom-columns=NAME:.metadata.name,STATUS:.status.conditions[?(@.type=='Ready')].status").Output()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(nodesOutput).NotTo(o.BeEmpty(), "No worker nodes found")
+
+		var readyWorkerNode string
+		for _, line := range strings.Split(strings.TrimSpace(nodesOutput), "\n") {
```
ardaguclu: Isn't there any simpler way to filter the ready nodes and pick the first one?

test/extended/cli/debug.go
```diff
@@ -278,4 +280,83 @@ spec:
...
+			fields := strings.Fields(line)
+			if len(fields) >= 2 && fields[1] == "True" {
+				readyWorkerNode = fields[0]
+				break
+			}
+		}
+		o.Expect(readyWorkerNode).NotTo(o.BeEmpty(), "No ready worker node found")
+
+		done := make(chan error, 1)
+		go func() {
```
ardaguclu: Any failure in these tests bring about a release blocker bug that needs to be investigated. Running some test cases in goroutines make investigation harder. I don't think, we need goroutines. You can run debug node by passing this flag https://github.com/openshift/oc/blob/0f09f6e9cc7fdaab8942ad118d0717a6d527746b/pkg/cli/debug/debug.go#L264.

test/extended/cli/debug.go
```diff
@@ -278,4 +280,83 @@ spec:
...
+			}
+			return len(pods.Items) >= 1, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred(), "Expected at least one debug pod with label debug.openshift.io/managed-by=oc-debug")
+
+		debugErr := <-done
+		o.Expect(debugErr).NotTo(o.HaveOccurred())
+
+		done = make(chan error, 1)
+		go func() {
```
ardaguclu: same. We don't need goroutines. These can be managed in a sequential order.

test/extended/cli/debug.go
ngopalak-redhat: Thanks, I've made sure not to label the node itself. Instead, the test labels a new namespace that's created specifically for the test run and is deleted afterward.

The reason for this is to create the privileged pod required by oc node debug. Since we can't safely use the default namespace in parallel tests, I used the same temporary namespace where the test is running to avoid any conflicts.

ardaguclu: ah, you are right. I missed you are labeling namespace not node.

test/extended/cli/debug.go
ngopalak-redhat: The tests above this one don't need "get" on cluster scoped resources. This test requires "nodes" which is a cluster scoped resource. Hence making sure it has the access before calling "get".

test/extended/cli/debug.go
ngopalak-redhat: Removed go-routines

test/extended/cli/debug.go
ngopalak-redhat: Used the flag as suggested

test/extended/cli/debug.go
ngopalak-redhat: Tried several ways. JsonPath is not supported fully as documented here : https://kubernetes.io/docs/reference/kubectl/jsonpath/ (See last section in the page)
Current method worked

test/extended/cli/debug.go
ngopalak-redhat: Addressed it. I spent sometime learning about the arguments used and got rid of ocNodeClient.

ngopalak-redhat:
cc @haircommander

test/extended/cli/debug.go
haircommander: you could go programatically:
```
oc.AsAdmin().KubeClient().CoreV1().Nodes().List(context.Background(), metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/worker"})
```
and go through the node status conditions?

haircommander: generally, anyway we can use the api instead of parsing strings, the better IMO

ngopalak-redhat: Thanks. I have adopted the API

test/extended/cli/debug.go
```diff
@@ -278,4 +279,54 @@ spec:
...
+
+		err = oc.AsAdmin().Run("debug").Args("node/"+readyWorkerNode, "--keep-labels=true", "--preserve-pod=true", "--", "sleep", "1").Execute()
+		pods, err := oc.AdminKubeClient().CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(pods.Items).To(o.HaveLen(1))
+		o.Expect(pods.Items[0].Labels).To(o.HaveKeyWithValue("debug.openshift.io/managed-by", "oc-debug"))
+		oc.AsAdmin().Run("delete").Args("pod", pods.Items[0].Name, "-n", ns).Output()
+
+		// Make sure the pod is deleted
+		pods, err = oc.AdminKubeClient().CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
```
ardaguclu: This causes flaky behavior. Because deletion operation sometimes takes time. I'd recommend using waitpoll until ensuring that the pod is deleted.

test/extended/cli/debug.go
```diff
@@ -278,4 +279,54 @@ spec:
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(pods.Items).To(o.HaveLen(0))
+
+		err = oc.AsAdmin().Run("debug").Args("node/"+readyWorkerNode, "--preserve-pod=true", "--", "sleep", "1").Execute()
+
+		// Tests the code fix in https://github.com/openshift/oc/pull/2074
+		o.Expect(err).NotTo(o.HaveOccurred())
+		pods, err = oc.AdminKubeClient().CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
+		o.Expect(pods.Items).To(o.HaveLen(1))
+		o.Expect(pods.Items[0].Labels).To(o.HaveKeyWithValue("debug.openshift.io/managed-by", "oc-debug"))
```
ardaguclu, line 338: It is better to check that there is no other label exists.

test/extended/cli/debug.go
```diff
@@ -278,4 +279,54 @@ spec:
...
+		oc.AsAdmin().Run("delete").Args("pod", pods.Items[0].Name, "-n", ns).Output()
```
ardaguclu: I don't think we need deletion in here?. When NS is garbage collected, this pod will be removed anyway.

test/extended/cli/debug.go
ngopalak-redhat: Thanks. Added the check

test/extended/cli/debug.go
ngopalak-redhat: Thanks, added poll wait

test/extended/cli/debug.go
```diff
@@ -278,4 +279,67 @@ spec:
...
+			}
+		}
+		o.Expect(readyWorkerNode).NotTo(o.BeEmpty(), "No ready worker node found")
+
+		err = oc.AsAdmin().Run("debug").Args("node/"+readyWorkerNode, "--keep-labels=true", "--preserve-pod=true", "--", "sleep", "1").Execute()
+		pods, err := oc.AdminKubeClient().CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(pods.Items).To(o.HaveLen(1))
+		o.Expect(pods.Items[0].Labels).To(o.HaveKeyWithValue("debug.openshift.io/managed-by", "oc-debug"))
+		o.Expect(pods.Items[0].Labels).To(o.HaveLen(1))
```
ardaguclu, line 316: when --keep-labels=true is passed, I think we can't guarantee that the label length is 1.

test/extended/cli/debug.go
```diff
@@ -278,4 +279,67 @@ spec:
...
+
+		oc.AsAdmin().Run("delete").Args("pod", pods.Items[0].Name, "-n", ns).Output()
```
ardaguclu, line 318: Why do we delete this one too?

test/extended/cli/debug.go
ngopalak-redhat: For this test case its always 1

test/extended/cli/debug.go
ngopalak-redhat: Similar pod is created with the same label after this one when we test without the keep-labels. Hence deleting this one so that we don't land up fetching the old pod

test/extended/cli/debug.go
ngopalak-redhat: Removed the last deletion

test/extended/cli/debug.go
ardaguclu: I think, we get the pod's live object already. We can get the length of labels and;

```go
o.Expect(pods.Items[0].Labels).To(o.HaveLen(numberoflabels + 1))
```

ardaguclu: Because are we sure that our test suite in sometime start adding some labels to all the pods created?.

ngopalak-redhat: This is true when we use "keep-labels" with "oc debug podname", here its "oc debug node/" so we don't inherit any pod labels from other pods.

ardaguclu: I mean in the future, if new functionality applies labels to the nodes. This test will start failing. In my opinion this change would increase the resiliency of the test.

test/extended/cli/debug.go
```diff
@@ -278,4 +279,67 @@ spec:
...
+			pods, err := oc.AdminKubeClient().CoreV1().Pods(ns).List(ctx, metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
+			if err != nil {
+				return false, err
+			}
+			return len(pods.Items) == 0, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred(), "Expected debug pod to be deleted")
+
+		// Make sure the pod is deleted
+		pods, err = oc.AdminKubeClient().CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
```
ardaguclu: In the PollUntilContextTimeout, we have already verified that there is no other pod. Why do we additionally listing here?

ngopalak-redhat: My apologies, I thought I had removed that and added the new poll, but I unintentionally left it there. It's gone now.

test/extended/cli/debug.go
ngopalak-redhat: The doc says:
```
    --keep-labels=false:
	If true, keep the original pod labels
```
Adding labels to nodes won't affect this. This flag implies that we retain the labels of pod we are debugging and not the node labels. In this case we are debugging a node directly and hence there are no labels getting added.

ardaguclu: ah, yes you are right. In any case, there will be only one label https://github.com/openshift/oc/blob/c5c63346aa2af4b88e2f7757caf4c49cf33238e5/pkg/cli/debug/debug.go#L1066

ardaguclu:
This looks good to me. Before adding `lgtm`, I'll wait the results of the CI runs. Once I'll add lgtm, you can ping TRT for approval.

Thank you

test/extended/cli/debug.go
```diff
@@ -278,4 +279,62 @@ spec:
...
+				if nodeStatus.Type == corev1.NodeReady && nodeStatus.Status == corev1.ConditionTrue {
+					readyWorkerNode = node.Name
+					break
+				}
+			}
+		}
+		o.Expect(readyWorkerNode).NotTo(o.BeEmpty(), "No ready worker node found")
+
+		err = oc.AsAdmin().Run("debug").Args("node/"+readyWorkerNode, "--keep-labels=true", "--preserve-pod=true", "--", "sleep", "1").Execute()
+		pods, err := oc.AdminKubeClient().CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
```
ardaguclu, line 312: This still assumes that the creation of the pod will happen instantly which may contradict with the reality. Maybe it is better to WaitPoll here as well to at least retry a couple of times to prevent any flakiness. WDYT?

test/extended/cli/debug.go
```diff
@@ -278,4 +279,62 @@ spec:
...
+			}
+			return len(pods.Items) == 0, nil
+		})
+		o.Expect(err).NotTo(o.HaveOccurred(), "Expected debug pod to be deleted")
+
+		err = oc.AsAdmin().Run("debug").Args("node/"+readyWorkerNode, "--preserve-pod=true", "--", "sleep", "1").Execute()
+
+		// Tests the code fix in https://github.com/openshift/oc/pull/2074
+		o.Expect(err).NotTo(o.HaveOccurred())
+		pods, err = oc.AdminKubeClient().CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{LabelSelector: "debug.openshift.io/managed-by=oc-debug"})
```
ardaguclu, line 336: Same WaitPoll to prevent flaky behavior. WDYT?

test/extended/cli/debug.go
ngopalak-redhat: The oc debug and kubectl debug commands are designed to be blocking (without the interactive and terminal sessions attached). The command will not complete, and the script will not proceed to the next line, until the pod has been successfully created and executed the command. In this case, the sleep command simply delays the completion of the debug session, not the pod creation itself.

test/extended/cli/debug.go
ngopalak-redhat: Same as https://github.com/openshift/origin/pull/30093#discussion_r2278044137

test/extended/cli/debug.go
ardaguclu: You are right. That is correct.

ardaguclu:
Thank you
/lgtm
I think, you need to attach Jira ticket and an approval from TRT

ngopalak-redhat:
microshift tests failed with this change and hence couldn't merge. I have added an image to the debug pod now to fix the issue

ngopalak-redhat:
@ardaguclu I had to make a change for microshift platform. Added an image that is used by other tests. Initially tried with "ShellImage()" which didn't work. So used "agnhost" which worked across all platforms.
Please review

ardaguclu:
/hold
I think, it is better to merge this after release cut

ngopalak-redhat:
@ardaguclu Is it a good time to merge this?

ngopalak-redhat:
@ardaguclu Sorry, I thought this merged after your review, but needed more approvals. Do you know who can give "acknowledge-critical-fixes-only, verified" approvals?

ngopalak-redhat:
> @ardaguclu Sorry, I thought this merged after your review, but needed more approvals. Do you know who can give "acknowledge-critical-fixes-only, verified" approvals?

Thanks @ardaguclu . But the CI kept failing overnight and hence it didn't merge. I have rebased with lastest main, let's see if the tests pass now.

ardaguclu:
/verified by "oc debug ensure that the label is set for node debug"

test/extended/cli/debug.go
jacobsee: @ardaguclu @ngopalak-redhat  I am running into an issue in the 1.35 rebase as a result of this - please consider accepting the removal of this check: https://github.com/openshift/origin/pull/30576

ngopalak-redhat: Reviewed #30576

PR #30110, opened 2025-08-13
OCPEDGE-1483: Add TNF E2E tests for network failure

clobrano:
This commit introduces a new e2e test for TNF to validate etcd recovery after a network disruption in a two-node cluster.

The test simulates a network partition by using iptables to block traffic between the two nodes for a short, configurable, duration.

It then verifies that the cluster correctly handles the split-brain scenario, with one node becoming the leader and the other rejoining as a learner before being promoted back to a voting member.

Additionally, the node disruption pod creation logic has been refactored from `triggerNodeReboot` to a more generic `createNodeDisruptionPod` to accommodate different types of disruptions, including network partitions and reboots.

### DIFF
--- a/test/extended/two_node/tnf_recovery.go
+++ b/test/extended/two_node/tnf_recovery.go
@@ -26,6 +26,7 @@ const (
 	memberIsLeaderTimeout        = 10 * time.Minute
 	memberRejoinedLearnerTimeout = 10 * time.Minute
 	memberPromotedVotingTimeout  = 15 * time.Minute
+	networkDisruptionDuration    = 15 * time.Second
 	pollInterval                 = 5 * time.Second
 )
 
@@ -33,9 +34,9 @@ var _ = g.Describe("[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:Dual
 	defer g.GinkgoRecover()
 
 	var (
-		oc                       = util.NewCLIWithoutNamespace("").AsAdmin()
-		etcdClientFactory        *helpers.EtcdClientFactoryImpl
-		survivedNode, targetNode corev1.Node
+		oc                   = util.NewCLIWithoutNamespace("").AsAdmin()
+		etcdClientFactory    *helpers.EtcdClientFactoryImpl
+		peerNode, targetNode corev1.Node
 	)
 
 	g.BeforeEach(func() {
@@ -52,25 +53,29 @@ var _ = g.Describe("[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:Dual
 
 		// Select the first index randomly
 		randomIndex := rand.Intn(len(nodes.Items))
-		survivedNode = nodes.Items[randomIndex]
+		peerNode = nodes.Items[randomIndex]
 		// Select the remaining index
 		targetNode = nodes.Items[(randomIndex+1)%len(nodes.Items)]
-		g.GinkgoT().Printf("Randomly selected %s (%s) to be shut down and %s (%s) to take the lead\n", targetNode.Name, targetNode.Status.Addresses[0].Address, survivedNode.Name, survivedNode.Status.Addresses[0].Address)
 
 		kubeClient := oc.KubeClient()
 		etcdClientFactory = helpers.NewEtcdClientFactory(kubeClient)
 
 		g.GinkgoT().Printf("Ensure both nodes are healthy before starting the test\n")
 		o.Eventually(func() error {
-			return helpers.EnsureHealthyMember(g.GinkgoT(), etcdClientFactory, survivedNode.Name)
-		}, nodeIsHealthyTimeout, pollInterval).ShouldNot(o.HaveOccurred(), "expect to ensure Node A healthy without error")
+			return helpers.EnsureHealthyMember(g.GinkgoT(), etcdClientFactory, peerNode.Name)
+		}, nodeIsHealthyTimeout, pollInterval).ShouldNot(o.HaveOccurred(), fmt.Sprintf("expect to ensure Node '%s' healthiness without errors", peerNode.Name))
 
 		o.Eventually(func() error {
 			return helpers.EnsureHealthyMember(g.GinkgoT(), etcdClientFactory, targetNode.Name)
-		}, nodeIsHealthyTimeout, pollInterval).ShouldNot(o.HaveOccurred(), "expect to ensure Node B healthy without error")
+		}, nodeIsHealthyTimeout, pollInterval).ShouldNot(o.HaveOccurred(), fmt.Sprintf("expect to ensure Node '%s' healthiness without errors", targetNode.Name))
 	})
 
-	g.It("Should recover from graceful node shutdown with etcd member re-addition", func() {
+	g.It("should recover from graceful node shutdown with etcd member re-addition", func() {
+		// Note: In graceful shutdown, the targetNode is deliberately shut down while
+		// the peerNode remains running and becomes the etcd leader.
+		survivedNode := peerNode
+		g.GinkgoT().Printf("Randomly selected %s (%s) to be shut down and %s (%s) to take the lead\n",
+			targetNode.Name, targetNode.Status.Addresses[0].Address, peerNode.Name, peerNode.Status.Addresses[0].Address)
 		g.By(fmt.Sprintf("Shutting down %s gracefully in 1 minute", targetNode.Name))
 		err := util.TriggerNodeRebootGraceful(oc.KubeClient(), targetNode.Name)
 		o.Expect(err).To(o.BeNil(), "Expected to gracefully shutdown the node without errors")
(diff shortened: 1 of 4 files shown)


clobrano:
This shall be merged after #29986 

/hold

clobrano:
#29986 merged, this PR is ready for review

/unhold

test/extended/util/nodes.go
```diff
@@ -95,19 +95,30 @@ func DebugAllNodesRetryWithOptionsAndChroot(oc *CLI, debugNodeNamespace string,
...
+	command_preable := fmt.Sprintf("echo 'temporarily disabling interface %s on %s for %v'; exec chroot /host sh -c ", networkInterface, target.Name, disruptionDuration)
+	interface_down_cmd := fmt.Sprintf("ip link set %s down", networkInterface)
+	timeout_cmd := fmt.Sprintf("sleep %d", int(disruptionDuration.Seconds()))
+	interface_up_cmd := fmt.Sprintf("ip link set %s up", networkInterface)
+	command := fmt.Sprintf("%s '%s ; %s ; %s'", command_preable, interface_down_cmd, timeout_cmd, interface_up_cmd)
+
+	return command, triggerNodeDisruption(kubeClient, target.Name, 0, true, command)
+}
+
+func triggerNodeDisruption(kubeClient kubernetes.Interface, nodeName string, attempt int, useHostNetwork bool, command string) error {
```
jaypoulz: I don't love the `userHostNetwork` bool.
I would prefer some kind of enum that allows me to specify which kind of disruption.
-> NodeCrash
-> NetworkPartition
etc.

I think this will allow us to reuse the pattern as we add more disruption types (e.g. kubelet failure, etc)

test/extended/util/nodes.go
```diff
@@ -116,6 +127,7 @@ func triggerNodeReboot(kubeClient kubernetes.Interface, nodeName string, attempt
 			},
 		},
 		Spec: corev1.PodSpec{
+			HostNetwork:   useHostNetwork,
```
jaypoulz: this podspec is still called "reboot", which is confusing

zz_generated.manifests/test-reporting.yaml
```diff
@@ -193,6 +193,9 @@ spec:
     - testName: '[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:DualReplica][Suite:openshift/two-node][Disruptive]
         Two Node with Fencing etcd recovery Should recover from graceful node shutdown
         with etcd member re-addition'
+    - testName: '[sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:DualReplica][Suite:openshift/two-node][Disruptive]
```
jaypoulz, line 196: This is super nitpicky, but i feel like the naming would be cleaner if it was:
```
Two Node with Fencing etcd **s**hould recover from network disruption with etcd member re-addition
 ```

File: test/extended/util/nodes.go (hunk 1 above)
Comment (clobrano) [reply test/extended/util/nodes.go]: I understand; that boolean isn't my preference either. This function isn't exposed, though. Instead, you should use it through other functions like TriggerNetworkDisruption, TriggerNodeRebootGraceful, and TriggerNodeRebootUngraceful. This approach seems similar to one based on enums.

For example, to handle a kubelet disruption, I'd create a new TriggerKubeletDisruption function, which ensures a clear separation of concerns. As the list of supported disruptions expands, a switch statement based on an enum could become too large.

File: test/extended/two_node/tnf_recovery.go [hunk 4]
```diff
@@ -251,7 +299,76 @@ func validateEtcdRecoveryState(e *helpers.EtcdClientFactoryImpl, survivedNode *c
...
+
+		// This might happen if the disruption didn't occurred yet, or (unlikely) we get this snapshot when the learner has been already promoted
+		if !learnerA && !learnerB {
+			return fmt.Errorf("both nodes are non-learners (should have exactly one learner): %s(started=%v, learner=%v), %s(started=%v, learner=%v)",
+				nodeA.Name, startedA, learnerA, nodeB.Name, startedB, learnerB)
+		}
+
+		// Once we get one leader and one learner, we don't care if the latter has started already, but the first must
+		// already been started
+		nonLearnerStarted := (startedA && !learnerA) || (startedB && !learnerB)
```
jaypoulz: Can we call this variable votingMemberStarted to avoid the double negative?

jaypoulz: or leaderStarted

test/extended/util/nodes.go
jaypoulz: I get what you mean. Maybe a name tweak would suffice. Instead of "triggerNodeDisruption", could we rename the function to something like: "createNodeDisruptionPod" or "execNodeDisruptionPod"

This would make it clear to that usingHostNetwork is a change that affects the created pod, not something that dictates the kind of disruption. (The kind of disruption is defined by the command / all of the arguments together).

clobrano: Good idea :+1:

clobrano:
Addressed the comments, moreover I changed the disruption command to something more specific that limits _only_ the traffic between the two nodes and not the entire interface.

clobrano:
/pj-rehearse periodic-ci-openshift-release-master-nightly-4.20-e2e-metal-ovn-two-node-fencing-recovery-techpreview

eggfoobar:
/payload-job periodic-ci-openshift-release-master-nightly-4.20-e2e-metal-ovn-two-node-fencing-recovery-techpreview

test/extended/util/nodes.go
```diff
@@ -95,19 +105,43 @@ func DebugAllNodesRetryWithOptionsAndChroot(oc *CLI, debugNodeNamespace string,
...
+	preambleCmd := fmt.Sprintf("echo 'temporarily disabling network connection between %s and %s for %v'; exec chroot /host sh -c ", target.Name, peer.Name, disruptionDuration)
+
+	peerIP := getNodeInternalAddress(peer)
+
+	blockTrafficCmd := fmt.Sprintf("sudo iptables -I INPUT -j DROP -s %s && sudo iptables -I OUTPUT -j DROP -d %s", peerIP, peerIP)
+	cleanupCmd := fmt.Sprintf("sudo iptables -D INPUT -j DROP -s %s; sudo iptables -D OUTPUT -j DROP -d %s", peerIP, peerIP)
+	sleepCmd := fmt.Sprintf("sleep %d", int(disruptionDuration.Seconds()))
+	disruptionCmd := fmt.Sprintf("%s 'trap \"%s\" EXIT; %s ; %s'", preambleCmd, cleanupCmd, blockTrafficCmd, sleepCmd)
+
+	return disruptionCmd, createNodeDisruptionPod(kubeClient, target.Name, 0, hostNetwork, disruptionCmd)
```
jaypoulz: Very clever!

jaypoulz (approved):
/lgtm

jaypoulz:
/unhold

This looks good to me. A few more follow up things:
1. I saw the test run successfully in a rehearsal, so that's great. I put a hold on this because I wanted to make sure that was done, but Egli beat us to it :)
2. You will need an approval from someone in #forum-ocp-release-oversight

clobrano:
@xueqzhan, who can add the "verified" label?

test/extended/util/nodes.go
```diff
@@ -14,6 +14,16 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+// hostNetworkMode represents the networking mode for disruption pods
+type hostNetworkMode int
```
eggfoobar: Small suggestion here for naming consistency, since we have hostNetwork and podNetwork, the type is en enum on the mode you wish to execute, something like this might be better here.

```go
// networkMode represents the networking mode for disruption pods
type networkMode int

const (
	// hostNetworkMode enables host networking for the disruption pod
	hostNetworkMode networkMode = iota
	// podNetworkMode disables host networking for the disruption pod
	podNetworkMode
)
```

test/extended/util/nodes.go
```diff
@@ -95,19 +105,43 @@ func DebugAllNodesRetryWithOptionsAndChroot(oc *CLI, debugNodeNamespace string,
...
+		if addr.Type == corev1.NodeInternalIP {
+			return addr.Address
+		}
+	}
+	// fallback
+	return node.Status.Addresses[0].Address
 }
 
-func triggerNodeReboot(kubeClient kubernetes.Interface, nodeName string, attempt int, command string) error {
+func createNodeDisruptionPod(kubeClient kubernetes.Interface, nodeName string, attempt int, networkMode hostNetworkMode, command string) error {
```
eggfoobar: since this function is wearing many hats, we should update it the name to reflect it's core use case, maybe something like `runCmdPod`?

eggfoobar (commented):
Just some small clarity things, looking good.

test/extended/util/nodes.go
clobrano: the function does create the node disruption pod, so the name seems pretty accurate to me

eggfoobar: Oh is not also being used to trigger reboots?

clobrano: This function creates a pod with the goal to execute a given command, which might disrupt the network or reboot the node, either gracefully or ungracefully.

clobrano: but the function gets the command as input

eggfoobar: Right exactly, that's my point, the function is not creating a disruption pod or a reboot pod, it's creating a privileged pod with a passed in command. The calling function `TriggerNodeRebootGraceful` or `TriggerNetworkDisruption` is what decides what command to execute, the function is just a host command runner.

lol However, it also causes disruption so I understand it can go either way, It's not worth holding back this PR any longer, thanks man!

eggfoobar:
/payload-job periodic-ci-openshift-release-master-nightly-4.20-e2e-metal-ovn-two-node-fencing-recovery-techpreview

test/extended/util/nodes.go
clobrano: Yeah, you're also right :grin: 
Let's say that I like to keep the "disruptive" name in this function just because it is the way we use it now :)

Thanks!

clobrano:
/verified by [sig-etcd][apigroup:config.openshift.io][OCPFeatureGate:DualReplica][Suite:openshift/two-node][Disruptive] Two Node with Fencing etcd recovery should recover from network disruption with etcd member re-addition [Serial]

PR #30134, opened 2025-08-19
SREP-1622: Fix endpoint admission test namespace creation for ROSA

Amarthya-v:
Check if namespace exists before creation to prevent unnecessary webhook calls while maintaining same test behavior.

Fixes: [SREP-1622](https://issues.redhat.com//browse/SREP-1622)

### DIFF
--- a/test/extended/networking/endpoint_admission.go
+++ b/test/extended/networking/endpoint_admission.go
@@ -186,9 +186,15 @@ func testOneEndpointSlice(oc *exutil.CLI, client kubernetes.Interface, addrType,
 }
 
 func getClientForServiceAccount(adminClient kubernetes.Interface, clientConfig *rest.Config, namespace, name string) (*kubernetes.Clientset, *rest.Config, error) {
-	_, err := adminClient.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}, metav1.CreateOptions{})
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return nil, nil, err
+	_, err := adminClient.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, nil, err
+		}
+		_, err = adminClient.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}, metav1.CreateOptions{})
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 
 	_, err = adminClient.CoreV1().ServiceAccounts(namespace).Create(context.Background(), &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: name}}, metav1.CreateOptions{})


arghosh93:
/retitle OCPBUGS-60654: Fix endpoint admission test namespace creation for ROSA

test/extended/networking/endpoint_admission.go
```diff
@@ -186,9 +186,16 @@ func testOneEndpointSlice(oc *exutil.CLI, client kubernetes.Interface, addrType,
 }
 
 func getClientForServiceAccount(adminClient kubernetes.Interface, clientConfig *rest.Config, namespace, name string) (*kubernetes.Clientset, *rest.Config, error) {
-	_, err := adminClient.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}, metav1.CreateOptions{})
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return nil, nil, err
+	_, err := adminClient.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			_, err = adminClient.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}, metav1.CreateOptions{})
```
arghosh93: If the webhook does not allow creating a namespace, then I would expect this to also fail if the namespace does not exist by that time. 
However I think kube-system namespace should always be there. Are you counting on that fact?

Amarthya-v: @arghosh93  Yes, this fix specifically addresses the `kube-system` namespace case, which Always exists in OpenShift/Kubernetes clusters and is being unnecessarily created by the test, triggering the ROSA webhook
now it gets checked first before attempting creation.

For test-specific namespaces that don't exist, the create path still works as before. The webhook only blocks creation in managed namespaces like `kube-system`.

The failing ROSA jobs were specifically hitting this error when trying to create `kube-system`:

'''admission webhook "namespace-validation.managed.openshift.io" denied the request:
Prevented from accessing Red Hat managed namespaces

This fix avoids that unnecessary webhook call entirely for existing namespaces.'''

So the fix targets the specific scenario causing ROSA CI failures without breaking the general functionality.

kyrtapz:
@Amarthya-v why is the bug closed?

test/extended/networking/endpoint_admission.go
```diff
@@ -186,9 +186,16 @@ func testOneEndpointSlice(oc *exutil.CLI, client kubernetes.Interface, addrType,
...
+			if err != nil {
+				return nil, nil, err
+			}
+		} else {
```
kyrtapz: NIT: there is no need for the `else` here.
To decrease the nestedness maybe we could go with:
```
	if err != nil {
		if !errors.IsNotFound(err) {
		    return nil, nil, err
	    }
		_, err = adminClient.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}, metav1.CreateOptions{})
		if err != nil {
			return nil, nil, err
		}
	}

```

test/extended/networking/endpoint_admission.go
```diff
@@ -210,4 +217,4 @@ func getClientForServiceAccount(adminClient kubernetes.Interface, clientConfig *
 	}
 
 	return kubeClientset, saClientConfig, nil
-}
```
kyrtapz: NIT: missing newline

Amarthya-v:
> @Amarthya-v why is the bug closed?

The bug was closed because Trevor reverted the webhook temporarily to unblock CI. 

This PR is the permanent fix - once merged, Trevor can safely re-enable the webhook without breaking ROSA CI.

It's part of a coordinated plan: revert → fix test → re-enable webhook.

test/extended/networking/endpoint_admission.go
Amarthya-v: @kyrtapz Updated! Much cleaner now. Thanks for the suggestion.

test/extended/networking/endpoint_admission.go
Amarthya-v: Added the missing newline. thanks!

kyrtapz:
> > @Amarthya-v why is the bug closed?
> 
> The bug was closed because Trevor reverted the webhook temporarily to unblock CI.
> 
> This PR is the permanent fix - once merged, Trevor can safely re-enable the webhook without breaking ROSA CI.
> 
> It's part of a coordinated plan: revert → fix test → re-enable webhook.

You need to address the `jira/invalid-bug` label on the PR.
https://github.com/openshift/origin/pull/30134#issuecomment-3204908322

Amarthya-v:
> > > @Amarthya-v why is the bug closed?
> > 
> > 
> > The bug was closed because Trevor reverted the webhook temporarily to unblock CI.
> > This PR is the permanent fix - once merged, Trevor can safely re-enable the webhook without breaking ROSA CI.
> > It's part of a coordinated plan: revert → fix test → re-enable webhook.
> 
> You need to address the `jira/invalid-bug` label on the PR. [#30134 (comment)](https://github.com/openshift/origin/pull/30134#issuecomment-3204908322)

feedback addressed and Jira issue resolved.

Amarthya-v:
> /lgtm

Thanks Patryk!! the failure in [ci/prow/e2e-vsphere-ovn](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/30134/pull-ci-openshift-origin-main-e2e-vsphere-ovn/1958896839969214464) My PR doesn’t modify networking, and I see the same failure in other jobs today. This looks like infra/kernel flake rather than my change. Requesting test override.

PR #30144, opened 2025-08-20
OCPBUGS-60244: Wait 30s after the upgrade is complete to validate workloads

sdodson:
This accounts for the fact that it occasionally takes 1-2s for workloads to become ready after the nodes are marked ready at the end of an upgrade.


dinhxuanvu:
@sdodson Ready for merge :)

PR #30146, opened 2025-08-21
NO-JIRA: Stop setting feature gate conditional

petr-muller:
https://github.com/openshift/oc/pull/2063 promoted `oc adm upgrade status` and removed the conditional check of the `OC_ENABLE_CMD_UPGRADE_STATUS` envvar, so the tests should no longer set it either.


petr-muller:
/hold

I'll want to inspect the results to be sure

petr-muller:
I have inspected the jobs, the tests continue to work as expected

/hold cancel

PR #30148, opened 2025-08-21
CNTRLPLANE-945: oidc/keycloak: make keycloak client use environment proxy for HTTP requests

everettraven:
We encountered an issue in https://github.com/openshift/release/pull/68209 where requests to configure the Keycloak instance deployed for the OIDC tests failed to resolve the hostname of route for the Keycloak instance.

This appeared to be because all requests to routes on a baremetal cluster must go through a proxy. This PR updates the HTTP client we use for interacting with Keycloak instances to respect any proxies configured from the typical `HTTP_PROXY` `HTTPS_PROXY` and `NO_PROXY` environment variables.

### DIFF
--- a/test/extended/authentication/keycloak_client.go
+++ b/test/extended/authentication/keycloak_client.go
@@ -31,6 +31,7 @@ func keycloakClientFor(keycloakURL string) (*keycloakClient, error) {
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: true,
 		},
+		Proxy: http.ProxyFromEnvironment,
 	}
 
 	return &keycloakClient{


everettraven:
Tested this locally on a metal cluster and we were able to successfully configure keycloak through the keycloak route when the `HTTP_PROXY` and `HTTPS_PROXY` environment variables were set.

everettraven:
https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-authentication-operator-778-openshift-origin-30148-e2e-metal-ovn-ipv4-external-oidc-configure-techpreview/1958596586057502720 is evidence that  this change fixes the issue we were encountering in the metal jobs in https://github.com/openshift/release/pull/68209

All the feature-gated tests that use this functionality passed against a baremetal cluster, meaning we were successfully able to hit the Keycloak instances and configure them.

everettraven:
This suite is currently only run in tech preview jobs, and shouldn't negatively impact CI/CR signal or the default product.

PR #30152, opened 2025-08-21
TRT-2260: Create High CPU intervals for nodes

xueqzhan:
Assisted by cursor

### DIFF
--- a/e2echart/e2e-chart-template.html
+++ b/e2echart/e2e-chart-template.html
@@ -180,6 +180,10 @@
         return eventInterval.source === "Alert"
     }
 
+    function isHighCPU(eventInterval) {
+      return eventInterval.source === "CPUMonitor" && eventInterval.message.reason === "HighCPUUsage"
+    }
+
     function pathologicalEvents(item) {
         if (item.message.annotations["pathological"] === "true") {
             if (item.message.annotations["interesting"] === "true") {
@@ -490,6 +494,9 @@
         timelineGroups[timelineGroups.length - 1].data.sort(function (e1 ,e2){
             return e1.label < e2.label ? -1 : e1.label > e2.label;
         })
+
+        timelineGroups.push({group: "high-cpu", data: []})
+        createTimelineData("HighCPU", timelineGroups[timelineGroups.length - 1].data, eventIntervals, isHighCPU, regex)
 
         timelineGroups.push({group: "disruption", data: []})
         createTimelineData(disruptionValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isEndpointConnectivity, regex)
--- a/test/extended/testdata/bindata.go
+++ b/test/extended/testdata/bindata.go
@@ -57245,6 +57245,10 @@ var _e2echartE2eChartTemplateHtml = []byte(`<html lang="en">
         return eventInterval.source === "Alert"
     }
 
+    function isHighCPU(eventInterval) {
+      return eventInterval.source === "CPUMonitor" && eventInterval.message.reason === "HighCPUUsage"
+    }
+
     function pathologicalEvents(item) {
         if (item.message.annotations["pathological"] === "true") {
             if (item.message.annotations["interesting"] === "true") {
@@ -57555,6 +57559,9 @@ var _e2echartE2eChartTemplateHtml = []byte(`<html lang="en">
         timelineGroups[timelineGroups.length - 1].data.sort(function (e1 ,e2){
             return e1.label < e2.label ? -1 : e1.label > e2.label;
         })
+
+        timelineGroups.push({group: "high-cpu", data: []})
+        createTimelineData("HighCPU", timelineGroups[timelineGroups.length - 1].data, eventIntervals, isHighCPU, regex)
 
         timelineGroups.push({group: "disruption", data: []})
         createTimelineData(disruptionValue, timelineGroups[timelineGroups.length - 1].data, eventIntervals, isEndpointConnectivity, regex)


pkg/monitortests/node/highcpumetriccollector/monitortest.go
```diff
@@ -0,0 +1,212 @@
...
+		return intervals, err
+	}
+
+	timeRange := prometheusv1.Range{
+		Start: startTime,
+		End:   time.Now(),
+		Step:  30 * time.Second, // Sample every 30 seconds for better granularity
+	}
+
+	// Query for CPU usage percentage per instance
```
kannon92: We have an existing monitor test that detects if metrics are down. 

I think generally when we hit high CPU load, the metrics become unreliable or even down.

Should we consider that if the metrics are down we emit an interval here?

kannon92: I'm off VPN but I'd line the bug I am slowly looking at around MetricsEndPointDown which is flaky on over 75% of our tests. So I think this happens quite often so wanted to bring this up.

kannon92:
Drilling into one of these jobs:

https://sippy.dptools.openshift.org/sippy-ng/job_runs/1958624156723449856/pull-ci-openshift-origin-main-e2e-gcp-ovn-techpreview/openshift_origin/30152/intervals?end=2025-08-21T23%3A30%3A54Z&filterText=&intervalFile=e2e-events_20250821-222003.json&overrideDisplayFlag=0&selectedSources=E2EFailed&selectedSources=APIServerGracefulShutdown&selectedSources=NodeState&selectedSources=MetricsEndpointDown&selectedSources=NodeMonitor&start=2025-08-21T21%3A39%3A13Z

I think this is working correctly.

It took me a little time to figure out it is in NodeMonitor.

pkg/monitortests/node/highcpumetriccollector/monitortest.go
xueqzhan: Thanks Kevin for the review. We want to clearly correlate high CPU with e2e tests. So having this interval created directly from the metric is preferred.

xueqzhan:
> Drilling into one of these jobs:
> 
> https://sippy.dptools.openshift.org/sippy-ng/job_runs/1958624156723449856/pull-ci-openshift-origin-main-e2e-gcp-ovn-techpreview/openshift_origin/30152/intervals?end=2025-08-21T23%3A30%3A54Z&filterText=&intervalFile=e2e-events_20250821-222003.json&overrideDisplayFlag=0&selectedSources=E2EFailed&selectedSources=APIServerGracefulShutdown&selectedSources=NodeState&selectedSources=MetricsEndpointDown&selectedSources=NodeMonitor&start=2025-08-21T21%3A39%3A13Z
> 
> I think this is working correctly.
> 
> It took me a little time to figure out it is in NodeMonitor.

We decided to put this in its own source section.

pkg/monitortests/node/highcpumetriccollector/monitortest.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	routeClient, err := routeclient.NewForConfig(restConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = kubeClient.CoreV1().Namespaces().Get(ctx, "openshift-monitoring", metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return []monitorapi.Interval{}, nil
+	}
```
sosiouxme: no need to handle any other errors?

pkg/monitortests/node/highcpumetriccollector/monitortest.go
```diff
@@ -0,0 +1,212 @@
...
+
+	prometheusClient, err := metrics.NewPrometheusClient(ctx, kubeClient, routeClient)
+	if err != nil {
+		return nil, err
+	}
+
+	intervals, err := prometheus.EnsureThanosQueriersConnectedToPromSidecars(ctx, prometheusClient)
+	if err != nil {
+		return intervals, err
+	}
```
sosiouxme: would be clearer to me that the intervals only matter if there is an error:
```suggestion
	if intervals, err := prometheus.EnsureThanosQueriersConnectedToPromSidecars(ctx, prometheusClient); err != nil {
		return intervals, err
	}
```

pkg/monitortests/node/highcpumetriccollector/monitortest.go
```diff
@@ -0,0 +1,212 @@
...
+
+			msg := monitorapi.NewMessage().
+				Reason(monitorapi.IntervalReason("HighCPUUsage")).
+				HumanMessage(fmt.Sprintf("CPU usage above %.1f%% threshold on instance %s", w.cpuThreshold, instance)).
+				WithAnnotation("cpu_threshold", fmt.Sprintf("%.1f", w.cpuThreshold))
+
+			intervalTmpl := monitorapi.NewInterval(monitorapi.SourceNodeMonitor, monitorapi.Warning).
+				Locator(lb).
+				Message(msg).
+				Display()
```
sosiouxme: it seems a little convoluted to build this up when the only values needed in `createCPUInterval` are the instance name and locator. it would be useful if you could directly use it to create an interval with an added interval, but it seems you can't do that? i would just pass the values directly...

pkg/monitortests/node/highcpumetriccollector/monitortest.go
```diff
@@ -0,0 +1,212 @@
...
+func (w *highCPUMetricCollector) EvaluateTestsFromConstructedIntervals(ctx context.Context, finalIntervals monitorapi.Intervals) ([]*junitapi.JUnitTestCase, error) {
+	logger := logrus.WithField("MonitorTest", "HighCPUMetricCollector")
+
+	// Filter for high CPU intervals
+	highCPUIntervals := finalIntervals.Filter(func(eventInterval monitorapi.Interval) bool {
+		return eventInterval.Source == monitorapi.SourceNodeMonitor &&
+			eventInterval.Message.Reason == monitorapi.IntervalReason("HighCPUUsage")
+	})
+
+	logger.Infof("collected %d high CPU intervals for analysis", len(highCPUIntervals))
```
sosiouxme: do we need to do this here because the collection might be called more than once? otherwise seems like it would be simpler to log at the time of collection.

pkg/monitortests/testframework/highcputestanalyzer/monitortest.go
sosiouxme, line 1: would be clearer to me if this could just be combined into `highcpumetriccollector` but i guess that may be more trouble than it's worth.

xueqzhan: This will be too much for this story.

pkg/monitortests/node/highcpumetriccollector/monitortest.go
xueqzhan: Blame any convoluted logic on cursor. Anything smart is from me. :)

dgoodwin:
Please include these in the spyglass intervals files we render by default, I think it's in the BelongsInSpyglass function. 

My vision for this was always that you could create new intervals and they'd just be visible, that's true if the user knows to load the huge "all" events file, but everyone typically works with the spyglass files. I almost wonder if that function should include everything with Display true, but that should be a separate PR for another day.

My weak attempt to correlate high cpu tests is picking up almost too much to be useful, but maybe it will help find something. I'm curious to see how this goes with https://github.com/openshift/origin/pull/30171 just merged. Will be interesting to see. 

Anyhow if we can get these into the default files I think this looks good to go, super high value Ken thank you, and I love the inclusion of the peak CPU.

xueqzhan:
> Please include these in the spyglass intervals files we render by default, I think it's in the BelongsInSpyglass function.
> 
> My vision for this was always that you could create new intervals and they'd just be visible, that's true if the user knows to load the huge "all" events file, but everyone typically works with the spyglass files. I almost wonder if that function should include everything with Display true, but that should be a separate PR for another day.
> 
> My weak attempt to correlate high cpu tests is picking up almost too much to be useful, but maybe it will help find something. I'm curious to see how this goes with #30171 just merged. Will be interesting to see.
> 
> Anyhow if we can get these into the default files I think this looks good to go, super high value Ken thank you, and I love the inclusion of the peak CPU.

That function is indeed allowing most by default. But that is the not the function to add a section to the spyglass view. Let's hope my latest commit work. :)

PR #30168, opened 2025-08-26
Restore retry for the flaky conntrack UDP test

pliurh:
A new test case, "[sig-network] Conntrack should be able to preserve UDP traffic when server pod cycles for a NodePort service", is added to the list of tests that are retried on failure. This is to mitigate the impact of a known flake in this test.

OCPBUGS-59496


pliurh:
This is a bug that we need to fix in ovn-k.

PR #30169, opened 2025-08-26
OCPBUGS-60881: Exclude RHEL nodes in verify that nodes have no unexpected reboots test

gpei:
RHEL nodes are not totally managed by the MCO, so there would be rebooting happened without rebootRequest event, so skipping RHEL node in this test.


gpei:
Incorrect branch was used, close it.

PR #30180, opened 2025-08-26
OCPBUGS-60881: extended/machines: add --no-pager to journalctl cmds

sdodson:
When trying to determine what's different between RHEL8 and RHEL9 I stumbled across the fact that if the terminal is narrow you'll get a unicode emdash which may be the extra character we're seeing. This assumes that for some reason the terminal in a pod running on RHEL8 is more narrow than that on RHEL9.

### DIFF
--- a/test/extended/machines/display_reboots_pod.yaml
+++ b/test/extended/machines/display_reboots_pod.yaml
@@ -11,7 +11,7 @@ spec:
     - command: ['/bin/bash', '-ec']
       args:
         - |
-          chroot /host-root journalctl --list-boots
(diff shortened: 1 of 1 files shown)


sdodson:
@gpei no idea how to test this, do you know how we could or should we just give it a shot?

sdodson:
If we can't come up with a way to test I'll add it to the other command as well and go ahead with the merge.

gpei:
@sdodson no idea how to test this, installer QE team never added e2e tests to the origin repo...

sdodson:
/retitle OCPBUGS-60881: extended/machines: add --no-pager to journalctl cmds

PR #30196, opened 2025-08-30
trt-2230: Add monitor test annotation

neisw:


dgoodwin:
This is great, helps us actually locate where the test is from a little more clearly, we can scan for how often these kill jobs, locate perma or heavily flaking mon tests nobody came back to deal with, and implications for the currently unusable flake on failure feature in comp readiness.

/lgtm

PR #30010, opened 2025-07-22
CORENET-6232: add PreconfiguredUDNAddresses FG tests

RamLavi:
This PR adds tests to check the PreconfiguredUDNAddresses FG:
- create VM with Preconfigured IPs, then restart.
- create VM with Preconfigured MAC, then restart.
- create VM with IPs and MAC, then migrate

PR #30197, opened 2025-09-01
CORENET-6363: Add PreconfiguredUDNAddresses duplicate IP detection tests

RamLavi:
This PR is a followup PR to https://github.com/openshift/origin/pull/30010, adding tests to check duplicate IP detection:
- create VM with Preconfigured IPs, assert you can't have IP collision

### DIFF
--- a/test/extended/networking/livemigration.go
+++ b/test/extended/networking/livemigration.go
@@ -559,14 +559,26 @@ func duplicateVM(cli *kubevirt.Client, vmNamespace, vmName string) {
 	Expect(json.Unmarshal([]byte(originalVMIRawAnnotations), &originalVMIAnnotations)).To(Succeed())
 
 	var vmiCreationOptions []kubevirt.Option
+	var vmiExpectations []func()
 	if requestedIPs, hasIPRequests := originalVMIAnnotations[kvIPRequestsAnnot]; hasIPRequests {
 		vmiCreationOptions = append(
 			vmiCreationOptions,
 			kubevirt.WithAnnotations(ipRequests(requestedIPs)),
 		)
+		vmiExpectations = append(vmiExpectations, func() {
+			waitForVMPodEventWithMessage(
+				cli,
+				vmNamespace,
+				duplicateVMName,
+				"IP is already allocated",
+				2*time.Minute,
+			)
+		})
 	}
 	Expect(cli.CreateVMIFromSpec(vmNamespace, duplicateVMName, vmiSpec, vmiCreationOptions...)).To(Succeed())
-	waitForVMPodEventWithMessage(cli, vmNamespace, duplicateVMName, "IP is already allocated", 2*time.Minute)
+	for _, expectation := range vmiExpectations {
+		expectation()
+	}
 }
 
 func waitForVMPodEventWithMessage(vmClient *kubevirt.Client, vmNamespace, vmName, expectedEventMessage string, timeout time.Duration) {


RamLavi:
/test e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview
running this as baseline, expecting it to fail both tests.

RamLavi:
> /test e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview running this as baseline, expecting it to fail both tests.

[failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/30197/pull-ci-openshift-origin-main-e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview/1962779023872364544) as expected

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2744

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2744

RamLavi:
> /testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview [openshift/ovn-kubernetes#2744](https://github.com/openshift/ovn-kubernetes/pull/2744)

[results](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-origin-30197-openshift-ovn-kubernetes-2744-e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview/1963304554246180864): ip detection passes, MAC detection currently fails.

test/extended/networking/livemigration.go
```diff
@@ -537,6 +570,75 @@ func verifyVMMAC(virtClient *kubevirt.Client, vmName, expectedMAC string) {
...
+		vmiExpectations = append(vmiExpectations, func() {
+			waitForVMPodEventWithMessage(
+				cli,
+				vmNamespace,
+				duplicateVMName,
+				"IP is already allocated",
+				2*time.Minute,
+			)
+		})
+	} else if hasMACAddressRequest(cli, vmName) {
```
maiqueb: don't we want to drop the "else" ? Otherwise, how will we be able to create a VM w/ MAC & IP requests ? 

As this stands, I think it'll get either MAC **or** (exclusive or) IP.

test/extended/networking/livemigration.go
```diff
@@ -537,6 +570,75 @@ func verifyVMMAC(virtClient *kubevirt.Client, vmName, expectedMAC string) {
...
+
+func waitForVMPodEventWithMessage(vmClient *kubevirt.Client, vmNamespace, vmName, expectedEventMessage string, timeout time.Duration) {
+	GinkgoHelper()
+	By(fmt.Sprintf("Waiting for event containing %q on VM %s/%s virt-launcher pod", expectedEventMessage, vmNamespace, vmName))
+
+	Eventually(func(g Gomega) []string {
+		const vmLabelKey = "vm.kubevirt.io/name"
+		podNames, err := vmClient.GetPodsByLabel(vmLabelKey, vmName)
+		g.Expect(err).NotTo(HaveOccurred(), "Failed to get pods by label %s=%s", vmLabelKey, vmName)
+		g.Expect(podNames).To(HaveLen(1), "Expected exactly one virt-launcher pod for VM %s/%s, but found %d pods: %v", vmNamespace, vmName, len(podNames), podNames)
```
maiqueb, line 592: question: don't we need to filter out some states ? Let's take the "migration" scenario into consideration; a completed migration will leave the original (src) pod as `completed`. Won't that pod be caught in this filter ? 

Do we want to address that scenario here ?

test/extended/networking/livemigration.go
RamLavi: I removed the MAC related code. We will reintroduce it when we add the MAC test.
Wanted to again clarify though - the pod will not err on both MAC and IP conflicts, it will return err on the first problem it finds and rollover.

maiqueb: yeah, still, that's the production code. 
The test, imho, should be prepared to work independently of that implementation detail. 

not a must, I won't insist further.

RamLavi:
[Changes](https://github.com/openshift/origin/compare/5e64ef96a91d4214b113fd8c882899cf238360ac..f14d9ef5e68cc23993cff9d5264347561097b91e): Remove code regarding MAC duplicate test.

test/extended/networking/livemigration.go
RamLavi: hmm let's pick up this thread in the follow up PR where we add the MAC dupliacate

test/extended/networking/livemigration.go
RamLavi: depends.. Right now we not looking for event of vm-pods during migration/restart, so I would keep things simple. 
Should the occasion arise (look for events during migration/restart/whatnot), we could change `GetPodsByLabel` --> `GetPodsByLabelAndPhase` ...

maiqueb: fair enough for me.

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2744

test/extended/networking/kubevirt/client.go
```diff
@@ -93,6 +95,65 @@ func (c *Client) GetJSONPath(resource, name, jsonPath string) (string, error) {
 	}
 	return strings.TrimSuffix(strings.TrimPrefix(output, `"`), `"`), nil
 }
+
+func (c *Client) GetPodsByLabel(labelKey, labelValue string) ([]string, error) {
+	output, err := c.oc.AsAdmin().Run("get").Args("pods", "-n", c.oc.Namespace(), "-l", fmt.Sprintf("%s=%s", labelKey, labelValue), "-o", "jsonpath={.items[*].metadata.name}").Output()
```
qinqon: You can use `-o name` instead of the jsonpath also do we have to care of running vs completed pods during migration we can end up returning more pods than needed?

test/extended/networking/kubevirt/client.go
```diff
@@ -93,6 +95,65 @@ func (c *Client) GetJSONPath(resource, name, jsonPath string) (string, error) {
...
+	}
+
+	for _, opt := range opts {
+		opt(newVMI)
+	}
+
+	newVMIYAML, err := yaml.Marshal(newVMI)
+	if err != nil {
+		return err
+	}
```
qinqon, line 157: Why don't we use Unstructured stuff for this ?

test/extended/networking/livemigration.go
```diff
@@ -537,6 +553,65 @@ func verifyVMMAC(virtClient *kubevirt.Client, vmName, expectedMAC string) {
...
+
+		virtLauncherPodName := podNames[0]
+		eventMessages, err := vmClient.GetEventsForPod(virtLauncherPodName)
+		g.Expect(err).NotTo(HaveOccurred(), "Failed to get events for pod %s", virtLauncherPodName)
```
qinqon, line 596: Are we sure we want to retry o error ?

test/extended/networking/livemigration.go
```diff
@@ -537,6 +553,65 @@ func verifyVMMAC(virtClient *kubevirt.Client, vmName, expectedMAC string) {
...
+
+		return eventMessages
+	}).WithPolling(time.Second).WithTimeout(timeout).Should(
+		ContainElements(strings.Split(expectedEventMessage, " ")),
```
qinqon: Why do we have to split `"IP is already allocated",` ?

qinqon (changes requested):

test/extended/networking/livemigration.go
maiqueb: because that's what's output by the `GetEventsForPod` function - a slice of strings, with each "word" a separate slice element. 

Maybe it can be corrected, but I just found it simpler to do it here.

maiqueb: For instance, the lane is failing right now on the MAC conflict tests because:
```
Expected to find an event containing "MAC address conflict detected:"
Expected
    <[]string | len:28, cap:28>: [
        "Successfully",
        "assigned",
        "e2e-network-segmentation-e2e-2361/virt-launcher-myvm-duplicate-btp47",
        "to",
        "worker-2",
        "failed",
        "to",
        "update",
        "pod",
        "e2e-network-segmentation-e2e-2361/virt-launcher-myvm-duplicate-btp47:",
        "failed",
        "to",
        "reserve",
        "MAC",
        "address",
        "\"02:0a:0b:0c:0d:10\"",
        "for",
        "owner",
        "\"e2e-network-segmentation-e2e-2361/myvm-duplicate\"",
        "on",
        "network",
        "attachment",
        "\"e2e-network-segmentation-e2e-2361/blue\":",
        "MAC",
        "address",
        "already",
        "in",
        "use",
    ]
to contain elements
    <[]string | len:4, cap:4>: ["MAC", "address", "conflict", "detected:"]
the missing elements were
    <[]string | len:2, cap:2>: ["conflict", "detected:"]}
```

@qinqon this shows you the output data.

@RamLavi I think it does not make sense to "fix" the tests since we have decided (out of band) to drop the MAC address related tests from this PR, and to later on push them.

qinqon: Don't split the string, we can end up with false possitives since the elemnts of the array can be at the testing array but a different places.

maiqueb: definitely. @RamLavi would you follow up on this one ? 

We need to avoid calling the `strings.Fields(...)` in the `GetEventsForPod` function.

test/extended/networking/kubevirt/client.go
RamLavi: >You can use -o name instead of the jsonpath

ACK

>also do we have to care of running vs completed pods during migration we can end up returning more pods than needed?

@maiqueb also raised this [here](https://github.com/openshift/origin/pull/30197#discussion_r2321228532) - I think that right now we should keep it simple (we only check restarted VMs, not migrated ones), but having said that - since both of you suggested it - if you still think it's needed to be added now - I'll do it. your call.

test/extended/networking/kubevirt/client.go
RamLavi: not sure I follow - can you elaborate?

test/extended/networking/livemigration.go
RamLavi: why not? if k8s service is down or something we should retry.. My reasoning is: it's not what's the test is about, we should try and get the events and fail only if the relevant even isn't there.

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2744

test/extended/networking/livemigration.go
RamLavi: well the thing was that jsonpath returns space separated words.. 
changed to using custom-columns

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2744

RamLavi:
[Change](https://github.com/openshift/origin/compare/f14d9ef5e68cc23993cff9d5264347561097b91e..f43feb841a9e6411a8e13b4034d6e7881cae0405): Address @qinqon 's review

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2744

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2744

RamLavi:
/testwith openshift/origin/main/e2e-metal-ipi-ovn-bgp-virt-dualstack-techpreview https://github.com/openshift/ovn-kubernetes/pull/2750

kyrtapz:
@RamLavi verify is failing: https://storage.googleapis.com/test-platform-results/pr-logs/pull/30197/pull-ci-openshift-origin-main-verify/1964652088877125632/build-log.txt

test/extended/networking/kubevirt/client.go
```diff
@@ -93,6 +95,81 @@ func (c *Client) GetJSONPath(resource, name, jsonPath string) (string, error) {
 	}
 	return strings.TrimSuffix(strings.TrimPrefix(output, `"`), `"`), nil
 }
+
+func (c *Client) GetPodsByLabel(labelKey, labelValue string) ([]string, error) {
+	output, err := c.oc.AsAdmin().Run("get").Args("pods", "-n", c.oc.Namespace(), "-l", fmt.Sprintf("%s=%s", labelKey, labelValue), "-o", "name").Output()
```
kyrtapz, line 100: NIT: Why do need an admin?
Why can't you use a lister?
```
c.oc.KubeClient().CoreV1().Pods("").List(...selector..)
```
it would mean that there is no need to parse the output.

test/extended/networking/kubevirt/client.go
```diff
@@ -93,6 +95,81 @@ func (c *Client) GetJSONPath(resource, name, jsonPath string) (string, error) {
...
+		if line != "" {
+			podName := strings.TrimPrefix(line, "pod/")
+			podNames = append(podNames, podName)
+		}
+	}
+	return podNames, nil
+}
+
+func (c *Client) GetEventsForPod(podName string) ([]string, error) {
+	output, err := c.oc.AsAdmin().Run("get").Args("events", "-n", c.oc.Namespace(), "--field-selector", fmt.Sprintf("involvedObject.name=%s,involvedObject.kind=Pod", podName), "-o", "custom-columns=MESSAGE:.message", "--no-headers").Output()
```
kyrtapz, line 120: NIT: same as above, this should work:
```
c.oc.KubeClient().CoreV1().Events("").List(...)
```

test/extended/networking/kubevirt/client.go
RamLavi: I was following the logic of former code. So in order to fix this properly, I will issue a follow-up PR to fix it in all client functions

test/extended/networking/kubevirt/client.go
RamLavi: will do in follow up PR

ormergi:
I think the PR miss test annotations updates, dont we need to add changes by `make updated`?

maiqueb:
> I think the PR miss test annotations updated, dont we need to add changes by `make updated`?

Well, I don't think these are used anymore. I just ran it, and nothing happened. But, in the process I managed to mess up the branch, and re-pushed here, effectively losing the LGTM tag. 

@kyrtapz could you take another look ?... Did I mess something up ?.. 

Really sorry.

PR #30200, opened 2025-09-02
Check for control-plane operator in hypershift for oauth test

dinhxuanvu:
In HyperShift env, cluster-authentication-operator doesn't run in management cluster. Instead, control-plane operator will handle authenication operations.

### DIFF
--- a/test/extended/oauth/requestheaders.go
+++ b/test/extended/oauth/requestheaders.go
@@ -66,8 +66,33 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
 		// In some rare cases, CAO might be damaged when entering this test. If it is - the results
 		// of this test might flaky. This check ensures that we capture such situation early and
 		// investigate why it wasn't ready before this test.
-		e2e.Logf("Ensuring CAO is available==True, progressing==False, degraded==False")
-		waitForAuthenticationProgressing(oc, configv1.ConditionFalse)
+		// If cluster is Hypershift, skip the test if it's specifically ARO HCP given CAO is not running
+		// in Hypershift management cluster
+		ctx := context.Background()
+		isHypershift, _ := exutil.IsHypershift(ctx, oc.AdminConfigClient())
+		if isHypershift {
+			_, hcpNamespace, err := exutil.GetHypershiftManagementClusterConfigAndNamespace()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			managementOC := exutil.NewHypershiftManagementCLI(hcpNamespace)
+			// List deployments with the correct label that actually exists on the deployment
+			deployments, err := managementOC.AdminKubeClient().AppsV1().Deployments(hcpNamespace).List(ctx, metav1.ListOptions{
+				LabelSelector: "hypershift.openshift.io/managed-by=control-plane-operator",
+			})
+			o.Expect(err).NotTo(o.HaveOccurred())
+			var isCPODeployment bool
+			for _, deployment := range deployments.Items {
+				// Look for the control-plane-operator container directly in the deployment spec
+				for _, container := range deployment.Spec.Template.Spec.Containers {
+					if container.Name == "control-plane-operator" {
+						isCPODeployment = true
+					}
+				}
+			}
+			o.Expect(isCPODeployment).To(o.Equal(true), "control-plane-operator deployment exists")
+		} else {
+			e2e.Logf("Ensuring CAO is available==True, progressing==False, degraded==False")
+			waitForAuthenticationProgressing(oc, configv1.ConditionFalse)
+		}
 
 		controlPlaneTopology, err := exutil.GetControlPlaneTopology(oc)
 		o.Expect(err).NotTo(o.HaveOccurred())


dinhxuanvu:
/payload-job periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-aws-ovn-conformance-serial

test/extended/oauth/requestheaders.go
```diff
@@ -66,8 +66,33 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
...
+			var isCPODeployment bool
+			for _, deployment := range deployments.Items {
+				// Look for the control-plane-operator container directly in the deployment spec
+				for _, container := range deployment.Spec.Template.Spec.Containers {
+					if container.Name == "control-plane-operator" {
+						isCPODeployment = true
+					}
+				}
+			}
+			o.Expect(isCPODeployment).To(o.Equal(true), "control-plane-operator deployment exists")
```
everettraven: Is checking existence of the CPO sufficient here?

It looks like on standalone we are intentionally looking for a signal that the cluster-authentication-operator is no longer progressing (presumably meaning the OAuth server is up and ready to accept requests). Is there an equivalent status condition or something we could use to identify that everything auth related is up and ready for connections?

dinhxuanvu: I'm simply testing out the CPO check to see if it works. From the initial look, this test may not even be compatible with HyperShift and may need to be skipped in HCP case. There is already a check in the code to skip on ExternalTopologyMode which is the case for HCP given CAO component doesn't run on management cluster. The topology check is placed after CAO availability check and that's why the test fails. That check should have been placed before CAO check.

dinhxuanvu: The CPO did work as the test ultimately got skipped when it reached then ExternalTopologyMode check and no longer failed at CAO check. I think it is a reasonable solution.

dinhxuanvu:
@everettraven Hey Bryce, I think this PR is good to go unless you have any further comments. PTAL. Thanks.

test/extended/oauth/requestheaders.go
```diff
@@ -66,8 +66,33 @@ var _ = g.Describe("[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [I
...
+		ctx := context.Background()
+		isHypershift, _ := exutil.IsHypershift(ctx, oc.AdminConfigClient())
+		if isHypershift {
+			_, hcpNamespace, err := exutil.GetHypershiftManagementClusterConfigAndNamespace()
+			o.Expect(err).NotTo(o.HaveOccurred())
+			managementOC := exutil.NewHypershiftManagementCLI(hcpNamespace)
+			// List deployments with the correct label that actually exists on the deployment
+			deployments, err := managementOC.AdminKubeClient().AppsV1().Deployments(hcpNamespace).List(ctx, metav1.ListOptions{
+				LabelSelector: "hypershift.openshift.io/managed-by=control-plane-operator",
+			})
```
everettraven: In practice, how many deployments does this return?

I'm wondering if we could reasonably do just a `Get()` instead of a `List()` here?

test/extended/oauth/requestheaders.go
everettraven: If we skip this test when on HCP due to the control plane topology, should we just move this check below the topology check?

Is it always the case that when it is a HyperShift cluster that the control plane topology is the external topology mode?

test/extended/oauth/requestheaders.go
dinhxuanvu: There are a few including CPO, cluster-autoscaler, openshift-route-controller-manager and other controllers that are deployed by CPO. So `List` is an appropriate use here.

test/extended/oauth/requestheaders.go
dinhxuanvu: I can move the topology check to the front and skip this test entirely for external mode regardless if it is HyperShift or not. However, I thought given we delegate the auth control to CPO in HCP, it may be good to check for CPO availability in HyperShift case and let the current logic flow remaining the same. If you think it is better just to skip entirely for external mode in all cases, I'm fine with that.

Yes, for HCP, it is external mode for control plane topology: https://github.com/openshift/hypershift/blob/main/support/globalconfig/infrastructure.go#L44

everettraven: Keeping the existing logic flow means we check for CPO deployment and then skip?

dinhxuanvu: Yes. The current flow is to check CAO first and then check topology (and skipped if external). I keep it the same and just add a check for CPO instead CAO in case it is HCP.

everettraven: Seems pointless to add a check there then, no?

If we will always end up skipping it seems like we are just wasting cycles on performing that logic.

dinhxuanvu: I'm fine with moving the topology check to the top and skip the whole test.

everettraven: Yeah, I think that sounds more reasonable in this case. Thanks!

dinhxuanvu: Done

dinhxuanvu:
/verified by periodic-ci-openshift-hypershift-release-4.20-periodics-e2e-aws-ovn-conformance-serial

melvinjoseph86:
@alebedev87 based on https://github.com/openshift/release/pull/75813, do we need cherry pick to 4.19 release also?

alebedev87:
@melvinjoseph86 : yes. Strangely, [the cherry-pick from 4.20 PR doesn't work](https://github.com/openshift/origin/pull/30848#issuecomment-4066246765).

PR #30210, opened 2025-09-04
OCPBUGS-61194: Mark etcd net overload logging test as a flake

dgoodwin:
This is failing too often on gcp and aws jobs where nothing else failed. Informational only, so we'll leave it a flake for now.


dgoodwin:
/verified later @dgoodwin

This change is inconsequential and triggering a flake is difficult, we need to merge and let soak to ensure it works, but the change is extremely simple and should be no problem.

PR #30222, opened 2025-09-08
NO-JIRA: Add intervals and a test for systemd-coredumps

dgoodwin:
In https://issues.redhat.com/browse/OCPBUGS-61224 we found haproxy is core dumping right before we get disruption. Scanning journals for systemd-coredump lines seems like a good idea. 

Flake only for now, who knows what this might pick up initially.

Assisted-by: Cursor AI

### DIFF
--- a/pkg/monitor/monitorapi/types.go
+++ b/pkg/monitor/monitorapi/types.go
@@ -272,7 +272,8 @@ const (
 	ReasonHighGeneration    IntervalReason = "HighGeneration"
 	ReasonInvalidGeneration IntervalReason = "GenerationViolation"
 
-	ReasonEtcdBootstrap IntervalReason = "EtcdBootstrap"
+	ReasonEtcdBootstrap     IntervalReason = "EtcdBootstrap"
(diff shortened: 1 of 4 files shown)


pkg/monitor/monitorapi/types.go
```diff
@@ -272,7 +272,8 @@ const (
 	ReasonHighGeneration    IntervalReason = "HighGeneration"
 	ReasonInvalidGeneration IntervalReason = "GenerationViolation"
 
-	ReasonEtcdBootstrap IntervalReason = "EtcdBootstrap"
+	ReasonEtcdBootstrap     IntervalReason = "EtcdBootstrap"
+	ReasonProcessDumpedCore                = "ProcessDumpedCore"
```
smg247:
```suggestion
	ReasonProcessDumpedCore         IntervalReason = "ProcessDumpedCore"
```

dgoodwin:
Whoops, compile issue in the test after that constant change.

PR #30227, opened 2025-09-09
OCPBUGS-60507: Bump to latest openshift/kubernetes (v1.33.4)

bertinatto:
/assign @kevinrizza


sdodson:
/verified by the fact that we won't ship 4.21 on 1.33 so this won't matter

PR #30234, opened 2025-09-10
OCPBUGS-61196: Add test that the ServiceCIDR API is blocked

danwinship:
(Test should fail until https://github.com/openshift/cluster-network-operator/pull/2605 merges.)

### DIFF
--- a/test/extended/networking/servicecidr.go
+++ b/test/extended/networking/servicecidr.go
@@ -1,0 +1,38 @@
+package networking
+
+import (
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+
+	exutil "github.com/openshift/origin/test/extended/util"
+)
+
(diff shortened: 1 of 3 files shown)


kyrtapz:
/lgtm
Lets retest after https://github.com/openshift/cluster-network-operator/pull/2605 gets in  
/hold

danwinship:
/assign @bertinatto 

(the e2e test to confirm that the ServiceCIDR-blocking VAP actually works. needs a top-level approve)

bertinatto (commented):
/approve

danwinship:
/verified by @danwinship

@kyrtapz can I get a re-lgtm? (I had to re-push for needs-rebase, and then because it turns out this won't pass on microshift because the VAP doesn't get installed there (which we'll probably want to deal with later).)

danwinship:
/hold cancel
failures are in permafailing jobs

PR #30239, opened 2025-09-11
MULTIARCH-5391: Add Disruptive ImageStreamImportMode Tests into a Disruptive Suite

jatinsu:
This PR is to move certain disruptive tests introduced in [MULTIARCH-5318](https://github.com/openshift/origin/pull/29524) into it's own disruptive suite


test/extended/images/distruptiveimportmode.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/openshift/library-go/pkg/image/imageutil"
+	exutil "github.com/openshift/origin/test/extended/util"
+	"github.com/openshift/origin/test/extended/util/operator"
+)
+
+var _ = g.Describe("[sig-imageregistry][Suite:openshift/test-image-stream-import-mode][OCPFeatureGate:ImageStreamImportMode][Serial][Disruptive] ImageStream API", func() {
+	defer g.GinkgoRecover()
+	oc := exutil.NewCLI("imagestream-api")
+
+	g.It("import mode should be PreserveOriginal or Legacy depending on desired.architecture field in the CV [apigroup:image.openshift.io]", func() {
```
Prashanth684: this test is not disruptive - it does not trigger an apiserver rollout so it can be kept in serial

zz_generated.manifests/test-reporting.yaml
```diff
@@ -461,9 +461,9 @@ spec:
         pods are running on the Arbiter node'
   - featureGate: ImageStreamImportMode
     tests:
-    - testName: '[sig-cluster-lifecycle][OCPFeatureGate:ImageStreamImportMode] ClusterVersion
-        API desired architecture should be valid when architecture is set in release
-        payload metadata [apigroup:config.openshift.io]'
+    - testName: '[sig-cluster-lifecycle][Suite:openshift/test-image-stream-import-mode][OCPFeatureGate:ImageStreamImportMode][Disruptive][Serial]
+        ClusterVersion API desired architecture should be valid when architecture
```
Prashanth684: did you change these manually? these are changed by running the make update. This should not be disruptive

zz_generated.manifests/test-reporting.yaml
```diff
@@ -473,6 +473,15 @@ spec:
...
         ImageStream API import mode should be PreserveOriginal or Legacy depending
         on desired.architecture field in the CV [apigroup:image.openshift.io]'
+    - testName: '[sig-imageregistry][Suite:openshift/test-image-stream-import-mode][OCPFeatureGate:ImageStreamImportMode][Serial][Disruptive]
+        ImageStream API import mode should be Legacy if the import mode specified
+        in image.config.openshift.io config is Legacy [apigroup:image.openshift.io]'
+    - testName: '[sig-imageregistry][Suite:openshift/test-image-stream-import-mode][OCPFeatureGate:ImageStreamImportMode][Serial][Disruptive]
+        ImageStream API import mode should be PreserveOriginal if the import mode
+        specified in image.config.openshift.io config is PreserveOriginal [apigroup:image.openshift.io]'
+    - testName: '[sig-imageregistry][Suite:openshift/test-image-stream-import-mode][OCPFeatureGate:ImageStreamImportMode][Serial][Disruptive]
+        ImageStream API import mode should be PreserveOriginal or Legacy depending
```
Prashanth684: this should also not be disruptive

zz_generated.manifests/test-reporting.yaml
jatinsu: No, I ran `make update`. I think what happened is when I made my last changes I forgot to run it again.

test/extended/images/distruptiveimportmode.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func TestImageConfigImageStreamImportModeLegacy(t g.GinkgoTInterface, oc *exutil.CLI) {
+	ctx := context.Background()
+	if !exutil.IsTechPreviewNoUpgrade(ctx, oc.AdminConfigClient()) {
+		g.Skip("skipping, this feature is only supported on TechPreviewNoUpgrade clusters")
+	}
+	if isSNO, err := exutil.IsSingleNode(ctx, oc.AdminConfigClient()); err == nil && isSNO {
+		g.Skip("skipping this test for SNO as it involves an openshift-apiserver disruption")
+	}
```
everettraven: If we are going to support this feature on SNO clusters, we should ensure it is thoroughly tested on SNO clusters as well.

Since you are now using a test suite marked as disruptive, I'd expect to see this removed so that all your tests run on every supported platform.

test/extended/images/distruptiveimportmode.go
```diff
@@ -0,0 +1,164 @@
...
+	t.Log("image controller configuration propagated")
+	return nil
+}
+
+func TestImageConfigImageStreamImportModeLegacy(t g.GinkgoTInterface, oc *exutil.CLI) {
+	ctx := context.Background()
+	if !exutil.IsTechPreviewNoUpgrade(ctx, oc.AdminConfigClient()) {
+		g.Skip("skipping, this feature is only supported on TechPreviewNoUpgrade clusters")
+	}
```
everettraven: I don't think you need this gate.

Using the `[OCPFeatureGate: ...]` annotation should automatically exclude this test from the set of tests to run if the feature gate is not enabled.

Presumably, it is currently only enabled in TPNU clusters?

Prashanth684: correct. this can be removed

test/extended/images/distruptiveimportmode.go
Prashanth684: Yes we need to remove this as well - right now we can't test this on SNO because it's a separate suite, but once we define jobs for this suite we can test this

everettraven: Not having a job for it doesn't mean you can't remove this - using a new suite means that no other suite should run these tests.

That being said, it definitely is a bit of a painful cycle to add the test suite and then try to create jobs in openshift/release to run them without being able to test the new suite in CI.

I think there are a few things you can do to test the changes in this pull request:
- Create a manually triggered only presubmit job that runs this suite in openshift/origin and get it merged (rehearsals will fail, thats okay if it is only ever manually triggered and temporary). Run that presubmit against this PR
- Merge this implemented how you want and then use rehearsals on your openshift/release PR to test. This is a _slow_ feedback loop but one that you  _can_ do if you'd like
- Build a payload image from this PR and use gangway-cli to trigger runs using this payload image, injecting the necessary arguments to the openshift-tests binary to run the suite. If you need more information on gangway-cli, let me know and I can try to wrangle some resources together to share.

jatinsu: Sorry for the delay, I ran a presubmit job against this PR and [it passed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/30239/pull-ci-openshift-origin-main-e2e-test-image-stream-import-mode-techpreview/1973252036904357888). The reason why it's not ran against the most recent commit is because I accidentally named the test suite in the [release repo](https://github.com/openshift/release/blob/660a6be6268dcf674f3feb376ecaf94f661bcaec/ci-operator/config/openshift/origin/openshift-origin-main.yaml#L287) `test-image-stream-import-mode-techpreview`, so I had to change the name of the test suite in this pr to run the presubmit job :melting_face:

jatinsu:
@everettraven @neisw Would you mind taking a look at this? I ran the presubmit job against these changes and it [passed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/30239/pull-ci-openshift-origin-main-e2e-test-image-stream-import-mode-techpreview/1973252036904357888). The reason why it's not ran against the most recent commit is because I accidentally named the test suite in the [release repo](https://github.com/openshift/release/blob/660a6be6268dcf674f3feb376ecaf94f661bcaec/ci-operator/config/openshift/origin/openshift-origin-main.yaml#L287) `test-image-stream-import-mode-techpreview`, so I had to alter the name of the test suite in this pr to run the presubmit job and then change it back :melting_face:

zz_generated.manifests/test-reporting.yaml
everettraven, line 1: Not sure I would have expected this to be a new file?

Looks like maybe this needs a rebase based on https://github.com/openshift/origin/commit/4d9b126311b570f2360dc182af51160418dd8edf#diff-eaa77465a08bec46c1a6aa46b98ac7de4933436695637240cfdcc28f5cfaf2a0 ?

everettraven:
Aside from some weirdness with generated files (looks like this needs a rebase?) the general shape looks right to me.

zz_generated.manifests/test-reporting.yaml
jatinsu: I think it's because this PR predates that commit, so that change was never incorporated. I just rebased this PR, thanks for letting me know!

test/extended/util/annotate/generated/zz_generated.annotations.go
everettraven, line 1: This also seems off for the same reason as my last comment - unless I'm missing something?

jatinsu: No you are correct, I just didn't notice that these files were deleted, sorry about that.

PR #30265, opened 2025-09-17
OCPBUGS-61855: CRD Apply Alert Allowance

dtfranz:
Adds an allowed alert for `InstallPlanStepAppliedWithWarnings` to unblock kubernetes 1.34 rebase efforts. The servicemeshoperator VirtualServices CRD is generating warnings from the api server because 1.34 now gives warnings for unknown string formats. Without this exception the rebase efforts may be blocked until the CRD can be fixed. This allowance will be removed once the issue has been resolved.

I don't yet know if the upgrade tests will need this as well, but I'm limiting the scope here for the time being.


dgoodwin:
How can we ensure the exception doesn't get forgotten like virtually all of these do. :) 

Could someone file two release blocker bugs, one for the underlying issue, one dependant for the removal of exception. That would probably be our best bet at not missing this.

PR #30267, opened 2025-09-17
CNF-18661: update the commatrix vendor version to contain new changes like adding new logs

aabughosh:
update the commatrix vendor version to contain new changes like adding new logs


mgencur:
It would be nice to update to the latest version to get changes from https://github.com/openshift-kni/commatrix/pull/213
Thanks.

PR #30288, opened 2025-09-22
MCO-1875: Bump timeout for vSphereBootimageUpdate tests

RishabhSaini:
helpers: vSphere 2x upload is time consuming hence bump timeout


yuqi-zhang:
Since this is a timeout bump, probably best verified via followup CI runs.

/verified later @RishabhSaini

